PR #124, opened 2023-01-30
add citation

natolambert:
I'm working on an RLHF tutorial, and will cite tools for it, so it would be good to have.

Feel free @lvwerra and @younesbelkada to propose an author order. I just threw one together.


lvwerra (approved):

younesbelkada (approved):
Thanks for adding this!

PR #127, opened 2023-02-01
add `set_seed` to __init__.py

lvwerra:
This adds `set_seed` to the main `__init__.py` so we can `from trl import set_seed`. I think we should move important stuff the users is supposed to use there so they don't have to remember paths. Wdyt @younesbelkada? 

cc @natolambert


younesbelkada (approved):
💯  on this, thank you @lvwerra !

natolambert:
Yep, sorry I missed the point on the main init. I see now :).

PR #128, opened 2023-02-01
update docs with Seq2seq models, set_seed, and create_reference_model

lvwerra:
This PR is just a small update to the docs.


younesbelkada (approved):
Thanks for taking care of updating the documentation 💯

PR #132, opened 2023-02-06
[`bug`] Update gpt2-sentiment.py

younesbelkada:
This PR updates a small bug that has been introduced in https://github.com/lvwerra/trl/pull/118 

cc @lvwerra


lvwerra (approved):
Good catch! It's not in the notebook or other scripts?

lvwerra:
It's actually the same in the T5 script - mind fixing it too?

younesbelkada:
Thanks! Added it on t5 script and checked that this was not used on the notebooks

PR #139, opened 2023-02-08
fix forward batching for seq2seq and right padding models.

lvwerra:
This PR fixes issues with forward batching of models that pad to the right (e.g. `bloom`) and seq2seq models (e.g. `T5`).

### DIFF
--- a/examples/sentiment/scripts/t5-sentiment.py
+++ b/examples/sentiment/scripts/t5-sentiment.py
@@ -42,7 +42,7 @@ from trl.core import LengthSampler
 # We first define the configuration of the experiment, defining the model, the dataset,
 # the training parameters, and the PPO parameters.
 # Check the default arguments in the `PPOConfig` class for more details.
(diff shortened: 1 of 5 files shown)


younesbelkada (approved):
This looks great to me! 💯 
[optional] My only comment:
- We can maybe not rely on `self.datacollator` and create the `attention_mask` ourselves (for e.g. `attention_mask = q.ne(self.tokenizer.pad_token_id) * 1.0` ) ? This way if a user wants to build something on top of that can directly see where the correct `attention_mask` is created. i.e. passing a full-ones to `data_collator` to retrieve the correct one is IMO not very intuitive - maybe at least worth commenting it

lvwerra:
Both runs seem to be converging normally:
GPT2: https://wandb.ai/lvwerra/trl/runs/9ufqpv3w
T5: https://wandb.ai/lvwerra/trl/runs/4rphjty6

PR #144, opened 2023-02-11
fix bug in batched_forward_pass

ArvinZhuang:
fix the bug that will cause the devices not match issue in the `batched_forward_pass` method.

The reason:
`self. data_collator ` returned tensors will be on CPU, thus the later `self.model(**input_kwargs)` will give error as model is on GPU.

The proposed solution:
do `.to(self.accelerator.device)` after `self.data_collator`

My transformers version: 4.26.1

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -568,10 +568,7 @@ class PPOTrainer(BaseTrainer):
         else:
             input_ids = [torch.cat([q, r]) for q, r in zip(queries, responses)]
             input_data = self.data_collator(
-                [
-                    {"input_ids": ids, "attention_mask": torch.ones_like(ids)}
-                    for ids in input_ids
-                ]
+                [{"input_ids": ids, "attention_mask": torch.ones_like(ids)} for ids in input_ids]
             ).to(self.accelerator.device)
 
         for i in range(int(bs / fbs)):
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -553,13 +553,12 @@ class PPOTrainer(BaseTrainer):
         # automatically padded by the datacollator to indicate padding tokens
         if self.is_encoder_decoder:
             input_data = self.data_collator(
-                [{"input_ids": q, "attention_mask": torch.ones_like(q).to(self.accelerator.device)} for q in queries]
-            )
+                [{"input_ids": q, "attention_mask": torch.ones_like(q)} for q in queries]
+            ).to(self.accelerator.device)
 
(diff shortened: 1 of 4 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -553,13 +553,12 @@ def batched_forward_pass(
         # automatically padded by the datacollator to indicate padding tokens
         if self.is_encoder_decoder:
             input_data = self.data_collator(
-                [{"input_ids": q, "attention_mask": torch.ones_like(q).to(self.accelerator.device)} for q in queries]
-            )
+                [{"input_ids": q, "attention_mask": torch.ones_like(q)} for q in queries]
+            ).to(self.accelerator.device)
 
-            attention_mask = [torch.ones_like(r) for r in responses]
```
younesbelkada, line 559: Hum why this has been removed? 🤔

younesbelkada (commented):
Wow thanks a lot for fixing the bug! 
I am ok with this fix in the principle that it is a safety checker to make sure all the returned tensors will be on the correct device (regardless where the dataloader will send the device)
Let's run the tests and see! 
Wdyt @lvwerra ?

trl/trainer/ppo_trainer.py
ArvinZhuang: because seems attention_mask variable is never used

younesbelkada: Perfect, can you run the styling checks so that the testing suite will be executed?

ArvinZhuang: just did

younesbelkada (commented):
Thanks for fixing! Could you revert your changes in all the scripts inside `examples/` after that we should be good to merge!

ArvinZhuang:
> Thanks for fixing! Could you revert your changes in all the scripts inside `examples/` after that we should be good to merge!
Hi 
these changes are automatically generated by style && quality, but I reverted them anyway

younesbelkada (approved):
Thanks for iterating! 🚀

ArvinZhuang:
Also I notice this `"Forward batch size > 1 is not well supported yet for encoder-decoder models`.
Could you please give me more hits why this is the case? Looks like forward batch size =1 will greatly slow down the inference, I probably can give it a shot to fix this.

Rebecca-Qian:
Had just prepared a PR to fix this 😄 thanks @ArvinZhuang

ArvinZhuang:
> Had just prepared a PR to fix this 😄 thanks @ArvinZhuang
@Rebecca-Qian You are welcome :) Im glad to contribute as well

lvwerra (approved):
Thanks for the fix!

PR #147, opened 2023-02-15
[`core`] Add `torch_dtype` support

younesbelkada:
# What does this PR do?

Currently on the main branch, passing the `transformers` kwargs directly on `from_pretrained` is broken due to a slient bug. This PR fixes the bug and adds some tests, so that any user can benefit from args such as `torch_dtype` or `load_in_8bit` support

cc @lvwerra

### DIFF
--- a/tests/models/test_modeling_value_head.py
+++ b/tests/models/test_modeling_value_head.py
@@ -122,7 +122,7 @@ class VHeadModelTester:
             with tempfile.TemporaryDirectory() as tmp_dir:
                 model.save_pretrained(tmp_dir)
 
-                model_from_save = self.trl_model_class.from_pretrained(tmp_dir, max_shard_size="1MB")
+                model_from_save = self.trl_model_class.from_pretrained(tmp_dir)
 
             # Check if the weights are the same
             for key in model_from_save.state_dict():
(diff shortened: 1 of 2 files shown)


tests/models/test_modeling_value_head.py
```diff
@@ -384,6 +409,35 @@ def test_push_to_hub(self):
                     f"Parameter {name} is not the same after push_to_hub and from_pretrained",
                 )
 
+    def test_transformers_kwargs(self):
```
lvwerra, line 412: Is this test duplicate? the only difference seems to be the FSMT check

tests/models/test_modeling_value_head.py
```diff
@@ -264,6 +264,31 @@ def test_raise_error_not_causallm(self):
             pretrained_model = AutoModelForCausalLM.from_pretrained(model_id)
             _ = AutoModelForCausalLMWithValueHead.from_pretrained(pretrained_model.transformer)
 
+    def test_transformers_kwargs(self):
```
lvwerra, line 267: should we name it:
```suggestion
    def test_transformers_bf16_kwargs(self):
```

tests/models/test_modeling_value_head.py
younesbelkada: Yeah it's just that for seq2seq models `fmst` does not have any `lm_head` and as I suppose that this model is not very much used (by looking at the number of downloads) I decided to skip it.

lvwerra: ah ok, can you also name it ? then we can merge
```suggestion
    def test_transformers_bf16_kwargs(self):
```

younesbelkada: Yes sure!

lvwerra (approved):

PR #152, opened 2023-02-16
[Doc] Add how to use Lion optimizer

younesbelkada:
# What does this PR do?

This PR adds LION optimizer from: https://arxiv.org/abs/2302.06675 / code source is: https://github.com/google/automl/blob/master/lion/lion_pytorch.py

For now it does not really seem to improve convergence in `gpt-sentiment`, but thought it would be nice to showcase that it can be applied for `trl` and used out of the box

EDIT: it seems to converge nicely after dividing the learning rate by 3, as suggested by the paper

cc @lvwerra @kashif

I think this is a nice-to-have artifact in this lib, as it reduces quite nicely the memory footprint of training (got 10% reduction)


docs/source/customization.mdx
```diff
@@ -54,6 +54,22 @@ optimizer = bnb.optim.Adam8bit(model.parameters(), lr=config.learning_rate)
 ppo_trainer = PPOTrainer(config, model, model_ref, tokenizer, optimizer=optimizer)
 ```
 
+### Use LION optimizer
+
+You can use the new LION optimizer from Google as well, first take the source code of the optimizer definition [here](https://github.com/lucidrains/lion-pytorch/blob/main/lion_pytorch/lion_pytorch.py), and copy it so that you can import the optimizer. Make sure to initialize the optimizer by considering the trainable parameters only for a more memory efficient training:
```
lvwerra, line 59: Can you add a link to the paper/blog post in the first reference of "Lion optimizer from Google"?

docs/source/customization.mdx
```diff
@@ -54,6 +54,22 @@ optimizer = bnb.optim.Adam8bit(model.parameters(), lr=config.learning_rate)
...
+```python
+optimizer = Lion(filter(lambda p: p.requires_grad, self.model.parameters()), lr=self.config.learning_rate)
+
+...
+ppo_trainer = PPOTrainer(config, model, model_ref, tokenizer, optimizer=optimizer)
+```
+We advice you to use the learning rate that you would use for `Adam` divided by 3 as pointed out [here](https://github.com/lucidrains/lion-pytorch#lion---pytorch). We observed an improvement when using this optimizer compared to classic Adam, check the full logs [here](https://wandb.ai/distill-bloom/trl/runs/lj4bheke?workspace=user-younesbelkada):
```
lvwerra, line 66:
```suggestion
We advice you to use the learning rate that you would use for `Adam` divided by 3 as pointed out [here](https://github.com/lucidrains/lion-pytorch#lion---pytorch). We observed an improvement when using this optimizer compared to classic Adam (check the full logs [here](https://wandb.ai/distill-bloom/trl/runs/lj4bheke?workspace=user-younesbelkada)):
```

lvwerra (approved):
Just two minor comments, otherwise looks good! ❤️

docs/source/customization.mdx
younesbelkada, line 59:
```suggestion
You can use the new [LION optimizer from Google](https://arxiv.org/abs/2302.06675) as well, first take the source code of the optimizer definition [here](https://github.com/lucidrains/lion-pytorch/blob/main/lion_pytorch/lion_pytorch.py), and copy it so that you can import the optimizer. Make sure to initialize the optimizer by considering the trainable parameters only for a more memory efficient training:
```

younesbelkada: Sure!

PR #157, opened 2023-02-16
[`core`] fix saving multi-gpu

younesbelkada:
Currently on the `main` branch of `trl` saving a `PPOTrainer` in a multi-GPU setting is broken. The fix is to first call `self.accelerator.unwrap_model(self.model)` before calling `save_pretrained`

cc @lvwerra

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -906,6 +906,7 @@ class PPOTrainer(BaseTrainer):
             f.write(model_card_content)
 
     def _save_pretrained(self, save_directory: str) -> None:
+        self.accelerator.unwrap_model(self.model).save_pretrained(save_directory)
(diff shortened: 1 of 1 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -906,6 +906,7 @@ def create_model_card(self, path: str, model_name: Optional[str] = "TRL Model")
             f.write(model_card_content)
 
     def _save_pretrained(self, save_directory: str) -> None:
+        self.accelerator.unwrap_model(self.model).save_pretrained(save_directory)
         self.model.save_pretrained(save_directory)
```
lvwerra, line 910: We should remove the following line then, right?
```suggestion
```

younesbelkada: yes

lvwerra (approved):

PR #158, opened 2023-02-16
Added pip install and wandb init

### LINKED ISSUE #150: Slight issues in the example notebook
I noticed some minor things in the example colab notebook. 

Installs where not properly setup and you needed to initialise wandb.
I made a notebook with the necessary changes if it would be useful.
https://colab.research.google.com/github/lvwerra/trl/blob/main/examples/sentiment/notebooks/gpt2-sentiment.ipynb

It was basically one cell with

`!pip install transformers trl wandb`

and one with 

```
import wandb
wandb.init()
```

BirgerMoell:
Added a cell for installing the packages and one for init for wandb.
I noticed now that some outputs was included since I ran the script. Is that ok?
Otherwise I can probably figure out how to remove it.

Fixes #150

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -907,6 +907,6 @@ class PPOTrainer(BaseTrainer):
             f.write(model_card_content)
 
     def _save_pretrained(self, save_directory: str) -> None:
-        self.model.save_pretrained(save_directory)
+        self.accelerator.unwrap_model(self.model).save_pretrained(save_directory)
         self.tokenizer.save_pretrained(save_directory)
         self.create_model_card(save_directory)


younesbelkada (approved):
Thanks for fixing! 
For me this looks very good, better if you clean up the output logs indeed but ok for me to merge as it is

younesbelkada:
Arghf I think one of the commits removed the stdout of the cells that are below that are needed, could you revert that? 🙏

BirgerMoell:
I used the package nbstripout to remove the output logs. This did however remove all the output logs on commit. 

Here is how to install it. Then it just works on commit.
nbstripout --install

I hope this is better. Otherwise we can maybe just commit the previous version that kept your logs intact.

younesbelkada:
Yeah I think keeping the previous commit is better here!

BirgerMoell:
I made a new commit instead that seems clean. Was a bit messy to clean this up. 
That one should just be the changes without any added output logs. 
https://github.com/lvwerra/trl/pull/159

PR #162, opened 2023-02-17
Add toxcitiy example

younesbelkada:
# What does this PR do?

This PR adds the toxicity example and replaces #142 

TODO:

- update docs with tips

cc @lvwerra


docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+We decided to use a context window of 10 to 15 tokens for the training of the larger models, as we think the results will be more impressive if the model is able to generate toxic prompts even with slightly long contexts.
+
+
+<div style="text-align: center">
+<img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/trl-long-vs-short-context.png">
+</div>
+
+### How to deal with OOM issues
+
+Oour goal was to train models up to 6B parameters, which is about 24GB in float32! Here are few tricks we have used to be able to train the largest model on a single 40GB-RAM GPU:
```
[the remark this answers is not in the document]
lvwerra: In general I would write the text not in past tense: write it as if you were doing the experiment right now.

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+Also we have observed that training with larger `mini_batch_size` leads to smoother convergence and better results on the test set:
+
+<div style="text-align: center">
+<img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/trl-gpt-j-mbs-run.png">
+</div>
+
+## Results
+
+We tested our models on a new dataset, the `OxAISH-AL-LLM/wiki_toxic` dataset that can be found [here](https://huggingface.co/datasets/OxAISH-AL-LLM/wiki_toxic). We feed to each model a toxic prompt from it (a sample with the label "toxic"), and generate 30 new tokens as it is done on the training loop and measure the toxicity score using `evaluate`'s [`toxicity` metric](https://huggingface.co/spaces/ybelkada/toxicity).
```
lvwerra, line 136:
```suggestion
We tested our models on a new dataset, the [`OxAISH-AL-LLM/wiki_toxic`](https://huggingface.co/datasets/OxAISH-AL-LLM/wiki_toxic) dataset. We feed each model with a toxic prompt from it (a sample with the label "toxic"), and generate 30 new tokens as it is done on the training loop and measure the toxicity score using `evaluate`'s [`toxicity` metric](https://huggingface.co/spaces/ybelkada/toxicity).
```

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+| `ybelkada/gpt-j-6b-detoxified-20shdl-4mbs` | **0.1510** | **0.2798** |
+
+<div class="column" style="text-align:center">
+  <div class="column">
+    <img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/trl-toxicity-final-plot.png" style="width:50%">
+  </div>
+  <div class="column">
+    <img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/trl-toxicity-without-std.png" style="width:50%">
+  </div>
+</div>
```
lvwerra, line 157: I would probably show the one without std as the std is so large :) Alternatively you could show the distribution in a violin plot.

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+The evaluation script can be found [here](https://gist.github.com/younesbelkada/781b862ffbe1233a1c117a07716d4711).
```
lvwerra, line 159: What about adding it to the script folder?

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+
```
lvwerra, line 160: I think a few examples would also be useful. E.g. you could do something similar to the table here:
https://github.com/lvwerra/trl#advanced-example-imdb-sentiment

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+### Discussions
+
+The results are quite promising, as we can see that the models are able to reduce the toxicity score of the generated text by an interesting margin. The gap is impressive for `gpt-neo-2B` model but we hoped to see a larger gap for the `gpt-j-6B` model. We believe that many tricks could solve these issues at scale, starting with training with larger `mini_batch_size` and probably allowing to back-propagate through more layers (i.e. use less shared layers).
```
lvwerra, line 163:
```suggestion
The results are quite promising, as we can see that the models are able to reduce the toxicity score of the generated text by an interesting margin. The gap is clear for `gpt-neo-2B` model but we less so for the `gpt-j-6B` model. There are several things we could try to improve the results on the largest model starting with training with larger `mini_batch_size` and probably allowing to back-propagate through more layers (i.e. use less shared layers).
```

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+We also think we could have trained the models on a "more toxic" dataset as the one we used is much cleaner than the dataset we used for testing our models (from our observation). 
+A hypothesis we made is that larger models tends to be more toxic. Therefore, one could have also played with the KL-penalty term, to allow the model to deviate a bit more from its original distribution. We also believe that fine-tuning a model that is known to be toxic (i.e. trained on a toxic dataset) could also lead to better results.
+We have also observed that training the model with larger context helps getting better results for larger models. Therefore one could have also played with this factor for the larger model and produce a better model.
```
lvwerra, line 166: I feel like there is a lot of uncertainty in those statements. I would maybe go for something like:

In addition to human feedback this could be a useful additional signal when training large language models to ensure there outputs are less toxic as well as useful.

lvwerra (commented):
Hi @younesbelkada did a first pass over the text and looks good. Left a few comments and suggestions.

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,169 @@
...
+```
+
+### Reward function
+
+
+The reward function is one of the most important part of training a model with reinforcement learning. It is the function that will tell the model if it is doing well or not.
+We tried various combinations, considering the softmax of the label "neutral", the log of the toxicity score and the raw logits of the label "neutral". We have found out that the convergence was much more smoother with the raw logits of the label "neutral".
+```python
+logits = toxicity_model(**toxicity_inputs).logits.float()
+toxicity_labels = (logits[:, 0]).tolist()
```
younesbelkada, line 62:
```suggestion
rewards = (logits[:, 0]).tolist()
```

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,169 @@
...
+```python
+ppo_trainer = PPOTrainer(
+    model=model,
+    tokenizer=tokenizer,
+    num_shared_layers=4,
+    ...
+)
+```
+
+- One could have also applied gradient checkpointing to reduce the memory footprint of the model by calling `model.pretrained_model.enable_gradient_checkpointing()`.
```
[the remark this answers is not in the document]
younesbelkada: Yeah we can do that in a follow up PR!

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+| [`gpt-j-6b-toxicity.py`](https://github.com/lvwerra/trl/blob/main/examples/toxicity/scripts/gpt-j-6b-toxicity.py) | Detoxify `GPT-J-6B` using PPO | x | 
+
+## Context
+
+A Language Model is not initially trained to be toxic, but it can be trained to be toxic by feeding it toxic prompts. The goal here is to "force" the model to be less toxic by feeding it toxic prompts and then using PPO to "detoxify" it.
+We believe that if this technique would be applied to language models that are meant to be toxic (e.g. GPT-4chan) the results would be even more impressive.
+
+### Computing toxicity scores
+
+We used [`facebook/roberta-hate-speech-dynabench-r4-target`](https://huggingface.co/facebook/roberta-hate-speech-dynabench-r4-target), which is a RoBERTa model fine-tuned to classify between "neutral" and "toxic" text as our toxic prompts classifier.
```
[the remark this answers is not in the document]
younesbelkada: Thanks for the suggestion! Added it acconrdingly

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+| `gpt2` | 0.01602 |
+| `facebook/opt-350m` | 0.01628 |
+| `bigscience/bloom-560m` | 0.00767 |
+| `EleutherAI/gpt-neo-125M` | **0.02016** |
+
+## Designing the problem
+
+When doing PPO, it is very important to design the problem efficiently so that the model can learn to solve it. Let's cover the topics that were important for the model to converge.
+
+### Pre-processing the dataset
```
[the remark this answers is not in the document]
younesbelkada: Makes sense! Added an example and a brief explanation of the dataset

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+<div style="text-align: center">
+<img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/trl-long-vs-short-context.png">
+</div>
+
+### How to deal with OOM issues
+
+Oour goal was to train models up to 6B parameters, which is about 24GB in float32! Here are few tricks we have used to be able to train the largest model on a single 40GB-RAM GPU:
+
+- Use `bfloat16` precision: Simply load your model in `bfloat16` when calling `from_pretrained` and you can reduce the size of the model by 2:
```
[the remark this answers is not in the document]
younesbelkada: Yes I think it is different from mixed precision, here I meant training the model in full `bfloat16` - maybe I can add few sentences to clarify this

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+```python
+model = AutoModelForCausalLM.from_pretrained("EleutherAI/gpt-j-6B", torch_dtype=torch.bfloat16)
+```
+
+- Use shared layers: Since PPO algorithm requires to have both the active and reference model to be on the same device, we have decided to use shared layers to reduce the memory footprint of the model. This can be achieved by just speifying `num_shared_layers` argument when creating a `PPOTrainer`:
```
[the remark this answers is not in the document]
younesbelkada: Yes makes sense!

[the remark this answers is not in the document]
younesbelkada: Isn't the other way around? We don't train the first 4 layers and train the rest

docs/source/detoxifying_a_lm.mdx
younesbelkada: Added it just above! :)

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+The results are quite promising, as we can see that the models are able to reduce the toxicity score of the generated text by an interesting margin. The gap is clear for `gpt-neo-2B` model but we less so for the `gpt-j-6B` model. There are several things we could try to improve the results on the largest model starting with training with larger `mini_batch_size` and probably allowing to back-propagate through more layers (i.e. use less shared layers).
+We also think we could have trained the models on a "more toxic" dataset as the one we used is much cleaner than the dataset we used for testing our models (from our observation). 
+A hypothesis we made is that larger models tends to be more toxic. Therefore, one could have also played with the KL-penalty term, to allow the model to deviate a bit more from its original distribution. We also believe that fine-tuning a model that is known to be toxic (i.e. trained on a toxic dataset) could also lead to better results.
+We have also observed that training the model with larger context helps getting better results for larger models. Therefore one could have also played with this factor for the larger model and produce a better model.
```
younesbelkada, line 183:
```suggestion

To sum up, in addition to human feedback this could be a useful additional signal when training large language models to ensure there outputs are less toxic as well as useful.
```

docs/source/detoxifying_a_lm.mdx
younesbelkada: Thanks for the feedback! Proposed something below

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,169 @@
...
+## Training the model!
+
+We have trained 6 models in total:
+
+- [`ybelkada/gpt-neo-125m-detoxified-small-context`](https://huggingface.co/ybelkada/gpt-neo-125m-detoxified-small-context)
+- [`ybelkada/gpt-neo-125m-detoxified-long-context`](https://huggingface.co/ybelkada/gpt-neo-125m-detoxified-long-context)
+- [`ybelkada/gpt-neo-2B-detoxified-20sh-30steps`](https://huggingface.co/ybelkada/gpt-neo-2B-detoxified-20sh-30steps)
+- [`ybelkada/gpt-neo-2.7B-detoxified-20shdl`](https://huggingface.co/ybelkada/gpt-neo-2.7B-detoxified-20shdl)
+- [`ybelkada/gpt-j-6b-detoxified-20shdl-4mbs`](https://huggingface.co/ybelkada/gpt-j-6b-detoxified-20shdl-4mbs)
+- [`ybelkada/gpt-j-6b-detoxified-1000-20shdl`](https://huggingface.co/ybelkada/gpt-j-6b-detoxified-1000-20shdl)
```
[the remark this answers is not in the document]
younesbelkada: Yes this is much cleaner, I have modified the model names and will update the model cards accordingly

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,169 @@
+# Detoxifying a LM using PPO
+
+Let us check step by step on how to detoxify a Language Model using PPO, using a dataset of toxic prompts!
```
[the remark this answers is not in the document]
younesbelkada: Thanks, fixed!

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,170 @@
...
+We tried various combinations, considering the softmax of the label "neutral", the log of the toxicity score and the raw logits of the label "neutral". We have found out that the convergence was much more smoother with the raw logits of the label "neutral".
+```python
+logits = toxicity_model(**toxicity_inputs).logits.float()
+toxicity_labels = (logits[:, 0]).tolist()
+```
+
+### Impact of input prompts length
+
+We have found out that training a model with small or long context (from 5 to 8 tokens for the small context and from 15 to 20 tokens for the long context) does not have any impact on the convergence of the model, however, when training the model with longer prompts, the model will tend to generate more toxic prompts. 
+We decided to use a context window of 10 to 15 tokens for the training of the larger models, as we think the results will be more impressive if the model is able to generate toxic prompts even with slightly long contexts.
```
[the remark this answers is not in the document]
younesbelkada: Ah yes you are right!

docs/source/detoxifying_a_lm.mdx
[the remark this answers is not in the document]
younesbelkada: Thanks, yes makes sense!

younesbelkada:
Thanks a mile for the extensive review @lvwerra ! Should have addressed them now and left few minor questions :)

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,189 @@
...
+- [`ybelkada/gpt-neo-2.7B-detox`](https://huggingface.co/ybelkada/gpt-neo-2.7B-detox)
+- [`ybelkada/gpt-j-6b-detox`](https://huggingface.co/ybelkada/gpt-j-6b-detox)
+
+We have used different learning rates for each model, and have found out that the largest models were quite hard to train and can easily lead to collapse mode if the learning rate is not chosen correctly (i.e. if the learning rate is too high):
+
+<div style="text-align: center">
+<img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/trl-collapse-mode.png">
+</div>
+
+The final training run of `ybelkada/gpt-j-6b-detoxified-1000-20shdl` looks like this:
```
lvwerra, line 134: Do you need to update the model name?

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,189 @@
...
+<div style="text-align: center">
+<img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/trl-toxicity-examples.png">
+</div>
+
+The evaluation script can be found [here](https://github.com/lvwerra/trl/blob/main/examples/toxicity/scripts/evaluate-toxicity.py).
+
+### Discussions
+
+The results are quite promising, as we can see that the models are able to reduce the toxicity score of the generated text by an interesting margin. The gap is clear for `gpt-neo-2B` model but we less so for the `gpt-j-6B` model. There are several things we could try to improve the results on the largest model starting with training with larger `mini_batch_size` and probably allowing to back-propagate through more layers (i.e. use less shared layers).
+We also think we could have trained the models on a "more toxic" dataset as the one we used is much cleaner than the dataset we used for testing our models (from our observation).
```
lvwerra, line 182:
```suggestion
We also think we could have trained the models using a "more toxic" dataset as the one we used is much cleaner than the dataset we used for testing our models (from our observation). 
```

docs/source/detoxifying_a_lm.mdx
```diff
@@ -0,0 +1,189 @@
...
+</div>
+
+The evaluation script can be found [here](https://github.com/lvwerra/trl/blob/main/examples/toxicity/scripts/evaluate-toxicity.py).
+
+### Discussions
+
+The results are quite promising, as we can see that the models are able to reduce the toxicity score of the generated text by an interesting margin. The gap is clear for `gpt-neo-2B` model but we less so for the `gpt-j-6B` model. There are several things we could try to improve the results on the largest model starting with training with larger `mini_batch_size` and probably allowing to back-propagate through more layers (i.e. use less shared layers).
+We also think we could have trained the models on a "more toxic" dataset as the one we used is much cleaner than the dataset we used for testing our models (from our observation). 
+A hypothesis we made is that larger models tends to be more toxic. Therefore, one could have also played with the KL-penalty term, to allow the model to deviate a bit more from its original distribution. We also believe that fine-tuning a model that is known to be toxic (i.e. trained on a toxic dataset) could also lead to better results.
```
lvwerra, line 183: I would still skip this paragraph. From your plot we can see that the large model is actually not more toxic.

lvwerra (approved):
A few minor comments, then we can merge :)

docs/source/detoxifying_a_lm.mdx
younesbelkada: Nice catch, updated the model name on the Hub and here as well

docs/source/detoxifying_a_lm.mdx
younesbelkada: Makes sense! Removed in [c40c250](https://github.com/lvwerra/trl/pull/162/commits/c40c250788a4abbcd69add62b6a6c36a98ca2ffd)

PR #163, opened 2023-02-17
`peft` integration

### LINKED ISSUE #145: Peft support
Hi, I was trying to train my model using  trl + LoRA(using Peft), but it looks like the two are incompatible for now.
Are there any plans for compatibility with LoRA(Peft)?
Thanks

edbeeching:
- adds command line args to the gpt2 sentiment example so we can  more easily benchmark different models.
- Add a new example using lora / peft `examples/sentiment/scripts/gpt2-sentiment_peft.py`
- some minor compatibility changes

Note: that gpt2 models use Conv1D layers instead of Linear, which is currently not supported in 8bit mode. Benchmarks of memory usage and convergence should be undertaken on some other models such as GPT-J

This should resolve issue #145 

Fixes #145


examples/sentiment/scripts/gpt2-sentiment_peft.py
```diff
@@ -0,0 +1,233 @@
...
+    task_type="CAUSAL_LM",
+    modules_to_save=["v_head"],
+)
+
+ref_model.config = (
+    ref_model.pretrained_model.config
+)  # We should probably change the AutoModelForCausalLMWithValueHead to have a config property
+ref_model.prepare_inputs_for_generation = (
+    ref_model.pretrained_model.prepare_inputs_for_generation
+)  # Again, this is required by LoRA, should we also add this method to AutoModelForCausalLMWithValueHead?
```
younesbelkada, line 175: Indeed better to upstream this in `PreTrainedModelWrapper`, we can also do it in a follow up PR

examples/sentiment/scripts/gpt2-sentiment_peft.py
```diff
@@ -0,0 +1,233 @@
...
+        texts = [q + r for q, r in zip(batch["query"], batch["response"])]
+        pipe_outputs = sentiment_pipe(texts, **sent_kwargs)
+        rewards = [torch.tensor(output[1]["score"]) for output in pipe_outputs]
+
+        #### Run PPO step
+        stats = ppo_trainer.step(query_tensors, response_tensors, rewards)
+    ppo_trainer.log_stats(stats, batch, rewards)
+
+
+# model.push_to_hub(f"{script_args.model_name}-ppo-sentiment")
```
younesbelkada, line 233:
```suggestion
model.push_to_hub(f"{script_args.model_name}-ppo-sentiment")
```

trl/models/modeling_value_head.py
```diff
@@ -156,11 +156,12 @@ def forward(
             kwargs (`dict`, `optional`):
                 Additional keyword arguments, that are passed to the wrapped model.
         """
+        kwargs["output_hidden_states"] = True  # this had already been set in the LORA / PEFT examples
         base_model_output = self.pretrained_model(
             input_ids=input_ids,
             past_key_values=past_key_values,
             attention_mask=attention_mask,
-            output_hidden_states=True,  # We force the model to output hidden states
+            # output_hidden_states=True,  # We force the model to output hidden states
```
younesbelkada, line 164: Great!

trl/trainer/ppo_trainer.py
```diff
@@ -20,6 +20,7 @@
 from typing import List, Optional, Union
 
 import datasets
+import peft
```
younesbelkada, line 23:
```suggestion
import datasets
from peft.peft_model import PeftModelForCausalLM
```

trl/trainer/ppo_trainer.py
```diff
@@ -171,7 +172,7 @@ def __init__(
             raise ValueError(
                 f"tokenizer must be a PreTrainedTokenizer or PreTrainedTokenizerFast, got {type(tokenizer)}"
             )
-        if not isinstance(model, PreTrainedModelWrapper):
+        if not isinstance(model, (PreTrainedModelWrapper, peft.peft_model.PeftModelForCausalLM)):
```
younesbelkada, line 175:
```suggestion
        if not isinstance(model, (PreTrainedModelWrapper, PeftModelForCausalLM)):
```

younesbelkada (commented):
Thanks a mile for working on this! Let's make `peft` + `trl` integration happen! 🔥 
I left few comments, and I have some open questions

- Should we make `peft` as a soft-dependency? We can do something similar than what we do in `transformers` : `if is_peft_available():`
- Should we add some tests? Just to make sure we train only the adapters

Let me know what do you think and if I can help on that!

trl/trainer/ppo_trainer.py
younesbelkada: Btw maybe the check can be better if we use `PeftModel` instead, so that it also works for seq2Seq models

trl/trainer/ppo_trainer.py
younesbelkada: We can also consider using `PeftModel` instead, as it will also contain the seq2seq models

edbeeching:
I just pushed a commit that resolves most of these changes.
Remaining questions / todo:

- For the upstreaming of properties / methods missing from PretrainedModelBase I added them as follows:
`        model.config = model.pretrained_model.config
        model.prepare_inputs_for_generation = model.pretrained_model.prepare_inputs_for_generation`
Alternatively, I could implement these actual methods?

- I agree with making peft a soft-dependency, but I have not implemented this yet.
- I agree with tests, but I have not implemented these yet.

I will implement the rest of the changes later in the week.

.github/workflows/tests.yml
```diff
@@ -45,3 +45,26 @@ jobs:
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        # cpu version of pytorch
+        pip install .[test]
+        pip install git+https://github.com/huggingface/peft.git
```
younesbelkada, line 67:
```suggestion
        pip install git+https://github.com/huggingface/peft.git@main
```

examples/sentiment/scripts/gpt2-sentiment_peft.py
```diff
@@ -0,0 +1,244 @@
...
+    bias="none",
+    task_type="CAUSAL_LM",
+)
+
+pretrained_model = prepare_model_for_int8_training(pretrained_model)
+pretrained_model = get_peft_model(pretrained_model, lora_config)
+
+model = AutoModelForCausalLMWithValueHead.from_pretrained(pretrained_model)
+
+print(model)
```
lvwerra, line 179: This seems a bit verbose, no?

examples/sentiment/scripts/gpt2-sentiment_peft.py
```diff
@@ -0,0 +1,244 @@
...
+print_trainable_parameters(model)
+
+# GPT-2 tokenizer has a pad token, but it is not eos_token by default. We need to set it to eos_token.
+# only for this model.
+tokenizer.pad_token = tokenizer.eos_token
+
+optimizer = torch.optim.Adam(filter(lambda p: p.requires_grad, model.parameters()), lr=config.learning_rate)
```
lvwerra, line 186: Why do we need to specify the optimizer manually? why not `AdamW`?

trl/models/__init__.py
```diff
@@ -13,8 +13,24 @@
...
+    (
+        AutoModelForCausalLMWithValueHead,
+        AutoModelForSeq2SeqLMWithValueHead,
+    )
+    if not is_peft_available()
+    else (
+        AutoModelForCausalLMWithValueHead,
+        AutoModelForSeq2SeqLMWithValueHead,
+        PeftModelForCausalLM,
+        PeftModelForSeq2SeqLM,
```
lvwerra, line 34: Wouldn't we need a value head for those models as well?

trl/trainer/ppo_trainer.py
```diff
@@ -428,14 +431,28 @@ def step(
...
         timing["time/ppo/forward_pass"] = time.time() - t
 
         t = time.time()
         rewards, non_score_reward = self.compute_rewards(scores, all_logprobs, ref_logprobs, masks)
         timing["time/ppo/compute_rewards"] = time.time() - t
 
+        if self.is_peft_model:
+            self.model.train()
+            self.model.gradient_checkpointing_enable()
+
```
lvwerra, line 455: Isn't that something we want to do outside of the trainer or at least make optional? otherwise GC is always turned on, no?

.github/workflows/tests.yml
```diff
@@ -45,3 +45,26 @@ jobs:
     - name: Test with pytest
       run: |
         make test
+
```
lvwerra, line 48: I think we can add `peft` as an dev/test dependency which is always installed in the test. then we don't need an extra peft CI (we already have 10 tests that run :)). wdyt?

lvwerra (commented):
Added a few comments/questions, overall looks great!

examples/sentiment/scripts/gpt2-sentiment_peft.py
edbeeching: @younesbelkada should we move the lambda into the PPOTrainer?

trl/models/__init__.py
edbeeching: No I don't think so, as we only use peft on the pretrained model.

examples/sentiment/scripts/gpt2-sentiment_peft.py
younesbelkada: Yes I can , this is better I think indeed!

examples/sentiment/scripts/gpt2-sentiment_peft.py
younesbelkada: Fixed in [7915cf4](https://github.com/lvwerra/trl/pull/163/commits/7915cf401a22e46486522055b39913c2985fc0c3)

examples/sentiment/scripts/gpt2-sentiment_peft.py
younesbelkada: Fixed in [7915cf4](https://github.com/lvwerra/trl/pull/163/commits/7915cf401a22e46486522055b39913c2985fc0c3)

trl/models/__init__.py
younesbelkada: Fixed in [e2aff14](https://github.com/lvwerra/trl/pull/163/commits/e2aff14f2ccee3099d2eeb2af204d6b99b39ffb0)

trl/trainer/ppo_trainer.py
younesbelkada: It is recommended for users to enable GC when using `peft` to maximize efficiency, maybe we can have a check that warns users to turn on GC if `peft` is available

.github/workflows/tests.yml
younesbelkada: I proposed a hack to run the tests on the CI runners, however I think we should keep at least one runner that runs the tests without `peft` to check that everything works fine w/o `peft`

trl/trainer/ppo_trainer.py
younesbelkada: let's just educate users on how to properly use this instead on the example

trl/models/modeling_base.py
```diff
@@ -74,19 +95,50 @@ class and the arguments that are specific to trl models.
...
         # either `AutoModelForCausalLM` or `AutoModelForSeq2SeqLM`
         if isinstance(pretrained_model_name_or_path, str):
-            pretrained_model = cls.transformers_parent_class.from_pretrained(
-                pretrained_model_name_or_path, *model_args, **pretrained_kwargs
-            )
-        elif isinstance(pretrained_model_name_or_path, PreTrainedModel):
+            try:
+                peft_filename = hf_hub_download(pretrained_model_name_or_path, "adapter_config.json")
+            except:  # noqa
+                peft_filename = None
```
lvwerra, line 106: We only need to do this if peft is available, right?

.github/workflows/tests.yml
```diff
@@ -33,6 +33,28 @@ jobs:
...
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        # cpu version of pytorch
+        pip install .[test]
+        pip install git+https://github.com/huggingface/peft.git@main
+    - name: Test with pytest
+      run: |
+        make test
+  tests_no_peft:
```
lvwerra, line 49: Why is this step necessary? If the tests run with peft they should also run without, no?

examples/sentiment/scripts/gpt2-sentiment_peft.py
```diff
@@ -0,0 +1,234 @@
...
+        response_tensors.append(response.squeeze()[-gen_len:])
+    batch["response"] = [tokenizer.decode(r.squeeze()) for r in response_tensors]
+
+    # Compute sentiment score
+    texts = [q + r for q, r in zip(batch["query"], batch["response"])]
+    pipe_outputs = sentiment_pipe(texts, **sent_kwargs)
+    rewards = [torch.tensor(output[1]["score"]) for output in pipe_outputs]
+
+    # Run PPO step
+    model.train()
```
lvwerra, line 227: Do we need to set the model to `train()` every step? we never set it to `eval()`

examples/sentiment/scripts/gpt2-sentiment_peft.py
```diff
@@ -0,0 +1,234 @@
...
+set_seed(config.seed)
+
+# Now let's build the main base model! We'll use the `AutoModelForCausalLM` class and load the model in 8 bit mode.
+pretrained_model = AutoModelForCausalLM.from_pretrained(config.model_name, load_in_8bit=True, device_map="auto")
+
+tokenizer = AutoTokenizer.from_pretrained(config.model_name)
+
+"""### Apply LoRA
+Here comes the magic with `peft`! Let's load a `PeftModel` and specify that we are going to use low-rank adapters (LoRA) using `get_peft_model` utility function from `peft`.
+"""
```
lvwerra, line 144: Small nit: let's just also to normal code comments here for consistency.

lvwerra (commented):
PR is in good shape! a few comments:
- docs are missing - i assume we do that in a follow-up PR?
- the saving loading mechanism is not so clear to me yet. the way it is setup currently only adapter and v_head are saved if using peft. what if I want to share the full model with someone?

trl/models/modeling_base.py
younesbelkada: yes!

younesbelkada:
Thanks @lvwerra for the review! 
1- yes we can do it in a follow-up PR regarding docs
2- As the base model is kept untouched during training, you can just share the model-id of the base model and use it - that is also how the models are pushed using `peft` , we never share the entire model but the adapters only, as you can retrieve the model id of the base model using `PeftConfig`. Here is a nice example: https://huggingface.co/crumb/FLAN-OPT-6.7b-LoRA

examples/sentiment/scripts/gpt2-sentiment_peft.py
younesbelkada: I think this is required but not sure why, @edbeeching I will run some tests wo `model.train()` and let you know!

younesbelkada: Here is a run I made without `model.train()`: https://wandb.ai/distill-bloom/trl/runs/fgnwyu1i?workspace=user-younesbelkada I think we can remove that

younesbelkada: Interestingly keeping `model.train()` yields to freeing extra 2GB memory when running `nvidia-smi`

younesbelkada: After discussing with @edbeeching , seems that this is needed as if `model.train()` is not called, it won't backpropagate properly

trl/models/modeling_base.py
younesbelkada: Yes! fixed in [f3ab16a](https://github.com/lvwerra/trl/pull/163/commits/f3ab16a47cb59fa637d263a772e6b8c742015ccf)

.github/workflows/tests.yml
younesbelkada: Removed !

lvwerra (approved):
LGTM! 🚀

PR #97, opened 2023-01-23
[core] improve API

younesbelkada:
# What does this PR do?

(please ignore the branch name)

This PR adds safety checkers on the positional arguments of `PPOTrainer` making the API clearer for users if they pass unvalid arguments when initializing the trainer. This PR adds also nice tests

cc @lvwerra

PR #167, opened 2023-02-20
Fix recent `peft` issue

younesbelkada:
# What does this PR do?

This PR fixes a small issue, to make `trl` compatible with a recent change that has been introduced on `peft` https://github.com/huggingface/peft/pull/97

cc @edbeeching


lvwerra (approved):

younesbelkada:
Thanks for approving
Actually let's close this PR as this change has been already done on #97 !

PR #177, opened 2023-02-27
[`core`] Add `max_grad_norm` support

younesbelkada:
# What does this PR do?

This PR adds the support of clipping gradients norm, that was not supported before but implemented in other frameworks such as clean RL: https://github.com/vwxyzjn/cleanrl/blob/2e41da2a3649c50f27121d74896110fe8f69dd52/cleanrl/ppo.py#L286 

This PR also adds a nice test

I can confirm multi-GPU training works with this PR (ran 1 PPO step of `gpt2-sentiment` with 2 GPU)

cc @edbeeching @lvwerra @natolambert

### DIFF
--- a/tests/test_ppo_trainer.py
+++ b/tests/test_ppo_trainer.py
@@ -645,7 +645,7 @@ class PPOTrainerTester(unittest.TestCase):
 
     def test_ppo_trainer_max_grad_norm(self):
         """
-        Test if the loss trainer works fine
+        Test if the `max_grad_norm` feature works as expected
         """
         # initialize dataset
         dummy_dataset = self._init_dummy_dataset()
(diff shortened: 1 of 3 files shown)


tests/test_ppo_trainer.py
```diff
@@ -643,6 +643,41 @@ def test_batched_forward_pass(self, name):
         self.assertLessEqual(abs_diff_masked_tensors(logprobs_0, logprobs_2[:1], mask_0, mask_2[:1]), 1e-4)
         self.assertLessEqual(abs_diff_masked_tensors(values_0, values_2[:1], mask_0, mask_2[:1]), 1e-4)
 
+    def test_ppo_trainer_max_grad_norm(self):
+        """
+        Test if the loss trainer works fine
```
younesbelkada, line 648:
```suggestion
        Test if the `max_grad_norm` feature works as expected
```

edbeeching (approved):
LGTM, thanks.

lvwerra (approved):
LGTM, thanks for adding @younesbelkada!

PR #182, opened 2023-02-28
[`core`] Fix DeepSpeed zero-3 issue

younesbelkada:
# What does this PR do?

This is an attempt to fix https://github.com/lvwerra/trl/issues/171 

For now the sentiment script hangs, so I need to investigate

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -244,16 +244,40 @@ class PPOTrainer(BaseTrainer):
         else:
             self.kl_ctl = FixedKLController(self.config.init_kl_coef)
 
-        (
-            self.model,
-            self.ref_model,
-            self.optimizer,
-            self.data_collator,
-            self.dataloader,
-            self.lr_scheduler,
-        ) = self.accelerator.prepare(
-            self.model, self.ref_model, self.optimizer, self.data_collator, self.dataloader, self.lr_scheduler
+        # Safety checkers for DS integration
+        is_deepspeed_zero_3 = (
+            self.accelerator.distributed_type == "DEEPSPEED"
+            and hasattr(self.accelerator.state, "deepspeed_plugin")
+            and self.accelerator.state.deepspeed_plugin.zero_stage == 3
         )
+
+        if is_deepspeed_zero_3:
+            (
+                self.model,
+                self.optimizer,
+                self.data_collator,
+                self.dataloader,
+                self.lr_scheduler,
+            ) = self.accelerator.prepare(
+                self.model, self.optimizer, self.data_collator, self.dataloader, self.lr_scheduler
+            )
+            # 8 bit models are already set on the correct device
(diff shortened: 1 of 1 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -244,17 +244,41 @@ def __init__(
...
-            self.model,
-            self.ref_model,
-            self.optimizer,
-            self.data_collator,
-            self.dataloader,
-            self.lr_scheduler,
-        ) = self.accelerator.prepare(
-            self.model, self.ref_model, self.optimizer, self.data_collator, self.dataloader, self.lr_scheduler
+        # Safety checkers for DS integration
+        is_deepspeed_zero_3 = (
```
pacman100, line 248: This changes is irrespective of DS Stage, it should be applied for all DS Stages

trl/trainer/ppo_trainer.py
```diff
@@ -244,17 +244,41 @@ def __init__(
...
+            self.accelerator.distributed_type == "DEEPSPEED"
+            and hasattr(self.accelerator.state, "deepspeed_plugin")
+            and self.accelerator.state.deepspeed_plugin.zero_stage == 3
```
pacman100, line 251:
```suggestion
```

pacman100 (commented):
Thank you @younesbelkada for fixing TRL+DS integration. Left comment. Sentiment pipeline related changes have shared offline.

trl/trainer/ppo_trainer.py
```diff
@@ -265,17 +265,43 @@ def __init__(
...
+            # 8 bit models are already set on the correct device
+            if not getattr(self.ref_model.pretrained_model, "is_loaded_in_8bit", False):
+                # DS integration only allows for single model and as `ref_model` is only used for
+                # `KL devergence loss`,i.e, in eval model, just have it be on the respective device and
+                # there is no need to pass it to the `accelerator.prepare` call
+                self.ref_model = self.ref_model.to(self.accelerator.device)
+
+            # this hack seems to be needed for DS stage 3 to work
+            if self.accelerator.state.deepspeed_plugin.zero_stage == 3:
+                self.model.train()
```
younesbelkada, line 292: Based on the offline discussion I had with @pacman100 , I confirm this hack is needed.

trl/trainer/ppo_trainer.py
```diff
@@ -265,17 +265,43 @@ def __init__(
...
+        if is_deepspeed_used:
+            (
+                self.model,
+                self.optimizer,
+                self.data_collator,
+                self.dataloader,
+                self.lr_scheduler,
+            ) = self.accelerator.prepare(
+                self.model, self.optimizer, self.data_collator, self.dataloader, self.lr_scheduler
+            )
```
lvwerra, line 282: we could move this block outside the if-else and then just call `self.ref_model = self.accelerator.prepare(self.ref_model)` in the else clause, no?

lvwerra (approved):
One small comment, otherwise looks good!

PR #184, opened 2023-03-01
Fix reference to example

jordimas:


lvwerra (approved):
Thanks for fixing the link! ❤️

PR #185, opened 2023-03-01
Clarifications of acronyms and initialisms

meg-huggingface:


younesbelkada (approved):
Thanks a mile for fixing!

Issue #183, opened 2023-03-01
About gpt2-sentiment.py


YTL7:
Can you run this project(examples/sentiment/scripts/gpt2-sentiment.py) successfully? Why do I have many bugs[TypeError: new(): invalid data type 'str']

younesbelkada:
Hello @Ytlskys 
Can you share with us more details about your environment? What is the command you are using?
Thanks!

Ytlskys:
Thank you for your attention！
datasets=2.0.0  torch= 1.12.1+cu116  tqdm=4.64.1  transformers= 4.26.1accelerate=0.16.0
Traceback (most recent call last):
  File "D:\Python\Test\test\T\trl-main\examples\sentiment\scripts\gpt2-sentiment.py", line 135, in <module>
    for epoch, batch in tqdm(enumerate(ppo_trainer.dataloader)):
  File "E:\Python\python3.10.5\lib\site-packages\tqdm\std.py", line 1195, in __iter__
    for obj in iterable:
  File "E:\Python\python3.10.5\lib\site-packages\accelerate\data_loader.py", line 378, in __iter__
    current_batch = next(dataloader_iter)
  File "E:\Python\python3.10.5\lib\site-packages\torch\utils\data\dataloader.py", line 681, in __next__
    data = self._next_data()
  File "E:\Python\python3.10.5\lib\site-packages\torch\utils\data\dataloader.py", line 721, in _next_data
    data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
  File "E:\Python\python3.10.5\lib\site-packages\torch\utils\data\_utils\fetch.py", line 49, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "E:\Python\python3.10.5\lib\site-packages\torch\utils\data\_utils\fetch.py", line 49, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "E:\Python\python3.10.5\lib\site-packages\datasets\arrow_dataset.py", line 1765, in __getitem__
    return self._getitem(
  File "E:\Python\python3.10.5\lib\site-packages\datasets\arrow_dataset.py", line 1750, in _getitem
    formatted_output = format_table(
  File "E:\Python\python3.10.5\lib\site-packages\datasets\formatting\formatting.py", line 532, in format_table
    return formatter(pa_table, query_type=query_type)
[...]

PR #190, opened 2023-03-03
Add 1.12.1 torch compatibility in sum method

PanchenkoYehor:
The current code needs at least 1.13 torch which is not very convenient, because 1.13 torch deprecated CUDA 10.2 and patch 1.13 contains several major updates. There is a high chance that not every project is ready to update their environments and CUDA drivers to use 0.3.1 trl version

### DIFF
--- a/examples/summarization/scripts/reward_summarization.py
+++ b/examples/summarization/scripts/reward_summarization.py
@@ -16,7 +16,6 @@ from transformers import (
 from transformers.utils import PaddingStrategy
 
 
-
 # Define and parse arguments.
 @dataclass
 class ScriptArguments:
--- a/examples/summarization/scripts/reward_summarization.py
+++ b/examples/summarization/scripts/reward_summarization.py
@@ -1,18 +1,20 @@
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional, Union
+
+import evaluate
+import numpy as np
+import torch.nn as nn
 from datasets import load_dataset
 from transformers import (
-    AutoTokenizer,
     AutoModelForSequenceClassification,
-    TrainingArguments,
-    Trainer,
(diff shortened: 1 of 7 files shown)


PanchenkoYehor:
The relevant open issue that currently resolves by updating environment #183

younesbelkada (commented):
Hi @PanchenkoYehor 
Thanks a lot for the fix!
Can you run the styling checks? `make style` + `make quality`
Thanks!

trl/core.py
```diff
@@ -106,14 +104,17 @@ def whiten(values, shift_mean=True):
 
 def masked_mean(values, mask, axis=None):
     """Compute mean of tensor with a masked values."""
-    return (values * mask).sum(axis=axis) / mask.sum(axis=axis)
+    try:
+        return (values * mask).sum(axis=axis) / mask.sum(axis=axis)
+    except:
+        return (values * mask).sum() / mask.sum()
```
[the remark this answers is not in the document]
PanchenkoYehor: Good point, I will change it

lvwerra:
Looks great, thanks for fixing!

lvwerra (approved):

examples/summarization/scripts/reward_summarization.py
```diff
@@ -16,6 +16,7 @@
 from transformers.utils import PaddingStrategy
 
 
+
```
younesbelkada, line 19:
```suggestion
```

trl/core.py
```diff
@@ -26,7 +26,6 @@
 except ImportError:
     from collections import Mapping
```
younesbelkada, line 28:
```suggestion


```

younesbelkada (approved):
Thanks a lot for this!

Issue #40, opened 2022-07-18
No updating on the ref_model?


hdvvip:
Hi, thanks for the great repo,

I wonder that in the original [PPO paper](https://arxiv.org/pdf/1707.06347.pdf), the authors update the ref_model to become the fine-tuned model after every iteration (theta_old = theta). I attached the image below for your convenience.
![Selection_1430](https://user-images.githubusercontent.com/42698038/179431550-60fbed31-a6c7-4894-bbf6-202517137c41.png)

So, there must be a code line where ref_moel = model?
Why didn't you update your ref_model as shown in the original PPO paper?
Thank you.

hdvvip:
Ok I understood, you used [logprob](https://github.com/lvwerra/trl/blob/4fe9988eb8adf0227c26432f8eb3e57a66556350/trl/ppo.py#L156) of the current network as theta_old

train_stats = self.train_minibatch(logprobs[idx].unsqueeze(0), values[idx].unsqueeze(0),
                                                   rewards[idx].unsqueeze(0), queries[idx].unsqueeze(0),
                                                   responses[idx].unsqueeze(0),
                                                   torch.cat([queries[idx],responses[idx]]).unsqueeze(0))

This works similarly to update theta_old after every iteration.

PR #192, opened 2023-03-03
Exploring RM loss potential bug.

### LINKED ISSUE #191: Learning to Summarize Reward Model Loss indexing
Been improving on this for H4, and I think there may be a slight bug. At each reward calculation, which input has a higher reward needs to be known to pass it into the model.

In TRL: https://github.com/lvwerra/trl/blob/a05ddbdd836d3217c80a4b3e679ba984bfd4fa24/examples/summarization/scripts/reward_summarization.py#L185

Here's the original paper, note how the indexing depends on which is selected. Or, maybe this is handled elsewhere in the script (I didn't see it).
image

natolambert:
close #191 if so.
Update: I actually don't think this is right, having looked at it more. We need to know _which part was selected_ rather than _which has a higher reward_, because at early stage of training, it'll be wrong. I'll finish this PR when I figure it out in H4.

### DIFF
--- a/examples/summarization/scripts/reward_summarization.py
+++ b/examples/summarization/scripts/reward_summarization.py
@@ -49,10 +49,7 @@ class ScriptArguments:
         },
     )
     num_train_epochs: Optional[int] = field(
-        default="5",
-        metadata={
-            "help": "The number of training epochs for the reward model. OpenAI used 5."
-        }
+        default="5", metadata={"help": "The number of training epochs for the reward model. OpenAI used 5."}
     )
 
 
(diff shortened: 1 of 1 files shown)


natolambert:
`make style` changed some other lines, not sure why. oops.

natolambert:
I think this should be good, but I haven't gotten `run_rm.py` working yet.

Was trying this config:
```
# Model arguments
# For definitions, see: src/h4/training/config.py
cache_dir: null
config_name: null
config_overrides: null
model_name_or_path: gpt2
model_revision: main
model_type: null
tokenizer_name: null
torch_dtype: null # Use `null` or `float32` for training to avoid NaNs from low-precision types
use_auth_token: false
use_fast_tokenizer: true

# Data training arguments
# For definitions, see: src/h4/training/config.py
block_size: 2048
dataset_config_name: comparisons
dataset_name: openai/summarize_from_feedback
keep_linebreaks: true
max_eval_samples: null
max_train_samples: null
overwrite_cache: false
pre_tokenized: false
preprocessing_num_workers: null
streaming: false
train_file: null
validation_file: null
validation_split_percentage: 5

```
(continues)

But the dataset wasn't formatted correctly. @lewtun want to push a running example from #40 to base the config off?

```
  File "/opt/conda/envs/h4/lib/python3.8/site-packages/datasets/arrow_dataset.py", line 3236, in _map_single
    batch = apply_function_on_filtered_inputs(
  File "/opt/conda/envs/h4/lib/python3.8/site-packages/datasets/arrow_dataset.py", line 3112, in apply_function_on_filtered_inputs
    processed_inputs = function(*fn_args, *additional_args, **fn_kwargs)
  File "scripts/training/run_rm.py", line 236, in convert_to_text_classification_format
    for meta, responses, label in zip(examples["meta"], examples["responses"], examples["label"]):
  File "/opt/conda/envs/h4/lib/python3.8/site-packages/datasets/formatting/formatting.py", line 280, in __getitem__
    value = self.data[key]
KeyError: 'meta'
```

natolambert:
In digging further, I saw the labels are handled elsewhere so this wasn't needed.
Not sure how to delete the PR.

PR #195, opened 2023-03-05
Allow running evaluate-toxicity with cpu

jordimas:
Allow running evaluate-toxicity with cpu


younesbelkada (approved):
Thanks a lot for your work on this!

PR #197, opened 2023-03-06
[`core`] Fix quality issue

younesbelkada:
Fixes quality check that were not being run on the examples to avoid extra files being modified such as in https://github.com/lvwerra/trl/pull/190

cc @lvwerra

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,8 @@ test:
 
 quality:
 	black --check --line-length 119 --target-version py38 tests trl examples
-	isort --check-only tests trl
-	flake8  tests trl
+	isort --check-only tests trl examples
+	flake8  tests trl examples
 
 style:
 	black --line-length 119 --target-version py38 tests trl examples setup.py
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -4,7 +4,7 @@ test:
 	python -m pytest -n auto --dist=loadfile -s -v ./tests/
 
 quality:
-	black --check --line-length 119 --target-version py38 tests trl
+	black --check --line-length 119 --target-version py38 tests trl examples
 	isort --check-only tests trl
 	flake8  tests trl
```
lvwerra, line 10: Let's apply all tests to the same files:

```suggestion
	black --check --line-length 119 --target-version py38 tests trl examples
	isort --check-only tests trl examples
	flake8  tests trl examples

```

lvwerra: Let's do the same in `style` and remove `setup.py`

lvwerra (commented):
Small nit, otherwise good.

Makefile
younesbelkada: Proposed something in [4ad0581](https://github.com/lvwerra/trl/pull/197/commits/4ad05816af684111d5a867e43db086fffb2b37f8)

lvwerra: much cleaner indeed :)

lvwerra (approved):

PR #198, opened 2023-03-06
feat(ci): enable `pip` cache

SauravMaheshkar:
This PR sets the `cache` and `cache-dependency-path` argument of `actions/setup-python` to enable pip caching.

### DIFF
--- a/.github/workflows/clear_cache.yml
+++ b/.github/workflows/clear_cache.yml
@@ -1,0 +1,33 @@
+name: "Cleanup Cache"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
+    
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+        
+      - name: Cleanup
+        run: |
+          gh extension install actions/gh-actions-cache
+          
+          REPO=${{ github.repository }}
+
+          echo "Fetching list of cache key"
+          cacheKeysForPR=$(gh actions-cache list -R $REPO | cut -f 1 )
+
+          ## Setting this to not fail the workflow while deleting cache keys. 
+          set +e
+          echo "Deleting caches..."
+          for cacheKey in $cacheKeysForPR
+          do
+              gh actions-cache delete $cacheKey -R $REPO --confirm
+          done
+          echo "Done"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,6 +16,10 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: "3.8"
+          cache: "pip"
+          cache-dependency-path: |
+            setup.py
+            requirements.txt
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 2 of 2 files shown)


lvwerra:
Hi @SauravMaheshkar, thanks for opening a PR! If I understand it correctly this integration hashes the requirements/setup files to cache dependencies. In that case we would be quite blind to breaking changes from new versions of a library, right?

SauravMaheshkar:
@lvwerra as far as I understand, the action reuses cache if the files listed under the `cache-dependency-path` haven't been changed for while. If the files have been recently changed then it updates the cache. So we should not be blind to breaking changes from new versions.

lvwerra:
I think the issue is that we don't change the files in question often so it is possible that there are breaking changes that we don't notice because we never updated the cached files. Is there an option to set a cache lifetime?

SauravMaheshkar:
AFAIK there isn't a option to set cache lifetime however we can use the `cache-hit` flag to check if a cache hit has occurred. [Source](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#cache-hit)

lvwerra:
Hi @SauravMaheshkar, not sure I follow how the cache-hit would help. I would like to at least clear the cache once every 1-2 days since libraries might be updated that the we won't install due to caching.

SauravMaheshkar:
Hey @lvwerra thanks for the feedback, you do make an interesting case. Upon searching around for a bit, we can create another workflow [similar to the one in the Github Docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries) which can clear all the caches every `n` days.

SauravMaheshkar:
Do you want me to work on that as well in this PR ?

lvwerra:
Hi @SauravMaheshkar, yes, if we can delete the cache every 1-2 days adding caching would be a good feature.

SauravMaheshkar:
@lvwerra I added a proposal for a workflow in e19ce2072fa0dc17c4dc8b60d009ed08f9d7d8f4

lvwerra:
Awesome work @SauravMaheshkar , that looks great to me! What do you think @younesbelkada ?

younesbelkada (commented):
Thanks a lot for this awesome feature!
I am totally ok for this PR, I just have few (basic) questions
- Where the duration of the cache refreshing mechanism (you stated each 1 or 2 days) can be changed on the file you have added?
- Why the changes on `tests.yml` are needed? It seems to be applied only on python 3.8, do we need to change it as well for other python versions?
Thanks a lot!

lvwerra:
For the first question `- cron: "0 0 * * *"` specifies refresh at midnight (see [here](https://crontab.guru)). Also curious about the second one!

younesbelkada:
> For the first question - cron: "0 0 * * *" specifies refresh at midnight (see [here](https://crontab.guru/)).
Awesome thank you!

SauravMaheshkar:
> Thanks a lot for this awesome feature! I am totally ok for this PR, I just have few (basic) questions
> 
> * Where the duration of the cache refreshing mechanism (you stated each 1 or 2 days) can be changed on the file you have added?
> * Why the changes on `tests.yml` are needed? It seems to be applied only on python 3.8, do we need to change it as well for other python versions?
>   Thanks a lot!

We can run the tests on other python versions as well. The changes proposed by this PR is independent of the python version. We can test for more python versions as well.

SauravMaheshkar:
> Hi @SauravMaheshkar just to double check: it seems like the cache is only a pip download cache and the libraries still need to be installed (which takes a significant amount of time). Is this expected?

Yes it's just a download cache AFAIK

PR #206, opened 2023-03-09
[`core`] Update dependency

younesbelkada:
# What does this PR do?

As `peft` did a new release, this PR adds a new extra dependency for `tests` and creates `trl[peft]`

cc @lvwerra

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,8 +66,9 @@ REQUIRED_PKGS = [
     "datasets",
 ]
 EXTRAS = {
-    "test": ["parameterized", "pytest", "pytest-xdist", "accelerate"],
-    "dev": ["parameterized", "pytest", "pytest-xdist", "black", "isort", "flake8>=3.8.3"],
+    "test": ["parameterized", "pytest", "pytest-xdist", "accelerate", "peft"],
+    "peft": ["parameterized", "pytest", "pytest-xdist", "accelerate", "peft"],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -66,8 +66,9 @@
     "datasets",
 ]
 EXTRAS = {
-    "test": ["parameterized", "pytest", "pytest-xdist", "accelerate"],
-    "dev": ["parameterized", "pytest", "pytest-xdist", "black", "isort", "flake8>=3.8.3"],
+    "test": ["parameterized", "pytest", "pytest-xdist", "accelerate", "peft"],
+    "peft": ["parameterized", "pytest", "pytest-xdist", "accelerate", "peft"],
```
lvwerra, line 70: I think the test stuff is not needed and accelerate is already a core dependency, no?
```suggestion
    "peft": ["peft"],
```

setup.py
```diff
@@ -66,8 +66,9 @@
     "datasets",
 ]
 EXTRAS = {
-    "test": ["parameterized", "pytest", "pytest-xdist", "accelerate"],
-    "dev": ["parameterized", "pytest", "pytest-xdist", "black", "isort", "flake8>=3.8.3"],
+    "test": ["parameterized", "pytest", "pytest-xdist", "accelerate", "peft"],
+    "peft": ["peft"],
```
lvwerra, line 70: actually there is a minimal version that works right? so we should set it here.

younesbelkada, line 70:
```suggestion
    "peft": ["peft>=0.2.0"],
```

lvwerra (approved):

PR #207, opened 2023-03-09
pass pretrained_kwargs to peft and hf_hub_download

kashif:
We need to be able to pass args to `hf_hub_download` and or Peft `from_pretrained` e.g. `cache_dir` or `local_files_only` etc.


younesbelkada (commented):
Thanks a lot for your PR! 
I think some kwargs conflicts with the pretrained kwargs, maybe we should separate the kwargs into 4 different kwargs to be on the safe zone:
- PeftConfig.from_pretrained kwargs
- hf_hub_download kwargs
- PreTrainedModel.from_pretrained kwargs
- trl_model_class.from_pretrained kwargs
Thanks a lot!

Issue #215, opened 2023-03-13
Extra code in toxicity example


natolambert:
In the toxicity [script](https://github.com/lvwerra/trl/blob/b75d83ab28b59307916beb425207d46406502f11/examples/summarization/scripts/reward_summarization.py) should the `optimizer` be passed to the PPOTrainer -- or omitted?

Found this because I'm dealing with optimizer setup for H4 by copying the code over.
```
  File "/home/nathan/h4/scripts/training/run_rl.py", line 479, in <module>
    main()
  File "/home/nathan/h4/scripts/training/run_rl.py", line 251, in main
    optimizer = Adam(filter(lambda p: p.requires_grad, model.parameters()), lr=training_args.rl_learning_rate)
  File "/opt/conda/envs/h4/lib/python3.10/site-packages/torch/optim/adam.py", line 137, in __init__
    super(Adam, self).__init__(params, defaults)
  File "/opt/conda/envs/h4/lib/python3.10/site-packages/torch/optim/optimizer.py", line 61, in __init__
    raise ValueError("optimizer got an empty parameter list")
ValueError: optimizer got an empty parameter list
```

PR #216, opened 2023-03-14
Small changes when integrating into H4

natolambert:
Two changes:
1. Pass the optimizer in the sentiment example (currently variable was not passed into trainier).
2. [I think] fix the kwarg option for wandb config of `Accelerate`. See this [docs page](https://docs.wandb.ai/guides/integrations/accelerate), where `init_kwargs` is handled differently. In trying to use this with the code as is, `wandb` is getting read as a `kwarg` and not handled correctly by [this line](https://github.com/lvwerra/trl/blob/b75d83ab28b59307916beb425207d46406502f11/trl/trainer/ppo_trainer.py#L180). If this is different in Tensorboard, it may just be incompatible.

Let me know if I'm wrong!

Fixes: #215

### DIFF
--- a/examples/toxicity/scripts/gpt-j-6b-toxicity.py
+++ b/examples/toxicity/scripts/gpt-j-6b-toxicity.py
@@ -134,7 +134,7 @@ tokenizer.pad_token = tokenizer.eos_token
 
 # We then build the PPOTrainer, passing the model, the reference model, the tokenizer
 ppo_trainer = PPOTrainer(
-    config, model, ref_model=ref_model, tokenizer=tokenizer, dataset=dataset, data_collator=collator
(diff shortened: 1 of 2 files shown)


natolambert:
Closes #https://github.com/lvwerra/trl/issues/215 if correct on point 1 @younesbelkada !

natolambert:
I tested the logging change with my code in H4 #https://github.com/huggingface/h4/pull/73, and it fixed my problem!

natolambert:
I'll test `tensorboard` today. FYI this is needed for the script in H4, so I'll be motivated to get this working soon.

If `tensorboard` doesn't work, I'll prolly do an if statement.

younesbelkada (approved):
Thanks a lot for fixing! 🔥

younesbelkada:
Thanks a lot for experimenting @natolambert ! LGTM

Issue #218, opened 2023-03-14
Add gradient accumulation


edbeeching:
With larger sequences and batches, we quickly go out of memory when the batch size is greater than 1.

lvwerra:
We could probably make use of the `accelerate`[ context manager for gradient accumulation](https://huggingface.co/docs/accelerate/usage_guides/gradient_accumulation#letting-accelerate-handle-gradient-accumulation)!

PR #220, opened 2023-03-14
Grad accumulation and memory bugfix

edbeeching:
- Adds command line arg passing to sentiment and toxicity examples
- Adds gradient accumulation as a command line arg #218 
- Fixes tensors not being detached from graph when stored in stats, leading to an overuse of memory.
- Updates forward_batch_size -> minibatch size in a number of examples

By the way, I think there may be other places we use excessive memory due to storing attached tensors for too long. I will investigate further.

### DIFF
--- a/examples/sentiment/scripts/gpt-neox-20b_peft/gpt-neo-20b_sentiment_peft.py
+++ b/examples/sentiment/scripts/gpt-neox-20b_peft/gpt-neo-20b_sentiment_peft.py
@@ -84,7 +84,7 @@ config = PPOConfig(
     log_with=script_args.log_with,
     mini_batch_size=script_args.mini_batch_size,
     batch_size=script_args.batch_size,
-    accelerator_kwargs=dict(gradient_accumulation_steps=script_args.gradient_accumulation_steps),
+    gradient_accumulation_steps=script_args.gradient_accumulation_steps,
 )
 
 # We then define the arguments to pass to the sentiment analysis pipeline.
--- a/examples/sentiment/scripts/gpt-neox-20b_peft/gpt-neo-20b_sentiment_peft.py
+++ b/examples/sentiment/scripts/gpt-neox-20b_peft/gpt-neo-20b_sentiment_peft.py
@@ -89,7 +89,7 @@ config = PPOConfig(
 
 # We then define the arguments to pass to the sentiment analysis pipeline.
 # We set `return_all_scores` to True to get the sentiment score for each token.
-sent_kwargs = {"return_all_scores": True, "function_to_apply": "none", "batch_size": config.forward_batch_size}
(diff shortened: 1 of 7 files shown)


edbeeching:
I forgot to run style / quality. I am not on my dev machine at the moment. I will run this in an hour.

examples/sentiment/scripts/gpt-neox-20b_peft/gpt-neo-20b_sentiment_peft.py
```diff
@@ -78,13 +82,14 @@ class ScriptArguments:
     model_name=script_args.model_name,
     learning_rate=script_args.learning_rate,
     log_with=script_args.log_with,
-    forward_batch_size=16,
-    batch_size=256,
+    mini_batch_size=script_args.mini_batch_size,
+    batch_size=script_args.batch_size,
+    accelerator_kwargs=dict(gradient_accumulation_steps=script_args.gradient_accumulation_steps),
```
lvwerra, line 87: I would add `gradient_accumlation_steps` a dedicated kwarg of the `PPOConfig`. The `accelerator_kwargs` should be a bit a power feature in my opinion as you need to know accelerate to know what to pass whereas GA is a pretty standard thing that the PPOTrainer can expose. What do you think?

lvwerra (commented):
Thanks @edbeeching, looks very clean to me. Just one small comment.

examples/sentiment/scripts/gpt-neox-20b_peft/gpt-neo-20b_sentiment_peft.py
edbeeching: Sounds good. I've updated the PR and rebased on main.

trl/trainer/ppo_trainer.py
```diff
@@ -177,10 +177,12 @@ def __init__(
...
         # Step 1: Initialize Accelerator
-        self.accelerator = Accelerator(log_with=config.log_with, **config.accelerator_kwargs)
-        self.accelerator.init_trackers(
-            config.tracker_project_name, config=config.to_dict(), init_kwargs=config.tracker_kwargs
+        self.accelerator = Accelerator(
+            log_with=config.log_with,
+            gradient_accumulation_steps=config.gradient_accumulation_steps,
+            **config.accelerator_kwargs,
         )
+        self.accelerator.init_trackers(config.tracker_project_name, config=config.to_dict(), **config.tracker_kwargs)
```
younesbelkada, line 185:
```suggestion
        self.accelerator.init_trackers(config.tracker_project_name, config=config.to_dict(), init_kwargs=config.tracker_kwargs)
```

younesbelkada (commented):
Thanks a lot for this! 
I left a single comment, otherwise the fix proposed in https://github.com/lvwerra/trl/pull/216 will fail I belive

lvwerra (approved):

PR #221, opened 2023-03-15
[`peft`] Fix DP issues

younesbelkada:
# What does this PR do?

This PR fixes issues related to DP with `peft`

Adds also instructions on how to properly run DP + `peft`

cc @lvwerra

### DIFF
--- a/docs/source/sentiment_tuning_peft.mdx
+++ b/docs/source/sentiment_tuning_peft.mdx
@@ -40,13 +40,19 @@ accelerate launch scripts/gpt2-sentiment_peft.py # launches training
 
 You can scale up to as many GPUs as you want, as long as you are able to fit the training process in a single device. The only tweak you need to apply is to load the model as follows:
 ```python
-current_device = torch.distributed.get_rank()
+from accelerate import Accelerator
+...
+
+current_device = Accelerator().process_index
 
 pretrained_model = AutoModelForCausalLM.from_pretrained(
     config.model_name, load_in_8bit=True, device_map={"": current_device}
 )
 ```
-And make sure that the rewards are computed on `current_device` as well.
+As the `Accelerator` object from `accelerate` will take care of initializing the distributed setup correctly.
+Make sure to initialize your accelerate config by specifying that you are training in a multi-gpu setup, by running `accelerate config` and make sure to run the training script with `accelerator launch your_script.py`.
+
+Finally make sure that the rewards are computed on `current_device` as well.
 
 ## Naive pipeline parallelism (NPP) for large models (>60B models)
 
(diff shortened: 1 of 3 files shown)


lvwerra:
Looks good as a temporary fix but we should really change the API a bit to make this easier. :)

younesbelkada:
I see, what you are suggesting is to simplify the model loading API a bit right?

lvwerra:
Exactly, otherwise our API becomes more and more dark magic :D 

I think for NPP, PEFT, Int8 it should all become:

```python
model = AutoModelForCausalLMWithValueHead.from_pretrained(ckpt, method_specific_kwargs)
```
Internally we can then check if the kwargs are consistent and work in that combination and also have useful defaults for some of the approaches.

younesbelkada:
Final training run for gpt2-peft in DP: https://wandb.ai/distill-bloom/trl/runs/anb919vh?workspace=user-younesbelkada

lvwerra:
Looking good :)

trl/trainer/ppo_trainer.py
```diff
@@ -461,15 +461,21 @@ def step(
                 model_inputs["decoder_attention_mask"] = self.accelerator.pad_across_processes(
                     model_inputs["decoder_attention_mask"], dim=1, pad_index=0, pad_first=pad_first
                 )
+            else:
+                model_inputs['labels'] = self.accelerator.pad_across_processes(
```
younesbelkada, line 465: `labels` were used for decoder based models @lvwerra and I mistakenly deleted then in the previous PR

docs/source/sentiment_tuning_peft.mdx
```diff
@@ -36,6 +36,24 @@ accelerate config # will prompt you to define the training configuration
...
+
+You can scale up to as many GPUs as you want, as long as you are able to fit the training process in a single device. The only tweak you need to apply is to load the model as follows:
+```python
+from accelerate import Accelerator
+...
+
+current_device = Accelerator().process_index
+
+pretrained_model = AutoModelForCausalLM.from_pretrained(
+    config.model_name, load_in_8bit=True, device_map={"": current_device}
```
lewtun, line 49: Using an empty key in `device_map` seems a bit like magic to me - could we have a one-liner to explain why?

younesbelkada: Sure yes, I will add it

trl/trainer/ppo_trainer.py
lvwerra, line 465: We actually don't need labels, right? In `prepare_model_inputs` we pop them for encoder-decoder, I think we should do the same for encoders there.

lvwerra (approved):

PR #231, opened 2023-03-20
[`core`] refactor peft API

younesbelkada:
This PR slightly refactors the way we use / load peft models

For now I think it is important to let users pass the `PeftConfig` to `from_pretrained` , as it is an essential component for create peft models

I also think that we should be able to pass `prepare_int8_training` kwargs directly to `from_pretrained` as well, since sometimes you need some tweaks to enable int8 training

This PR is also fully backward compatible with the previous API

cc @lvwerra

### DIFF
--- a/docs/source/sentiment_tuning_peft.mdx
+++ b/docs/source/sentiment_tuning_peft.mdx
@@ -26,6 +26,30 @@ pip install wandb
 
 Note: if you don't want to log with `wandb` remove `log_with="wandb"` in the scripts/notebooks. You can also replace it with your favourite experiment tracker that's [supported by `accelerate`](https://huggingface.co/docs/accelerate/usage_guides/tracking).
 
+## How to use it?
+
+Simply declare a `PeftConfig` object in your script and pass it through `.from_pretrained` to load the TRL+PEFT model. 
+
+```python
+from peft import LoraConfig
+from trl import AutoModelForCausalLMWithValueHead
+
+model_id = "edbeeching/gpt-neo-125M-imdb"
+lora_config = LoraConfig(
+    r=16,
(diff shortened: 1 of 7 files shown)


examples/sentiment/scripts/gpt2-sentiment_peft.py
```diff
@@ -144,8 +144,16 @@ def collator(data):
 # Now let's build the main base model! We'll use the `AutoModelForCausalLM` class and load the model in 8 bit mode.
 current_device = Accelerator().process_index
```
younesbelkada, line 145: this has to be improved

younesbelkada:
GPT2 sentiment script: https://wandb.ai/distill-bloom/trl/runs/jvg0lxz7?workspace=user-younesbelkada
will run the NPP script with the new API soon

younesbelkada:
Final GPT run: https://wandb.ai/distill-bloom/trl/runs/vkgszndu?workspace=user-younesbelkada
NPP run: https://wandb.ai/distill-bloom/trl/runs/9mdl90we?workspace=user-younesbelkada

docs/source/sentiment_tuning_peft.mdx
```diff
@@ -26,6 +26,38 @@ pip install wandb
...
+)
+```
+And if you want to load your model in 8bit precision:
+```python
+pretrained_model = AutoModelForCausalLM.from_pretrained(
+    config.model_name, 
+    peft_config=lora_config,
+    load_in_8bit=True,
+)
+```
```
lvwerra, line 59: That's not necessary, right?

lvwerra (approved):
One comment otherwise looks good to me!

PR #234, opened 2023-03-21
Reduce memory consumption in batched_forward_pass

ohashi56225:
This PR reduces memory consumption in `batched_forward_pass` of PPOTrainer, by avoiding the storage of logits when they are not necessary.

Before this PR, `batched_forward_pass` stored all of the model's `logits` all the time like other tensors such as `values` and `logprobs`. Here, `logits` tensors have a much larger size (batch_size * tokens * **vocabulary_size**) compared to `logprobs` and `values` tensors (batch_size × tokens), consuming a significant amount of cuda memory.

I have modified `batched_forward_pass` to avoid unnecessary storage of `logits`, which is only required when calculating entropy in the `loss` method.

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -654,6 +654,8 @@ class PPOTrainer(BaseTrainer):
                 List of tensors containing the encoded queries, shape (`batch_size`, `query_length`)
             responses (`torch.LongTensor`):
                 List of tensors containing the encoded responses, shape (`batch_size`, `response_length`)
+            return_logits (`bool`, *optional*, defaults to `False`):
+                Whether to return all_logits. Set to `False` if logits are not needed to reduce memory consumption.
         Returns:
             (tuple):
                 - all_logprobs (`torch.FloatTensor`): Log probabilities of the responses,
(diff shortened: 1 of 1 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -644,6 +644,7 @@ def batched_forward_pass(
         queries: torch.Tensor,
         responses: torch.Tensor,
         model_inputs: dict,
+        return_logits: bool = False,
```
younesbelkada, line 647: Could you add a sentence stating what this arg does below on the docstring

younesbelkada (approved):
Thanks a lot for fixing and for taking care of the memory consumption
This looks very good to me!
Would love to hear @lvwerra 's thoughts here

trl/trainer/ppo_trainer.py
ohashi56225: Sure!
How about this sentence as the docstring text?

```
return_logits (`bool`, *optional*, defaults to `False`):
    Whether to return all_logits. Set to `False` if logits are not needed to reduce memory consumption.
```

younesbelkada: Awesome, this looks great!

ohashi56225: Thanks, I added the docstring!

younesbelkada: Great thank you!

lvwerra (approved):
Looks great, thanks!

PR #236, opened 2023-03-21
Adds checkpointing saving and loading

edbeeching:
This PR adds the saving and loading of the model, optimizer and scheduler states.
I don't think it is really needed to store the ref model weights, but it was not clear how to get accelerate.save_state to ignore this.

I updated the example gpt2-sentiment to demonstrate the usage, we will probably want to revert these changes before merging.
I have not tested with a peft model yet.

Early feedback would be great @younesbelkada

### DIFF
--- a/trl/models/modeling_value_head.py
+++ b/trl/models/modeling_value_head.py
@@ -200,15 +200,15 @@ class AutoModelForCausalLMWithValueHead(PreTrainedModelWrapper):
         to the state dictionary of the wrapped model by prepending the key with `v_head.`.
         """
         if not self.is_peft_model:
-            pretrained_model_state_dict = self.pretrained_model.state_dict(*args, **kwargs)
-        else:
-            # if it is a peft model, only save the v_head
-            pretrained_model_state_dict = {}
+            return super(PreTrainedModelWrapper, self).state_dict(*args, **kwargs)
+
+        # if it is a peft model, only save the v_head
+        model_state_dict = {}
 
         v_head_state_dict = self.v_head.state_dict(*args, **kwargs)
         for k, v in v_head_state_dict.items():
-            pretrained_model_state_dict[f"v_head.{k}"] = v
-        return pretrained_model_state_dict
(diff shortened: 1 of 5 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -589,6 +622,10 @@ def collator(data):
         if self.lr_scheduler is not None:
             self.lr_scheduler.step()
 
+        self.current_step += 1
+        if self.config.save_steps and self.current_step % self.config.save_steps == 0:
+            self.accelerator.save_state()
```
younesbelkada, line 627: Nice! Does this needs to be executed on the main rank in case of distributed training? Maybe the `accelerator` already takes care of that

trl/models/modeling_value_head.py
```diff
@@ -200,15 +200,15 @@ def state_dict(self, *args, **kwargs):
         to the state dictionary of the wrapped model by prepending the key with `v_head.`.
         """
         if not self.is_peft_model:
-            pretrained_model_state_dict = self.pretrained_model.state_dict(*args, **kwargs)
-        else:
-            # if it is a peft model, only save the v_head
-            pretrained_model_state_dict = {}
+            return super(PreTrainedModelWrapper, self).state_dict(*args, **kwargs)
```
younesbelkada, line 203:
```suggestion
            state_dict = super(PreTrainedModelWrapper, self).state_dict(*args, **kwargs)
            v_head_state_dict = self.v_head.state_dict(*args, **kwargs)
            for k, v ...
```

younesbelkada (commented):
Looks great! 
I left some suggestions to make the tests pass
Can you also add a test to check that after saving the trainer things work as expected?
Thanks!

trl/models/modeling_value_head.py
younesbelkada: I think something like that might be needed

lvwerra:
Thanks for working on this @edbeeching! Could you provide a small code snippet on how saving and resuming would work for a user?

PR #238, opened 2023-03-21
adds early stopping

### LINKED ISSUE #232: Feature request: PPO early stopping, important for training stability
Hi TRL contributors,

Thanks a lot for creating this nice library.
While using TRL, I realised that an important feature is still missing, that's the early stopping during a PPO step.
Without this feature, the optimisation is prone to KL/loss spikes such as those mentioned in #101 .
In my experiments, I found this issue is alleviated through reducing the default num PPO epochs.

This feature is already implemented in the [OpenAI official PPO](https://github.com/openai/spinningup/blob/master/spinup/algos/pytorch/ppo/ppo.py#L269) and [AllenAI RL4LMs](https://github.com/allenai/RL4LMs/blob/main/rl4lms/algorithms/ppo/ppo.py#L279).

Looking forward to your feedback.

Kind regards,
Shaojie

edbeeching:
Adds early stopping to the PPO loop. Fixes #232 

I used a value of 0.1 as the threshold as I observed an initial spike of 0.2 before we had instabilities in gpt2-xl earlier this week:
image

Note that RL4LM [use a value of 0.5 ](https://github.com/allenai/RL4LMs/blob/main/scripts/training/task_configs/dialog/gpt2_ppo.yml)

My only other concern is gradient accumulation. I think it would it be better to zero the gradients as soon as we see a splike in KL, or to leave them in the as is. I have zeroed them for now, but it would be great to have feedback on this.

### DIFF
--- a/trl/trainer/ppo_config.py
+++ b/trl/trainer/ppo_config.py
@@ -77,6 +77,10 @@ class PPOConfig(object):
             Seed value for random generations
         optimize_cuda_cache (`bool`, *optional*, defaults to `False`):
             Optimize CUDA cache for slightly more memory-effcient training
+        early_stopping (`bool`, *optional*, defaults to `False`):
+            Whether to stop the PPO opimization loop early is the KL too high
+        target_kl (`float`, *optional*, defaults to `0.01`):
+            Stop early if we exceed this value by over 50%
     """
 
     def __init__(
@@ -106,6 +110,8 @@ class PPOConfig(object):
         max_grad_norm: Optional[float] = None,
         seed: Optional[int] = 0,
         optimize_cuda_cache: Optional[bool] = False,
+        early_stopping: Optional[bool] = False,
+        target_kl: Optional[float] = 0.1,
     ):
         self.model_name = model_name
         self.steps = steps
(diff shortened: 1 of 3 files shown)


trl/trainer/ppo_config.py
```diff
@@ -77,6 +77,10 @@ class PPOConfig(object):
             Seed value for random generations
         optimize_cuda_cache (`bool`, *optional*, defaults to `False`):
             Optimize CUDA cache for slightly more memory-effcient training
+        early_stopping (`bool`, *optional*, defaults to `False`):
+            Whether to stop the PPO opimization loop early is the KL too high
+        target_kl (`float`, *optional*, defaults to `0.01`):
```
lvwerra, line 82: the default in the code below is `0.1`

lvwerra (commented):
Generally looks nice and clean to me! My main worry is that this breaks logging because if we early stop the values might not all have the same shape. Did you double check that and if not could you run a quick test?

edbeeching:
Good point, I ran a benchmark with early stopping enabled and a threshold of 0.001 [here](https://wandb.ai/edbeeching/trl/runs/huqvs18b?workspace=user-edbeeching) and the logging seems to work ok.

Issue #235, opened 2023-03-21
Add warning when KL-divergence is negative


lvwerra:
A common failure mode of the training is that the generation kwargs are not correctly set which leads to negative KL-divergence which is almost always a precursor for failed training. We should at least show a warning when this happens.

janpawlowskiof:
Would you care to explain how is the negative KL-divergence caused by incorrect generation kwargs? 

I am trying to train a fine-tuned polish T5 model with kwargs `{"top_k": 0.0, "top_p": 1.0, "do_sample": True, "eos_token_id": 1, "max_new_tokens": 128}`, but objective/kl immediately becomes negative and then everything goes to hell, and i cannot figure out what is the thing I'm missing.
![image](https://user-images.githubusercontent.com/17483828/226899673-b0a95765-ad92-4a01-b891-7b70ced9dc38.png)

younesbelkada:
From my understanding and discussions with @lvwerra , manually adding `eos_token_id:-1` is needed in order for `generate` to continue generation. Otherwise whenever the model predicts `1` as the next token, the generation will not continue, thus leads to much shorter sequences sometimes and being the rootcause of unusual training behaviors

janpawlowskiof:
Thanks for your response! 
Unfortunately I did try this as I was slightly desperate, but with no effect. I got exactly the same behavior. Also I have a preview of generated sentences and they are rather long, so I do not think that this is the cause of my issues.

lvwerra:
Does it work if you just plug in your model into the T5 example? Trying to figure out the source of the issue (model or code). Or vice versa plugin the model used in the T5 example into your code. Even if the rewards don't make much sense with the switched language the negative KL shouldn't occur in any case.

janpawlowskiof:
![image](https://user-images.githubusercontent.com/17483828/227198311-c6663f4a-3522-4a27-83fe-4356f4badc6d.png)

Blue plot is my dataset, purple plot is original dataset from t5-sentiment sample (like it says in the legend).

From what it seems, this is at least somewhat dependent on the dataset that I am using. I am unsure if I can share the dataset, but I will try to figure this out and one I have a while to spare, I will maybe try to create a minimal example where this happens.

lvwerra:
It's a bit hard to know what's going on exactly. In general, a negative KL means that the reference model assigns a higher probability to the sampled tokens than the policy which was used to sample it, which means something with sampling during generation is off.

[...]

PR #239, opened 2023-03-22
[`core`] Add warning when negative KL

younesbelkada:
# What does this PR do?

Fixes: #235 

The warning message will be displayed only once, if there is a repetition, as described in the official documentation of `warnings` package: https://docs.python.org/2/library/warnings.html / https://stackoverflow.com/questions/22661745/why-only-one-warning-in-a-loop

cc @lvwerra

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -750,6 +750,7 @@ class PPOTrainer(BaseTrainer):
             input_kwargs = {key: value[i * fbs : (i + 1) * fbs] for key, value in model_inputs.items()}
             query_batch = queries[i * fbs : (i + 1) * fbs]
             response_batch = responses[i * fbs : (i + 1) * fbs]
+
             logits, _, values = model(**input_kwargs)
 
             if self.is_encoder_decoder:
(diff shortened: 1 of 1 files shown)


lvwerra (commented):
Great idea, however it's the `"objective/kl"` that becomes negative with the wrong generation kwargs :)

younesbelkada:
Indeed! Just fixed it :D

trl/trainer/ppo_trainer.py
```diff
@@ -993,6 +985,14 @@ def record_step_stats(self, kl_coef: float, **data):
 
         mean_non_score_reward = masked_mean(data["non_score_reward"], mask)
 
+        if mean_kl.item() < 0.0:
+            # warn users
+            warnings.warn(
+                f"KL divergence is starting to become negative: {mean_kl.item()} - this might be a precursor for failed training."
```
lvwerra, line 991: for better readability :) 
```suggestion
                f"KL divergence is starting to become negative: {mean_kl.item():.2f} - this might be a precursor for failed training."
```

lvwerra (approved):

PR #254, opened 2023-03-27
[`distributed`] Fix early stopping and DP

younesbelkada:
# What does this PR do?

This PR fixes an issue where the training was hanging in multi-GPU setting and using earyl stopping
In fact, in a multi-GPU setting, we need to compute the average of `policykl` across all GPUs, and call early stopping

I can confirm this solution worked, the way I checked was to add a print after the `self.optimizer_zero_grad` and forcing the coefficient to be `0` :
`if policykl > 0 * self.config.target_kl:`

cc @edbeeching @lvwerra 

```bash
4it [00:03,  1.36it/s]early stopping successfulearly stopping successful
5it [00:04,  1.21it/s]early stopping successful
```

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -642,7 +642,7 @@ class PPOTrainer(BaseTrainer):
                         dist.all_reduce(policykl, dist.ReduceOp.SUM)
                         policykl /= self.accelerator.num_processes
 
-                        if policykl > 0 * self.config.target_kl:
+                        if policykl > 1.5 * self.config.target_kl:
                             early_stop = True
                             self.optimizer.zero_grad()
                             break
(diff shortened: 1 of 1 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -625,10 +625,9 @@ def collator(data):
                     vpreds,
                     batch["masks"],
                 )
-                if self.config.early_stopping and train_stats["policy/policykl"] > 1.5 * self.config.target_kl:
-                    early_stop = True
-                    self.optimizer.zero_grad()
-                    break
+                if self.config.early_stopping:
+                    policykl = train_stats["policy/policykl"]
+                    early_stop = self._early_stop(policykl)
```
edbeeching, line 630: Do we not want to break here if early stop is True? Otherwise we might remain in the inner loop for a few more iterations.

lvwerra:
If it looks good for @edbeeching I am happy to merge it.

trl/trainer/ppo_trainer.py
younesbelkada: Agreed! this should be fixed now

younesbelkada:
Should have addressed the last comment by now!

PR #262, opened 2023-03-29
[`t5`] Fix negative kl issue

### LINKED ISSUE #256: t5-sentiment example collapses on master
I just rerun the t5-sentiment example and rn on master it has negative kl divergence in new version and does not learn in general.
This seems not have happened since the v0.4.1 release.

The t5-sentiment.py script itself does not seem to be the culprit as i tested master version with it reverted to v0.4.1 version and behavior is identical. 

image

younesbelkada:
Fixes #256 

This PR fixes issues related with negative KL and T5 sentiment example. The first fix was related to the sentiment script that was incorrectly ported.

Before this PR, the padding side of tokenizers was always hardcoded to `left` in `_batched_generate`. in encoder-decoder models, they should be set to their native `padding_side` (i.e. `right` for T5), as the padding is performed on the encoder tokens, and these models have been trained with this specific padding side. I think that the culprit is the way the positional attention bias is computed, that does not take into account the starting position if `padding_side=left`: https://github.com/huggingface/transformers/blob/main/src/transformers/models/t5/modeling_t5.py#L436-L451 

Thus forcing padding_side=left for encoder-decoder models should probably be avoided, as most of Enc-dec models pad the encoder input to the right (to verify).

To illustrate the fix: 

- [run on `main` with the modification on the example script only](https://wandb.ai/distill-bloom/trl/runs/0extfcp0)
- [run on this branch](https://wandb.ai/distill-bloom/trl/runs/nxsxr2y3) 

You can see that using `padding_side=left` led to unstable KL, whereas the proposed fix seems to lead to smoother KL

Screenshot 2023-03-29 at 17 16 25

cc @lvwerra

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -417,7 +417,8 @@ class PPOTrainer(BaseTrainer):
         outputs = []
 
         padding_side_default = self.tokenizer.padding_side
-        self.tokenizer.padding_side = "left"
+        if not self.is_encoder_decoder:
+            self.tokenizer.padding_side = "left"
 
         # in case we have fewer examples than bs
         batch_size = min(len(query_tensors), batch_size)
@@ -444,7 +445,12 @@ class PPOTrainer(BaseTrainer):
             generations = self.accelerator.unwrap_model(self.model).generate(**padded_inputs, **generation_kwargs)
 
             for generation, mask in zip(generations, padded_inputs["attention_mask"]):
-                output = generation[(1 - mask).sum() :]  # remove padding
+                if not self.is_encoder_decoder:
+                    output = generation[(1 - mask).sum() :]  # remove padding
+                else:
+                    output = generation
+
+
                 if not return_prompt and not self.is_encoder_decoder:
                     output = output[(mask).sum() :]  # remove prompt
                 outputs.append(output)
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -450,7 +450,6 @@ class PPOTrainer(BaseTrainer):
                 else:
                     output = generation
 
-
                 if not return_prompt and not self.is_encoder_decoder:
                     output = output[(mask).sum() :]  # remove prompt
                 outputs.append(output)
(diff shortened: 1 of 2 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -444,7 +445,11 @@ def _generate_batched(
             generations = self.accelerator.unwrap_model(self.model).generate(**padded_inputs, **generation_kwargs)
 
             for generation, mask in zip(generations, padded_inputs["attention_mask"]):
-                output = generation[(1 - mask).sum() :]  # remove padding
+                if not self.is_encoder_decoder:
+                    output = generation[(1 - mask).sum() :]  # remove padding
+                else:
+                    output = generation
```
lvwerra, line 451: maybe we could remove the special token here?

younesbelkada: I am not sure if this is correct as the token is removed here right after: https://github.com/lvwerra/trl/blob/ed87942a47f26d15e823ca7674737be02e48cc0a/trl/trainer/ppo_trainer.py#L832 
Also made a run with `generation[1:]` and the KL becomes negative: https://wandb.ai/younesbelkada/trl/runs/vjbydeqv -  so I think we shouldn't remove the special token here

janpawlowskiof: Is that not suspicious that such a slight change breaks the training? If no then why is that expected? I'm asking this as I myself am having trouble with my t5 model and negative kl divergence even on v.0.4.1 release.

lvwerra (approved):

avacaondata:
With other encoder-decoder models such as MarianMT models (BART architecture) I'm still experiencing negative kl, in fact it is becoming increasingly negative: -5, -19, -24.

Opdoop:
I experiencing negative kl warning with Alpaca-7B on the sentiment script.

lzy37ld:
In gpt2_sentiment.py, I have tried to modify the generation_kwargs a bit, like set the top_p to 0.9, and it said that my kl fall into negative and maybe because generation kwargs are not correctly set.

I wonder is there any specific restriction on the generation kwargs during ppo?

Gently pin:@younesbelkada

PR #277, opened 2023-04-06
Adding pointer back to Meta's LLaMA.

meg-huggingface:


younesbelkada (approved):
Thanks a lot for fixing and clarifying!

PR #281, opened 2023-04-07
Loss spike check

edbeeching:
Clamps the loss to a threshold `loss_spike_threshold`

### DIFF
--- a/examples/sentiment/scripts/gpt2-sentiment.py
+++ b/examples/sentiment/scripts/gpt2-sentiment.py
@@ -67,7 +67,9 @@ class ScriptArguments:
     )
     early_stopping: Optional[bool] = field(default=False, metadata={"help": "whether to early stop"})
     target_kl: Optional[float] = field(default=0.1, metadata={"help": "kl target for early stopping"})
-    loss_spike_threshold: Optional[float] = field(default=1000.0, metadata={"help": "A loss threshold at which we do no backprop"})
+    loss_spike_threshold: Optional[float] = field(
+        default=1000.0, metadata={"help": "A loss threshold at which we do no backprop"}
+    )
 
 
 parser = HfArgumentParser(ScriptArguments)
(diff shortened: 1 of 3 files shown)


younesbelkada (approved):
Thanks a lot for adding this great feature! 🔥

examples/sentiment/scripts/gpt2-sentiment.py
```diff
@@ -67,6 +67,7 @@ class ScriptArguments:
     )
     early_stopping: Optional[bool] = field(default=False, metadata={"help": "whether to early stop"})
     target_kl: Optional[float] = field(default=0.1, metadata={"help": "kl target for early stopping"})
+    loss_spike_threshold: Optional[float] = field(default=1000.0, metadata={"help": "kl target for early stopping"})
```
lvwerra, line 70: I think the metadata needs to be updated :)

lvwerra (commented):
A general question regarding this: we already clamp the policy gradient loss in `pg_losses2` but then we take the max of the two. .Since advantages should be whitened I guess the reason for a spike would be a very negative `ratio` right?

Looking at the following run it seems like we do get an increase in `approxkl` which is the log-mean of the ratio (the ratio is hard to read on the wandb log). 

https://wandb.ai/lvwerra/trl/runs/ie2h4q8p

To me that seams to indicate that the issue is that we get a very large ratio paired with a negative advantage. Maybe we can get the raw ratios from the logs to see if there was a particular high value in there. Maybe that would be the better place to clip? What do you think?

examples/sentiment/scripts/gpt2-sentiment.py
edbeeching: Are the ratios not clipped with ppo_clip?

edbeeching:
Hmm, are the ratios not clipped with ppo_clip?

PR #284, opened 2023-04-09
Fix swapped helper texts

philipp-classen:
The lines were swapped and the text didn't match.


younesbelkada (approved):
Thanks a lot for fixing!

PR #295, opened 2023-04-10
Log Token distribution of Query / Response

natolambert:


trl/trainer/ppo_config.py
```diff
@@ -73,6 +73,10 @@ class PPOConfig(object):
             "help": "Log with either 'wandb' or 'tensorboard', check  https://huggingface.co/docs/accelerate/usage_guides/tracking for more details"
         },
     )
+    log_token_distribution: Optional[bool] = field(
```
lvwerra, line 76: I would not add another config and just always log it. It's cheap to do.

trl/trainer/ppo_trainer.py
```diff
@@ -1094,6 +1096,16 @@ def record_step_stats(self, kl_coef: float, **data):
...
 
+        if self.config.log_token_distribution:
+            # Log text properties
+            query_lens = torch.tensor([len(query) for query in data["queries"]], dtype=torch.float)
+            response_lens = torch.tensor([len(response) for response in data["responses"]], dtype=torch.float)
+
+            stats["tokens/query_len_mean"] = torch.mean(query_lens).cpu().numpy().item()
+            stats["tokens/query_len_std"] = torch.std(query_lens).cpu().numpy().item()
+            stats["tokens/responses_len_mean"] = torch.mean(response_lens).cpu().numpy().item()
+            stats["tokens/responses_len_std"] = torch.std(response_lens).cpu().numpy().item()
```
lvwerra, line 1107: similar to rewards i would also log the distribution so we can look at histograms in the logs.

lvwerra (commented):
Thanks for adding this @natolambert! Generally looks good, added two comments.

natolambert:
KK should be good, new logs here https://wandb.ai/huggingface/trl/runs/jhz21u5g

natolambert:
KK should be good, example here: https://wandb.ai/huggingface/trl/runs/r2w2101u?workspace=user-natolambert

PR #303, opened 2023-04-14
[`core`] Officially Support Reward Modeling

younesbelkada:
# What does this PR do?

With Reward modeling being an important piece of PPO algorithm, it would be cool to support an "official" RewardTrainer in `trl`.

The `RewardTrainer` simply inherits from `transformers.Trainer`, but with some constraints. Users should be responsible to create a paired dataset that contains `input_ids_j`, `input_ids_k`, `attention_mask_j`, `attention_mask_k`, if they want to use the default `RewardDataCollatorWithPadding` data collator. 
Also I propose to add the possibility to create the PEFT model under the hood, if a user passes a `PeftConfig` to the Trainer. 

This PR adds a first version of it, adds also nice tests and cool documentation about that

cc @lvwerra

### DIFF
--- a/trl/trainer/reward_trainer.py
+++ b/trl/trainer/reward_trainer.py
@@ -1,0 +1,139 @@
+# Copyright 2023 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Callable, Dict, List, Optional, Tuple, Union
+
+import torch
+import torch.nn as nn
+from datasets import Dataset
+from transformers import DataCollator, PreTrainedModel, PreTrainedTokenizerBase, Trainer, TrainingArguments
+from transformers.trainer_callback import TrainerCallback
+from transformers.trainer_utils import EvalPrediction
+
+from ..import_utils import is_peft_available
+from .utils import RewardDataCollatorWithPadding
+
+
+if is_peft_available():
+    from peft import get_peft_model
+
+
+class RewardTrainer(Trainer):
+    r"""
+    The RewardTrainer can be used to train your custom Reward Model. It is a subclass of the
+    `transformers.Trainer` class and inherits all of its attributes and methods. It is recommended to use
+    an `AutoModelForSequenceClassification` as the reward model. The reward model should be trained on a dataset
+    of paired examples, where each example is a tuple of two sequences. The reward model should be trained to
+    predict which example in the pair is more relevant to the task at hand.
+
+    """
+
+    def __init__(
+        self,
+        model: Union[PreTrainedModel, nn.Module] = None,
+        args: TrainingArguments = None,
+        data_collator: Optional[DataCollator] = None,
+        train_dataset: Optional[Dataset] = None,
+        eval_dataset: Optional[Union[Dataset, Dict[str, Dataset]]] = None,
+        tokenizer: Optional[PreTrainedTokenizerBase] = None,
+        model_init: Optional[Callable[[], PreTrainedModel]] = None,
+        compute_metrics: Optional[Callable[[EvalPrediction], Dict]] = None,
+        callbacks: Optional[List[TrainerCallback]] = None,
+        optimizers: Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
+        preprocess_logits_for_metrics: Optional[Callable[[torch.Tensor, torch.Tensor], torch.Tensor]] = None,
+        **kwargs,
+    ):
+        """
+        Initialize RewardTrainer.
+
+        Args:
+            model (`transformers.PreTrainedModel`):
+                The model to train, preferably an `AutoModelForSequenceClassification`.
+            args (`transformers.TrainingArguments`):
+                The arguments to use for training.
+            data_collator (`transformers.DataCollator`):
+                The data collator to use for training. If None is specified, the default data collator (`RewardDataCollatorWithPadding`) will be used
+                which will pad the sequences to the maximum length of the sequences in the batch, given a dataset of paired sequences.
+            train_dataset (`datasets.Dataset`):
+                The dataset to use for training.
+            eval_dataset (`datasets.Dataset`):
+                The dataset to use for evaluation.
+            tokenizer (`transformers.PreTrainedTokenizerBase`):
+                The tokenizer to use for training. This argument is required if you want to use the default data collator.
+            model_init (`Callable[[], transformers.PreTrainedModel]`):
+                The model initializer to use for training. If None is specified, the default model initializer will be used.
+            compute_metrics (`Callable[[transformers.EvalPrediction], Dict]`):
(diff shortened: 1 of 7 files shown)


trl/trainer/reward_trainer.py
```diff
@@ -0,0 +1,139 @@
...
+            model = get_peft_model(model, peft_config)
+
+        if use_reward_data_collator and data_collator is None:
+            if max_length is None or tokenizer is None:
+                raise ValueError(
+                    "max_length or a tokenizer must be specified when using the default RewardDataCollatorWithPadding"
+                )
+            data_collator = RewardDataCollatorWithPadding(tokenizer, max_length=max_length)
+
+            if args.remove_unused_columns:
```
lvwerra, line 107: That will cause an error for probably everyone who is using the library for the first time (since i think it's `True` by default.) what do you think about setting it to `False` and yield a warning.

trl/trainer/reward_trainer.py
```diff
@@ -0,0 +1,139 @@
...
+                The callbacks to use for training.
+            optimizers (`Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR]`):
+                The optimizer and scheduler to use for training.
+            preprocess_logits_for_metrics (`Callable[[torch.Tensor, torch.Tensor], torch.Tensor]`):
+                The function to use to preprocess the logits before computing the metrics.
+            kwargs (`Dict`):
+                Additional keyword arguments to pass to the `RewardTrainer` class. The supported arguments are:
+                    - `use_reward_data_collator` (`bool`, defaults to `False`): Whether to use the default data collator (`RewardDataCollatorWithPadding`).
+                    - `max_length` (`int`, defaults to `None`): The maximum length of the sequences in the batch. This argument is required if you want to use the default data collator.
+                    - `peft_config` (`Dict`, defaults to `None`): The PEFT configuration to use for training. If you pass a PEFT configuration, the model will be wrapped in a PEFT model.
```
lvwerra, line 87: Shouldn't we add them as proper kwargs?

lvwerra: I think the kwarg for the collator could be `reward_data_collator ` or `data_collator ` and be `None` by default in which case the `RewardDataCollatorWithPadding` is used and otherwise the user could pass a custom data collator. What do you think?

lvwerra (commented):
Generally looks really good and clean to me. Left a few comments to try to make it a bit more user friendly.

lvwerra:
Maybe @lewtun would also be interested to have a look to see if there is feedback from the H4 team.

trl/trainer/reward_trainer.py
younesbelkada: SGTM!

trl/trainer/reward_trainer.py
```diff
@@ -0,0 +1,139 @@
...
+
+from ..import_utils import is_peft_available
+from .utils import RewardDataCollatorWithPadding
+
+
+if is_peft_available():
+    from peft import get_peft_model
+
+
+class RewardTrainer(Trainer):
```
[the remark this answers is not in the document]
younesbelkada: Done!

trl/trainer/reward_trainer.py
younesbelkada: Agreed! That would make a nicer API

younesbelkada: Done!

docs/source/reward_trainer.mdx
```diff
@@ -0,0 +1,51 @@
+# Reward Modeling
+
+TRL supports custom reward modeling for anyone to perform reward modeling on their dataset and model. 
+
+## Expected dataset format
+
+The reward trainer expects a very specific format for the dataset. The dataset should contain two 4 entries at least
+if you don't use the default `RewardDataCollatorWithPadding` data collator. The entries should be named:
+
+- `input_ids_j`
```
lewtun, line 10: Small suggestion: since reward modelling datasets typically come from pairs of chosen/rejected examples, we could use these names to make it clear to users which variables correspond to what:

```suggestion
- `input_ids_chosen`
```

An alternative would be to use good/bad like Anthropic does in their papers

docs/source/reward_trainer.mdx
```diff
@@ -0,0 +1,51 @@
...
+- `attention_mask_j`
+- `input_ids_k`
```
lewtun, line 12: Similarly to the above if you agree (and then it should be propagated through):

```suggestion
- `input_ids_rejected`
```

docs/source/reward_trainer.mdx
```diff
@@ -0,0 +1,51 @@
...
+- `attention_mask_k`
+
+The `j` and `k` suffixes are used to denote the two sentences in the paired dataset.
+
+## Using the `RewardTrainer`
+
+After standardizing your dataset, you can use the `RewardTrainer` as a classic HugingFace Trainer. 
+You should pass an `AutoModelForSequenceClassification` model to the `RewardTrainer`.
+
+### Leveraging `peft` library to train a reward model
```
lewtun, line 22: nit:
```suggestion
### Leveraging the `peft` library to train a reward model
```

lewtun (commented):
Awesome feature @younesbelkada 🔥 

I left some tiny nits and a feature request to compute accuracy by default :)

trl/trainer/reward_trainer.py
```diff
@@ -0,0 +1,154 @@
...
```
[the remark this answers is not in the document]
lvwerra: Accuracy is not a very hard metric, maybe we can just build it from scratch here :)

docs/source/reward_trainer.mdx
younesbelkada: Sounds good to me!

docs/source/reward_trainer.mdx
```diff
@@ -0,0 +1,51 @@
...
+- `input_ids_j`
+- `attention_mask_j`
+- `input_ids_k`
+- `attention_mask_k`
+
+The `j` and `k` suffixes are used to denote the two sentences in the paired dataset.
+
+## Using the `RewardTrainer`
+
+After standardizing your dataset, you can use the `RewardTrainer` as a classic HugingFace Trainer.
```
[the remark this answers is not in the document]
younesbelkada: I added few lines in [4bcd96e](https://github.com/lvwerra/trl/pull/303/commits/4bcd96eee41ab64f6e20599fec0c5cebe865a10d) but not sure if what I said is 100% correct, would love to have a second look here!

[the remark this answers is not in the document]
younesbelkada: Will also add an example now

trl/trainer/reward_trainer.py
```diff
@@ -0,0 +1,167 @@
...
+            self.use_reward_data_collator = False
+        super().__init__(
+            model,
+            args,
+            data_collator,
+            train_dataset,
+            eval_dataset,
+            tokenizer,
+            model_init,
+            compute_metrics,
```
lvwerra, line 139: i am not sure this works: if we overwrite the class method with our own metric the compute metrics in the parent class is never used, no? 

what about defining a `compute_accuracy` function outside the class and pass if `compute_metrics` from the init is `None`

younesbelkada: Sounds like a great plan!

docs/source/reward_trainer.mdx
```diff
@@ -0,0 +1,56 @@
...
+    task_type=TaskType.SEQ_CLS,
+    inference_mode=False,
+    r=8,
+    lora_alpha=32,
+    lora_dropout=0.1,
+)
+
+...
+
+trainer = RewardTrainer(
```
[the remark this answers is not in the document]
younesbelkada: Now I pass it! Thanks for the heads up

lvwerra (approved):

PR #315, opened 2023-04-23
Give a key to the wandb PPOConfig config entry

JulesGM:
There is a lot of stuff with very generic keys in the `PPOConfig` dict, and the user may have logged a `wandb` config dict elsewhere. I know I had that problem. To counter that, I pass the PPOConfig dict in a dict under the key `trl_ppo_trainer_config`, to prevent collisions & be very clear.

PR #330, opened 2023-04-28
[`PPOTrainer`] Fix tensorboard issue

younesbelkada:
# What does this PR do?

Addresses https://github.com/lvwerra/trl/pull/315#issuecomment-1525832522

With #315 being merged, the tensorboard logging was slightly broken. This PR fixes that issue by adding a small check

cc @lvwerra


JulesGM:
that's cool. would it make sense to add `trl_ppo_trainer_` as a prefix to the keys for tensorboard? that way there is no confusion in tensorboard as well

lvwerra (approved):

PR #331, opened 2023-04-28
TRL Currently breaks with bfloat16. Fixes `mini_batch_data` to support BFloat16

JulesGM:
https://github.com/lvwerra/trl/blob/main/trl/trainer/ppo_trainer.py#L641 breaks `BFloat16`. Indeed, hf `datasets.Dataset` don't support `bfloat16` it would appear.  We instead create a dict friendly `torch.utils.data.Dataset` that has the same interface, & it works flawlessly.

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -90,6 +90,24 @@ outputs = model(**inputs, labels=inputs["input_ids"])
 ```
 """
 
+class DictDataset(torch.utils.data.Dataset):
+    def __init__(self, data):
+        self._data = data
+
+    def __getitem__(self, *argc, **kwargs):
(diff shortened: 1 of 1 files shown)


JulesGM:
@younesbelkada

younesbelkada (commented):
Hi @JulesGM 
Thanks for the PR! 
Could you clarify in which case do you face this issue with ideally a reproducible script?

JulesGM:
I'm in a bit of a rush right now, but have you ever tried bfloat16 with the current code-base ?

younesbelkada:
Closing for now as there is no clear reproducer of the issue yet, feel free to re-open if you can send us a simple and clear reproducer of the issue you are facing

PR #339, opened 2023-05-04
Fix argument's description

vinhkhuc:
Thanks for the great project.
This PR fixes a minor issue in the argument's description.


younesbelkada (approved):
Thanks a lot for this!

PR #359, opened 2023-05-11
stack_llama: add parameter to control max_length (to mitigate OOM errors)

teticio:
As I "only" have 24 Gb VRAM, I run out of memory when training the reward model. The dataset was previously truancated to a hardcoded value of 512; I have made this a script argument.


lvwerra (approved):
Thanks, looks good to me!

younesbelkada (approved):
Awesome work! Thanks a lot for this!

PR #363, opened 2023-05-12
add is_trainable in kwargs

Opdoop:
# What does this PR do?

This PR enables continue training of peft model by add `is_trainable` as a `kwarg` on `from_pretained` of `PreTrainedWrapper`. (ref. https://github.com/lvwerra/trl/pull/342#discussion_r1191286149)

`is_trainable` is a parameter of `PeftModel.from_pretrained()` which enable continue training. `LoraConfig` do not have `is_trainable` parameter. 

Use case example: https://github.com/huggingface/peft/issues/282#issuecomment-1500804381
Source: https://github.com/huggingface/peft/blob/main/src/peft/peft_model.py#L145

cc @lvwerra @younesbelkada


younesbelkada (approved):
Thanks a lot for the clean work! 
This looks great to me!

lvwerra (approved):

PR #375, opened 2023-05-18
 best-of-n sampler class

metric-space:
# Rationale

Follow up work to this closed ticket:  https://github.com/lvwerra/trl/issues/140
This is to create a class that can aid in `best-of-n sampling` of models using a reward pipeline

Usage example (see the last most cell) : https://gist.github.com/metric-space/a315a97d581039cd47888c24caba9f65


metric-space:
Just was hoping to get initial feedback. The class part does seem a bit clunky but I'll leave it to you folks to tell me more  on how to improve the abstraction

trl/extras/best_of_n_sampler.py
```diff
@@ -0,0 +1,76 @@
...
+        if not isinstance(model, (SUPPORTED_ARCHITECTURES)):
+            raise ValueError(
+                f"model must be a PreTrainedModelWrapper, got {type(model)} - supported architectures are: {SUPPORTED_ARCHITECTURES}"
+            )
+
+        self.model = model
+        self.tokenizer = tokenizer
+
+        if reward_pipeline is None:
+            raise ValueError("reward_pipeline must be provided")
```
younesbelkada, line 38: I would add a check that it is an instance from pipeline here

trl/extras/best_of_n_sampler.py
```diff
@@ -0,0 +1,76 @@
...
+
```
younesbelkada, line 39: After thinking a bit, can we just make the reward a callable instead of using pipeline?

younesbelkada (commented):
This looks awesome to me, thanks a lot for working on this feature!
I would add a section in the documentation to explain what are inside `extras` and how to properly use `BestOfNSampler` class to educate future users. 
Also, what if someone doesn't want to rely on pipeline for the reward function? So after thinking a bit, I think that for broader usage we should accept a callable for the reward pipeline to make it very customizable. What do you think?

trl/extras/best_of_n_sampler.py
metric-space: I wasn't aware there would be such a use-case, thank you for that and will do

metric-space: I think the reasoning to make more broader makes sense. 

Just to clarify, and I may be skipping a few steps and with the constraint of making the API look less like a patchwork doll, do you envision removing all the pipeline based parameters :`reward_pipeline` and  `reward_kwargs` and with this also removes any assumption we can make about the pipeline output in this line (and we remove this as well) 

`[self.reward_extractor(output) for output in self.reward_pipeline(output, **self.reward_kwargs)]` ?

This would result in placing the onus on the user to give us a callable that takes care of all the three concerns (model, model params and output extraction) so all we have to do is plug it here: https://github.com/lvwerra/trl/pull/375/files#diff-667e046b6a4c1712553bdec501359fe107cb415f05e5b2526b1df83fa2b4d14bR72

younesbelkada (approved):
Amazing work @metric-space !! 🔥 Thanks so much for adding this!

lvwerra (commented):
Thanks for adding this - the PR looks very clean! The main question I have is regarding the API and `best_of_lot`. I think it might be easier to use if instead of a filter function we have the user provide a scoring function (`[str]`->`[float]`). This will also allow us in the future for example to go for top-k instead of top-1 filtering. What do you think?

metric-space:
Thank you for the reviews folks. @lvwerra I see what you mean. It does seem that would add more utility to what the class does and takes a bit of burden away from the user.

I am wondering if  `top_k` is the only new parameter desired or are there are other potential parameters like `sampling`? If sampling is a desired thing, this would lay some grounding for my wonderment over the mean portion of @natolambert 's pseudocode that I never understood https://github.com/lvwerra/trl/issues/140#issuecomment-1522126764

natolambert:
Hmmm @metric-space is your question around if it is used in the optimization loop? I guess I never studied it too deeply (my understanding is that in WebGPT they use it to boost performance after training models, so optimization / gradients not needed). Does that make sense? Can you help me with a more specific question?

metric-space:
@natolambert that does make sense. I think my question was around the `zero mean` aspect mentioned in your previously mentioned pseudo code which now that I think about it, is probably orthogonal to what I'm trying to gather here

metric-space:
@lvwerra  @younesbelkada  Hey folks, just a slight nudge on this

trl/extras/best_of_n_sampler.py
```diff
@@ -0,0 +1,74 @@
...
+        self,
+        model: PreTrainedModelWrapper,
+        tokenizer: Union[PreTrainedTokenizer, PreTrainedTokenizerFast],
+        queries_to_scores: Callable[
+            [List[str]], List[float]
+        ],  # callable that takes a list of generated texts and returns the associated reward scores
+        length_sampler: Any,
+        sample_size: int = 4,
+        seed: Optional[int] = None,
+        postop_top_k: int = 1,
```
lvwerra, line 21: maybe we can call this something that's easier to grasp. Maybe `n_candidates` or `return_n`?

trl/extras/best_of_n_sampler.py
```diff
@@ -0,0 +1,74 @@
...
+        generation_config: Optional[GenerationConfig] = None,
+    ) -> None:
```
lvwerra, line 23: can you add a docstring?

trl/extras/best_of_n_sampler.py
```diff
@@ -0,0 +1,74 @@
...
+        self.sample_size = sample_size
+        self.postop_top_k = postop_top_k
+
+    def generate(
+        self,
+        query_tensor: Union[torch.Tensor, List[torch.Tensor]],
+        skip_special_tokens: bool = True,
+        device: Optional[Union[str, torch.device]] = None,
+        **generation_kwargs,  # way to override generation config
+    ) -> List[List[str]]:
```
lvwerra, line 51: also docstring for this :)

trl/extras/best_of_n_sampler.py
```diff
@@ -0,0 +1,74 @@
...
+        self.gen_config = generation_config
+        self.sample_size = sample_size
+        self.postop_top_k = postop_top_k
+
+    def generate(
+        self,
+        query_tensor: Union[torch.Tensor, List[torch.Tensor]],
+        skip_special_tokens: bool = True,
+        device: Optional[Union[str, torch.device]] = None,
+        **generation_kwargs,  # way to override generation config
```
lvwerra, line 50: with docstring we can remove comments

```suggestion
        **generation_kwargs,
```

trl/extras/best_of_n_sampler.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class BestOfNSampler(object):
+    def __init__(
+        self,
+        model: PreTrainedModelWrapper,
+        tokenizer: Union[PreTrainedTokenizer, PreTrainedTokenizerFast],
+        queries_to_scores: Callable[
+            [List[str]], List[float]
+        ],  # callable that takes a list of generated texts and returns the associated reward scores
```
lvwerra, line 17: with docstrings we can remove comment

```suggestion
        ],
```

lvwerra (commented):
Looks great! Left a small comment regarding docstrings and variable naming. Could you add tests for the class as well? Then it should be good to merge. Let us know if you need help with the tests!

metric-space:
@younesbelkada  @lvwerra  Thank you for the reviews and feedback. Work on testing this new class is underway, stay tuned

younesbelkada (commented):
Awesome! 
As a side note, some CI tests are flaky but we skipped them, you should merge your branch with upstream main to see that everything works

metric-space:
Hey folks, I did add in a few tests. Please let me know if those need further work. @younesbelkada apologies for the third ping but there is a bit of code updates beyond the tests additions hence the need for a review again

just a heads up 
https://github.com/lvwerra/trl/pull/375/files#diff-667e046b6a4c1712553bdec501359fe107cb415f05e5b2526b1df83fa2b4d14bR97-R100
this may a look repetitive but this is to stop pytorch from warning about how tensors are copied

younesbelkada (approved):
Thanks a lot for this clean implementation !

younesbelkada:
I see you have addressed all comments from @lvwerra and you added nice tests, merging the PR!
Again thanks a lot for this and looking forward to contributing again 🎉

metric-space:
You're welcome and thank you for the opportunity

PR #380, opened 2023-05-20
from_pretrain with peft adapter on the hub (# 379)

glerzing:
I also added a warning in case an adapter is found but a peft_config is also given. And I tried to simplify a bit since the same processing is done in PeftConfig.from_pretrained.

is_peft_available() is indirectly checked. If it is not installed, it will trigger an exception and trained_adapter_config will be set to None.

Let me know what you think. If you don't think this is an improvement, I can revert, no problem.

### DIFF
--- a/trl/models/modeling_base.py
+++ b/trl/models/modeling_base.py
@@ -140,13 +140,12 @@ class PreTrainedModelWrapper(nn.Module):
         # either `AutoModelForCausalLM` or `AutoModelForSeq2SeqLM`
         if isinstance(pretrained_model_name_or_path, str):
             try:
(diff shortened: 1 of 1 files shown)


younesbelkada (commented):
Thank you very much for this! 
I have one comment, I am not really comfortable using `try/except` directly on the `PeftConfig.from_pretrained` call as this might catch a lot of exceptions that we don't want to catch. How about using `hf_hub_download` as it was done before?

glerzing:
I just pushed a new version. I also added a few logging.info(). Let me know if that's a bad idea, I can revert that too.

younesbelkada (approved):
Thanks very much! This looks good to me

PR #389, opened 2023-05-26
fix dataloader typo in ppo_trainer.py

LZY-the-boys:
`dataloader must be a torch.utils.data.Dataset`: `dataloader` should be `dataset`


lvwerra (approved):
Thanks for the fix!

PR #392, opened 2023-05-29
Fixed some type annotations of trl.trainer.PPoTrainer

JulesGM:
- Ref model should be Optional
- The usual annotation for the Huggingface tokenizers is PreTrainedTokenizerBase. Not using that messes up people's annotation checks.
- Fixed the comments wrt the other two points


JulesGM:
Should be good now. Are there tests I should be running?

JulesGM:
forgot to sync

younesbelkada (approved):
Thanks a lot!

lvwerra (approved):

PR #393, opened 2023-05-29
keep state_dict kwargs instead of popping it in save_pretrained

rizar:
This PR fixes the following issue: when `save_pretrained(...)` was called for `AutoModelForCausalLMWithValueHead` with an explicitly given `state_dict`, this `state_dict` was deleted from `kwargs` and not passed further. I believe the intended behavior was to keep `state_dict` in `kwargs` if it's already in there.


lvwerra (approved):
Looks good to me but I'll let @younesbelkada double check :)

younesbelkada (approved):
LGTM! Merging once the CIs are green!

Issue #386, opened 2023-05-24
Lora weights are not merged properly in merge_peft_adapter


vvasily:
It seems that merge_peft_adapter doesn't work properly. Base and output models will be with the same weights.

Weights of Lora are not merged and at least `target.merge()` should be added before this line.
https://github.com/lvwerra/trl/blob/25fa1bd8803c9998d95a3c38b2bb27a85235ce7b/examples/stack_llama/scripts/merge_peft_adapter.py#L58

SangMyeongWoh:
I also have same think with you.
target.merge() should be called somewhere.

younesbelkada:
Hi 
If you use the latest version of `peft` you can just call `model.merge_and_unload()` and the model will be automatically merged. Let us know if you face some issues with that

Issue #327, opened 2023-04-27
RLHF pipeline for StackLLama with pretrained models doesn't work


vvasily:
I try to run RLHF pipeline for StackLLama with your pretrained models and it failed.
It seems that the reward model has improper output layer.

I'm following this [README](https://github.com/lvwerra/trl/tree/main/examples/stack_llama/scripts)
To got the pretrained models I do:

1) Download the original llama-7b model

2) To get LLAMA_SE_MODEL, I downloaded your adapter: [trl-lib/llama-7b-se-peft](https://huggingface.co/trl-lib/llama-7b-se-peft) and merged it with llama-7b via your merge_peft_adapter.py script:
`python merge_peft_adapter.py --adapter_model_name=/home/user/llama-7b-se-peft/ --base_model_name=/home/user/llama7b/ --output_name=/home/user/llama-7b-se-ckpt/`
As I understand now LLAMA_SE_MODEL is saved in /home/user/llama-7b-se-ckpt/

3) After that to get LLAMA_SE_RM_MODEL, I downloaded your adapter: [trl-lib/llama-7b-se-rm-peft](https://huggingface.co/trl-lib/llama-7b-se-rm-peft) and merged it with LLAMA_SE_MODEL (step 2) via your merge_peft_adapter.py script:
`python merge_peft_adapter.py --adapter_model_name=/home/user/llama-7b-se-rm-peft --base_model_name=/home/user/llama-7b-se-ckpt/ --output_name=/home/user/llama-7b-se-rm-ckpt`
As I understand now LLAMA_SE_RM_MODEL is placed in /home/user/llama-7b-se-rm-ckpt

After that I tried to start rl_training.py script:
`python examples/stack_llama/scripts/rl_training.py --model_name=/home/user/llama-7b-se-ckpt --reward_model_name=/home/user/llama-7b-se-rm-ckpt --adafactor=False --tokenizer_name=/home/user/llama-7b-se-ckpt --save_freq=100 --output_max_length=128 --batch_size=1 --gradient_accumulation_steps=8 --batched_gen=True --ppo_epochs=4 --seed=0 --learning_rate=1.4e-5 --early_stopping=True --output_dir=llama-se-rl-finetune-128-8-8-1.4e-5_adam
`
And get the following error related to reward model:
  File "examples/stack_llama/scripts/rl_training.py", line 255, in <listcomp>
    rewards = [torch.tensor(output[0]["score"] - script_args.reward_baseline) for output in pipe_outputs]
[...]

Issue #297, opened 2023-04-11
[stackllama] merge_peft_adapter miss fc_head in reward model


akk-123:
reward model use AutoModelForSequenceClassification to add a fc layer on base_model, and saved adapter_model.bin include `base_model.model.score.weight`, but after use merge_peft_adapter, saved model miss the fc layer

PR #398, opened 2023-06-01
StackLLaMA: correctly merge peft model

### LINKED ISSUE #368: Llama Reward Model is incorrectly merged
As mentioned in #287, `merge_peft_adapter` saves the Llama RM as a `LlamaForCausalLM` see [here](https://github.com/lvwerra/trl/blob/main/examples/stack_llama/scripts/merge_peft_adapter.py#L35)

But the reward model is trained and should be a `LlamaForSequenceClassification` and running `rl_training.py` gives the obvious warnings
```
Some weights of the model checkpoint at ./llama-7b-se-rm were not used when initializing LlamaForSequenceClassification: ['lm_head.weight']
- This IS expected if you are initializing LlamaForSequenceClassification from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
- This IS NOT expected if you are initializing LlamaForSequenceClassification from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
Some weights of LlamaForSequenceClassification were not initialized from the model checkpoint at /home/toolkit/huggingface/llama-7b-rm and are newly initialized: ['score.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
```

We should instead check whether we are merging the rm and then save as a the correct model 

Also the `score.weight` is not being loaded as mentioned in #297 , see more info below

--- update --

It seems that `merge_peft_adapter` should be using `merge_and_unload()` which correctly overrides the score. But I haven't yet managed to get good results using the adapter weights on the hub

mnoukhov:
fixes #368 #386 #297 and partially #327
correctly merge weights with peft's `merge_and_unload`
if merging a reward model, load `sequence classification`

Note that `trl-lib/llama-7b-se-rm-peft` does not contain the `score` weight so I will release my own pretrained reward model adapter weights

### DIFF
--- a/examples/stack_llama/scripts/merge_peft_adapter.py
+++ b/examples/stack_llama/scripts/merge_peft_adapter.py
@@ -1,10 +1,8 @@
 from dataclasses import dataclass, field
 from typing import Optional
 
-import peft
 import torch
 from peft import PeftConfig, PeftModel
-from peft.utils import _get_submodules
 from transformers import AutoModelForCausalLM, AutoModelForSequenceClassification, AutoTokenizer, HfArgumentParser
 
 
--- a/examples/stack_llama/scripts/merge_peft_adapter.py
+++ b/examples/stack_llama/scripts/merge_peft_adapter.py
@@ -5,12 +5,7 @@ import peft
 import torch
 from peft import PeftConfig, PeftModel
 from peft.utils import _get_submodules
-from transformers import (
-    AutoModelForCausalLM,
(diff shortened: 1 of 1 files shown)


mnoukhov:
Done!

younesbelkada (approved):
Thanks a lot!

lvwerra (approved):

PR #399, opened 2023-06-01
StackLLaMA: fix supervised finetuning and reward model training

mnoukhov:
for supervised finetuning
- removed tokenizer hacks since they are no longer necessary with the updated `llamatokenizer`
- black + isort

for reward modeling
- added `tokenizer_name` so it can be separately specified from model
- again removed old llama tokenizer hacks
- added `eval_first_step` option to add an eval loop after the first step to make nicer graphs
- black + isort

### DIFF
--- a/examples/stack_llama/scripts/reward_modeling.py
+++ b/examples/stack_llama/scripts/reward_modeling.py
@@ -14,15 +14,10 @@ from transformers import (
     HfArgumentParser,
     PreTrainedTokenizerBase,
     Trainer,
+    TrainerCallback,
     TrainingArguments,
 )
 from transformers.utils import PaddingStrategy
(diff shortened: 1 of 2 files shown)


younesbelkada (commented):
Thanks for your hardwork @mnoukhov ! 
I have the same comments as here: https://github.com/lvwerra/trl/pull/398#pullrequestreview-1457294425 
If you run the styling checks we should be all good!

mnoukhov:
Ran the checks and added the configs to my workspace config for the future :)

younesbelkada (approved):
Thanks a lot!

lvwerra (approved):
Thanks a lot for updating!

PR #400, opened 2023-06-01
StackLlama: fixed RL training and added args

mnoukhov:
added steps argument and break to respect max training epochs added more PPOConfig args to script args
removed llama tokenizer hacks
black + isort

~switched to `LlamaTokenizer` from `AutoTokenizer`~  added `return_token_type_ids=False` to pipeline kwargs because
- the latter loads `LlamaTokenizerFast`
- `LlamaTokenizerFast` will output `token_type_ids` see https://github.com/huggingface/transformers/issues/23818 and https://github.com/huggingface/transformers/pull/23909
-  these `token_type_ids` cause an error in our reward model `pipeline`, namely `TypeError: LlamaForSequenceClassification.forward() got an unexpected keyword argument 'token_type_ids'`

### DIFF
--- a/examples/stack_llama/scripts/rl_training.py
+++ b/examples/stack_llama/scripts/rl_training.py
@@ -20,16 +20,11 @@ from accelerate import Accelerator
 from datasets import load_dataset
 from peft import LoraConfig
 from tqdm import tqdm
-from transformers import Adafactor, AutoTokenizer, HfArgumentParser, pipeline
+from transformers import Adafactor, HfArgumentParser, LlamaTokenizer, pipeline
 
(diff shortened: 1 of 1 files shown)


ArthurZucker:
You can use `return_token_type_ids=False ` as a quick fix to prevent them from being passed

younesbelkada (commented):
THank you so much for your contribution! Could you just run the styling checks? After that we should be good for merging
```bash
make style && make quality
```

mnoukhov:
Fixed style and quality and Switched back to Autotokenizer, thanks for the tip @ArthurZucker

younesbelkada (approved):
Thanks so much!

lvwerra (approved):

PR #406, opened 2023-06-05
Remove unused imports in docs.

vwxyzjn:
Minor minor improvement on docs :)

### DIFF
--- a/docs/source/quickstart.mdx
+++ b/docs/source/quickstart.mdx
@@ -23,13 +23,13 @@ from trl import PPOTrainer, PPOConfig, AutoModelForCausalLMWithValueHead
 from trl.core import LengthSampler, respond_to_batch
 
 # 1. load a pretrained model
-model = AutoModelForCausalLMWithValueHead.from_pretrained('gpt2')
-model_ref = AutoModelForCausalLMWithValueHead.from_pretrained('gpt2')
-tokenizer = GPT2Tokenizer.from_pretrained('gpt2')
+model = AutoModelForCausalLMWithValueHead.from_pretrained("gpt2")
+model_ref = AutoModelForCausalLMWithValueHead.from_pretrained("gpt2")
+tokenizer = GPT2Tokenizer.from_pretrained("gpt2")
 tokenizer.pad_token = tokenizer.eos_token
 
 # 2. initialize trainer
-ppo_config = {'batch_size': 1}
+ppo_config = {"batch_size": 1}
 config = PPOConfig(**ppo_config)
 ppo_trainer = PPOTrainer(config, model, model_ref, tokenizer)
 
--- a/docs/source/quickstart.mdx
+++ b/docs/source/quickstart.mdx
@@ -19,7 +19,7 @@ The following code illustrates the steps above.
 # 0. imports
 import torch
 from transformers import GPT2Tokenizer
-from trl import PPOTrainer, PPOConfig, AutoModelForCausalLMWithValueHead, create_reference_model
+from trl import PPOTrainer, PPOConfig, AutoModelForCausalLMWithValueHead
 from trl.core import respond_to_batch
 
 # 1. load a pretrained model
--- a/docs/source/quickstart.mdx
+++ b/docs/source/quickstart.mdx
@@ -21,7 +21,6 @@ import torch
 from transformers import GPT2Tokenizer
 
 from trl import AutoModelForCausalLMWithValueHead, PPOConfig, PPOTrainer
-from trl.core import LengthSampler, respond_to_batch
 
 
 # 1. load a pretrained model
--- a/docs/source/quickstart.mdx
+++ b/docs/source/quickstart.mdx
@@ -19,8 +19,10 @@ The following code illustrates the steps above.
 # 0. imports
 import torch
 from transformers import GPT2Tokenizer
-from trl import PPOTrainer, PPOConfig, AutoModelForCausalLMWithValueHead
+
+from trl import AutoModelForCausalLMWithValueHead, PPOConfig, PPOTrainer
 from trl.core import LengthSampler, respond_to_batch
+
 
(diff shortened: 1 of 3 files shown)


docs/source/quickstart.mdx
```diff
@@ -19,13 +19,14 @@ The following code illustrates the steps above.
...
 from transformers import GPT2Tokenizer
-from trl import PPOTrainer, PPOConfig, AutoModelForCausalLMWithValueHead, create_reference_model
+from trl import PPOTrainer, PPOConfig, AutoModelForCausalLMWithValueHead
 from trl.core import respond_to_batch
 
 # 1. load a pretrained model
 model = AutoModelForCausalLMWithValueHead.from_pretrained('gpt2')
 model_ref = AutoModelForCausalLMWithValueHead.from_pretrained('gpt2')
 tokenizer = GPT2Tokenizer.from_pretrained('gpt2')
+tokenizer.pad_token = tokenizer.eos_token
```
vwxyzjn, line 29: Without this I was getting

```
    raise ValueError(
ValueError: Asking to pad but the tokenizer does not have a padding token. Please select a token to use as `pad_token` `(tokenizer.pad_token = tokenizer.eos_token e.g.)` or add a new pad token via `tokenizer.add_special_tokens({'pad_token': '[PAD]'})`.
```

docs/source/quickstart.mdx
```diff
@@ -34,15 +35,15 @@ ppo_trainer = PPOTrainer(config, model, model_ref, tokenizer)
...
+query_tensor = tokenizer.encode(query_txt, return_tensors="pt").to(model.pretrained_model.device)
 
 # 4. generate model response
 response_tensor  = respond_to_batch(model, query_tensor)
 response_txt = tokenizer.decode(response_tensor[0,:])
 
 # 5. define a reward for response
 # (this could be any reward such as human feedback or output from another model)
-reward = [torch.tensor(1.0)]
+reward = [torch.tensor(1.0, device=model.pretrained_model.device)]
```
vwxyzjn, line 46: Without the `, device=model.pretrained_model.device` the code still works but the reward must be transferred to the device somewhere in the trainer.

docs/source/quickstart.mdx
```diff
@@ -19,13 +19,14 @@ The following code illustrates the steps above.
 # 0. imports
 import torch
 from transformers import GPT2Tokenizer
-from trl import PPOTrainer, PPOConfig, AutoModelForCausalLMWithValueHead, create_reference_model
+from trl import PPOTrainer, PPOConfig, AutoModelForCausalLMWithValueHead
 from trl.core import respond_to_batch
```
lvwerra, line 23: we should also replace `respond_to_batch` with `ppo_trainer.generate`. could you test if that works?

docs/source/quickstart.mdx
younesbelkada: Awesome thanks for adding this!

docs/source/quickstart.mdx
vwxyzjn: Thanks for the comment. I replaced it with `ppo_trainer.generate` as follows, but I have to use `query_tensor[0]` instead of `query_tensor`. 

```python
response_tensor = ppo_trainer.generate(query_tensor[0], return_prompt=False)
response_txt = tokenizer.decode(response_tensor[0,:])
```

Furthermore it gives two extra warnings:

```
The attention mask and the pad token id were not set. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.
Setting `pad_token_id` to `eos_token_id`:50256 for open-end generation.
/admin/home/costa/.pyenv/versions/trl/lib/python3.8/site-packages/transformers/generation/utils.py:1346: UserWarning: Using `max_length`'s default (20) to control the generation length. This behaviour is deprecated and will be removed from the config in v5 of Transformers -- we recommend using `max_new_tokens` to control the maximum length of the generation.
```


I then followed the example in sentiment analysis to do the following

```
generation_kwargs = {
    "min_length": -1,
    "top_k": 0.0,
    "top_p": 1.0,
    "do_sample": True,
    "pad_token_id": tokenizer.eos_token_id,
}
output_min_length = 4
output_max_length = 16
output_length_sampler = LengthSampler(output_min_length, output_max_length)
response_tensor = ppo_trainer.generate([item for item in query_tensor], return_prompt=False, length_sampler=output_length_sampler, **generation_kwargs)
response_txt = tokenizer.decode(response_tensor[0])
```

which removes the warning. 

Please let me know which one do you prefer :)

lvwerra: Oh yes, you can add `max_new_tokens=10` (or whatever number) to the generation kwargs.

vwxyzjn:
Thanks @lvwerra for the review. I added an end-to-end test as well.

tests/test_e2e.py
```diff
@@ -0,0 +1,9 @@
+import subprocess
+
+
+def test_dqn_jax():
```
lvwerra, line 4: yay, we support Jax now? :D

lvwerra (approved):
LGTM! 🚀  just one small comment!

tests/test_e2e.py
vwxyzjn: Whooops :) Fixed now

younesbelkada (commented):
Thanks a lot for the great work and welcome to TRL! 🚀 
We just need to address:
```
examples/hello_world.py:6:1: F401 'trl.core.LengthSampler' imported but unused
examples/hello_world.py:6:1: F401 'trl.core.respond_to_batch' imported but unused
```
So that the code quality checker will be happy - you can just do
```
make style && make quality
```
After that we can merge the PR! 💪 
Nice test btw :D !

younesbelkada (approved):
Awesome work!! 🚀

PR #423, opened 2023-06-09
KL Estimator

vwxyzjn:
This PR uses the better KL estimator from http://joschu.net/blog/kl-approx.html. The estimator should have a lower variance and is unbiased.


vwxyzjn:
Closing the issue for now.

PR #425, opened 2023-06-09
[CI] Cache entire environement

thomasw21:


thomasw21:
This might be a bad idea as the cache required for this to work is an order of magnitude bigger.

PR #373, opened 2023-05-17
Multi adapter RL (MARL) - a single model for RM & Value Head

younesbelkada:
# What does this PR do?

This PR introduces a new concept (initiated by @pacman100 & @philschmid) - were you would use a single base model to apply both RM & PPO leveraging multiple adapters.

image

Users needs first to train RM adapters using PEFT (script will be provided soon), then push these adapters either on the Hub or save them locally.

After that, just load the RM adapter using the flag `reward_modeling_adapter_id` when creating a `PPOTrainer`. Note that this feature is exclusive the fine-tuning PPO with adapters 

```python
model = AutoModelForCausalLMWithValueHead.from_pretrained(
    model_name,
    load_in_8bit=True,
    device_map={"": 0},
    peft_config=lora_config,
    reward_modeling_adapter_id=rm_adapter_id,
)
```

Then at rollout time call `ppo_trainer.model.compute_reward_score` to compute the raw reward scores of the RM. That's all you need!

With that you could fine-tune llama-7b on a single Nvidia T4 using RLHF (with 8bit base model loading) - meaning this would unlock many possibility for users and practitioneers.

## TODOs

- [x] nice CI tests
- [x] precise docs

cc @lvwerra @lewtun

PR #428, opened 2023-06-13
Multi Adapter Value Function 

mnoukhov:
PR that extends #373 to also make the value function a separate model. Specifically, add a new set of LoRA parameters for the value function and learn them completely separately from the policy instead of making the value function a separate head on the policy. By default, also initialize the value function with the policy parameters to bootstrap it.

The upside is the value function is not affected by the policy and alignment tax should be reduced. The downside is you need two full forward passes to get the policy and value logits. 

Current issues
- I implemented this in StackLLaMA but #373 uses a different example
- Can't be merged into #373 because I rebased on the main branch
- Getting an error `path/to/torch/utils/checkpoint.py:31: UserWarning: None of the inputs have requires_grad=True. Gradients will be None` that is also present in #373, I believe


mnoukhov:
Yep that's what I was thinking too. Once your branch is moved, I'll edit this to change the base branch to main and change from draft to real PR. I just started this as a draft and set your branch as the base branch to make my changes more clear i.e. setting `requires_grad=False` in https://github.com/lvwerra/trl/pull/428/files#diff-be4f28845d341065eb254da5db9fa8fee2e6b5347071ef98448f27acd70fe8a7R461-R463. 

But looking at the Files Changed actually doesn't correctly show the diff between my branch and yours for some reason. Let me know if you'd prefer if I rebased on your branch or did something else to make this draft more useful

PR #436, opened 2023-06-14
Multi Adapter RL - fixes

mnoukhov:
fixes to #373 

@younesbelkada I decided to move all my multi-adapter fixes into this PR so that you don't need to re-implement the changes yourself. This PR doesn't have any of the multi-adapter value function things from #428, only fixes within the scope of #373

the main changes are 
- add the reward adapter before calling `init` so that all the modules are already created
  - makes adding the reward adapter a class method 
  - this feels like it better follows the design structure I've seen in other huggingface PretrainedModels but I'm happy to change it back
- refactor `compute_reward_score` to use `model.forward` so it works seamlessly with accelerate in multi-gpu, this can be changed to just `accelerator.unwrap_model` otherwise
- correctly set the `model.score` layer's dtype
- set reward model lora parameters to `requires_grad = False` by using `inference_mode = True` in the `rm_adapter_peft_config`


mnoukhov:
@younesbelkada the fixes are now in a good state

let me know how you feel about the refactoring of `compute_reward_score` particularly the changes to the `forward` method to allow for `use_score = True`. This was added to avoid doing `accelerator.unwrap_model ... compute_reward_score`. I'm not too familiar with `accelerate` but assumed that this would be able to leverage it for better parallism. If that isn't the case, then I'm happy to change it back and make a simpler `compute_reward_score` method that leaves the original work intact

The other thing to check would be how I'm calculating the score now. I switched to just using the code from [LlamaForSequenceClassification.forward](https://github.com/huggingface/transformers/blob/v4.29.1/src/transformers/models/llama/modeling_llama.py#L800), as that would directly correspond to the way the reward model was trained. Because of this, I return a scalar per example instead of a matrix and don't index it with `[-1,1]` as you do [here](https://github.com/lvwerra/trl/blob/add-multiple-adapters/examples/multi-adapter-rl/rl_finetuning.py#L112). I wasn't really sure why this was necessary but let me know if I'm missing something.

younesbelkada:
Hi @mnoukhov 
The PR got closed by mistake, as I just merged #373 
Can you open a new PR on main with the same changes ? 🙏 Thanks a lot !

PR #444, opened 2023-06-16
[`SFTTrainer`] Fix non packed dataset

younesbelkada:
# What does this PR do?

This PR properly educates users on how to correctly use `formatting_func` method when someone uses a non-packed dataset. 
Since the dataset processing calls `dataset.map(xxx, batched=True)` under the hood, it is important to return an array of processed texts to properly process all texts from the dataset example batch, otherwise it will lead to silent bugs that are hard to understand such as the one described in https://github.com/lvwerra/trl/issues/439

```python
from trl import SFTTrainer
import transformers

dataset = load_dataset("tatsu-lab/alpaca", split="train")

model = transformers.AutoModelForCausalLM.from_pretrained("facebook/opt-350m")
tokenizer = transformers.AutoTokenizer.from_pretrained("facebook/opt-350m")

def formatting_prompts_func(examples):
    output_text = []
    for i in range(len(examples["instruction"])):
        instruction = examples["instruction"][i]
        input_text = examples["input"][i]
        response = examples["output"][i]

        if len(input_text) >= 2:
            text = f'''Below is an instruction that describes a task, paired with an input that provides further context. Write a response that appropriately completes the request.
            
            ### Instruction:
            {instruction}
            
            ### Input:
            {input_text}
            
            ### Response:
            {response}
            '''
        else:
            text = f'''Below is an instruction that describes a task, paired with an input that provides further context. Write a response that appropriately completes the request.
            
            ### Instruction:
            {instruction}
            
            ### Response:
            {response}
            '''
        output_text.append(text)

    return output_text

trainer = SFTTrainer(
    model,
    tokenizer=tokenizer,
    train_dataset=dataset,
    max_seq_length=256,
    packing=False,
)

trainer.train()
```

The PR adds a sanity check when processing the dataset, and adds the argument `padding=True,` to always return a sequence of length `max_seq_len` and correctly appends the attention mask to the output dataset as well.

### DIFF
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -65,8 +65,11 @@ Let us assume your dataset has two fields, `question` and `answer`. Therefore yo
 ```python
 ...
 def formatting_prompts_func(example):
-    text = f"### Question: {example['question']}\n ### Answer: {example['answer']}"
-    return text
+    output_texts = []
+    for i in range(len(example['question'])):
+        text = f"### Question: {example['question'][i]}\n ### Answer: {example['answer'][i]}"
+        output_texts.append(text)
+    return output_texts
 
 trainer = SFTTrainer(
     model,
@@ -76,6 +79,7 @@ trainer = SFTTrainer(
 
 trainer.train()
 ```
+To preperly format your input make sure to process all the examples by looping over them and returning a list of processed text. Check out a full example on how to use SFTTrainer on alpaca dataset [here]()
 
 ### Packing dataset ([`ConstantLengthDataset`])
 
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -79,7 +79,7 @@ trainer = SFTTrainer(
 
 trainer.train()
 ```
-To preperly format your input make sure to process all the examples by looping over them and returning a list of processed text. Check out a full example on how to use SFTTrainer on alpaca dataset [here]()
+To preperly format your input make sure to process all the examples by looping over them and returning a list of processed text. Check out a full example on how to use SFTTrainer on alpaca dataset [here](https://github.com/lvwerra/trl/pull/444#issue-1760952763)
 
 ### Packing dataset ([`ConstantLengthDataset`])
 
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -276,27 +276,43 @@ class SFTTrainer(Trainer):
         self, tokenizer, dataset, dataset_text_field, max_seq_len, formatting_func=None
     ):
         use_formatting_func = formatting_func is not None and dataset_text_field is None
+        self._dataset_sanity_checked = False
 
         # Inspired from: https://huggingface.co/learn/nlp-course/chapter7/6?fw=pt
         def tokenize(element):
+            input_batch = []
+            attention_masks = []
+
             outputs = tokenizer(
                 element[dataset_text_field] if not use_formatting_func else formatting_func(element),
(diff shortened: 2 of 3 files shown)


docs/source/sft_trainer.mdx
```diff
@@ -76,6 +79,7 @@ trainer = SFTTrainer(
 
 trainer.train()
 ```
+To preperly format your input make sure to process all the examples by looping over them and returning a list of processed text. Check out a full example on how to use SFTTrainer on alpaca dataset [here]()
```
younesbelkada, line 82:
```suggestion
To preperly format your input make sure to process all the examples by looping over them and returning a list of processed text. Check out a full example on how to use SFTTrainer on alpaca dataset [here](https://github.com/lvwerra/trl/pull/444#issue-1760952763)
```

docs/source/sft_trainer.mdx
```diff
@@ -65,8 +65,11 @@ Let us assume your dataset has two fields, `question` and `answer`. Therefore yo
 ```python
 ...
 def formatting_prompts_func(example):
-    text = f"### Question: {example['question']}\n ### Answer: {example['answer']}"
-    return text
+    output_texts = []
+    for i in range(len(example['question'])):
+        text = f"### Question: {example['question'][i]}\n ### Answer: {example['answer'][i]}"
+        output_texts.append(text)
+    return output_texts
```
vwxyzjn, line 72: Oh interesting. So previously, we were dumping an entire dataset to the prompt?

younesbelkada: Exactly yes :D the previous examples on the documentation were wrong and we were dumping the entire mini-batches when processing the dataset .. :/

trl/trainer/sft_trainer.py
```diff
@@ -276,27 +276,43 @@ def _prepare_non_packed_dataloader(
...
 
         # Inspired from: https://huggingface.co/learn/nlp-course/chapter7/6?fw=pt
         def tokenize(element):
+            input_batch = []
+            attention_masks = []
+
             outputs = tokenizer(
                 element[dataset_text_field] if not use_formatting_func else formatting_func(element),
                 truncation=True,
+                padding=True,
```
PhilDakin, line 289: @younesbelkada  this code is still incorrect - consider the case where **all** samples in the dataset are less than `max_seq_len`. Each batch will be padded to the largest element in the batch, but no data will pass the `if length == max_seq_len` check below.

Perhaps:
```
                padding='max_length',

younesbelkada: Yes you are correct thanks a lot for flagging, do you want to open a PR for that? happy to do it otherwise

ahmadmustafaanis:
Will this fine-tune the complete model end-to-end, or will this example fine-tune just a portion of it, like in Lora?

lvwerra:
The above example will train the full model but there are also options to use LoRA.

hy-chen:
Is this merged in main and then reverted? because padding is still False in _prepare_non_packed_dataloader in main (0.7.9)

hy-chen:
> Is this merged in main and then reverted? because padding is still False in _prepare_non_packed_dataloader in main (0.7.9)

Actually padding was turned off by this PR: https://github.com/huggingface/trl/pull/512

Now running SFT on alpaca gives 
`ValueError: Unable to create tensor, you should probably activate truncation and/or padding with 'padding=True' 'truncation=True' to have batched tensors with the same length. Perhaps your features (`labels` in this case) have excessive nesting (inputs type `list` where type `int` is expected).`

haochuan-li:
Any update on this issue?

PR #446, opened 2023-06-16
Fix implicit exclusion of training/evaluation data that is under max_seq_len tokens when packing=False.

PhilDakin:
@younesbelkada currently the code implicitly eliminates all data under the passed max length.

This seems wrong.


PhilDakin:
Actually, I see this should already be fixed via the inclusion of padding here: https://github.com/lvwerra/trl/pull/444/files#diff-67e157adfcd37d677fba66f610e3dfb56238cc550f221e8683fcfa0556e0f7caR289.

Closing.

PR #448, opened 2023-06-20
Pre-commit

vwxyzjn:
You can reproduce the existing pipeline with `pre-commit run --all-files` (note that pre-commit however only applies to git's staged and already committed files). I didn't fix the typo pointed out by `codespell` because it might create too many merge conflicts with existing PRs.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
-.PHONY: test commit
+.PHONY: test precommit
 
 check_dirs := examples tests trl
 
 test:
 	python -m pytest -n auto --dist=loadfile -s -v ./tests/
 
-commit:
+precommit:
 	pre-commit run --all-files
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,6 @@
-.PHONY: quality style test
+.PHONY: test
 
 check_dirs := examples tests trl
 
 test:
 	python -m pytest -n auto --dist=loadfile -s -v ./tests/
-
-quality:
-	black --check --line-length 119 --target-version py38 $(check_dirs)
-	isort --check-only $(check_dirs)
-	flake8 $(check_dirs)
-
-style:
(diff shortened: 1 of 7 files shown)


lvwerra:
Can you elaborate how to use it in a workflow? Can you register this as a git hook or would you run it manually?

vwxyzjn:
There are two workflows to use it

1. A lot of people use `pre-commit install` to register it with git hooks so it runs automatically when you commit. Personally I don't like it because it doesn't let you confirm your changes
2. I like to use `pre-commit run --all-files` as a drop-in replacement for `make format`. The only difference is currently we manually specify `check_dirs`, whereas pre-commit's `check_dirs` will include all the committed and staged files (and don't include the untracked files).

lvwerra:
Sounds good to me. What to you think about adding `pre-commit run --all-files` to the Makefile e.g. as `make format`. A few commands less to remember :) 

What do you think @younesbelkada ?

younesbelkada:
the changes sounds good, I also agree we should add something in the lines that @lvwerra mentioned on the Makefile

vwxyzjn:
@lvwerra @younesbelkada good idea! Done with `make commit`

Makefile
```diff
@@ -1,15 +1,9 @@
...
-quality:
-	black --check --line-length 119 --target-version py38 $(check_dirs)
-	isort --check-only $(check_dirs)
-	flake8 $(check_dirs)
-
-style:
-	black --line-length 119 --target-version py38 $(check_dirs)
-	isort $(check_dirs)
\ No newline at end of file
+commit:
```
lvwerra, line 8: small nit, let's rename it otherwise it might sound like the actually commits something :) 

```suggestion
precommit:
```

vwxyzjn: Done!

younesbelkada (approved):
Thank you @vwxyzjn !

PR #452, opened 2023-06-21
[`ConstantLengthDataset`] Fix packed dataset issue

younesbelkada:
# What does this PR do?

Fixes https://github.com/lvwerra/trl/issues/450 

As reported in that issue, there is an inconsistent behaviour when using `packing=True` and the number of iterations showed in the logs of training. In fact, the inconsistency comes from the fact that in the packed case, sequences are packed all together on the fly until `max_seq_length` is reached, leading to reducing the number of total expected samples.

The fix is to force-set the `infinite` argument to `True` if a user explicity decides to train a model with `max_steps` strategy but properly warn them that the argument has been overriden and warn the user when the dataset is moved to the next iteration.

In the case of epochs training strategy, we should force-set infinite argument to False, otherwise the training will run forever.

Added also nice CI tests to make sure that this behavior will stay consistent in future commits.  Added also a line in the documentation.

cc @lvwerra @vwxyzjn

### DIFF
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -98,6 +98,8 @@ trainer = SFTTrainer(
 trainer.train()
 ```
 
+Note that if you use a packed dataset and if you pass `max_steps` in the training arguments you will probably train your models for more than few epochs. 
+
 #### Customize your prompts using packed dataset
 
 If your dataset has several fields that you want to combine, for example if the dataset has `question` and `answer` fields and you want to combine them, you can pass a formatting function to the trainer that will take care of that. For example:
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -98,7 +98,7 @@ trainer = SFTTrainer(
 trainer.train()
 ```
 
-Note that if you use a packed dataset and if you pass `max_steps` in the training arguments you will probably train your models for more than few epochs. 
+Note that if you use a packed dataset and if you pass `max_steps` in the training arguments you will probably train your models for more than few epochs, depending on the way you have configured the packed dataset and the training protocol. Double check that you know and understand what you are doing. 
 
 #### Customize your prompts using packed dataset
 
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -480,7 +480,7 @@ class SFTTrainerTester(unittest.TestCase):
             self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
 
             # make sure the trainer did 5 steps
-            self.assertTrue("pytorch_model.bin" in os.listdir(tmp_dir + "/checkpoint-2"))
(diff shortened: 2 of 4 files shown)


docs/source/sft_trainer.mdx
```diff
@@ -98,6 +98,8 @@ trainer = SFTTrainer(
 trainer.train()
 ```
 
+Note that if you use a packed dataset and if you pass `max_steps` in the training arguments you will probably train your models for more than few epochs.
```
lvwerra, line 101: That's not really true, right? It could also be shorter - depends on the dataset and max_steps you choose :)

trl/trainer/sft_trainer.py
```diff
@@ -219,6 +219,17 @@ def __init__(
             preprocess_logits_for_metrics=preprocess_logits_for_metrics,
         )
 
+        if self.args.max_steps > 0 and packing:
+            warnings.warn(
+                "You passed `packing=True` to the SFTTrainer, and you are training your model with `max_steps` strategy. Will force-set `infinite` argument of the packed dataset to `True` "
```
lvwerra, line 224: Since most users might not know the infinite setting maybe we can say something like:
```suggestion
                "You passed `packing=True` to the SFTTrainer, and you are training your model with `max_steps` strategy. The dataset will be iterated until the `max_steps` are reached."
```

trl/trainer/sft_trainer.py
```diff
@@ -219,6 +219,17 @@ def __init__(
...
+            )
+            self.train_dataset.infinite = True
+        elif self.args.max_steps == -1 and packing:
+            warnings.warn(
```
lvwerra, line 228: I think in that case we don't have to warn - the model will train as expected, no?

trl/trainer/utils.py
```diff
@@ -244,6 +244,7 @@ def __iter__(self):
                 except StopIteration:
                     if self.infinite:
                         iterator = iter(self.dataset)
+                        warnings.warn("The dataset reached end. The iterator is reset")
```
lvwerra, line 247:
```suggestion
                        warnings.warn("The dataset reached end and the iterator is reset to the start.")
```

tests/test_sft_trainer.py
```diff
@@ -423,6 +423,65 @@ def test_data_collator_completion_lm(self):
         result_text = self.tokenizer.decode(batch["input_ids"][0, last_pad_idx + 1 :])
         self.assertTrue(result_text == "I have not been masked correctly.")
 
+    def test_sft_trainer_infinite_with_model(self):
```
lvwerra, line 426: What do you think about double checking here for both cases if we trained as many steps as we expect?

trl/trainer/sft_trainer.py
younesbelkada: Makes sense, removed it

lvwerra (approved):

PR #457, opened 2023-06-22
Make shuffle optional

lopez-hector:
Added parameter that allows you to turn off shuffling.

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -187,6 +187,8 @@ class ConstantLengthDataset(IterableDataset):
                 Number of characters per token used to estimate number of tokens in text buffer.
             eos_token_id (`int`, *optional*, defaults to `0`):
                 Id of the end of sequence token if the passed tokenizer does not have an EOS token.
+            shuffle ('bool', *optional*, defaults to True)
+                Shuffle the examples before they are returned
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


trl/trainer/utils.py
```diff
@@ -187,6 +187,8 @@ class ConstantLengthDataset(IterableDataset):
                 Number of characters per token used to estimate number of tokens in text buffer.
             eos_token_id (`int`, *optional*, defaults to `0`):
                 Id of the end of sequence token if the passed tokenizer does not have an EOS token.
+            shuffle ('bool', *optional*, defaults to True)
```
younesbelkada, line 190:
```suggestion
            shuffle ('bool', *optional*, defaults to `False`)
```

trl/trainer/utils.py
```diff
@@ -200,6 +202,7 @@ def __init__(
         num_of_sequences=1024,
         chars_per_token=3.6,
         eos_token_id=0,
+        shuffle=True,
```
younesbelkada, line 205:
```suggestion
        shuffle=False,
```

younesbelkada (commented):
Thank you for your great suggestion!
Let's put the default to False to avoid breaking changes for users. What do you think ?

younesbelkada (approved):
Sorry Ignore my comments, this looks great to me

PR #464, opened 2023-06-23
Debug the tortuous logic in `_prepare_dataset` function

BeibinLi:
There are two issues with the previous `_prepare_dataset` function.

1. Tortuous and burdensome logic: the `is_already_dataset` variable is confusing and not helpful. So, remove it.
2. The comments and the logics do not match. 

For instance, in the previous version, the comments said "check if torch dataset ... and do nothing". However, when "dataset" is a torch.utils.data.Dataset and `packing = True`? It will still move into the _prepare_non_packed_dataloader(...) function call. 

The corrected version will do nothing if the dataset is already a torch dataloader/dataset/ConstantLengthDataset.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -232,10 +232,7 @@ class SFTTrainer(Trainer):
             raise ValueError("The dataset should not be None")
             
         # check if torch dataset / dataloader and do nothing
-        if isinstance(dataset,
-                      (torch.utils.data.IterableDataset, 
-                       torch.utils.data.Dataset, 
-                       ConstantLengthDataset)):
+        if isinstance(dataset, (torch.utils.data.IterableDataset, torch.utils.data.Dataset, ConstantLengthDataset)):
             return dataset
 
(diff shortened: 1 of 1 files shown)


younesbelkada (commented):
Thanks a lot @BeibinLi for proposing this refactor, can you run the styling checks in order to trigger the CI tests? 
```
make style && make quality
```
Or alternatively
```
make precommit
```
If you cloned the repo after: https://github.com/lvwerra/trl/pull/448
After that we should be good to merge

BeibinLi:
> Thanks a lot @BeibinLi for proposing this refactor, can you run the styling checks in order to trigger the CI tests?
> 
> ```
> make style && make quality
> ```
> 
> Or alternatively
> 
> ```
> make precommit
> ```
> 
> If you cloned the repo after: #448 After that we should be good to merge

Thanks! I just re-ran the styling checks and pushed the latest code.

younesbelkada (approved):
Thanks a lot @BeibinLi !
This is much more cleaner now!

PR #468, opened 2023-06-26
[`CI`] Fix CI RM

younesbelkada:
Currently a CI is failing (weird that it was not failing in the previous PRs - probably a recent release introduced a silent bug?)

The fix seems to simply correct the `gradient_accumulation steps` value. In fact putting `gradient_accumulation_steps > max_steps ` will lead to gradients that gets never updated

Related: https://github.com/lvwerra/trl/pull/416

Let's see if this fixes it 🤞 

cc @lvwerra


lvwerra (approved):

lvwerra:
LGTM! 🚀

Issue #471, opened 2023-06-27
Reproducing StackLLaMA with Multi-Adapters


mnoukhov:
Now that we've repro'd the StackLLaMA results in #401, I think it would be useful to repro StackLLaMA with the more compute-efficient Multi-Adapter setup #373 to see if we can reduce compute requirements and also as a sort of integration test for Multi-Adapter

mnoukhov:
The multi-adapter fixes make the code better and everything runs but it doesn't repro StackLLaMA. Below is the regular run in brown and the multi-adapter run in yellow. 

The initial values are very similar, which is a good sign, but the training doesn't proceed as expected. 

![image](https://github.com/lvwerra/trl/assets/3391297/4a79c74f-fffd-4d5c-ab4f-6aff92c861d4)

These issues could be related to the KL estimation and perhaps #423 could help. 
Alternatively, the issue could be the difference in quanitization between reward model and rlhf training. Reward modelling uses a `bf16` base LLaMA model whereas we do RLHF training with an `int8` LLaMA model. Currently, I am retraining a reward model with `int8` to see if that's the issue.  


Also, note that we do see reduced memory usage with Multi-Adapter but sadly memory spikes seem just as high

![image](https://github.com/lvwerra/trl/assets/3391297/21f5f8b8-89c1-4ca0-89ca-

mnoukhov:
Hey @dh2shin, if this isn't specifically related to multi-adapter, can you move the discussion back to #401

dh2shin:
@mnoukhov Hi Michael, how has the int8 experiments gone for the reward model? Are you getting decent accuracy.

mnoukhov:
I was running into issues with `ddp` and `peft` (#480 ) but even after getting around them, I haven't managed to make things work. The main thing I found was that you have to train a new reward model adapter on top of the `int8`.

Even after all this, I'm not managing to repro the original results. There could be a regression in the codebase or it could just be too much instability. I'm still investigating but moving to a smaller scale task in the meantime

Nipi64310:
Hi @mnoukhov ,
[...]

PR #472, opened 2023-06-27
Multi Adapter Fixes

mnoukhov:
Multi-Adapter RL, as it currently is, does not reproduce StackLLaMA results and has some design choices that can be improved. This PR fixes some basics and improves the design. This is the first step in the StackLLaMA repro #471 but I won't add more to the current PR to keep it manageable. 

the main changes are
  - add the reward adapter before calling init so that all the modules are already created
      - makes adding the reward adapter a class method
      - this feels like it better follows the design structure I've seen in other huggingface PretrainedModels but I'm happy to change it back
- refactor `compute_reward_score` to use model.forward so it works seamlessly with accelerate in multi-gpu, this can be changed to just accelerator.unwrap_model otherwise
     - I assume that we want to parallelize  
- correctly set the `model.score` layer's dtype
    - otherwise fails if it isn't the same dtype as the `model.pretrained_model`
- set reward model lora parameters to `requires_grad = False` by using `inference_mode = True` in the rm_adapter_peft_config


younesbelkada (commented):
Hi @mnoukhov 
Thanks a lot for the PR ! currently a CI test fails:
```bash
FAILED tests/test_ppo_trainer.py::PPOTrainerTester::test_peft_model_ppo_adapter_rm_trainer - TypeError: add_and_load_reward_modeling_adapter() missing 1 required positional argument: 'adapter_model_id'
```
Can you please double check? 🙏 Thanks!

mnoukhov:
Sorry I forgot about this PR. Re-merged with main and re-opening. Should have fixed the test error

mnoukhov:
@younesbelkada can you re-open this or should I make a new PR?

PR #474, opened 2023-06-27
Update sft_trainer.py

### LINKED ISSUE #473: SFTTrainer: passing a PeftModel as a model argument without passing the config breaks
[https://github.com/lvwerra/trl/blob/main/trl/trainer/sft_trainer.py#L127C36-L127C47](https://github.com/lvwerra/trl/blob/main/trl/trainer/sft_trainer.py#L127C36-L127C47) Requires that you pass a PEFT config even though you've already instantiated the PeftModel yourself, otherwise it tries ignores the if and tries to build a regular transformers model at https://github.com/lvwerra/trl/blob/main/trl/trainer/sft_trainer.py#L148 taking your PeftModel as argument, which makes no sense.

I suggest adding a check at https://github.com/lvwerra/trl/blob/main/trl/trainer/sft_trainer.py#L147 to make sure that not isinstance(model, PeftModel):

```python
elif not isinstance(model, PreTrainedModel) and not isinstance(model, PeftModel):
```

JulesGM:
Allows the user to give their own peft model arg. https://github.com/lvwerra/trl/issues/473

Fixes #473

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -144,7 +144,7 @@ class SFTTrainer(Trainer):
 
             if callbacks is None:
                 callbacks = [PeftSavingCallback]
-        elif not isinstance(model, PreTrainedModel) and not isinstance(model, PeftModel):
+        elif not isinstance(model, (PreTrainedModel, PeftModel)):
             model = AutoModelForCausalLM.from_pretrained(model)
 
         if tokenizer is None:
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -144,7 +144,7 @@ class SFTTrainer(Trainer):
 
             if callbacks is None:
                 callbacks = [PeftSavingCallback]
-        elif not isinstance(model, PreTrainedModel):
+        elif not isinstance(model, PreTrainedModel) and not isinstance(model, PeftModel):
             model = AutoModelForCausalLM.from_pretrained(model)
 
         if tokenizer is None:


trl/trainer/sft_trainer.py
```diff
@@ -144,7 +144,7 @@ def __init__(
 
             if callbacks is None:
                 callbacks = [PeftSavingCallback]
-        elif not isinstance(model, PreTrainedModel):
+        elif not isinstance(model, PreTrainedModel) and not isinstance(model, PeftModel):
```
younesbelkada, line 147:
```suggestion
        elif isinstance(model, str):
```

younesbelkada (commented):
Great catch! Thanks for fixing, proposed an alternative, what do you think?

JulesGM:
The goal of still using the SFTTrainer is to use the peft saving callback, and maybe the dataset saving stuff I guess

trl/trainer/sft_trainer.py
JulesGM: what about

```python
elif not isinstance(model, (PretrainedModel, PeftModel):
```
which is the same thing but more compact I guess. 

`str` fails on `pathlib.Path` to a local save of a pretrained model, though I don't know if that's supported to begin with

JulesGM: but I'm fine with that too

younesbelkada: I see, can you change it with the suggesiton you made then ? 🙏 Then we can merge I think!

JulesGM: ok I pushed the suggestion

younesbelkada (approved):
Thank you for the clean work !

PR #476, opened 2023-06-28
Disable mlm by default in DataCollatorForCompletionOnlyLM, add ignore_index and docstring

BramVanroy:
As discussed [here](https://github.com/lvwerra/trl/pull/445#issuecomment-1601544929), `DataCollatorForCompletionOnlyLM ` inherits from the MLM data collator but that means that `mlm` is set to True by default. In reality that does not matter (because the appropriate methods are overridden) but it can still be confusing for the user. So now `mlm` is explicitly set to False (with the option to be changed by the user).

The user now also has the option to decide which index to use to ignore `ignore_index`.

A docstring is added for DataCollatorForCompletionOnlyLM, which was missing. I urge future PRs to always include good documentation and docstrings, otherwise the overview gets lost quickly.

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -60,12 +60,15 @@ class DataCollatorForCompletionOnlyLM(DataCollatorForLanguageModeling):
     Args:
         response_template (`str`): the template form that indicates the start of the response, typically something like
             '### Response:\n'
+        mlm (`bool`, *optional*, defaults to `False`): Whether or not to use masked language modeling in the underlying
+            `DataCollatorForLanguageModeling` class. Note that this option currently has no effect but is present
+             for flexibility and backwards-compatibility.
         ignore_index (`int`, *optional*, defaults to `-100`):
             The index to use to ignore the initial tokens with
     """
 
-    def __init__(self, response_template: str, *args, ignore_index: int = -100, **kwargs):
-        super().__init__(*args, mlm=False, **kwargs)
+    def __init__(self, response_template: str, *args, mlm: bool = False, ignore_index: int = -100, **kwargs):
+        super().__init__(*args, mlm=mlm, **kwargs)
         self.response_template = response_template
         self.ignore_index = ignore_index
 
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -63,6 +63,7 @@ class DataCollatorForCompletionOnlyLM(DataCollatorForLanguageModeling):
         ignore_index (`int`, *optional*, defaults to `-100`):
             The index to use to ignore the initial tokens with
     """
+
     def __init__(self, response_template: str, *args, ignore_index: int = -100, **kwargs):
         super().__init__(*args, mlm=False, **kwargs)
         self.response_template = response_template
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -52,9 +52,24 @@ class FixedKLController:
 
 
 class DataCollatorForCompletionOnlyLM(DataCollatorForLanguageModeling):
-    def __init__(self, response_template, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    """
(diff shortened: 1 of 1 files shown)


trl/trainer/utils.py
```diff
@@ -81,7 +96,7 @@ def torch_call(self, examples: List[Union[List[int], Any, Dict[str, Any]]]) -> D
             response_token_ids_end_idx = response_token_ids_start_idx + len(response_token_ids)
 
             # Make pytorch loss function ignore all tokens up through the end of the response key
-            labels[i, :response_token_ids_end_idx] = -100
+            labels[i, :response_token_ids_end_idx] = self.ignore_index
```
lvwerra, line 99: not sure if this makes sense here, since `-100` is the default value ignored by torch loss functions.

trl/trainer/utils.py
```diff
@@ -52,9 +52,24 @@ def update(self, current, n_steps):
...
-        super().__init__(*args, **kwargs)
+    """
+    Data collator used for completion tasks. It ensures that all the tokens of the labels are set to an 'ignore_index'
+     up to the prompt response template tokens ('response_template'). This ensure that the loss is only
+     calculated on the completion of the reponse.
+
+    Args:
+        response_template (`str`): the template form that indicates the start of the response, typically something like
+            '### Response:\n'
+        mlm (`bool`, *optional*, defaults to `False`):
```
lvwerra, line 63: we don't support mlm so I don't think we need to support it here. so i would just hardcode it to `False`

lvwerra (commented):
Thanks @BramVanroy for fixing - I left two small comments!

trl/trainer/utils.py
lvwerra: Let's go with the hardcoded variant. I'll discuss with @younesbelkada when he gets back from vacation in case he feels there is a use-case where someone wants to use it.

lvwerra (approved):

BramVanroy:
@lvwerra It doesn't seem straightforward to do the quality fixes automatically from my end. The makefile contains a `precommit` key but that seems specific to Github (with the related yaml file). Unlike `transformers` I am missing a `make style`. The CONTRIBUTING guide mentions `make commit` but that command does not exist.

lvwerra:
Did you install the `pre-commit` library?

```bash
pip install pre-commit
make precommit
```

trl/trainer/utils.py
```diff
@@ -52,9 +52,22 @@ def update(self, current, n_steps):
...
+
+    Args:
+        response_template (`str`): the template form that indicates the start of the response, typically something like
+            '### Response:\n'
+        ignore_index (`int`, *optional*, defaults to `-100`):
+            The index to use to ignore the initial tokens with
+    """
+
+    def __init__(self, response_template: str, *args, ignore_index: int = -100, **kwargs):
+        super().__init__(*args, mlm=False, **kwargs)
```
lvwerra, line 68: E.g. here we pass the `mlm` explicitly which is the reason the tests fail:
https://github.com/lvwerra/trl/blob/2b531b92238f9fc40c6cc9cf27b8bbd3ae2727b1/trl/trainer/sft_trainer.py#L170

We should check if mlm is a key in `kwargs` and only update with `False` if not.

BramVanroy: Doesn't that make it the same as the initial PR (not hardcoded) but just less transparent?

Original:

```
    def __init__(self, response_template: str, *args, mlm: bool = False, ignore_index: int = -100, **kwargs):
```

Requested refactor:

```
    def __init__(self, response_template: str, *args, ignore_index: int = -100, **kwargs):
        super().__init__(*args, mlm=kwargs.get("mlm", False), **kwargs)
```

From a user/typing/IDE completion, I'd argue the first one is the better choice, no?

lvwerra: True, let's go with the original one. Sorry for going in a circle to find the best solution :)

BramVanroy: No problem ;-) Re-added

PR #479, opened 2023-06-29
correctly implement gradient checkpointing to multi-adapter example

mnoukhov:
pass `gradient_checkpointing` to both `prepare_model` and `TrainingArgs` 
switch from discontinued `int8` to new peft `kbit` api
pass `max_length` to `RewardTrainer`


younesbelkada (approved):
Thanks a lot for the fix!

PR #481, opened 2023-06-30
Fix: avoid filter out smaller examples than `max_seq_len`

mrm8488:
- Fix this issue: https://github.com/lvwerra/trl/issues/467

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -293,7 +293,7 @@ class SFTTrainer(Trainer):
             for length, input_ids, attention_mask in zip(
                 outputs["length"], outputs["input_ids"], outputs["attention_mask"]
             ):
-                if length == max_seq_len:
+                if length <= max_seq_len:
                     input_batch.append(input_ids)
                     attention_masks.append(attention_mask)
 


BramVanroy:
I agree that the current implementation seems wrong, but can't we get rid of [that whole loop](https://github.com/lvwerra/trl/blob/eb6eaa65cde6b5a5f58aa731abea1b48a75c96aa/trl/trainer/sft_trainer.py#L293-L298)? The tokenizer is already using `max_seq_len` as the `max_length`, so longer items will be truncated anyway: items that are longer should not even occur anymore.

mrm8488:
> I agree that the current implementation seems wrong, but can't we get rid of [that whole loop](https://github.com/lvwerra/trl/blob/eb6eaa65cde6b5a5f58aa731abea1b48a75c96aa/trl/trainer/sft_trainer.py#L293-L298)? The tokenizer is already using `max_seq_len` as the `max_length`, so longer items will be truncated anyway: items that are longer should not even occur anymore.

Think in the same way

lvwerra:
If the tokenizer already truncates longer examples and pads shorter ones we can just append without the check, no?

BramVanroy:
@lvwerra My point was that I do not think that we need the loop at all. Can't we just have this?

```python
outputs = tokenizer(
                element[dataset_text_field] if not use_formatting_func else formatting_func(element),
                truncation=True,
                padding=True,
                max_length=max_seq_len,
                return_overflowing_tokens=False,
                return_length=True,
            )

if use_formatting_func and not self._dataset_sanity_checked:
    if not isinstance(formatting_func(element), list):
        raise ValueError(
            "The `formatting_func` should return a list of processed strings since it can lead to silent bugs."
        )
    else:
        self._dataset_sanity_checked = True

return {"input_ids": outputs["input_ids"], "attention_mask": outputs["attention_mask"]}
```

Also, if I understand the tokenizer correctly, this will truncate longer sequences to `max_seq_len` and pad to the longest item _in the batch_ - so it is not padding to the `max_seq_len`. At least if I understand [this table](https://huggingface.co/docs/transformers/pad_truncation) correctly.

lvwerra:
Yes, the data collator will do the padding so I agree we can remove this loop altogether!

BramVanroy:
@lvwerra If the collator is _also_ doing padding, won't that lead to irregular padding?

So first, in the trainer, the dataset will be processed here and padding will be addeded depending on the batch. So for sample `n`, it will be padded up to the max length of a given batch. Let's say it ends up with 5 padding tokens.

But then, later on in the collator, the same sample might end up in a different batch. Which then, again will be padded up to the max_length of that batch -- but the max_length might be determined by an already padded sample. So if sample n is the longest sample in the batch, all other samples will be padded up to its length, even though the sample is artificially long with 5 padding tokens.

I'm not sure if it is clear what I am trying to say but it comes down to: doesn't this mean that batch-level padding occurs twice, which might be inconsistent between the trainer and the collator (like, shuffling or grouping by length).

lvwerra:
Oh didn't catch the padding in the tokenizer - indeed we shouldn't do it as the `Trainer` shuffles each epoch and we should apply padding on the fly with the data collator rather than doing it at the tokenization step.

BramVanroy:
@mrm8488 So maybe you could change the code in this PR?

1. Remove [the loop](https://github.com/lvwerra/trl/blob/eb6eaa65cde6b5a5f58aa731abea1b48a75c96aa/trl/trainer/sft_trainer.py#L293-L298) completely
2. Replace [this line](https://github.com/lvwerra/trl/blob/eb6eaa65cde6b5a5f58aa731abea1b48a75c96aa/trl/trainer/sft_trainer.py#L305) with the corresponding keys directly from `outputs`
3. Remove the padding and return length keys from [this line](https://github.com/lvwerra/trl/blob/eb6eaa65cde6b5a5f58aa731abea1b48a75c96aa/trl/trainer/sft_trainer.py#L276)

Thanks!

lvwerra:
Sounds good, just a small question: what do you mean by length keys?

BramVanroy:
> ```python
> return_length
> ```

I should have formatted that more clearly: `padding` and `return_length` parameters from the `tokenizer` call.

mrm8488:
> Exactly, @mrm8488 are you happy to add these changes?

Ofc, I will do it today! Thanks!

younesbelkada:
Thanks so much @mrm8488  !

DebarshiChanda:
Will this fix take time? I can take it up if needed

younesbelkada:
Thanks everyone for your contributions ! https://github.com/lvwerra/trl/pull/512 being merged we can clone this PR 
@mrm8488 @BramVanroy I have added you as co-authors for your help !

Issue #101, opened 2023-01-24
Spikes in PPO policy loss


lvwerra:
We sometimes experience huge loss spikes in the policy loss which either cause the training to fail or take a very long time to recover. It would be useful to investigate where they come from and how to mitigate them. cc @natolambert 

<img width="430" alt="Screenshot 2023-01-24 at 16 24 36" src="https://user-images.githubusercontent.com/8264887/214341251-ce24e511-89c1-49a2-b7e1-8d6e9abb8294.png">

<img width="1294" alt="Screenshot 2023-01-24 at 16 48 34" src="https://user-images.githubusercontent.com/8264887/214341267-287588c1-cda0-4fc4-80f0-5dddd5bfa712.png">

younesbelkada:
One idea could be that we don't mask out the logits corresponding to padding tokens, it is something I am having a look in https://github.com/lvwerra/trl/pull/100 - But I am not sure here if this is really the rootcause of this

natolambert:
Yeah, so something weird is going one with a simultaneous large drop in entropy, clip fraction, etc. Can we log the model outputs at that step? Is there any chance the model output gets stuck on something?

DaehanKim:
I corrected parsing pipeline output and loss spike still remains in sentiment-control notebook example. 
so there may be another reaseon for this unstability.

![image](https://user-images.githubusercontent.com/20675681/216659430-e828a210-203d-4b1e-a3c9-556653795ebb.png)

lvwerra:
Thanks @DaehanKim, yes there is an issue besides the order of the logits. I tracked it down to some changes done in #80 (no spikes at the beginning of the PR and spikes at time of merge) and I started tracking the issue down in #126. I'll report as well here if I figure it out!

lvwerra:
The issue with the loss spikes in the sentiment control notebook was that sometimes only a few new tokens would be generated (1-2) and this would cause the loss to spike. Not sure, yet, where exactly this behaviour comes from but we now know where to look: we can actively generate short sequences and investigate what causes the loss explosion.

tengxiaoliu:
I also experienced the spike loss in my case. I'm using the seq2seq t5 model as the backbone. The model is initialized with a supervised finetuned model. I find that the spike loss comes from steps that have a negative advantage and an extremely high ratio r(\theta). This falls in the situation 6 in the [figure](https://huggingface.co/blog/deep-rl-ppo) below.
[...]

PR #488, opened 2023-07-03
add ratio threshold to avoid spikes

lvwerra:
Investigating issue #101 further we found that they are usually caused by abnormally high ratios in the PPO loss. This adds a threshold to skip those batches. See below for a seed that had a spike before and after the fix:

Screenshot 2023-07-03 at 17 21 03
Screenshot 2023-07-03 at 17 20 39


vwxyzjn (approved):
This is a very nice seeding and repro thing you had in the PR description. The PR makes sense to me. Feel free to merge!

PR #493, opened 2023-07-04
FIX: contributing guidelines command

BramVanroy:
The contributing guideline mentions `make commit` but this should be `make precommit` as per the Makefile.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,6 +42,9 @@ Then before submitting your PR make sure the code quality follows the standards.
 make precommit
 ```
 
+Make sure to install `pre-commit` before running the command:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -39,7 +39,7 @@ make test
 Then before submitting your PR make sure the code quality follows the standards. You can run the following command to format:
 
 ```bash
-make commit
+make precommit
 ```
```
lvwerra, line 44: For some reason I can't suggest the closing of the code block, but you get the idea?
```suggestion

Make sure to install `pre-commit` before running the command:
```bash
pip install pre-commit
```

PR #499, opened 2023-07-06
Use `float` instead of `double` to avoid issues with MPS device

younesbelkada:
Addresses: https://github.com/lvwerra/trl/issues/440#issuecomment-1623160589 

We should use `float` instead of `double` to avoid issues with MPS device

cc @lvwerra


lvwerra:
Can you quickly run the IMDb example to check that we are not breaking anything?

lvwerra (approved):

younesbelkada:
Sure yes, running it now, will merge after few steps and making sure it converges

younesbelkada:
Here are the logs; https://wandb.ai/younesbelkada/trl/runs/zmimbndd?workspace=user- all seems good ! Merging!

PR #500, opened 2023-07-06
[`PPOTrainer`] Add prompt tuning support on TRL

younesbelkada:
Fixes https://github.com/lvwerra/trl/issues/490 

Prompt tunning appends new tokens in the beginning of the generated sentence and make the new vocabulary that is used to generate these tokens trainable. 
Before this PR that PEFT method was not supported out of the box. I propose a v1 to integrate prompt tuning into TRL.

Might be also related to https://github.com/lvwerra/trl/issues/470 

Added also a nice test that fails on main and does not fail on this branch

In my opinion we should zero-out the logits that corresponds to these new tokens as it is not possible to compute reference logits on these tokens, however we shouldn't mask out these tokens in the PPO step

cc @lvwerra

### DIFF
--- a/tests/test_ppo_trainer.py
+++ b/tests/test_ppo_trainer.py
@@ -872,6 +872,66 @@ class PPOTrainerTester(unittest.TestCase):
                 else:
                     self.assertTrue(param.grad is None, f"Parameter {name} has a gradient")
 
+    @require_peft
+    @mark.peft_test
+    def test_peft_model_ppo_trainer_prompt_tuning(self):
+        from peft import PromptTuningConfig, TaskType, get_peft_model
+        from transformers import AutoModelForCausalLM
+
+        pt_config = PromptTuningConfig(
+            task_type=TaskType.CAUSAL_LM,
+            num_virtual_tokens=10,
+            tokenizer_name_or_path=self.model_id,
+            inference_mode=False,
+            token_dim=32,
+        )
+        gpt2_model = AutoModelForCausalLM.from_pretrained(self.model_id)
+
+        # this line is very important
+        def make_inputs_require_grad(module, input, output):
+            output.requires_grad_(True)
+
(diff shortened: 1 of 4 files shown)


tests/test_ppo_trainer.py
```diff
@@ -872,6 +872,67 @@ def make_inputs_require_grad(module, input, output):
...
+        from peft import PromptTuningConfig, TaskType, get_peft_model
+        from transformers import AutoModelForCausalLM
+
+        pt_config = PromptTuningConfig(
+            task_type=TaskType.CAUSAL_LM,
+            num_virtual_tokens=10,
+            tokenizer_name_or_path=self.model_id,
+            inference_mode=False,
+            token_dim=32,
+            encoder_hidden_size=32,
```
younesbelkada, line 887: to remove on the new release : https://github.com/huggingface/peft/pull/459

lvwerra:
> In my opinion we should zero-out the logits that corresponds to these new tokens as it is not possible to compute reference logits on these tokens, however we shouldn't mask out these tokens in the PPO step


Not sure I follow: if we can't compute ref logits then the training will fail. I think this is maybe better suited for the SFT step?

younesbelkada:
> Not sure I follow: if we can't compute ref logits then the training will fail. 

The reference logits can in fact be computed but it can't be computed on the new prepended tokens that are created by the new embedding layer (as it is relevant only when inferring with the adapter model). But we can still compute the ref logits and logits on the model's input.

younesbelkada:
After a discussion with @lvwerra I agree that RLHF + prompt tuning might be a bit corner case and hard to support. As the new tokens will be pre-pended to the query it is a bit unclear how the model will learn the correct alignment during the PPO step given the "lack" of learnable signal. Will convert this PR to add a check. 
Did I missed anythin @lvwerra ?

lvwerra (approved):

PR #509, opened 2023-07-10
[`examples`] Big refactor of examples and documentation

younesbelkada:
This PR is an attempt to refactor most of the examples and adapt the documentation accordingly

Added also a small enhancement for the reward trainer to directly call `prepare_model_for_int8_training` method under the hood.

### Examples:

- [x] Add a script for SFTTrainer
- [x] Modify the docs accordingly for SFTTrainer
- [x] Add a script for RMTrainer
- [x] Modify the docs accordingly for RewardTrainer
- [x] Add one or two scripts for PPOTrainer 
- [x] Modify the docs accordingly for PPOTrainer
- [x] Try to reproduce stack llama only using the new scripts - let's move all research projects inside a dedicated folder
- [x] Deal with `examples/summarization` folder

Removing all scripts related to merging adapter weights since in PEFT you can simply do:

```python
model = model.merge_and_unload()
```

### Documentation

To discuss

cc @lvwerra @vwxyzjn


trl/trainer/reward_trainer.py
```diff
@@ -28,7 +28,7 @@
 
 
 if is_peft_available():
-    from peft import PeftModel, get_peft_model
+    from peft import PeftModel, get_peft_model, prepare_model_for_int8_training
```
lvwerra, line 31: what about int4?

examples/reward_modeling/train_reward_model.py
```diff
@@ -0,0 +1,169 @@
...
+# configuration with `accelerate config`
+#
+########################################################################
+
+########################################################################
+# NOTE for to train with a 8-bit model a more recent version of
+# transformers is required, full dependecies for this example:
+# pip install bitsandbytes datasets accelerate
+# pip install git+https://github.com/huggingface/peft.git
+########################################################################
```
lvwerra, line 55: I would just mention this in the readme? Feels a bit an odd place for documentation. what do you think?

lvwerra (commented):
Overall, looks really great! Left a few small comments and two main points:

## Doc structure

What do you think about simplifying/renaming the sections a bit in the docs:

- API
  - Model Classes
  - Trainer Classes
  - Reward Model Training
  - Supervised Fine-Tuning
  - Best-of-N Samppling

- Examples
  - Sentiment Tuning
  - Training with PEFT
  - Detoxifying LLMs
  - StackLlama
  - Multi-Adapter Training 

I think we could also do a better job at giving an overview of the docs in the landing page on the docs.

## Examples structure

For consistency I would call the reward examples folder `reward_trainer`. Having said that I wonder if we shouldn't further simplify the examples only to have these folders:
- research_projects
- notebooks
- scripts
Most of the folders only have 1-2 scripts anyway.

trl/trainer/reward_trainer.py
younesbelkada: It works for both, currently in PEFT there is a method  called `prepare_model_for_kbit_training`, but we did not made a release yet (they're the same method, just changed the name).

examples/reward_modeling/train_reward_model.py
younesbelkada: Agreed !

younesbelkada (commented):
Thank @lvwerra for the extensive review, should be all good now!

docs/source/index.mdx
```diff
@@ -4,6 +4,28 @@
...
 
-TRL supports decoder models such as GPT-2, BLOOM, GPT-Neo which can all be optimized using Proximal Policy Optimization (PPO). You can find installation instructions in the [installation guide](installation) and an introduction to the library in the [Quickstart section](quickstart). There is also a more [in-depth example](sentiment_tuning) to tune GPT-2 to produce positive movie reviews.
+<div style="text-align: center">
+<img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/TRL-readme.png">
+</div>
+
+Check the appropriate sections of the documentation depending on your needs:
+
+API documentation:
+
```
lvwerra, line 17: What do you think about adding 1 sentence here per bullet point for people who don't know e.g. what an `SFTTrainer` could do or what a StackLlama is :)

lvwerra (commented):
Looks great, I think we can merge soon! A few small nits:

## Short Sections Names

It would be really nice if we could make all the section heads single line in the sidebar. Suggestions:
- Training your own reward model -> Reward Model Training
- Best of N sampling - Better model output without reinforcement learning -> Best of N Sampling
- Multi Adapter RL (MARL) - a single base model for everything -> Multi Adapter RLHF or Multi Adapter RL?
- Using Llama with TRL -> Training StackLlama

## Section Capitalisation
Also we should be consistent in capitalisation. I am in favour of capitalising the sections heads. E.g. Reward Model Training rather than Reward model training. What do you think?

## SFTTrainer

It seems to be missing from the `Trainer Classes` docs page.

docs/source/index.mdx
younesbelkada: Makes sense!

younesbelkada:
Looks much cleaner now ! Addressed your comment @lvwerra  !

lvwerra (approved):

PR #511, opened 2023-07-11
Update modeling_base.py

DogitoErgoSum:
Update to prepare_model_for_kbit_training from int8_training


younesbelkada (commented):
hello @DogitoErgoSum thanks very much for your PR, unfortunately we cannot merge this yet as `prepare_model_for_kbit_training` is not on the latest release of PEFT library. We need to wait for the next release of PEFT and constraint the peft version to be greater than 0.4.0 on `setup.py` before merging this. Thanks !

PR #512, opened 2023-07-12
[`SFTTrainer`] Fix the sequence length check of `SFTTrainer`

younesbelkada:
# What does this PR do? 

Fixes https://github.com/lvwerra/trl/issues/467
Replaces https://github.com/lvwerra/trl/pull/481

The original code lead to having examples that were filtered out by the previous check. This ultimately lead to having examples being ignored when processing the data. Probably a copy-pasta from: https://huggingface.co/learn/nlp-course/chapter7/6?fw=pt where I borrowed the original code

The fix is to remove completely that check

cc @lvwerra


lvwerra:
We should also set `padding=False` and ` return_length=False`. Otherwise looks good.

PR #515, opened 2023-07-14
Use local process index for `_get_current_device()`

lewtun:
This PR fixes a bug in `_get_current_device()` where the global process index was being returned instead of the local one. 

With this fix, it is possible to run training in **multi-node** environments and avoid the dreaded `RuntimeError: CUDA error: invalid device ordinal` :)


younesbelkada (approved):
Amazing catch ! Thanks so much @lewtun 🔥

lvwerra (approved):

PR #516, opened 2023-07-14
[`PPOTrainer`] Print query tokens and associated KL

younesbelkada:
As discussed offline, we can add an utility logging to log the kl per token when the mean_kl is quite negative (we can also make that value modulable).

cc @lvwerra

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -638,6 +638,9 @@ class PPOTrainer(BaseTrainer):
 
         t = time.time()
         rewards, non_score_reward = self.compute_rewards(scores, all_logprobs, ref_logprobs, masks)
+
+        self._kl_warning(non_score_reward, responses, masks)
+
         timing["time/ppo/compute_rewards"] = time.time() - t
 
         # upcast to float32 to avoid dataset issues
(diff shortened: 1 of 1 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -1136,6 +1136,12 @@ def record_step_stats(self, kl_coef: float, **data):
                 " that the generation kwargs are set correctly, or review your training hyperparameters."
             )
 
+        if mean_kl.item() < -10.0:
+            warnings.warn(
+                f"KL divergence is becoming very negative: {mean_kl.item():.2f} - this might be a precursor for failed training."
+                f" Here is the mea, kl penalty score per token {kl_list}."
+            )
+
```
vwxyzjn, line 1144: Isn't this duplicate warning to lines 1131-1138?

lvwerra:
Adapted it a bit to print per token KL as well as the token for better debugging.

lvwerra:
cc @edbeeching and @lewtun as this could be useful for debugging

trl/trainer/ppo_trainer.py
```diff
@@ -782,6 +785,19 @@ def _early_stop(self, policykl):
...
+        for i, kl in enumerate(avg_kl):
+            if kl < -10:
+                warning_str = f"The average KL for sequence is below threshold ({kl.item():.2f}). Per token KL: "
+                for token_index in range(non_score_reward.shape[1]):
+                    if masks[i, token_index] == 1:
+                        # TODO: a bug adds an unnecessary masked token at position 0 which is why we offset -1
+                        token = self.tokenizer.convert_ids_to_tokens([responses[i][token_index - 1]])[0]
+                        per_token_kl = non_score_reward[i, token_index]
+                        warning_str += f"[{token}, {per_token_kl:.2f}] "
+                warnings.warn(warning_str)
```
vwxyzjn, line 799: Oh nice, this will be super helpful! One additional thing ti add is maybe to output the corresponding `ref_logprob` and `logprob`, but it might clutter the output too much so either way is fine with me.

PR #528, opened 2023-07-18
[DPO] make sure all the concated batches are on same device

### LINKED ISSUE #525: DPO evaluation error---tensors on two devices
Hi!

Thanks for the awesome codebase. I ran the DPO example `trl/examples/dpo.py` but encountered an error at the evaluation step: `Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!`

Here is a [colab notebook](https://colab.research.google.com/drive/11AVym7U3gkTn_qTfrnSDtA_AO5_zJkkd?usp=sharing) that shows this problem. To expose the problem faster, I set `training_args.eval_steps = 1`.

To solve it, a hotfix can be adding `.to(self.accelerator.device)` at a few places in [concatenated_forward](https://github.com/lvwerra/trl/blob/main/trl/trainer/dpo_trainer.py#L288-L296):

```python
all_logits = model(
    concatenated_batch["concatenated_input_ids"].to(self.accelerator.device),
    attention_mask=concatenated_batch["concatenated_attention_mask"].to(self.accelerator.device),
).logits.to(torch.float32)
all_logps = self._get_batch_logps(
    all_logits,
    concatenated_batch["concatenated_labels"].to(self.accelerator.device),
    average_log_prob=False,
)
```

However, I am not sure why the trainer does not handle the device change automatically. If this hotfix is fine, I can submit a pull request. Otherwise, I'm also happy to learn how to address this problem more generically.

Tianlin

kashif:
fixes #525

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -204,7 +204,7 @@ class DPOTrainer(Trainer):
                         pad_to_length(batch[k], max_length, pad_value=pad_value),
                     ),
                     dim=0,
-                )
+                ).to(self.accelerator.device)
         return concatenated_batch
 
     def dpo_loss(


younesbelkada (approved):
Thanks for the quick fix!

lvwerra (approved):

PR #531, opened 2023-07-18
Fix application of prompt function to non-CLDataset inputs

jganitkevitch:
Passing a standard dataset in non-packing mode gets NOOPed without applying the prompt function. This PR fixes that.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -231,8 +231,8 @@ class SFTTrainer(Trainer):
         if dataset is None:
             raise ValueError("The dataset should not be None")
 
-        # check if torch dataset / dataloader and do nothing
-        if isinstance(dataset, (torch.utils.data.IterableDataset, torch.utils.data.Dataset, ConstantLengthDataset)):
+        # do nothing if dataset is already a ConstantLengthDataset
+        if isinstance(dataset, ConstantLengthDataset):
             return dataset
 
         if not packing:


younesbelkada (commented):
Hi @jganitkevitch 
Thanks for the PR, maybe I am wrong but I think this behavior is intended no? When one passes a `torch.utils.data.Dataset` we assume that dataset is already processed by the user. Maybe I need you to share me a snippet that shows a situation where you pass a torch dataset + want to apply prompt function. What do you think?

jganitkevitch:
Certainly! This is for any case where we'd want to use `formatting_func` to process the data, but don't want to use a CLDataset. I think the example from the SFTTrainer page, minimally modified, would do:

```python
from trl import SFTTrainer

dataset = load_dataset("CodeAlpaca_20K", split="train")

model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")

def formatting_prompts_func(example):
    output_texts = []
    for i in range(len(example['prompt'])):
        text = f"### Question: {example['prompt'][i]}\n ### Answer: {example['completion'][i]}"
        output_texts.append(text)
    return output_texts

trainer = SFTTrainer(
    model,
    train_dataset=dataset,
)

trainer.train() 
```

younesbelkada (commented):
Hi @jganitkevitch 
Thanks for iterating, I ran this script (the original one was not working): 

```python
from trl import SFTTrainer

dataset = load_dataset("lucasmccabe-lmi/CodeAlpaca-20k", split="train")

model = AutoModelForCausalLM.from_pretrained("facebook/opt-350m")

def formatting_prompts_func(example):
    output_texts = []
    for i in range(len(example['instruction'])):
        text = f"### Question: {example['instruction'][i]}\n ### Answer: {example['output'][i]}"
        output_texts.append(text)
    return output_texts

trainer = SFTTrainer(
    model,
    train_dataset=dataset,
)

trainer.train() 
```

and it seems to not ignore the formatting function. The check `isinstance(dataset, (torch.utils.data.IterableDataset, torch.utils.data.Dataset, ConstantLengthDataset))` will return `False` because the dataset you created is a dataset from `datasets` library. Then since you did not passed `packing` argument the SFTTrainer's init method will successfully call the method `_prepare_non_packed_dataloader` and format the input correctly since [`formatting_func` is not `None` and `dataset_text_field` is None](https://github.com/lvwerra/trl/blob/b4f4558952fa3a3017dcc47eeb116ea4794e907e/trl/trainer/sft_trainer.py#L274).
I think the check `isinstance(dataset, (torch.utils.data.IterableDataset, torch.utils.data.Dataset, ConstantLengthDataset))` is important as it assumes users have already created a processed torch dataset, hence already tokenized the dataset beforehand; so it is important to do nothing in that case to avoid any unexpected behaviour. Does that makes sense to you? Let me know if anything is unclear

younesbelkada:
Closing this PR for now, feel free to re-open if you think the explanation above does not make sense !

PR #539, opened 2023-07-19
Relax reward trainer constraint

younesbelkada:
Fixes https://github.com/lvwerra/trl/issues/535

We should relax the constraint for users in case they implement their own Data collator when using `RewardTrainer`.

cc @lvwerra

### DIFF
--- a/trl/trainer/reward_trainer.py
+++ b/trl/trainer/reward_trainer.py
@@ -162,7 +162,7 @@ class RewardTrainer(Trainer):
     ) -> Union[torch.Tensor, Tuple[torch.Tensor, Dict[str, torch.Tensor]]]:
         if not self.use_reward_data_collator:
             warnings.warn(
-                "The current compute_loss is implemented for RewardDataCollatorWithPadding"
+                "The current compute_loss is implemented for RewardDataCollatorWithPadding,"
                 " if you are using a custom data collator make sure you know what you are doing or"
                 " implement your own compute_loss method."
             )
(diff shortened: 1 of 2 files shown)


trl/trainer/reward_trainer.py
```diff
@@ -161,8 +161,10 @@ def compute_loss(
         return_outputs=False,
     ) -> Union[torch.Tensor, Tuple[torch.Tensor, Dict[str, torch.Tensor]]]:
         if not self.use_reward_data_collator:
-            raise NotImplementedError(
-                "compute_loss is only implemented for RewardDataCollatorWithPadding, please implement your own compute_loss method if you are using a custom data collator"
+            warnings.warn(
+                "The current compute_loss is implemented for RewardDataCollatorWithPadding"
```
tomaarsen, line 165:
```suggestion
                "The current compute_loss is implemented for RewardDataCollatorWithPadding,"
```

tomaarsen (approved):
Small nit.

Other than that, perhaps we can propagate this to the DPO Trainer as well:
https://github.com/lvwerra/trl/blob/0e8d9f8504909fcc0571545b9f3777b285084e59/trl/trainer/dpo_trainer.py#L355-L358

and perhaps even in the prediction_step:
https://github.com/lvwerra/trl/blob/0e8d9f8504909fcc0571545b9f3777b285084e59/trl/trainer/dpo_trainer.py#L403-L406

younesbelkada:
Thanks @tomaarsen for your review, your comments makes sense! Will address them

lvwerra (approved):

PR #540, opened 2023-07-19
[DPO] remove response/pairs from the DPO side

### LINKED ISSUE #537: DPOTrainer ignores training & evaluation data if there are more than 2 responses
Hello!

First of all, thank you @younesbelkada and @lvwerra for maintaining this work, and thank you @kashif for spearheading the valuable DPO implementation.

## Bug overview
* The `DPODataCollatorWithPadding` processes `batch_size` samples at once, yielding `batch_size` chosen-rejected pairs. However, `batch_size` samples may have much more than `batch_size` pairs.

## Details
The issue originates in the `DPODataCollatorWithPadding.__call__` method:
https://github.com/lvwerra/trl/blob/7f297b38c665e6d2357dfd5ba552ef685afe1302/trl/trainer/utils.py#L328-L345
The `break` will pre-emptively stop processing of all pairs in `features` to ensure that the output has a length of `batch_size`. Because of this, any training or evaluation data with more than 2 responses will result in data not being used.

## How to reproduce
<details><summary>See reproduction steps</summary>

Replace `DPODataCollatorWithPadding.__call__` with this variant, which tracks how many samples are used and how many are ignored:
```python
    def __call__(self, features: List[Dict[str, Any]]) -> Dict[str, Any]:
        batch = []
        done = False
        for feature in features:
            prompt = feature["prompt"]
            responses = feature["responses"]
            pairs = feature["pairs"]

            for p in pairs:
                if not done:

                    batch_element = self.tokenize_batch_element(prompt, responses[p[0]], responses[p[1]])
                    batch.append(batch_element)

                    done = len(batch) == self.batch_size
                    self.included += 1
                else:
                    self.ignored += 1

        print(f"{self.ignored / (self.ignored + self.included):%} of training pairs are ignored.")
        return self.collate(batch)
```
and add
```python
    included: int = 0
    ignored: int = 0
```
to the list of dataclass fields of the `DPODataCollatorWithPadding`.

Then, run this script, which uses a dummy dataset with many responses:
```python
from transformers import AutoModelForCausalLM, AutoTokenizer, TrainingArguments
from trl import DPOTrainer
from datasets import Dataset

model = AutoModelForCausalLM.from_pretrained("gpt2")

model_ref = AutoModelForCausalLM.from_pretrained("gpt2")

train_dataset_dict = {
    "prompt": [
        "trl",
        "is",
        "pretty",
        "great",
        "but",
        "it",
        "has",
        "a",
        "small",
        "bug",
    ],
    "responses": [
        [
            "good",
            "bad",
            "worse",
            "horrible",
            "not good",
            "poor",
            "awful",
            "substandard",
            "unsatisfactory",
        ],
    ]
    * 10,
    "pairs": [
        [(0, 1), (0, 2), (0, 3), (0, 4), (0, 5), (0, 6), (0, 7), (0, 8)],
    ]
    * 10,
}
train_dataset = Dataset.from_dict(train_dataset_dict)

    remove_unused_columns=False,
    output_dir="./tmp",
    report_to="none",
    num_train_epochs=10,
)

dpo_trainer = DPOTrainer(
    model,
    model_ref,
    args=training_args,
    beta=0.1,
    train_dataset=train_dataset,
    tokenizer=tokenizer,
)

dpo_trainer.train()
```
Towards the end of training, you'll see that `85.00%` of all pairs have been ignored in this example. 

</details>

## Proposed fixes
Before I start working on it, I'd like to discuss my proposed changes: 

### The fix
Spread out the training and evaluation datasets into a Dataset of `prompt`, `chosen`, `rejected` columns right in the `DPOTrainer.__init__`. This requires some small changes throughout this work, but will work conveniently otherwise: each dataset element will be one training sample. It also has a drawback: prompts and responses will need to be tokenized multiple times, unless we immediately store the `prompt`-`chosen`-`rejected` dataset in a tokenized way.

#### `prompt`, `chosen`, `rejected`
Beyond that, I can also implement a more drastic change: require the user to provide a Dataset with `prompt`, `chosen`, `rejected` columns. 

#### `prompt`, `responses`, `chosen` | `pairs`
A less drastic change is to allow users to provide a `chosen` column like:
```
"chosen": [0, 1, 0, 0]
```
where the `chosen` is simply the index of the best response, instead of `pairs` like:
```python
"pairs": [
        [(0, 1)],
        [(1, 0)],
        [(0, 2), (0, 1)],
        [(0, 1), (0, 2), (0, 3)],
    ]
```
Internally, we can use this `chosen` to create the same pairs. This is simpler for most users, but is less powerful, as the `pairs` approach allows for a ranking (e.g. also say that although response 1 is best, response 2 is better than response 3). I propose that the input dataset should *either* have a `pairs` or `chosen` column.

### What now?
I would love some feedback on your thoughts on:
* Is the proposed fix (internal `prompt`, `chosen`, `rejected` columns) a good idea?
* Should we change the input structures to `prompt`, `chosen`, `rejected`, `prompt`, `responses`, `chosen` | `pairs`, or keep it as-is?

---

- Tom Aarsen

kashif:
fixes #537

### DIFF
--- a/examples/dpo.py
+++ b/examples/dpo.py
@@ -95,11 +95,12 @@ def get_hh(split: str, sanity_check: bool = False, silent: bool = False, cache_d
 
     def gen():
         for prompt, values in data.items():
-            yield {
-                "prompt": prompt,
-                "responses": values["responses"],
-                "pairs": values["pairs"],
-            }
+            for p in values["pairs"]:
+                yield {
+                    "prompt": prompt,
+                    "chosen": values["responses"][p[0]],
+                    "rejected": values["responses"][p[1]],
+                }
 
     return Dataset.from_generator(gen)
 
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -326,23 +326,18 @@ class DPODataCollatorWithPadding:
         return padded_batch
 
     def __call__(self, features: List[Dict[str, Any]]) -> Dict[str, Any]:
-        batch = []
+        tokenized_batch = []
+
         for feature in features:
             prompt = feature["prompt"]
-            responses = feature["responses"]
-            pairs = feature["pairs"]
+            chosen = feature["chosen"]
+            rejected = feature["rejected"]
 
-            done = False
-            for p in pairs:
-                if done:
-                    break
+            batch_element = self.tokenize_batch_element(prompt, chosen, rejected)
+            tokenized_batch.append(batch_element)
 
-                batch_element = self.tokenize_batch_element(prompt, responses[p[0]], responses[p[1]])
-                batch.append(batch_element)
-
-                done = len(batch) == self.batch_size
-
-        return self.collate(batch)
+        # return collated batch
+        return self.collate(tokenized_batch)
 
 
 class ConstantLengthDataset(IterableDataset):


examples/dpo.py
```diff
@@ -95,11 +95,12 @@ def split_prompt_and_responses(ex):
 
     def gen():
```
tomaarsen, line 96: This function can be simplified a lot now:
```python
def get_hh(split: str, sanity_check: bool = False, silent: bool = False, cache_dir: str = None) -> Dataset:
    """Load the Anthropic Helpful-Harmless dataset from Hugging Face and convert it to the necessary format.

    The dataset is converted to a dictionary with the following structure:
    {
        'prompt': List[str],
        'chosen': List[str],
        'rejected': List[str],
    }

    Prompts should be structured as follows:
      \n\nHuman: <prompt>\n\nAssistant:
    Multiple turns are allowed, but the prompt should always start with \n\nHuman: and end with \n\nAssistant:.
    """
    dataset = load_dataset("Anthropic/hh-rlhf", split=split, cache_dir=cache_dir)
    if sanity_check:
        dataset = dataset.select(range(min(len(dataset), 1000)))

    def split_prompt_and_responses(sample) -> Dict[str, str]:
        prompt = extract_anthropic_prompt(sample["chosen"])
        return {
            "prompt": prompt,
            "chosen": sample["chosen"][len(prompt) :],
            "rejected": sample["rejected"][len(prompt) :],
        }

    return dataset.map(split_prompt_and_responses)
```
I can't push this to the PR directly or add this as a suggestion in the review, sadly.

kashif:
you should be able to push to my branch now... please feel free!

kashif:
let me fix up the docs and test

kashif:
@younesbelkada or @lvwerra  should be ready for review!

tomaarsen:
I appreciate that you left some time for people to find these issues before releasing. I'd love to look into the spamminess of the logs during training as well. Do you know when you intend to release?

- Tom Aarsen

tomaarsen (approved):

tomaarsen:
`wandb` support should be automatically implemented right? As long as it's installed and the `report_to` is set up right.
Is the issue simply that we don't call `self.callback_handler.on_log(self.args, self.state, self.control, ...)`?

I can help with this, as I have a good amount of experience with how Transformers does its `Trainer` via a SetFit logging/callbacks integration that I've been working on.

kashif:
yup! for some reason i was not getting it to log with the trainer and so I "forced" it to log things via the call to:

```py
     self.log_metrics("test", metrics)
```

would appreciate any insight here!

younesbelkada:
Thanks everyone! For the release probably mid-next week would be a nice time, but not sure sure @lvwerra

PR #541, opened 2023-07-19
Computes the KL penalty using the entire distribution

edbeeching:
I got so fed up with negative KL that I have added an option to calculate the KL for the entire distribution. 
It is memory intensive for large sequences with an allocation of `8*vocab_size` per token, but it should stabilize training and cannot be negative! Memory usage could be greatly reduced but would require a larger refactor.

Single seed runs for gpt2-sentiment:

image


lvwerra (approved):

younesbelkada (approved):
Thanks !

PR #544, opened 2023-07-19
Add `use_auth_token` arg to sft_trainer example

corey-lambda:
This is for huggingface models that require using an access token to download (like meta's llama2). Tested by running the following command:

```bash
> huggingface-cli login
> python trl/examples/scripts/sft_trainer.py \
    --model_name meta-llama/Llama-2-7b-hf \
    --dataset_name timdettmers/openassistant-guanaco \
    --load_in_4bit \
    --use_peft \
    --batch_size 4 \
    --gradient_accumulation_steps 2 \
    --use_auth_token
```

### DIFF
--- a/examples/scripts/sft_trainer.py
+++ b/examples/scripts/sft_trainer.py
@@ -54,7 +54,7 @@ class ScriptArguments:
     peft_lora_r: Optional[int] = field(default=64, metadata={"help": "the r parameter of the LoRA adapters"})
     peft_lora_alpha: Optional[int] = field(default=16, metadata={"help": "the alpha parameter of the LoRA adapters"})
     logging_steps: Optional[int] = field(default=1, metadata={"help": "the number of logging steps"})
-    use_auth_token: Optional[bool] = field(default=False, metadata={"help": "Use HF auth token to access the model"})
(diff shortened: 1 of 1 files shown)


examples/scripts/sft_trainer.py
```diff
@@ -54,6 +54,7 @@ class ScriptArguments:
     peft_lora_r: Optional[int] = field(default=64, metadata={"help": "the r parameter of the LoRA adapters"})
     peft_lora_alpha: Optional[int] = field(default=16, metadata={"help": "the alpha parameter of the LoRA adapters"})
     logging_steps: Optional[int] = field(default=1, metadata={"help": "the number of logging steps"})
+    use_auth_token: Optional[bool] = field(default=False, metadata={"help": "Use HF auth token to access the model"})
```
lvwerra, line 57: Good point, let's make this `True` by default so we don't have to update the examples and it should work out of the box.

```suggestion
    use_auth_token: Optional[bool] = field(default=True, metadata={"help": "Use HF auth token to access the model"})
```

lvwerra (approved):
Thanks a lot for the fix! Made a small change so this works out of the box without updating the existing examples.

PR #546, opened 2023-07-19
refactor grad accum

### LINKED ISSUE #545: Broken gradient accumulation
## Description

I ran `python examples/scripts/sentiment_tuning.py --log_with wandb` and compared it with `python examples/scripts/sentiment_tuning.py --log_with wandb --mini_batch_size 1 --gradient_accumulation_steps 128`.

Some gradient accumulation runs get the following errors and crash immediately. Other surviving runs do not learn anything 

image

```
278 │    778 │   a, weights = _ravel_and_check_weights(a, weights)                 │
279 │    779 │                                                                     │
280 │ ❱  780 │   bin_edges, uniform_bins = _get_bin_edges(a, bins, range, weights) │
281 │    781 │                                                                     │
282 │    782 │   # Histogram is an integer or a float array depending on the weigh │
283 │    783 │   if weights is None:                                               │
284 │                                                                              │
285 │ /admin/home/costa/.pyenv/versions/trl/lib/python3.8/site-packages/numpy/lib/ │
286 │ histograms.py:426 in _get_bin_edges                                          │
287 │                                                                              │
288 │    423 │   │   if n_equal_bins < 1:                                          │
289 │    424 │   │   │   raise ValueError('`bins` must be positive, when an intege │
290 │    425 │   │                                                                 │
291 │ ❱  426 │   │   first_edge, last_edge = _get_outer_edges(a, range)            │
292 │    427 │                                                                     │
293 │    428 │   elif np.ndim(bins) == 1:                                          │
294 │    429 │   │   bin_edges = np.asarray(bins)                                  │
295 │                                                                              │
296 │ /admin/home/costa/.pyenv/versions/trl/lib/python3.8/site-packages/numpy/lib/ │
297 │ histograms.py:323 in _get_outer_edges                                        │
298 │                                                                              │
299 │    320 │   else:                                                             │
300 │    321 │   │   first_edge, last_edge = a.min(), a.max()                      │
301 │    322 │   │   if not (np.isfinite(first_edge) and np.isfinite(last_edge)):  │
302 │ ❱  323 │   │   │   raise ValueError(                                         │
303 │    324 │   │   │   │   "autodetected range of [{}, {}] is not finite".format │
304 │    325 │                                                                     │
305 │    326 │   # expand empty range to avoid divide by zero                      │
306 ╰──────────────────────────────────────────────────────────────────────────────╯
307 ValueError: autodetected range of [nan, nan] is not finite
```

This is interesting because we have a test case asserting the **gradient equivalence** of these two hyperparameters... Gonna dig deeper.

https://github.com/lvwerra/trl/blob/d0fe348a0a354bb10e0d728d8e9a4d46d4df0cbc/tests/test_ppo_trainer.py#L1018-L1019
https://github.com/lvwerra/trl/blob/d0fe348a0a354bb10e0d728d8e9a4d46d4df0cbc/tests/test_ppo_trainer.py#L1042-L1043

vwxyzjn:
Closes #545 and #548. This PR does a few things:

* **introduces "micro_batch"**, a more clarified terminology for gradient accumulation. (E.g., see the terminology adoption in [nanoGPT](https://github.com/karpathy/nanoGPT/blob/eba36e84649f3c6d840a93092cb779a260544d08/train.py#L289))
    ```python
    import numpy as np
    batch_size = 8
    mini_batch_size = 4
    gradient_accumulation_steps = 2
    micro_batch_size = mini_batch_size // gradient_accumulation_steps
    data = np.arange(batch_size)
    for epoch in range(2):
        b_inds = np.random.permutation(batch_size)
        print("epoch: ", epoch, "b_inds: ", b_inds)
        for mini_batch_start in range(0, batch_size, mini_batch_size):
            mini_batch_end = mini_batch_start + mini_batch_size
            mini_batch_inds = b_inds[mini_batch_start:mini_batch_end]
            
            # `optimizer.zero_grad()` set optimizer to zero for gradient accumulation
            for micro_batch_start in range(0, mini_batch_size, micro_batch_size):
                micro_batch_end = micro_batch_start + micro_batch_size 
                micro_batch_inds = mini_batch_inds[micro_batch_start:micro_batch_end]
                print("actual training pass on", data[micro_batch_inds])
            # `optimizer.step()`
    ```
    ```
    epoch:  0 b_inds:  [3 7 2 1 4 5 6 0]
    actual training pass on [3 7]
    actual training pass on [2 1]
    actual training pass on [4 5]
    actual training pass on [6 0]
    epoch:  1 b_inds:  [6 5 2 0 7 3 1 4]
    actual training pass on [6 5]
    actual training pass on [2 0]
    actual training pass on [7 3]
    actual training pass on [1 4]
    ``` 
* **Only calculate advantages once**. As pointed out by #548, there is no need to calculate advantages at every forward pass. This PR refactors the learner only to calculate the advantage once.

I then conducted end-to-end experiments, which demonstrated near-identical learning curves when running with `accelerate launch --num_processes 1 examples/scripts/sentiment_tuning.py --log_with wandb --gradient_accumulation_steps 128` and `python examples/scripts/sentiment_tuning.py --log_with wandb` in the main branch.

image

To further confirmed it works across the board, I also tested with `gpt2-xl` and multi-GPU seemed to work perfectly

image

Potential tests to do in the future 
* Models other GPT2 such as starcoder
* DeepSpeed and peft tests 
* test if whitening the global advantages matters in the case of multi-GPU.

### DIFF
--- a/trl/core.py
+++ b/trl/core.py
@@ -124,6 +124,11 @@ def masked_var(values, mask, unbiased=True):
     variance = masked_mean(centered_values**2, mask)
     if unbiased:
         mask_sum = mask.sum()
+        if mask_sum == 0:
+            raise ValueError(
+                "The sum of the mask is zero, which can happen when `mini_batch_size=1`;"
+                "try increase the `mini_batch_size` or `gradient_accumulation_steps`"
+            )
         # note that if mask_sum == 1, then there is a division by zero issue
         # to avoid it you just need to use a larger minibatch_size
         bessel_correction = mask_sum / (mask_sum - 1)
--- a/trl/core.py
+++ b/trl/core.py
@@ -120,9 +120,9 @@ def masked_var(values, mask, unbiased=True):
     centered_values = values - mean
     variance = masked_mean(centered_values**2, mask)
     if unbiased:
-        mask_sum = mask.sum(dim=1)
-        # avoid division by 0 in case there is only one element
-        mask_sum[mask_sum == 1] = 0
+        mask_sum = mask.sum()
+        # note that if mask_sum == 1, then there is a division by zero issue
+        # to avoid it you just need to use a larger minibatch_size
         bessel_correction = mask_sum / (mask_sum - 1)
         variance = variance * bessel_correction
     return variance
--- a/trl/core.py
+++ b/trl/core.py
@@ -99,7 +99,7 @@ def logprobs_from_logits(logits, labels):
 
 def whiten(values, shift_mean=True):
     """Whiten values."""
-    mean, var = torch.mean(values), torch.var(values, unbiased=False)
+    mean, var = torch.mean(values), torch.var(values)
     whitened = (values - mean) * torch.rsqrt(var + 1e-8)
     if not shift_mean:
         whitened += mean
@@ -127,7 +127,7 @@ def masked_var(values, mask, unbiased=True):
 
 def masked_whiten(values, mask, shift_mean=True):
     """Whiten values with masked values."""
-    mean, var = masked_mean(values, mask), masked_var(values, mask, unbiased=False)
+    mean, var = masked_mean(values, mask), masked_var(values, mask)
     whitened = (values - mean) * torch.rsqrt(var + 1e-8)
     if not shift_mean:
         whitened += mean
--- a/trl/core.py
+++ b/trl/core.py
@@ -99,7 +99,7 @@ def logprobs_from_logits(logits, labels):
 
 def whiten(values, shift_mean=True):
     """Whiten values."""
-    mean, var = torch.mean(values), torch.var(values)
+    mean, var = torch.mean(values), torch.var(values, unbiased=False)
     whitened = (values - mean) * torch.rsqrt(var + 1e-8)
     if not shift_mean:
         whitened += mean
@@ -108,12 +108,21 @@ def whiten(values, shift_mean=True):
 
 def masked_mean(values, mask, axis=None):
     """Compute mean of tensor with a masked values."""
-    return (values * mask).mean(axis=axis)
+    if axis is not None:
+        return (values * mask).sum(axis=axis) / mask.sum(axis=axis)
+    else:
+        return (values * mask).sum() / mask.sum()
 
 
 def masked_var(values, mask, unbiased=True):
     """Compute variance of tensor with masked values."""
-    return (values * mask).var(unbiased=unbiased)
+    mean = masked_mean(values, mask)
+    centered_values = values - mean
+    variance = masked_mean(centered_values**2, mask)
+    if unbiased:
+        bessel_correction = mask.sum() / (mask.sum() - 1 + 1e-8)
+        variance = variance * bessel_correction
+    return variance
 
 
 def masked_whiten(values, mask, shift_mean=True):
--- a/trl/core.py
+++ b/trl/core.py
@@ -120,7 +120,10 @@ def masked_var(values, mask, unbiased=True):
     centered_values = values - mean
     variance = masked_mean(centered_values**2, mask)
     if unbiased:
-        bessel_correction = mask.sum() / (mask.sum() - 1 + 1e-8)
+        mask_sum = mask.sum(dim=1)
+        # avoid division by 0 in case there is only one element
+        mask_sum[mask_sum == 1] = 0
+        bessel_correction = mask_sum / (mask_sum - 1)
         variance = variance * bessel_correction
     return variance
 
--- a/trl/core.py
+++ b/trl/core.py
@@ -88,11 +88,14 @@ def pad_to_size(tensor, size, dim=1, padding=50256):
         return torch.nn.functional.pad(tensor, (0, size - t_size), "constant", padding)
 
 
-def logprobs_from_logits(logits, labels):
+def logprobs_from_logits(logits, labels, gather=True):
     """
     See: https://github.com/pytorch/pytorch/issues/563#issuecomment-330103591
     """
     logp = F.log_softmax(logits, dim=2)
+
+    if not gather:
+        return logp
     logpy = torch.gather(logp, 2, labels.unsqueeze(2)).squeeze(-1)
     return logpy
 
--- a/trl/core.py
+++ b/trl/core.py
@@ -108,26 +108,17 @@ def whiten(values, shift_mean=True):
 
 def masked_mean(values, mask, axis=None):
     """Compute mean of tensor with a masked values."""
-    if axis is not None:
-        return (values * mask).sum(axis=axis) / mask.sum(axis=axis)
-    else:
-        return (values * mask).sum() / mask.sum()
+    return (values * mask).mean(axis=axis)
 
 
 def masked_var(values, mask, unbiased=True):
     """Compute variance of tensor with masked values."""
-    mean = masked_mean(values, mask)
-    centered_values = values - mean
-    variance = masked_mean(centered_values**2, mask)
-    if unbiased:
-        bessel_correction = mask.sum() / (mask.sum() - 1)
-        variance = variance * bessel_correction
-    return variance
+    return (values * mask).var(unbiased=unbiased)
 
 
 def masked_whiten(values, mask, shift_mean=True):
     """Whiten values with masked values."""
-    mean, var = masked_mean(values, mask), masked_var(values, mask)
+    mean, var = masked_mean(values, mask), masked_var(values, mask, unbiased=False)
     whitened = (values - mean) * torch.rsqrt(var + 1e-8)
     if not shift_mean:
         whitened += mean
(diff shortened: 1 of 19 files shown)


vwxyzjn:
By following accelerate's docs, we should do

```
for batch in training_dataloader:
    with accelerator.accumulate(model):
        inputs, targets = batch
        outputs = model(inputs)
        loss = loss_function(outputs, targets)
        accelerator.backward(loss)
        optimizer.step()
        scheduler.step()
        optimizer.zero_grad()
```

I believe the implementation is correct but for some reason if I do `python examples/scripts/sentiment_tuning.py  --gradient_accumulation_steps 128` it gives me `'ppo/loss/policy': nan, 'ppo/loss/value': nan, 'ppo/loss/total': nan, 'ppo/policy/entropy': nan,` in `stats`... not sure what did I miss.

vwxyzjn:
I fixed all the test case except 
image

trl/trainer/ppo_trainer.py
```diff
@@ -649,76 +662,82 @@ def step(
...
-                        early_stop = self._early_stop(policykl)
-                        if early_stop:
-                            break
+            b_inds = np.random.permutation(bs)
+            for mini_batch_start in range(0, bs, self.config.mini_batch_size):
+                mini_batch_end = mini_batch_start + self.config.mini_batch_size
+                mini_batch_inds = b_inds[mini_batch_start:mini_batch_end]
+
+                # set optimizer to zero for gradient accumulation
+                for micro_batch_start in range(0, self.config.mini_batch_size, self.config.micro_batch_size):
```
zfang, line 698: Hi, 

I'm trying to understand the motivation behind the introduction of `micro_batch` in addition to `mini_batch`.

For example, given the contrived PPO config:
```
mini_batch_size = 4
gradient_accumulation_steps = 2
micro_batch_size = mini_batch_size / gradient_accumulation_steps = 2
```
(A side question: why even make `micro_batch_size` configurable when you will overwrite it with `mini_batch_size / gradient_accumulation_steps`?)

Based on my understanding:
Original:
- Train a batch of 4 (`mini_batch_size`), and zero out gradients for every 2 (`gradient_accumulation_steps`) batches;

New:
- Train a batch of 2 (`micro_batch_size`), and always zero out gradients;
- This is equivalent to setting `mini_batch_size = 2` and `gradient_accumulation_steps = 1` in the original implementation.

So it looks like
1. The behaviors between the original and new implementations are different based on the same `mini_batch_size` and `gradient_accumulation_steps` config values, which could make this change not backward compatible;
2. The additional level of micro batch for loop appears to be unnecessary and further increase code complexity when you can simply use different `mini_batch_size` and `gradient_accumulation_steps` config values to achieve the same behavior;
3. `gradient_accumulation_steps` in the new implementation loses its intended meaning because we do not accumulate gradients at all with micro batches.

Thanks and sorry if I misunderstood anything,

Felix

vwxyzjn: Hi Felix, thanks for the thoughtful reply. 

> A side question: why even make micro_batch_size configurable when you will overwrite it with mini_batch_size / gradient_accumulation_steps?

Because then `micro_batch_size` will be recorded to wandb for experiment tracking to my understanding. 

> Train a batch of 2 (micro_batch_size), and always zero out gradients;... gradient_accumulation_steps in the new implementation loses its intended meaning because we do not accumulate gradients at all with micro batches.

It looks this way but in reality, it's doing something different. Because we hand off gradient accumulation to accelerate, you are supposed to call `optimizer.zero_grad()` even though typically you wouldn't (see https://huggingface.co/docs/accelerate/usage_guides/gradient_accumulation#the-finished-code)

> The additional level of micro batch for loop appears to be unnecessary and further increase code complexity when you can simply use different mini_batch_size and gradient_accumulation_steps config values to achieve the same behavior;

I guess this is more of a stylistic thing. We really have three levels of "batch sizes":

1. `batch_size` is the amount of rollout data, e,g., 8 data points
2. `minibatch_size` is the amount of data that you are actually doing a  `zero_grad()`, `loss.backward()`, and `optimizer.step()`, e.g., 4 data points
3. `microbatch_size` if the `minibatch_size` is too large to fit in memory, so we `zero_grad()`, partition the `minibatch` and do multiple `loss.backward()` and then `optimizer.step()`. 

Having the terminology like this makes it clear what the real `minibatch_size` is, so that we do not confound it with `gradient_accumulation_steps`. 

From the user's perspective, one set of hyperparameters should always reliably reproduce similar learning curves. By the new API design in this PR, `batch_size=8, minibatch_size=4` will reliably reproduce similar learning curves regardless of how many `gradient_accumulation_steps` there is, so when doing experiment management / analysis we can just group /filter by `batch_size` and `minibatch_size`. The existing implementation, however, would require us to calculate what's the "real size of the data we perform an `optimizer.step()` on".

trl/core.py
```diff
@@ -123,7 +123,10 @@ def masked_var(values, mask, unbiased=True):
     centered_values = values - mean
     variance = masked_mean(centered_values**2, mask)
     if unbiased:
-        bessel_correction = mask.sum() / (mask.sum() - 1)
+        mask_sum = mask.sum()
+        # note that if mask_sum == 1, then there is a division by zero issue
+        # to avoid it you just need to use a larger minibatch_size
+        bessel_correction = mask_sum / (mask_sum - 1)
```
younesbelkada, line 129: Maybe let's add a warning here in case `mask_sum==1`?

trl/trainer/ppo_trainer.py
```diff
@@ -659,83 +672,90 @@ def step(
 
         timing["time/ppo/forward_pass"] = time.time() - t
 
-        t = time.time()
+        with torch.no_grad():
```
younesbelkada, line 675: Why weren't doing `torch.no_grad()` at the first place? Perhaps we just forgot to add it?

trl/trainer/ppo_trainer.py
```diff
@@ -977,19 +996,16 @@ def train_minibatch(
...
         self.accelerator.backward(loss)
-
         if self.config.max_grad_norm is not None:
-            torch.nn.utils.clip_grad_norm_(
-                filter(lambda p: p.requires_grad, self.model.parameters()),
-                self.config.max_grad_norm,
-            )
-
-        t = time.time()
+            if self.accelerator.sync_gradients:
```
younesbelkada, line 1005: Nice!

younesbelkada (approved):
Thanks a lot for this refactor ! I think this makes sense, I agree this should make the new version to `0.5.0` for the introduced small breaking change. Let's maybe add 🚨🚨🚨 in the PR title so that it will be clear for developers that rely on TRL that this is a breaking change.
I also agree with: https://github.com/lvwerra/trl/pull/546/files#r1270503430 
I added one open question
Let's wait for @lvwerra 's review before merging
Once this gets merged, let's close: https://github.com/lvwerra/trl/pull/584
Thanks!

trl/trainer/ppo_trainer.py
vwxyzjn: Haha I guess we just forgot about it. OAI also did remove the gradient :) (https://github.com/openai/lm-human-preferences/blob/cbfd210bb8b08f6bc5c26878c10984b90f516c66/lm_human_preferences/train_policy.py#L339)

trl/core.py
vwxyzjn: I think we should just error out actually — it's gonna create nans anyway. Let's wait for @lvwerra's review and address this comment together

trl/trainer/ppo_trainer.py
lvwerra: I could swear it was there at some point :) thanks for putting it back!

trl/core.py
lvwerra: Yeah error out sounds good to me here!

vwxyzjn:
The latest commit works out! 

image

trl/trainer/ppo_trainer.py
```diff
@@ -659,83 +663,90 @@ def step(
...
-                        policykl = train_stats["policy/policykl"]
-                        early_stop = self._early_stop(policykl)
-                        if early_stop:
-                            break
+            b_inds = np.random.permutation(bs)
+            for backward_batch_start in range(0, bs, self.config.backward_batch_size):
+                backward_batch_end = backward_batch_start + self.config.backward_batch_size
+                backward_batch_inds = b_inds[backward_batch_start:backward_batch_end]
+
+                # set optimizer to zero for gradient accumulation
```
lvwerra, line 706: I don't think that belongs here anymore.
```suggestion
                # set optimizer to zero for gradient accumulation
```

lvwerra (approved):
PR looks super clean to me! Thanks a lot for fixing this!

younesbelkada (approved):
Thanks very much for this great work and investigation @vwxyzjn !

PR #547, opened 2023-07-20
ADD: num_proc to SFTTrainer

BramVanroy:
This tiny PR adds the optional argument `num_proc` to the SFTTrainer so that the mapped tokenization can occur in parallel if the user wishes that.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -89,7 +89,7 @@ class SFTTrainer(Trainer):
         packing (`Optional[bool]`):
             Used only in case `dataset_text_field` is passed. This argument is used by the `ConstantLengthDataset` to pack the sequences
             of the dataset.
-        num_proc (`Optional[int]`):
+        dataset_num_proc (`Optional[int]`):
             The number of workers to use to tokenize the data. Only used when `packing=False`. Defaults to None.
     """
 
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -294,7 +294,7 @@ class SFTTrainer(Trainer):
             return {"input_ids": outputs["input_ids"], "attention_mask": outputs["attention_mask"]}
 
         tokenized_dataset = dataset.map(
-            tokenize, batched=True, remove_columns=dataset.column_names, num_proc=self.num_proc
+            tokenize, batched=True, remove_columns=dataset.column_names, num_proc=self.dataset_num_proc
         )
 
         return tokenized_dataset
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -92,8 +92,8 @@ class SFTTrainer(Trainer):
         dataset_num_proc (`Optional[int]`):
             The number of workers to use to tokenize the data. Only used when `packing=False`. Defaults to None.
         dataset_batch_size (`int`):
-            Number of examples per batch provided to tokenize. If batch_size <= 0 or batch_size == None,
-             provide the full dataset as a single batch to tokenize. Defaults to 1000
+            The number of examples to tokenize per batch. If batch_size <= 0 or batch_size == None,
+            tokenize the full dataset as a single batch. Defaults to 1000.
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -89,6 +89,8 @@ class SFTTrainer(Trainer):
         packing (`Optional[bool]`):
             Used only in case `dataset_text_field` is passed. This argument is used by the `ConstantLengthDataset` to pack the sequences
             of the dataset.
+        num_proc (`Optional[int]`):
```
younesbelkada, line 92:
```suggestion
        dataset_num_proc (`Optional[int]`):
```

trl/trainer/sft_trainer.py
```diff
@@ -112,6 +114,7 @@ def __init__(
         infinite: Optional[bool] = False,
         num_of_sequences: Optional[int] = 1024,
         chars_per_token: Optional[float] = 3.6,
+        num_proc: Optional[int] = None,
```
younesbelkada, line 117:
```suggestion
        dataset_num_proc: Optional[int] = None,
```

trl/trainer/sft_trainer.py
```diff
@@ -160,6 +163,7 @@ def __init__(
                 f"You didn't pass a `max_seq_length` argument to the SFTTrainer, this will default to {max_seq_length}"
             )
 
+        self.num_proc = num_proc
```
younesbelkada, line 166:
```suggestion
        self. dataset_num_proc = dataset_num_proc
```

trl/trainer/sft_trainer.py
```diff
@@ -289,6 +293,8 @@ def tokenize(element):
 
             return {"input_ids": outputs["input_ids"], "attention_mask": outputs["attention_mask"]}
 
-        tokenized_dataset = dataset.map(tokenize, batched=True, remove_columns=dataset.column_names)
+        tokenized_dataset = dataset.map(
+            tokenize, batched=True, remove_columns=dataset.column_names, num_proc=self.num_proc
```
younesbelkada, line 297:
```suggestion
            tokenize, batched=True, remove_columns=dataset.column_names, num_proc=self. dataset_num_proc
```

younesbelkada (commented):
Thanks a lot @BramVanroy , this sounds like a great idea, I propose to change the variable name to some more explicit so that users are aware it is about the dataset, what do you think?

trl/trainer/sft_trainer.py
```diff
@@ -89,6 +89,11 @@ class SFTTrainer(Trainer):
         packing (`Optional[bool]`):
             Used only in case `dataset_text_field` is passed. This argument is used by the `ConstantLengthDataset` to pack the sequences
             of the dataset.
+        dataset_num_proc (`Optional[int]`):
+            The number of workers to use to tokenize the data. Only used when `packing=False`. Defaults to None.
+        dataset_batch_size (`int`):
+            Number of examples per batch provided to tokenize. If batch_size <= 0 or batch_size == None,
+             provide the full dataset as a single batch to tokenize. Defaults to 1000
```
tomaarsen, line 96:
```suggestion
            The number of examples to tokenize per batch. If batch_size <= 0 or batch_size == None,
            tokenize the full dataset as a single batch. Defaults to 1000.
```
Normalized the text to start in the same way as the `datset_num_proc`, simplified it a bit, removed a space too many at the start of the second sentence, and added a missing dot.

tomaarsen (approved):
Looks good, just some small nits with the docstring of `dataset_batch_size`.

lvwerra (approved):

tomaarsen:
For reference, I'm not a great fan of `**dataset_kwargs` simply because there's no autocomplete/autosuggest on IDEs for these parameters. Most users will never know they exist. But the way we have it now: it'll autocomplete and both parameters are individually documented.

BramVanroy:
@tomaarsen Good point and I definitely agree! It's sometimes a tough call to make though, in this case because [`map`](https://huggingface.co/docs/datasets/v2.13.1/en/package_reference/main_classes#datasets.Dataset.map) has _a lot_ of potential arguments. For users it would be best if they were all copied to the trainer's init but from a maintaining perspective, this becomes tedious and potentially "dangerous" to maintain in case the signature of `map` changes. Which means you'd have to pin the version of `datasets` quite prohibitively and verify at every new release that this indeed still works with SFTTrainer's arguments.

So generally I agree with you and user experience should be a high priority. But sometimes there are arguments for and against. In this case I am not sure what is best but for now having just batch_size and num_workers as the two only explicit arguments is perfectly fine with me!

PR #549, opened 2023-07-20
ADD: option to disable preparing data in trainer

BramVanroy:
Currently the tokenization is done _inside_ the SFTTrainer. While this is easy for users it takes a way some flexibility. It is also different from what users may be used to from the `transformers` Trainer.

With this backward-compatible change, an argument is added to the SFTTrainer, `prepare_datasets` that is enabled by default. If it is turned off, the trainer will _not_ automatically process the given datasets. In that case it is up to the user to provide a tokenized dataset.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -94,6 +94,10 @@ class SFTTrainer(Trainer):
         dataset_batch_size (`int`):
             The number of examples to tokenize per batch. If batch_size <= 0 or batch_size == None,
             tokenize the full dataset as a single batch. Defaults to 1000.
+        prepare_datasets (`bool`):
+            Whether to automatically prepare the given datasets based on the other given arguments, e.g. tokenization,
+            packing. You may want to disable this if your dataset has already been processed (e.g. tokenized) in
+            the format that you like. Defaults to `False`.
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


lvwerra:
Thanks @BramVanroy for the PR. However, I am not sure we want to add this right now. We want to give a painless high level experience to the user with the `STFTrainer`. For more sophisticated use-cases one can always use the `Trainer` that covers this use-case already. Also this is currently not compatible with the `ConstantLengthDataset` which expects raw texts and we would need a lot of exceptions with the preprocessing. Not sure this is worth it. What do you think? also cc @younesbelkada

younesbelkada (commented):
Thanks a lot for your PR, I also agree with @lvwerra - IMO this introduces too much modularity and flexibility, currently if a user wants to process a dataset by their own, they can create and process their own dataset and convert it to torch Dataset / dataloader: https://github.com/lvwerra/trl/blob/main/trl/trainer/sft_trainer.py#L244

BramVanroy:
@lwverra If I understand the flow correctly, a `ConstantLengthDataset` dataset will never be created because if `prepare_datasets=False` the dataset will not need to be created - the user just supplies the dataset.

`prepare_datasets=True` by default, which yields the same behavior as is currently the case. But allowing users to set it to False caters to more expert users who want to have more control over their dataset.

So this PR changes nothing for the basic user but provides more options for experts while keeping simplicity (i.e. they can just do dataset.map in their own code and then pass the result to the SFTTrainer). Seems like a win-win to me!

But of course it is fine if this is not something in the spirit of `trl` that is fine. I will close the PR.

PR #554, opened 2023-07-22
[DPO] remove unnecessary batch size arg to Collator

kashif:
removed unused argument to collator


younesbelkada (approved):
Thanks for cleaning this up !

Issue #462, opened 2023-06-23
Improve stability of default hyperparameters


vwxyzjn:
Oops accidentally pushed to main directly at [b56e8b3](https://github.com/lvwerra/trl/commit/b56e8b327733baa81c3ef0d6508f08e1b3e33939). This issue explained the changes.

As discovered in https://github.com/lvwerra/trl/issues/454#issuecomment-1604227258, we were probably doing too many minibatch updates. So this commit checks what happens if we do it with no minibatch updates, similar to how openai did it. When performing a benchmark, I found increased stability — only 1 out of 10 random seeds crashed.

Furthermore, I used a lower batch size and a target KL of 6 (the default setting in OAI's repo).

<img width="786" alt="image" src="https://github.com/lvwerra/trl/assets/5555347/bb25705e-ab89-440f-9446-adfe2131798f">

natolambert:
This makes sense as a reason that weird kl's could happen too.

vwxyzjn:
Thanks @natolambert!

PR #560, opened 2023-07-24
Add score scaling/normalization/clipping

zfang:
## Summary
Add score (aka reward) scaling/normalization/clipping to improve PPO training stability based on Section 5.3.1 of [Secrets of RLHF in Large Language Models Part I: PPO](https://arxiv.org/pdf/2307.04964.pdf) and https://github.com/OpenLMLab/MOSS-RLHF:
Screen Shot 2023-07-23 at 10 29 29 PM
Screen Shot 2023-07-24 at 12 18 39 PM

## Tests
The following is tested on a Google Colab notebook with an Nvidia T4 GPU.  My notebook gets disconnected by itself after a few hours while I was getting 1 iteration per minute, so my runs crashed fairly early.

#### Command for baseline:
```
python examples/scripts/sentiment_tuning.py --log_with wandb
```
Command for score scaling/normalization/clipping:
```
python examples/scripts/sentiment_tuning.py --log_with wandb --use_score_scaling --use_score_norm --score_clip 0.5
```

#### Screenshots of wandb:

Screen Shot 2023-07-23 at 10 20 37 PM
Screen Shot 2023-07-23 at 10 20 51 PM
Screen Shot 2023-07-23 at 10 21 56 PM

### `multi_adapter_rl_v2.py`
Command for baseline:
```
python examples/scripts/multi_adapter_rl_v2.py --model_name ../llama-7b --log_with wandb --use_safetensors
```
Command for score scaling/normalization/clipping:
```
python examples/scripts/multi_adapter_rl_v2.py --model_name ../llama-7b --log_with wandb --use_safetensors --use_score_scaling --use_score_norm --score_clip 0.5
```

#### Screenshots of wandb:

Screen Shot 2023-07-25 at 9 28 14 PM
Screen Shot 2023-07-25 at 9 28 28 PM
Screen Shot 2023-07-25 at 9 28 47 PM
Screen Shot 2023-07-25 at 9 29 02 PM
Screen Shot 2023-07-25 at 9 29 11 PM

### DIFF
--- a/examples/scripts/sentiment_tuning.py
+++ b/examples/scripts/sentiment_tuning.py
@@ -56,7 +56,9 @@ class ScriptArguments:
     target_kl: Optional[float] = field(default=0.1, metadata={"help": "kl target for early stopping"})
     seed: Optional[int] = field(default=0, metadata={"help": "the random seed"})
     use_score_scaling: Optional[bool] = field(default=False, metadata={"help": "Use score scaling"})
-    use_score_norm: Optional[bool] = field(default=False, metadata={"help": "Use score normalization"})
+    use_score_norm: Optional[bool] = field(
+        default=False, metadata={"help": "Use score normalization. Only applicable if use_score_scaling is True"}
+    )
     score_clip: Optional[float] = field(default=None, metadata={"help": "Score clipping"})
 
 
--- a/examples/scripts/sentiment_tuning.py
+++ b/examples/scripts/sentiment_tuning.py
@@ -81,6 +81,8 @@ config = PPOConfig(
     score_clip=script_args.score_clip,
 )
 
+# set seed before initializing value head for deterministic eval
+set_seed(config.seed)
 
 # We then define the arguments to pass to the sentiment analysis pipeline.
 # We set `return_all_scores` to True to get the sentiment score for each token.
@@ -133,9 +135,6 @@ dataset = build_dataset(config)
 def collator(data):
     return dict((key, [d[key] for d in data]) for key in data[0])
 
-
-# set seed before initializing value head for deterministic eval
-set_seed(config.seed)
 
 # Now let's build the model, the reference model, and the tokenizer.
 if not script_args.use_peft:
--- a/examples/scripts/sentiment_tuning.py
+++ b/examples/scripts/sentiment_tuning.py
@@ -45,7 +45,6 @@ class ScriptArguments:
         default=1, metadata={"help": "the number of gradient accumulation steps"}
     )
     early_stopping: Optional[bool] = field(default=False, metadata={"help": "whether to early stop"})
-    target_kl: Optional[float] = field(default=6, metadata={"help": "kl target for early stopping"})
     use_peft: Optional[bool] = field(default=False, metadata={"help": "whether to use peft"})
     use_seq2seq: Optional[bool] = field(default=False, metadata={"help": "whether to use seq2seq models"})
     kl_penalty: Optional[str] = field(
@@ -56,6 +55,9 @@ class ScriptArguments:
     )
     target_kl: Optional[float] = field(default=0.1, metadata={"help": "kl target for early stopping"})
     seed: Optional[int] = field(default=0, metadata={"help": "the random seed"})
+    use_score_scaling: Optional[bool] = field(default=False, metadata={"help": "Use score scaling"})
+    use_score_norm: Optional[bool] = field(default=False, metadata={"help": "Use score normalization"})
+    score_clip: Optional[float] = field(default=None, metadata={"help": "Score clipping"})
 
 
 parser = HfArgumentParser(ScriptArguments)
@@ -72,6 +74,9 @@ config = PPOConfig(
     target_kl=script_args.target_kl,
     kl_penalty=script_args.kl_penalty,
     seed=script_args.seed,
+    use_score_scaling=script_args.use_score_scaling,
+    use_score_norm=script_args.use_score_norm,
+    score_clip=script_args.score_clip,
 )
 
 
--- a/docs/source/customization.mdx
+++ b/docs/source/customization.mdx
@@ -195,3 +195,21 @@ Note that using `python -m torch.distributed.launch --nproc_per_node=1 reward_su
 ValueError: Some specified arguments are not used by the HfArgumentParser: ['--local-rank=0']
 ERROR:torch.distributed.elastic.multiprocessing.api:failed (exitcode: 1) local_rank: 0 (pid: 194889) of binary: /home/ubuntu/miniconda3/envs/trl/bin/python
 ```
+
+## Use score scaling/normalization/clipping
+As suggested by [Secrets of RLHF in Large Language Models Part I: PPO](https://arxiv.org/abs/2307.04964), we support score (aka reward) scaling/normalization/clipping to improve training stability via `PPOConfig`:
+```python
+from trl import PPOConfig
+
+ppo_config = {
+    use_score_scaling=True,
+    use_score_norm=True,
+    score_clip=0.5,
+}
+config = PPOConfig(**ppo_config)
+```
+
+To run `sentiment_tuning.py`, you can use the following command:
+```
+python examples/scripts/sentiment_tuning.py --log_with wandb --use_score_scaling --use_score_norm --score_clip 0.5
+```
--- a/examples/scripts/multi_adapter_rl_v2.py
+++ b/examples/scripts/multi_adapter_rl_v2.py
@@ -1,0 +1,147 @@
+# coding=utf-8
+# Copyright 2023 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from dataclasses import dataclass, field
+from typing import Optional
+
+import torch
+from datasets import load_dataset
+from peft import LoraConfig
+from tqdm import tqdm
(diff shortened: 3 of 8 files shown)


younesbelkada (commented):
Thanks a lot for working on this and adding this new nice feature. 
I am ok with this PR in principle that it is backward compatible with the existing setup and works in distributed setting as well (from the code of `RunningMoments` & `get_global_statistics`)
Would love to hear from @lvwerra & @vwxyzjn to hear their thoughts on this 
Can you also run the styling checks?

```bash
make precommit
```

Thanks!

younesbelkada (commented):
Can you also add few lines in the documentation explaining this feature to users ? The details could go in a dedicated section here: https://github.com/lvwerra/trl/blob/main/docs/source/customization.mdx 
Also can you share the behaviour of the `env/rewards_mean` and `env/rewards_std` ? 
Thanks!

examples/scripts/sentiment_tuning.py
```diff
@@ -45,7 +45,6 @@ class ScriptArguments:
         default=1, metadata={"help": "the number of gradient accumulation steps"}
     )
     early_stopping: Optional[bool] = field(default=False, metadata={"help": "whether to early stop"})
-    target_kl: Optional[float] = field(default=6, metadata={"help": "kl target for early stopping"})
```
younesbelkada, line 48: This field seems to have been removed by mistake?

zfang: Hi Younes,

You will find that `target_kl` already exists on L57 with a much smaller value.

I dug deeper and found that `PPOConfig` has two configs `target` and `target_kl`, where `target` has a default value of 6. So I assume the first duplicate `target_kl` config here was meant to be `target`. However, `target` is NOT used to populate PPOConfig at L64, so I just removed it.

Regards,

Felix

zfang:
> Can you also add few lines in the documentation explaining this feature to users ? The details could go in a dedicated section here: https://github.com/lvwerra/trl/blob/main/docs/source/customization.mdx Also can you share the behaviour of the `env/rewards_mean` and `env/rewards_std` ? Thanks!

Screen Shot 2023-07-24 at 11 35 11 AM
The `rewards` here are actually independent of score scaling/normalization/clipping because they are logged independently:
```
    rewards = [torch.tensor(output[1]["score"]) for output in pipe_outputs]

    # Run PPO step
    stats = ppo_trainer.step(query_tensors, response_tensors, rewards)
    ppo_trainer.log_stats(stats, batch, rewards)

...

    def log_stats(
        self,
        stats: dict,
        batch: dict,
        rewards: List[torch.FloatTensor],
    ):
            ...

            logs["env/reward_mean"] = torch.mean(rewards).cpu().numpy().item()
            logs["env/reward_std"] = torch.std(rewards).cpu().numpy().item()
            logs["env/reward_dist"] = rewards.cpu().numpy()

            ...
```

examples/scripts/sentiment_tuning.py
younesbelkada: great point, thank you !

younesbelkada (approved):
Thanks a lot for this great work, this looks very nice on my side, let's see what others will say !

examples/scripts/sentiment_tuning.py
lvwerra: I think this is actually a bug from here: https://github.com/lvwerra/trl/commit/1620da371a77e0c7030ff3093fa0d7814de575a0
we overloaded the `target_kl` term - we should rename it!

lvwerra: cc @edbeeching

examples/scripts/sentiment_tuning.py
```diff
@@ -56,6 +55,9 @@ class ScriptArguments:
     )
     target_kl: Optional[float] = field(default=0.1, metadata={"help": "kl target for early stopping"})
     seed: Optional[int] = field(default=0, metadata={"help": "the random seed"})
+    use_score_scaling: Optional[bool] = field(default=False, metadata={"help": "Use score scaling"})
+    use_score_norm: Optional[bool] = field(default=False, metadata={"help": "Use score normalization"})
```
lvwerra, line 59: Maybe we should clarify that this only works if `use_score_scaling` is also `True` otherwise it's actually ignored. we change the logic a bit in general

lvwerra (commented):
Very clean PR, thanks! left a few questions :)

examples/scripts/sentiment_tuning.py
zfang: Done

examples/scripts/sentiment_tuning.py
edbeeching: @lvwerra as much as I love introducing bugs into trl. I think this time it was @younesbelkada , in the **Big refactor of examples and documentation** (https://github.com/lvwerra/trl/pull/509). [Here](https://github.com/lvwerra/trl/commit/5c7bfbc8d9aeabee893290cc02121d7260636978#diff-ecb04b8f3ca12c6ad0d07746a2f836b9cacbe468fe7fb148b4148fb77af5a73e)

I agree to rename to `early_stop_kl`, or something

lvwerra:
> The rewards here are actually independent of score scaling/normalization/clipping because they are logged independently:

Is that really true? inside step we only log `scores` which are normalized with this PR. also if it weren't true then it looks like we have a strong performance degradation.

zfang:
> > The rewards here are actually independent of score scaling/normalization/clipping because they are logged independently:
> 
> Is that really true? inside step we only log `scores` which are normalized with this PR. also if it weren't true then it looks like we have a strong performance degradation.

Hi @lvwerra, 

Could you elaborate on the performance degradation?

In `sentiment-tuning.py` (and similarly `multi_adapter_rl_v2.py`), we have
```python
for epoch, batch in tqdm(enumerate(ppo_trainer.dataloader)):
    query_tensors = batch["input_ids"]

    # Get response from gpt2
    response_tensors = ppo_trainer.generate(query_tensors, return_prompt=False, **generation_kwargs)
    batch["response"] = tokenizer.batch_decode(response_tensors)

    # Compute sentiment score
    rewards = [torch.tensor(output[1]["score"]) for output in pipe_outputs]

    # Run PPO step
    stats = ppo_trainer.step(query_tensors, response_tensors, rewards)
    ppo_trainer.log_stats(stats, batch, rewards)
```

`env/reward_mean` and `env/reward_std` are logged inside of `ppo_trainer.log_stats`, which is based on the raw `rewards` from `sentiment_pipe`. I think the reason we observe different reward curves is randomness. Even in `ScriptArguments` we provide the `seed` config, it is only used in `PPOConfig` but not `LengthSampler` which can impact the input. I do observe near identical reward curves with `multi_adapter_rl_v2.py`
Screen Shot 2023-07-26 at 9 15 38 AM


`ppo/mean_scores` and `ppo/std_scores` are the per-batch (normalized) score stats that are logged inside of `ppo_trainer.step`. From the wandb screenshots you can see that the scores have a mean close to zero with a 0.5 std (clipping value).

It's not obvious to me whether score scaling/normalization/clipping improves or degrades the performance. It's meant to improve training stability but I guess I haven't run the training long enough to observe possible divergences (well Google Colab would crash on me). In general I observe more smooth curves.

I do observe better loss curves on the value head/function and assume that that can be attributed to the more stable and smooth reward scores. In `sentiment-tuning.py` I also observe better KL divergence and thus better non-score rewards assume that this is because normalized score rewards make them less dominant over the non-score rewards. In my opinion this makes it easier to configure the KL coefficient because we know what the range of score rewards are to expect.

Regardless, the configs are optional and are backward compatible.

Regards,

Felix

lvwerra:
Hi @zfang I was mainly referring to this plot that you shared: 
Screenshot 2023-08-08 at 14 05 54
It appears that the rewards are considerably lower and was wondering if that's due to scaling. Curious to hear your thoughts.

zfang:
> Hey @zfang thanks for the PR! Sometimes random seeds could impact the results a lot. E.g., [#462 (comment)](https://github.com/lvwerra/trl/issues/462#issue-1771440473) Could you run the experiment for 10 random seeds?

Actually I just made a change in `sentiment_tuning.py` to move `set_seed` before calling `build_dataset`. Re-running `sentiment-tuning.py` now.

zfang:
> Hi @zfang I was mainly referring to this plot that you shared: Screenshot 2023-08-08 at 14 05 54 It appears that the rewards are considerably lower and was wondering if that's due to scaling. Curious to hear your thoughts.

Update: I do observe consistent patterns of difference in `env/reward_std` and `env/reward_mean` with `sentiment-tuning.py`, but not `multi_adapter_rl_v2.py`

zfang:
> Hi @zfang I was mainly referring to this plot that you shared: Screenshot 2023-08-08 at 14 05 54 It appears that the rewards are considerably lower and was wondering if that's due to scaling. Curious to hear your thoughts.

Hi @lvwerra and @vwxyzjn,

After some investigations, I have the root cause. 

Based on the following code snippet
```python
for epoch, batch in tqdm(enumerate(ppo_trainer.dataloader)):
    query_tensors = batch["input_ids"]

    # Get response from gpt2
    response_tensors = ppo_trainer.generate(query_tensors, return_prompt=False, **generation_kwargs)
    batch["response"] = tokenizer.batch_decode(response_tensors)

    # Compute sentiment score
    rewards = [torch.tensor(output[1]["score"]) for output in pipe_outputs]

    # Run PPO step
    stats = ppo_trainer.step(query_tensors, response_tensors, rewards)
    ppo_trainer.log_stats(stats, batch, rewards)
``` 
We have the dependencies of `rewards` <- `pipe_outputs` <- `texts` <- `batch["response"]` <- `response_tensors` <- `ppo_trainer`. Because `ppo_trainer.step` updates `ppo_trainer` differently between baseline and score normalization, we start to get different `batch["response"]` and thus different `rewards`. 

In other words, because `gpt2-imdb` generates slightly different responses (perhaps less overly "positive" at the expense of higher KL loss) after PPO with score normalization, we start to also see different sentiment scores from `distilbert-imdb`.

On a high level I think that makes sense: we normalize the sentiment scores so it's less dominant over the KL loss, and thus we observe that with score normalization the model is less eager to optimize for sentiment scores in comparison to the KL loss. This can be adjusted by using a smaller `init_kl_coef`.

Screenshot 2023-08-08 at 4 18 44 PM
Screenshot 2023-08-08 at 4 18 54 PM


Hopefully that makes sense to you.

Regards,

Felix

lvwerra (approved):

PR #563, opened 2023-07-24
[`PPO`] fix corner cases with PPO batch size and forward_batch_size

younesbelkada:
Fixes: https://github.com/lvwerra/trl/issues/553

When users uses different combinations of batch size and forward batch size it might lead to breaking behaviours as explained on the attached issue.

This PR fixes it by adding the suggestion proposed by one of the user which is perfectly backward compatible

cc @lvwerra

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import inspect
+import math
 import os
 import time
 import typing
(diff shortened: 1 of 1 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -882,7 +883,7 @@ def batched_forward_pass(
         all_masks = []
         all_values = []
 
-        for i in range(int(bs / fbs)):
+        for i in range(math.ceil(bs / fbs)):
             input_kwargs = {key: value[i * fbs : (i + 1) * fbs] for key, value in model_inputs.items()}
```
lvwerra, line 887: That will lead to out of index errors further down no? What about giving a warning that grad-acc*mbs should be equal to batch size if it's not?

shubhlohiya: Can you mention lines where you see a potential issue? 
The only lines where `i` is referenced in this function is lines #886-888, where there shouldn't be any problem because in Python slicing for an example list `items`, `items[a:b] == items[a:]` if `b>=len(items)`, and for the last minibatch which might have length less than `fbs`, the change on line #903 takes care of the inner loop indexing according to current minibatch size. 

So, I don't see any possible out-of-index errors at least in this function.

lvwerra: Good point, was expecting an error there. Thanks for pointing this out!

lvwerra (approved):

PR #565, opened 2023-07-24
Add `DataCollatorForCompletionOnlyLM` in the docs

younesbelkada:
As per title and as discussed offline, this PR adds a section on the docs about how to FT a model using SFTTrainer on completions only.

cc @lvwerra


lvwerra (approved):
Awesome, thanks!

PR #566, opened 2023-07-24
Fix model output dim in reward trainer example

### LINKED ISSUE #558: Potential bug of model output dim in reward trainer example
Hi!

I've been exploring the `examples/scripts/reward_trainer.py` and believe I've found a potential bug that I'd like to address.

## Potential bug.
To my understanding, the reward model should produce a scalar output. However, when we load the pre-trained model in the [example script](https://github.com/lvwerra/trl/blob/main/examples/scripts/reward_trainer.py#L75-L80), the model will be a binary classifier by default, with a 2-dim logit output:

```python
import torch
inputs = tokenizer("Hello, my dog is cute", return_tensors="pt")
model = model.cpu()
with torch.no_grad():
    out = model(**inputs)
print(out.logits.shape) # torch.Size([1, 2])
```

During training, this issue was hidden by the [the loss computation part](https://github.com/lvwerra/trl/blob/main/trl/trainer/reward_trainer.py#L169-L176) of the reward trainer:

```python
    rewards_chosen = model(
    )[0]

    rewards_rejected = model(
    )[0]

    loss = -torch.nn.functional.logsigmoid(rewards_chosen - rewards_rejected).mean()
```
There, the `rewards_rejected` and `rewards_chosen` are dim [sample, 2] matrices, which I think should be dim [sample, 1] instead. The `.mean()` operator hides the issue as it returns a scalar value.

Here is a [colab](https://colab.research.google.com/drive/1szzGC8H1DQ7329_idVnnOnPIPE6D7J4j?usp=sharing) that shows the dimensionality problem.

## Proposed fix

To resolve this, in the model training example, we can use:

```python
model = AutoModelForSequenceClassification.from_pretrained(
    script_args.model_name,
    quantization_config=quantization_config,
    device_map=device_map,
    trust_remote_code=script_args.trust_remote_code,
    num_labels=1 # <---- make sure that the model output is a scalar
)
```
I'll submit a pull request if it is OK. 

In addition, while exploring `examples/scripts/reward_trainer.py`, I noticed a few other places that can be improved. For example, there is currently no validation steps that show the genearlization performance of the reward model; the `log_with` option in `ScriptArguments` is currently not being used anywhere and therefore has no effect. I'm happy to include the validation and fix the logging glitch in my pull request as well.

Tianlin

liutianlin0121:
Fixes #558

In `examples/scripts/reward_trainer.py`,
- set `num_labels=1` when loading the pre-trained model, so that the reward model produces a scalar output.
- add validation dataset
- correct a logging glitch

### DIFF
--- a/examples/scripts/reward_trainer.py
+++ b/examples/scripts/reward_trainer.py
@@ -46,7 +46,7 @@ class ScriptArguments:
     logging_steps: Optional[int] = field(default=500, metadata={"help": "the number of update steps between two logs"})
     eval_split: Optional[str] = field(
         default="none", metadata={"help": "the dataset split to evaluate on; default to 'none' (no evaluation)"}
-        )
+    )
     learning_rate: Optional[float] = field(default=1.41e-5, metadata={"help": "the learning rate"})
     batch_size: Optional[int] = field(default=64, metadata={"help": "the batch size"})
     num_train_epochs: Optional[int] = field(default=1, metadata={"help": "the number of training epochs"})
@@ -128,9 +128,7 @@ train_dataset = train_dataset.filter(
 if script_args.eval_split == "none":
     eval_dataset = None
 else:
-    eval_dataset = load_dataset(
-        script_args.dataset_name, split=script_args.eval_split
-    )
+    eval_dataset = load_dataset(script_args.dataset_name, split=script_args.eval_split)
 
     eval_dataset = eval_dataset.map(
         preprocess_function,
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Thanks a lot for this great contribution!

examples/scripts/reward_trainer.py
```diff
@@ -77,11 +79,14 @@ class ScriptArguments:
...
     device_map=device_map,
     trust_remote_code=script_args.trust_remote_code,
+    num_labels=1,
 )
 
 # Step 2: Load the dataset and pre-process it
 tokenizer = AutoTokenizer.from_pretrained(script_args.model_name)
-dataset = load_dataset(script_args.dataset_name, split="train")
+train_dataset = load_dataset(script_args.dataset_name, split="train")
+eval_dataset = load_dataset(script_args.dataset_name, split="test")
```
lvwerra, line 88: I think this should be optional - not every dataset has a `test` split. Maybe we can pass a `eval-split` to the cli arguments. by default it is none (no eval set used) and alternatively pass a split (e.g. `validation` or `test`). 

what do you think?

liutianlin0121: Good point. Done.

lvwerra (approved):

PR #571, opened 2023-07-25
[distributed training] conditional statement error for GPU allocation

halfrot:
should remove `not`
'''
        if not getattr(self.model, "is_sequential_parallel", False):
            self.current_device = self.accelerator.device
        else:
            self.current_device = torch.device("cuda:0")
'''

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -331,7 +331,7 @@ class PPOTrainer(BaseTrainer):
             self.highest_reward = torch.tensor(-float("inf"))
 
         # post process for PP
-        if not getattr(self.model, "is_sequential_parallel", False):
+        if getattr(self.model, "is_sequential_parallel", False):
             self.current_device = self.accelerator.device
         else:
             self.current_device = torch.device("cuda:0")


younesbelkada (commented):
Hi @halfrot 
Thanks for the PR, I am not sure if this is the right fix, the idea with sequential parallel is that if users are training a model in that case, the current device should be the first GPU device, hence `torch.device("cuda:0")

soumyasanyal:
Hi @halfrot @younesbelkada,

I'm facing the same issue as above. What is the intended fix for this case?

My issue:
I'm using `accelerate` with PPO training. Following example scripts, I've set `device_map = {'': Accelerator().local_process_index}` in my base model definition. As I can check [here](https://github.com/huggingface/trl/blob/3b4c24946b7d5580fd354b0e3800fc1047b82a41/trl/models/modeling_value_head.py#L269), this somehow sets `is_sequential_parallel = True` (I'm not sure how this exact process follows, didn't debug yet). But given that `is_sequential_parallel` is `True`, [this](https://github.com/huggingface/trl/blob/v0.8.6/trl/trainer/ppo_trainer.py#L355) code executes to `False`, and `self.current_device` is updated to `cuda:0`. This eventually leads to the issue that `response_tensors` are not in the right device as expected.

PR #573, opened 2023-07-25
Update sft_trainer.py

plaggy:
Looks like a minor bug in _prepare_non_packed_dataloader. Worked for me after this change.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -280,7 +280,7 @@ class SFTTrainer(Trainer):
     def _prepare_non_packed_dataloader(
         self, tokenizer, dataset, dataset_text_field, max_seq_len, formatting_func=None
     ):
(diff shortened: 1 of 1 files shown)


lvwerra (approved):
Thanks for the fix, that looks good to me. I let @younesbelkada confirm.

younesbelkada (commented):
Hi @plaggy 
Thanks for the PR, can you elaborate on the issue you were getting?
I think the current condition is correct, one should use the formatting function only if `formatting_func` is not `None` and `dataset_text_field` is `None`, the formatting function should take care of getting the proper text fields

plaggy:
Hi @younesbelkada, thank you for looking into it! Totally my bad, didn't look properly into the behaviour without the text field specified. Sorry for the false alarm :)

younesbelkada:
No problem ! @plaggy thanks again !

PR #576, opened 2023-07-26
Adds a Rejection Sampling Trainer and sentiment tuning example

edbeeching:
Still a draft PR as the following needs to be added:

- logging stats for timings of generation, forward, backward and opt steps
- Testing with encoder models
- validate the masking is done correctly on the loss
- configs renaming, currently I use a PPOConfig 
- the trainer uses an AutoModelWithValueHead. The value head is not needed as we are just doing text generation.
- results seem a bit too good to be true. perhaps validate in other settings

Tested in a 8 GPU 1 node setting:
https://wandb.ai/edbeeching/trl/runs/hq5sn0z2?workspace=user-edbeeching

Comparison to PPO sentiment tuning:
image

### DIFF
--- a/examples/scripts/sentiment_tuning_rs.py
+++ b/examples/scripts/sentiment_tuning_rs.py
@@ -1,0 +1,203 @@
+# coding=utf-8
+# Copyright 2023 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


lvwerra:
Idea looks really nice! I wonder if we should do the `best_of_n` outside `step` maybe with the [existing tools](https://github.com/lvwerra/trl/blob/main/trl/extras/best_of_n_sampler.py) and have a generic `IterativeSFTTrainer`? What do you think?

lvwerra:
Hi @edbeeching, as mentioned in https://github.com/huggingface/trl/pull/704#issuecomment-1705001973 I think it would be a great solution to have an `IterativeSFTTrainer` that we can use for several of these use cases that require some steps between optimizations. We then could accomodate the specific implementations in an example script rather than having dedicated trainers for each one of them. What do you think?

edbeeching:
Hey @lvwerra , ye I think this is a good idea. I have not had much bandwidth to look at this, but I will find some time later this week to take a look.

RylanSchaeffer:
@lvwerra @edbeeching I'm interested in this Rejection Sampling Trainer (motivated by Llama 2 and Llama 3) and I'm not clear how this can be implemented in the Iterative Trainer.

Does the Iterative Trainer support this Rejection Sampling-styled training approach? If so, is there documentation or a tutorial that demonstrates this?

PR #584, opened 2023-07-27
Fix grad accumulation

younesbelkada:
Fixes https://github.com/lvwerra/trl/issues/583

As reported by @Uncomfy, the tests used to pass within a single training step but fails if one uses larger dataset - leading to the gradient accumulation not working properly. 

Indeed one needs to perform gradient accumulation as showed in the accelerate tutorial

I propose to merge this PR as a hotfix before waiting for https://github.com/lvwerra/trl/pull/546 to get merged.

cc @lvwerra @edbeeching

### DIFF
--- a/tests/test_ppo_trainer.py
+++ b/tests/test_ppo_trainer.py
@@ -1153,6 +1153,11 @@ class PPOTrainerTester(unittest.TestCase):
 
         model_grad_acc = gpt2_model_clone.v_head.summary.weight.grad.clone()
         self.assertTrue(torch.allclose(model_grad_acc, model_grad, rtol=1e-3, atol=1e-3))
+        self.assertTrue(
+            torch.allclose(
+                gpt2_model_clone.v_head.summary.weight, gpt2_model.v_head.summary.weight, rtol=1e-3, atol=1e-3
+            )
+        )
 
     @unittest.skip("Fix by either patching `whomai()` to work in the staging endpoint or use a dummy prod user.")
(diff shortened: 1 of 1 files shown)


edbeeching (approved):
Thanks for fixing this, I mentioned it the other day but it makes sense to fix asap.
LGTM

tests/test_ppo_trainer.py
```diff
@@ -1151,7 +1150,6 @@ def test_grad_accumulation(self):
             reward = [torch.tensor(1.0), torch.tensor(1.0)]
             # train model by running a step twice
             _ = ppo_trainer.step([q for q in query_tensor], [r for r in response_tensor], reward)
-            break
 
         model_grad_acc = gpt2_model_clone.v_head.summary.weight.grad.clone()
         self.assertTrue(torch.allclose(model_grad_acc, model_grad, rtol=1e-3, atol=1e-3))
```
vwxyzjn, line 1155: A quick comment here: at some point I tried printing out the grads and they are all zeros here — I think it’s because we already did `optimizer.step`. We should be comparing the weights instead: 

```
self.assertTrue(torch.allclose(gpt2_model_clone.v_head.summary.weight, gpt2_model.v_head.summary.weight, rtol=1e-3, atol=1e-3))
```

lvwerra (approved):

tests/test_ppo_trainer.py
younesbelkada: Thanks, added that suggestion !

PR #585, opened 2023-07-27
Add missing max_seq_length arg to example sft_trainer.py

SharkWipf:
I believe this was supposed to be here already, as the `--seq_length` argument doesn't currently seem to be getting used, and there is no other way I can see to set max_seq_length.
I may be mistaken though, I'm not familiar with the code yet.


younesbelkada (approved):
Great point ! Thanks for this contribution !

Issue #579, opened 2023-07-26
Seq2Seq model training with DPO


gaetanlop:
Hello, 

I was looking at the DPO implementation to train T5 on a summarization task. It looks like both the DPOCollator and the DPOTrainer won't work for encoder-decoder architectures. Is it on purpose? Does DPO not work well for seq2seq models (I didn't find any experiments with T5 in their paper)?

If it is relevant I can try to work on an implementation that supports seq2seq models (it won't require that much changes from what I have seen).

younesbelkada:
Great point! 
Quickly looking at the DPOTrainer source code I don't think there is anything blocking for enc-dec models. I did not tested or tried anything though. Perhaps you can try to adapt the script: https://github.com/lvwerra/trl/blob/main/examples/dpo.py to a T5-like model and see if you face into any issue ! If it does not work out of the box then we'll need to think of how to adapt the trainer for seq2seq models
Maybe @kashif, @TevenLeScao & @tomaarsen can also share their thoughts as I am not 100% familiar with DPO
Thanks !

tomaarsen:
Hello!

I've also not experimented with any models other than ones that are loaded with `AutoModelForCausalLM` - I've only tested with GPT2 and some other tiny models.

Did you get this error?
```python
  File "[path]\transformers\models\t5\modeling_t5.py", line 987, in forward
    raise ValueError(f"You have to specify either {err_msg_prefix}input_ids or {err_msg_prefix}inputs_embeds")
ValueError: You have to specify either decoder_input_ids or decoder_inputs_embeds
```

- Tom Aarsen

gaetanlop:
Hello @tomaarsen @younesbelkada, T5 like models require labels in order to do teacher forcing + the prompt and the chosen/rejected responses should not be concatenated together. This is the case in the current DPO implementation. I have worked on something yesterday. I will make a pull request if that's okay for you.

younesbelkada:
Hi @gaetanlop @tomaarsen 
Thanks for all your replies,
That sounds great, looking forward to your PR !

kashif:
@gaetanlop thanks for this... yes currently we only tested with decoder only models and thus things were contacted together to avoid multiple forward calls... but you are right in the enc-dec setting those helpers would need to be extended appropriately... I do not think there is inherently anything in DPO stopping the use of enc-dec models so your PR will be a useful extension!

gaetanlop:
Thanks for the confirmation @kashif!

[...]

PR #586, opened 2023-07-28
Seq2Seq model support for DPO

gaetanlop:
Draft PR following #579. The goal is to make DPO support encoder-decoder models like T5. 

Things to add:
- Improve the collator so that it also returns the text version of the prompt/chosen/rejected samples.
- Add decoder input ids to the collator for models that do not directly create them from labels.
- Find a better way to know if the model is an encoder-decoder than directly passing a param in init.
- Modify the dpo example if needed.


trl/trainer/dpo_trainer.py
```diff
@@ -97,6 +97,8 @@ def __init__(
         preprocess_logits_for_metrics: Optional[Callable[[torch.Tensor, torch.Tensor], torch.Tensor]] = None,
         max_length: Optional[int] = None,
         max_prompt_length: Optional[int] = None,
+        max_target_length: Optional[int] = None,
+        is_encoder_decoder: bool = None,
```
tomaarsen, line 101:
```suggestion
```

tomaarsen: You can use `model.config.is_encoder_decoder`:
https://github.com/huggingface/transformers/blob/main/src/transformers/configuration_utils.py#L277-L278
All Configs have this option.

Do note however that in this `__init__` we might have a `model_init` instead of a `model`, so we'll have to work around that.

trl/trainer/dpo_trainer.py
```diff
@@ -128,13 +130,23 @@ def __init__(
                 )
                 max_prompt_length = 128
 
+            if max_target_length is None & is_encoder_decoder:
```
tomaarsen, line 133:
```suggestion
            if max_target_length is None and is_encoder_decoder:
```
+ you can use `model.is_encoder_decoder` if `model` is passed.

trl/trainer/dpo_trainer.py
```diff
@@ -189,24 +202,46 @@ def concatenated_inputs(self, batch: Dict[str, Union[List, torch.LongTensor]]) -
...
+                if k.startswith("rejected") and isinstance(batch[k], torch.Tensor):
+                    pad_value = self.label_pad_token_id if "labels" in k else self.padding_value
+                    concatenated_key = k.replace("rejected", "concatenated")
+                    concatenated_batch[concatenated_key] = torch.cat(
+                        (
+                            concatenated_batch[concatenated_key],
+                            pad_to_length(batch[k], max_length, pad_value=pad_value),
+                        ),
+                        dim=0,
+                    ).to(self.accelerator.device)
```
tomaarsen, line 244:
```suggestion
        if self.is_encoder_decoder:
            max_length = max(batch["chosen_labels"].shape[1], batch["rejected_labels"].shape[1])
            pad_value = self.label_pad_token_id 
        else:
            max_length = max(batch["chosen_input_ids"].shape[1], batch["rejected_input_ids"].shape[1])
            pad_value = self.label_pad_token_id if "labels" in k else self.padding_value

        for k in batch:
            if k.startswith("chosen") and isinstance(batch[k], torch.Tensor):
                concatenated_key = k.replace("chosen", "concatenated")
                concatenated_batch[concatenated_key] = pad_to_length(batch[k], max_length, pad_value=pad_value)
        for k in batch:
            if k.startswith("rejected") and isinstance(batch[k], torch.Tensor):
                concatenated_key = k.replace("rejected", "concatenated")
                concatenated_batch[concatenated_key] = torch.cat(
                    (
                        concatenated_batch[concatenated_key],
                        pad_to_length(batch[k], max_length, pad_value=pad_value),
                    ),
                    dim=0,
                ).to(self.accelerator.device)
            if self.is_encoder_decoder:
                concatenated_batch["concatenated_input_ids"] = batch["prompt_input_ids"].repeat(2,1)
                concatenated_batch["concatenated_attention_mask"] = batch["prompt_attention_mask"].repeat(2,1)
```
I believe this is equivalent.

trl/trainer/dpo_trainer.py
```diff
@@ -287,10 +323,19 @@ def concatenated_forward(
...
+            all_logits = model(
+                concatenated_batch["concatenated_input_ids"],
+                attention_mask=concatenated_batch["concatenated_attention_mask"],
+                labels = concatenated_batch["concatenated_labels"]
+            ).logits.to(torch.float32)
+        else:
+            all_logits = model(
+                concatenated_batch["concatenated_input_ids"],
+                attention_mask=concatenated_batch["concatenated_attention_mask"],
+            ).logits.to(torch.float32)
```
tomaarsen, line 337:
```suggestion
        model_kwargs = {"labels": concatenated_batch["concatenated_labels"]} if self.is_encoder_decoder else {}
        all_logits = model(
            concatenated_batch["concatenated_input_ids"],
            attention_mask=concatenated_batch["concatenated_attention_mask"],
            **model_kwargs,
        ).logits.to(torch.float32)
```
We can simplify this too, but this is also a bit of a style preference.

trl/trainer/utils.py
```diff
@@ -229,97 +231,124 @@ def tokenize_batch_element(
...
-                # adapted from https://stackoverflow.com/questions/73256206
-                if "prompt" in k:
-                    to_pad = [torch.LongTensor(ex[k][::-1]) for ex in batch]
-                else:
-                    to_pad = [torch.LongTensor(ex[k]) for ex in batch]
-                if k.endswith("_input_ids"):
+            if self.is_encoder_decoder:
+                to_pad = [torch.LongTensor(ex[k]) for ex in batch]
+            
+                if (k.startswith("prompt")) & (k.endswith("input_ids")):
```
tomaarsen, line 320:
```suggestion
                if k.startswith("prompt") and k.endswith("input_ids"):
```

trl/trainer/utils.py
```diff
@@ -229,97 +231,124 @@ def tokenize_batch_element(
...
                     padding_value = self.tokenizer.pad_token_id
-                elif k.endswith("_labels"):
-                    padding_value = self.label_pad_token_id
                 elif k.endswith("_attention_mask"):
-                    padding_value = self.padding_value
+                    padding_value = 0
+                elif (k.startswith("chosen")) | (k.startswith("rejected")):
```
tomaarsen, line 324:
```suggestion
                elif k.startswith("chosen") or k.startswith("rejected"):
```

tomaarsen (commented):
Hey! Thanks for starting this work. I wrote a few changes/simplifications that might be useful.

Another thing - this project uses `pre-commit`, which automatically runs some formatting scripts every time that you commit. You can use it with `pip install pre-commit` and `pre-commit install` to set it up (for this git repository only).
Otherwise the code quality CI will keep complaining.

If you're curious, this is what it runs: https://github.com/lvwerra/trl/blob/main/.pre-commit-config.yaml (p.s. maybe we can move to using ruff as well? It can replace isort).

gaetanlop:
Thanks @tomaarsen I will incorporate that.

trl/trainer/dpo_trainer.py
gaetanlop: I don't think this will work as k is not defined at this point for the pad_value

trl/trainer/dpo_trainer.py
gaetanlop: Indeed, thanks!

trl/trainer/dpo_trainer.py
tomaarsen: Oh, whoops. I missed that variable.
That said, the pad_value can be moved down somewhat in my solution, and it'll still remove some duplicated code.

gaetanlop: Yes I agree

trl/trainer/dpo_trainer.py
gaetanlop: Oh yes thanks I forget configs had this option

gaetanlop:
The PR is still not ready, I need to add docstrings for the different arguments added and fix the code quality issue that comes from some added imports.

tomaarsen:
You can try running `pre-commit run --all` and see if that makes the desired changes. For me it does. If that doesn't work, perhaps you can allow me to somehow push a commit to your branch and then I can resolve it for you, as I have the commit locally. I'm not quite sure how that works, though.

- Tom Aarsen

gaetanlop:
Thanks a lot @tomaarsen it worked!

kashif:
thanks @gaetanlop gonna start having a look. So perhaps you can also add something about this in the `dpo.mdx` docs too? Also if I understand now both model and ref models need to be either seq-2-seq or decoder only but not mixed?

gaetanlop:
Hey @kashif, yes good idea for the dpo.mdx docs. I am still unsure of how we should detect if a model_init is an encoder-decoder. I don't think there is a straightforward way to do it.
Exact, if I remember well in the DPO paper the model and the ref_model are the same model. Making it mixed will require further changes. Are you aware of a recent paper using DPO with different model and ref model architectures?

younesbelkada:
Hi @gaetanlop 
Thanks for the great work on this PR! Do you need any help finishing it up? 🙏 Let us know if you need any assistance!

gaetanlop:
Hi @younesbelkada, sorry have been busy with other things lately. I need to add information to the dpo.mdx and resolve different conflicts. I will do this over the weekend!

lvwerra:
Merge conflicts will be resolved probably if you merge `main` into your branch.

gaetanlop:
Hello @younesbelkada @lvwerra, if I do not miss anything I think it is ready for a review

younesbelkada (approved):
This looks great to me, thanks for your great work on adding Seq2Seq model support for DPO!

PR #587, opened 2023-07-28
Update the example sft_trainer.py

ZeusFSX:
Added save the model, because by default it saves only checkpoints not the final version.


younesbelkada (approved):
Thanks a lot for this contribution!

Issue #588, opened 2023-07-28
Could not find response key, But there are related keys in IDS


1518630367:
RuntimeError: Could not find response key [396, 673, 29901] in token IDs tensor([ 
         -100,     1,   673,   278,  1494,  2999,  7348,  1139,   491,  6820,
          278,  1556,  8210,  2933, 29889,   673,   881,   367,   697,  4249,
          518, 29909, 29892,   350, 29892,   315, 29892,   360, 29892,   382,
        29962,    13,    13, 16492, 29901,  1128,  1784, 10901, 11195,  1188,
         1140,   293,  1298,  6471,   526,   727,   297,  2211, 29899, 12531,
         2913, 29973,    13, 29871,    13,   319, 29897, 29871, 29955,    13,
           13,   350, 29897, 29871, 29941, 29906,    13,    13,   315, 29897,
        29871, 29896, 29946,    13,    13,   360, 29897, 29871, 29945,    13,
           13,   382, 29897, 29871, 29906, 29955,    13, 29871,    13,   **396,
          673, 29901**, 29871, 29941, 29906])

1518630367:
The code:
supervised_finetuning_trainer = SFTTrainer(
    base_model,
    train_dataset=data["train"],
    args=training_args,
    tokenizer=tokenizer,
    peft_config=qlora_config,
    dataset_text_field="text",
    max_seq_length=2048,
    data_collator=DataCollatorForCompletionOnlyLM(tokenizer=tokenizer, 
                                                  response_template="# Answer:")
)

lvwerra:
Can you provide the full script? To me it looks like your `text` field is already tokenized but it should be raw text.

1518630367:
This is my  full script .

> Can you provide the full script? To me it looks like your `text` field is already tokenized but it should be raw text.

Yes, This is my full script

tkreiman:
I looked through the source code, and it looks like the DataCollatorForCompletionOnlyLM is wrong. The code is currently written:

```
if self.response_token_ids == examples[i]["input_ids"][idx : idx + len(self.response_token_ids)]:
                        response_token_ids_start_idx = idx

                if response_token_ids_start_idx is None:
                    raise RuntimeError(
                        f'Could not find response key {self.response_token_ids} in token IDs {batch["labels"][i]}'
                    )
```

It should instead be something like:

`a = np.array(self.response_token_ids)
[...]

PR #594, opened 2023-07-31
Fix comparison in DataCollatorForCompletionOnlyLM (#588)

RyujiTamaki:
# What does this PR do?

Fixes https://github.com/lvwerra/trl/issues/588

This PR fixes issue https://github.com/lvwerra/trl/issues/588 encountered in the DataCollatorForCompletionOnlyLM class. Originally, the class compared self.response_token_ids with examples[i]["input_ids"], which caused a RuntimeError under certain circumstances.

This PR changes the comparison target to batch["labels"][i] from examples[i]["input_ids"] and add unit tests.
The added unit test reproduces the bug reported in the issue. Additionally, I have confirmed that the changes I made cause the added unit test to pass.

## Before submitting
- [x] `make precommit`
- [x] Add tests

Please review and let me know if any changes are required.

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import copy
 import os
 import tempfile
 import unittest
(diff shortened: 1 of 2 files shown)


younesbelkada (approved):
This looks great! thanks a lot @RyujiTamaki for leading this and fixing the bug with `padding_side="left"` !

PR #595, opened 2023-07-31
Use `accelerate` to clip gradient norm

lvwerra:
Follows best practice here: https://huggingface.co/docs/accelerate/v0.1.0/accelerator.html.

Should only really make a difference in case accelerator uses the following:
```python
 if self.state.use_fp16 and self.native_amp
```


vwxyzjn:
@lvwerra already added it in #546 :) https://github.com/lvwerra/trl/blob/a21d2c2d22e0045dfc845ef2f5b0ff8f02bf9c5d/trl/trainer/ppo_trainer.py#L1004-L1006

lvwerra:
Oops! Thanks @vwxyzjn!

PR #604, opened 2023-08-02
[`Modeling`] Add token support for `hf_hub_download`

younesbelkada:
Fixes https://github.com/lvwerra/trl/issues/602

As per title, let's add the `token` argument in the from_pretrained method to allow users to pass their HF tokens to download from private repos.

cc @lvwerra


lvwerra:
Isn't it usually `use_auth_token` in the `from_pretrained` method? Happy to add it in general :)

younesbelkada:
Use auth token is deprecated : https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L1714-1724 but happy to add it if you think it is relevant (maybe for consistency with transformers) wdyt?

lvwerra (approved):
didn't know that! no then it's fine.

PR #611, opened 2023-08-03
[DPO] stack-llama-2 training scripts

kashif:
Training scripts for the DPO training of stack-llama-2 via a 7B model

### DIFF
--- a/examples/research_projects/stack_llama_2/scripts/README.md
+++ b/examples/research_projects/stack_llama_2/scripts/README.md
@@ -1,0 +1,17 @@
+# DPO pipeline for the creation of StackLlaMa 2: a Stack exchange llama-v2-7b model
+
+## Prerequisites
+
+Install all the dependencies in the `requirements.txt`:
+
+```
+$ pip install -U -r requirements.txt
+```
+
+## Training
+
+There were two main steps to the training process:
+1. Supervised fine-tuning of the base llama-v2-7b model to create llama-v2-7b-se:
+    - `torchrun --nnodes 1  --nproc_per_node 8 examples/stack_llama_2/scripts/sft_llama2.py`
+1. Run the DPO trainer:
+    - `torchrun --nnodes 1  --nproc_per_node 8 examples/stack_llama_2/scripts/sft_llama2.py`
--- a/examples/research_projects/stack_llama_2/scripts/README.md
+++ b/examples/research_projects/stack_llama_2/scripts/README.md
@@ -21,9 +21,20 @@ There were two main steps to the DPO training process:
 1. Run the DPO trainer using the model saved by the previous step:
     - `accelerate launch examples/stack_llama_2/scripts/dpo_llama2.py --model_name_or_path="sft/final_checkpoint" --output_dir="dpo"`
 
+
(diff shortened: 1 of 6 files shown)


examples/research_projects/stack_llama_2/scripts/README.md
```diff
@@ -0,0 +1,17 @@
...
+$ pip install -U -r requirements.txt
+```
+
+## Training
+
+There were two main steps to the training process:
+1. Supervised fine-tuning of the base llama-v2-7b model to create llama-v2-7b-se:
+    - `torchrun --nnodes 1  --nproc_per_node 8 examples/stack_llama_2/scripts/sft_llama2.py`
+1. Run the DPO trainer:
+    - `torchrun --nnodes 1  --nproc_per_node 8 examples/stack_llama_2/scripts/sft_llama2.py`
```
lvwerra, line 17: I think since we are using `accelerate` the recommended way is to run `accelerate config` and then just `accelerate launch script.py`

if you want to can pass args at runtime:

```bash
accelerate launch --multi_gpu \
    --num_machines $NNODES \
    --num_processes $WORLD_SIZE \
    script.py
```

examples/research_projects/stack_llama_2/scripts/sft_llama2.py
```diff
@@ -0,0 +1,272 @@
...
+trainer.model.save_pretrained(output_dir)
+
+# Free memory for merging weights
+del base_model
+torch.cuda.empty_cache()
+
+model = AutoPeftModelForCausalLM.from_pretrained(
+    output_dir, device_map="auto", torch_dtype=torch.bfloat16
+)
+model = model.merge_and_unload()
```
lvwerra, line 269: For future work: could we just save the adapter and continue training the same one in DPO? Would save us from storing huge ckpts.

examples/research_projects/stack_llama_2/scripts/README.md
lvwerra, line 17: A frequent request is get is how to use the model after training. Maybe we can have a small example snippet like it is common in model cards here?

examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
```diff
@@ -0,0 +1,245 @@
...
+    script_args = parser.parse_args_into_dataclasses()[0]
+
+    # 1. load a pretrained model
+    model = AutoPeftModelForCausalLM.from_pretrained(
+        script_args.model_name_or_path,
+        low_cpu_mem_usage=True,
+        torch_dtype=torch.float16,
+        load_in_4bit=True,
+    )
+    model.config.use_cache = False
```
pacman100, line 171: `is_trainable` is False by default. So, none of the parameters are trainable unless I am missing something. @younesbelkada, can you please double check? Also, @kashif, you can check using `model.print_trainable_parameters()`

kashif: oh ok let me check!

lvwerra (approved):

PR #620, opened 2023-08-06
Use log_with argument in SFT example

hitorilabs:
It's in the script arguments, but not used

```
log_with: Optional[str] = field(default=None, metadata={"help": "use 'wandb' to log with wandb"})
```


lvwerra (approved):
Good catch, thanks for fixing!

Issue #624, opened 2023-08-08
Would DPOTrainer with a single model in memory be possible?


vincentmin:
Currently the DPOTrainer needs two models in memory: 1. the model to be trained, 2. the reference model. The reference model is frozen here. I was thinking that it could be possible to load the model only once in memory and have a single LoRA adapter. When the model needs to be used in reference mode, we simply turn off the adapter. The base model would be frozen and only the LoRA adapter would be trainable.

Is there a theoretical obstruction to doing this? If not, is there a significant overhead for having to turn the adapter on and off during training? If the answer is no to both of these questions, I think it would be interesting to create an implementation of such a DPOTrainer as it would cut the required vram in half and therefore further democratise llm finetuning.

lvwerra:
Indeed we could and it's what we are doing in the `PPOTrainer`. What do you think @younesbelkada and @kashif?

kashif:
@vincentmin Yes in theory that would be indeed possible and something we should do! Do you want to have a crack at it?

vincentmin:
Good to hear this is possible. I am restricted in my free time at the moment (life with a newborn 😄), so I'll let someone else take a crack at it.

vincentmin:
Actually, let me give this a shot. It seems to me it shouldn't take so much time.

Do you prefer a new DPOTrainer class, or to modify the input arguments to the current DPOTrainer?

lvwerra:
Probably similar to the `PPOTrainer` should work: if model is a peft model and there's no reference model we deactivate the adapter for the reference model.

vincentmin:
PR is live! #640

PR #639, opened 2023-08-11
Disable dropout in DPO Training

NouamaneTazi:
Helps stabilising training.
For now, I only added `disable_dropout` to DPO Training, but let me know if you think it should go to `DPOTrainer` with a config to disable

cc @kashif

PR #640, opened 2023-08-11
Allow for ref_model=None in DPOTrainer

vincentmin:
PR for #624. The goal is to allow DPOTrainer to allow for `ref_model=None`. If the model to be trained is a `peft` model, or a `peft_config` is provided, we use the base model for improved memory consumption. Otherwise, a reference model is created on the fly by copying the `model`.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -39,7 +39,7 @@ class DPOTrainer(Trainer):
         model (`transformers.PreTrainedModel`):
             The model to train, preferably an `AutoModelForSequenceClassification`.
         ref_model (`PreTrainedModelWrapper`):
-            Hugging Face transformer model with a casual language modelling head. Used for implicit reward computation and loss.
+            Hugging Face transformer model with a casual language modelling head. Used for implicit reward computation and loss. If no `ref_model` is provided, `model` without adapters is used as the reference model (only possible if peft_config is provided).
         beta (`float`, defaults to 0.1):
             The beta factor in DPO loss. Higher beta means less divergence from the initial policy.
         args (`transformers.TrainingArguments`):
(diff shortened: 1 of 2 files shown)


vincentmin:
@lvwerra I used a bit of a different approach from the `PPOTrainer`. In `PPOTrainer` `self.ref_model` can be None in which case `self.accelerator.unwrap_model(self.model).pretrained_model.disable_adapter()` is used. Instead, I just set `self.ref_model=model.get_base_model()`. Let me know if you see any problems with this approach.

kashif:
will do! thanks! i wanted to test it out together with the PR #639

younesbelkada (commented):
Great work @vincentmin for adding this support, there is a small merge conflict, would you be happy to take a look? Otherwise happy to do it!

vincentmin:
@younesbelkada Thanks. The conflict is resolved.

younesbelkada (commented):
Hmm a CI test is failing, you might need to add a simple check here: https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L158 to make the CI happy

trl/trainer/dpo_trainer.py
```diff
@@ -155,13 +167,13 @@ def __init__(
 
         if disable_dropout:
             disable_dropout_in_model(model)
-            disable_dropout_in_model(ref_model)
+            if self.ref_model:
```
younesbelkada, line 170:
```suggestion
            if self.ref_model is not None:
```

younesbelkada (commented):
Thanks a lot ! Left a last suggestion, what do you think?

tannonk:
@vincentmin Thanks for the effort on this PR! Any chance you could provide the script/run command you used to test the implementation? I'm yet to get this working from my end.

vincentmin:
If you modify [this script](https://github.com/huggingface/trl/blob/main/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py) by setting `model_ref=None` on line 141, it should work. I haven't had a chance to try this myself, so let me know if you encounter an error.

vincentmin:
@lvwerra @younesbelkada Could you give your input here? @tannonk 's proposed fix sounds reasonable to me, but I do not have the time to verify it at the moment. If this is indeed a correct fix, I would expect this to be necessary for the PPOTrainer as well.

kashif:
thanks @tannonk I can test it out

PR #644, opened 2023-08-14
Handle potentially long sequences with DataCollatorForCompletionOnlyLM

### LINKED ISSUE #643: Exception handling in DataCollatorForCompletionOnlyLM
Hey,

While testing out the new `DataCollatorForCompletionOnlyLM` on custom data, I had some issues relating to the exception handling of batch instances in which we fail to find the `response_template` or `instruction_template` (https://github.com/huggingface/trl/blob/3b2c820db68aecca36c2ff5d152de59deec8fe5c/trl/trainer/utils.py#L142C3-L142C3). 

It turned out that I had some exceptionally long responses in my training data which didn't actually contain both a valid `response_template` and `instruction_template`.

**Possible fix:**

My current work around was to replace the `RuntimeError` with a more informative `warning` and simply set the labels for a problematic instance to `self.ignore_index`, so that it will be ignored for the loss calculation but training will still run successfully (provided this doesn't happen for every instance 😉 ), e.g.:

```python
if len(human_token_ids_idxs) == 0:
    warnings.warn(
        f'Could not find instruction key `{self.instruction_template}` in the following instance {self.tokenizer.decode(batch["input_ids"][i])} ' \
        f'This instance will be ignored in loss calculation.'
    )
    batch["labels"][i, :] = self.ignore_index
```

Would something like this be more suitable than the current exception handling?

tannonk:
This PR resolves #643 and provides a patch in `DataCollatorForCompletionOnlyLM` for handling long sequences which may or may not contain a valid `response_template` or a valid `instruction_template`.

For problematic instances where no `response_template` or `instruction_template` is found, we set the labels to the `ignore_idx`. As a result, problematic instances are ignored in the loss computation, but still allow for training to continue.

### DIFF
--- a/tests/test_data_collator_completion_only.py
+++ b/tests/test_data_collator_completion_only.py
@@ -49,15 +49,14 @@ class DataCollatorForCompletionOnlyLMTester(unittest.TestCase):
 ### User: How much is 2+2? I'm asking because I'm not sure. And I'm not sure because I'm not good at math.
 """
         self.response_template = "\n### Assistant:"
-
-        # check DataCollatorForCompletionOnlyLM with response template only
+        # check DataCollatorForCompletionOnlyLM using response template only
         self.tokenized_instruction = self.tokenizer.encode(self.instruction, add_special_tokens=False)
         self.collator = DataCollatorForCompletionOnlyLM(self.response_template, tokenizer=self.tokenizer)
         encoded_instance = self.collator.torch_call([self.tokenized_instruction])
         result = torch.all(encoded_instance["labels"] == -100)
         self.assertTrue(result, "Not all values in the tensor are -100.")
 
-        # check DataCollatorForCompletionOnlyLM with response template and instruction template
+        # check DataCollatorForCompletionOnlyLM using response template and instruction template
         self.instruction_template = "\n### User:"
         self.collator = DataCollatorForCompletionOnlyLM(
             self.response_template, self.instruction_template, tokenizer=self.tokenizer
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -132,7 +132,7 @@ class DataCollatorForCompletionOnlyLM(DataCollatorForLanguageModeling):
                     ):
                         response_token_ids_idxs.append(assistant_idx + len(self.response_token_ids))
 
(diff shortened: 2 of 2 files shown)


tests/test_data_collator_completion_only.py
```diff
@@ -40,3 +41,29 @@ def test_data_collator_finds_response_template_llama2_tokenizer(self):
         # Pass already tokenized (w context) and truncated response_template so token_ids are like in the instruction + response
         self.collator = DataCollatorForCompletionOnlyLM(self.tokenized_response_w_context, tokenizer=self.tokenizer)
         self.collator.torch_call([self.tokenized_instruction])
+
+    def test_data_collator_handling_of_long_sequences(self):
+        self.tokenizer = AutoTokenizer.from_pretrained("upstage/Llama-2-70b-instruct-v2")
```
lvwerra, line 46: Can we change this to
```suggestion
        self.tokenizer = AutoTokenizer.from_pretrained("trl-internal-testing/dummy-GPT2-correct-vocab")
```
So we don't depend on external repositories. Could you also change it in the test above, it seems that slipped review when it was added.

lvwerra (approved):
Looks good to me, thanks for fixing! Just a small nit then we can merge.

lvwerra:
Also could you run the code quality tests with `make precommit`? That should fix the CI.

tannonk:
Hey @lvwerra, thanks for the review. The requested changes have been made and `precommit` passes on my end.

lvwerra:
Looks like the tests are not passing, yet.

younesbelkada (commented):
Hi @tannonk 
Thanks a lot for the PR! 
In my opinion, currently raising a `RuntimeError` really helps for understanding whether things worked correctly or not for users, I am afraid the warning is not a strong enough indicator to users. Maybe it is better to keep it and add a flag `allow_ignore_not_matched` for advanced users and keep the previous behaviour untouched, what do you think? cc @lvwerra

younesbelkada (approved):
Discussed with @lvwerra , indeed if the sequence is too long the training would completely break and this could happen in the middle of a training, which is too odd. Let's merge it! Thanks again @tannonk for your work on this!

PR #649, opened 2023-08-16
fix PeftConfig loading from a remote repo.

w32zhong:
I am using peft 0.4.0 (currently the most updated one), it requires `PeftConfig.from_pretrained()` to be passed a directory, instead of the config file path.


lvwerra:
Thanks for the PR! I let @younesbelkada review it as he built that part of the repo - he's back from vacation next week.

trl/models/modeling_base.py
```diff
@@ -263,7 +264,7 @@ class and the arguments that are specific to trl models. The kwargs
                                 "pytorch_model.bin.index.json",
                                 token=token,
                             )
-                        except ValueError:  # not continue training, do not have v_head weight
+                        except:  # not continue training, do not have v_head weight
```
younesbelkada, line 267: Can you elaborate why this change is needed?

younesbelkada (commented):
This looks good I think, left one comment, also can you run `make precommit` ? Thanks!

trl/models/modeling_base.py
w32zhong: Hi @younesbelkada, thanks for getting back! This change has two reasons:

* At least in huggingface-hub 0.10.1, the error for "not found" is huggingface_hub.utils._errors.EntryNotFoundError, not a ValueError. This makes it fail to catch hf_hub_download() error when an entry is not found.
* This is also consistent with the code above (when trying to catch a similar scenario): https://github.com/huggingface/trl/pull/649/files#diff-be4f28845d341065eb254da5db9fa8fee2e6b5347071ef98448f27acd70fe8a7L166

younesbelkada: Thanks, that makes sense, would you be happy to import `EntryNotFoundError` from huggingface_hub similarly as here: https://github.com/huggingface/transformers/blob/main/src/transformers/utils/hub.py#L43 and replace the bare excepts with `except EntryNotFoundError:`? 🙏

w32zhong: Hi @younesbelkada, I am thinking this would be a little tricky, as you might notice, there are [many error codes](https://github.com/huggingface/transformers/blob/6b82d936d49956ba7b43c5ee590f4868de373b65/src/transformers/utils/hub.py#L47C25-L47C25), and the logics / handler is the same for any of them? So bare excepts will be reasonable in this case? Or we can just use the general `Exception` form?

w32zhong: okay, I see only two error codes are sufficient for the purpose. I have made the proposed changes and applied it to other places in that file as well. Please review it in my most recent commit.

younesbelkada (commented):
Hi @w32zhong 
Thanks a lot for iterating, sadly some CIs are failing, you can try to run:

```bash
pytest tests/test_peft_models.py::PeftModelTester::test_continue_training_peft_model
```

Can you please try to have a look 🙏 Let me know if you need any help !

w32zhong:
Looks like this test failed:
```sh
pytest tests/test_reward_trainer.py::RewardTrainerTester::test_reward_trainer_assert_value_error
```
But a local run gives me error on the missing wandb api key. Any idea how to proceed?

younesbelkada:
Hi @w32zhong 
Thanks a lot for iterating, regarding the failing test it is unrelated to your PR, it should be solved with https://github.com/huggingface/trl/pull/676 
Once #676 gets merged, you can merge your branch with upstream main and the CI will be green once again

younesbelkada:
Hi @w32zhong 
You can now merge with main branch and the CI will be green

w32zhong:
@younesbelkada done.

younesbelkada (approved):
Looking great thank you!

PR #656, opened 2023-08-17
[`SFTTrainer`] Flash attention support for SFTTrainer

younesbelkada:
Related to https://github.com/huggingface/transformers/pull/25265

Users can easily benefit from flash attention, this PR adds a new argument in `SFTTrainer` to take care of that and properly document and raise errors when relevant. 

This leads to some interesting speedups and memory saving that I will detail after doing some experiments, hence putting this PR as draft for now

Only available if you use pytorch nightlies and use `packing=True` as SDPA + flash attention does not support padding

cc @lvwerra @fxmarty @vwxyzjn

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -347,6 +347,7 @@ class SFTTrainer(Trainer):
             training_contexts.append(
                 torch.backends.cuda.sdp_kernel(enable_flash=True, enable_math=False, enable_mem_efficient=False)
             )
+            training_contexts.append(torch.cuda.amp.autocast())
 
         with ContextManagers(training_contexts):
             super().train(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


fxmarty:
Are trl users using padding?

younesbelkada:
adding `packing=True` should lead to having no padding in the input texts as it will concatenate text chunks until reaching `max_seq_length`

trl/trainer/sft_trainer.py
```diff
@@ -29,6 +30,7 @@
 )
 from transformers.trainer_callback import TrainerCallback
 from transformers.trainer_utils import EvalPrediction
+from transformers.utils import ContextManagers
```
lvwerra, line 33: I made this 😂

younesbelkada: hahahah nice!

younesbelkada:
Closing for now as https://github.com/huggingface/transformers/pull/25598 might be merged

Issue #667, opened 2023-08-21
Error in trl 0.5.0


coralexbadea:
If you copy paste the code from hugging face SFTTrainer page ([Supervised Fine-tuning Trainer](https://huggingface.co/docs/trl/main/en/sft_trainer)) it results in error:
```
name ‘PeftModel’ is not defined
```

Code:
```
trainer = SFTTrainer(
“facebook/opt-350m”,
train_dataset=dataset,
dataset_text_field=“text”,
max_seq_length=512,
)
trainer.train()
```

If I downgrade to 0.4.0 the issue is no longer there.

younesbelkada:
Thanks @coralexbadea , I have managed to reproduce, meanwhile you can install PEFT `pip install peft` to overcome the issue

PR #668, opened 2023-08-21
[`SFTTrainer`] Check correctly for condition

younesbelkada:
Fixes https://github.com/huggingface/trl/issues/667 
Fixes https://github.com/huggingface/trl/issues/735

if PEFT is not installed SFTTrainer will throw an error when checking `isinstance(model, (PreTrainedModel, PeftModel))` as `PeftModel` is defined nowhere in the script if PEFT is not available

cc @lvwerra


lvwerra (approved):

lvwerra:
Maybe we can mock in the tests that peft is not installed and test that it still runs?

younesbelkada:
Will be doing it in a follow up PR (to unlock #735)

PR #672, opened 2023-08-22
Update README.md

insublee:
Added instructions for google colab


lvwerra:
Since we don't link notebooks on the Readme I think `installation.mdx` would be better suited for the information. What do you think?

PR #675, opened 2023-08-22
propagating eval_batch_size to TrainingArguments

### LINKED ISSUE #671: Tutorial script missing eval_batch_size
Following this tutorial https://huggingface.co/blog/dpo-trl, I ran into out of memory issues with the DPO script. The problem was that the script missed copying per_device_eval_batch_size to Training Arguments [here](https://github.com/huggingface/trl/blob/main/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py#L166C12-L166C12). This makes it use the default size of 8 for eval_batch_size instead of 1 as set in this script. The below addition fixes the issue.

```
        per_device_train_batch_size=script_args.per_device_train_batch_size,
-->        per_device_eval_batch_size=script_args.per_device_eval_batch_size,
```

rahuljha:
The script missed copying `per_device_eval_batch_size` to `training_args`, which makes it use the default size of 8 for eval_batch_size instead of 1 as set in this script. This PR fixes the issue by copying the `per_device_eval_batch_size` from `script_args` to `training_args`.

fixes #671

### DIFF
--- a/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
+++ b/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
@@ -164,6 +164,7 @@ if __name__ == "__main__":
     # 4. initialize training arguments:
     training_args = TrainingArguments(
         per_device_train_batch_size=script_args.per_device_train_batch_size,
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Thanks!

PR #678, opened 2023-08-23
Update sft_llama2.py

msaad02:
Add argument num_workers. Fixed error on line 103 if streaming set = False


younesbelkada (approved):
Thanks!

Issue #680, opened 2023-08-23
PPOTrainer gives an error when using CUDA


backpropper:
The step function in the PPOTrainer gives an error when using a GPU. Specifically, the `scores` tensor [here](https://github.com/huggingface/trl/blob/c115453fba481cc8a9212a4abd56904b08316913/trl/trainer/ppo_trainer.py#L593) is not loaded to the GPU which makes it a `cpu` tensor and gives an error when gathering all the stats [here](https://github.com/huggingface/trl/blob/c115453fba481cc8a9212a4abd56904b08316913/trl/trainer/ppo_trainer.py#L865).

mnoukhov:
I haven't encountered this. Could be a recent regression?

Can you give basic steps to repro the error?

PR #681, opened 2023-08-23
fix device issue

backpropper:
Fix for #680


trl/trainer/ppo_trainer.py
```diff
@@ -590,7 +590,8 @@ def step(
         bs = self.config.batch_size
 
         queries, responses, scores = self._step_safety_checker(bs, queries, responses, scores)
-        scores = torch.tensor(scores)
+        scores_device = scores[0].device
+        scores = torch.tensor(scores).to(device=scores_device)
```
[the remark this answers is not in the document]
backpropper: It doesn't work as scores is a list of one-dimensional tensors and pytorch doesn't allow concatenating 1D tensors [here](https://stackoverflow.com/questions/75644077/pytorch-runtimeerror-zero-dimensional-tensor-at-position-0-cannot-be-concate). It also requires specifying the `device` when creating the `rewards` tensor (modified one file in the PR but will require changes to other examples as well)

The error is a bit sneaky one, it only occurs when you call the script using `accelerate` using the default config given [here](https://huggingface.co/docs/accelerate/main/en/basic_tutorials/launch#:~:text=For%20example%2C%20here%20is%20how%20to%20use%20accelerate,to%20manually%20pass%20in%20the%20right%20configuration%20parameters.)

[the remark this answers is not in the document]
malik-ali: @lvwerra I just found this as well (see #700). You can't `torch.cat` but you could use `torch.stack`. 

Either way, one way this issue shows up is in the distributed gather of stats across multiple gpus. This requires the `scores` to be on a cuda device and fails with the following error:

```
RuntimeError: Tensors must be CUDA and dense
```

Simple fix:

```
scores = torch.tensor(scores, device=self.current_device)
```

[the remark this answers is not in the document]
younesbelkada: Hmm the score tensor should be sent to the correct device here no? https://github.com/huggingface/trl/blob/main/trl/trainer/ppo_trainer.py#L558 
Does the issue happen if you correctly select "Multi-GPU" when launching `accelerate config` and running right after with accelerate launch?

younesbelkada (approved):
Looks good to me, thanks for explaining the issue! I would like to have a second look from @lvwerra and @vwxyzjn before merging

trl/trainer/ppo_trainer.py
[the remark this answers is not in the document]
malik-ali: @younesbelkada 

What you linked is correct but occurs in the `_step_safety_checker` function. The issue is on the following block of code https://github.com/huggingface/trl/blob/main/trl/trainer/ppo_trainer.py#L593:
```
        queries, responses, scores = self._step_safety_checker(bs, queries, responses, scores)
        scores = torch.tensor(scores)
```
Notice this modifies the return value of `_step_safety_checker` and puts all the tensors on the cpu. This was added recently to include reward scaling in [this commit](https://github.com/huggingface/trl/commit/3b2c820db68aecca36c2ff5d152de59deec8fe5c)

malik-ali:
I also think you should implement this as @lvwerra suggested, without accessing the first entry of the list. I think the rewards list could be empty and this would crash. It would be cleaner to use `self.current_device` or to use `stack`:

```
scores = torch.tensor(scores, device=self.current_device)
# or
scores = torch.stack(scores)
```

examples/scripts/sentiment_tuning.py
```diff
@@ -197,7 +197,7 @@ def collator(data):
     # Compute sentiment score
     texts = [q + r for q, r in zip(batch["query"], batch["response"])]
     pipe_outputs = sentiment_pipe(texts, **sent_kwargs)
-    rewards = [torch.tensor(output[1]["score"]) for output in pipe_outputs]
+    rewards = [torch.tensor(output[1]["score"], device=device) for output in pipe_outputs]
```
malik-ali, line 200: I think you don't need this change. The `_step_safety_checker` puts all the tensors on the right device

trl/trainer/ppo_trainer.py
```diff
@@ -590,7 +590,8 @@ def step(
         bs = self.config.batch_size
 
         queries, responses, scores = self._step_safety_checker(bs, queries, responses, scores)
-        scores = torch.tensor(scores)
+        scores_device = scores[0].device
```
malik-ali, line 593: It would be safer to not access the first entry in the list and just use `self.current_device`

malik-ali (changes requested):

trl/trainer/ppo_trainer.py
lvwerra: that sounds like a good idea!

backpropper: I agree the fix is much better. Can you suggest a change to the PR, I can add it then.

malik-ali: Thanks for fixing this issue!

younesbelkada (commented):
Thanks, the changes look good to me, can you try to push a new commit after running `make precommit` ?

backpropper:
Thanks I just did :)

younesbelkada (approved):
This looks clean to me, thanks!

Issue #701, opened 2023-08-27
DPO with chatglm2-6b


valkryhx:
When training a chatglm2-6b model (sft with qlora) with DPO, the loss did not seem to get low ?

Below is part of training log, I find there is only small jitter on the training loss and the rewards/margins  seems really really small ?

Is it the expected behavior?

And ,if I set ref_model=None in the DPOTrainer ,when eval, the eval_loss and eval_rewards and etc will always keep the same during training. It does not seem right...

{'loss': 0.6924, 'learning_rate': 5e-06, 'rewards/chosen': 0.008880767971277237, 'rewards/rejected': 0.007253284566104412, 'rewards/accuracies': 0.44999998807907104, 'rewards/margins': 0.0016274836380034685, 'logps/rejected': -82.50346374511719, 'logps/chosen': -81.14619445800781, 'logits/rejected': -2.220956802368164, 'logits/chosen': -2.227485179901123, 'epoch': 0.01}
{'loss': 0.6951, 'learning_rate': 1e-05, 'rewards/chosen': 0.006737555377185345, 'rewards/rejected': 0.010529976338148117, 'rewards/accuracies': 0.5, 'rewards/margins': -0.0037924200296401978, 'logps/rejected': -98.62718200683594, 'logps/chosen': -75.55088806152344, 'logits/rejected': -2.2945332527160645, 'logits/chosen': -2.1720192432403564, 'epoch': 0.02}
{'loss': 0.6942, 'learning_rate': 9.9999992772835e-06, 'rewards/chosen': -0.0009587286040186882, 'rewards/rejected': 0.0010564230615273118, 'rewards/accuracies': 0.44999998807907104, 'rewards/margins': -0.0020151520147919655, 'logps/rejected': -86.17352294921875, 'logps/chosen': -69.18254089355469, 'logits/rejected': -2.2097086906433105, 'logits/chosen': -2.1535096168518066, 'epoch': 0.03}
{'loss': 0.6928, 'learning_rate': 9.999997109134203e-06, 'rewards/chosen': 0.0020176221150904894, 'rewards/rejected': 0.0011578561970964074, 'rewards/accuracies': 0.550000011920929, 'rewards/margins': 0.0008597659762017429, 'logps/rejected': -67.88219451904297, 'logps/chosen': -85.24455261230469, 'logits/rejected': -2.137803792953491, 'logits/chosen': -2.1222000122070312, 'epoch': 0.03}
[...]

PR #703, opened 2023-08-29
[DPO] fix DPO ref_model=None

### LINKED ISSUE #699: Unable to use `ref_model=None` with DPOTrainer
Hi,

While trying to run DPO with `ref_model=None` ([PR #640](https://github.com/huggingface/trl/pull/640)), I ran into a couple of issues.

Following the Stack-Llama-2 example, `dpo_llama2.py` runs ok with all the default settings. However, this example loads the active and ref model separately and passes both to the `DPOTrainer`. 

If I understand correctly, setting `ref_model=None`, should avoid copying the model and instead disable the adapters to compute the reference model score. But it seems like [this check](https://github.com/huggingface/trl/blob/d3bbee3ab84d4d5dcb413003bf93541df5b4e54c/trl/trainer/dpo_trainer.py#L116) consistently fails and the model gets copied in any case. This also produces the following warning with no updates to the loss:

```python
 UserWarning: None of the inputs have requires_grad=True. Gradients will be None
  warnings.warn("None of the inputs have requires_grad=True. Gradients will be None")
```

To reproduce, replace https://github.com/huggingface/trl/blob/d3bbee3ab84d4d5dcb413003bf93541df5b4e54c/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py#L141C5-L146C6 with `model_ref=None`.

---

Libraries:

```
accelerate               0.21.0
bitsandbytes             0.41.1
peft                     0.5.0
torch                    2.0.1
torchvision              0.15.2
transformers             4.33.0.dev0
trl                      0.6.1.dev0
```

Any insights on this would be greatly appreciated!

kashif:
by @tannonk fixes #699  and #701

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -29,7 +29,7 @@ from .utils import DPODataCollatorWithPadding, disable_dropout_in_model, pad_to_
 
 
 if is_peft_available():
-    from peft import get_peft_model, prepare_model_for_int8_training
+    from peft import PeftModel, get_peft_model, prepare_model_for_int8_training
 
 
 class DPOTrainer(Trainer):


younesbelkada (approved):
Thank you for the quick fix @kashif !

trl/trainer/dpo_trainer.py
```diff
@@ -113,7 +113,7 @@ def __init__(
                 model = prepare_model_for_int8_training(model)
             model = get_peft_model(model, peft_config)
 
-        self.is_peft_model = getattr(model, "is_peft_model", False)
+        self.is_peft_model = getattr(model, "active_adapter", False)
```
younesbelkada, line 116: nit, this can be also:

```suggestion
        self.is_peft_model = is_peft_available() and isinstance(model, PeftModel)
```

PR #719, opened 2023-08-31
Add margin to RM training

### LINKED ISSUE #718: Add margin to reward trainer, similar to LLAMA-2
In the LLAMA-2 paper, an additional margin term is added to the loss function of the reward model. This allows to add a higher weight to pairs with a higher score difference. 

image

This should also be possible with the TRL RewardTrainer. Adding that functionality probably requires to rebuild the trainer and collator. 

Does this make sense?

jvhoffbauer:
Fix #718


jvhoffbauer:
@younesbelkada could you give me a quick hint if the interface is ok that way? 

Especially, as I solely use the margin if it is provided, but implicitly assume it is zero, if it is not provided. Is that ok? 

Also, the tests seem to be timed out. Is that something I can fix or unrelated?

younesbelkada (commented):
@jvhoffbauer , the interface looks really great to me IMO as it seems to preserve previous behaviour 
Can you give an example of how to use this interface to enable margin? Also would be nice to add few lines in the documentation about it ! What do you think?
Regarding the failing tests, don't worry the time out issues happen sometimes and it is not related to your PR

jvhoffbauer:
Awesome! I will look into this tomorrow/eow. Should I also add a test?

younesbelkada:
Great! If possible yes, a simple test that checks computing the loss with margin works as expected would be really great ! 🙏

jvhoffbauer:
Done. Let me know what you think! 

One thing: I noticed that the rewards are actually tensors with shape [1, 2] even if I process just one sample. Is that correct? 
```
batch = [[dummy_dataset[0]]]
batch = trainer.data_collator(batch)
loss = trainer.compute_loss(trainer.model, batch, return_outputs=True)

# Output: 
{
  'rewards_chosen': tensor([[ 0.2797, -0.1509]], grad_fn=<IndexBackward0>)
  'rewards_rejected': tensor([[0.0570, 0.0088]], grad_fn=<IndexBackward0>)
}
```

docs/source/reward_trainer.mdx
```diff
@@ -14,16 +14,22 @@ The [`RewardTrainer`] expects a very specific format for the dataset since the m
...
 
-After preparing your dataset, you can use the [`RewardTrainer`] in the same way as the `Trainer` class from 🤗 Transformers. 
+<<<<<<< HEAD
+After standardizing your dataset, you can use the `RewardTrainer` as a classic Hugging Face Trainer.
+You should pass an `AutoModelForSequenceClassification` model to the `RewardTrainer`.
+=======
+After preparing your dataset, you can use the [`RewardTrainer`] in the same way as the `Trainer` class from 🤗 Transformers.
 You should pass an `AutoModelForSequenceClassification` model to the [`RewardTrainer`], along with a [`RewardConfig`] which configures the hyperparameters of the training.
 
+> > > > > > > origin/main
```
younesbelkada, line 31: Hmm there is a merge conflict here that has not been properly dealt with, can you please have a look 🙏

younesbelkada (commented):
Thanks a lot, this looks great ! I left one single comment with respect to a merge conflict you might forgot to fix, apart from that LGTM !

docs/source/reward_trainer.mdx
```diff
@@ -56,10 +62,22 @@ trainer.train()
 
 ```
 
+### Adding a margin to the loss
+
+As in the LLAMA-2 paper, you can add a margin to the loss by adding a `margin` entry to the dataset. The reward collator will automatically pass it through and the loss will be computed accordingly.
```
lewtun, line 67: Small suggestion: 

```suggestion
As in the [Llama 2 paper](https://huggingface.co/papers/2307.09288), you can add a margin to the loss by adding a `margin` column to the dataset. The reward collator will automatically pass it through and the loss will be computed accordingly.
```

lewtun (commented):
Thanks a lot for adding this @jvhoffbauer - it's going to be really interesting to see if we can use this on datasets like SHP :)

docs/source/reward_trainer.mdx
jvhoffbauer: Good points. Thanks!

docs/source/reward_trainer.mdx
jvhoffbauer: Should be ready

younesbelkada (approved):
Looking great thank you very much for your great contribution !

lewtun (approved):
Thanks a lot for iterating @jvhoffbauer - this LGTM 🔥 !

Issue #480, opened 2023-06-29
Error with Multi-GPU peft Reward Training


mnoukhov:
There is an issue when you combine all four:
- `peft` quantization
- gradient checkpointing
- multi-gpu ddp
- two gradients on the same parameters (as you have in the loss function for Reward Trainer)

This is reproducible if you correctly enable gradient checkpointing in `examples/multi-adapter-rl` as shown in PR #479 and then run in a multi-gpu setup

`accelerate launch --multi_gpu reward_modeling.py --gradient_checkpointing True`

you will receive the error 
> RuntimeError: Expected to mark a variable ready only once. This error is caused by one of the following reasons: 1) Use of a module parameter outside the `forward` function. Please make sure model parameters are not shared across multiple concurrent forward-backward passes. or try to use _set_static_graph() as a workaround if this module graph does not change during training loop.2) Reused parameters in multiple reentrant backward passes. For example, if you use multiple `checkpoint` functions to wrap the same part of your model, it would result in the same set of parameters been used by different reentrant backward passes multiple times, and hence marking a variable ready multiple times. DDP does not support such use cases in default. You can try to use _set_static_graph() as a workaround if your module
graph does not change over iterations. Parameter at index 127 has been marked as ready twice. This means that multiple autograd engine  hooks have fired for this particular parameter during this iteration. You can set the environment variable
TORCH_DISTRIBUTED_DEBUG to either INFO or DETAIL to print parameter names for further debugging.

With `TORCH_DISTRIBUTED_DEBUG=DETAIL`, we find the affected parameter is a LoRA parameter. It is not related to https://github.com/pytorch/pytorch/issues/60844 because `find_unused_parameters` is set to False.

This is likely a problem between `peft` and `accelerate`/`ddp` but I'm putting the issue here because it affects `RewardTrainer` and quantization + multi gpu + gradient checkpointing are a common combination

Receiling:
We have the same problem. 
With the training scripts from `peft`, multi-gpu + gradient checkpointing is working normally.
But with `RewardTrainer` from `trl`, we have the aforementioned error, i.e., "RuntimeError: Expected to mark a variable ready only once. "

younesbelkada:
Hi @mnoukhov @Receiling 
After discussing with @pacman100 , the fix should go on transformers, and should be similar to https://github.com/huggingface/transformers/pull/24247 
we should add a flag in training arguments and make the default behavior use `use_reentrant=True` for backward compatiblity

[...]

PR #725, opened 2023-09-01
Enable gradient checkpointing to be disabled for reward modelling

lewtun:
Edit: to be merged after #726 

This PR allows users to disable gradient checkpointing for PEFT models in the `RewardTrainer`. The motivation for this is to bypass the `RuntimeError: Expected to mark a variable ready only once.` issue noted in https://github.com/huggingface/trl/issues/480 at the expense of using more vRAM. More generally, I think it's good when a library allows users to configure settings like these, since there are tradeoffs involved (i.e. computation time for memory).

With this change, it is now possible to train PEFT reward models in multi-GPU contexts as follows:

```shell
# The example script didn't work when num_gpus > 1
accelerate launch --multi_gpu --num_processes 4 examples/scripts/reward_trainer.py --use_peft --load_in_8bit --gradient_checkpointing false
```

When gradient checkpointing is activated the script still fails due to https://github.com/huggingface/trl/issues/480 (which requires a deeper-level fix):

```shell
# Current behaviour is the same
accelerate launch --multi_gpu --num_processes 4 examples/scripts/reward_trainer.py --use_peft --load_in_8bit --gradient_checkpointing true
```

```
  File "/fsx/lewis/git/trl/examples/scripts/reward_trainer.py", line 177, in <module>
    trainer.train()
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/transformers/trainer.py", line 1539, in train
    return inner_training_loop(
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/transformers/trainer.py", line 1809, in _inner_training_loop
    tr_loss_step = self.training_step(model, inputs)
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/transformers/trainer.py", line 2665, in training_step
    self.accelerator.backward(loss)
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/accelerate/accelerator.py", line 1923, in backward
    loss.backward(**kwargs)
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/torch/_tensor.py", line 487, in backward
    torch.autograd.backward(
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/torch/autograd/__init__.py", line 200, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/torch/autograd/function.py", line 274, in apply
    return user_fn(self, *args)
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/torch/utils/checkpoint.py", line 157, in backward
    torch.autograd.backward(outputs_with_grad, args_with_grad)
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/torch/autograd/__init__.py", line 200, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: Expected to mark a variable ready only once. This error is caused by one of the following reasons: 1) Use of a module parameter outside the `forward` function. Please make sure model parameters are not shared across multiple concurrent forward-backward passes. or try to use _set_static_graph() as a workaround if this module graph does not change during training loop.2) Reused parameters in multiple reentrant backward passes. For example, if you use multiple `checkpoint` functions to wrap the same part of your model, it would result in the same set of parameters been used by different reentrant backward passes multiple times, and hence marking a variable ready multiple times. DDP does not support such use cases in default. You can try to use _set_static_graph() as a workaround if your module graph does not change over iterations.
Parameter at index 93 has been marked as ready twice. This means that multiple autograd engine  hooks have fired for this particular parameter during this iteration. You can set the environment variable TORCH_DISTRIBUTED_DEBUG to either INFO or DETAIL to print parameter names for further debugging.
```

**Question:** this change overrides the default value of the `RewardTrainer` from `true` to `false` which may lead to unexpected behaviour with user code. Should I add a warning in the logger?

Update: I think I can solve the question above with my proposal to have dedicated training arguments per trainer. I'll open a separate PR for that.

### DIFF
--- a/docs/source/reward_trainer.mdx
+++ b/docs/source/reward_trainer.mdx
@@ -6,29 +6,27 @@ Check out a complete flexible example inside [`examples/scripts`](https://github
 
 ## Expected dataset format
 
-The reward trainer expects a very specific format for the dataset. Since the model will be trained to predict which sentence is the most relevant, given two sentences. We provide an example from the [`Anthropic/hh-rlhf`](https://huggingface.co/datasets/Anthropic/hh-rlhf) dataset below:
+The [`RewardTrainer`] expects a very specific format for the dataset since the model will be trained on pairs of examples to predict which of the two is preferred. We provide an example from the [`Anthropic/hh-rlhf`](https://huggingface.co/datasets/Anthropic/hh-rlhf) dataset below:
 
 <div style="text-align: center">
 <img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/rlhf-antropic-example.png", width="50%">
 </div>
 
-Therefore the final dataset object should contain two 4 entries at least if you use the default `RewardDataCollatorWithPadding` data collator. The entries should be named:
+Therefore the final dataset object should contain two 4 entries at least if you use the default [`RewardDataCollatorWithPadding`] data collator. The entries should be named:
 
 - `input_ids_chosen`
 - `attention_mask_chosen`
 - `input_ids_rejected`
 - `attention_mask_rejected`
 
-The `j` and `k` suffixes are used to denote the two sentences in the paired dataset.
-
 ## Using the `RewardTrainer`
 
-After standardizing your dataset, you can use the `RewardTrainer` as a classic Hugging Face Trainer. 
-You should pass an `AutoModelForSequenceClassification` model to the `RewardTrainer`.
+After preparing your dataset, you can use the [`RewardTrainer`] in the same way as the `Trainer` class from 🤗 Transformers. 
+You should pass an `AutoModelForSequenceClassification` model to the [`RewardTrainer`], along with a [`RewardConfig`] which configures the hyperparameters of the training.
 
-### Leveraging the `peft` library to train a reward model
+### Leveraging 🤗 PEFT to train a reward model
 
-Just pass a `peft_config` in the key word arguments of `RewardTrainer`, and the trainer should automatically take care of converting the model into a PEFT model!
+Just pass a `peft_config` in the keyword arguments of [`RewardTrainer`], and the trainer should automatically take care of converting the model into a PEFT model!
 
 ```python
 from peft import LoraConfig, task_type
--- a/examples/scripts/reward_trainer.py
+++ b/examples/scripts/reward_trainer.py
@@ -34,7 +34,7 @@ class ScriptArguments:
     """
 
     model_name: Optional[str] = field(default="facebook/opt-350m", metadata={"help": "the model name"})
-    dataset_name: Optional[str] = field(default="Anthropic/hh-rlhf", metadata={"help": "the model name"})
+    dataset_name: Optional[str] = field(default="Anthropic/hh-rlhf", metadata={"help": "the dataset name"})
     dataset_text_field: Optional[str] = field(default="text", metadata={"help": "the text field of the dataset"})
     log_with: Optional[str] = field(default=None, metadata={"help": "use 'wandb' to log with wandb"})
     logging_steps: Optional[int] = field(default=500, metadata={"help": "the number of update steps between two logs"})
@@ -115,7 +115,6 @@ train_dataset = train_dataset.map(
 train_dataset = train_dataset.filter(
     lambda x: len(x["input_ids_chosen"]) <= script_args.seq_length
     and len(x["input_ids_rejected"]) <= script_args.seq_length,
-    num_proc=4,
 )
 
 if script_args.eval_split == "none":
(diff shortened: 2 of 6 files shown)


examples/scripts/sentiment_tuning.py
```diff
@@ -149,7 +150,8 @@ def collator(data):
         task_type="CAUSAL_LM",
     )
     ref_model = None
-    device_map = {"": 0}
+    # Copy the model to each device
+    device_map = {"": Accelerator().local_process_index}
```
lewtun, line 154: I noticed none of the example scripts actually work in multi-GPU settings unless we add this trick

examples/scripts/reward_trainer.py
```diff
@@ -103,26 +102,27 @@ def preprocess_function(examples):
...
 train_dataset = train_dataset.map(
     preprocess_function,
     batched=True,
     num_proc=4,
 )
 train_dataset = train_dataset.filter(
     lambda x: len(x["input_ids_chosen"]) <= script_args.seq_length
-    and len(x["input_ids_rejected"]) <= script_args.seq_length
+    and len(x["input_ids_rejected"]) <= script_args.seq_length,
+    num_proc=4,
```
lvwerra, line 125: why do you change to four?

examples/scripts/reward_trainer.py
```diff
@@ -32,11 +33,10 @@
 tqdm.pandas()
 
 
-# Define and parse arguments.
 @dataclass
 class ScriptArguments:
     """
-    The name of the Casual LM model we wish to fine with RewardTrainer
+    Hyperparameters to fine-tune a reward model on a given dataset.
```
lvwerra, line 39:
```suggestion
    Hyperparameters to fine-tune a reward model on a given dataset with the `RewardTrainer`.
```

lvwerra (approved):
Thanks for adding this. Just a few small nits, looks generally good to me.

examples/scripts/reward_trainer.py
lewtun: Because the HHH dataset is quite large and I was bored of waiting :) Will revert it back to 1

examples/scripts/sentiment_tuning.py
younesbelkada: Makes sense!

examples/scripts/reward_trainer.py
```diff
@@ -63,6 +64,7 @@ class ScriptArguments:
 
 parser = HfArgumentParser(ScriptArguments)
 script_args = parser.parse_args_into_dataclasses()[0]
+print(script_args)
```
younesbelkada, line 67:
```suggestion
```

younesbelkada (approved):
Thanks a lot for workaround and for adding this! LGTM!

examples/scripts/reward_trainer.py
lewtun: Haha nice catch, `print()` is the best debugger :D

docs/source/reward_trainer.mdx
```diff
@@ -6,29 +6,27 @@ Check out a complete flexible example inside [`examples/scripts`](https://github
 
 ## Expected dataset format
 
-The reward trainer expects a very specific format for the dataset. Since the model will be trained to predict which sentence is the most relevant, given two sentences. We provide an example from the [`Anthropic/hh-rlhf`](https://huggingface.co/datasets/Anthropic/hh-rlhf) dataset below:
+The [`RewardTrainer`] expects a very specific format for the dataset since the model will be trained on pairs of examples to predict which of the two is preferred. We provide an example from the [`Anthropic/hh-rlhf`](https://huggingface.co/datasets/Anthropic/hh-rlhf) dataset below:
```
lewtun, line 9: @lvwerra @younesbelkada FYI I made a few minor tweaks to the docs since your last review

trl/trainer/training_configs.py
```diff
@@ -39,3 +41,9 @@ class RewardConfig(TrainingArguments):
             "help": "The maximum length of the sequences in the batch. This argument is required if you want to use the default data collator."
         },
     )
+    gradient_checkpointing: Optional[bool] = field(
+        default=True,
```
lewtun, line 45: Here we override the default value of `transformers.TrainingArguments` to be `True` for backwards compatibility with how PEFT prepares k-bit models

younesbelkada (approved):
Thanks a lot @lewtun !!

PR #726, opened 2023-09-01
Refactor RewardTrainer hyperparameters into dedicated dataclass

lewtun:
This PR migrates the `max_length` arg of the `RewardTrainer` into a dedicated `RewardTrainingArguments` class that can also be used for storing future hyperparameters.

To be backwards compatible, I've left the variable in the trainer's init, with a warning that this will be removed in some future version.

Tested with:

```shell
accelerate launch --multi_gpu --num_processes 2 examples/scripts/reward_trainer.py --batch_size 1
```

### DIFF
--- a/trl/trainer/reward_trainer.py
+++ b/trl/trainer/reward_trainer.py
@@ -18,12 +18,13 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 import torch
 import torch.nn as nn
 from datasets import Dataset
-from transformers import DataCollator, PreTrainedModel, PreTrainedTokenizerBase, Trainer, TrainingArguments
+from transformers import DataCollator, PreTrainedModel, PreTrainedTokenizerBase, Trainer
 from transformers.trainer_callback import TrainerCallback
 from transformers.trainer_pt_utils import nested_detach
 from transformers.trainer_utils import EvalPrediction
 
 from ..import_utils import is_peft_available
+from .training_args import RewardTrainingArguments
 from .utils import PeftSavingCallback, RewardDataCollatorWithPadding, compute_accuracy
 
 
@@ -51,7 +52,7 @@ class RewardTrainer(Trainer):
     def __init__(
         self,
         model: Union[PreTrainedModel, nn.Module] = None,
-        args: TrainingArguments = None,
+        args: RewardTrainingArguments = None,
         data_collator: Optional[DataCollator] = None,
         train_dataset: Optional[Dataset] = None,
         eval_dataset: Optional[Union[Dataset, Dict[str, Dataset]]] = None,
@@ -73,7 +74,7 @@ class RewardTrainer(Trainer):
         Args:
             model (`transformers.PreTrainedModel`):
                 The model to train, preferably an `AutoModelForSequenceClassification`.
-            args (`transformers.TrainingArguments`):
+            args (`RewardTrainingArguments`):
                 The arguments to use for training.
             data_collator (`transformers.DataCollator`):
                 The data collator to use for training. If None is specified, the default data collator (`RewardDataCollatorWithPadding`) will be used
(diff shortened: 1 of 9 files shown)


trl/trainer/reward_trainer.py
```diff
@@ -94,18 +95,21 @@ def __init__(
                 The optimizer and scheduler to use for training.
             preprocess_logits_for_metrics (`Callable[[torch.Tensor, torch.Tensor], torch.Tensor]`):
                 The function to use to preprocess the logits before computing the metrics.
-            max_length (`int`, defaults to `None`):
```
lewtun, line 97: I followed the `transformers` convention to remove deprecated args from public doc strings

trl/trainer/reward_trainer.py
```diff
@@ -94,11 +95,14 @@ def __init__(
...
-                The maximum length of the sequences in the batch. This argument is required if you want to use the default data collator.
             peft_config (`Dict`, defaults to `None`):
                 The PEFT configuration to use for training. If you pass a PEFT configuration, the model will be wrapped in a PEFT model.
         """
+        if max_length is not None:
+            warnings.warn(
+                "The `max_length` argument is deprecated and will be removed in a future version. Please use the `RewardTrainingArguments` to set `max_length` instead.",
```
lewtun, line 103: Should we specify a precise version, e.g. 0.9.0?

trl/trainer/training_args.py
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from dataclasses import dataclass, field
+from typing import Optional
+
+from transformers import TrainingArguments
+
+
+@dataclass
+class RewardTrainingArguments(TrainingArguments):
```
lewtun, line 23: Another name for this could be `RewardConfig` to be more aligned with `PPOConfig`

lewtun: I opted for `RewardConfig` in https://github.com/huggingface/trl/pull/726/commits/e6e346d321654de7a689d44a05ef51d9e2005a34

I've also placed this in a `training_configs.py` module - let me know if you'd prefer a dedicated module per config (currently `PPOConfig` lives in `ppo_config.py`)

lewtun:
I think this is good to go, so gently pinging @lvwerra @younesbelkada for a final review 🙏

younesbelkada (approved):
Thanks a lot @lewtun ! This looks great to me, I just have one suggestion about the name of the class, what `RewardTrainingArguments` ? I am also happy with the current naming if @lvwerra agrees

trl/trainer/reward_trainer.py
```diff
@@ -122,11 +126,20 @@ def __init__(
...
                 max_length = 512
+            elif args.max_length is None:
+                warnings.warn(
+                    "When using RewardDataCollatorWithPadding, you should set `max_length` in RewardConfig."
+                    " It will be set to `512` by default, but you should do it yourself in the future.",
+                    UserWarning,
+                )
+                max_length = 512
+            else:
+                max_length = args.max_length
```
lvwerra, line 142: I think in the case where `max_length` is not `None` then it is still overwritten when `args.max_length is None`, right? I think in that case we should keep the original value to be backwards compatible.

lewtun: Oh yes, good catch - I'll fix that

lewtun:
> Thanks a lot @lewtun ! This looks great to me, I just have one suggestion about the name of the class, what `RewardTrainingArguments` ? I am also happy with the current naming if @lvwerra agrees

Are you suggesting we use `RewardTrainingArguments` instead of `RewardConfig`? I originally had the former, but switched to the latter to be aligned with `PPOConfig`. I don't have a strong opinion, but whatever choice we make should stay consistent for all the other trainers, e.g. `SFTTrainingArguments` vs `SFTConfig` etc.

One advantage of `XConfig` classes is that it's less keystrokes :D

trl/trainer/reward_trainer.py
lewtun: Fixed in https://github.com/huggingface/trl/pull/726/commits/5b67e0d40d96811b3202ea5bf11628698c44b49b

lvwerra (approved):
LGTM! 🚀

younesbelkada (approved):
Thanks @lewtun , makes sense, let's merge it! 🚀

PR #727, opened 2023-09-01
Add epsilon to score normalization

zfang:
## Summary
- Add epsilon to score normalization for numerical stability;
- Call `Tensor.to` on running mean and std

## Test
Running
```commandline
python examples/scripts/sentiment_tuning.py --log_with wandb --use_score_scaling --use_score_norm --score_clip 0.5
```
Screenshot 2023-09-01 at 9 20 03 PM


lvwerra (approved):
Thanks a lot for the PR! Looks good to me, I'll let @younesbelkada also have a look!

younesbelkada (approved):
Very nice ! Thanks !

PR #737, opened 2023-09-04
Introducing the Iterative Trainer

gaetanlop:
This PR is a follow-up to a requested Iterative Trainer in #704 and #576. It introduces a way to finetune models with methods that require some steps between optimizations.


gaetanlop:
@younesbelkada  @lvwerra what do you think of the structure of the trainer? Should we add a generate function or should the generation step be done outside of the trainer?

gaetanlop:
Hello @younesbelkada @lvwerra, any news on this? I think that's a required step before adding an example for #704

trl/trainer/iterative_trainer.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+class IterativeTrainer:
+    """
+    The IterativeTrainer can be used to finetune models with methods that requires some steps between optimization.
+
+    Attributes:
+        **config** (`IterativeConfig`) -- Configuration object for IterativeTrainer.
+        **model** (`PreTrainedModel`) -- Model to be optimized, Hugging Face transformer model with a causal language modeling head.
+            Check the documentation of `PreTrainedModelWrapper` for more details.
```
younesbelkada, line 35:
```suggestion
```

younesbelkada: Here we're using a `PreTrainedModel` right?

trl/trainer/iterative_trainer.py
```diff
@@ -0,0 +1,232 @@
...
+        set_seed(config.seed)
+
+        # Step 0: check positional arguments validity
+        if not isinstance(config, IterativeConfig):
+            raise ValueError(f"config must be a IterativeConfig, got {type(config)}")
+        if not isinstance(tokenizer, (PreTrainedTokenizerBase)):
+            raise ValueError(
+                f"tokenizer must be a PreTrainedTokenizerBase like a PreTrainedTokenizer or a PreTrainedTokenizerFast, got {type(tokenizer)}"
+            )
+
```
younesbelkada, line 82: You can also check here is the model is an instance of `PreTrainedModel`

trl/trainer/iterative_trainer.py
```diff
@@ -0,0 +1,232 @@
...
+            self.current_device = self.accelerator.device
+        else:
+            self.current_device = torch.device("cuda:0")
+
+        PPODecorators.optimize_cuda_cache = self.config.optimize_cuda_cache
+
+        self.is_encoder_decoder = hasattr(self.model, "is_encoder_decoder")
+
+    def prepare_model_inputs(self, input_ids: torch.Tensor, attention_mask: torch.Tensor, labels: torch.Tensor):
+        if self.is_encoder_decoder:
```
younesbelkada, line 139: Can it be an encoder decoder? from the docstring above it seems on decoder-based models are supported

trl/trainer/iterative_trainer.py
```diff
@@ -0,0 +1,232 @@
...
+
+        # Step 0: check positional arguments validity
+        if not isinstance(config, IterativeConfig):
+            raise ValueError(f"config must be a IterativeConfig, got {type(config)}")
+        if not isinstance(tokenizer, (PreTrainedTokenizerBase)):
+            raise ValueError(
+                f"tokenizer must be a PreTrainedTokenizerBase like a PreTrainedTokenizer or a PreTrainedTokenizerFast, got {type(tokenizer)}"
+            )
+
+        # Step 1: Initialize Accelerator
```
younesbelkada, line 83: You can also add a stronger check, one can check if `model.can_generate()` and pass a warning if that method returns `False`

younesbelkada (commented):
Thanks a lot for your great work ! 
Design wise the PR looks great, I just left minor questions, I expect this API to be public, therefore we need to add some documentation, would you be happy doing so? I can also help you if any !
Thanks!

gaetanlop:
Hello @younesbelkada thanks for your feedback, I have added the requested modifications to handle seq2seq models and added some documentation. Also, I have added:

- Made the attention_mask and the labels optional in the step function (labels should only be needed for encoder-decoder models).
- a safety checker method to verify that input_ids, attention_mask and labels are list of tensors.
- a logging method to be able to log metrics during training.

It should be ready to be merged.

trl/trainer/iterative_trainer.py
```diff
@@ -0,0 +1,333 @@
...
+        set_seed(config.seed)
+
+        # Step 0: check positional arguments validity
+        if not isinstance(config, IterativeConfig):
+            raise ValueError(f"config must be a IterativeConfig, got {type(config)}")
+        if not isinstance(tokenizer, (PreTrainedTokenizerBase)):
+            raise ValueError(
+                f"tokenizer must be a PreTrainedTokenizerBase like a PreTrainedTokenizer or a PreTrainedTokenizerFast, got {type(tokenizer)}"
+            )
+        if not isinstance(model, (PreTrainedModel)):
```
younesbelkada, line 89:
```suggestion
        if not isinstance(model, PreTrainedModel):
```

tests/test_iterative_trainer.py
```diff
@@ -0,0 +1,60 @@
...
+        return dummy_dataset
+
+    def setUp(self):
+        # initialize trainer
+        self.iterative_config = IterativeConfig(step_batch_size=2, log_with=None)
+        self.model.train()
+        self.data_collator = DataCollatorForLanguageModeling(self.tokenizer, mlm=False)
+        return super().setUp()
+
+    def test_iterative_step(self):
```
younesbelkada, line 49: Can you also add a test with a seq2seq model to make sure this feature will not be broken by future PRs? 🙏

younesbelkada (commented):
Thanks a lot for your great work @gaetanlop !!
I left three comments, otherwise looks really great ! Looking forward to merge this PR !

gaetanlop:
Hey @younesbelkada, thanks for your feedback. I made the requested changes.

gaetanlop:
Hi @younesbelkada, still interested by this trainer? I don't think this will be useful for #704 as the SFTTrainer will do the job for the training phase, but that would be useful for rejection sampling and for this paper from google deepmind (https://arxiv.org/pdf/2306.13649.pdf) which seems to be the new state of the art for LLMs knowledge distillation. It uses on policy samples for each optimization steps.

trl/trainer/iterative_trainer.py
```diff
@@ -0,0 +1,333 @@
...
+                    raise ValueError(f"Elements in {name} must be tensors - got {type(tensor_list[0])}")
+
+        return input_ids, attention_mask, labels
+
+    @PPODecorators.empty_cuda_cache()
+    def step(
+        self,
+        input_ids: List[torch.LongTensor],
+        attention_mask: Optional[List[torch.LongTensor]],
+        labels: Optional[List[torch.LongTensor]],
```
lvwerra, line 239: There is some design decision we have to take here. The `SFTTrainer` takes texts as input whereas the `PPOTrainer.step` uses tokenized inputs. I tend think that texts would be more natural but a compromise could be that we have an additional `texts` kwarg and make all args kwargs. Then the user can optionally pass tokenized or text inputs. What do you think?

trl/trainer/iterative_trainer.py
```diff
@@ -0,0 +1,333 @@
...
+            data. Check the documentation of `transformers.PreTrainedTokenizer` and
+            `transformers.PreTrainedTokenizerFast` for more details.
+        **optimizer** (`torch.optim.Optimizer`, *optional*) -- Optimizer to be used for training. If no optimizer is
+            provided, the trainer will create an Adam optimizer with the learning rate specified in the configuration
+            object.
+        **data_collator** (Union[DataCollatorForLanguageModeling, DataCollatorForSeq2Seq], *optional*) -- Data collator to be used for training and
+            passed along the dataloader.
+    """
+
+    def __init__(
```
lvwerra, line 53: What about passing a learning rate schedule?

trl/trainer/iterative_trainer.py
```diff
@@ -0,0 +1,333 @@
...
+    DataCollatorForSeq2Seq,
+    PreTrainedModel,
+    PreTrainedTokenizerBase,
+)
+
+from ..core import PPODecorators, set_seed
+from . import IterativeConfig
+
+
+class IterativeTrainer:
```
lvwerra, line 35: What do you think about calling it `IterativeSFTTrainer`?

lvwerra (commented):
Hi @gaetanlop, thanks for working on this! Generally I think it is in a good shape, just a few comments on some design decisions.

lvwerra:
Hi @gaetanlop the PR is in a good shape, do you have bandwidth to work on the final details?

gaetanlop:
Hey @lvwerra sorry have been very busy with work the last 4 days. Thanks for your review, I will add the final details this weekend.

gaetanlop:
Hi @lvwerra, I have added the requested changes.

gaetanlop:
Hi @lvwerra, what do you think of the changes? Do not hesitate if you require other modifications before merging.

trl/trainer/iterative_sft_config.py
```diff
@@ -0,0 +1,79 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from dataclasses import dataclass, field
+from typing import Optional
+
+from ..core import flatten_dict
+
+
+@dataclass
+class IterativeSFTConfig(object):
```
lvwerra, line 21: note that we ported the CLI args to `tyro` since you opened the PR. not a big change, you can for example look at https://github.com/huggingface/trl/blob/main/trl/trainer/ppo_config.py

trl/trainer/iterative_sft_trainer.py
```diff
@@ -0,0 +1,393 @@
...
+            if texts_labels is not None:
+                if not isinstance(texts_labels, list):
+                    raise ValueError(f"'text_labels' must be a list of strings - got {type(texts_labels)}")
+                if not isinstance(texts_labels[0], str):
+                    raise ValueError(f"Elements in 'text_labels' must be strings - got {type(texts_labels[0])}")
+
+        return input_ids, attention_mask, labels, texts, texts_labels
+
+    @PPODecorators.empty_cuda_cache()
+    def step(self, **kwargs):
```
lvwerra, line 273: Why are you using `kwargs` here instead of listing the kwargs? it makes the function more prone to unexpected behaviour (e.g. if a kwarg is misspelled).

trl/trainer/iterative_sft_trainer.py
```diff
@@ -0,0 +1,393 @@
...
+            raise ValueError(
+                "No 'labels' or 'text_labels' are provided. When using an encoder-decoder architecture, 'labels' or 'text_labels' must be passed."
+            )
+
+        input_ids, attention_mask, labels, texts, texts_labels = self._step_safety_checker(
+            input_ids, attention_mask, labels, texts, texts_labels
+        )
+
+        if texts is not None:
+            input_ids = [self.tokenizer(text, return_tensors="pt")["input_ids"][0] for text in texts]
```
lvwerra, line 316: this will be not make use of multiple CPU cores. what about:

```python
input_ids = self.tokenizer(texts, return_tensors="pt").input_ids
```

trl/trainer/iterative_sft_trainer.py
```diff
@@ -0,0 +1,393 @@
...
+        if texts_labels is not None:
+            labels = [self.tokenizer(text_labels, return_tensors="pt")["input_ids"][0] for text_labels in texts_labels]
```
lvwerra, line 318: same as above.

trl/trainer/iterative_sft_trainer.py
```diff
@@ -0,0 +1,393 @@
...
+    PreTrainedModel,
+    PreTrainedTokenizerBase,
+)
+
+from ..core import PPODecorators, set_seed
+from ..import_utils import is_torch_greater_2_0
+from . import IterativeSFTConfig
+
+
+class IterativeSFTTrainer:
```
lvwerra, line 36: I wonder if we should inherit from `transformers.Trainer` which would give us a lot of init logic as well as e.g. saving models or evaluating for free. 

As discussed with @lewtun offline, maybe we should inherit from the `Trainer` everywhere to have the same base logic for all trainers

trl/trainer/iterative_sft_trainer.py
```diff
@@ -0,0 +1,393 @@
...
+            labels = input_ids
+
+        model_inputs = self.prepare_model_inputs(input_ids, attention_mask, labels)
+
+        model_inputs_names = list(model_inputs.keys())
+
+        batch_dict = {}
+        batch_dict.update(model_inputs)
+
+        def collator(data):
```
lvwerra, line 331: I think we are missing a bit of preprocessing here. 

- truncation of long docs
- padding if not all sequences have the same length

We could use the [`DataCollatorForLanguageModeling`](https://huggingface.co/docs/transformers/main_classes/data_collator#transformers.DataCollatorForLanguageModeling) to implement some of that logic. Maybe we need to pass some additional kwargs to the `step`

lvwerra (commented):
Hi @gaetanlop, sorry for the delay, people working on TRL have been a bit busy with other projects the past few weeks. We plan to do a release this week and adding this PR would be nice (if you have time of course). 

My main two points:
- should we inherit from `Trainer` to get a lot of upstream functionality for free
- there is some preprocessing missing I believe making sure we pad/truncate sequence

Let me know what you think. cc @younesbelkada

trl/trainer/iterative_sft_trainer.py
```diff
@@ -0,0 +1,393 @@
...
+        self.accelerator.init_trackers(
+            config.tracker_project_name,
+            config=dict(trl_ppo_trainer_config=config.to_dict()) if not is_using_tensorboard else config.to_dict(),
+            init_kwargs=config.tracker_kwargs,
+        )
+
+        self.model = model
+        self.model_params = filter(lambda p: p.requires_grad, self.model.parameters())
+        self.is_encoder_decoder = hasattr(self.model, "is_encoder_decoder")
+        self.is_peft_model = getattr(self.model, "is_peft_model", False)
```
younesbelkada, line 121:
```suggestion
        self.is_peft_model = is_peft_available() and isinstance(self.model, PeftModel)
```
And make sure to import peftModel if peft is available on the top of the file

younesbelkada, line 121: Since here we rely on `PreTrainedModel` that's not the way to check if the model is a peft model.

trl/trainer/iterative_sft_trainer.py
younesbelkada: Yes, let's list all supported kwargs here

younesbelkada (commented):
Thanks a lot for your great work ! Agreed with all the points shared by @lvwerra 

My comments are similar than @lvwerra 's comments:

1- Let's inherit from `transformers.Trainer` to benefit from all features from the trainer, in the current implementation saving / pushing to hub mechanism are missing for instance

2- let's properly tokenize a batch of sequences instead of tokenizing one by one

3- Let's use explcit arguments in step to make sure we avoid unexpected behaviours

4- We should use `DataCollatorForLanguageModeling` to properly handle padding 

5- We can't take `model.is_peft_model` as model is a `PreTrainedModel` , I proposed an alternative to check if the model is a peft model.

Thanks!

trl/trainer/iterative_sft_trainer.py
gaetanlop: Padding is already done inside `prepare_model_inputs`. I have added truncation

trl/trainer/iterative_sft_config.py
gaetanlop: I have removed the `IterativeSFTConfig` as it was not useful anymore as we inherit from `Trainer`

trl/trainer/iterative_sft_trainer.py
gaetanlop: Indeed, thanks

gaetanlop:
Hi @younesbelkada @lvwerra, thanks for the review. I have made the changes.

The padding part was already done inside the `prepare_model_inputs` function. However, truncation was not set. I have added two keyword arguments to the init (truncation mode and max_length). If the user provides texts instead of input ids then the truncation is directly done when tokenizing. In any case, we also truncate inside the `prepare_model_inputs` function in case the user didn't truncate its model_inputs before passing it to the `step` function.

I have removed the `IterativeSFTConfig` as we can directly use the `TrainingArguments` if we inherit from the HF Trainer.

younesbelkada (commented):
Looking great, thanks! I left one question, I think that `is_encoder_decoder` has 0 effect as it is an attribute of `PreTrainedModelWrapper`: https://github.com/huggingface/trl/blob/main/trl/models/modeling_value_head.py#L286 
Can you try to remove it and see if the tests pass? Or alternatively try the alternative I suggested

trl/trainer/iterative_sft_trainer.py
```diff
@@ -0,0 +1,335 @@
...
+            warnings.warn(
+                f"The current model class {type(model)} is not compatible with `.generate()`"
+                "Please make sure that this is intended."
+            )
+        if optimizers[1] is None and args.max_steps == -1:
+            raise ValueError(
+                "When no scheduler is provided, you need to set the total number of training steps to perform `max_steps`"
+            )
+
+        self.is_encoder_decoder = hasattr(model, "is_encoder_decoder")
```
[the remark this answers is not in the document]
gaetanlop: Indeed, that's an error

gaetanlop:
@younesbelkada Looks like tests do not start

trl/trainer/iterative_sft_trainer.py
```diff
@@ -0,0 +1,335 @@
...
+        **model** (`PreTrainedModel`) -- Model to be optimized, either an 'AutoModelForCausalLM' or an 'AutoModelForSeq2SeqLM'.
+            Check the documentation of `PreTrainedModel` for more details.
+        **args** (`transformers.TrainingArguments`): -- The arguments to use for training.
+        **tokenizer** (`PreTrainedTokenizerBase`) -- Tokenizer to be used for encoding the
+            data. Check the documentation of `transformers.PreTrainedTokenizer` and
+            `transformers.PreTrainedTokenizerFast` for more details.
+        **optimizers** (`Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR]`): -- The optimizer and scheduler to use for training.
+        **data_collator** (Union[DataCollatorForLanguageModeling, DataCollatorForSeq2Seq], *optional*) -- Data collator to be used for training and
+            passed along the dataloader.
+        **eval_dataset** (`datasets.Dataset`): The dataset to use for evaluation.
```
lvwerra, line 53: Currently, we never evaluate, right?

trl/trainer/iterative_sft_trainer.py
```diff
@@ -0,0 +1,335 @@
...
+            data_collator=data_collator,
+            eval_dataset=eval_dataset,
+            tokenizer=tokenizer,
+            compute_metrics=compute_metrics,
+            optimizers=optimizers,
+            preprocess_logits_for_metrics=preprocess_logits_for_metrics,
+        )
+
+        self.optimizer, self.lr_scheduler = optimizers
+
```
lvwerra, line 129: isnt' that redundant? the parent class should set the optimizer/scheduler, no?

lvwerra (commented):
Almost go to merge I would say! One question remaining regarding evaluation. I think we can probably just keep track of how many optimization steps have been run and call the `Trainer` evaluation when it's time.

trl/trainer/iterative_sft_trainer.py
gaetanlop: Indeed, thanks for catching this, I have removed and made the necessary changes

trl/trainer/iterative_sft_trainer.py
gaetanlop: Indeed, the user had to call the evaluate function of the iterative trainer. I have made some changes

gaetanlop:
I made some changes to enable simple logging and evaluation. The function `_maybe_log_save_evaluate` of the **HF Trainer** could not be used in our case so I made a simple version of the function that has the following behavior: 
- We log and evaluate only if the user has specified **logging_steps** and **eval_steps** respectively (we do not have access to the number of epochs so step args are the only thing that matters).
- Logging and evaluation are done every logging_steps and eval_steps respectively.
- We only keep track of the loss and the learning rate.

lvwerra (approved):
Thanks! LGTM! 🚀

younesbelkada (approved):
Thanks a lot for your great work! Let's 🚢 it !

PR #745, opened 2023-09-07
[DPO] fix ref_model

i4never:
https://github.com/huggingface/accelerate/blob/cec5ae8e4d078186e1d13413a08857ee56fee355/src/accelerate/accelerator.py#L1675
`_prepare_deepspeed` return tuple


kashif:
nice catch @i4never LGTM!

lvwerra (approved):

PR #748, opened 2023-09-08
Ensure `RewardConfig` is backwards compatible

lewtun:
In #726 we replaced the `RewardTrainer.args` with a dedicated `RewardConfig` to collect all hyperparameters in a dedicated class (in particular `max_length`).

Unfortunately, that implementation wasn't backwards compatible because `RewardTrainer.args` was of type `transformers.TrainingArguments` and that object doesn't have the `max_length` attribute. This meant that part of the logic which checks for the presence of `args.max_length` would throw an error for any users who hadn't updated their code to use `RewardConfig`.

This PR fixes that by first checking the type of `RewardTrainer.args` and raising a warning if it's `transformers.TrainingArguments`.

Happy to add a unit test for this, but I felt it was a bit of overkill.

### DIFF
--- a/trl/trainer/reward_trainer.py
+++ b/trl/trainer/reward_trainer.py
@@ -18,7 +18,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 import torch
 import torch.nn as nn
 from datasets import Dataset
-from transformers import DataCollator, PreTrainedModel, PreTrainedTokenizerBase, Trainer
+from transformers import DataCollator, PreTrainedModel, PreTrainedTokenizerBase, Trainer, TrainingArguments
 from transformers.trainer_callback import TrainerCallback
 from transformers.trainer_pt_utils import nested_detach
 from transformers.trainer_utils import EvalPrediction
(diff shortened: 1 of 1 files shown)


trl/trainer/reward_trainer.py
```diff
@@ -128,15 +139,24 @@ def __init__(
...
-            if max_length is None and args.max_length is not None:
-                max_length = args.max_length
+            if type(args) == TrainingArguments:
+                if max_length is None:
+                    warnings.warn(
+                        "When using RewardDataCollatorWithPadding, you should set `max_length` in RewardConfig."
+                        " It will be set to `512` by default, but you should do it yourself in the future.",
+                        UserWarning,
+                    )
+                    max_length = 512
```
lewtun, line 149: Side remark for a later PR, but I think this should ideally be set to be the model's max context size as the default

trl/trainer/reward_trainer.py
```diff
@@ -98,15 +98,26 @@ def __init__(
             peft_config (`Dict`, defaults to `None`):
                 The PEFT configuration to use for training. If you pass a PEFT configuration, the model will be wrapped in a PEFT model.
         """
-        if max_length is not None and args.max_length is not None:
-            raise ValueError(
-                "You cannot specify both `max_length` and `args.max_length`. Please use the `RewardConfig` to set `max_length` once."
-            )
-        if max_length is not None and args.max_length is None:
+        if type(args) == TrainingArguments:
```
younesbelkada, line 101:
```suggestion
        if isinstance(args, TrainingArguments):
```

trl/trainer/reward_trainer.py
```diff
@@ -128,15 +139,24 @@ def __init__(
...
-            if max_length is None and args.max_length is None:
-                warnings.warn(
-                    "When using RewardDataCollatorWithPadding, you should set `max_length` in RewardConfig."
-                    " It will be set to `512` by default, but you should do it yourself in the future.",
-                    UserWarning,
-                )
-                max_length = 512
-            if max_length is None and args.max_length is not None:
-                max_length = args.max_length
+            if type(args) == TrainingArguments:
```
younesbelkada, line 142:
```suggestion
            if isinstance(args, TrainingArguments):
```

younesbelkada (approved):
Thanks a lot ! I left two comments, otherwise looking great ! Thanks for taking care of the backward compatibility lewis!

trl/trainer/reward_trainer.py
lewtun: I tried this initially, but realised it wont' work because `args` is a subclass of `TrainingArguments` and thus is an instance of `TrainingArguments` which makes the if-statement always true.  See e.g. this: https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance

Happy to refactor to a try / except clause if you prefer :)

younesbelkada: I see perfect, thanks a lot lewis for explaining!

younesbelkada (commented):
Thanks again for your work on this @lewtun !

PR #751, opened 2023-09-11
[`core`] Fix import of `randn_tensor`

younesbelkada:
Fixes https://github.com/huggingface/autotrain-advanced/issues/259

With https://github.com/huggingface/diffusers/pull/4829 being merged it changed a bit the structure of diffusers lib. the method `randn_tensor` which always lives inside `diffusers.utils.torch_utils` has been removed from `diffusers.utils`' init. Therefore causes import issues for users that are using the latest diffusers or in the next release. This PR fixes this

cc @kashif @vwxyzjn

### DIFF
--- a/trl/models/modeling_sd_base.py
+++ b/trl/models/modeling_sd_base.py
@@ -24,7 +24,7 @@ from diffusers import DDIMScheduler, StableDiffusionPipeline, UNet2DConditionMod
 from diffusers.loaders import AttnProcsLayers
 from diffusers.models.attention_processor import LoRAAttnProcessor
 from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion import rescale_noise_cfg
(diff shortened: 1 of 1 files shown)


kashif:
shall we also update the required diffusers version `>= 0.21.0` which should be out in the next few days
`

vwxyzjn:
LGTM. Let’s wait for tests to pass and merge.

younesbelkada:
Thanks everyone!

> shall we also update the required diffusers version >= 0.21.0 which should be out in the next few days

I think this is not strictly needed as that method always lived there since 0.18.0: https://github.com/huggingface/diffusers/blob/v0.18.0/src/diffusers/utils/torch_utils.py#L36

younesbelkada:
Yes @neo we will have a release soon

PR #752, opened 2023-09-11
Add benchmark CI

vwxyzjn:
This PR adds the benchmark CI workflow so that we can follow up with separate PRs to trigger it.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,85 @@
+name: "Benchmark on Comment"
+
+# https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  Benchmark:
+    strategy:
+      fail-fast: true
+      matrix:
+        python-version: [3.8]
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,85 @@
...
+  Benchmark:
+    strategy:
+      fail-fast: true
+      matrix:
+        python-version: [3.8]
+        os: [self-hosted]
+
+    name: Benchmark
+    # Only run if it#s a PR and the comment contains /Benchmark
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/Benchmark') && contains('["vwxyzjn", "younesbelkada", "lvwerra"]', github.actor)
```
vwxyzjn, line 18: Note there is a restriction that only `vwxyzjn", "younesbelkada", "lvwerra"` can trigger CI

younesbelkada: Sounds great ! We can always add more maintainers later if needed

younesbelkada (approved):
Thanks a lot ! 🚀

PR #757, opened 2023-09-12
fix const-length-dataset information leak

JackeyLove1:
To fix the information leakage in autoregressive language models with ConstantLengthDataset, we can fix the problem by separating input from labels. When training an autoregressive model, we typically want the model to predict the next token. Thus, there should be a one-position offset between the input and the labels. For example ,in the sentence "The cat is on bed.", if the input is [The, cat, is], then the labels should be [cat, is, on]. After shift the example array, the labels are the next token of input_ids.

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -554,9 +554,11 @@ class ConstantLengthDataset(IterableDataset):
                 random.shuffle(examples)
             for example in examples:
                 self.current_size += 1
+                input_example = example[:-1]
+                label_example = example[1:]
                 yield {
-                    "input_ids": torch.LongTensor(example),
-                    "labels": torch.LongTensor(example),
+                    "input_ids": torch.LongTensor(input_example),
+                    "labels": torch.LongTensor(label_example),
                 }
 
 


younesbelkada (commented):
Hi @JackeyLove1 
Thanks a lot for your PR, unfortunately I am not sure these changes are required to make things work, in transformers this is done under the hood by the model itself, check for instance what is done for llama: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L842-L852 
If we want to cover the case someone uses `ConstantLengthDataset` as a standalone object independently from transformers, IMO it should be the users responsibility to correctly shift the labels and inputs. What do you think?

JackeyLove1:
> Hi @JackeyLove1 Thanks a lot for your PR, unfortunately I am not sure these changes are required to make things work, in transformers this is done under the hood by the model itself, check for instance what is done for llama: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L842-L852 If we want to cover the case someone uses `ConstantLengthDataset` as a standalone object independently from transformers, IMO it should be the users responsibility to correctly shift the labels and inputs. What do you think?

PR #758, opened 2023-09-12
[WIP] Extend DeepSpeed integration to ZeRO-{1,2,3}

lewtun:


docs/source/customization.mdx
```diff
@@ -32,17 +32,6 @@ All of the trainers in TRL can be run on multiple GPUs together with DeepSpeed Z
 accelerate launch --config_file=examples/accelerate_configs/deepspeed_zero{1,2,3}.yaml --num_processes {NUM_GPUS} path_to_your_script.py --all_arguments_of_the_script
 ```
 
-Note that for ZeRO-3, a small tweak is needed to initialize your reward model on the correct device via the `zero3_init_context_manager` context manager. Here is an example adapted for the [`sentiment_tuning`](https://github.com/huggingface/trl/blob/main/examples/scripts/sentiment_tuning.py):
```
lewtun, line 35: As discussed offline, this is no longer needed for `deepspeed>=0.9.5`

examples/accelerate_configs/deepspeed_zero1.yaml
```diff
@@ -8,7 +8,7 @@ distributed_type: DEEPSPEED
 downcast_bf16: 'no'
 machine_rank: 0
 main_training_function: main
-mixed_precision: 'no'
+mixed_precision: 'bf16'
```
lewtun, line 11: We can now set this as the default since we initialise both the reference and active models with DeepSpeed

examples/scripts/sentiment_tuning.py
```diff
@@ -38,6 +38,9 @@ class ScriptArguments:
     # NOTE: gpt2 models use Conv1D instead of Linear layers which are not yet supported in 8 bit mode
     # models like gpt-neo* models are more suitable.
     model_name: Optional[str] = field(default="lvwerra/gpt2-imdb", metadata={"help": "the model name"})
+    reward_model_name: Optional[str] = field(
```
lewtun, line 41: I've added this arg to make it easier to configure the running of this script

trl/trainer/ppo_trainer.py
```diff
@@ -1389,27 +1384,18 @@ def _show_tokens(self, tokens, masks):
...
         print(text)
 
-    def _prepare_deepspeed_zero3(self, model: PreTrainedModelWrapper):
+    def _prepare_deepspeed(self, model: PreTrainedModelWrapper):
         # Adapted from accelerate: https://github.com/huggingface/accelerate/blob/739b135f8367becb67ffaada12fe76e3aa60fefd/src/accelerate/accelerator.py#L1473
-        # TODO: figure out if any other parameters are needed to optimize inference
         deepspeed_plugin = self.accelerator.state.deepspeed_plugin
-        batch_size_per_device = deepspeed_plugin.deepspeed_config["train_micro_batch_size_per_gpu"]
-        # See DeepSpeed docs for definition of these parameters: https://deepspeed.readthedocs.io/en/latest/zero3.html
-        config_kwargs = {
```
lewtun, line 1398: All these parameters are set automatically by `accelerate` and this don't need duplicating. One check I need to make is the inclusion of gradient accumulation

vwxyzjn:
@lewtun nice work! I love that there are gpt2 runs across different zero stages. Could you also test zero 2 + 3 on larger models such as falcon 7b or cerebras GPT 6.7B?

docs/source/customization.mdx
lewtun: Indeed, ZeRO-3 hangs after 10 steps without the context manager. Curiously, this happens exactly after the `pipleine()` method throws a warning that one is using the pipeline sequentially and presumably this is the root cause

lewtun:
> @lewtun nice work! I love that there are gpt2 runs across different zero stages. Could you also test zero 2 + 3 on larger models such as falcon 7b or cerebras GPT 6.7B?

Yes, I'm running the Cerebras models as we speak and will report back when the runs are done :)

lewtun:
Update on running 3 x 6.7B models with DeepSpeed on `sentiment_tuning.py`:

* ZeRO-2 works fine and doesn't OOM
* ZeRO-3 works fine and doesn't OOM

Here's the command I used to test:

```shell
accelerate launch --config_file=examples/accelerate_configs/deepspeed_zero{2,3}.yaml examples/scripts/sentiment_tuning.py --batch_size 32 --mini_batch_size 32 --
log_with wandb --model_name cerebras/Cerebras-GPT-6.7B --reward_model_name cerebras/Cerebras-GPT-6.7B
```

Interestingly, although ZeRO-3 is less memory intensive, the savings aren't as high as I would have expected on a single node:

Screenshot 2023-09-12 at 15 48 25

examples/scripts/sentiment_tuning.py
```diff
@@ -175,7 +177,19 @@ def collator(data):
...
+        sentiment_pipe = pipeline("sentiment-analysis", model=script_args.reward_model_name, device=device)
+else:
+    sentiment_pipe = pipeline("sentiment-analysis", model=script_args.reward_model_name, device=device)
+
+# Some tokenizers like GPT-2's don't have a padding token by default, so we set one here.
+if sentiment_pipe.tokenizer.pad_token_id is None:
+    sentiment_pipe.tokenizer.pad_token_id = tokenizer.pad_token_id
+
+if sentiment_pipe.model.config.pad_token_id is None:
+    sentiment_pipe.model.config.pad_token_id = tokenizer.pad_token_id
```
younesbelkada, line 192: Do you know why this was not needed before?

docs/source/customization.mdx
```diff
@@ -32,7 +32,7 @@ All of the trainers in TRL can be run on multiple GPUs together with DeepSpeed Z
...
+Note that for ZeRO-3, a small tweak is needed to initialize your reward model on the correct device via the `zero3_init_context_manager()` context manager. In particular, this is needed to avoid DeepSpeed hanging after a fixed number of training steps. Here is a snippet of what's involved from the [`sentiment_tuning`](https://github.com/huggingface/trl/blob/main/examples/scripts/sentiment_tuning.py) example:
```
younesbelkada, line 35:
```suggestion
Note that for ZeRO-3, a small tweak is needed to initialize your reward model on the correct device via the `zero3_init_context_manager()` context manager. In particular, this is needed to avoid DeepSpeed hanging after a fixed number of training steps. Here is a snippet of what is involved from the [`sentiment_tuning`](https://github.com/huggingface/trl/blob/main/examples/scripts/sentiment_tuning.py) example:
```

younesbelkada (approved):
Looks really great to me, thanks a lot for the great investigation and spending some time on benchmarking and testing, to make sure we now support DS Zero 1, 2, 3!

examples/scripts/sentiment_tuning.py
lewtun: It's usually not needed if you've already trained a proper reward model because this comes with a proper padding token. However, if you want to plug and play with any causal LM on the Hub then this is typically needed to avoid throwing errors in the pipeline

lewtun:
Hi @uahmad235 !

Here's answers to your questions:

1. I'm using a single node of 8 x A100s with 80GB vRAM each
2. I have 1TB CPU RAM
3. I am using the default deepspeed configs from the repo with no changes (i.e. no offloading)

It will likely be tight to fit 3 x 7B models on 2 x A6000s, so one possibility would be to quantize the reward model by passing `load_in_8bit=True`. Alternatively, you could try shrinking the batch size & increasing the number of gradient accumulation steps.

Hope that helps!

uahmad235:
Thanks for the info @lewtun.
I am actually using the default reward model (i.e., `lvwerra/distilbert-imdb`) so it is not too heavy on the memory.
And yeah, I tried using `load_in_8bit` with `batchsize=4`. It runs into runtime error at `ppo_trainer.step()`. Here's the issue i get:
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0!

However, using `torch_dtype=torch.bfloat16` does help a little with memory but not enough that it can allow me to make gradient updates even with `batch_size=2`

Seems like i might have to go for a pair of A100s.

PR #759, opened 2023-09-12
Add default Optim to DPO example

natolambert:
Comparison here: https://wandb.ai/huggingface/huggingface/reports/RMSProp-DPO--Vmlldzo1Mzc2NDQ3

### DIFF
--- a/examples/dpo.py
+++ b/examples/dpo.py
@@ -67,6 +67,13 @@ class ScriptArguments:
         },
     )
 
+    # optimizer settings
+    warmup_steps: Optional[int] = field(default=150, metadata={"help": "Number of warmup steps for optimizer"})
+    optim: Optional[str] = field(
+        default="RMSprop",
+        metadata={"help": "Optimizer to use. Default is RMSprop, if none" "passed defaults to Transformers trainer."},
+    )
+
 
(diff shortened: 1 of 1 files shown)


vwxyzjn:
Hi @natolambert, thanks for the PR. There is a lot of metric in the wandb report; which are the ones we should look at?


image

natolambert:
Mostly this PR is just one of many things to get closer to the original implementation. Thanks for sharing the screenshot (I didn't realize it was private). 
@vwxyzjn the goal of this was to show _**there is not a big difference**_. We'll have more signal soon as we scale it.

examples/dpo.py
```diff
@@ -145,6 +145,13 @@ def split_prompt_and_responses(sample) -> Dict[str, str]:
         report_to=script_args.report_to,
     )
 
+    # Trainer to match original paper
+    warmup_steps = 150
```
lewtun, line 149: Should we make this configurable by the user?

tomaarsen: Makes sense to me

natolambert: fixed

younesbelkada (approved):
Thank you @natolambert !

PR #760, opened 2023-09-12
Imrpove benchmark ci

vwxyzjn:
https://github.com/huggingface/trl/pull/754#issuecomment-1715990325 actually triggered an infinite loop — this is because the comment contained "Here are the benchmark results" which somehow triggered the `contains(github.event.comment.body, '/Benchmark')` condition. This PR tighten things up further with `/benchmark-core-trl-experiments` command, making accidentally triggering the CI less likely.

Screenshot 2023-09-12 at 4 00 16 PM


younesbelkada (approved):
Nice catch, thanks very much !

PR #763, opened 2023-09-13
[PPOTrainer] - add comment of zero masking (from second query token)

zuoxingdong:
It took a while to understand why zero-masked tokens are one less than the length of query tokens. 

If I got it correctly, it is because the first logit (and state-value) from the outputs refers to the second token in the query. 

Hope this comment can be helpful to others who may encounter a similar question in the first-pass reading of the code :)


younesbelkada (approved):
Great, makes sense, thank you!

PR #764, opened 2023-09-13
Support fork in benchmark CI

vwxyzjn:
Basically just use `gh pr checkout xx` to support dealing with fork in the benchmark CI


younesbelkada (approved):
Thanks a lot for supporting forks for benchmark CIs !

PR #766, opened 2023-09-13
init custom eval loop for further DPO evals

natolambert:
To add:
- [x] save generations to `game_log` like PPOTrainer
- [x] [Moving to separate PR] compute KL distance in each evaluation loop

cc @tomaarsen feel free to directly modify when you're working tomorrow.


natolambert:
Here's an example of what this looks like! I'm only running one batch for inference in evals because it slows everything down. Open to feedback here, exciting!

Screenshot 2023-09-13 at 4 31 38 PM

trl/trainer/dpo_trainer.py
```diff
@@ -164,6 +180,7 @@ def __init__(
                 )
                 max_target_length = 128
 
+            self.max_length = max_length
```
younesbelkada, line 183: Do you know why before this PR we were using `self.config.max_length`?

younesbelkada (approved):
Looks great thank you ! I just left one question, if that's the case maybe let's use `self.config.max_length` all over the place? What do you think?

trl/trainer/dpo_trainer.py
tomaarsen: It seems fine to stick with `self.config.max_length` imo.

natolambert: the function with `self.config.max_length` was not used, and `self.config.max_length` did not exist.
I think maybe it should be `self.args.max_length`, let me check.

natolambert: I double checked, `max_length` is not in `TrainerArguments`, so it needs to be handled in the `DPOTrainer` init

PR #770, opened 2023-09-14
Update custom eval loop to aid DPO debugging

tomaarsen:
Hello!

This is intended to be pushed directly on top of [dpo_custom_eval](https://github.com/huggingface/trl/tree/dpo_custom_eval) i.e. on top of #766, but I don't have the permissions for that.

## Pull Request overview
* `sample_during_eval` is now `generate_during_eval` - I think `sample` is a bit too vague.
* `return_tokens` was unused, so I removed it.
* Prevent test failures due to `wandb` import without having `wandb` as a mandatory dependency. I added import utils for W&B & a test.
* Optimize random batch selection.
* Separate prompt and Policy/Reference responses in game log table.

This PR is a WIP.

- Tom Aarsen

### DIFF
--- a/examples/dpo.py
+++ b/examples/dpo.py
@@ -157,7 +157,7 @@ if __name__ == "__main__":
         max_length=script_args.max_length,
         max_target_length=script_args.max_target_length,
         max_prompt_length=script_args.max_prompt_length,
-        sample_during_eval=True,
+        generate_during_eval=True,
     )
 
     # 6. train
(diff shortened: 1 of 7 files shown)


tomaarsen:
Bad news @natolambert, the `islice` still seems to iterate over all elements until `random_index`. 

image

tomaarsen:
That's easy to resolve though. The new approach takes ~0.005 seconds regardless of which index is used.

image

This also doesn't restrict us to batches anymore. We can just go for 1 sample now, for example.

Issue #788, opened 2023-09-18
Typo in SFTTrainer example


filippobistaffa:
I believe the default value should be `"none"` instead of `None`:

https://github.com/huggingface/trl/blob/0a6c42c12c637bb7f28782fa72ec45dd64bce0bd/examples/scripts/sft_trainer.py#L43

younesbelkada:
I think it can be also a `NoneType`, looking at the accelerate code: https://github.com/huggingface/accelerate/blob/main/src/accelerate/accelerator.py#L381 
Do you face an issue when running the script with default arguments?

filippobistaffa:
With the code as it is now I get the following error:
`wandb.errors.UsageError: api_key not configured (no-tty). call wandb.login(key=[your_api_key])`
If I change this line
https://github.com/huggingface/trl/blob/0a6c42c12c637bb7f28782fa72ec45dd64bce0bd/examples/scripts/sft_trainer.py#L108
to `report_to="none"`, I don't get that error.

filippobistaffa:
While I was inspecting a log, I found this message
`The default value for the training argument `--report_to` will change in v5 (from all installed integrations to none). In v5, you will need to use `--report_to all` to get the same behavior as now.`
So it seems that the default is indeed "all".

younesbelkada:
Makes sense thanks for investigating ! Would you be happy to open a PR to fix it ? otherwise happy to do it !

filippobistaffa:
I will do it, no worries :)

younesbelkada:
Thanks !

PR #792, opened 2023-09-19
Changed the default value of the `log_with` argument

filippobistaffa:
This change avoids setting report_to="all" (the default behavior in transformers v4), which could lead to unexpected error messages for inexperienced users. Note that the default value of report_to will change anyway to "none" in transformers v5.

Fixes https://github.com/huggingface/trl/issues/788


younesbelkada (approved):
Thanks a lot!

PR #795, opened 2023-09-19
Add deepspeed experiment

vwxyzjn:


vwxyzjn:
**[COSTA BENCHMARK BOT]**: Here are the results

vwxyzjn:
**[COSTA BENCHMARK BOT]**: Here are the results

benchmark/benchmark_level2.sh
```diff
@@ -1,4 +1,4 @@
-# compound
+# compound: gpt2xl + grad_accu
```
lewtun, line 1: For my own understanding, is this `compound` arg documented somewhere?

benchmark/benchmark_level2.sh
```diff
@@ -8,4 +8,16 @@ python benchmark/benchmark.py \
     --slurm-gpus-per-task 1 \
     --slurm-ntasks 1 \
     --slurm-total-cpus 12 \
-    --slurm-template-path benchmark/trl.slurm_template
\ No newline at end of file
+    --slurm-template-path benchmark/trl.slurm_template
+
+# compound: Cerebras-GPT-6.7B + deepspeed zero2 + grad_accu
+python benchmark/benchmark.py \
+    --command "accelerate launch --config_file examples/accelerate_configs/deepspeed_zero2.yaml examples/scripts/sentiment_tuning.py --ppo_config.exp_name sentiment_tuning_Cerebras-GPT-6.7B_grad_accu_deepspeed_stage2  --ppo_config.batch_size 32  --ppo_config.mini_batch_size 32 --ppo_config.log_with wandb --ppo_config.model_name cerebras/Cerebras-GPT-6.7B --ppo_config.reward_model sentiment-analysis:cerebras/Cerebras-GPT-6.7B" \
```
lewtun, line 15: Eventually I think we should do the "proper" thing and fine-tune these models on IMDB so we have a genuine good policy / reward model. Of course, not necessary for this PR, but perhaps good to be as realistic as possible for the benchmark

lewtun (approved):
Thanks a lot for adding this sweet benchmark 🚀 ! I left a comment about adding a benchmark for ZeRO-3 but that can also be a separate PR if you prefer

benchmark/benchmark_level2.sh
vwxyzjn: The `compound` comment simply means we are using more features at once (e.g., in this case, we are using a larger model and gradiant accumulation at the same time) :)

benchmark/benchmark_level2.sh
vwxyzjn: I think that sounds good. Perhaps we can set up an end-to-end example where we train the reward model and then the policy model at the same time.

benchmark/benchmark_level2.sh
```diff
@@ -8,4 +8,16 @@ python benchmark/benchmark.py \
     --slurm-gpus-per-task 1 \
     --slurm-ntasks 1 \
     --slurm-total-cpus 12 \
-    --slurm-template-path benchmark/trl.slurm_template
\ No newline at end of file
+    --slurm-template-path benchmark/trl.slurm_template
+
+# compound: Cerebras-GPT-6.7B + deepspeed zero2 + grad_accu
```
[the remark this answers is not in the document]
vwxyzjn: Let's probably do this in a separate PR.

PR #799, opened 2023-09-20
[`DPO`] Revert "Add default Optim to DPO example (#759)"

younesbelkada:
This reverts commit d603e7c52704054a9e7f306ae63acdafaa3d179a.

Temporarly reverts #759 as the example script seems to fail on main, making it not usable for users, flagged in https://github.com/huggingface/trl/commit/d603e7c52704054a9e7f306ae63acdafaa3d179a#commitcomment-127787118

Let's revert the PR until we find a better fix 

cc @natolambert

### DIFF
--- a/examples/dpo.py
+++ b/examples/dpo.py
@@ -67,13 +67,6 @@ class ScriptArguments:
         },
     )
 
-    # optimizer settings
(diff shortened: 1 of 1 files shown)


younesbelkada:
We should maybe add directly the RMSProp support in transformers

PR #811, opened 2023-09-23
Small fixes to the PPO trainer doc and script.

namin:
One outstanding issue is that ppo_trainer.save_model doesn't exist. How do we actually save the model after training?

### DIFF
--- a/docs/source/ppo_trainer.mdx
+++ b/docs/source/ppo_trainer.mdx
@@ -24,7 +24,7 @@ Resulting in the following subset of the dataset:
 
 ```py
 ppo_dataset_dict = {
-    "prompt": [
+    "query": [
         "Explain the moon landing to a 6 year old in a few sentences.",
         "Why aren’t birds real?",
         "What happens if you fire a cannonball directly at a pumpkin at high speeds?",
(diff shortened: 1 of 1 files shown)


lvwerra:
Hi @namin, thanks for the PR. However, I don't quite understand why the fixes are needed and what the changes fix exactly besides the typo in the docs. Can you elaborate?

namin:
Sorry, I thought I needed these changes to get the script working, but maybe it was before installing from main instead of release. I reverted those changes, and kept only the doc changes.

Thank you!

younesbelkada (approved):
The changes look great to me, thanks!

namin:
Thank you. I still have a lingering question about saving the model. The doc says to use `save_model` but this method is not available on the PPO Trainer. So looking at the source code, I use `_save_pretrained`, but that is an internal method. WDYT? Should the PPO Trainer expose `save_model`?

lvwerra:
The `PPOTrainer` inherits from the `PyTorchModelHubMixin` and as such has a `save_pretrained` method by default that makes use of the private `_save_pretrained` method:

```python
    >>> import torch
    >>> import torch.nn as nn
    >>> from huggingface_hub import PyTorchModelHubMixin


    >>> class MyModel(nn.Module, PyTorchModelHubMixin):
    ...     def __init__(self):
    ...         super().__init__()
    ...         self.param = nn.Parameter(torch.rand(3, 4))
    ...         self.linear = nn.Linear(4, 5)

    ...     def forward(self, x):
    ...         return self.linear(x + self.param)
    >>> model = MyModel()

    # Save model weights to local directory
    >>> model.save_pretrained("my-awesome-model")
```

PR #817, opened 2023-09-25
Use uniform config

vwxyzjn:
This PR converts various configs to the new format. Things should be backward-compatible at this PR. 

I did not refactor `stable_diffusion_tuning.py` nor `reward_trainer.py` because that refactor might involve documentation changes. We should do them in separate PRs. For example, we would probably do `python stable_diffusion_tuning.py --ddpo_config.train_batch_size 32` instead of `python stable_diffusion_tuning.py --train_batch_size 32`. Let's probably do them in separate PRs.


lvwerra (approved):
Thanks a lot @vwxyzjn - LGTM! I'll also let @younesbelkada have a look :)

PR #821, opened 2023-09-26
Add RMSProp back to DPO

natolambert:
To be merged only after https://github.com/huggingface/transformers/pull/26425 is merged, and we update `setup.py` for the needed version of this.

I installed the transformers PR from source and ran with the DPO script, looks to be running 🤗


younesbelkada (approved):
Great work, thanks a lot for taking care of this @natolambert !

PR #830, opened 2023-10-04
testing doc-builder

mishig25:
testing huggingface/doc-builder#423


vwxyzjn:
@mishig25 thanks for the fix. The CI passes but it seems the docs is still 500. https://moon-ci-docs.huggingface.co/docs/trl/pr_830/en/index

image

PR #833, opened 2023-10-04
Set trust remote code to false by default

lewtun:
In the maintained examples, `trust_remote_code` is set to `True` by default which means they cannot be run with the Falcon models due to the existence of custom modelling code in the model repos (takes precedence over `transformers` modelling code).

This PR sets it to `False` by default, which is also the default in `transformers`


younesbelkada (approved):
Thanks!

PR #838, opened 2023-10-06
Clarify docstrings, help messages, assert messages in merge_peft_adapter.py

larekrow:
This PR clarifies various docstrings, help messages, and assert messages within `merge_peft_adapter.py`. In addition, it also corrects an assertion to the intended test condition.

No functionality change was made.


younesbelkada (approved):
Thanks for clarifying!

PR #842, opened 2023-10-07
Standardise example scripts

lewtun:
This PR standardises all the example scripts to follow the `run_xxx.py` convention, where `xxx` typically refers to the algorithm instead of the task (i.e. have just 1 PPO example instead of calling it "sentiment tuning"). The resulting structure is as follows:

```
examples/scripts
├── run_ddpo.py
├── run_dpo.py
├── run_ppo.py
├── run_ppo_multi_adapter.py
├── run_reward_modeling.py
└── run_sft.py
```

IMO this makes it a bit easier for newcomers to know what each script does by filename instead of guessing whether e.g. multi adapter RL refers to PPO or something else.

I also deleted an old and duplicate multi adapter RL script `multi_adapter_rl.py` which seems to be outdated.

Eventually, we could harmonize the scripts so that the SFT and reward models produced by `run_sft.py` and `run_reward_modeling.py` are the same ones that feed into `run_ppo.py` and `run_dpo.py`. This would give a true end to end pipeline that is maintained & solid for many people to work from :)


vwxyzjn (commented):
Love the standardization! Very nice change. I assume `multi_adapter_rl.py` is deprecated in favor of ` 
multi_adapter_rl_v2.py` (the now `run_ppo_multi_adapter.py`)?

vwxyzjn (approved):

vwxyzjn:
**[COSTA BENCHMARK BOT]**: Here are the results

lvwerra (approved):
Generally looks great, thanks! Small nit: I don't like the `run_xxx.py` naming that much, I think just `xxx.py` would do the job and be less redundant.

lewtun:
> Generally looks great, thanks! Small nit: I don't like the `run_xxx.py` naming that much, I think just `xxx.py` would do the job and be less redundant.

Good idea! Done in https://github.com/huggingface/trl/pull/842/commits/a6d1d90c7526bd05ae656250f9d9610545eb7bed

I'll merge if all the tests still pass

docs/source/sentiment_tuning.mdx
```diff
@@ -38,11 +38,11 @@ To run in multi-GPU setup with DDP (distributed Data Parallel) change the `devic
 
 ## Benchmarks
 
-Below are some benchmark results for `examples/scripts/sentiment_tuning.py`. To reproduce locally, please check out the `--command` arguments below.
+Below are some benchmark results for `examples/scripts/ppo.py`. To reproduce locally, please check out the `--command` arguments below.
 
 ```bash
 python benchmark/benchmark.py \
-    --command "python examples/scripts/sentiment_tuning.py --ppo_config.log_with wandb" \
+    --command "python examples/scripts/ppo.pynfig.log_with wandb" \
```
lewtun, line 45: This look off - will fix

vwxyzjn:
LG!

PR #843, opened 2023-10-07
Support both old and new diffusers import path

osanseviero:


trl/models/modeling_sd_base.py
```diff
@@ -24,7 +24,10 @@
 from diffusers.loaders import AttnProcsLayers
 from diffusers.models.attention_processor import LoRAAttnProcessor
 from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion import rescale_noise_cfg
-from diffusers.utils.torch_utils import randn_tensor
+try:
+    from diffusers.utils import randn_tensor
+except ImportError:
+    from diffusers.utils.torch_utils import randn_tensor
```
younesbelkada, line 30:
```suggestion
import diffusers
_diffusers_greater_21_4 = hasattr(diffusers.utils, "randn_tensor")
if _diffusers_greater_21_4:
    from diffusers.utils import randn_tensor
else:
    from diffusers.utils.torch_utils import randn_tensor
```

younesbelkada (commented):
Thanks ! I left a single comment, WDYT?

younesbelkada (commented):
Thanks! can you run the styling checks?
```bash
make precommit
```

younesbelkada:
Note the lazy import has been added in 0.21.0 and the latest version before that is 0.20.2

https://github.com/huggingface/diffusers/releases/tag/v0.21.0

trl/models/modeling_sd_base.py
```diff
@@ -24,7 +24,14 @@
...
 from diffusers.pipelines.stable_diffusion.pipeline_stable_diffusion import rescale_noise_cfg
-from diffusers.utils.torch_utils import randn_tensor
+
+from ..import_utils import is_diffusers_greater_than
+
+
+if is_diffusers_greater_than("0.20.2"):
+    from diffusers.utils.torch_utils import randn_tensor
+else:
+    from diffusers.utils import randn_tensor
```
lvwerra, line 34: If I understood Patrick correctly, `randn_tensor` is an internal method that could be deprecated, delete, renamed at any time. Should we not use it and instead add our own?

younesbelkada: Makes sense yeah !

trl/torch_utils.py
lvwerra, line 1: Why is there an empty file? :)

lvwerra (approved):
LGTM 🚀 Just that question about the empty file.

trl/torch_utils.py
younesbelkada: oops will remove it

PR #847, opened 2023-10-09
Optionally logging reference response

vwxyzjn:
This PR allows us to log the reference response in wandb as well, which can help us make a direct comparison of the data.

https://wandb.ai/costa-huang/trl/runs/0m8ylgjy

image

tested it to work with PEFT as well (needed to refactor the logic a bit) see https://wandb.ai/costa-huang/trl/runs/10i8zl27


lvwerra (commented):
Generally like the idea of adding reference responses. Since this touches a lot of PEFT logic I'll let @younesbelkada make sure it doesn't break anything :)

trl/trainer/ppo_trainer.py
```diff
@@ -229,12 +230,17 @@ def __init__(
         elif ref_model is None and not self.is_peft_model:
             self.ref_model = create_reference_model(self.model, num_shared_layers=num_shared_layers)
         elif self.is_peft_model:
-            self.ref_model = None
+            self.ref_model = self.model
```
younesbelkada, line 233: this is breaking no? Maybe some workflows use the check `if ppo_trainer.ref_model is None` . Do you think there is a way to achieve what you want to do in the PR without modifying this condition? 🙏 
What do you think?

vwxyzjn: Thanks for the commnet, I have addressed it by creating a separate var

trl/trainer/ppo_trainer.py
```diff
@@ -235,6 +236,12 @@ def __init__(
                 f"ref_model must be a PreTrainedModelWrapper or `None`, got {type(ref_model)} - supported "
                 f"architectures are: {SUPPORTED_ARCHITECTURES} "
             )
+        self.optional_peft_ref_model = self.model if self.is_peft_model else self.ref_model
```
younesbelkada, line 239:
```suggestion
```

trl/trainer/ppo_trainer.py
```diff
@@ -437,19 +445,32 @@ def generate(
...
                 query_tensor,
                 length_sampler=length_sampler,
                 batch_size=batch_size,
                 return_prompt=return_prompt,
                 **generation_kwargs,
             )
+            if generate_ref_response:
+                with self.optional_peft_ctx():
+                    ref_response = self._generate_batched(
+                        self.accelerator.unwrap_model(self.optional_peft_ref_model),
```
younesbelkada, line 467:
```suggestion
                        ref_model,
```

trl/trainer/ppo_trainer.py
```diff
@@ -437,19 +445,32 @@ def generate(
...
+                If set to `True` the reference response is also generated, defaults to `False`.
 
         Returns:
             `torch.LongTensor`: A tensor of shape (`batch_size`, `gen_len`) containing response tokens.
         """
 
         if isinstance(query_tensor, List):
-            return self._generate_batched(
+            response = self._generate_batched(
+                self.accelerator.unwrap_model(self.model),
```
younesbelkada, line 457:
```suggestion
                self.model,
```

younesbelkada (commented):
Hi @vwxyzjn thanks a lot for iterating! I really like the `optional_peft_ctx` context manager! 
 I left a suggestion to use `self.accelerator.unwrap_model` once and make use of a local variable to pick up the correct ref model. What do you think?

younesbelkada (approved):
Very clean! Thanks very much Costa for your work on this! 🙏

PR #850, opened 2023-10-10
Fixes reward and text gathering in distributed training

edbeeching:
This PR fixes two issues:

- **Rewards** were being wrongly summed instead of gathered. While this did not affect the mean, other stats such as std and the distribution plots in wandb were affected.
- **Texts** were not being gathered across processes, leading to text only being reported from the main process.

I have tested in two settings [single node 8 GPU](https://wandb.ai/edbeeching/trl/runs/5ovy2waw/overview?workspace=user-edbeeching) and [single node 1 GPU](https://wandb.ai/edbeeching/trl/runs/6zz4yyh2?workspace=user-edbeeching). I launched both with `--ppo-config.mini-batch-size=1 --ppo-config.batch-size=2` to confirm the gather is working correctly.

This may impact #847


trl/trainer/ppo_trainer.py
```diff
@@ -1310,17 +1320,21 @@ def log_stats(
...
+                    import torch.distributed as dist
 
-                table_rows = [list(r) for r in zip(*batch_list, rewards.cpu().tolist())]
-                logs.update({"game_log": wandb.Table(columns=[*columns_to_log, "reward"], rows=table_rows)})
-            # All reduce rewards if distributed
-            if self.is_distributed:
-                import torch.distributed as dist
+                    dist.barrier()
+                    gathered_batch_list = []
+                    for b in batch_list:
```
vwxyzjn, line 1328: Similar as above.

trl/trainer/ppo_trainer.py
```diff
@@ -1350,7 +1364,23 @@ def log_stats(
...
+                        raise ValueError(
+                            f"Columns to log {columns_to_log} are not present in the batch {batch.keys()}."
+                        )
+
+                    batch_list = [batch[column_to_log] for column_to_log in columns_to_log]
+
+                    dist.barrier()
+                    for b in batch_list:
+                        gather = [None for _ in range(dist.get_world_size())]
+                        dist.all_gather_object(gather, b)
```
vwxyzjn, line 1383: What is this doing?

trl/trainer/ppo_trainer.py
```diff
@@ -1297,6 +1297,16 @@ def log_stats(
...
 
+            # Gather rewards if distributed
+            if self.is_distributed:
+                import torch.distributed as dist
+
+                dist.barrier()
+                world_size = dist.get_world_size()
+                tensors = [torch.zeros_like(rewards) for _ in range(world_size)]
+                dist.all_gather(tensors, rewards)
+                rewards = torch.cat(tensors)
```
vwxyzjn, line 1308: Maybe all these line of code can be replaced with self.accelerator.gather(rewards)


```suggestion
            # Gather rewards if distributed
            rewards = self.accelerator.gather(rewards).flatten()
```

vwxyzjn (commented):
Very nice PR! Thanks @edbeeching this will make logging much better

trl/trainer/ppo_trainer.py
edbeeching: Ah nice, I was not aware accelerate has this helper method, I will clean up.

trl/trainer/ppo_trainer.py
edbeeching: Here I am sending the query & response from process n>1 to process 0. This is two way, perhaps there is a way to do this one way with send / recv? I don't often use those operations.

lewtun:
Begone stale bot 🪄!

edbeeching:
Yes, I will get back to this when I have the time.

edbeeching:
@vwxyzjn I have rebased and updated the PR if you can take another look.

trl/trainer/ppo_trainer.py
```diff
@@ -1368,16 +1368,6 @@ def log_stats(
...
-        else:
-            if self.is_distributed:
-                import torch.distributed as dist
-
-                if not isinstance(rewards, torch.Tensor):
-                    rewards = torch.tensor(rewards).to(self.current_device)
-
-                dist.barrier()
-                dist.all_reduce(rewards, op=torch.distributed.ReduceOp.SUM)
-
```
vwxyzjn, line 1380: @lvwerra do these code do anything?

vwxyzjn (approved):
@edbeeching I have refactored just a bit :) There seems to be some unused code that I will have @lvwerra take a look at but otherwise LGTM.

Thanks a lot for the PR!

trl/trainer/ppo_trainer.py
lvwerra: cc @younesbelkada :)

lvwerra (approved):
Looks good to me, I'll let @younesbelkada have a look too!

trl/trainer/ppo_trainer.py
```diff
@@ -1336,17 +1337,16 @@ def log_stats(
                     raise ValueError(f"Columns to log {columns_to_log} are not present in the batch {batch.keys()}.")
 
                 batch_list = [batch[column_to_log] for column_to_log in columns_to_log]
+                if self.is_distributed:
+                    self.accelerator.wait_for_everyone()
+                    gathered_batch_list = []
+                    for b in batch_list:
```
lewtun, line 1343: turbo nit to follow HF conventions to use descriptive names
```suggestion
                    for batch in batch_list:
```

lewtun (approved):
LGTM with a turbo nit about variable names :)

I also have the same question as @vwxyzjn about the remove of some distributed code, but @younesbelkada is the best person to answer that

lvwerra:
Is this ready to be merged?

vwxyzjn:
Merged! Thanks @edbeeching @lewtun @lvwerra !

PR #851, opened 2023-10-10
[SFTTrainer] Support logging response in wandb

vwxyzjn:
# What does this PR do?

This PR attempts to log some sample and reference responses in wandb, which gives concrete examples for our inspection and also makes training more informative. Basically, it's going to log something as follows:

image

Some datasets such as `timdettmers/openassistant-guanaco` do not really have a query/response structure, so I basically give the model the first half of the token in the dataset and let it generate the remaining tokens; I also added the second half of the token as the reference response to test out the SFT policy.

## Question:

I wasn't quite sure how to get the batching working with dataset and generate, though... The `dataloader` complains

```
ValueError: Unable to create tensor, you should probably activate truncation and/or padding with 'padding=True' 'truncation=True' to have batched tensors with the same length. Perhaps your features (`remaining_input_ids` in this case) have excessive nesting (inputs type `list` where type `int` is expected).
```

if I try to use a `batch_size > 1` for the dataloader.

### DIFF
--- a/examples/scripts/sft_trainer.py
+++ b/examples/scripts/sft_trainer.py
@@ -12,15 +12,19 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from collections import defaultdict
 from dataclasses import dataclass, field
 from typing import Optional
 
+import pandas as pd
 import torch
+import tyro
 from accelerate import Accelerator
 from datasets import load_dataset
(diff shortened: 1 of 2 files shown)


vwxyzjn:
update: after chatting with @lvwerra, we think maybe the change in `SFTTrainer` is unnecessary. Will checkout how can we further improve it.

JamesSand:
Hi, I have met the same problem. Does anyone solve that?

I have tried to use the demo code provided in the readme, but it still does not work....

PR #853, opened 2023-10-11
Fix version check in import_utils.py

adampauls:
IIUC, the check is currently comparing 3 to 3.8 and so is always true. I noticed this because Pyright [crashes](https://github.com/microsoft/pyright/issues/6116) because it special cases the version check.


younesbelkada (approved):
LGTM thanks

lvwerra (approved):

PR #857, opened 2023-10-11
dont use get_peft_model if model is already peft

abhishekkrthakur:


lewtun (approved):
Thanks for the fix, LGTM 🚀 !

younesbelkada (approved):
Thanks a lot abhishek !

Issue #816, opened 2023-09-25
Add RSO  (Statistical Rejection Sampling Improves Preference Optimization) Examples/Trainer `https://arxiv.org/abs/2309.06657`


philschmid:
DeepMind proposed a new method, "Statistical Rejection Sampling Optimization (RSO)" to align and learn from Human Preferences, which looks very interesting. https://arxiv.org/abs/2309.06657. 

It seems like we can reuse a lot of components from DPO using the same loss, what do you think about adding the missing utils, e.g. `conduct_rejection_sampling` and then create an example for HHH. 

Requirements are: 

- [ ]  Collect pairwise human preferences data, with each example containing a prompt, an accepted response, and a rejected response. -> HHH
- [ ] Train a pairwise Reward Model on the preference dataset to predict which response is “accepted” for a given prompt. -> could be part of the exmaples
- [ ]  Train an SFT Model on a different instruction dataset -> part of the examples 
- [ ] Apply RSO 
      - [ ] create new dataset -> new script 
      - [ ] train SFT again -> reuse script with different loss? 
      
RSO implementation
1. Sample responses using the SFT Model and rank each response with the Reward Model
2. Use Statistical rejection sampling to conduct samples favoring higher rewards. Label the selected responses as preference pairs based on the relative reward.
4. Iteratively train the SFT Model using the new labeled preference pairs with logistic loss.

philschmid:
> > 3. Iteratively train the SFT Model using the new labeled preference pairs with logistic loss.
> 
> This would be the exact same DPO implementation, right?

Thats my understanding as well.

philschmid:
@lvwerra please reopen

PR #866, opened 2023-10-12
[DPO] add SLiC hinge loss to DPOTrainer

kashif:
from the [RSO paper]( https://arxiv.org/pdf/2309.06657.pdf) equation (12)

#816

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -55,8 +55,8 @@ class DPOTrainer(Trainer):
             reference model is provided, the trainer will create a reference model with the same architecture as the model to be optimized.
         beta (`float`, defaults to 0.1):
             The beta factor in DPO loss. Higher beta means less divergence from the initial policy.
-        loss_type (`str`, defaults to `sigmoid`):
-            The type of DPO loss to use. Either `sigmoid` the default DPO loss or `hinge` loss from SLiC paper.
+        loss_type (`str`, defaults to `"sigmoid"`):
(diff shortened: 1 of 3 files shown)


younesbelkada (approved):
Thanks a lot for adding this @kashif !

trl/trainer/dpo_trainer.py
```diff
@@ -223,6 +226,7 @@ def __init__(
         self.padding_value = padding_value
 
         self.beta = beta
+        self.loss_type = loss_type
```
lvwerra, line 229: I think we should raise an error if someone provided the wrong loss string. I guess currently they would get a "Losses referenced before definition" kind of error.

Otherwise looks good to me!

kashif: ok fixing

trl/trainer/dpo_trainer.py
```diff
@@ -364,6 +364,8 @@ def dpo_loss(
             losses = -F.logsigmoid(self.beta * logits)
         elif self.loss_type == "hinge":
             losses = torch.relu(1 - self.beta * logits)
+         else:
+            raise ValueError(f"Unknown loss type: {self.loss_type}")
```
lvwerra, line 368: Small nit to make it more user-friendly :) 

```suggestion
            raise ValueError(f"Unknown loss type: {self.loss_type}. Should be one of ['sigmoid', 'hinge']")
```

lvwerra: btw i think the indent is off

Issue #870, opened 2023-10-13
Integrate NEFT into `SFTTrainer`


lvwerra:
NEFT is a simple trick where noise is applied to the embeddings during instruction tuning which improves the performance significantly. See https://twitter.com/tomgoldsteincs/status/1712498076340932855 

We could integrate it into the `SFTTrainer` and apply that trick to the model during training.

Necessary code (see [here](https://github.com/neelsjain/NEFTune#code)):

```python
from torch.nn import functional as F

def NEFTune(model, noise_alpha=5)
    def noised_embed(orig_embed, noise_alpha):
        def new_func(x):
            # during training, we add noise to the embedding
            # during generation, we don't add noise to the embedding
            if model.training:
                embed_init = orig_embed(x)
                dims = torch.tensor(embed_init.size(1) * embed_init.size(2))
                mag_norm = noise_alpha/torch.sqrt(dims)
                return embed_init + torch.zeros_like(embed_init).uniform_(-mag_norm, mag_norm)
            else:
                return orig_embed(x)
        return new_func
    ##### NOTE: this is for a LLaMA model ##### 
    ##### For a different model, you need to change the attribute path to the embedding #####
    model.base_model.model.model.embed_tokens.forward = noised_embed(model.base_model.model.model.embed_tokens, noise_alpha)
return model
```

cc @younesbelkada

PR #871, opened 2023-10-13
[`SFTTrainer`] Adds NEFTune into `SFTTrainer`

younesbelkada:
# What does this PR do?

This PR adds NEFTune: a new technique for enhancing Supervised fine-tuning results results proposed in: https://arxiv.org/abs/2310.05914

Screenshot 2023-10-13 at 17 36 38

I propose a very simple API which is as simple as passing a valid `neftune_noise_alpha` argument when initializing the `SFTTrainer`. To avoid any surprising behaviour, we should revert to the original forward method at the end of the training. This is handled inside `def train()` that is a wrapper around `Trainer`'s `train` method. 

```python
from trl import SFTTrainer

dataset = load_dataset("imdb", split="train")

trainer = SFTTrainer(
    "facebook/opt-350m",
    train_dataset=dataset,
    dataset_text_field="text",
    max_seq_length=512,
    neftune_noise_alpha=5,
)
trainer.train()
```

I still need to add few lines in the documentation. 

Fixes: https://github.com/huggingface/trl/issues/870

cc @lvwerra @neelsjain @YuxinWenRick

### DIFF
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -396,6 +396,34 @@ After loading your model, you can either train it as it is, or attach adapters a
 
 In contrary to Flash Attention 1, the integration makes it possible to train your model on an arbitrary dataset that also includes padding tokens.
 
+### Enhance model's performances using NEFTune
+
+NEFTune has been introduced by the paper "NEFTune: Noisy Embeddings Improve Instruction Finetuning" from Jain et al. it consists of adding noise to the embedding vectors during training. According to the abstract of the paper:
+
+>  Standard finetuning of LLaMA-2-7B using Alpaca achieves 29.79% on AlpacaEval, which rises to 64.69% using noisy embeddings. NEFTune also improves over strong baselines on modern instruction datasets. Models trained with Evol-Instruct see a 10% improvement, with ShareGPT an 8% improvement, and with OpenPlatypus an 8% improvement. Even powerful models further refined with RLHF such as LLaMA-2-Chat benefit from additional training with
+
+<div style="text-align: center">
+<img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/neft-screenshot.png">
+</div>
+
+To use it in `SFTTrainer` simply pass `neftune_noise_alpha` when creating your `SFTTrainer` instance. Note that to avoid any surprising behaviour, the NEFTune is disabled after training to retrieve back the original behaviour of the embedding layer.
+
+```python
+from datasets import load_dataset
+from trl import SFTTrainer
+
+dataset = load_dataset("imdb", split="train")
+
+trainer = SFTTrainer(
+    "facebook/opt-350m",
+    train_dataset=dataset,
+    dataset_text_field="text",
+    max_seq_length=512,
+    neftune_noise_alpha=5,
+)
+trainer.train()
+```
+
 ## Best practices
 
 Pay attention to the following best practices when training a model with that trainer:
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -398,15 +398,15 @@ In contrary to Flash Attention 1, the integration makes it possible to train you
 
 ### Enhance model's performances using NEFTune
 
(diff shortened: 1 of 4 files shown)


BenjaminBossan:
I wonder if the implementation would be cleaner by using a post forward hook for the embedding layer instead of replacing the `forward` method completely.

younesbelkada:
@imrankh46 that feature is not merged yet in TRL main branch to use it please run:

```bash
pip install -U git+https://github.com/huggingface/trl.git@add-neftune
```

@BenjaminBossan , yes this is possible indeed, however I found it easier to understand for future users to have a standalone forward method. Would it also hurt existing hooks that accelerate attaches in case we manipulate forward post hooks?

BenjaminBossan:
> yes this is possible indeed and I think it is cleaner, however I found it easier to understand for future users to have a standalone forward method. Would it also hurt existing hooks that accelerate attaches in case we manipulate forward post hooks?

My thinking was that with a forward hook, you don't need to monkey patch (which can break stuff sometimes) and there is no need to call `torch.nn.functional.embedding` explicitly (what if the embedding layer of a model does something extra?). Whether it's easier to read or not, I don't know.

Existing hooks shouldn't be affected. When registering the hook, you get back a handle for this specific hook, which allows you to remove the hook once you don't need it anymore.

docs/source/sft_trainer.mdx
```diff
@@ -396,6 +396,40 @@ After loading your model, you can either train it as it is, or attach adapters a
...
+    "facebook/opt-350m",
+    train_dataset=dataset,
+    dataset_text_field="text",
+    max_seq_length=512,
+    neftune_noise_alpha=5,
+)
+trainer.train()
+```
+
+We have tried NEFTune for supervised finetuning `mistralai/Mistral-7B-v0.1` model and proved that using NEFTune led to having a performance boost of ~25%.
```
lewtun, line 427:
```suggestion
We have tested NEFTune by training `mistralai/Mistral-7B-v0.1` on the [OpenAssistant dataset](https://huggingface.co/datasets/timdettmers/openassistant-guanaco) and validated that using NEFTune led to a performance boost of ~25% on MT Bench.
```

docs/source/sft_trainer.mdx
```diff
@@ -396,6 +396,40 @@ After loading your model, you can either train it as it is, or attach adapters a
...
+
+<div style="text-align: center">
+<img src="https://huggingface.co/datasets/trl-internal-testing/example-images/resolve/main/images/trl-neftune-mistral-7b.png">
+</div>
+
```
lewtun, line 432:
```suggestion

Note however, that the amount of performance gain is _dataset dependent_ and in particular, applying NEFTune on synthetic datasets like [UltraChat](https://huggingface.co/datasets/stingning/ultrachat) typically produces smaller gains.
```

lvwerra (approved):
Looks good to me! 🚀

cuongtran-uva:
> @imrankh46 that feature is not merged yet in TRL main branch to use it please run:
> 
> ```shell
> pip install -U git+https://github.com/huggingface/trl.git@add-neftune
> ```
> 
> @BenjaminBossan , yes this is possible indeed and I think it is cleaner, however I found it easier to understand for future users to have a standalone forward method. Would it also hurt existing hooks that accelerate attaches in case we manipulate forward post hooks?

I tried to update the trl  with neftune by using your command but it was unsuccessful. 
 
```
Cloning https://github.com/huggingface/trl.git (to revision add-neftune) to /tmp/pip-req-build-copawfzq
  Running command git clone --quiet https://github.com/huggingface/trl.git /tmp/pip-req-build-copawfzq
  WARNING: Did not find branch or tag 'add-neftune', assuming revision or ref.
  Running command git checkout -q add-neftune
  error: pathspec 'add-neftune' did not match any file(s) known to git.
  error: subprocess-exited-with-error
```

lvwerra:
The branch got deleted after merging. now you can use `@main` instead!

daehuikim:
> The branch got deleted after merging. now you can use `@main` instead!

```$ pip install -U git+https://github.com/huggingface/trl.git@main```
This works for me. Thanks!

```
$ pip freeze | grep trl
trl @ git+https://github.com/huggingface/trl.git@393dbf67494771170cb8fede30b6e77c39d85841
```
Installation check

PR #882, opened 2023-10-17
[reward_modeling] Cleaning example script

gaetanlop:
### Small changes to reward_modeling example script

Script arguments 'load_in_8_bit' and 'load_in_4_bit' were repeated multiple times.

Also removed truncation=True in the preprocess_function as it is already removing samples with higher lengths than reward_config.max_length in the filtering step.
It avoids problems with the scenario when reward_config.max_length is equal to the tokenizer max length. In this scenario, if truncation is set to True, we won't filter samples that have a higher maximum length than reward_config.max_length.

### DIFF
--- a/examples/scripts/reward_modeling.py
+++ b/examples/scripts/reward_modeling.py
@@ -72,10 +72,6 @@ class ScriptArguments:
             modules_to_save=["scores"],
         ),
     )
-    load_in_8bit: bool = False
-    """load the model in 8 bits precision"""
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Makes sense, thanks @gaetanlop !

younesbelkada:
can you push an empty commit to trigger the CI?

gaetanlop:
Done @younesbelkada

PR #885, opened 2023-10-17
[DPO] use ref model logprobs if it exists in the data

kashif:
- [x] Refactor the trainer so that it can use the logprobs from the data rather than a reference model if it exists in the dataset
- [x] Added a flag that adds reference model logprobs to the dataset before training in the dataloader creation phase
- [x] fix confusion about padding_value and label_pad_token_id
- [x] fix from #932 
- [x] fix for #1025


lvwerra (approved):
I understand this only affects internals except the new flag? Looks good to me, maybe @lewtun wants to have a look too since he's been using this quite a bit recently.

edbeeching:
Hey @kashif , a couple of questions:

1. Is the user expected to pass `ref_model=None` and `precompute_logprobs=True` to the DPOTrainer? As in this case isn't another instance of the model instantiated through a deepcopy?: https://github.com/huggingface/trl/blob/5096ff25fda682eca99b68a7fcc24fe215712e67/trl/trainer/dpo_trainer.py#L166
3. Will the pre-computation of logprobs work correctly in a distributed setting, as if we have 8 GPUs and DDP=8, I think the dataset.map will run on all GPUs? https://github.com/huggingface/trl/blob/5096ff25fda682eca99b68a7fcc24fe215712e67/trl/trainer/dpo_trainer.py#L307

trl/trainer/dpo_trainer.py
```diff
@@ -250,10 +253,17 @@ def __init__(
...
-            if not hasattr(self.accelerator.unwrap_model(self.model), "disable_adapter"):
-                raise ValueError(
-                    "You are using a `peft` version that does not support `disable_adapter`. Please update your `peft` version to the latest version."
-                )
+            # if dataset has reference model logps and sample outputs, we don't need a reference model
+            if (
+                "reference_chosen_logps" in self.train_dataset[0]
+                and "reference_rejected_logps" in self.train_dataset[0]
+            ):
+                self.ref_model = None
```
lewtun, line 261: Isn't this line redundant since we already check if `ref_model is None` above?

lewtun (commented):
Thanks a lot for optimising the DPO trainer @kashif 🔥 !

Overall the PR looks great, but I have a few questions about whether the log probs are precomputed on CPU vs GPU. 

It would also be great to see some docs on how to use this feature, e.g. showing the 3 main cases:

* Log probs are already precomputed and stored in a dataset
* Log probs are precomputed at the start of training
* Log probs are computed on the fly (previous behaviour)

Would it also make sense if I run this PR through a DPO training run to check we don't have any major regressions?

younesbelkada:
Note the current CI will be red because of the recent transformers release, fixing it here: https://github.com/huggingface/trl/pull/946

kashif:
@edbeeching  
1. yes was a bug, so now when ref_model=None and precomputed_log_probs=True we do not deep copy the main model and instead use the main model at the start to precompute the log_probs
2. so it should run the model on the GPU device as it uses the model after the preparation from the accelerator

trl/trainer/dpo_trainer.py
```diff
@@ -192,14 +194,10 @@ def __init__(
                 max_target_length = 128
 
             data_collator = DPODataCollatorWithPadding(
-                tokenizer,
```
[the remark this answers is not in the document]
kashif: so i realized the data collator was doing a bunch of things apart from the final padding of the chosen sentences in a batch and had to move that functionality out to the dataset.map helpers which is also needed for calculating the ref log_probs... Thus after moving that functionality out the collator became a lot simpler... let me check how to add a warning...

tests/test_dpo_trainer.py
```diff
@@ -122,7 +125,9 @@ def test_dpo_trainer(self, name, loss_type):
                 if param.sum() != 0:
                     self.assertFalse(torch.equal(param, new_param))
 
-    def test_dpo_trainer_without_providing_ref_model(self):
+    @parameterized.expand([["gpt2", False], ["t5", True]])
```
lewtun, line 128: Should this be `True` since we want to test `precompute_ref_log_probs=True`?
```suggestion
    @parameterized.expand([["gpt2", True], ["t5", True]])
```

tests/test_dpo_trainer.py
```diff
@@ -161,7 +167,9 @@ def test_dpo_trainer_without_providing_ref_model(self):
 
     @require_peft
     @mark.peft_test
-    def test_dpo_trainer_without_providing_ref_model_with_lora(self):
+    @parameterized.expand([["gpt2", True], ["t5", False]])
```
lewtun, line 170: Similar question here as above:
```suggestion
    @parameterized.expand([["gpt2", True], ["t5", True]])
```

trl/trainer/dpo_trainer.py
```diff
@@ -290,7 +294,219 @@ def _prepare_deepspeed(self, model: PreTrainedModelWrapper):
...
+                batch["rejected_decoder_input_ids"] = model.prepare_decoder_input_ids_from_labels(
+                    labels=batch["rejected_labels"]
+                )
+                batch["chosen_decoder_input_ids"] = model.prepare_decoder_input_ids_from_labels(
+                    labels=batch["chosen_labels"]
+                )
+
+        return batch
+
+    def compute_reference_logps(self, batch) -> Dict:
```
lewtun, line 436: turbo nit to align with the `precompute_ref_log_probs` in the config:

```suggestion
    def compute_reference_log_probs(self, batch) -> Dict:
```

Also can we add the type to `batch`?

trl/trainer/dpo_trainer.py
```diff
@@ -290,7 +294,219 @@ def _prepare_deepspeed(self, model: PreTrainedModelWrapper):
...
+        """Computes logps for reference model for a single batch of a DPO specific dataset."""
```
lewtun, line 437:
```suggestion
        """Computes log probabilities of the reference model for a single batch of a DPO specific dataset."""
```

trl/trainer/dpo_trainer.py
```diff
@@ -290,7 +294,219 @@ def _prepare_deepspeed(self, model: PreTrainedModelWrapper):
...
+            chosen_tokens["attention_mask"] = new_attention_mask_c
+
+            eos_indices_rejected = [i for i, x in enumerate(rejected_tokens["input_ids"]) if x == eos_token_id]
+            new_attention_mask_r = [
+                0 if i in eos_indices_rejected else p for i, p in enumerate(rejected_tokens["attention_mask"])
+            ]
+            rejected_tokens["attention_mask"] = new_attention_mask_r
+
+            # add EOS token to end of prompt
+            chosen_tokens["input_ids"].append(self.tokenizer.eos_token_id)
```
lewtun, line 364: For my understanding, why do we add the EOS token at the end of the prompt? For context, when we use chat templates one often has a structure like this:

```
<|system|>
{{ system_prompt }}</s>
<|user|>
{{ instruction }}</s>
<|assistant|>
{{ model_answer }}</s>
```

where the EOS token `</s>` indicates the end of a turn. My understanding of the DPO implementation is that you have a triple like `(prompt, chosen_completion, rejected_completion)`, so my question is whether users should be removing the EOS token from the user turn of their chat templates or not

trl/trainer/dpo_trainer.py
```diff
@@ -290,7 +300,217 @@ def _prepare_deepspeed(self, model: PreTrainedModelWrapper):
         model.eval()
         return model
 
-    def concatenated_inputs(self, batch: Dict[str, Union[List, torch.LongTensor]]) -> Dict[str, torch.LongTensor]:
+    def get_train_dataloader(self) -> DataLoader:
+        # tokenize the dataset and compute reference logps for training datasets
+        self.train_dataset = self.train_dataset.map(self.tokenize_batch_element)
+        if self.precompute_ref_logps:
+            self.train_dataset = self.train_dataset.map(
```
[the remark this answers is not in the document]
lewtun: Do you happen to know what the answer to the above question is (ie do we run inference on CPU or GPU)?

lewtun (approved):
Thanks for iterating @kashif ! The PR LGTM and I've left a few nits & questions regarding the `map()` and EOS tokens

trl/trainer/dpo_trainer.py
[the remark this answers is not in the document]
lvwerra: It should run wherever the `ref_model` is at this point, which is already prepared via `accelerate` no?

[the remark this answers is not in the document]
kashif: yes however if there is explicitly a ref_model then (since its no being passed to the accelerate we have to move it to the accelerate device) which is what i am doing

UniverseFly:
This PR is really saving the life! Precomputing the logprobs can make DPO way more memory-efficient, and I guess it would be even faster if we loaded two models at the same time and used ZeRO stage 3. Hope it can get merged soon. Just out of curiosity, can this also be applied to the PPO pipeline to avoid loading reward and policy models at the same time? For example, we first load the policy model to generate responses and dump the old logprobs (then unload), load the reward model to score the responses (then unload), and finally load the policy model again to update the weights.

younesbelkada (approved):
Thanks a lot for making DPO more accessible! Let's 🚢 it !

trl/trainer/dpo_trainer.py
```diff
@@ -374,42 +377,321 @@ def _prepare_deepspeed(self, model: PreTrainedModelWrapper):
         model.eval()
         return model
 
-    def concatenated_inputs(self, batch: Dict[str, Union[List, torch.LongTensor]]) -> Dict[str, torch.LongTensor]:
+    def get_train_dataloader(self) -> DataLoader:
+        # tokenize the dataset and compute reference logps for training datasets
+        self.train_dataset = self.train_dataset.map(self.tokenize_batch_element)
+        if self.precompute_ref_log_probs:
```
i4never, line 383: Maybe it should be `if not self.precompute_ref_log_probs` here? Same in `get_eval_data_loader`

kashif: opps so i wanted to keep the old default behaviour ... let me fix the eval data loader

kashif: no its also fine in the `get_eval_dataloader`

i4never: My mistake. Great jobs. Hope this could be merged soon.

younesbelkada (approved):
LGTM, what do you think @lvwerra @lewtun @edbeeching ?

trl/trainer/dpo_trainer.py
```diff
@@ -110,11 +112,13 @@ class DPOTrainer(Trainer):
         compute_metrics (`Callable[[EvalPrediction], Dict]`, *optional*):
             The function to use to compute the metrics. Must take a `EvalPrediction` and return
             a dictionary string to metric values.
+        precompute_ref_log_probs (`bool`, defaults to `False`):
```
lewtun, line 115: How about raising a `ValueError` in the init if DeepSpeed ZeRO-3 is activated with a message like `"`precompute_ref_log_probs` is not supported with DeepSpeed ZeRO-3."`?

trl/trainer/dpo_trainer.py
kangqiyue: > For my understanding, why do we add the EOS token at the end of the prompt? For context, when we use chat templates one often has a structure like this:
> 
> ```
> <|system|>
> {{ system_prompt }}</s>
> <|user|>
> {{ instruction }}</s>
> <|assistant|>
> {{ model_answer }}</s>
> ```
> 
> where the EOS token `</s>` indicates the end of a turn. My understanding of the DPO implementation is that you have a triple like `(prompt, chosen_completion, rejected_completion)`, so my question is whether users should be removing the EOS token from the user turn of their chat templates or not

I think the  chosen_tokens here is the concatenated inputs of 'prompt + chosen', instead of the prompt only. please check if it is right or not? I think the EOS token is append to the end of the whole "prompt + chosen_completion" here

lewtun:
This looks good to go - let merge it 🔥 !

PR #887, opened 2023-10-18
Add whiten ops before compute advatanges

SingL3:
1. From LLaMA 2 paper, it says:
```
We also find it important to whiten the final linear scores (shown here by reversing the sigmoid with the logit function) in order to increase stability and balance properly with the KL penalty term (β) above.
```
2. This function is taken from [alpaca_farm](https://github.com/tatsu-lab/alpaca_farm/blob/64e489c67ea502ab5fa944bebde3078c9722f6ee/src/alpaca_farm/rl/ppo_trainer.py#L86)

### DIFF
--- a/trl/trainer/ppo_config.py
+++ b/trl/trainer/ppo_config.py
@@ -115,6 +115,8 @@ class PPOConfig:
     """Use score normalization. Only applicable if use_score_scaling is True"""
     score_clip: Optional[float] = None
     """Score clipping"""
+    whiten_rewards: bool = False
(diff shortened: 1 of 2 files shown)


vwxyzjn:
@SingL3, thanks for the change! It makes sense. I have kicked started our benchmark to help understand its effect

image

vwxyzjn (approved):
No noticeable difference in performance with `whiten_rewards=True`. We can even the set the default value to `True`. WDYT @lvwerra?

PR #888, opened 2023-10-18
Re-add summarize example

vwxyzjn:
Ran the training with gpt2 and got ~60% validation accuracy using 94k comparison data, which reproduces the reported validation accuracy in early TRL. Here is the [run](
https://wandb.ai/costa-huang/huggingface/runs/t8q7izac).

image
image

The reference valid accuracy is higher (64%) and requires less data (64k comparison data)

image


vwxyzjn:
I added @mnoukhov's script which seems to work much better. His original setting with LORA, pythia1b, and 3 epochs it can obtain near 70% validation accuracy! I then tried full-param tuning and got similar accuracy with just 1 epoch, so maybe we can focus on this particular setting (which is OAI's original setting as well) 

[570dd3f](https://github.com/huggingface/trl/pull/888/commits/570dd3f104ca377a43b6ba6e6a9b4a6df12fb49e) refactors the code a bit to be shorter and seemed to perform very similarly. 

image

examples/scripts/reward_trainer_summarize.py
```diff
@@ -13,155 +13,217 @@
...
+    torch_dtype = torch.float32
+
+    model = AutoModelForSequenceClassification.from_pretrained(
+        args.model_name,
+        num_labels=1,
+        torch_dtype=torch_dtype,
     )
 
+    model.config.torch_dtype = torch_dtype
+    model.config.use_cache = True
```
mnoukhov, line 171: I didn't use gradient checkpointing so I set this to true but I think it should be `not gradient_checkpointing` officially

examples/scripts/reward_trainer_summarize.py
```diff
@@ -0,0 +1,229 @@
...
+from trl import RewardTrainer
+
+
+tqdm.pandas()
+
+### fix from https://github.com/huggingface/trl/issues/274
+# Define and parse arguments.
+@dataclass
+class ScriptArguments:
+    model_name: str = "EleutherAI/pythia-1b"
```
mnoukhov, line 34: I used pythia-1b-deduped, which might perform better

examples/scripts/reward_trainer_summarize.py
```diff
@@ -0,0 +1,229 @@
...
+    def preprocess_function(examples):
+        new_examples = {
+            "input_ids_chosen": [],
+            "attention_mask_chosen": [],
+            "input_ids_rejected": [],
+            "attention_mask_rejected": [],
+        }
+        for prompt, chosen, rejected in zip(examples["prompt"], examples["chosen"], examples["rejected"]):
+            tokenized_chosen = tokenizer(
+                prompt + "\n\n" + chosen, padding="max_length", truncation=True, max_length=args.seq_length
```
mnoukhov, line 197: so it turns out the chosen and rejected have two spaces after TLDR: instead of one.  It's not a big issue but you can remove it with `chosen = chosen[:6] + chosen[7:]` and same with rejected.

examples/scripts/reward_trainer_summarize.py
```diff
@@ -0,0 +1,229 @@
...
+    torch_dtype = torch.float32
+
+    model = AutoModelForSequenceClassification.from_pretrained(
+        args.model_name,
+        num_labels=1,
+        torch_dtype=torch_dtype,
+    )
+
+    model.config.torch_dtype = torch_dtype
+    model.config.use_cache = True
```
mnoukhov, line 171: I didn't use gradient checkpointing so I set this to true but I think it should be `not gradient_checkpointing` officially

examples/scripts/reward_trainer_summarize.py
```diff
@@ -0,0 +1,229 @@
...
+            "attention_mask_chosen": [],
+            "input_ids_rejected": [],
+            "attention_mask_rejected": [],
+        }
+        for prompt, chosen, rejected in zip(examples["prompt"], examples["chosen"], examples["rejected"]):
+            tokenized_chosen = tokenizer(
+                prompt + "\n\n" + chosen, padding="max_length", truncation=True, max_length=args.seq_length
+            )
+            tokenized_rejected = tokenizer(
+                prompt + "\n\n" + rejected, padding="max_length", truncation=True, max_length=args.seq_length
```
mnoukhov, line 200: I found that padding to `max_length` isn't necessary and can make performance very slightly slower, I tend to get better results without padding here (and just letting the data collator do padding)

mnoukhov (commented):
some small implementation details

examples/scripts/reward_trainer_summarize.py
vwxyzjn: Let's try that!

examples/scripts/reward_trainer_summarize.py
vwxyzjn: Done

examples/scripts/reward_trainer_summarize.py
vwxyzjn: probably just keep as is for now

examples/scripts/reward_trainer_summarize.py
vwxyzjn: > I tend to get better results

Do you mean 1) faster training or 2) higher validation accuracy? These would mean different things 👀

mnoukhov: Generally faster training. Maximum accuracy seems unchanged

PR #889, opened 2023-10-18
[`NEFTune`] Make use of forward hooks instead

younesbelkada:
As per title, this looks indeed much cleaner than monkey patching the forward pass of the embedding layer

No changes on the API are required

Stronger checks added on the CI 

cc @BenjaminBossan @lewtun @lvwerra

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -580,6 +580,8 @@ class SFTTrainerTester(unittest.TestCase):
                 packing=True,
             )
 
+            trainer.model = trainer._activate_neftune(trainer.model)
+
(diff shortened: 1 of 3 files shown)


BenjaminBossan (commented):
Looks very good.

I have one small concern (would be the same without this PR): If a user never calls `train`, the hook would never be removed, right? Is that a situation that could surprise users or is it not realistic? As it's only applied in training mode, maybe it's fine, I just wanted to bring it up.

To address this, would it be possible to call `_activate_neftune` inside of `train` instead of `__init__` or is it required to apply it earlier than `train`?

PR #890, opened 2023-10-18
Remove duplicate key in `reward_modeling.py`

vwxyzjn:
To make `reward_modeling.py` runnable again


younesbelkada (approved):
makes sense, thanks!

PR #895, opened 2023-10-19
Typo fixed: "scores" should be "score"

jangmino:
## Description

For XXXForSequenceClassification, we should define the correct classification head module name:
modules_to_save=["score"]


jangmino:
This pull request is too misc. So... it should be closed.

PR #904, opened 2023-10-22
[NEFTune] Replace absolute noise with proportional noise

jackstaff:
Generally speaking, using proportional noise is better than absolute noise for embedding layers.

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -754,7 +754,7 @@ def neftune_forward(self, input: torch.Tensor):
     embeddings = self._trl_old_forward(input)
 
     if self.training:
-        noise = (2*torch.rand_like(embeddings)-1)*self.neftune_noise_alpha
+        noise = torch.zeros_like(embeddings).uniform_(-self.neftune_noise_alpha, self.neftune_noise_alpha)
         embeddings = embeddings*(1+noise)
 
     return embeddings
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -751,13 +751,10 @@ def neftune_forward(self, input: torch.Tensor):
         noise_alpha (`float`):
             The noise alpha value to use for the NEFTune forward pass.
     """
-    embeddings = torch.nn.functional.embedding(
-        input, self.weight, self.padding_idx, self.max_norm, self.norm_type, self.scale_grad_by_freq, self.sparse
-    )
+    embeddings = self._trl_old_forward(input)
 
     if self.training:
-        dims = torch.tensor(embeddings.size(1) * embeddings.size(2))
-        mag_norm = self.neftune_noise_alpha / torch.sqrt(dims)
-        embeddings = embeddings + torch.zeros_like(embeddings).uniform_(-mag_norm, mag_norm)
+        noise = (2*torch.rand_like(embeddings)-1)*self.neftune_noise_alpha
+        embeddings = embeddings*(1+noise)
 
     return embeddings


lvwerra:
I only saw the absolute approach discussed in the paper - do you have evidence that proportional noise is better?

younesbelkada:
Hey @jackstaff 
Note that https://github.com/huggingface/trl/pull/889 being merged, we now swtiched to using forward hooks instead of patching the forward pass. I also second what @lvwerra said

jackstaff:
@lvwerra @younesbelkada 

No, I have no proof that this is better, I'm just relying on my imagination.
When I saw the NEFTune paper, my instinct asked me, why isn't it called "direct data augmentation"? Data enhancement is to generate multiple copies of data with some degree of distortion for training. The data enters the word segmenter and enters the embedding vector space... And NEFTune directly distorts the data in the embedding vector space... Based on this The imaginary understanding in my mind, I think this Alpha hyperparameter should be twisted around the input data, rather than purely random numbers.
The final effect is easier to understand and the parameters are better adjusted (I only make a simple comparison of individual models and cannot make a final judgment on whether it is better)

Because it will be troublesome to modify the semantics of the API after it is released, maybe it can support both at the same time? For reference only, you decide.

lvwerra:
Hi @jackstaff, happy to consider it if you can provide some experiments that show that it performs better than the current approach. Without any evidence I am reluctant to add it as an option, as it might even deteriorate the performance for users.

jackstaff:
Hi @lvwerra ,
    I'll try to compare the performance differences myself, but due to my limited resources it's impossible to draw a complete conclusion. This Pull request will be closed at the same time, thank you

jackstaff:
I created repository for verify:  [Direct Data Augmentation](https://github.com/jackstaff/dda)

Just for the record:
----------------------------------------------------------------------------------------------------------------------
From:  "Neel Jain"<[njain17@umd.edu](mailto:njain17@umd.edu)>;
Date:  Thu, Oct 26, 2023 09:27 AM
To:  "reco"<[reco@jackstaff.org](mailto:reco@jackstaff.org)>;
Subject:  Re: Fw:Re: [huggingface/trl] [NEFTune] Replace absolute noise with proportional noise (PR #904)
 
Hello Reco,

This is an interesting idea and worth trying. We used our scaling rules from FreeLB, which is robust optimization technique. I think your idea might be good, but it’s hard to know without empirical evidence.

Best,
Neel Jain

On Wed, Oct 25, 2023 at 4:50 AM reco <[reco@jackstaff.org](mailto:reco@jackstaff.org)> wrote:
Dear Neel Jain,

I hope this email finds you well.

I recently read the relevant information on NEFTune, and I was very impressed by its conceptual brilliance.

In particular, I was wondering if you have considered setting the neftune_alpha parameter to proportional noise. This could be done by simply changing the following line of code in train.py:
FROM:
        data['inputs_embeds'] = embeds_init + delta
TO:
       data['inputs_embeds'] = embeds_init*(1+proportion_delta)

This would ensure that the added noise is a perturbation of the input data, with the maximum amplitude of each data perturbation controlled by the proportional hyperparameter. This approach, in my opinion, offers several advantages over the absolute value-based hyperparameter setting used in the paper:

It is more intuitive and easier to understand, as it can be interpreted as a deviation from the input data.
It is more robust to changes in the input data, as the maximum amplitude of the noise is always relative to the input data.
It may allow for finer-grained control over the amount of noise added to the data, as the proportional hyperparameter can be adjusted in smaller increments.
(By adjusting the neftune_alpha parameter from small to large, we can even see the process of the model from overfitting to underfitting.)

I have not had the opportunity to test this approach myself, but I am curious to see how it performs in practice. I would be grateful for any feedback you have on this idea.

Thank you for your time and consideration.


Best Regards.
Reco.


Code Reference:
FROM:
        dims = torch.tensor(embeds_init.size(1) * embeds_init.size(2))
        mag_norm = args.neftune_alpha / torch.sqrt(dims)
        embeds= embeds_init+ torch.zeros_like(embeds_init).uniform_(-mag_norm, mag_norm)

TO:
        embeds= embeds_init*(1+ torch.zeros_like(embeds_init).uniform_(-args.neftune_alpha, args.neftune_alpha))
 
------------------ Original ------------------

PR #910, opened 2023-10-23
[Update reward_trainer.py] append PeftSavingCallback if callbacks is not None

zuoxingdong:
There are some cases in which users may specify some callbacks. Then `PeftSavingCallback` should be appended. The current implementation will ignore it in this case.


lvwerra (approved):
Looks good to me, I'll let @younesbelkada have a look as well :)

younesbelkada (approved):
Makes sense, thanks!

Issue #891, opened 2023-10-18
`reward_modeling.py` errors on multi-GPU training 


vwxyzjn:
To repro `accelerate launch --num_processes 8  examples/scripts/reward_modeling.py`

https://wandb.ai/costa-huang/huggingface/runs/ordvc1sq/logs?workspace=user-costa-huang

```
  File "examples/scripts/reward_trainer_summarize.py", line 165, in <module>
    trainer.train()
  File "/admin/home/costa/.pyenv/versions/3.8.11/envs/trl/lib/python3.8/site-packages/transformers/trainer.py", line 1506, in train
    return inner_training_loop(
  File "/admin/home/costa/.pyenv/versions/3.8.11/envs/trl/lib/python3.8/site-packages/transformers/trainer.py", line 1801, in _inner_training_loop
    tr_loss_step = self.training_step(model, inputs)
  File "/admin/home/costa/.pyenv/versions/3.8.11/envs/trl/lib/python3.8/site-packages/transformers/trainer.py", line 2659, in training_step
    self.accelerator.backward(loss)
  File "/admin/home/costa/.pyenv/versions/3.8.11/envs/trl/lib/python3.8/site-packages/accelerate/accelerator.py", line 1923, in backward
    loss.backward(**kwargs)
  File "/admin/home/costa/.pyenv/versions/3.8.11/envs/trl/lib/python3.8/site-packages/torch/_tensor.py", line 487, in backward
    torch.autograd.backward(
  File "/admin/home/costa/.pyenv/versions/3.8.11/envs/trl/lib/python3.8/site-packages/torch/autograd/__init__.py", line 200, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
  File "/admin/home/costa/.pyenv/versions/3.8.11/envs/trl/lib/python3.8/site-packages/torch/autograd/function.py", line 274, in apply
Traceback (most recent call last):
  File "examples/scripts/reward_trainer_summarize.py", line 165, in <module>
    return user_fn(self, *args)
  File "/admin/home/costa/.pyenv/versions/3.8.11/envs/trl/lib/python3.8/site-packages/torch/utils/checkpoint.py", line 157, in backward
    torch.autograd.backward(outputs_with_grad, args_with_grad)
  File "/admin/home/costa/.pyenv/versions/3.8.11/envs/trl/lib/python3.8/site-packages/torch/autograd/__init__.py", line 200, in backward
    trainer.train()
[...]

Issue #835, opened 2023-10-05
[RewardTrainer] Enable gradient checkpointing for all multi-GPU training modes


lewtun:
We currently have a few issues like https://github.com/huggingface/trl/issues/831 and https://github.com/huggingface/trl/issues/480 where gradient checkpointing + DDP does not work with the `RewardTrainer`.

Let's use this issue to collect the various training modes we'd like to support and track the status of their fixes. Ideally these would all be treated as integration tests so we know if there's regressions in future.

| Mode | Supported? |
| :---: | :---: |
| LoRA (unquantized) | ❌ |
| LoRA (4-bit) | ❌ |
| LoRA (8-bit) | ❌ |
| DDP | ❌ |
| ZeRO-1 | ✅ |
| ZeRO-2 | ❌ |
| ZeRO-3 | ❌ |

More details on each mode are provided below.

## 1. LoRA without quantization

Currently, this mode gives a warning that gradients are `None` on the inputs (i.e. model doesn't learn):

```
/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/torch/utils/checkpoint.py:31: UserWarning: None of the inputs have requires_grad=True. Gradients will be None
  warnings.warn("None of the inputs have requires_grad=True. Gradients will be None")
```

Command to reproduce:

```shell
ACCELERATE_LOG_LEVEL=info TRANSFORMERS_VERBOSITY=info accelerate launch --config_file=examples/accelerate_configs/multi_gpu.yaml examples/scripts/reward_trainer.py --use_peft True --gradient_checkpointing=True --gradient_accumulation_steps 1
```

## 2. LoRA with 4-bit or 8-bit quantization

```
Traceback (most recent call last):
  File "/fsx/lewis/git/trl/examples/scripts/reward_trainer.py", line 168, in <module>
    trainer.train()
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/transformers/trainer.py", line 1591, in train
    return inner_training_loop(
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/transformers/trainer.py", line 1892, in _inner_training_loop
    tr_loss_step = self.training_step(model, inputs)
  File "/fsx/lewis/miniconda/envs/trl/lib/python3.10/site-packages/transformers/trainer.py", line 2787, in training_step
    self.accelerator.backward(loss)
[...]

PR #912, opened 2023-10-24
[`core` / `DDP`] Fix RM trainer + DDP + quantization + propagate `gradient_checkpointing_kwargs` in SFT & DPO

younesbelkada:
Needs : https://github.com/huggingface/peft/pull/1036 / https://github.com/huggingface/transformers/pull/27020 

Fixes: https://github.com/huggingface/trl/issues/891
Fixes: https://github.com/huggingface/trl/issues/835

To avoid issues with PEFT + DDP, we need to call the gradient checkpointing method with `use_reentrant=False` that you can pass over the argument `gradient_checkpointing_kwargs` directly in trainer with https://github.com/huggingface/transformers/pull/27020 .

For users that do not have the correct transformers version they need to update transformers with the correct version to get that feature, otherwise `gradient_checkpointing_kwargs` will get ignored.

cc @lvwerra

### DIFF
--- a/examples/scripts/reward_modeling.py
+++ b/examples/scripts/reward_modeling.py
@@ -47,18 +47,18 @@ class ScriptArguments:
     reward_config: RewardConfig = field(
         default_factory=lambda: RewardConfig(
             output_dir="output",
-            per_device_train_batch_size=64,
+            per_device_train_batch_size=1,
             num_train_epochs=1,
             gradient_accumulation_steps=16,
             gradient_checkpointing=True,
-            gradient_checkpointing_kwargs={"use_reentrant": True},
+            gradient_checkpointing_kwargs={"use_reentrant": False},
             learning_rate=1.41e-5,
             report_to="tensorboard",
             remove_unused_columns=False,
             optim="adamw_torch",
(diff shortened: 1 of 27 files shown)


examples/scripts/reward_modeling.py
```diff
@@ -103,7 +103,7 @@ class ScriptArguments:
 
 # Step 2: Load the dataset and pre-process it
 tokenizer = AutoTokenizer.from_pretrained(args.model_name)
-train_dataset = load_dataset(args.dataset_name, split="train")
+train_dataset = load_dataset(args.dataset_name, split="train[:50]")
```
vwxyzjn, line 106: Could this be not hardcoded?

vwxyzjn (commented):
LG! One minor comment.

examples/scripts/reward_modeling.py
```diff
@@ -98,7 +99,7 @@ class ScriptArguments:
...
```
younesbelkada, line 102:
```suggestion
train_dataset = load_dataset(args.dataset_name, split="train")
```
I used that for testing purpose, will remove it!

PR #913, opened 2023-10-24
deactivate MacOS CI

lvwerra:
Since it is super flaky recently let's deactivate it until we find a better and more stable solution.


younesbelkada (approved):
Makes sense, thanks!

Issue #914, opened 2023-10-24
[BUG] Running examples/scripts/dpo.py crashes


JeanKaddour:
# Error
Running `examples/scripts/dpo.py` with no changes to the args crashes with the following error message 

```
 File "/home/jean/trl/examples/scripts/dpo.py", line 170, in <module>
    dpo_trainer.train()
  File "/home/jean/anaconda3/envs/rlhf/lib/python3.10/site-packages/transformers/trainer.py", line 1591, in train
    return inner_training_loop(
  File "/home/jean/anaconda3/envs/rlhf/lib/python3.10/site-packages/transformers/trainer.py", line 1984, in _inner_training_loop
    self._maybe_log_save_evaluate(tr_loss, model, trial, epoch, ignore_keys_for_eval)
  File "/home/jean/anaconda3/envs/rlhf/lib/python3.10/site-packages/transformers/trainer.py", line 2339, in _maybe_log_save_evaluate
    self._save_checkpoint(model, trial, metrics=metrics)
  File "/home/jean/anaconda3/envs/rlhf/lib/python3.10/site-packages/transformers/trainer.py", line 2471, in _save_checkpoint
    self.state.save_to_json(os.path.join(output_dir, TRAINER_STATE_NAME))
  File "/home/jean/anaconda3/envs/rlhf/lib/python3.10/site-packages/transformers/trainer_callback.py", line 106, in save_to_json
    json_string = json.dumps(dataclasses.asdict(self), indent=2, sort_keys=True) + "\n"
  File "/home/jean/anaconda3/envs/rlhf/lib/python3.10/dataclasses.py", line 1238, in asdict
    return _asdict_inner(obj, dict_factory)
  File "/home/jean/anaconda3/envs/rlhf/lib/python3.10/dataclasses.py", line 1245, in _asdict_inner
    value = _asdict_inner(getattr(obj, f.name), dict_factory)
  File "/home/jean/anaconda3/envs/rlhf/lib/python3.10/dataclasses.py", line 1273, in _asdict_inner
    return type(obj)(_asdict_inner(v, dict_factory) for v in obj)
  File "/home/jean/anaconda3/envs/rlhf/lib/python3.10/dataclasses.py", line 1273, in <genexpr>
    return type(obj)(_asdict_inner(v, dict_factory) for v in obj)
  File "/home/jean/anaconda3/envs/rlhf/lib/python3.10/dataclasses.py", line 1275, in _asdict_inner
[...]

PR #919, opened 2023-10-27
hotfix for dpo trainer

mnoukhov:
addresses #914

DPO trainer logs a `wandb.Table` which gets added to the trainer's `state.log_history` and then can't be converted to a json when the trainer is trying to save the state.

This removes the `wandb.Table` from the log history, in a kind of ugly way.


lvwerra (approved):

lvwerra:
Thanks a lot for the fix. We can merge when tests pass. @kashif for visibility.

Issue #923, opened 2023-10-30
NEFTune failed when using accelerate


SingL3:
I have launch a training mission on 2 nodes of A100 80G with the following config:
```yaml
compute_environment: LOCAL_MACHINE                                                                                             
debug: false                                                                                                                   
distributed_type: FSDP
downcast_bf16: 'no'
fsdp_config:
  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
  fsdp_backward_prefetch_policy: BACKWARD_PRE
  fsdp_forward_prefetch: true
  fsdp_offload_params: false
  fsdp_sharding_strategy: 1
  fsdp_state_dict_type: SHARDED_STATE_DICT
  fsdp_sync_module_states: true
  fsdp_use_orig_params: true
machine_rank: 0
main_training_function: main
mixed_precision: 'bf16'
num_machines: 1
num_processes: 8
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```
```bash
LAUNCHER="accelerate launch \
    --config_file configs/fsdp_config.yaml \
    --main_process_ip $MASTER_ADDR \
    --main_process_port $MASTER_PORT \
    --machine_rank $RANK \
    --num_processes $NUM_PROCESSES \
    --num_machines $NNODES \
```
Trainer:
```python
from typing import Optional, Union, Callable, Dict, List, Tuple
from functools import wraps

import torch
import torch.nn as nn
from torch.utils.data import Dataset

from transformers import (
        Trainer,
        PreTrainedModel,
        TrainingArguments,
        DataCollator,
        PreTrainedTokenizerBase,
        TrainerCallback,
        EvalPrediction
    )
from peft import PeftModel

def neftune_post_forward_hook(module, input, output):
    """
    Implements the NEFTune forward pass for the model using forward hooks. Note this works only for
    torch.nn.Embedding layers. This method is slightly adapted from the original source code
    that can be found here: https://github.com/neelsjain/NEFTune
    Simply add it to your model as follows:
    ```python
    model = ...
    model.embed_tokens.neftune_noise_alpha = 0.1
    model.embed_tokens.register_forward_hook(neftune_post_forward_hook)
    ```
    Args:
        module (`torch.nn.Module`):
            The embedding module where the hook is attached. Note that you need to set
            `module.neftune_noise_alpha` to the desired noise alpha value.
        input (`torch.Tensor`):
[...]

PR #924, opened 2023-10-30
Generalize NEFTune for FSDP, DDP, ...

younesbelkada:
# What does this PR do?

Fixes https://github.com/huggingface/trl/issues/923

In fact one always needs to unwrap the model before adding / removing hooks from the embeddings otherwise `get_input_embeddings()` will fail

cc @lvwerra


lvwerra (approved):
Thanks, looks good. Ideally we would test this properly at some point.

PR #929, opened 2023-10-31
[DPO] Add reward ratio to reduce randomness

gaetanlop:
Hello,

Approaches such as [RSO](https://huggingface.co/papers/2309.06657) directly sample from the policy to generate preference pairs based on reward model scores. For some pairs, the reward model might be sure of its decision (ie high difference between chosen reward and rejected reward) but for some others, it might be the opposite (low difference between chosen and rejected reward). 

This PR adds a `reduce_randomness` parameter to the DPO Trainer to try to account for the above issue. Setting this parameter to True will weigh the impact of each sample in the loss by how sure the reward model is in its preference choice (see formula below).

```py
torch.sigmoid(chosen_reward-rejected_reward)
```

 I think it was first introduced in [P3O paper](https://arxiv.org/pdf/2310.00212.pdf).

cc @kashif, @younesbelkada, @tomaarsen


lvwerra:
Hi @gaetanlop, thanks a lot for the PR! Did you run some experiments to verify that the approach improves training? Generally we are happy to add such features given there is some evidence that they improve performance.

Issue #878, opened 2023-10-16
DPO training produces nan for rewards


kangqi-ni:
I am using DPO trainer to train a vicuna-based fine-tuned model on a custom dataset. However, after a few steps, the trainer started to output nan for rewards. Did everyone run into similar issues or know how to fix this? 

Edit: I tried using different models or use a public dataset "lvwerra/stack-exchange-paired". I still ran into nan values starting on the second step.

![Screenshot from 2023-10-15 19-48-23](https://github.com/huggingface/trl/assets/64222208/863e4252-b488-4128-b8cb-8f935a39f681)

lvwerra:
cc @kashif

alex-ht:
I have the same problem with fp16, but not with bf16.

fe1ixxu:
The same issue with fp16

worden1:
Same issue with fp16 on V100, tried fp32 and load reference model with float32 then seems the nan fixed.

younesbelkada:
I think the fix is to upcast the logits in fp32 to avoid underflow & overflow here: https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L405-L408 Since the rewards are never used for backprop it should be totally fine. Just made: https://github.com/huggingface/trl/pull/942

alex-ht:
> I think the fix is to upcast the logits in fp32 to avoid underflow & overflow here: https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L405-L408 Since the rewards are never used for backprop it should be totally fine. Just made: #942

Should we convert logits to fp32 before applying logsoftmax? Megatron-LM converts it before calculating cross entropy.
https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/model/gpt_model.py#L36
https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L437

alex-ht:
All logps are expected to be in torch.FloatTensor, so why upcast them to fp32 again?

kashif:
I have removed the line that cast the `logits` to `float32` unnecessarily and have tested the `dpo.py` script in both `float16` and `bfloat16` mode so I believe that should fix the issue... @kangqi-ni can you tell me how to reproduce your error exactly?

OmarMohammed88:
@younesbelkada  i tried your solution but it still produce nan values after set of steps

younesbelkada:
[...]

PR #942, opened 2023-11-02
Upcast log probs in fp32 to avoid `NaN` reward

younesbelkada:
Fixes https://github.com/huggingface/trl/issues/878

As stated by the contributors, it seems the NaN issues disappears when training the model in bf16 / fp32 - hence I believe the fix is to upcast the logprobs in fp32 before computing the rewards to avoid underflow/overflow issues

This should have no effect on training as the reward is not used for backprop

cc @lvwerra

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -402,8 +402,8 @@ class DPOTrainer(Trainer):
         else:
             raise ValueError(f"Unknown loss type: {self.loss_type}. Should be one of ['sigmoid', 'hinge']")
 
-        chosen_rewards = self.beta * (policy_chosen_logps - reference_chosen_logps).detach()
-        rejected_rewards = self.beta * (policy_rejected_logps - reference_rejected_logps).detach()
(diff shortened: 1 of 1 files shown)


younesbelkada:
@lewtun what do you think?

lewtun (commented):
Seems like a sensible change to me @younesbelkada - thanks for adding it! Before we merge, can we create an integration / unit test that reproduces the issue in indeed just for `fp16` and `bf16` isn't affected?

I realise our CI doesn't currently run on GPU, but even having a slow test we can check manually would be good IMO

younesbelkada:
Yeah makes sense, let me try to do that

lvwerra (approved):

lvwerra:
also cc @kashif

kashif:
I believe I fixed this issue by removing the cast of the logits to `float()` now the logits are in the actual type of the model: https://github.com/huggingface/trl/pull/885/files#diff-5bbdb5d54108f2162b47bc54dc23c7b8e7744d2941118e60a44c161a4acc0ee8L532-R860

lvwerra:
Then maybe we can close the issue for now until someone reports the issue again?

kashif:
yes lets close it for now

PR #945, opened 2023-11-02
[`SFTrainer`] Allow users to pass a simple `formatting_func` when `packing=False` at their own risk

### LINKED ISSUE #944: API - Unconsistant API for `formatting_func` between packed and not packed
Currently users need to provide different `formatting_func` depending on if the dataset is `packing=True` or no. In the first case users can pass a method that takes a dict as input and return a simple string, for the non `packing` case users need to write a method that takes a dict and returns a list - API-wise this is very confusing

cc @lvwerra

younesbelkada:
Fixes #944 

Currently the API for using `formatting_func` is not consistent across `packing=True` and `packing=False`. The reason behind forcing users to use a function that processes a batched input and returns a list is because for non-packed dataset we call `dataset.map(xxx, batched=True)` under the hood, that returns a dict of list for each element.

As changing the API will be breaking for users that already use this setup, I propose to offer users the possibility to use a simpler prompt formatting function at their own risk (slower pre-tokenization step) - by passing `process_batched_dataset=False` at the trainer's init. 

Updated the documentation with more clarification and added a test

cc @lvwerra

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import dataclasses
-import inspect
 import warnings
 from functools import wraps
 from typing import Callable, Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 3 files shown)


vwxyzjn (approved):
Hey @younesbelkada nice PR! This makes sense to me and could make processing custom dataset more convenient!

trl/trainer/sft_trainer.py
```diff
@@ -357,7 +363,7 @@ def tokenize(element):
             )
 
             if use_formatting_func and not self._dataset_sanity_checked:
-                if not isinstance(formatting_func(element), list):
+                if not isinstance(formatting_func(element), list) and self.process_batched_dataset:
                     raise ValueError(
                         "The `formatting_func` should return a list of processed strings since it can lead to silent bugs."
```
lvwerra, line 368:
```suggestion
                        "The `formatting_func` should return a list of processed strings."
```

lvwerra (commented):
I think the formatting function should always have the same format (irrespective of batched or not batched) and we should take care of it under the hood. Why not keep it a `List[str] -> List[str]` format?

PR #951, opened 2023-11-03
Adds model kwargs to SFT and DPO trainers

edbeeching:
- Adds model as string support to the DPOTrainer.
- Adds model_kwargs optional args to SFT and DPO trainers.
- Refactors some of the SFT model instantiation logic as I thought the implementation was overly complicated.

This will greatly simplify the alignment handbook so it would be great to have it in the next trl release.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -143,16 +143,16 @@ class DPOTrainer(Trainer):
         disable_dropout: bool = True,
         generate_during_eval: bool = False,
         compute_metrics: Optional[Callable[[EvalLoopOutput], Dict]] = None,
-        model_kwargs: Optional[Dict] = None,
-        ref_model_kwargs: Optional[Dict] = None,
+        model_init_kwargs: Optional[Dict] = None,
+        ref_model_init_kwargs: Optional[Dict] = None,
     ):
-        if model_kwargs is None:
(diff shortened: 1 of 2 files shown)


younesbelkada (commented):
LGTM ! Can you just add a simple check for users that pass both a valid model and model_kwargs to make sure we throw an appropriate error telling users that passing model_kwargs with non-str model is not supported? 🙏 Thanks!

trl/trainer/dpo_trainer.py
```diff
@@ -100,12 +107,17 @@ class DPOTrainer(Trainer):
         compute_metrics (`Callable[[EvalPrediction], Dict]`, *optional*):
             The function to use to compute the metrics. Must take a `EvalPrediction` and return
             a dictionary string to metric values.
+        model_kwargs: (`Optional[Dict]`, *optional*):
```
lvwerra, line 110: I'd be in favour of calling it `model_init_kwargs` to make it clear where they are used.

lvwerra (commented):
Looks generally good to me, just a small nit regarding naming.

trl/trainer/dpo_trainer.py
edbeeching: done

younesbelkada (approved):
Looks great thank you for this @edbeeching !

Issue #952, opened 2023-11-03
[DPO] should we refactor to be more in line with Trainer?


mnoukhov:
@kashif is there a reason why the DPO trainer has such a strange setup using `store_metrics` to forcibly store metrics during the prediction step and then delete them whenever `log` is called?  This feels quite brittle and from what I've seen, the regular `Trainer` uses a different setup. `Trainer.evaluate` and `predict` generally returns the metrics and uses `compute_metrics` to calculate things.

Would it make sense to refactor to be more in line with this?

TODO
- change to logging with `compute_metrics` #954 
- make eval use eval_dataloader instead of random batches
- switch to `Seq2SeqTrainer` so we can use `predict_with_generate` instead of `generate_during_eval`

mnoukhov:
I'll get on it and start a draft

mnoukhov:
Still being addressed in a follow-up to #954

mnoukhov:
Still being worked on

younesbelkada:
@mnoukhov let us know with @lvwerra @kashif if you need any help !

PR #954, opened 2023-11-03
[DPO] Refactor eval logging of dpo trainer

mnoukhov:
Address part 1 of #952 by removing `store_metrics` and switching to more standard `compute_metrics` letting `Trainer` handle logging

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -294,6 +294,8 @@ class DPOTrainer(Trainer):
                 )
 
             self.use_dpo_data_collator = True
+
+            args.label_names = ["chosen_labels", "rejected_labels"]
         else:
             self.use_dpo_data_collator = False
 
@@ -685,7 +687,7 @@ class DPOTrainer(Trainer):
         return initial_output
 
 
-def compute_dpo_metrics(self, eval_preds: EvalPrediction):
+def compute_dpo_metrics(eval_preds: EvalPrediction):
     preds, labels = eval_preds
     chosen_rewards = preds["chosen_rewards"]
     rejected_rewards = preds["rejected_rewards"]
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -15,6 +15,7 @@
 import inspect
 import random
 import warnings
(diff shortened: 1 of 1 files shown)


lvwerra:
Thanks for working on this! Just ping me whenever you would like to run the CI.

mnoukhov:
@lvwerra, I completed this part of the refactor, can you run the CI?
 
@kashif as part of this, I removed logging all the `logits` since it seemed a bit excessive to pass around the `batch x seq_len x vocab size` array and taking the `mean` didn't feel like a meaningful thing to log since we already have `logps`. Let me know if you'd like this reversed

kashif:
@mnoukhov are you also logging sample outputs as well here?

mnoukhov:
@kashif I didn't change sample output logging in this PR, so it's still there. I was going to address the sample logging issue (2 in #952) in a second PR to keep things managable

kashif:
so this PR LGTM! feed free to merge

kashif:
@mnoukhov so folks really want the additional metrics to be logged during training which isn't possible at the moment, so lets revert this for now and continue this refactoring to figure that out.

mnoukhov:
I don't think this change prevents additional metrics from being logged. It feels like it actually makes it easier to log

kashif:
Ok so let's try to add the previous behavior if possible... I couldn't figure it out at first glance... I think the trainer's training loop doesn't seem to have a way to do it... can you kindly confirm?

mnoukhov:
Basically what #1046 wants is to log metrics during training, which is not how `Trainer` usually works but it is a common ask e.g. https://discuss.huggingface.co/t/logging-training-accuracy-using-trainer-class/5524/3. The reverted code does not do this correctly.

AFAIK the problem with trying to put `self.log` into `compute_loss` is it doesn't really make sense whenever you use multi-gpu or gradient accumulation. In both cases you're calling `compute_loss` more than once per step and the previous code just logged the metrics for one subset of that batch (e.g. the last minibatch in gradient accumulation or one gpu's batch). @lewtun, is this right?

If you want metrics for the training set, I would recommend doing `evaluate` on the training set any time it is also done on the test set, using a callback.

PR #957, opened 2023-11-05
fix: dpo trainer ds config

mengban:
ref_model and model shouldnt share the same ds config, so we shouldnt modify the ds config directly. or else, it will cause sth wrong when init deepspeed engine


younesbelkada (approved):
This should be good ! Thanks!

younesbelkada:
@mengban before merging can you run the styling checks ? `make precommit`

mengban:
> @mengban before merging can you run the styling checks ? `make precommit`

ok, I have submitted a new commit to solve the problem of import sorting.

younesbelkada (approved):
Thanks again!

PR #974, opened 2023-11-09
[`Tests`] Add non optional packages tests

younesbelkada:
# What does this PR do?

This PR adds in the test workflow a workflow to test TRL testing suite without optional packages as we were not able to flag some minor issues that can occur during releases, e.g. https://github.com/huggingface/trl/pull/967

cc @lvwerra


lvwerra (approved):
Looks good to me!

PR #982, opened 2023-11-10
[Multi-Adapter PPO] Fix and Refactor reward model adapter 

mnoukhov:
A simpler and cleaner version of #472 focused on the main issues

`examples/multi_adapter_ppo.py` fails with multi-gpu
1.  you can't call `model.compute_reward_score` if `model` is wrapped in `DDP`
- fix by using `unwrap_model`

2. running crashes with `RuntimeError: Expected to have finished reduction in the prior iteration before starting a new one....`
- Caused because the reward score is being calculated with gradient
- Add a `torch.no_grad()` and make sure that the reward adapter's `score` has `requires_grad=False`

in `modeling_base.py`, create the reward adapter before initializing the model and pass score into `init`
- makes adding the reward adapter a class method that can be extended
- makes use of `PeftModel.load_adapter` which includes a lot of useful logic

### DIFF
--- a/trl/models/modeling_base.py
+++ b/trl/models/modeling_base.py
@@ -29,7 +29,6 @@ from ..import_utils import is_peft_available, is_transformers_greater_than, is_x
 
 if is_peft_available():
     from peft import (
-        LoraConfig,
         PeftConfig,
         PeftModel,
         PeftModelForCausalLM,
@@ -38,7 +37,6 @@ if is_peft_available():
         get_peft_model,
         prepare_model_for_kbit_training,
     )
-    from peft.peft_model import set_peft_model_state_dict
 
 if is_transformers_greater_than("4.33.0"):
     from transformers.integrations.deepspeed import is_deepspeed_zero3_enabled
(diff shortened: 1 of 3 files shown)


trl/models/modeling_base.py
```diff
@@ -68,7 +66,7 @@ class PreTrainedModelWrapper(nn.Module):
             The list of arguments that are supported by the wrapper class.
     """
     transformers_parent_class = None
-    supported_args = None
+    supported_args = ("score_module", "supports_rm_adapter", "rm_adapter_name")
```
younesbelkada, line 69: Do you know if these are used? It seems we always overwrite them: https://github.com/huggingface/trl/blob/main/trl/models/modeling_value_head.py#L90

younesbelkada (commented):
Great clean up! Thanks a ton for working on this @mnoukhov !
I left only one question, what do you think?

trl/models/modeling_base.py
younesbelkada, line 69:
```suggestion
    supported_args = None
```

mnoukhov: I made this change so it would be a bit more obvious the args being passed in. I'm happy to change it back if you'd like

I actually don't think we use `supported_args` anywhere. I assumed it would be a future feature for some sort of argparsing. To be more in line with `transformers` we would probably want a `config` class instead of supported args anyways so it isn't a big deal either way.

younesbelkada: OK I see, thanks! 
Yes it is not a big deal, technically I have set it to `None` so that the `PreTrainedModelWrapper` can't be used as it is (I should have set that class as an abstract class). Yes it would be great if you can change this back then we can merge I think

mnoukhov: Done!

PR #995, opened 2023-11-14
SDXL support for DDPO 

kabachuha:
Should work, but need to test myself on an A100

### DIFF
--- a/examples/scripts/ddpo.py
+++ b/examples/scripts/ddpo.py
@@ -35,6 +35,10 @@ class ScriptArguments:
     """the pretrained model to use"""
     pretrained_revision: str = "main"
     """the pretrained model revision to use"""
+    sdxl: bool = False
+    """whether to use SDXL"""
+    sdxl_vae: str = "madebyollin/sdxl-vae-fp16-fix"
+    """the name of the pretrained SDXL VAE model to use"""
     hf_hub_model_id: str = "ddpo-finetuned-stable-diffusion"
     """HuggingFace repo to save model weights to"""
     hf_hub_aesthetic_model_id: str = "trl-lib/ddpo-aesthetic-predictor"
@@ -44,7 +48,6 @@ class ScriptArguments:
 
     ddpo_config: DDPOConfig = field(
         default_factory=lambda: DDPOConfig(
-            sdxl=False,
             num_epochs=200,
             train_gradient_accumulation_steps=1,
             sample_num_steps=50,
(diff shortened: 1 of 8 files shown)


younesbelkada:
Great work @kabachuha ! Let us know when this is ready for review, cc @sayakpaul

kabachuha:
It's ready, just need to spend a bit time to rent an A100 and make a test, maybe not with the aesthetic model, but with jpeg compressibility instead (aesthetic model's resolution is too low for sdxl)

kabachuha:
@younesbelkada this would be great! <3

younesbelkada (commented):
Thanks a lot! I just ran `python ddpo_sdxl_compressibility.py` and got:

```bash
, line 443, in check_inputs
    raise ValueError(f"`height` and `width` have to be divisible by 8 but are {height} and {width}.")
ValueError: `height` and `width` have to be divisible by 8 but are 1024 and 1.
```
Can you help guide me how to successfully run the script?

kabachuha:
Oops, forgot about the sdxl dims setting for the script 👀

Anyway, when I come home (in ~6 hours), I'm going to make a small run on remote to ensure that it fully works (with pipeline calls and rl cycles) and report back

younesbelkada:
Awesome thank you very much @kabachuha !

kabachuha:
The cloud has really bad internet today 😩. Still, I think I put the right params into the places in the last commit. Would it be possible for you @younesbelkada to test it again?

younesbelkada:
OK thanks @kabachuha ! let me know how it goes!

kabachuha:
Almost made it work (pipeline, parameters transfer). The only remaining part is the loss calculation. Hopefully, will be ready this weekend

kabachuha:
It kinda works, but there's the `Trying to backward through the graph a second time` error in the loss calculation function, need to think about it

kabachuha:
Well, it works with `retain_graph=True` in the trainer, but it OOMs even on a A100, so idk how to test it

AmericanPresidentJimmyCarter:
This should really be unet and text encoder agnostic. It's as relevent for SD 1.x and 2.x as it is for SDXL.

anonymous-atom:
Hey @kabachuha any updates about your script ? Does it works ? I want to try it with some loss function.

anonymous-atom:
Hi @kabachuha , I also tried to test the code from your ddpo-sdxl form. I ran the ddpo_sdxl_compressibilty.py file and it's using DefaultDDPOStableDiffusionPipeline instead of the XL DDPO pepeline you implemented.

Issue #960, opened 2023-11-06
[BUG] Gather Stat is not working for DeepSpeed?


allanj:
I guess the issue is mainly caused by this line:

https://github.com/huggingface/trl/blob/6ff0fac2c118250e1a14a0a793e96c5c3f654885/trl/trainer/ppo_trainer.py#L333

The `self.accelerator.distributed_type` is `DEEPSPEED` when I use DeepSpeed Zero stage 2 with `accelerate`.

It works when I change this line to
```python3
self.is_distributed = self.accelerator.distributed_type == "MULTI_GPU"  or self.accelerator.distributed_type == "DEEPSPEED"
```
But I think, to make this more general, we can try something like

```python3
self.is_distributed = self.accelerator.num_process > 1
```

Any suggestions?

seanexp:
looks like a valid fix to me! 👍

allanj:
maybe I will try to make a pull request for this. Also, I feel like there are some redundant for loop that we can avoid in the code

younesbelkada:
Hi everyone, yes this looks like a reasonable change, would you be happy to open a quick PR for it?

allanj:
> Hi everyone, yes this looks like a reasonable change, would you be happy to open a quick PR for it?

I have created the simple PR fix. Please help review: https://github.com/huggingface/trl/pull/997

PR #997, opened 2023-11-15
make distributed true for multiple process

allanj:
This PR is to fix the issue in https://github.com/huggingface/trl/issues/960#issuecomment-1810865206

The issue will cause "gather stat" not to work for other distributed scenarios other than `MULTI_GPU`.

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -330,7 +330,7 @@ class PPOTrainer(BaseTrainer):
         # In a distributed setup, only logging needs to be performed on the main process
         # check: https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html
         # or: https://discuss.pytorch.org/t/use-distributed-data-parallel-correctly/82500/11
-        self.is_distributed = self.accelerator.distributed_type == "MULTI_GPU"
+        self.is_distributed = self.accelerator.num_processes > 0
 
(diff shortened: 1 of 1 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -330,7 +330,7 @@ def __init__(
         # In a distributed setup, only logging needs to be performed on the main process
         # check: https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html
         # or: https://discuss.pytorch.org/t/use-distributed-data-parallel-correctly/82500/11
-        self.is_distributed = self.accelerator.distributed_type == "MULTI_GPU"
+        self.is_distributed = self.accelerator.num_processes > 0
```
younesbelkada, line 333:
```suggestion
        self.is_distributed = self.accelerator.num_processes > 1
```

younesbelkada (commented):
Thanks for the fix! I left one comment, what do you think?

trl/trainer/ppo_trainer.py
allanj: Oh, yes. That's correct. My bad

allanj:
It's kind of my first time to make PR here. So I guess one more review needed? @younesbelkada

younesbelkada:
No this is great! Merging!

younesbelkada (approved):

PR #1012, opened 2023-11-20
Update utils.py

ZihanWang314:
update compute_accuracy to deal with the cases where str_chosen and str_rej got the same scores, which is probably what the developers don't want

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -674,7 +674,7 @@ def compute_accuracy(eval_pred) -> Dict[str, float]:
         warnings.warn(
             "Some of the predictions are equal, which is probably not what you want. In our basic setting, we set the equal predictions to be wrong."
         )
-    predictions = predictions[:, 0] < predictions[:, 1]
+    predictions = np.argmax(predictions, axis=1)
 
     accuracy = np.array(predictions == labels, dtype=float).mean().item()
     return {"accuracy": accuracy}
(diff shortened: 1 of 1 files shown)


lvwerra:
Can you elaborate a bit more why this is bad and what bad side effects this has?

ZihanWang314:
When a model gives exactly same scores for the sentence to accept and the sentence to reject, the current evaluation metric will set this pair to be "correct". However, the above scenario is usually caused by the model having some problems in encoding sentences (e.g. over-quantized), or the two sentences exactly being the same, which we both don't expect. I think putting a warning here and defining this pair as "wrong" or "random" would be helpful for the developers to observe these problems.

lvwerra:
I see, thanks! In that case I'd prefer to just add a warning but leave the computation logic the same. Does that make sense?

ZihanWang314:
sure it makes sense. Thanks for your suggestions!

trl/trainer/utils.py
```diff
@@ -670,6 +670,10 @@ def compute_accuracy(eval_pred) -> Dict[str, float]:
     predictions, labels = eval_pred
     # Here, predictions is rewards_chosen and rewards_rejected.
     # We want to see how much of the time rewards_chosen > rewards_rejected.
+    if np.array(predictions[:, 0] == predictions[:, 1], dtype=float).sum() > 0:
+        warnings.warn(
+            "Some of the predictions are equal, which is probably not what you want. In our basic setting, we set the equal predictions to be wrong."
```
lvwerra, line 675: What do you think about updating the text to the following?

```suggestion
            f"There are {np.array(predictions[:, 0] == predictions[:, 1]).sum()} out of {len(predictions[:, 0])} instances where the predictions for both options are equal. As a consequence the accuracy can be misleading."
```

ZihanWang314: It seems good!

lvwerra (approved):

PR #1020, opened 2023-11-21
Remove duplicate data loading in rl_training.py

viethoangtranduong:
We load dataset twice, but in line 149 (new), we do  `ds = train_dataset.map` anyway


younesbelkada (approved):
Makes sense, thanks!

PR #1022, opened 2023-11-22
[DPO] IPO Training loss

kashif:
Implemented the IPO training loss from the paper:
A General Theoretical Paradigm to Understand Learning from Human Preferences
https://arxiv.org/pdf/2310.12036.pdf

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -464,24 +464,24 @@ class DPOTrainer(Trainer):
         reference_rejected_logps: torch.FloatTensor,
     ) -> Tuple[torch.FloatTensor, torch.FloatTensor, torch.FloatTensor]:
         """Compute the IPO loss for a batch of policy and reference model log probabilities.
+        The beta is the regularization parameter for the IPO loss, denoted by tau in the paper.
 
         Args:
             policy_chosen_logps: Log probabilities of the policy model for the chosen responses. Shape: (batch_size,)
             policy_rejected_logps: Log probabilities of the policy model for the rejected responses. Shape: (batch_size,)
             reference_chosen_logps: Log probabilities of the reference model for the chosen responses. Shape: (batch_size,)
             reference_rejected_logps: Log probabilities of the reference model for the rejected responses. Shape: (batch_size,)
-            beta: Regularization parameter for the IPO loss, denoted by tau in the paper.
 
         Returns:
             A tuple of three tensors: (losses, chosen_rewards, rejected_rewards).
-            The losses tensor contains the DPO loss for each example in the batch.
+            The losses tensor contains the IPO loss for each example in the batch.
             The chosen_rewards and rejected_rewards tensors contain the rewards for the chosen and rejected responses, respectively.
         """
-        pi_logratios = policy_chosen_logps - policy_rejected_logps
-        ref_logratios = reference_chosen_logps - reference_rejected_logps
+        pi_logratios = policy_chosen_logps + reference_rejected_logps
+        ref_logratios = policy_rejected_logps + reference_chosen_logps
 
         logits = pi_logratios - ref_logratios
-
+        # eqn (17) of the paper:
         losses = (logits - 1 / (2 * self.beta)) ** 2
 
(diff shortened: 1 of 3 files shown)


younesbelkada (approved):
Thanks for this great addition @kashif !

trl/trainer/dpo_trainer.py
```diff
@@ -422,13 +422,13 @@ def dpo_loss(
         reference_free: bool = False,
     ) -> Tuple[torch.FloatTensor, torch.FloatTensor, torch.FloatTensor]:
         """Compute the DPO loss for a batch of policy and reference model log probabilities.
+        The beta is a temperature parameter for the DPO loss, typically something in the range of 0.1 to 0.5. We ignore the reference model as beta -> 0.
```
lewtun, line 425: This comment is a bit cryptic since `beta` isn't part of the args. Would it be better to move this as a comment directly above the code, e.g. here: https://github.com/huggingface/trl/pull/1022/files#diff-5bbdb5d54108f2162b47bc54dc23c7b8e7744d2941118e60a44c161a4acc0ee8L454-L455

trl/trainer/dpo_trainer.py
```diff
@@ -422,13 +422,13 @@ def dpo_loss(
...
 
         Args:
             policy_chosen_logps: Log probabilities of the policy model for the chosen responses. Shape: (batch_size,)
             policy_rejected_logps: Log probabilities of the policy model for the rejected responses. Shape: (batch_size,)
             reference_chosen_logps: Log probabilities of the reference model for the chosen responses. Shape: (batch_size,)
             reference_rejected_logps: Log probabilities of the reference model for the rejected responses. Shape: (batch_size,)
-            beta: Temperature parameter for the DPO loss, typically something in the range of 0.1 to 0.5. We ignore the reference model as beta -> 0.
```
lewtun, line 431: Just for my understanding, `beta` is removed here since it was actually not part of the function signature?

trl/trainer/dpo_trainer.py
```diff
@@ -456,6 +456,39 @@ def dpo_loss(
...
 
+    def ipo_loss(
+        self,
+        policy_chosen_logps: torch.FloatTensor,
+        policy_rejected_logps: torch.FloatTensor,
+        reference_chosen_logps: torch.FloatTensor,
+        reference_rejected_logps: torch.FloatTensor,
+    ) -> Tuple[torch.FloatTensor, torch.FloatTensor, torch.FloatTensor]:
+        """Compute the IPO loss for a batch of policy and reference model log probabilities.
+        The beta is the regularization parameter for the IPO loss, denoted by tau in the paper.
```
lewtun, line 467: Similar comment as a above, I would move this to the code below where we define the equation. 

We could describe what `beta` is in the docs of the trainer attributes

trl/trainer/dpo_trainer.py
```diff
@@ -456,6 +456,39 @@ def dpo_loss(
...
+            A tuple of three tensors: (losses, chosen_rewards, rejected_rewards).
+            The losses tensor contains the IPO loss for each example in the batch.
+            The chosen_rewards and rejected_rewards tensors contain the rewards for the chosen and rejected responses, respectively.
+        """
+        pi_logratios = policy_chosen_logps + reference_rejected_logps
+        ref_logratios = policy_rejected_logps + reference_chosen_logps
+
+        logits = pi_logratios - ref_logratios
+        # eqn (17) of the paper:
+        losses = (logits - 1 / (2 * self.beta)) ** 2
```
lewtun, line 485: I checked the equations and this makes sense to me

lewtun (approved):
Thanks for adding this so quickly @kashif, it's very elegant that you can now just swap out the losses for each method!

I left a nit about where to mention `beta`, but otherwise this LGTM

trl/trainer/dpo_trainer.py
kashif: yes as it was not part of the argument... I'll move it to a comment in the code good idea!

trl/trainer/dpo_trainer.py
kashif: added to the docstring of the Trainer too

lvwerra (approved):
LGTM 🚀

Issue #746, opened 2023-09-07
SFTTrainer: DataLoader warnings when using Packing


pacman100:
1. Code: https://github.com/pacman100/DHS-LLM-Workshop/tree/main/code_assistant/training
2. command:
```
train.py \
--model_name "meta-llama/Llama-2-70b-chat-hf" \
--dataset_name "smangrul/code-chat-assistant-v1" \
--max_seq_len 2048 \
--max_steps 1000 \
--logging_steps 25 \
--eval_steps 200 \
--save_steps 500 \
--bf16 True \
--packing True \
--output_dir "temp/experiments/full-finetune-llama-chat-asst" \
--per_device_train_batch_size 1 \
--gradient_accumulation_steps 2 \
--dataset_text_field "content" \
--use_gradient_checkpointing True \
--learning_rate 5e-5  \
--lr_scheduler_type "cosine" \
--weight_decay 0.01 \
--warmup_ratio 0.03 \
--use_flash_attn True
"
```

3. logs:
```
{'loss': 0.9431, 'learning_rate': 4.936026311617316e-05, 'epoch': 0.29}
  11 {'loss': 0.8537, 'learning_rate': 4.882595527372152e-05, 'epoch': 0.37}
  12 {'loss': 0.6112, 'learning_rate': 4.813553074106761e-05, 'epoch': 0.44}
  13 {'loss': 0.5148, 'learning_rate': 4.7293513441455364e-05, 'epoch': 0.51}
  14 {'loss': 0.5078, 'learning_rate': 4.630542059139924e-05, 'epoch': 0.59}
  15  29%|██▊       | 2/7 [00:01<00:02,  1.86it/s]
  16 {'eval_loss': 0.9526195526123047, 'eval_runtime': 2.2525, 'eval_samples_per_second': 363.152, 'eval_steps_per_second': 3.108, 'epoch': 0.59}
  17 {'loss': 0.4727, 'learning_rate': 4.517772654979023e-05, 'epoch': 0.66}
  18 {'loss': 0.3182, 'learning_rate': 4.391782039544238e-05, 'epoch': 0.74}
  19 {'loss': 0.2365, 'learning_rate': 4.253395751104748e-05, 'epoch': 0.81}
  20 {'loss': 0.2251, 'learning_rate': 4.10352054907785e-05, 'epoch': 0.88}
  21 {'loss': 0.2237, 'learning_rate': 3.943138472597549e-05, 'epoch': 0.96}
[...]

PR #1027, opened 2023-11-23
Add `PackedIterableDataset`, `peft_module_casting_to_bf16` util method and `append_concat_token` flag

pacman100:
### What does this PR do?
1. Fixes https://github.com/huggingface/trl/issues/746 by introducing `PackedIterableDataset`
2. Inline with QLoRA code when using bf16 [here](https://github.com/artidoro/qlora/blob/main/qlora.py#L396-L405), added `peft_module_casting_to_bf16`.
3. Added `append_concat_token` flag so that datasets which already have eos_token at the end of each sample can skip appending the eos/concat token.

Details:
Issues with `ConstantLengthDataset`: 
1. Learning rate schedulers like Cosine, linear_decay ... will run improperly because the length of the dataset is set to the initial dataset while an epoch will have much fewer steps because of packing.
For example, I've the below code which uses the existing `ConstantLengthDataset` and epoch=1. The logs show that epoch only has 207 steps instead of the original 1360 steps due to packing. Because of this, cosine lr_scheduler doesn't go down to 0 at the end of the epoch.
Code: https://github.com/pacman100/DHS-LLM-Workshop/pull/14
Commnad:
```
cd /DHS-LLM-Workshop/chat_assistant/training

accelerate launch --config_file configs/fsdp_config.yaml \
train.py \
--model_name_or_path "meta-llama/Llama-2-7b-chat-hf" \
--dataset_name "smangrul/code-chat-assistant-v1" \
--max_seq_len 4096 \
--num_train_epochs 1 \
--save_strategy "epoch" \
--logging_steps 25 \
--eval_steps 100 \
--bf16 True \
--packing True \
--output_dir "experiments/full-finetune-llama7B-fullshard" \
--dataset_text_field "content" \
--learning_rate 5e-5  \
--lr_scheduler_type "cosine" \
--weight_decay 0.01 \
--warmup_ratio 0.03 \
--use_flash_attn True
```

Output logs:
```
15%|█▌        | 207/1360 [04:09<22:52,  1.19s/it]
```

wandb_plots:
Screenshot 2023-11-23 at 3 17 58 PM

So, epoch based training using `ConstantLengthDataset` is not preferable with decay based lr_scheduler and packing=True.

2. When `max_steps` is passed which has more steps than the initial `len(dataset)`, it results in https://github.com/huggingface/trl/issues/746. `PackedIterableDataset` is meant to fix this issue as it doesn't contain the length attribute and works as expected when passing `max_steps`.

### DIFF
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -449,3 +449,8 @@ Pay attention to the following best practices when training a model with that tr
 ## ConstantLengthDataset
 
 [[autodoc]] trainer.ConstantLengthDataset
+
+
+## PackedIterableDataset
+
+[[autodoc]] trainer.PackedIterableDataset
--- a/trl/trainer/__init__.py
+++ b/trl/trainer/__init__.py
@@ -20,9 +20,13 @@ from .utils import (
     AdaptiveKLController,
     FixedKLController,
     ConstantLengthDataset,
+    PackedIterableDataset,
     DataCollatorForCompletionOnlyLM,
     RunningMoments,
     disable_dropout_in_model,
+    peft_module_casting_to_bf16,
+    SaveDeepSpeedPeftModelCallback,
+    save_peft_deepspeed_ckpt,
 )
 
 # isort: on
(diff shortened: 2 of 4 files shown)


pacman100:
cc @younesbelkada, @lvwerra

pacman100:
Closing this in favour of https://github.com/huggingface/trl/pull/1110

PR #1035, opened 2023-11-26
[DPO] cDPO loss

kashif:
Add the cDPO loss.


trl/trainer/dpo_trainer.py
```diff
@@ -309,6 +312,7 @@ def make_inputs_require_grad(module, input, output):
         self.padding_value = padding_value
 
         self.beta = beta
+        self.label_smoothing = label_smoothing
```
lvwerra, line 315: Maybe add a warning if the parameter is set but IPO is used since then it has no effect.

PR #1048, opened 2023-12-01
Improve PreTrainedModelWrapper._get_current_device

billvsme:
If using ```dummy_accelerator = Accelerator()```, it will cause the ```ACCELERATE_USE_DEEPSPEED``` environment variable to be set in advance, resulting in the regeneration of ```DeepSpeedPlugin``` and causing a change in the type of ```trainer.accelerator.state.deepspeed_plugin.hf_ds_config```
 [source code: regeneration DeepSpeedPlugin](https://github.com/huggingface/accelerate/blob/3499cf25aa431ed3eeb7969f1d9b501ef20d0911/src/accelerate/accelerator.py#L284-L292)
[source code: HfTrainerDeepSpeedConfig  -> HfDeepSpeedConfig ](https://github.com/huggingface/accelerate/blob/3499cf25aa431ed3eeb7969f1d9b501ef20d0911/src/accelerate/utils/dataclasses.py#L600)

which may lead to errors similar to https://github.com/microsoft/DeepSpeed/issues/3733, https://github.com/hiyouga/LLaMA-Factory/issues/286. If this situation occurs, it will be difficult to identify the problem. It feels that using the ```LOCAL_RANK``` environment variable may be a better approach.

### DIFF
--- a/trl/models/modeling_base.py
+++ b/trl/models/modeling_base.py
@@ -18,7 +18,7 @@ from copy import deepcopy
 
 import torch
 import torch.nn as nn
-from accelerate import Accelerator
+from accelerate import PartialState
 from huggingface_hub import hf_hub_download
 from huggingface_hub.utils import EntryNotFoundError, HFValidationError, LocalEntryNotFoundError
 from safetensors.torch import load_file as safe_load_file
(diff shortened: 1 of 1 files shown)


lvwerra:
Maybe @muellerzr would know if this is a viable alternative? Otherwise I can also try to double check myself.

trl/models/modeling_base.py
```diff
@@ -386,18 +386,18 @@ def _get_checkpoint_from_hub(
...
                 The current device.
         """
-        local_rank = int(os.environ.get('LOCAL_RANK', '0'))
+        state = PartialState()
         if is_xpu_available():
-            return f"xpu:{local_rank}"
+            return f"xpu:{state.local_process_index}"
         else:
-            return local_rank if torch.cuda.is_available() else "cpu"
+            return state.local_process_index if torch.cuda.is_available() else "cpu"
```
muellerzr, line 400: This whole thing can be replaced with `state.local_process_index` unless `xpu` is needed.

muellerzr: I'd say that can be in another PR then just to keep things isolated a bit :) LG2M then!

muellerzr (approved):

lvwerra (approved):

muellerzr:
@billvsme ideally can you run it again and verify this fixes the problem with `PartialState()`? (want to make sure we're not in the same position)

billvsme:
@muellerzr Yes, solved. (I changed a annotation after you approved and submitted it once)

PR #1049, opened 2023-12-01
Fix DPOTrainer + PEFT 2

rdk31:
The same problem as described in #877, just with the reference model.


kashif (approved):
Nice! thank you! makes it clearer as well

PR #1052, opened 2023-12-02
enable multiple eval datasets

peter-sk:
The standard Trainer class from the transformers library (and the documentation of SFTTrainer) allow for multiple validation datasets to be passed as a dictionary from dataset name to dataset.

This does not work however in current SFTTrainer code. This PR fixes this.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -241,17 +241,22 @@ class SFTTrainer(Trainer):
                 chars_per_token,
             )
         if eval_dataset is not None:
-            eval_dataset = self._prepare_dataset(
-                eval_dataset,
-                tokenizer,
-                packing,
-                dataset_text_field,
-                max_seq_length,
-                formatting_func,
-                infinite,
-                num_of_sequences,
-                chars_per_token,
-            )
+            _multiple = isinstance(eval_dataset, dict)
+            _eval_datasets = eval_dataset if _multiple else {"singleton": eval_dataset}
+            for _eval_dataset_name, _eval_dataset in _eval_datasets.items():
+                _eval_datasets[_eval_dataset_name] = self._prepare_dataset(
+                    _eval_dataset,
+                    tokenizer,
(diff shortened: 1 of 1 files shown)


pop-srw:
I have the same problem. When multiple eval datasets were pass as dict, It cause an error in dataset.map https://github.com/huggingface/trl/blob/a60ceefa694d62565789b03e8fa35244bc46c9ba/trl/trainer/sft_trainer.py#L379
I think this PR is really useful.

peter-sk:
> I have the same problem. When multiple eval datasets were pass as dict, It cause an error in dataset.map
> 
> https://github.com/huggingface/trl/blob/a60ceefa694d62565789b03e8fa35244bc46c9ba/trl/trainer/sft_trainer.py#L379
> 
> I think this PR is really useful.

Yes, exactly this. And this error occurs naturally both when using packing and when not.

lvwerra (approved):
Looks good to me but I'll let @younesbelkada have a look as well!

lvwerra:
If you could add a test for this that would be awesome!

peter-sk:
@lvwerra 
Sure, I can add a test. Are there any guidelines for this? I have tried it for transformers, but not for trl. Any hints are appreciated!

peter-sk:
@lvwerra 
I added a test.

peter-sk:
@lvwerra @younesbelkada 
Just a careful push :-)

younesbelkada (approved):
Looks great to me, thanks for this contribution!

younesbelkada:
@peter-sk the tests are taking forever, I think that something went wrong with the test you designed, can you please have a quick look?

peter-sk:
@younesbelkada 
I guess the training dataset was getting iterated (for hundreds of thousands of iterations).
Now, I am using twice the evaluation dataset. This test works fine in my machine.
I also downsized the test.

younesbelkada (approved):
Thanks again!

PR #1062, opened 2023-12-05
Update doc for the computer_metrics argument of SFTTrainer

albertauyeung:
The `compute_metrics` argument defaults to `None` in the `SFTTrainer` class, but the docstring was inaccurate. This PR updates the descriptions of the argument to match the definition in the constructor.


lvwerra (approved):
Looks good to me. Also letting @younesbelkada have a look.

younesbelkada (approved):
Thanks for your contribution!

filippo82:
@lvwerra @younesbelkada does this apply to the `DPOTrainer` documentation too? Let me clarify: shall the [documentation](https://huggingface.co/docs/trl/dpo_trainer#trl.DPOTrainer.compute_metrics) of the `compute_metrics` parameter of the `DPOTrainer` say something like

> If not specified, only the XXX will be computed during evaluation.

?

filippo82:
@younesbelkada I just want to first dig a bit more into the code base to really understand what it should be expected 🙏🏻

PR #1067, opened 2023-12-06
Add missing `loss_type` in `ValueError` message

alvarobartt:
Hi to whoever is reading this! 🤗

## Description

This PR only adds the IPO loss within the available ones within the exception message of `ValueError` when `self.loss_type` is neither `hinge`, `sigmoid` or `ipo`.


younesbelkada (approved):
Thank you very much !

PR #1079, opened 2023-12-11
Add args to SFT example

lewtun:
Adds a few args to the SFT example to enable simpler training of Mixtral from the command line


younesbelkada (approved):
Thanks ! failing CI are unrelated to this PR

PR #1096, opened 2023-12-14
[Feature] Add Ascend NPU accelerator support 

statelesshz:
### What does this PR do?
Since npu already supports Transformer/Accelerate and Peft out of the box, this would be a good addition to run TRL out of the box on npu.

cc @younesbelkada and @lvwerra


statelesshz:
@younesbelkada @lvwerra I tried using `dpo.py` to do a sanity check on ascend [npu,](url) here is the training log: https://gist.github.com/statelesshz/4437ccec89c8f0d9df11444e8ca578cd

younesbelkada (commented):
Looks great thank you very much for adding the NPU support! can you run the styling checks? `make precommit`

statelesshz:
@younesbelkada Thanks for your reminder. CI is green:)

younesbelkada (approved):
Looks great!

PR #1097, opened 2023-12-14
Add type hints to core.py

zachschillaci27:
Hi all,

Just a quick PR to add type hinting to `core.py`

I noticed a few methods are not currently used in the codebase, such as `build_bert_batch_from_txt` and `listify_batch`. Not sure if they are being used elsewhere or we should clean them up. Let me know!

Zach

### DIFF
--- a/trl/core.py
+++ b/trl/core.py
@@ -15,13 +15,14 @@ import gc
 import random
 import warnings
 from contextlib import contextmanager
-from typing import List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
+import torch.nn as nn
 import torch.nn.functional as F
 from torch.nn.utils.rnn import pad_sequence
-from transformers import top_k_top_p_filtering
+from transformers import PreTrainedTokenizerBase, top_k_top_p_filtering
 
 from .import_utils import is_xpu_available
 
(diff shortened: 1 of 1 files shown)


lvwerra:
Hi @zachschillaci27, thanks for helping cleanup. Indeed those functions are not used anymore and can be removed!

zachschillaci27:
> Hi @zachschillaci27, thanks for helping cleanup. Indeed those functions are not used anymore and can be removed!

Got it, done!

lvwerra (approved):

lvwerra:
Could you run `make precommit`  (requires `pre-commit` to be installed) to fix the code quality checks?

trl/core.py
```diff
@@ -188,37 +189,9 @@ def stats_to_np(stats_dict):
     return new_dict
 
 
-def listify_batch(tensor):
-    """Turns the first dimension of a tensor into a list."""
-    return [tensor[i] for i in range(tensor.shape[0])]
-
-
-def build_bert_batch_from_txt(text_list, tokenizer, device):
-    """Create token id and attention mask tensors from text list for BERT classification."""
```
younesbelkada, line 197: Nice catch, these methods are not used at all in TRL

younesbelkada (approved):
Great clean up ! Thanks !

trl/core.py
zachschillaci27: Happy to help!

PR #1101, opened 2023-12-15
Update `description` in `setup.py`

alvarobartt:
## Description

The description in `setup.py` was outdated and I've aligned it with the one within the repository.


younesbelkada (approved):
Thanks!

lvwerra (approved):
Thanks for fixing! :)

PR #1110, opened 2023-12-19
`peft_module_casting_to_bf16` util method, `append_concat_token` flag, remove callback `PeftSavingCallback`

pacman100:
### What does this PR do?
1. Inline with QLoRA code when using bf16 [here](https://github.com/artidoro/qlora/blob/main/qlora.py#L396-L405), added `peft_module_casting_to_bf16`.
2. Added `append_concat_token` flag so that datasets which already have `eos_token` at the end of each sample can skip appending the eos/concat token
3. Removes the callback `PeftSavingCallback` and bumps the minimum transformers version to `4.31.0` which handles this by default. The current `PeftSavingCallback` would fail when using PEFT+DeepSpeed and the trainer handles this properly.

### DIFF
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -279,33 +279,6 @@ trainer = SFTTrainer(
 trainer.train()
 ```
 
-Note that in case of training adapters, we manually add a saving callback to automatically save the adapters only:
-```python
-class PeftSavingCallback(TrainerCallback):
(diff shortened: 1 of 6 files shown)


pacman100:
cc @lewtun and @younesbelkada

younesbelkada (approved):
Thanks a lot ! 
OK for me to bump transformers version to 4.31

lvwerra (approved):

PR #1114, opened 2023-12-20
Make prepending of bos token configurable.

pacman100:
### What does this PR do?
1. Tokenization using chat template doesn't add special tokens such as bos_token by default. https://github.com/huggingface/transformers/blob/17506d1256c1780efc9e2a5898a828c10ad4ea69/src/transformers/tokenization_utils_base.py#L1750
2. Related issue elsewhere: https://github.com/vllm-project/vllm/issues/2012
3. This leads to redundant bos/special tokens being added by the tokenizer in SFT training here: 

https://github.com/huggingface/trl/blob/f100ca34cca92e3a09a04b0f07950703b0aa7fd0/trl/trainer/utils.py#L429

This PR makes it configurable while retaining backwards compatibility.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -243,7 +243,8 @@ class SFTTrainer(Trainer):
             if data_collator is None:
                 data_collator = DataCollatorForLanguageModeling(tokenizer=tokenizer, mlm=False)
 
-        dataset_kwargs = dataset_kwargs or {}
+        if dataset_kwargs is None:
+            dataset_kwargs = {}
         if train_dataset is not None:
             train_dataset = self._prepare_dataset(
                 train_dataset,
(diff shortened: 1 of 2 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -240,6 +243,7 @@ def make_inputs_require_grad(module, input, output):
             if data_collator is None:
                 data_collator = DataCollatorForLanguageModeling(tokenizer=tokenizer, mlm=False)
 
+        dataset_kwargs = dataset_kwargs or {}
```
younesbelkada, line 246:
```suggestion
        if dataset_kwargs is None:
            dataset_kwargs = {}
```

pacman100: `dataset_kwargs = dataset_kwargs or {}`  should handle this.

younesbelkada: Yes but I would say let's make it slightly clearer !

younesbelkada (approved):
Thanks !

PR #1118, opened 2023-12-20
fix gradient checkpointing when using PEFT

pacman100:
### What does this PR do?
1. Enables proper handling of `gradient_checkpointing_kwargs` when using PEFT without the quantization, such as PEFT+DeepSpeed.


younesbelkada (approved):
Great, thanks !

PR #1120, opened 2023-12-20
[`xxxTrainer`] Add tags to all trainers in TRL

younesbelkada:
# What does this PR do?

As per title, this PR adds the tag `"trl-xxx"` whenever someone pushes a model with SFTTrainer, PPOTrainer or DPOTrainer

Example repo here: https://huggingface.co/ybelkada/out-test

cc @lvwerra

### DIFF
--- a/trl/trainer/ddpo_trainer.py
+++ b/trl/trainer/ddpo_trainer.py
@@ -15,6 +15,7 @@
 import os
 from collections import defaultdict
 from concurrent import futures
+from functools import wraps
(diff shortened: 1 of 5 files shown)


lvwerra (approved):
Looks great! Just to confirm the example only used `sft` and not `trl-sft`, right?

lvwerra:
Also, should we add it for DDPO as well?

younesbelkada:
Done !

Issue #955, opened 2023-11-04
DPO trainer with deepspped offload cpu config cause error: AssertionError: CPUAdam param is on cuda:0 and must be 'cpu', make sure you enabled 'offload_optimizer': 'cpu' in your ZeRO config.


hanyullai:
Working with DPO alone or using CPU offload within SFT separately results in no issues. However, when using DPO in combination with DS stage2 + CPU Offload, I encounter the following error: 
miniconda3/envs/dpo/lib/python3.9/site-packages/deepspeed/ops/adam/cpu_adam.py, line 150, in step
assert p.device == device, f"CPUAdam param is on {p.device} and must be 'cpu', make " AssertionError: CPUAdam param is on cuda:0 and must be 'cpu', ensure you enabled 'offload_optimizer': 'cpu' in your ZeRO config.

The framework I am using is [LLaMA-Factory](https://github.com/hiyouga/LLaMA-Factory/tree/main) with no modifications. This problem appears to be unrelated to the repository's code. This leads me to believe it's potentially due to environmental problems related to Pytorch, Trl, Transformers, or Deepspeed. Upon debugging into the Trl library, it seems that DPOTrainer does not close the gradient of ref_model when initializing (even though all gradients are 0), causing an error to be thrown at [_prepare_deepspeed](https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L294C39-L294C57). In Deepspeed, at this [line](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/ops/adam/cpu_adam.py#L147), ref_model should continue.

The versions of my environment are as follows: 
torch==2.1.0 
transformers==4.33.2 
trl==0.7.2 
deepspeed==0.11.1

My ds_report is displayed below:
JIT compiled ops requires ninja
ninja .................. [OKAY]
op name ................ installed .. compatible
 [WARNING]  async_io requires the dev libaio .so object and headers but these were not found.
 [WARNING]  async_io: please install the libaio-dev package with apt
 [WARNING]  If libaio is already installed (perhaps from source), try setting the CFLAGS and LDFLAGS environment variables to where it can be found.
async_io ............... [NO] ....... [NO]
fused_adam ............. [NO] ....... [OKAY]
cpu_adam ............... [NO] ....... [OKAY]
cpu_adagrad ............ [NO] ....... [OKAY]
cpu_lion ............... [NO] ....... [OKAY]
[...]

PR #1123, opened 2023-12-21
[dpo_trainer] Fixed a compatibility bug with deepspeed when initializing reference_model

Emperorizzis:
When training a model using DeepSpeed + DPO (with Warmup), the following error occurs:
(This issue has also been mentioned in [#955](https://github.com/huggingface/trl/issues/955)).

`File "xxx/deepspeed/runtime/lr_schedules.py", line 661, in __init__ self.warmup_num_steps = max(2, warmup_num_steps) TypeError: '>' not supported between instances of 'str' and 'int'`

The dpo_trainer initializes both the model and the reference_model using the same deepspeed config. Training hyperparameters are set through TrainingArguments & Trainer, but the reference_model only requires deepspeed for forward computation. Scheduler-related parameters not only cause issues but are also redundant, they should be removed.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -398,6 +398,7 @@ class DPOTrainer(Trainer):
         # Otherwise, we assume the reference model fits in memory and is initialized on each device with ZeRO disabled (stage 0)
         if config_kwargs["zero_optimization"]["stage"] != 3:
             config_kwargs["zero_optimization"]["stage"] = 0
+        config_kwargs.pop('scheduler', None)
         model, *_ = deepspeed.initialize(model=model, config=config_kwargs)
         model.eval()
         return model


kashif:
nice catch @Emperorizzis checking!

kashif:
@Emperorizzis  can you also share your deepspeed config to make sure I can reproduce it? thanks!

Emperorizzis:
> @Emperorizzis can you also share your deepspeed config to make sure I can reproduce it? thanks!

Of Course ~

Below is my deepspeed config:
```json
{
  "bfloat16": {
      "enabled": true
  },
  "fp16": {
      "enabled": false,
      "loss_scale": 0,
      "loss_scale_window": 1000,
      "initial_scale_power": 16,
      "hysteresis": 2,
      "min_loss_scale": 1
  },
  "optimizer": {
      "type": "AdamW",
      "params": {
          "lr": "auto",
          "weight_decay": "auto",
          "betas": "auto",
          "eps": "auto",
          "torch_adam": true,
          "adam_w_mode": true
      }
  },
  "scheduler": {
      "type": "WarmupDecayLR",
      "params": {
          "warmup_min_lr": 1e-7,
          "warmup_max_lr": 1e-6,
          "warmup_num_steps": "auto",
          "total_num_steps": "auto"
      }
  },
  "zero_optimization": {
      "stage": 3,
      "overlap_comm": true,
      "contiguous_gradients": true,
      "sub_group_size": 1e12,
      "reduce_bucket_size": "auto",
      "stage3_prefetch_bucket_size": "auto",
      "stage3_param_persistence_threshold": "auto",
      "stage3_max_live_parameters": 1e9,
      "stage3_max_reuse_distance": 1e9,
      "stage3_gather_16bit_weights_on_model_save": true
  },
  "gradient_accumulation_steps": "auto",
  "gradient_clipping": "auto",
  "steps_per_print": 1e5,
  "train_batch_size": "auto",
  "train_micro_batch_size_per_gpu": "auto",
  "wall_clock_breakdown": false
}
```

Also, to save memory, I loaded the model and reference_model after initializing the TrainingArguments (which automatically selected the deepspeed backend).
(I noticed that the [dpo.py script](https://github.com/huggingface/trl/blob/main/examples/scripts/dpo.py#L130) in the example loads the model before initializing the TrainingArguments, which can run out of memory when the model is very large, such as 70B or larger.)

kashif:
thanks @Emperorizzis so one issue is that the Deepspeed 3 together with pre-computing the ref log-probs does not work as the private HF Trainer methods initialize the dataloader before the model etc. and we currently raise an error:

https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L358-L36

Emperorizzis:
> thanks @Emperorizzis so one issue is that the Deepspeed 3 together with pre-computing the ref log-probs does not work as the private HF Trainer methods initialize the dataloader before the model etc. and we currently raise an error:
> 
> https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L358-L36

Thanks for replying ~ So the reference_model does need to be initialized from different deepspeed configs. (At least remove the superfluous data config)

PR #1126, opened 2023-12-22
clear up the parameters of supervised_finetuning.py

sywangyi:
no_gradient_checkpointing is always false


sywangyi:
@younesbelkada

younesbelkada (approved):
Thanks a lot @sywangyi ! 🚀

PR #1127, opened 2023-12-22
[DPO] rename kto loss

kashif:
The kto loss in the DPOTrainer is a special-case of the general KTO loss, and as such we will implement a separate KTOTrainer and rename the special case loss to `kto_pair`


lvwerra (approved):
Sounds good. @younesbelkada lets merge this before the release?

younesbelkada (approved):
Makes sense, thanks!

PR #1128, opened 2023-12-22
Add npu support for ppo example

zhangsibo1129:
Add npu support for `examples/script/ppo_multi_adapter.py`


younesbelkada (commented):
Thanks a lot for adding NPU support for PPO example! Can you run the styling checks? `make precommit`

younesbelkada (approved):
Thanks again for adding the NPU support!

PR #1129, opened 2023-12-22
fix_ddpo_demo

zhangsibo1129:
In `examples/scripts/ddpo.py` , the `AestheticScorer.mlp` loads weights to cuda device by default 
```python
self.mlp.load_state_dict(state_dict)
state_dict = torch.load(cached_path)
```
which throws an error for non-cuda devices, so loading weights on cpu first 

```python
self.mlp.load_state_dict(state_dict, map_location=torch.device('cpu'))
state_dict = torch.load(cached_path)
```
and then copy to target device 
```python
if is_npu_available():
    scorer = scorer.npu()
elif is_xpu_available():
    scorer = scorer.xpu()
else:
    scorer = scorer.cuda()
```

### DIFF
--- a/examples/scripts/ddpo.py
+++ b/examples/scripts/ddpo.py
@@ -99,7 +99,7 @@ class AestheticScorer(torch.nn.Module):
             cached_path = hf_hub_download(model_id, model_filename)
         except EntryNotFoundError:
             cached_path = os.path.join(model_id, model_filename)
-        state_dict = torch.load(cached_path)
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Thanks for fixing the ddpo demo !

younesbelkada:
@zhangsibo1129 can you run the styling checks? `make precommit`

younesbelkada (approved):
Thanks again!

PR #1133, opened 2023-12-22
[`xxxTrainer`] multi-tags support for tagging

younesbelkada:
# What does this PR do?

Instead of pushing a single tag, this PR changes the current behaviour of `xxxTrainer` tagging to push multiple tags (e.g. `trl` & `sft`) instead of a single tag

cc @lvwerra @osanseviero 

example here: https://huggingface.co/ybelkada/test-multi-tag


lvwerra (approved):

osanseviero (commented):
Thank you!

PR #1153, opened 2023-12-28
Handle last token from generation prompt

pablovicente:
Some tokenizers, like Llama BPE tokenizer, might merge tokens into one such as the case where there is a space follow by some characters like `$`. 

In a previous PR we handled the case for chosen/rejected but one small improvement is to remove the extra space token from the end of the prompt used for generation in `get_batch_samples`. It should left to the model if the next generated token is just the space token or one such that it combines the space with other token into one. To do so, we need to manually check `prompt_ids` for chosen/rejected and by itself and chose the shortest knowing that they can only differ by 1 token. 

Given the following example, `prompt + chosen` will leave the space after `[\INST]` unchanged but it wont be the case for `prompt+rejected` since it starts with `$`.    

```
"prompt": "[INST] How is the stock price? [/INST] "
"chosen": "46 as of 10am EST"
"rejected":  "$46 as of 10am EST"
```

@kashif

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -593,35 +593,28 @@ class DPOTrainer(Trainer):
                 raise ValueError(f"rejected should be an str but got {type(rejected)}")
             rejected_tokens = self.build_tokenized_answer(prompt, rejected)
 
-            # Last prompt token might get merged by tokenizer
+            # Last prompt token might get merged by tokenizer and
             # it should not be included for generation if that happens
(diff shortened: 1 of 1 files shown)


kashif:
@pablovicente are the tests passing for you?

pablovicente:
> @pablovicente are the tests passing for you?

`test_dpo_trainer.py` tests pass. I see issues on the sft_trainer but have not made changes there. Are dpo_trainer tests pass for you?

younesbelkada (commented):
@kashif we can merge this no? Wdyt? 🙏

kashif (approved):

PR #1154, opened 2023-12-29
Change device access order for speedup of calculating metrics in DPOTrainer

brcps12:
Currently, when calculating metrics in `DPOTrainer`, the values are moved to the CPU and the mean is calculated. However, when doing this, the cost of moving the values to the CPU is very large.

So I changed the device order to improve speed. Below are the `torch profiler` results before and after changing the order in my experiment.

cpu

As shown in the figure, the operation to move to the CPU has a similar cost to the forward of the model.

gpu

After making the change, the time taken has decreased significantly.


younesbelkada (approved):
Thanks very much, very clean!

kashif:
great thank you for the fix!

PR #1160, opened 2023-12-31
Logs metrics on all distributed processes when using DPO & FSDP

AjayP13:
Logs metrics on all distributed processes when using FSDP. 

All processes need access to the metrics in order for things like "Early Stopping" / "Load Best Model At End" to work. You get a `KeyError` on workers #1 - #N, otherwise.

There is no need for this guard as the logging callbacks themselves check if they are the main process before printing metrics to the terminal.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -989,8 +989,7 @@ class DPOTrainer(Trainer):
         loss, metrics = self.get_batch_loss_metrics(model, inputs, train_eval="train")
 
         # force log the metrics
-        if self.accelerator.is_main_process:
-            self.store_metrics(metrics, train_eval="train")
+        self.store_metrics(metrics, train_eval="train")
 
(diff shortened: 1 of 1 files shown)


lvwerra:
cc @kashif

younesbelkada (commented):
Does this affect classic DPO on multi-GPU + distributed without FSDP? If not, I think we can safely merge it, would you be able to quickly try that out @AjayP13 ? 🙏

AjayP13:
@younesbelkada Yes, I tested this distributed with FSDP enabled and distributed with FSDP disabled (DDP, I believe) and both seemed to work fine.

younesbelkada (commented):
cc @kashif wdyt? should be all good I would say no?

kashif:
yes all good!

kashif (approved):

younesbelkada:
Thanks @kashif @AjayP13 !

PR #1170, opened 2024-01-02
Correct shapes in docstring of PPOTrainer's train_minibatch method

nikihowe:
`train_minibatch` is called by `step` [here](https://github.com/huggingface/trl/blob/20428c48ba651b2bc11b2a73e9eb9568b1af3f96/trl/trainer/ppo_trainer.py#L795) (and only there), with minibatches of first dimension size `self.config.mini_batch_size`. To match this, the docstring in `train_minibatch` should say that the inputs are of shape `[mini_batch_size, XXX_length]`, not `[batch_size, XXX_length]`. Alternatively, to be very precise, one could instead say that they are of size `[self.config.mini_batch_size, XXX_length]`.

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -1045,15 +1045,15 @@ class PPOTrainer(BaseTrainer):
 
         Args:
             logprobs (`torch.FloatTensor`):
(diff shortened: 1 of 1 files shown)


vwxyzjn (approved):
This makes sense. Thanks for the cath!

PR #1171, opened 2024-01-02
Fix misleading variable "epoch" from the training loop from PPOTrainer Doc.

Jfhseh:
**The usage of the variable “epoch” is misleading (incorrect) in the original Doc:**
the dataloader does not contain the data for ALL epochs, but 1 epoch only, thus  
`"for epoch, batch in tqdm(enumerate(ppo_trainer.dataloader))" ` is misleading and does not actually stores the epoch #. 

The correct version comes from the TRL PPO notebook tutorial  (https://github.com/huggingface/trl/blob/main/examples/notebooks/gpt2-sentiment-control.ipynb), which uses an outer loop to capture the epochs.

I posted also the question on forum: https://discuss.huggingface.co/t/confusing-and-possibly-misleading-ppo-trainer-code-from-trl-api-doc-tutorial/67531
image

image

### DIFF
--- a/docs/source/ppo_trainer.mdx
+++ b/docs/source/ppo_trainer.mdx
@@ -115,22 +115,22 @@ We can then loop over all examples in the dataset and generate a response for ea
 
 ```py
 from tqdm import tqdm
-
-for epoch, batch in tqdm(enumerate(ppo_trainer.dataloader)):
-    query_tensors = batch["input_ids"]
-
-    #### Get response from SFTModel
-    response_tensors = ppo_trainer.generate(query_tensors, **generation_kwargs)
-    batch["response"] = [tokenizer.decode(r.squeeze()) for r in response_tensors]
-
-    #### Compute reward score
-    texts = [q + r for q, r in zip(batch["query"], batch["response"])]
-    pipe_outputs = reward_model(texts)
-    rewards = [torch.tensor(output[1]["score"]) for output in pipe_outputs]
-
-    #### Run PPO step
-    stats = ppo_trainer.step(query_tensors, response_tensors, rewards)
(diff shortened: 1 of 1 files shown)


docs/source/ppo_trainer.mdx
```diff
@@ -115,22 +115,22 @@ We can then loop over all examples in the dataset and generate a response for ea
...
-    #### Compute reward score
-    texts = [q + r for q, r in zip(batch["query"], batch["response"])]
-    pipe_outputs = reward_model(texts)
-    rewards = [torch.tensor(output[1]["score"]) for output in pipe_outputs]
-
-    #### Run PPO step
-    stats = ppo_trainer.step(query_tensors, response_tensors, rewards)
-    ppo_trainer.log_stats(stats, batch, rewards)
+for epoch in tqdm(range(ppo_trainer.config.ppo_epochs), "epoch: "):
+    for batch_id, batch in tqdm(enumerate(ppo_trainer.dataloader)):
```
lvwerra, line 119: since we don't use the `batch_id` we could remove the `enumerate` and `batch_id` altogether.

lvwerra:
Sounds good to me. Note that the reason people referred to this part as an `epoch` is because it constitutes a PPO optimization epoch. But I agree that it's a bit confusing. Left a minor comment.

docs/source/ppo_trainer.mdx
Jfhseh: Got it. Yes, I removed the batch_id.

lvwerra (approved):

younesbelkada (approved):
Thanks for clarifying the docs!

PR #1174, opened 2024-01-03
Address issue #1122

### LINKED ISSUE #1122: `_prepare_packed_dataloader` and `_prepare_non_packed_dataloader` have inconsistent keys
`_prepare_packed_dataloader`  outputs a `ConstantLengthDataset` with [keys](https://github.com/huggingface/trl/blob/830cadfc4c80bdced0d3753de392070e6760d1f5/trl/trainer/utils.py#L450)
```
'input_ids`, 'labels'
```
`_prepare_non_packed_dataloader` outputs some type of dataset, with [keys](https://github.com/huggingface/trl/blob/830cadfc4c80bdced0d3753de392070e6760d1f5/trl/trainer/sft_trainer.py#L393)
```
'input_ids`, 'attention_mask'
```
The lack of `'labels'` in the latter might be source of confusing bugs.

If this is intended behavior, I would at least suggest to post a warning in the documentation of the argument `packing` of `SFTTrainer`

maneandrea:
Issue [#1122](https://github.com/huggingface/trl/issues/1122)
takes care of an inconsistency between `_prepare_packed_dataloader`
and `_prepare_non_packed_dataloader`

Fixes #1122

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -452,7 +452,7 @@ class ConstantLengthDataset(IterableDataset):
                 yield {
                     "input_ids": torch.LongTensor(example),
                     "labels": torch.LongTensor(example),
-                    "attention_mask": [1 for _ in example],
+                    "attention_mask": torch.ones(len(example)),
                 }
 
 
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -452,6 +452,7 @@ class ConstantLengthDataset(IterableDataset):
                 yield {
                     "input_ids": torch.LongTensor(example),
                     "labels": torch.LongTensor(example),
+                    "attention_mask": [1 for _ in example],
                 }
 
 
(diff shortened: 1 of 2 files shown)


trl/trainer/utils.py
```diff
@@ -452,6 +452,7 @@ def __iter__(self):
                 yield {
                     "input_ids": torch.LongTensor(example),
                     "labels": torch.LongTensor(example),
+                    "attention_mask": [1 for _ in example],
```
lvwerra, line 455: I think the following would be better, since it will also create a torch tensor:

```suggestion
                    "attention_mask": torch.ones_like(example),
```

maneandrea:
Good point. I think `ones_like` requires a tensor as input so I just did `ones(len(example))`

younesbelkada (approved):
Sounds good, thanks!

younesbelkada:
Hi, I will revert this PR because it seems that now simple snippets do not work anymore
```python
from trl import SFTTrainer

dataset = load_dataset("imdb", split="train")

trainer = SFTTrainer(
    "facebook/opt-350m",
    train_dataset=dataset,
    dataset_text_field="text",
    max_seq_length=512,
)
trainer.train()
```
Let's figure out a more global fix. I will make a patch release

maneandrea:
What is exactly the error it gives? In my modest PC I cannot run that snippet

PR #1176, opened 2024-01-04
Remove tyro

vwxyzjn:
The command to run is simply `python examples/scripts/ppo.py --log_with wandb`. There is no evidence of significant regression in the new refactor, though the learning curves appear less smooth.

This was the learning curve before

### DIFF
--- a/benchmark/trl.slurm_template
+++ b/benchmark/trl.slurm_template
@@ -1,12 +1,12 @@
 #!/bin/bash
 #SBATCH --job-name=trl
-#SBATCH --partition=production-cluster
+#SBATCH --partition=hopper-prod
 #SBATCH --gpus-per-task={{gpus_per_task}}
 #SBATCH --cpus-per-gpu={{cpus_per_gpu}}
 #SBATCH --ntasks={{ntasks}}
 #SBATCH --output=slurm/logs/%x_%j.out
 #SBATCH --array={{array}}
-#SBATCH --exclude=ip-26-0-156-239,ip-26-0-148-151,ip-26-0-146-212,ip-26-0-145-137,ip-26-0-146-249,ip-26-0-146-149,ip-26-0-147-233,ip-26-0-145-154,ip-26-0-144-35,ip-26-0-144-189,ip-26-0-146-183,ip-26-0-147-120,ip-26-0-144-95,ip-26-0-145-193
+##SBATCH --exclude=ip-26-0-149-199
 {{nodes}}
 
 seeds={{seeds}}
--- a/benchmark/trl.slurm_template
+++ b/benchmark/trl.slurm_template
@@ -7,6 +7,9 @@
 #SBATCH --output=slurm/logs/%x_%j.out
 #SBATCH --array={{array}}
 ##SBATCH --exclude=ip-26-0-149-199
+
+module load cuda/12.1
+
 {{nodes}}
 
 seeds={{seeds}}
--- a/benchmark/trl.slurm_template
+++ b/benchmark/trl.slurm_template
@@ -13,4 +13,4 @@ seeds={{seeds}}
 seed=${seeds[$SLURM_ARRAY_TASK_ID % {{len_seeds}}]}
 
 echo "Running task $SLURM_ARRAY_TASK_ID with seed: $seed"
-srun {{command}} --ppo_config.seed $seed
+srun {{command}} --seed $seed
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -413,44 +413,48 @@ Note however, that the amount of performance gain is _dataset dependent_ and in 
 
 ### Accelerate fine-tuning 2x using `unsloth`
 
-You can further accelerate QLoRA / LoRA (2x faster, 60% less memory) and even full-finetuning (1.1x faster) using the [`unsloth`](https://github.com/unslothai/unsloth) library that is compatible with `SFTTrainer`. Currently `unsloth` supports only Llama (Yi, TinyLlama as well) and Mistral architectures.
-First install `unsloth` according to the [official documentation](https://github.com/unslothai/unsloth#installation-instructions---conda). Once installed, you can incorporate unsloth into your workflow in a very simple manner; instead of loading `AutoModelForCausalLM`, you just need to load a `FastLlamaModel` or `FastMistralModel` as follows:
+You can further accelerate QLoRA / LoRA (2x faster, 60% less memory) using the [`unsloth`](https://github.com/unslothai/unsloth) library that is fully compatible with `SFTTrainer`. Currently `unsloth` supports only Llama (Yi, TinyLlama, Qwen, Deepseek etc) and Mistral architectures. Some benchmarks on 1x A100 listed below:
+
+| 1 A100 40GB     | Dataset   | 🤗  | 🤗 + Flash Attention 2 | 🦥 Unsloth     | 🦥 VRAM saved |
+|-----------------|-----------|-----|-------------------------|-----------------|----------------|
+| Code Llama 34b  | Slim Orca | 1x  | 1.01x                   | **1.94x**       | -22.7%         |
+| Llama-2 7b      | Slim Orca | 1x  | 0.96x                   | **1.87x**       | -39.3%         |
+| Mistral 7b      | Slim Orca | 1x  | 1.17x                   | **1.88x**       | -65.9%         |
+| Tiny Llama 1.1b | Alpaca    | 1x  | 1.55x                   | **2.74x**       | -57.8%         |
+
+First install `unsloth` according to the [official documentation](https://github.com/unslothai/unsloth). Once installed, you can incorporate unsloth into your workflow in a very simple manner; instead of loading `AutoModelForCausalLM`, you just need to load a `FastLanguageModel` as follows:
 
 ```python
 import torch
-
 from transformers import TrainingArguments
 from trl import SFTTrainer
-from unsloth import FastLlamaModel, FastMistralModel
+from unsloth import FastLanguageModel
 
-max_seq_length = 2048 # Supports automatic RoPE Scaling, so choose any number.
-dtype = None # None for auto detection. Float16 for Tesla T4, V100, Bfloat16 for Ampere+
-load_in_4bit = True # Use 4bit quantization to reduce memory usage. Can be False.
+max_seq_length = 2048 # Supports automatic RoPE Scaling, so choose any number
 
-# Load Llama model
-model, tokenizer = FastLlamaModel.from_pretrained(
-    model_name = "unsloth/llama-2-7b", # Supports any llama model eg meta-llama/Llama-2-7b-hf
+# Load model
+model, tokenizer = FastLanguageModel.from_pretrained(
+    model_name = "unsloth/mistral-7b",
     max_seq_length = max_seq_length,
-    dtype = dtype,
-    load_in_4bit = load_in_4bit,
+    dtype = None, # None for auto detection. Float16 for Tesla T4, V100, Bfloat16 for Ampere+
+    load_in_4bit = True, # Use 4bit quantization to reduce memory usage. Can be False
     # token = "hf_...", # use one if using gated models like meta-llama/Llama-2-7b-hf
 )
 
 # Do model patching and add fast LoRA weights
-model = FastLlamaModel.get_peft_model(
+model = FastLanguageModel.get_peft_model(
     model,
     r = 16,
     target_modules = ["q_proj", "k_proj", "v_proj", "o_proj",
                       "gate_proj", "up_proj", "down_proj",],
     lora_alpha = 16,
-    lora_dropout = 0, # Currently only supports dropout = 0
-    bias = "none",    # Currently only supports bias = "none"
+    lora_dropout = 0, # Dropout = 0 is currently optimized
+    bias = "none",    # Bias = "none" is currently optimized
     use_gradient_checkpointing = True,
     random_state = 3407,
-    max_seq_length = max_seq_length,
 )
 
-args = TrainingArguments(output_dir="./output")
+args = TrainingArguments(output_dir = "./output")
 
 trainer = SFTTrainer(
     model = model,
@@ -459,7 +463,6 @@ trainer = SFTTrainer(
     dataset_text_field = "text",
     max_seq_length = max_seq_length,
 )
-
 trainer.train()
 ```
 
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -1,4 +1,4 @@
-# Supervised Fine-tuning Trainer 
+# Supervised Fine-tuning Trainer
 
 Supervised fine-tuning (or SFT for short) is a crucial step in RLHF. In TRL we provide an easy-to-use API to create your SFT models and train them with few lines of code on your dataset.
 
@@ -6,7 +6,7 @@ Check out a complete flexible example at [`examples/scripts/sft.py`](https://git
 
 ## Quickstart
 
-If you have a dataset hosted on the 🤗 Hub, you can easily fine-tune your SFT model using [`SFTTrainer`] from TRL. Let us assume your dataset is `imdb`, the text you want to predict is inside the `text` field of the dataset, and you want to fine-tune the `facebook/opt-350m` model. 
+If you have a dataset hosted on the 🤗 Hub, you can easily fine-tune your SFT model using [`SFTTrainer`] from TRL. Let us assume your dataset is `imdb`, the text you want to predict is inside the `text` field of the dataset, and you want to fine-tune the `facebook/opt-350m` model.
 The following code-snippet takes care of all the data pre-processing and training for you:
 
 ```python
@@ -50,7 +50,7 @@ The above snippets will use the default training arguments from the [`transforme
 
 ## Advanced usage
 
-### Train on completions only 
+### Train on completions only
 
 You can use the `DataCollatorForCompletionOnlyLM` to train your model on the generated prompts only. Note that this works only in the case when `packing=False`.
 To instantiate that collator for instruction data, pass a response template and the tokenizer. Here is an example of how it would work to fine-tune `opt-350m` on completions only on the CodeAlpaca dataset:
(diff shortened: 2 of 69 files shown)


benchmark/trl.slurm_template
lewtun, line 1: FYI you might want to add 

```
module load cuda/12.1
```

before the `srun` command to activate network optimisations. See https://github.com/huggingface/hfc/blob/main/slurm/example.slurm#L60-L61

examples/scripts/ppo.py
```diff
@@ -33,42 +32,17 @@
...
-            bias="none",
-            task_type="CAUSAL_LM",
-        ),
-    )
+    use_seq2seq: bool = field(default=False, metadata={"help": "whether to use seq2seq"})
     trust_remote_code: bool = field(default=False, metadata={"help": "Enable `trust_remote_code`"})
 
+    # LoraConfig
+    use_peft: bool = field(default=False, metadata={"help": "whether to use peft"})
+    lora_alpha: Optional[float] = field(default=16, metadata={"help": "the lora alpha parameter"})
```
lewtun, line 40: Not for this PR, but I think we may want to have something like a `ModelArguments` or `ModelConfig` class that collects all hyperparams associated with model loading since this tends to be pretty similar across SFT/RM/DPO etc

examples/scripts/reward_modeling.py
```diff
@@ -30,52 +44,25 @@
...
-        default_factory=lambda: LoraConfig(
-            r=16,
-            lora_alpha=16,
-            bias="none",
-            task_type="SEQ_CLS",
-            modules_to_save=["scores"],
-        ),
+    model_name: str = field(default="facebook/opt-350m", metadata={"help": "the model name"})
+    dataset_name: str = field(default="Anthropic/hh-rlhf", metadata={"help": "the dataset name"})
+    dataset_text_field: str = field(default="text", metadata={"help": "the text field of the dataset"})
```
lewtun, line 49: Not for this PR, but I think we should refactor the `RewardTrainer` to be similar to the `SFTTrainer` so that all these args can be captured in `RewardConfig` and the preprocessing done on the fly

lewtun (commented):
Thanks a lot for working on this @vwxyzjn ! I think it all looks good, modulo a question I have about `ppo_config.py`: shouldn't we also be removing the `tyro` import and related bits about `tyro.conf.Suppress`?

e.g. here https://github.com/huggingface/trl/pull/1176/files#diff-070022da4e4a14f0782200ebeab47af78224f10933e97853a0475d26007d4a1eR22-R31

examples/scripts/ddpo.py
```diff
@@ -11,59 +11,51 @@
...
-from transformers import CLIPModel, CLIPProcessor
+from transformers import CLIPModel, CLIPProcessor, HfArgumentParser
 
 from trl import DDPOConfig, DDPOTrainer, DefaultDDPOStableDiffusionPipeline
 from trl.import_utils import is_npu_available, is_xpu_available
 
 
 @dataclass
 class ScriptArguments:
-    hf_user_access_token: str
```
vwxyzjn, line 33: We don't have to require `hf_user_access_token`. If needed we should just login with HF cli.

examples/scripts/reward_modeling.py
```diff
@@ -12,15 +12,29 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""
```
vwxyzjn, line 15: Now need to add a launch command because we need to fill in things like `output_dir`

examples/scripts/reward_modeling.py
```diff
@@ -30,52 +44,26 @@
...
+    use_peft: bool = field(default=False, metadata={"help": "whether to use peft"})
+    lora_alpha: Optional[float] = field(default=16, metadata={"help": "the lora alpha parameter"})
+    lora_r: Optional[int] = field(default=16, metadata={"help": "the lora r parameter"})
 
 
-args = tyro.cli(ScriptArguments)
-args.reward_config.evaluation_strategy = "steps" if args.eval_split != "none" else "no"
+parser = HfArgumentParser((ScriptArguments, RewardConfig))
+args, reward_config = parser.parse_args_into_dataclasses()
+reward_config.gradient_checkpointing_kwargs = dict(use_reentrant=False)
```
vwxyzjn, line 65: required. otherwise errors out.

vwxyzjn:
* reward_modeling.py: https://wandb.ai/costa-huang/huggingface/runs/xcbgn6pt?workspace=user-costa-huang
* ddpo.py:
https://wandb.ai/costa-huang/stable_diffusion_training/runs/7qne3p91?workspace=user-costa-huang

vwxyzjn:
Also removed duplicate / slightly different doc strings to make things more aligned.
image

image

vwxyzjn:
Not necessarily in this PR but we can also do

```
parser = HfArgumentParser((TrainingArguments, BitsAndBytesConfig, LoraConfig))
```

but currently 

```
parser = HfArgumentParser((LoraConfig))
```
errors out. The reason is as follows @younesbelkada. So not sure if we need to change peft or `HfArgumentParser`.

```python
from typing import Literal
from transformers import HfArgumentParser

@dataclass
class TestConfig: 
    init_lora_weights: bool | Literal["gaussian", "loftq"] = field(
        default=True,
        metadata={
            "help": (
                "How to initialize the weights of the LoRA layers. Passing True (default) results in the default "
                "initialization from the reference implementation from Microsoft. Passing 'gaussian' results "
                "in Gaussian initialization scaled by the LoRA rank for linear and layers. Setting the initialization "
                "to False leads to completely random initialization and is discouraged."
                "Pass `'loftq'` to use LoftQ initialization"
            ),
        },
    )
parser = HfArgumentParser((TestConfig))
args = parser.parse_args_into_dataclasses()
```
```
python g.py --help
  File "/fsx/costa/trl/g.py", line 21, in <module>
    parser = HfArgumentParser((TestConfig))
  File "/fsx/costa/pyenv/versions/mambaforge-22.9.0-3/envs/trl-310/lib/python3.10/site-packages/transformers/hf_argparser.py", line 136, in __init__
    self._add_dataclass_arguments(dtype)
  File "/fsx/costa/pyenv/versions/mambaforge-22.9.0-3/envs/trl-310/lib/python3.10/site-packages/transformers/hf_argparser.py", line 263, in _add_dataclass_arguments
    self._parse_dataclass_field(parser, field)
  File "/fsx/costa/pyenv/versions/mambaforge-22.9.0-3/envs/trl-310/lib/python3.10/site-packages/transformers/hf_argparser.py", line 159, in _parse_dataclass_field
    raise ValueError(
ValueError: Only `Union[X, NoneType]` (i.e., `Optional[X]`) is allowed for `Union` because the argument parser only supports one type per argument. Problem encountered in field 'init_lora_weights'.
```

lewtun:
> Not necessarily in this PR but we can also do
> 
> ```
> parser = HfArgumentParser((TrainingArguments, BitsAndBytesConfig, LoraConfig))
> ```
> 
> but currently
> 
> ```
> parser = HfArgumentParser((LoraConfig))
> ```
> 
> errors out. The reason is as follows @younesbelkada. So not sure if we need to change peft or `HfArgumentParser`.
> 
> ```python
> from dataclasses import dataclass, field
> from typing import Literal
> from transformers import HfArgumentParser
> 
> @dataclass
> class TestConfig: 
>     init_lora_weights: bool | Literal["gaussian", "loftq"] = field(
>         default=True,
>         metadata={
>             "help": (
>                 "How to initialize the weights of the LoRA layers. Passing True (default) results in the default "
>                 "initialization from the reference implementation from Microsoft. Passing 'gaussian' results "
>                 "in Gaussian initialization scaled by the LoRA rank for linear and layers. Setting the initialization "
>                 "to False leads to completely random initialization and is discouraged."
>                 "Pass `'loftq'` to use LoftQ initialization"
>             ),
>         },
>     )
> parser = HfArgumentParser((TestConfig))
> args = parser.parse_args_into_dataclasses()
> ```
> 
>  ```
> python g.py --help
>   File "/fsx/costa/trl/g.py", line 21, in <module>
>     parser = HfArgumentParser((TestConfig))
>   File "/fsx/costa/pyenv/versions/mambaforge-22.9.0-3/envs/trl-310/lib/python3.10/site-packages/transformers/hf_argparser.py", line 136, in __init__
>     self._add_dataclass_arguments(dtype)
>   File "/fsx/costa/pyenv/versions/mambaforge-22.9.0-3/envs/trl-310/lib/python3.10/site-packages/transformers/hf_argparser.py", line 263, in _add_dataclass_arguments
>     self._parse_dataclass_field(parser, field)
>   File "/fsx/costa/pyenv/versions/mambaforge-22.9.0-3/envs/trl-310/lib/python3.10/site-packages/transformers/hf_argparser.py", line 159, in _parse_dataclass_field
>     raise ValueError(
> ValueError: Only `Union[X, NoneType]` (i.e., `Optional[X]`) is allowed for `Union` because the argument parser only supports one type per argument. Problem encountered in field 'init_lora_weights'.
> ```

If I'm not mistaken, we can encapsulate the BnB and LoRA configs into a single `ModelConfig` class that contains everything associated with model loading. This is what we do in the handbook [here](https://github.com/huggingface/alignment-handbook/blob/c69ae4b8a5436885b72475914b92dfb341876bf0/src/alignment/configs.py#L107), although admittedly it only covers the args we found most useful. This way we can reuse the same config class across all the scripts (except possibly PPO which requires one to also specify the reward model kwargs)

younesbelkada:
@vwxyzjn hmmm indeed maybe the fix for that specific problem should go in PEFT, maybe as a quick workaround we can go for @lewtun 's solution?

> encapsulate the BnB and LoRA configs into a single ModelConfig class that contains everything associated with model loading

vwxyzjn:
@lewtun that's a great idea. Let me refactor things using `ModelArguments` then!

vwxyzjn:
So everything looks good, maybe except the DPO train accuracies. It's approaching 100% which seems a bit unrealistic because `reward_modeling.py` has like 65% accuracy on the same dataset...

examples/scripts/dpo.py
```diff
@@ -12,77 +12,71 @@
...
 # limitations under the License.
-
-# Note: you need to install transformers from main to run this script. See https://huggingface.co/docs/transformers/installation#install-from-source
-# TODO: bump transformers version in requirements at next release.
-
-# 0. imports
+"""
+# regular:
+python examples/scripts/dpo.py \
+    --model_name_or_path=gpt2 \
```
lewtun, line 18: Not for this PR, but DPO doesn't really work well on base models so one should ideally be applying it to a fine-tuned chat model.

examples/scripts/dpo.py
```diff
@@ -12,77 +12,71 @@
...
+    --eval_steps 500 \
+    --output_dir="dpo_anthropic_hh" \
+    --optim rmsprop \
+    --warmup_steps 150 \
+    --report_to wandb \
+    --bf16 \
+    --logging_first_step \
+    --no_remove_unused_columns \
+    --use_peft \
+    --lora_r=64 \
```
lewtun, line 50: This is a better default
```suggestion
    --lora_r=16 \
```

examples/scripts/dpo.py
```diff
@@ -123,76 +117,61 @@ def split_prompt_and_responses(sample) -> Dict[str, str]:
...
-    model = AutoModelForCausalLM.from_pretrained(script_args.model_name_or_path)
-
-    if script_args.ignore_bias_buffers:
+    parser = HfArgumentParser((ScriptArguments, TrainingArguments, ModelConfig))
+    args, training_args, model_config = parser.parse_args_into_dataclasses()
+
+    ################
+    # Model & Tokenizer
+    ################
+    torch_dtype = (
```
lewtun, line 126: Not for this PR but we can likely simplify this boilerplate code by having a `get_tokenizer()` and `get_model()` helper functions which receive `ModelConfig` and return the desired objects

lewtun (commented):
Great job adding the `ModelConfig` @vwxyzjn ! This is looking really good and I left a few small comments on things we can remove / tidy up

Regarding the DPO training curves, the better metric to track for overfitting is the train/val loss IMO but in general DPO does overfit quite quickly and this turns out not to matter much in practice

examples/scripts/dpo.py
```diff
@@ -12,77 +12,71 @@
...
+python examples/scripts/dpo.py \
+    --model_name_or_path=gpt2 \
+    --per_device_train_batch_size 4 \
+    --max_steps 1000 \
+    --learning_rate 1e-3 \
+    --gradient_accumulation_steps 1 \
+    --logging_steps 10 \
+    --eval_steps 500 \
+    --output_dir="dpo_anthropic_hh" \
+    --optim rmsprop \
```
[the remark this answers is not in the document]
vwxyzjn: no need to specify, default is adamw_torch

examples/scripts/dpo.py
vwxyzjn: `get_tokenizer` would be nice! Let's do it in another PR because it relates to data processing as well (e.g., `truncation_side`)

```
def get_tokenizer(model_args: ModelArguments, data_args: DataArguments) -> PreTrainedTokenizer:
    """Get the tokenizer for the model."""
        model_args.model_name_or_path,
        revision=model_args.model_revision,
    )
    if tokenizer.pad_token_id is None:
        tokenizer.pad_token_id = tokenizer.eos_token_id

    if data_args.truncation_side is not None:
        tokenizer.truncation_side = data_args.truncation_side

    # Set reasonable default for models without max length
    if tokenizer.model_max_length > 100_000:
        tokenizer.model_max_length = 2048

    if data_args.chat_template is not None:
        tokenizer.chat_template = data_args.chat_template
    elif tokenizer.chat_template is None and tokenizer.default_chat_template is None:
        tokenizer.chat_template = DEFAULT_CHAT_TEMPLATE

    return tokenizer
```

vwxyzjn:
The CI failed but it seems unrelated to the change in this PR. See #1273

trl/trainer/utils.py
```diff
@@ -19,11 +19,17 @@
 
 import numpy as np
 import torch
+from accelerate import Accelerator
```
younesbelkada, line 22:
```suggestion
from accelerate import PartialState
```

trl/trainer/utils.py
```diff
@@ -676,3 +682,47 @@ def trl_sanitze_kwargs_for_tagging(model, tag_names, kwargs=None):
...
+    return quantization_config
+
+
+def get_kbit_device_map() -> Optional[Dict[str, int]]:
+    if is_xpu_available():
+        return {"": f"xpu:{Accelerator().local_process_index}"}
+    elif torch.cuda.is_available():
+        return {"": Accelerator().local_process_index}
+    else:
+        return None
```
younesbelkada, line 711:
```suggestion
        return {"": f"xpu:{PartialState().local_process_index}"}
    elif torch.cuda.is_available():
        return {"": PartialState().local_process_index}
    else:
        return None
```

docs/source/sft_trainer.mdx
```diff
@@ -426,13 +426,13 @@ To use Flash Attention 2, first install the latest `flash-attn` package:
 pip install -U flash-attn
 ```
 
-And add `use_flash_attention_2=True` when calling `from_pretrained`:
+And add `attn_implementation="flash_attention_2"` when calling `from_pretrained`:
```
younesbelkada, line 429: Could you add a documentation section here to explain what `ModelConfig` does and how to use it together with the other utility methods that you have exposed on main init? 🙏

trl/trainer/model_config.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+    lora_target_modules: Optional[List[str]] = field(
+        default=None,
+        metadata={"help": ("LoRA target modules.")},
+    )
+    lora_modules_to_save: Optional[List[str]] = field(
+        default=None,
+        metadata={"help": ("Model layers to unfreeze & train")},
+    )
+    load_in_8bit: bool = field(default=False, metadata={"help": "use 8 bit precision"})
```
younesbelkada, line 64:
```suggestion
    load_in_8bit: bool = field(default=False, metadata={"help": "use 8 bit precision for the base model - works only with LoRA"})
```

trl/trainer/model_config.py
```diff
@@ -0,0 +1,80 @@
...
+    load_in_4bit: bool = field(default=False, metadata={"help": "use 4 bit precision"})
```
younesbelkada, line 65:
```suggestion
    load_in_4bit: bool = field(default=False, metadata={"help": "use 4 bit precision for the base model - works only with LoRA"})
```

examples/scripts/sft.py
```diff
@@ -12,154 +12,104 @@
...
-    max_steps: Optional[int] = field(default=-1, metadata={"help": "the number of training steps"})
-    save_steps: Optional[int] = field(
-        default=100, metadata={"help": "Number of updates steps before two checkpoint saves"}
+    dataset_name: str = field(default="timdettmers/openassistant-guanaco", metadata={"help": "the dataset name"})
+    dataset_text_field: str = field(default="text", metadata={"help": "the text field of the dataset"})
+    max_seq_length: int = field(default=512, metadata={"help": "The maximum sequence length for SFT Trainer"})
+
+
+if __name__ == "__main__":
+    parser = HfArgumentParser((ScriptArguments, TrainingArguments, ModelConfig))
```
younesbelkada, line 68: Just to confirm - `--bf16` is passed directly through the terminal command and parsed thanks to the HfArgumentParser?

younesbelkada (commented):
Very nice clean up @vwxyzjn ! Thanks a lot for the refactor - just as a sanity check before merging, could you run `make test_examples` to make sure the example scripts will not fail on the GPU CI? Only DPO with DS-1 , 2 & 3 should fail but all other configurations should pass

trl/trainer/utils.py
younesbelkada: One should use `PartialState` whenever possible and not `Accelerator` in order to avoid initializing things that are not needed

examples/scripts/sft.py
vwxyzjn: that's correct. 

image

docs/source/sft_trainer.mdx
vwxyzjn: Done!

vwxyzjn:
Thanks @younesbelkada for the comment! `make test_examples` seems to launch 4 experiments and 2 of them failed with

```
RuntimeError: Expected tensor for argument #1 'indices' to have one of the following scalar types: Long, Int; but got torch.cuda.HalfTensor instead (while checking arguments for embedding)
[2024-01-26 14:57:32,456] torch.distributed.elastic.multiprocessing.api: [WARNING] Sending process 2212929 closing signal SIGTERM
```

Is this expected?

image

younesbelkada (approved):
Huge work @vwxyzjn ! Thanks so much for this refactor !

PR #1180, opened 2024-01-05
Fix reported KL in PPO trainer

### LINKED ISSUE #1161: Negative KL warning even if KL='abs' and KL='full'
I'm using PPOTrainer to RLHF a llama2 model, and I sometimes get negative KL in my experiments. I've noticed that PPOTrainer prints a warning about that, even if I set `kl_penalty="abs"` or `kl_penalty="full"`, even though (by my understanding) it should be impossible to get negative KL in those cases.

Two questions:
* Checking the code, it looks like the KL divergence that's reported to wandb, and that is also used to determine whether the "negative KL" warning is printed, is calculated separately from the KL that is used for training; and the reported KL is always using the standard estimate. Did I understand that correctly, and does that mean I can ignore the "negative KL" warning if I'm using "abs" or "full" KL? (I can see that the non-score reward is non-positive if I use abs or full KL, and if I'm reading the code correctly the does indicate that the KL used in training is likely positive.)
* Conceptually, is there any issue or drawback with using "abs" or "full" instead of "kl"?

Thank you so much!

mgerstgrasser:
PPO trainer currently separately calculates the KL that's reported in the stats and to wandb, and always uses the estimated KL for this. That's fine when the estimated KL is also used for training (`kl_penalty = 'kl'`), but leads to a mismatch when using `kl_penalty = 'abs'`, `kl_penalty = 'mse'`, or `kl_penalty = 'full'`. Additionally, this can lead to a warning about negative KL being triggered even in cases where the KL used in training is not and cannot be negative.

This PR changes it so that instead the KL that is calculated during training is kept and used for stats and reporting.

Tested using the PPO example script in both single-GPU and Deepspeed stage 2 training, leading to identical results in the `kl_penalty = 'kl'`case. Verified also that `kl_penalty = 'abs'` now does not trigger the negative KL warning anymore.

Closes #1161.

### DIFF
--- a/tests/test_ppo_trainer.py
+++ b/tests/test_ppo_trainer.py
@@ -579,7 +579,7 @@ class PPOTrainerTester(unittest.TestCase):
         logits = torch.exp(all_logprobs)
         vpreds = values + 0.1
 
-        score, non_score = ppo_trainer.compute_rewards(dummy_scores, all_logprobs, ref_logprobs, mask)
+        score, non_score, kls = ppo_trainer.compute_rewards(dummy_scores, all_logprobs, ref_logprobs, mask)
         values, advantages, returns = ppo_trainer.compute_advantages(values, score, mask)
 
         # just make sure a dummy loss is computed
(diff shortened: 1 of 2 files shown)


younesbelkada:
cc @vwxyzjn @lvwerra would you be able to have a look here whenever possible 🙏

lvwerra (approved):

younesbelkada:
thanks @vwxyzjn @lvwerra ! will merge then!

PR #1185, opened 2024-01-06
Fix instruction token masking

### LINKED ISSUE #1184: DataCollatorForCompletionOnlyLM instruction token masking fails if first occurence of instruction is marked differently
The current code in DataCollatorForCompletionOnlyLM assumes that the first deteced occurence of `instruction_template` comes before the first detected occurence of `response_template`. This is reasonable, since in current applications conversations are initiated by the user, not the assistant. However, this can fail if the first instruction is marked differently from all the other instructions, which can if a context-sensitive tokenizer such as Llama-2 tokenizes the instruction_template differently at the start of a string than in the middle.

In particular this happens in practice with TinyLlama: `<|user|>` gets tokenized as `529, 29989, 1792, 29989, 29958` at the start of a conversation, but as `29966, 29989, 1792, 29989, 29958` in later messages.

Reproduction snippet:
```
from trl import DataCollatorForCompletionOnlyLM

tokenizer = AutoTokenizer.from_pretrained("TinyLlama/TinyLlama-1.1B-Chat-v1.0")

chat = [
    {"role": "user", "content": "Which is bigger, the moon or the sun?"},
    {"role": "assistant", "content": "The sun."},
    {"role": "user", "content": "Really?"},
    {"role": "assistant", "content": "Yes."},
    {"role": "user", "content": "I don't believe you."},
    {"role": "assistant", "content": "That's okay."},
]

collator = DataCollatorForCompletionOnlyLM(response_template=[29966, 29989, 465, 22137, 29989, 29958, 13], instruction_template=[29871, 13, 29966, 29989, 1792, 29989, 29958], tokenizer=tokenizer)


collator([tokenizer.apply_chat_template(chat)])
# This doesn't mask user messages 2 & 3.
```

PR #1185 fixes this, and makes the above snippet mask out all the user messages correctly.

mgerstgrasser:
The current code in DataCollatorForCompletionOnlyLM assumes that the first deteced occurence of `instruction_template` comes before the first detected occurence of `response_template`. This is reasonable, since in current applications conversations are initiated by the user, not the assistant. However, this can fail if the first instruction is marked differently from all the other instructions, which can if a context-sensitive tokenizer such as Llama-2 tokenizes the instruction_template differently at the start of a string than in the middle.

This PR fixes this by checking if the first detected instruction is after the first detected response. If that is the case, we can assume we missed the first instruction. We then insert a new `human_token_ids_idxs` starting at 0.

Fixes #1184.

### DIFF
--- a/tests/test_data_collator_completion_only.py
+++ b/tests/test_data_collator_completion_only.py
@@ -31,11 +31,14 @@ class DataCollatorForCompletionOnlyLMTester(unittest.TestCase):
         self.instruction_template = "\n### User:"
         self.response_template = "\n### Assistant:"
 
-        # GPT2Tokenizer: [198, 21017, 11787, 25] -> [11787, 25]
+        # GPT2Tokenizer: [198, 21017, 11787, 25] -> [21017, 11787, 25]
         # Llama2Tokenizer: [29871, 13, 2277, 29937, 4911, 29901] -> [2277, 29937, 4911, 29901]
+        # Note: If this test is ever switched to Llama2Tokenizer, this should be double checked,
+        # and possibly switched back to [2:] instead of [1:].
+        # With GPT2Tokenizer, [1:] is correct - we want the 21017 token included, which is ###.
         self.tokenized_instruction_w_context = self.tokenizer.encode(
             self.instruction_template, add_special_tokens=False
-        )[2:]
+        )[1:]
 
         # GPT2Tokenizer: [198, 21017, 15286, 25] -> [15286, 25]
         # Llama2Tokenizer: [29871, 13, 2277, 29937, 4007, 22137, 29901] -> [2277, 29937, 4007, 22137, 29901]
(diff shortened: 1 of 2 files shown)


younesbelkada (approved):
amazing work @mgerstgrasser !
Would you be happy to add a small test here: https://github.com/huggingface/trl/blob/main/tests/test_data_collator_completion_only.py with a test similar than the one you shared on the description of the linked issue?

mgerstgrasser:
> amazing work @mgerstgrasser ! Would you be happy to add a small test here: https://github.com/huggingface/trl/blob/main/tests/test_data_collator_completion_only.py with a test similar than the one you shared on the description of the linked issue?

I'll have a look.

younesbelkada:
Thank you very much @mgerstgrasser !

mgerstgrasser:
@younesbelkada

I've added a test! :) That checks that the unmasked text is exactly as expected. The other tests in there don't do that, it seems they just check that nothing crashes, but I don't really see how that would be helpful in this case, since there wasn't any crash to begin with. Let me know in case this isn't what you had in mind though.

Also, I had to slightly change the instruction template for the test, as the way it was done was cutting off the `###` from the instruction template, which shouldn't happen I think. (I think even the `\n` should be part of the template in this case, but that doesn't matter for the test.) I don't know if that would still work as intended with a Llama-2 tokenizer - I've left a note in the code to double check things in case the test is ever switched to that.

younesbelkada (approved):
Thanks a lot for the detailed explanation and fix!

PR #1188, opened 2024-01-07
SFTTrainer: follow args.remove_unused_columns

### LINKED ISSUE #1186: SFTTrainer removes dataset columns even when `remove_unused_columns = False`
I've noticed that SFTTrainer removes dataset columns before passing samples to the data collator, even when `remove_unused_columns` is set to `False` in the training arguments.

This happens here:
https://github.com/huggingface/trl/blob/main/trl/trainer/sft_trainer.py#L410

Any objections to adding a check for `args.remove_unused_columns is True` here?

mgerstgrasser:
I've noticed that SFTTrainer removes dataset columns before passing samples to the data collator, even when remove_unused_columns is set to False in the training arguments.

This happens here:
https://github.com/huggingface/trl/blob/main/trl/trainer/sft_trainer.py#L410

This PR adds a check to see if `args.remove_unused_columns` is set to `True` (the default). If not, it sets `remove_columns` to `None` in the call to `dataset.map()`, i.e. it doesn't remove other columns.

This can be useful if you'd like access to dataset features other than the tokenizer output in your data collator.

Closes #1186.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -258,6 +258,7 @@ class SFTTrainer(Trainer):
                 formatting_func,
                 num_of_sequences,
                 chars_per_token,
+                remove_unused_columns=args.remove_unused_columns if args is not None else True,
                 **dataset_kwargs,
             )
         if eval_dataset is not None:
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Amazing fix @mgerstgrasser ! Thanks a lot for your contribution ! 🚀  (Failing test is unrelated to this PR)

PR #1190, opened 2024-01-08
Allow separate devices for target/ref models.

jondurbin:
Ensure tensors are on accelerator device, allowing ref model to be running on a different device than the target model.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -795,6 +795,8 @@ class DPOTrainer(Trainer):
         else:
             ref_logratios = reference_chosen_logps - reference_rejected_logps
 
+        pi_logratios = pi_logratios.to(self.accelerator.device)
+        ref_logratios = ref_logratios.to(self.accelerator.device)
         logits = pi_logratios - ref_logratios
 
         # The beta is a temperature parameter for the DPO loss, typically something in the range of 0.1 to 0.5.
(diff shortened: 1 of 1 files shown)


trl/trainer/dpo_trainer.py
```diff
@@ -830,6 +832,9 @@ def dpo_loss(
                 f"Unknown loss type: {self.loss_type}. Should be one of ['sigmoid', 'hinge', 'ipo', 'kto_pair']"
             )
 
+        chosen_rewards = self.beta * (policy_chosen_logps.to(self.accelerator.device) - reference_chosen_logps.to(self.accelerator.device)).detach()
+        rejected_rewards = self.beta * (policy_rejected_logps.to(self.accelerator.device) - reference_rejected_logps.to(self.accelerator.device)).detach()
+
         chosen_rewards = self.beta * (policy_chosen_logps - reference_chosen_logps).detach()
         rejected_rewards = self.beta * (policy_rejected_logps - reference_rejected_logps).detach()
```
younesbelkada, line 839:
```suggestion
```
As pointed out by @kashif offline, this seems to be a duplicate of above no?

younesbelkada (commented):
Thanks for your contribution ! left a single comment, what do you think?

trl/trainer/dpo_trainer.py
jondurbin: Oops, forgot to clean up the original!

kashif (approved):
thanks for that!

PR #1193, opened 2024-01-08
Allow swapping PEFT adapters for target/ref model.

jondurbin:
I tend to use QLoRA for larger base model fine-tunes (30b+), but for the DPO phase I end up having a bit of a dilemma.

If I don't want to load the model twice, I need to merge the adapter into the base model, then create another adapter on top of the merged model (so the existing unload mechanism works).

There are two ways to merge, the way recommended by Tim Dettmers, in which you load the model in 4 bit, then upscale it back to bf16, then merge the weights in.  Using this method leads to 1-2% increase in benchmark scores pretty universally:
https://github.com/jondurbin/qlora/blob/main/qmerge.py

Or the standard merge where you don't prequantize the weights:
https://github.com/jondurbin/qlora/blob/main/merge.py

If you use that quant->dequant->merge method, it means you've already quantized the base model weights, so doing another qlora on top of that with another quantized merge seems like a bad idea.  If you use the other merge method, it means you're losing quality.

So instead of requiring a fine-tuned model in DPO as the base so you can unload the adapter, this allows loading the adapter twice, once for training and once for reference, e.g.:
```python
model = AutoModelForCausalLM.from_pretrained(...)
model = PeftModel.from_pretrained(model, ..., adapter_name="train_adapter")
model.load_adapter(..., adapter_name="ref_adapter") # duplicate adapter, different name
dpo_trainer = DPOTrainer(
    model,
    args=training_args,
    ...
    model_adapter_name="train_adapter",
    ref_adapter_name="ref_adapter",
)
```

DPOTrainer will then use `set_adapter` to swap between target/ref model.

This allows only small overhead in vram for loading two adapters instead of one, and doesn't require merging the adapter first (and again at the end) and doing two LoRAs.

### DIFF
--- a/docs/source/dpo_trainer.mdx
+++ b/docs/source/dpo_trainer.mdx
@@ -174,7 +174,7 @@ If you use the recommended approach, which quantizes the model, you're now in a 
 
 ### Using option 3 - load adapter twice
 
-To avoid the downsides with option 2, at the expensive of slightly increased VRAM, you can load your fine-tuned adapter into the model twice, with different names, and set the model/ref adapter names in DPOTrainer.
+To avoid the downsides with option 2, at the expense of slightly increased VRAM, you can load your fine-tuned adapter into the model twice, with different names, and set the model/ref adapter names in DPOTrainer.
 
 For example:
 ```python
--- a/docs/source/dpo_trainer.mdx
+++ b/docs/source/dpo_trainer.mdx
@@ -177,7 +177,7 @@ If you use the recommended approach, which quantizes the model, you're now in a 
 To avoid the downsides with option 2, at the expensive of slightly increased VRAM, you can load your fine-tuned adapter into the model twice, with different names, and set the model/ref adapter names in DPOTrainer.
 
 For example:
-```
+```python
 # Load the base model.
 bnb_config = BitsAndBytesConfig(
     load_in_4bit=True,
--- a/docs/source/dpo_trainer.mdx
+++ b/docs/source/dpo_trainer.mdx
@@ -170,7 +170,7 @@ As suggested by [Tim Dettmers](https://twitter.com/Tim_Dettmers/status/169465419
 
 You can also just merge the adapters the standard way without quantizing the base model, but then you have 1-2% reduced performance (and evidently, more issues with empty responses).
 
-If you use the recommended approach, which quantizes the model, you're now in a situation where to use QLoRA for DPO, you will need to re-quantize the merged model again, or use an unquantized merge with lower overall performance.
+If you use the recommended approach, which quantizes the model, you're now in a situation where to use QLoRA for DPO, you will need to re-quantize the merged model again or use an unquantized merge with lower overall performance.
 
 ### Using option 3 - load the adapter twice
 
--- a/docs/source/dpo_trainer.mdx
+++ b/docs/source/dpo_trainer.mdx
@@ -172,7 +172,7 @@ You can also just merge the adapters the standard way without quantizing the bas
 
 If you use the recommended approach, which quantizes the model, you're now in a situation where to use QLoRA for DPO, you will need to re-quantize the merged model again, or use an unquantized merge with lower overall performance.
 
-### Using option 3 - load adapter twice
+### Using option 3 - load the adapter twice
 
 To avoid the downsides with option 2, at the expense of slightly increased VRAM, you can load your fine-tuned adapter into the model twice, with different names, and set the model/ref adapter names in DPOTrainer.
 
--- a/docs/source/dpo_trainer.mdx
+++ b/docs/source/dpo_trainer.mdx
@@ -166,7 +166,7 @@ You have three main options (plus several variants) for how the reference model 
 
 ### Downsides to merging QLoRA before DPO (approach 2)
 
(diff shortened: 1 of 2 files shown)


trl/trainer/dpo_trainer.py
```diff
@@ -16,7 +16,7 @@
 import random
 import warnings
 from collections import defaultdict
-from contextlib import nullcontext
+from contextlib import nullcontext, contextmanager
```
kashif, line 19:
```suggestion
from contextlib import contextmanager, nullcontext
```

kashif: ah you did it awesome!

kashif (approved):

docs/source/dpo_trainer.mdx
```diff
@@ -156,6 +156,66 @@ dpo_trainer.train()
...
+
+### Downsides to merging QLoRA before DPO (approach 2)
+
+As suggested by [Tim Dettmers](https://twitter.com/Tim_Dettmers/status/1694654191325573456), the best option for merging QLoRA adapters is to first quantize the base model, merge adapter, then convert back to bf16.  Something similar to [this script](https://github.com/jondurbin/qlora/blob/main/qmerge.py)
+
+You can also just merge the adapters the standard way without quantizing the base model, but then you have 1-2% reduced performance (and evidently, more issues with empty responses).
+
+If you use the recommended approach, which quantizes the model, you're now in a situation where to use QLoRA for DPO, you will need to re-quantize the merged model again, or use an unquantized merge with lower overall performance.
+
+### Using option 3 - load adapter twice
```
kashif, line 175:
```suggestion
### Using option 3 - load the adapter twice
```

docs/source/dpo_trainer.mdx
```diff
@@ -156,6 +156,66 @@ dpo_trainer.train()
...
+
+To avoid the downsides with option 2, at the expensive of slightly increased VRAM, you can load your fine-tuned adapter into the model twice, with different names, and set the model/ref adapter names in DPOTrainer.
```
kashif, line 177:
```suggestion
To avoid the downsides with option 2, at the expense of slightly increased VRAM, you can load your fine-tuned adapter into the model twice, with different names, and set the model/ref adapter names in DPOTrainer.
```

docs/source/dpo_trainer.mdx
```diff
@@ -156,6 +156,66 @@ dpo_trainer.train()
...
+3. Load the adapter twice with different names, then use `set_adapter` during training to swap between the adapter being DPO'd and the reference adapter - slightly less efficient compared to 2 (~adapter size VRAM overhead), but avoids the pitfalls.
+
+### Downsides to merging QLoRA before DPO (approach 2)
+
+As suggested by [Tim Dettmers](https://twitter.com/Tim_Dettmers/status/1694654191325573456), the best option for merging QLoRA adapters is to first quantize the base model, merge adapter, then convert back to bf16.  Something similar to [this script](https://github.com/jondurbin/qlora/blob/main/qmerge.py)
+
+You can also just merge the adapters the standard way without quantizing the base model, but then you have 1-2% reduced performance (and evidently, more issues with empty responses).
+
+If you use the recommended approach, which quantizes the model, you're now in a situation where to use QLoRA for DPO, you will need to re-quantize the merged model again, or use an unquantized merge with lower overall performance.
```
kashif, line 173:
```suggestion
If you use the recommended approach, which quantizes the model, you're now in a situation where to use QLoRA for DPO, you will need to re-quantize the merged model again or use an unquantized merge with lower overall performance.
```

docs/source/dpo_trainer.mdx
```diff
@@ -156,6 +156,66 @@ dpo_trainer.train()
...
+model.config.use_cache = False
+
+# Load the adapter.
+model = PeftModel.from_pretrained(
+    model,
+    "/path/to/peft",
+    is_trainable=True,
+    adapter_name="train",
+)
+# Load the adapter a second time, with a differet name, which will be our reference model.
```
kashif, line 207:
```suggestion
# Load the adapter a second time, with a different name, which will be our reference model.
```

docs/source/dpo_trainer.mdx
```diff
@@ -156,6 +156,66 @@ dpo_trainer.train()
...
+
+You have three main options (plus several variants) for how the reference model works when using PEFT, assuming the model that you would like to further enhance with DPO was tuned using (Q)LoRA.
+
+1. Simply create two instances of the model, each loading your adapter - works fine but is very inefficient.
+2. Merge the adapter into the base model, create another adapter on top, then leave the `model_ref` param null, in which case DPOTrainer will unload the adapter for reference inference - efficient, but has potential downsides discussed below.
+3. Load the adapter twice with different names, then use `set_adapter` during training to swap between the adapter being DPO'd and the reference adapter - slightly less efficient compared to 2 (~adapter size VRAM overhead), but avoids the pitfalls.
+
+### Downsides to merging QLoRA before DPO (approach 2)
+
+As suggested by [Tim Dettmers](https://twitter.com/Tim_Dettmers/status/1694654191325573456), the best option for merging QLoRA adapters is to first quantize the base model, merge adapter, then convert back to bf16.  Something similar to [this script](https://github.com/jondurbin/qlora/blob/main/qmerge.py)
```
kashif, line 169:
```suggestion
As suggested by [Tim Dettmers](https://twitter.com/Tim_Dettmers/status/1694654191325573456), the best option for merging QLoRA adapters is to first quantize the base model, merge the adapter, then convert back to bf16.  Something similar to [this script](https://github.com/jondurbin/qlora/blob/main/qmerge.py)
```

kashif (approved):

kashif:
Thanks @jondurbin

PR #1196, opened 2024-01-09
Fix FSDP error

### LINKED ISSUE #1195: RewardTrainer fails with FSDP
I've just run into an odd issue with FSDP & RewardTrainer. It seems then when using FSDP, the output of the (sequence classification) model's `forward` function isn't as expected. 
Normally, it returns a `SequenceClassifierOutputWithPast` where `logits` contains a tensor with the logits, and `loss` is empty or contains some sort of generator object..
When using FSDP, I'm getting a `dict` inside the loss field (and oddly enough that dict again contains a single key `logits`, althouh that's not the issue).

Not sure why this happens, but the net effect is that when the RewardTrainer tries to get the logits through `model(...)[0]` (see [here](https://github.com/huggingface/trl/blob/3267be0fcd424c8d224d7f28a45a64358c857ed0/trl/trainer/reward_trainer.py#L220)), in the non-FSDP case it gets the logits, while in the FSDP case it gets the dict from the now non-empty`loss` field, and then fails a few lines later.

Two questions:
1. This is easily fixed by doing `model(...)["logits"]` instead. Any problem with doing that?
2. Purely out of curiosity, does anyone know why this behaves differently with FSDP?

To reproduce: Run `examples/scripts/reward_modeling.py` with accelerate + FSDP.

`forward` output in a single process:
```
SequenceClassifierOutputWithPast(loss=<generator object gather.<locals>.gather_map.<locals>.<genexpr> at 0x15360f993040>, logits=tensor([[...]], device='cuda:0', grad_fn=<GatherBackward>), past_key_values=None, hidden_states=None, attentions=None)
```

And in FSDP:
```
SequenceClassifierOutputWithPast(loss={'logits': tensor([[...]], device='cuda:1', grad_fn=<ToCopyBackward0>)}, logits=tensor([[...]], device='cuda:1', grad_fn=<ToCopyBackward0>), past_key_values=None, hidden_states=None, attentions=None)
```

mgerstgrasser:
This fixes an error in reward training when the `loss` field of model output is non-empty, and `model(...)[0]` returns loss instead of logits. That can happen with FSDP. This PR changes it so the trainer explicitly grabs `model(...)["logits"]` instead.

Closes #1195

### DIFF
--- a/trl/trainer/reward_trainer.py
+++ b/trl/trainer/reward_trainer.py
@@ -220,11 +220,11 @@ class RewardTrainer(Trainer):
         rewards_chosen = model(
             input_ids=inputs["input_ids_chosen"],
             attention_mask=inputs["attention_mask_chosen"],
-        )[0]
+        )["logits"]
         rewards_rejected = model(
             input_ids=inputs["input_ids_rejected"],
             attention_mask=inputs["attention_mask_rejected"],
-        )[0]
+        )["logits"]
         # calculate loss, optionally modulate with margin
         if "margin" in inputs:
             loss = -nn.functional.logsigmoid(rewards_chosen - rewards_rejected - inputs["margin"]).mean()
--- a/trl/trainer/reward_trainer.py
+++ b/trl/trainer/reward_trainer.py
@@ -220,10 +220,12 @@ class RewardTrainer(Trainer):
         rewards_chosen = model(
             input_ids=inputs["input_ids_chosen"],
             attention_mask=inputs["attention_mask_chosen"],
+            return_dict=True,
         )["logits"]
         rewards_rejected = model(
             input_ids=inputs["input_ids_rejected"],
             attention_mask=inputs["attention_mask_rejected"],
+            return_dict=True,
         )["logits"]
         # calculate loss, optionally modulate with margin
         if "margin" in inputs:


trl/trainer/reward_trainer.py
```diff
@@ -220,11 +220,11 @@ def compute_loss(
         rewards_chosen = model(
             input_ids=inputs["input_ids_chosen"],
             attention_mask=inputs["attention_mask_chosen"],
```
younesbelkada, line 222:
```suggestion
            attention_mask=inputs["attention_mask_chosen"],
            return_dict=True,
```

trl/trainer/reward_trainer.py
```diff
@@ -220,11 +220,11 @@ def compute_loss(
...
-        )[0]
+        )["logits"]
         rewards_rejected = model(
             input_ids=inputs["input_ids_rejected"],
             attention_mask=inputs["attention_mask_rejected"],
```
younesbelkada, line 226:
```suggestion
            attention_mask=inputs["attention_mask_rejected"],
            return_dict=True,
```

younesbelkada (commented):
Thanks a lot for the fix, I propose to force-set `return_dict=True` in case users train their RM with `return_dict` being explicitly set to False in the config. Can you try that out? 🙏

mgerstgrasser:
> Thanks a lot for the fix, I propose to force-set `return_dict=True` in case users train their RM with `return_dict` being explicitly set to False in the config. Can you try that out? 🙏

Yes, that works! Thanks for spotting that.

younesbelkada (approved):
Thanks again for all your contributions @mgerstgrasser ! You are on fire!

Issue #1202, opened 2024-01-09
After upgrade from 0.7.5 to 0.7.8 DPOTrainer get RuntimeError: expected scalar type Float but found BFloat16


Andcircle:
After upgrade from 0.7.5 to 0.7.8 DPOTrainer get RuntimeError: expected scalar type Float but found BFloat16

I'm using KTO loss, haven't tried other loss yet

```
Traceback (most recent call last):
  File "cpt-hf/train/train_dpo.py", line 307, in <module>
    trainer.train()
  File "/home/user/.local/lib/python3.8/site-packages/transformers/trainer.py", line 1537, in train
    return inner_training_loop(
  File "/home/user/.local/lib/python3.8/site-packages/transformers/trainer.py", line 1914, in _inner_training_loop
    self._maybe_log_save_evaluate(tr_loss, model, trial, epoch, ignore_keys_for_eval)
  File "/home/user/.local/lib/python3.8/site-packages/transformers/trainer.py", line 2268, in _maybe_log_save_evaluate
    metrics = self.evaluate(ignore_keys=ignore_keys_for_eval)
  File "/home/user/.local/lib/python3.8/site-packages/transformers/trainer.py", line 3019, in evaluate
    output = eval_loop(
  File "/home/user/.local/lib/python3.8/site-packages/trl/trainer/dpo_trainer.py", line 1163, in evaluation_loop
    policy_output_decoded, ref_output_decoded = self.get_batch_samples(self.model, random_batch)
  File "/home/user/.local/lib/python3.8/site-packages/trl/trainer/dpo_trainer.py", line 1056, in get_batch_samples
    policy_output = model.generate(
  File "/usr/local/lib/python3.8/dist-packages/peft/peft_model.py", line 988, in generate
    outputs = self.base_model.generate(**kwargs)
  File "/usr/local/lib/python3.8/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/user/.local/lib/python3.8/site-packages/transformers/generation/utils.py", line 1764, in generate
    return self.sample(
  File "/home/user/.local/lib/python3.8/site-packages/transformers/generation/utils.py", line 2861, in sample
    outputs = self(
  File "/usr/local/lib/python3.8/dist-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
[...]

PR #1143, opened 2023-12-26
add peft_module_casting_to_bf16 in DPOTrainer

sywangyi:

PR #1203, opened 2024-01-09
[`DPOTrainer`] Fix peft + DPO + bf16 if one uses `generate_during_eval` or pre-computed logits

younesbelkada:
# What does this PR do ?

Fixes https://github.com/huggingface/trl/issues/1202 which is a regression 

https://github.com/huggingface/trl/pull/1143 introduced `peft_module_casting_to_bf16` for `DPOTrainer` which worked fine when ` generate_during_eval` is set to `False`.

`peft_module_casting_to_bf16` will cast the LayerNorm layers in fp32 for stability purposes for smoother training, in case one uses peft + 4-bit quantization (QLoRA).

If one uses `bf16=True` in the `TrainingArguments`, `Trainer` will automatcially compute the loss under the torch.cuda.amp.autocast regime, making the forward pass possible to work without any dtype mismatch issue

However, we need to make sure all other methods inside `DPOTrainer` that call the model's forward pass will also use that context manager to make sure there is no dtype mismatch

since this is a regression, will publish a patch release after merging this pr

cc @kashif @pacman100 @lvwerra

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -641,9 +641,9 @@ def peft_module_casting_to_bf16(model):
     for name, module in model.named_modules():
         if isinstance(module, BaseTunerLayer):
             module = module.to(torch.bfloat16)
-        if "norm" in name:
+        elif isinstance(module, torch.nn.LayerNorm):
             module = module.to(torch.float32)
-        if any(x in name for x in ["lm_head", "embed_tokens", "wte", "wpe"]):
+        elif any(x in name for x in ["lm_head", "embed_tokens", "wte", "wpe"]):
             if hasattr(module, "weight"):
                 if module.weight.dtype == torch.float32:
                     module = module.to(torch.bfloat16)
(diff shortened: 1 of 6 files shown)


trl/trainer/utils.py
```diff
@@ -641,9 +641,9 @@ def peft_module_casting_to_bf16(model):
     for name, module in model.named_modules():
         if isinstance(module, BaseTunerLayer):
             module = module.to(torch.bfloat16)
-        if "norm" in name:
+        elif isinstance(module, torch.nn.LayerNorm) or "norm" in name:
```
younesbelkada, line 644: @pacman100 is it correct to assume that you want to cast here all layernorms?

kashif (approved):

kashif:
thanks @younesbelkada awesome!

trl/trainer/utils.py
pacman100: Hello Younes, yes.

PR #1204, opened 2024-01-09
Fix shape descriptions in calculate_loss method

yuta0x89:
The content of `latents` is defined in the following code in `modeling_sd_base.py`:

```python
latents = self.prepare_latents(
    batch_size * num_images_per_prompt,
    num_channels_latents,
    height,
    width,
    prompt_embeds.dtype,
    device,
    generator,
    latents,
)
```

As a result, the shape of `latents` and `next_latents` mentioned in the description of the `calculate_loss` method in the `DDPOTrainer` class should be `[batch_size, num_channels_latents, height, width]`.

### DIFF
--- a/trl/trainer/ddpo_trainer.py
+++ b/trl/trainer/ddpo_trainer.py
@@ -343,11 +343,11 @@ class DDPOTrainer(BaseTrainer):
 
         Args:
             latents (torch.Tensor):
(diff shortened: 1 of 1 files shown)


lvwerra:
Letting @sayakpaul have a look too.

sayakpaul (approved):

PR #1208, opened 2024-01-09
Add support for ChatML dataset format in

philschmid:
# What does this PR do? 

This PR adds support for a standardized dataset to be automatically formated for training in the `SFTTrainer` using the `apply_chat_template` from `transformers`. 
This allow users to pass the dataset without the need of a `formatting_func` to the `SFTTrainer`. Example below

```Python

dataset = load_dataset("philschmid/dolly-15k-oai-style", split="train")

trainer = SFTTrainer(
    model=model,
    args=training_args,
    train_dataset=dataset,
    packing=True,
)
```

In the init method the SFTTrainer tries to finds the correct formatting function based on the dataset structure. Currently supported datasets are:
    - `ChatML` with [{"role": str, "content": str}]
    - `instruction` with [{"prompt": str, "completion": str}]

Based on the dataset it returns a callable which uses the `tokenizer` of the model and the corresponding `apply_chat_template` method. This allows continues fine-tuning for, e.g. Llama2-chat or other models which already have a defined format. 

The nice part about is that you can use the "extras" outside of the SFFTrainer, e.g. if you want to format DPO datasets with the methods.


trl/extras/dataset_formatting.py
```diff
@@ -0,0 +1,77 @@
...
+        if "messages" in dataset.features:
+            if dataset.features["messages"] == FORMAT_MAPPING["chatml"]:
+                return conversations_formatting_function(tokenizer, "messages")
+        if "conversations" in dataset.features:
+            if dataset.features["conversations"] == FORMAT_MAPPING["chatml"]:
+                return conversations_formatting_function(tokenizer, "conversations")
+        elif dataset.features == FORMAT_MAPPING["instruction"]:
+            return instructions_formatting_function
+        else:
+            logging.warning("Could not find a formatting function for the dataset. Please check the dataset format.")
```
younesbelkada, line 75: I think this should raise a ValueError, what do you think?

trl/extras/dataset_formatting.py
```diff
@@ -0,0 +1,77 @@
...
+
+    return None
```
younesbelkada, line 77:
```suggestion
```

younesbelkada (commented):
Amazing addition, thanks! I went through the PR and it looks great to me! 
Can you add few lines in the documentaiton to explain what it does under the hood? I think the doc section should live inside the SFTTrainer docs - wdyt?

trl/extras/dataset_formatting.py
philschmid: I didn't wanted to raise an Error since other parts in the init of the Trainer are also checks and would error anyway? But i can add it if you think it would improve the UX.

trl/extras/dataset_formatting.py
philschmid: You rather make the "return None" implict?

philschmid:
> Can you add few lines in the documentaiton to explain what it does under the hood? I think the doc section should live inside the SFTTrainer docs - wdyt?

I didn't worked on the documentation yet, since i wanted to see what you think. Will work on the docs next.

trl/extras/dataset_formatting.py
younesbelkada: I was thinking if we raise the value error above that return None would not be needed anymore

trl/extras/dataset_formatting.py
younesbelkada: Hmmm thinking more about it I think you are right, let's keep it as is. If we error out here the other value error of the sftrainer init will not be triggered (which states a clear error message)

trl/extras/dataset_formatting.py
younesbelkada: Nevermind, I think we should not raise a value error above so all good IMO

younesbelkada: yes correct ! sounds great to me

docs/source/sft_trainer.mdx
```diff
@@ -154,6 +154,48 @@ response_template_ids = tokenizer.encode(response_template_with_context, add_spe
...
+{"messages": [{"role": "system", "content": "You are helpful"}, {"role": "user", "content": "How far is the Moon from Earth?"}, {"role": "assistant", "content": "..."}]}
+```
+* instruction format
+```json
+{"prompt": "<prompt text>", "completion": "<ideal generated text>"}
+{"prompt": "<prompt text>", "completion": "<ideal generated text>"}
+{"prompt": "<prompt text>", "completion": "<ideal generated text>"}
+```
+
+If your dataset uses one of the above formats, you can directly pass it to the trainer without pre-processing. The [`SFTTrainer`] will then format the dataset for you using the defined format from the model's tokenizer with the [apply_chat_template](https://huggingface.co/docs/transformers/main/en/chat_templating#templates-for-chat-models) method.
```
younesbelkada, line 173: Maybe worth adding a line saying that the user need to make sure that tokenizer support apply_chat_template - otherwise it'll fail I think no?

younesbelkada (approved):
Looking great to me thanks! I just left one question for the documentation

docs/source/sft_trainer.mdx
younesbelkada: ok perfect then!

Rocketknight1: Yep, the default format for tokenizers with no `chat_template` or class-level `default_chat_template` is ChatML.

docs/source/sft_trainer.mdx
```diff
@@ -154,6 +154,48 @@ response_template_ids = tokenizer.encode(response_template_with_context, add_spe
...
+    "facebook/opt-350m",
+    args=training_args,
+    train_dataset=dataset,
+    packing=True,
+)
+```
+
+
+
 ### Format your input prompts
```
lvwerra, line 199: maybe we can make a bridge to the next section:

> If the dataset is not in one either of those format you can either preprocess the dataset to match the formatting or pass a formatting function to the `SFTTrainer` to do it for you. Let's have a look.

lvwerra (approved):
Thanks a lot @philschmid! Overall very clean, just left a few small nits.

tests/test_sft_trainer.py
```diff
@@ -85,6 +85,52 @@ def setUpClass(cls):
                 ],
             }
         )
+        cls.dummy_chatml_dataset = Dataset.from_dict(
+            {
+                "messages": [
+                    [
+                        {"role": "user", "content": "Hello"},
```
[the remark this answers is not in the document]
lvwerra: and let's test the instruction format too?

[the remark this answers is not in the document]
philschmid: yes changed all tests to have system messages and and SFTTrainer test for instruct dataset.

younesbelkada (approved):
Still looks really good to me! Is it ok for merge @philschmid ?

PR #1213, opened 2024-01-10
Update Unsloth SFT, DPO docs

danielhanchen:
@younesbelkada :)


younesbelkada (approved):
Thanks a lot for making the TRL docs better with respect to unsloth usage!

PR #1214, opened 2024-01-10
Fix args type

zspo:
1. adjust the parameter description.
2. Correct the parameter type in the comment.


younesbelkada (approved):
Thanks!

PR #1215, opened 2024-01-10
[`core` / `Docker`] Add  workflow to build TRL docker images

younesbelkada:
# What does this PR do?

As discussed internally, this is a first step towards creating our CI system that will run on self-hosted runners on GPU. I want to test everything on two scenarios:
- transformers / accelerate / peft all installed from source
- all the libraries above installed from latest pypi 
We currently use similar docker images in transformers & peft and this is very convenient for testing purpose and tracking down where the issues do come from. 
The dockerfile files are mostly copied from PEFT: https://github.com/huggingface/peft/blob/main/docker/peft-gpu/Dockerfile

cc @lvwerra

### DIFF
--- a/docker/trl-latest-gpu/Dockerfile
+++ b/docker/trl-latest-gpu/Dockerfile
@@ -1,0 +1,66 @@
+# Builds GPU docker image of PyTorch
+# Uses multi-staged approach to reduce size
+# Stage 1
+# Use base conda image to reduce time
+FROM continuumio/miniconda3:latest AS compile-image
+# Specify py version
+ENV PYTHON_VERSION=3.8
+# Install apt libs - copied from https://github.com/huggingface/accelerate/blob/main/docker/accelerate-gpu/Dockerfile
+RUN apt-get update && \
+    apt-get install -y curl git wget software-properties-common git-lfs && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists*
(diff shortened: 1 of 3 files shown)


docker/trl-latest-gpu/Dockerfile
```diff
@@ -0,0 +1,66 @@
+# Builds GPU docker image of PyTorch
+# Uses multi-staged approach to reduce size
+# Stage 1
+# Use base conda image to reduce time
+FROM continuumio/miniconda3:latest AS compile-image
+# Specify py version
+ENV PYTHON_VERSION=3.8
```
younesbelkada, line 7:
```suggestion
ENV PYTHON_VERSION=3.10
```

docker/trl-source-gpu/Dockerfile
```diff
@@ -0,0 +1,66 @@
+# Builds GPU docker image of PyTorch
+# Uses multi-staged approach to reduce size
+# Stage 1
+# Use base conda image to reduce time
+FROM continuumio/miniconda3:latest AS compile-image
+# Specify py version
+ENV PYTHON_VERSION=3.8
```
younesbelkada, line 7:
```suggestion
ENV PYTHON_VERSION=3.10
```

lvwerra (approved):

PR #1221, opened 2024-01-12
Refactor RewardConfig to own module

lewtun:
Based on discussion here: https://github.com/huggingface/trl/pull/1181#discussion_r1445606450

Basic idea is to have have a dedicated config per trainer, each in their own module. Since the code path was never exposed as an absolute path, I believe this is backwards compatible

### DIFF
--- a/trl/trainer/reward_config.py
+++ b/trl/trainer/reward_config.py
@@ -1,6 +1,4 @@
-# coding=utf-8
-# coding=utf-8
-# Copyright 2023 The HuggingFace Team. All rights reserved.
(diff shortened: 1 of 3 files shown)


trl/trainer/reward_config.py
```diff
@@ -1,6 +1,4 @@
-# coding=utf-8
-# coding=utf-8
-# Copyright 2023 The HuggingFace Team. All rights reserved.
+# Copyright 2024 The HuggingFace Team. All rights reserved.
```
younesbelkada, line 1: Nice :D !

younesbelkada (approved):
Makes sense, thanks !

PR #1226, opened 2024-01-14
Remove a repeating line in how_to_train.md

kykim0:


younesbelkada (approved):
Thanks for fixing the repeating line !

Issue #1216, opened 2024-01-11
Failed to load data in trl 0.7.8/0.7.9.


xkszltl:
This is a new regression introduced in trl 0.7.8 (and 0.7.9), 0.7.7 is fine.

We run into issues of `ValueError: too many dimensions 'str'` when loading data to the trainer.
Here's a simple LLAMA2+LoRA fine-tuning on IMDB dataset as minimal repro:

```py3
#!/usr/bin/env python3

import datasets
import peft
import transformers
import trl


model_dir = "models/Llama-2-7b-hf"

tokenizer = transformers.AutoTokenizer.from_pretrained(model_dir)
tokenizer.pad_token = tokenizer.eos_token
tokenizer.padding_side = "right"

model = transformers.AutoModelForCausalLM.from_pretrained(model_dir)

ds_train = datasets.load_dataset("imdb", split="train[:10]")

trainer = trl.SFTTrainer(
    model=model,
    args=transformers.TrainingArguments(
        output_dir="output",
        max_steps=1,
        remove_unused_columns=False,
    ),
    peft_config=peft.LoraConfig(
        lora_alpha=16,
        lora_dropout=0.1,
        r=8,
        bias="none",
        task_type="Causal_LM",
    ),
    train_dataset=ds_train,
    tokenizer=tokenizer,
    dataset_text_field="text",
    max_seq_length=8,
)
trainer.train()
```

0.7.7 works:
```py3
# CUDA_VISIBLE_DEVICES=0 ./test.py 
/usr/lib/python3/dist-packages/requests/__init__.py:87: RequestsDependencyWarning: urllib3 (2.1.0) or chardet (4.0.0) doesn't match a supported version!
  warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported "
Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:03<00:00,  1.52s/it]
Detected kernel version 3.10.0, which is below the recommended minimum of 5.5.0; this can cause the process to hang. It is recommended to upgrade the kernel to the minimum version or higher.
  0%|                                                                                                                                                                                                                                 | 0/1 [00:00<?, ?it/s]You're using a LlamaTokenizerFast tokenizer. Please note that with a fast tokenizer, using the `__call__` method is faster than using a method to encode the text followed by a call to the `pad` method to get a padded encoding.
[...]

PR #1229, opened 2024-01-15
[`core` / SFTTrainer] Fix breaking change

younesbelkada:
# What does this PR do?

fixes a breaking change on a recent PR: https://github.com/huggingface/trl/pull/1188 
Fixes: https://github.com/huggingface/trl/issues/1216 

In fact, if one uses the default data collator + `remove_unused_columns=False`, the training script fails for some datasets such as imdb as the collator will try to encode all the columns, including `"text"`. The fix should be to force-set remove_unused_columns=True, only in the case users do not pass a data_collator for non-packing datasets and explain to the users who they should proceed in case they decided to pass `remove_unused_columns=False`

cc @lvwerra

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -36,7 +36,6 @@ from transformers.modeling_utils import unwrap_model
 from transformers.trainer_callback import TrainerCallback
 from transformers.trainer_utils import EvalPrediction
 
-from ..extras.dataset_formatting import get_formatting_func_from_dataset
 from ..import_utils import is_peft_available
 from .utils import (
     ConstantLengthDataset,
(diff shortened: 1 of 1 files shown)


xkszltl:
Thanks for the action. 
Not sure about details yet, just a general comment: 

> only in the case users do not pass a data_collator for non-packing datasets

Long special condition may not be a good design and may not be accurate either.

younesbelkada:
I agree, the thing is that :
1- in previous releases it was possible to pass `remove_unused_columns=False` to trainingarguments + non-packed dataset without any problem - however we were silently ignoring the unusued columns - and #1188 breaks it
2- passing `remove_unused_columns=False` + default data collator will most likely fail before the default datacollator will try to tokenize text columns by default, so it'll always fail in that scenario. IMO it should be bad intent and we should leave the support of that feature only for users that they know what they're doing, by subclassing the default collator and passing it to SFTTrainer 
What do you think?

xkszltl:
2. Wait, maybe I didn't got it correctly in the first place? My intention is to tokenize the text column so that it can be trained with, is that supposed to be the case?

younesbelkada:
tokenizing the `text` column should be already handled properly by the `SFTTrainer`, when you pass `dataset_text_field="text"` it will tokenize the `text` field and remove the `text` column after tokenizing it

xkszltl:
(I could be using it in the totally wrong way because it was just used as a mitigation of the other issue I mentioned.)

trl/trainer/sft_trainer.py
```diff
@@ -249,6 +249,13 @@ def make_inputs_require_grad(module, input, output):
                     "You passed `packing=False` to the SFTTrainer, but you didn't pass a `dataset_text_field` or `formatting_func` argument."
                 )
 
+            if args is not None and not args.remove_unused_columns and data_collator is None:
+                warnings.warn(
+                    "You passed `remove_unused_columns=False` on a non-packed dataset while using `DataCollatorForLanguageModeling`. This might create some issues with the default collator. If you want to "
+                    "inspect dataset other columns, you can subclass `DataCollatorForLanguageModeling` and create your own data collator in order to inspect the unused dataset columns."
+                )
+                args.remove_unused_columns = True
```
lvwerra, line 257: I am not sure this is the right approach here, potentially overwriting a user choice (`args.remove_unused_columns = False`) or warning when no warning is needed (if there are no extra columns).

lvwerra: Maybe we can check if extra columns exist and suggesting setting removed unused columns to `True` if that's the case?

lvwerra (commented):
Thanks for fixing @younesbelkada! Left a comment regarding the warning logic.

lvwerra (approved):

PR #1234, opened 2024-01-16
Fix Padding Behaviour

nivibilla:
Hi @younesbelkada , just moving the PR from #1194 to here, I accidentally made my changes in main lol.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -410,7 +410,7 @@ class SFTTrainer(Trainer):
                 element[dataset_text_field] if not use_formatting_func else formatting_func(element),
                 add_special_tokens=add_special_tokens,
                 truncation=True,
-                padding=False,
+                padding="max_length",
                 max_length=max_seq_length,
                 return_overflowing_tokens=False,
                 return_length=False,


younesbelkada:
hi @nivibilla 
Thanks for the PR! 
i discussed offline with @lvwerra abotut these changes, would it be possible to send over a small reproducer of the issue you are facing? Perhaps the fix could be on the collator side but I am not sure.

nivibilla:
Hi Younes, the dataset I used was proprietary but I'll try get you a Colab with the issue

younesbelkada:
Ok that would be great! 🙏

nivibilla:
Hi, not sure what changed but I'm unable to reproduce the issue. Even on my old code with the dataset I originally got the issue with.

Here is the [colab](https://colab.research.google.com/drive/1sVRk84r1XzO_34sZRyZIqsZIENmZHDSd?usp=sharing) anyway. Main thing is to set the max_length to be > lengths of the samples in training. But it seems to work fine. Ive tested it with tinyllama on colab and tinymix(mixtral arch) personally. 

Im happy to close this pr since there seems to be no issue with the current code.

younesbelkada:
Ok thanks for the further investigation ! Yes let's close it for now - happy to make sure  your other PR gets landed in TRL !

Issue #1230, opened 2024-01-15
Packing without cross contamination 


nivibilla:
Related to #1194 , using packing deteriorated performance as  samples in my dataset are not independent. And the correlation might have caused the issue. However packing did help my training as it was almost 4x as fast. (My samples are very small in length ~ 512 tokens, so with 2k context my dataset size shrinks by 4x)

There is an implementation [here](https://github.com/MeetKai/functionary/tree/main/functionary/train/packing) of packing without cross contamination. Can this be added to trl?

It would solve the issue of contaminating the training. But also reduce training time.

younesbelkada:
This would be a great addition indeed ! I am labeling the issue as a good second issue - @nivibilla do you think you would have time to work on a PoC PR to add this feature? 
In terms of API, we could just add a flag `filter_contamination=True` in `SFTTrainer`'s init - wdyt @lvwerra ?

nivibilla:
Sure I can try do it.

younesbelkada:
awesome, let us know with @lvwerra if you have any question!

qgallouedec:
#1235 was not merged because

> it depends on monkey patching

see https://github.com/huggingface/trl/pull/1235#issuecomment-1904180330

As this feature seems to be demanded by the community, we leave this issue open, and encourage contributions for implementation.

qgallouedec:
This is now resolved. Just use `packing=True` with flash attention 2/3

PR #1235, opened 2024-01-16
Non Contaminated Packing

nivibilla:
As discussed in #1230 , I've done a quick & dirty implementation. And also included a sample notebook(not tested).

Will test when I can. Or if you have time pls feel free to test and also any feedback on code changes. Putting this as draft for the moment.

Thanks!


nivibilla:
cc @younesbelkada @lvwerra

nivibilla:
Note it only works with flash attention enabled.

nivibilla:
getting this error
`ValueError: The model did not return a loss from the inputs, only the following keys: logits. For reference, the inputs it received are input_ids,attention_mask.`

nivibilla:
colab [here](https://drive.google.com/file/d/1fSyrSFL9ZyJS69R_lmEW5Hy_Dqi97awt/view?usp=sharing), not sure what I'm doing wrong. @younesbelkada could you pls take a look

trl/trainer/sft_trainer.py
```diff
@@ -390,6 +401,10 @@ def _prepare_dataset(
                 append_concat_token,
                 add_special_tokens,
             )
+        elif non_contaminated_packing:
```
khai-meetkai, line 404: I think we should move **checking** `non_contaminated_packing` this to line 380 or we will never reach this line.
Should be like this:

```
if non_contaminated_packing:
      return self._prepare_non_contaminated_packed_dataloader(
          tokenizer, dataset, dataset_text_field, max_seq_length, add_special_tokens
      )

if not packing:
      return self._prepare_non_packed_dataloader(
          tokenizer,
          dataset,
          dataset_text_field,
          max_seq_length,
          formatting_func,
          add_special_tokens,
          remove_unused_columns,
      )

else:
      return self._prepare_packed_dataloader(
          tokenizer,
          dataset,
          dataset_text_field,
          max_seq_length,
          num_of_sequences,
          chars_per_token,
          formatting_func,
          append_concat_token,
          add_special_tokens,
      )
```

nivibilla: Ah yes of course, thanks!

nivibilla:
Thanks @khai-meetkai for fixing the bug, training loop runs on colab, however since t4 doesn't support flash attention, cant test loss. Will test on aws cluster tomorrow and then mark as ready.

image

trl/trainer/monkey_patch_packing.py
younesbelkada, line 1: I don't think we can add monkey patching utility methods here, do you think we can get non-contaminated packing without FA-2 and we leave it to super-users about monkey patching outside TRL ?

younesbelkada:
Thanks @khai-meetkai for the fix! @nivibilla I left a comment, wdyt?

trl/trainer/monkey_patch_packing.py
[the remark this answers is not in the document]
nivibilla: The authors from functionary state that the method is only correct when flash attention is enabled.

nivibilla: The monkey patching is needed to change the way the attention is calculated when using the new packed dataset. I just implemented the code from the functionary repo into here as its more convenient to use trl.

nivibilla:
@younesbelkada afaik the non contaminated packed dataset only works with flash attention and the monkey patching for the attention is necessary to make use of the dataset.

As an alternative, maybe it can be a separate script/tutorial notebook to let users use the new packing? Since it does improve training speeds by a lot. 5-10x in my case.

Palmik:
Does decontamination also handle the positional encoding? Meaning, will each packed sequence starts form 0?

khai-meetkai:
> Does decontamination also handle the positional encoding? Meaning, will each packed sequence start form 0?

Thanks to ROPE positional embedding, the position is relative. So the attention score between token at position i and token at the position j only depends on (j - i), so not necessarily starting from 0.

In a packed sequence of sequence 1 and sequence 2, we use extended attention_mask to indicate that tokens from sequence 1 only attend to tokens from sequence 1. There won't be cross-attention between sequence 1 and sequence 2.
You can read more information from: https://github.com/MeetKai/functionary/tree/main/functionary/train/packing

nivibilla:
@khai-meetkai regarding @younesbelkada 's comments. Is it possible to have a non Flash attention version? And without the monkey patch?

khai-meetkai:
> @khai-meetkai regarding @younesbelkada 's comments. Is it possible to have a non Flash attention version? And without the monkey patch?

**Non Flash-attention is possible** but **without monkey-patch is not possible**. The reason is the current code only **accepts attention_mask containing: 0, 1** when we pack sequences; we don't know which tokens are from which sequences to avoid contaminated cross-attention.
For example: sequence 1 = [1, 2, 3]; sequence 2 = [4, 5, 6, 7]
packed sequence = [1, 2, 3, 4, 5, 6, 7, pad_id, pad_id] (Here assume that we add padding tokens (pad_id))
current attention_mask = [1, 1, 1, 1, 1, 1, 1, 0, 0]
Our extended attention_mask=[1, 1, 1, 2, 2, 2, 2, 0, 0] --> our monkey-patched version handle this attention_mask

nivibilla:
> > @khai-meetkai regarding @younesbelkada 's comments. Is it possible to have a non Flash attention version? And without the monkey patch?
> 
> **Non Flash-attention is possible** but **without monkey-patch is not possible**. The reason is the current code only **accepts attention_mask containing: 0, 1** when we pack sequences; we don't know which tokens are from which sequences to avoid contaminated cross-attention.
> For example: sequence 1 = [1, 2, 3]; sequence 2 = [4, 5, 6, 7]
> packed sequence = [1, 2, 3, 4, 5, 6, 7, pad_id, pad_id] (Here assume that we add padding tokens (pad_id))
> current attention_mask = [1, 1, 1, 1, 1, 1, 1, 0, 0] --> contaminated cross-attention
> Our extended attention_mask=[1, 1, 1, 2, 2, 2, 2, 0, 0] --> our monkey-patched version handle this attention_mask

Makes sense. Unless the the way the attention works in HF itself is changed this is not possible.

nivibilla:
What about this idea @younesbelkada @khai-meetkai ?
1) Merge the patched attention into transformers assuming it doesnt break anything else. Or at least have an option to load with the patched attention.
2) Modify the code for the packing so that it works regardless of flash attention is enabled or not
    - (optional, we can otherwise force users to enable FA if one wants to use the new packing, however this would limit the use of this on older hardware such as colab's t4 gpus.)
3) Finally merge the non contaminated packing code into TRL

younesbelkada:
Hi @nivibilla @khai-meetkai 
Hmmm I don't think we'll be able to merge this if it depends on monkey patching, I thought there was a way to detect non-contaminated packing with classic attention mask. If you can find a workaround (maybe using a custom DataCollator) we can definitely merge this feature request but as currently it depends on monkey patched versions of modeling files that is not possible unfortunately. We cannot also add that patch to the current transformers code base as it is too specific to this usecase

nivibilla:
@younesbelkada makes sense. Can close this pr in that case.

danielhanchen:
I know people have asked if Unsloth could include packing as a stopgap measure to bridge the gap to TRL, so if it becomes a more hotly request feature, we can add it in :)

PR #1242, opened 2024-01-17
Add `setup_chat_format` for adding new special tokens to model for training chat models

philschmid:
# What does this PR do? 

This PR adds a new `util` method `setup_chat_format`, which automatically defines the `chat_template` for a tokenizer, adds special tokens, resizes the model embedding layer ([optional to a multiple of 64](https://x.com/karpathy/status/1621578354024677377?s=20))

It also introduces the `ChatMlSpecialTokens` dataclass, which is used in the `setup_chat_format`. This will make it easy to extend to different formats in the future, e.g., llama, but for now, we only add `chatml`. 

## Open Discussions

~Should we add more dummy tokens to the tokenizer when the embedding layer is extended to a multiple of x? This can lead to downstream [issues](https://github.com/OpenAccess-AI-Collective/axolotl/issues/227) with [llama.cpp](https://github.com/OpenAccess-AI-Collective/axolotl/pull/313)~


tests/test_model_utils.py
younesbelkada, line 1: Can you add this test in https://github.com/huggingface/trl/blob/main/tests/test_dataset_formatting.py ? 🙏

trl/models/utils.py
```diff
@@ -0,0 +1,79 @@
+from dataclasses import dataclass
+from typing import Literal, Tuple
+
+from transformers import PreTrainedModel, PreTrainedTokenizer
+
+
+@dataclass
+class ChatMlSpecialTokens:
```
younesbelkada, line 8: maybe create an abstract class with abstract methods so that future data classes can use the same template?

trl/models/utils.py
```diff
@@ -0,0 +1,79 @@
...
+    Setup chat format by adding special tokens to the tokenizer, setting the correct format, and extending the embedding layer of the model based on the new special tokens.
+
+    Args:
+      model (AutoModel): The model to be modified.
+      tokenizer (AutoTokenizer): The tokenizer to be modified.
+      format (Literal["chatml"], optional): The format to be set. Defaults to "chatml".
+      resize_to_multiple_of (int, optional): Number to resize the embedding layer to. Defaults to None.
+    Returns:
+      model (AutoModel): The modified model.
+      tokenizer (AutoTokenizer): The modified tokenizer.
```
younesbelkada, line 58:
```suggestion
      model (`~transformers.AutoModel`): The model to be modified.
      tokenizer (`~transformers.AutoTokenizer`): The tokenizer to be modified.
      format (`Literal["chatml"]`, *optional*): The format to be set. Defaults to "chatml".
      resize_to_multiple_of (int, *optional*): Number to resize the embedding layer to. Defaults to None.
    Returns:
      model (`~transformers.AutoModel`): The modified model.
      tokenizer (`~transformers.AutoTokenizer`): The modified tokenizer.
```

younesbelkada (commented):
Thanks a lot ! I left some early comments, in addition can you:
- document its usage in the SFTTrainer docs? 
- and also add a slow test here: https://github.com/huggingface/trl/blob/main/tests/slow/test_sft_slow.py#L36 with an end-to-end example
Thanks!

tests/test_model_utils.py
```diff
@@ -0,0 +1,29 @@
...
+        # Check if special tokens are correctly set
+        assert modified_tokenizer.eos_token == "<|im_end|>"
+        assert modified_tokenizer.pad_token == "<|im_end|>"
+        assert modified_tokenizer.bos_token == "<|im_start|>"
+        assert modified_tokenizer.eos_token == _chatml.eos_token
+        assert modified_tokenizer.pad_token == _chatml.pad_token
+        assert modified_tokenizer.bos_token == _chatml.bos_token
+        assert len(modified_tokenizer) == original_tokenizer_len + 2
+        assert self.model.get_input_embeddings().weight.shape[0] % 64 == 0
+        assert self.model.get_input_embeddings().weight.shape[0] == original_tokenizer_len + 64
```
younesbelkada, line 29:
```suggestion
        self.assertTrue(modified_tokenizer.eos_token == "<|im_end|>")
        self.assertTrue(modified_tokenizer.pad_token == "<|im_end|>")
        self.asserTrue(modified_tokenizer.bos_token == "<|im_start|>")
        self.assertTrue(modified_tokenizer.eos_token == _chatml.eos_token)
        self.asserTrue(modified_tokenizer.pad_token == _chatml.pad_token)
        self.assertTrue(modified_tokenizer.bos_token == _chatml.bos_token)
        self.assertTrue(len(modified_tokenizer) == original_tokenizer_len + 2)
        self.assertTrue(self.model.get_input_embeddings().weight.shape[0] % 64 == 0)
        self.assertTrue(self.model.get_input_embeddings().weight.shape[0] == original_tokenizer_len + 64)
```

tests/test_model_utils.py
philschmid: yes

trl/models/utils.py
philschmid: should we add this now or when we have more templates? I would rather add if there is more request and add a note?

docs/source/sft_trainer.mdx
```diff
@@ -154,6 +154,29 @@ response_template_ids = tokenizer.encode(response_template_with_context, add_spe
...
 ```
 
+### Add Special Tokens for Chat Format
+
+Adding special tokens to a language model is crucial for training chat models. These tokens are added between the different roles in a conversation, such as the user, assistant, and system and help the model recognize the structure and flow of a conversation. This setup is essential for enabling the model to generate coherent and contextually appropriate responses in a chat environment. 
+The [`setup_chat_format`] function in `trl` easily sets up a model and tokenizer for conversational AI tasks. This function:
+- Adds special tokens to the tokenizer, e.g. `<|im_start|>` and `<|im_end|>`, to indicate the start and end of a conversation.
+- Resizes the model’s embedding layer to accommodate the new tokens.
+- Sets the `chat_template` of the tokenizer, which is used to format the input data into a chat-like format. The default is `chatml` from OpenAI.
+- _optionally_ you can pass `resize_to_multiple_of` to resize the embedding layer to a multiple of the `resize_to_multiple_of` argument, e.g. 64.
```
younesbelkada, line 164:
```suggestion
- _optionally_ you can pass `resize_to_multiple_of` to resize the embedding layer to a multiple of the `resize_to_multiple_of` argument, e.g. 64. If you want to see more formats being supported in the future, please open a GitHub issue on [trl](https://github.com/huggingface/trl)
```

younesbelkada (approved):
Thanks a lot !

PR #1247, opened 2024-01-18
fix: fix loss_type and some args desc

zspo:
1. Fixed loss_type, "kto" -> "kto_pair"
2. Fixed some bugs related to parameter description


younesbelkada (approved):
Thanks for fixing!

Issue #1249, opened 2024-01-18
Dataloading speed significantly slower in multi GPU setting


JohnGiorgi:
Hi, I am trying to launch a fine-tuning job using a slightly modified version of `sft.py` with accelerate, looks something like:

```bash
accelerate launch \
    --config_file="./conf/accelerate/multi_gpu.yaml" \
    --num_processes 8 \
    examples/sft.py \
    --other_args
```

Within the script is a `formatting_func` and `DataCollatorForCompletionOnlyLM` data collator that are both passed to `SFTTrainer`, mostly following the [documentation](https://huggingface.co/docs/trl/main/en/sft_trainer).

At some point before training, the `SFTTrainer._prepare_non_packed_dataloader()` method is called. My issue is that the __time it takes to preprocess the dataset is significantl longer in the multi GPU setting, compared to a single GPU__. I might expect a little overhead to go distributed, but this is a significant difference (>10X):

```bash
# Single GPU setting, processes my dataset in about 20 minutes
accelerate launch \
    --config_file="./conf/accelerate/single_gpu.yaml" \
    --num_processes 1 \
    examples/sft.py \
    --other_args
 
# Multi GPU setting, reporting almost 5 hours to preprocess my data, a ~15X slowdown
accelerate launch \
    --config_file="./conf/accelerate/multi_gpu.yaml" \
    --num_processes 8 \
    examples/sft.py \
    --other_args
```
 
I got the accelerate configs right from the TRL repo [here](https://github.com/huggingface/trl/tree/main/examples/accelerate_configs). I figured the dataset was being loaded once per node and then batches dispatched to each GPU, so not really sure why I am seeing the slow-down before training even begins. I am not even sure this is a TRL issue but I figured it would be worth asking. Is there any kind of obvious gotcha that would lead to a slowdown like this? Anything one should take into account when moving to the multi-GPU setting with `accelerate + SFTTrainer`?

__Update__:

I confirmed this is not due to any changes in my script. Running the raw [`sft.py`](https://github.com/huggingface/trl/blob/main/examples/scripts/sft.py) script with default arguments and I get the following numbers:

```bash
# Single GPU
accelerate launch \
    --config_file="./conf/accelerate/single_gpu.yaml" \
    --num_processes 1 \
    examples/sft.py
# => data loading reports: 9846/9846 [00:07<00:00, 1340.85 examples/s]

# Multi GPU
accelerate launch \
    --config_file="./conf/accelerate/multi_gpu.yaml" \
[...]

PR #1255, opened 2024-01-19
Only load data on main process

JohnGiorgi:
This PR address #1249. With the fix, the new data preprocessing times become:

```bash
# Single GPU
accelerate launch \
    --config_file="./conf/accelerate/single_gpu.yaml" \
    --num_processes 1 \
    examples/sft.py
# => data loading reports: 9846/9846 [00:07<00:00, 1340.85 examples/s]

# Multi GPU
accelerate launch \
    --config_file="./conf/accelerate/multi_gpu.yaml" \
    --num_processes 8 \
    examples/sft.py
# => data loading reports: 9846/9846 [00:11<00:00, 878.93 examples/s]
```

So there is still some overhead in the distributed setting, but this is a big improvement over the 10X slowdown I was previously seeing.

My only note is that I wasn't exactly sure where to put the `PartialState().is_main_process` in the `SFTTrainer`, so that could maybe use a second set of eyes @younesbelkada. Thanks!

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -35,6 +35,7 @@ from transformers import (
 from transformers.modeling_utils import unwrap_model
 from transformers.trainer_callback import TrainerCallback
 from transformers.trainer_utils import EvalPrediction
+from accelerate import PartialState
 
 from ..extras.dataset_formatting import get_formatting_func_from_dataset
 from ..import_utils import is_peft_available
@@ -254,7 +255,7 @@ class SFTTrainer(Trainer):
 
         if dataset_kwargs is None:
             dataset_kwargs = {}
-        if train_dataset is not None:
+        if train_dataset is not None and PartialState().is_main_process:
             train_dataset = self._prepare_dataset(
                 train_dataset,
                 tokenizer,
@@ -267,7 +268,7 @@ class SFTTrainer(Trainer):
                 remove_unused_columns=args.remove_unused_columns if args is not None else True,
                 **dataset_kwargs,
             )
-        if eval_dataset is not None:
+        if eval_dataset is not None and PartialState().is_main_process:
             _multiple = isinstance(eval_dataset, dict)
             _eval_datasets = eval_dataset if _multiple else {"singleton": eval_dataset}
             for _eval_dataset_name, _eval_dataset in _eval_datasets.items():
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -19,6 +19,7 @@ from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
+from accelerate.state import PartialState
 from datasets import Dataset
 from datasets.arrow_writer import SchemaInferenceError
 from datasets.builder import DatasetGenerationError
@@ -35,7 +36,6 @@ from transformers import (
 from transformers.modeling_utils import unwrap_model
 from transformers.trainer_callback import TrainerCallback
 from transformers.trainer_utils import EvalPrediction
-from accelerate import PartialState
 
 from ..extras.dataset_formatting import get_formatting_func_from_dataset
 from ..import_utils import is_peft_available
@@ -253,30 +253,13 @@ class SFTTrainer(Trainer):
             if data_collator is None:
                 data_collator = DataCollatorForLanguageModeling(tokenizer=tokenizer, mlm=False)
 
-        # Avoid processing the dataset on all processes for multi-GPU training
-        is_main_process = PartialState().is_main_process
-        
-        if dataset_kwargs is None:
-            dataset_kwargs = {}
-        if train_dataset is not None and is_main_process:
-            train_dataset = self._prepare_dataset(
-                train_dataset,
-                tokenizer,
-                packing,
-                dataset_text_field,
-                max_seq_length,
-                formatting_func,
-                num_of_sequences,
-                chars_per_token,
-                remove_unused_columns=args.remove_unused_columns if args is not None else True,
-                **dataset_kwargs,
-            )
-        if eval_dataset is not None and is_main_process:
-            _multiple = isinstance(eval_dataset, dict)
-            _eval_datasets = eval_dataset if _multiple else {"singleton": eval_dataset}
-            for _eval_dataset_name, _eval_dataset in _eval_datasets.items():
-                _eval_datasets[_eval_dataset_name] = self._prepare_dataset(
-                    _eval_dataset,
+        # Pre-process the datasets only once per node. The remaining processes will use the cache.
+        with PartialState().main_process_first():
+            if dataset_kwargs is None:
+                dataset_kwargs = {}
+            if train_dataset is not None:
+                train_dataset = self._prepare_dataset(
+                    train_dataset,
                     tokenizer,
                     packing,
                     dataset_text_field,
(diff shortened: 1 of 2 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -254,7 +255,7 @@ def make_inputs_require_grad(module, input, output):
 
         if dataset_kwargs is None:
```
younesbelkada, line 256:
```suggestion
        is_main_process = PartialState().is_main_process
        
        if dataset_kwargs is None:
```

trl/trainer/sft_trainer.py
```diff
@@ -254,7 +255,7 @@ def make_inputs_require_grad(module, input, output):
...
             dataset_kwargs = {}
-        if train_dataset is not None:
+        if train_dataset is not None and PartialState().is_main_process:
```
younesbelkada, line 258:
```suggestion
        if train_dataset is not None and is_main_process:
```

trl/trainer/sft_trainer.py
```diff
@@ -267,7 +268,7 @@ def make_inputs_require_grad(module, input, output):
                 remove_unused_columns=args.remove_unused_columns if args is not None else True,
                 **dataset_kwargs,
             )
-        if eval_dataset is not None:
+        if eval_dataset is not None and PartialState().is_main_process:
```
younesbelkada, line 271:
```suggestion
        if eval_dataset is not None and is_main_process:
```

younesbelkada (commented):
Amazing work @JohnGiorgi thanks a lot for working on this! I just need to double check from which version `PartialState` has been introduced in accelerate . I also left some tiny suggestions - what do you think ?

JohnGiorgi:
@younesbelkada changes look good! I committed them. Do you think the place where it `is_main_process` check is happening in `SFTTrainer` makes sense?

younesbelkada:
I think it makes sense! 
@JohnGiorgi can you just run the styling checks? `make precommit`

JohnGiorgi:
Ah is something is up. Those speed gains are real but this change causes an indexing error when training begins:

```
IndexError: Invalid key: 8900 is out of bounds for size 0
    return inner_training_loop(
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/transformers/trainer.py", line 1821, in _inner_training_loop
    raise IndexError(f"Invalid key: {key} is out of bounds for size {size}")
IndexError: Invalid key: 9177 is out of bounds for size 0
    for step, inputs in enumerate(epoch_iterator):
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/accelerate/data_loader.py", line 448, in __iter__
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 630, in __next__
    data = self._next_data()
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 674, in _next_data
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/torch/utils/data/_utils/fetch.py", line 49, in fetch
    data = self.dataset.__getitems__(possibly_batched_index)
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/datasets/arrow_dataset.py", line 2804, in __getitems__
    batch = self.__getitem__(keys)
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/datasets/arrow_dataset.py", line 2800, in __getitem__
    return self._getitem(key)
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/datasets/arrow_dataset.py", line 2784, in _getitem
    pa_subtable = query_table(self._data, key, indices=self._indices if self._indices is not None else None)
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/datasets/formatting/formatting.py", line 583, in query_table
    _check_valid_index_key(key, size)
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/datasets/formatting/formatting.py", line 536, in _check_valid_index_key
    _check_valid_index_key(int(max(key)), size=size)
  File "/home/johngiorgi/dev/elms-ai/.venv/lib/python3.10/site-packages/datasets/formatting/formatting.py", line 526, in _check_valid_index_key
    raise IndexError(f"Invalid key: {key} is out of bounds for size {size}")
IndexError: Invalid key: 6081 is out of bounds for size 0
```

If I had to guess, at some point we go looking for the data on each GPU to fetch the latest batch and this is empty everywhere but the main process?

JohnGiorgi:
I have another solution which appears to work, basically:

- create a state object: `state = PartialState()`
- wrap all data loading logic in: `with state.main_process_first():`
- data is pre-processed on main process first, and then all other processes pick it up from the cache

This appears to work. I get the intended speedup and training begins successfully. Thoughts? @younesbelkada

younesbelkada (commented):
This looks great to me ! Thanks for working on it, one improvement could be to decorate the `_prepare_dataset(self, ..)` method with `@PartialState().main_process_first()` - could you try that out as well? 🙏 

cc @muellerz I would like you to have a second look from you if possible. For context, in this PR we make sure the datasets are processed on the main process first, which seems to lead to faster training pre-processing for multi-GPU envs

JohnGiorgi:
@younesbelkada made the change to use the decorator instead of the context manager, speedup is still there and training completes successfully when running `sft.py` with default args. Looks like some of the tests are failing though...

JohnGiorgi:
Awesome, thanks for the guidance @younesbelkada. I reverted! One thing:

- This change assumes all nodes have access to the same filesystem, otherwise it will break as only the main node will have access to the cached dataset. I think the simplest fix is to swap `main_process_first` with `local_main_process_first` so all nodes pre-process and cache the dataset
-

younesbelkada (commented):
I see, this makes sense, thanks for that! can you update it to `local_process_first` then?

JohnGiorgi:
Done!

trl/trainer/sft_trainer.py
```diff
@@ -252,27 +253,13 @@ def make_inputs_require_grad(module, input, output):
...
-                **dataset_kwargs,
-            )
-        if eval_dataset is not None:
-            _multiple = isinstance(eval_dataset, dict)
-            _eval_datasets = eval_dataset if _multiple else {"singleton": eval_dataset}
-            for _eval_dataset_name, _eval_dataset in _eval_datasets.items():
-                _eval_datasets[_eval_dataset_name] = self._prepare_dataset(
-                    _eval_dataset,
+        # Pre-process the datasets only once per node. The remaining processes will use the cache.
+        with PartialState().is_local_main_process():
```
younesbelkada, line 257:
```suggestion
        with PartialState().is_local_main_process:
```

younesbelkada (commented):
Thanks! I think this should fix the failing test - can you double check all the initial issues on your side are fixed? 🙏

JohnGiorgi:
Oof, thanks for catching that mistake with the attribute instead of the context manager @younesbelkada . Just tested it out and its working great, huge reduction in data loading speed in the distributed setting, virtually the same time as the non-distributed setting which is what I was aiming for :)

trl/trainer/sft_trainer.py
```diff
@@ -252,27 +253,13 @@ def make_inputs_require_grad(module, input, output):
...
-                **dataset_kwargs,
-            )
-        if eval_dataset is not None:
-            _multiple = isinstance(eval_dataset, dict)
-            _eval_datasets = eval_dataset if _multiple else {"singleton": eval_dataset}
-            for _eval_dataset_name, _eval_dataset in _eval_datasets.items():
-                _eval_datasets[_eval_dataset_name] = self._prepare_dataset(
-                    _eval_dataset,
+        # Pre-process the datasets only once per node. The remaining processes will use the cache.
+        with PartialState().is_local_main_process:
```
younesbelkada, line 257:
```suggestion
        if PartialState().is_local_main_process:
```

younesbelkada (commented):
Thanks again ! it seems there is still an issue :D can you try that out?

JohnGiorgi:
Works and I still get the speed up in the distributed setting with the vanilla `sft.py` script and default arguments!

younesbelkada (approved):
Thank you very much @JohnGiorgi for all your hardwork on this ! Great job !

PR #1265, opened 2024-01-22
[DPO] average_log_prob when loss is IPO

kashif:
so that the beta is not dependent on the sequence length with IPO


edbeeching (approved):
Thanks for the fix. LGTM!

PR #1276, opened 2024-01-26
FIx SFTTrainer bugs on TRL main

younesbelkada:
Currently the TRL gpu CI is broken after #1255 - this PR fixes it

cc @lvwerra FYI


trl/trainer/sft_trainer.py
```diff
@@ -254,7 +254,7 @@ def make_inputs_require_grad(module, input, output):
                 data_collator = DataCollatorForLanguageModeling(tokenizer=tokenizer, mlm=False)
 
         # Pre-process the datasets only once per node. The remaining processes will use the cache.
-        if PartialState().is_local_main_process:
+        with PartialState().local_main_process_first()::
```
younesbelkada, line 257:
```suggestion
        with PartialState().local_main_process_first():
```

PR #1278, opened 2024-01-26
Fix SFT tuner in CI

vwxyzjn:
I also tested with `python examples/scripts/sft.py --output_dir sft_openassistant-guanaco  --model_name meta-llama/Llama-2-7b-hf --dataset_name timdettmers/openassistant-guanaco --load_in_4bit --use_peft --per_device_train_batch_size 4 --gradient_accumulation_steps 2` which seems to work (https://wandb.ai/costa-huang/huggingface/runs/m35stukl/logs)


younesbelkada (approved):
Thanks a lot for the quick reaction ! Looks good!

PR #1284, opened 2024-01-27
fix padding in dpo trainer

pacman100:
### What does this PR do?
1. The padding token in DPO trainer is wrongly set to 0 when tokenizer uses a different token id for padding and tokenizer is passed. This happens because the conditional statement is checking if padding value is `None` but if the default is 0, it is never `None`.


pacman100:
cc @younesbelkada

younesbelkada (approved):
Thanks, in principle this looks good - I would love though a second look from @kashif - let me know what do you think and once that's good for you, feel free to merge!

kashif:
great catch thanks!

kashif (approved):

PR #1286, opened 2024-01-29
Add multiprocessing in the DPO trainer.

imraviagrawal:
This tiny PR add support for multi process preprocessing in the DPO trainer. 
Added support for num_proc to tokenize the training dataset.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -122,6 +122,8 @@ class DPOTrainer(Trainer):
         precompute_ref_log_probs (`bool`, defaults to `False`):
             Flag to precompute reference model log probabilities and evaluation datasets. This is useful if you want to train
             without the reference model and reduce the total GPU memory needed.
+        dataset_num_proc (`Optional[int]`):
+            The number of workers to use to tokenize the data. Defaults to None.
         model_init_kwargs: (`Optional[Dict]`, *optional*):
             Dict of Optional kwargs to pass when instantiating the model from a string
         ref_model_init_kwargs: (`Optional[Dict]`, *optional*):
@@ -162,6 +164,7 @@ class DPOTrainer(Trainer):
         generate_during_eval: bool = False,
         compute_metrics: Optional[Callable[[EvalLoopOutput], Dict]] = None,
         precompute_ref_log_probs: bool = False,
+        dataset_num_proc = None,
         model_init_kwargs: Optional[Dict] = None,
         ref_model_init_kwargs: Optional[Dict] = None,
         model_adapter_name: str = None,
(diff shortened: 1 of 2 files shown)


trl/trainer/dpo_trainer.py
```diff
@@ -162,6 +164,7 @@ def __init__(
         generate_during_eval: bool = False,
         compute_metrics: Optional[Callable[[EvalLoopOutput], Dict]] = None,
         precompute_ref_log_probs: bool = False,
+        dataset_num_proc = None,
```
younesbelkada:
```suggestion
        dataset_num_proc: int = 2,
```

younesbelkada (commented):
Thanks a lot ! Do you think it would make sense to initialize it with a decent value, e.g. 2 ?

trl/trainer/dpo_trainer.py
```diff
@@ -162,6 +164,7 @@ def __init__(
         generate_during_eval: bool = False,
         compute_metrics: Optional[Callable[[EvalLoopOutput], Dict]] = None,
         precompute_ref_log_probs: bool = False,
+        dataset_num_proc: bool = None,
```
younesbelkada, line 167:
```suggestion
        dataset_num_proc: int = 2,
```

trl/trainer/dpo_trainer.py
```diff
@@ -122,6 +122,8 @@ class DPOTrainer(Trainer):
         precompute_ref_log_probs (`bool`, defaults to `False`):
             Flag to precompute reference model log probabilities and evaluation datasets. This is useful if you want to train
             without the reference model and reduce the total GPU memory needed.
+        dataset_num_proc (`Optional[int]`):
```
younesbelkada, line 125:
```suggestion
        dataset_num_proc (`int`):
```

trl/trainer/dpo_trainer.py
imraviagrawal: thats reasonable

younesbelkada (commented):
Actually this looks good, let's make it optional ! 
Would you be happy to add a test here: https://github.com/huggingface/trl/blob/main/tests/test_dpo_trainer.py#L231 just to make sure training runs fine?
Also make sure to run the styling checks `make precommit`

trl/trainer/dpo_trainer.py
younesbelkada: Please see my latest comment, I think it is good to make it configurable and set it to None by default

imraviagrawal:
@younesbelkada take a look at the test case, we also did make precommit

younesbelkada (approved):
Thanks a lot for your work on this!

PR #1289, opened 2024-01-30
[`PEFT` + `DPO`] Raise value error if one passes a ref_model and a peft_config

younesbelkada:
Recently I have seen many tutorials that passes both the ref_model and a peft_config, this should be not allowed as it is very memory inefficient. This PR enhances the UX by raising a value error if peft_config if passed together with a ref_model and guiding the user on what should be the best practice.

cc @kashif


kashif (approved):

younesbelkada:
Oops actually this made the example tests fail for DPO + QLoRA .. will fix it now

Issue #1217, opened 2024-01-11
Padding side incorrect for Mistral DPO


hengjiUSTC:
For code here: https://github.com/huggingface/trl/blob/main/trl/trainer/utils.py#L538C4-L550 Does it always padding to right?

I get error when run dpo for mixtral 7b.

I am still seeing this error after set `tokenizer.padding_side  = 'left'`
This is my demo notebook: https://colab.research.google.com/drive/1sVqbYEOqjJYl7CzNzXzviEBB6A984cMq?usp=sharing

**Tokenizer already set with left padding**
![截屏2024-01-10 下午10 53 07](https://github.com/huggingface/transformers/assets/25971665/0ce9f431-8a8f-4236-8c71-846f183f455e)

**Train**
![截屏2024-01-11 上午10 28 53](https://github.com/huggingface/trl/assets/25971665/e8f0a944-9f22-4cba-88a6-a1712ba02f2c)

**Still have**: `ValueError: You are attempting to perform batched generation with padding_side='right' this may lead to unexpected behaviour for Flash Attention version of Mistral. Make sure to  call `tokenizer.padding_side  = 'left'` before tokenizing the input.`

![截屏2024-01-10 下午10 54 31](https://github.com/huggingface/transformers/assets/25971665/f031de6c-6b25-4be8-8b4a-7075e70785a5)

Seems https://github.com/huggingface/trl/blob/main/trl/trainer/utils.py#L538C4-L550 or other part of code might lead to bug?

transformers             4.36.2
trl                      0.7.7
peft                     0.6.0

hengjiUSTC:
Did some debugging, quiet confirmed that https://github.com/huggingface/trl/blob/main/trl/trainer/utils.py#L538C4-L550 gives incorrect padding side and lead to crash in transformers.

```
File ~/learn-llm/venv/lib/python3.10/site-packages/transformers/models/mistral/modeling_mistral.py:899, in MistralModel.forward(self, input_ids, attention_mask, position_ids, past_key_values, inputs_embeds, use_cache, output_attentions, output_hidden_states, return_dict)
    897     is_padding_right = attention_mask[:, -1].sum().item() != batch_size
    898     if is_padding_right:
--> 899         raise ValueError(
    900             "You are attempting to perform batched generation with padding_side='right'"
    901             " this may lead to unexpected behaviour for Flash Attention version of Mistral. Make sure to "
    902             " call `tokenizer.padding_side  = 'left'` before tokenizing the input. "
[...]

PR #1290, opened 2024-01-30
[`DPOTrainer`] Fix DPO trainer + mistral + FA2

younesbelkada:
Fixes: https://github.com/huggingface/trl/issues/1217 
Fixes: https://github.com/huggingface/transformers/issues/26877
Fixes: https://github.com/huggingface/trl/issues/1266

Simply setting `use_cache=False` circumvents all issues with FA-2 + DPO + Mistral. in fact we should bypass that check since we are not in text generation mode when computing the loss function. `use_cache` is retrieved from the model config by default which falls back always to `True`. The cache is not used anyway when purely computing the logits so this change is fully BC

cc @kashif @vwxyzjn 

To test that, I managed to repro the issue by adding `--attn_implementation "flash_attention_2"` in the dpo shell script on a A100 machine, and I confirm this PR fixes it. Unfortunately our CI runners are not compatible with FA2 so we cannot add a slow test to test that

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -964,6 +964,7 @@ class DPOTrainer(Trainer):
         all_logits = model(
             concatenated_batch["concatenated_input_ids"],
             attention_mask=concatenated_batch["concatenated_attention_mask"],
+            use_cache=False,
             **model_kwargs,
         ).logits
 


kashif (approved):

kashif:
thank you great help!

PR #1292, opened 2024-01-30
Fix DPO slow tests

younesbelkada:
As https://github.com/huggingface/trl/pull/1289 being merged, you can't both load a ref_model and a peft model at the same time, therefore we should remove that test from our slow tests 

cc @kashif


kashif (approved):

kashif:
https://github.com/huggingface/trl/pull/1289 is merged so we can remove this

PR #1297, opened 2024-01-31
Types: Fix PEP 484 implicit-optional compliance

akx:
This was done automatically with [hauntsaninja/no_implicit_optional](https://github.com/hauntsaninja/no_implicit_optional).

This will help avoid bugs such as the one fixed by #1295, since possibly-`None` values are marked as such by the type annotations.


younesbelkada (approved):
Great catch @akx ! Any pointers on how we could test that through a CI workflow?

younesbelkada:
Really appreciate your recent contributions (e.g. bitsandbytes) thanks a lot @akx !

akx:
> Great catch @akx ! Any pointers on how we could test that through a CI workflow?

Ruff can quickly catch simple mistakes (these included).

Mypy and Pyright are more fully-fledged type checkers which take more time and can be annoyingly strict in certain cases.

PR #1301, opened 2024-01-31
Codemod Unittest assertions to bare asserts

akx:
This PR runs https://github.com/akx/codemod-unittest-to-pytest-asserts to rewrite Unittest `self.assert...` assertions as plain assertions that can be introspected by Pytest's [assert rewriting](https://docs.pytest.org/en/7.1.x/how-to/assert.html).

### DIFF
--- a/tests/slow/test_dpo_slow.py
+++ b/tests/slow/test_dpo_slow.py
@@ -145,8 +145,8 @@ class DPOTrainerSlowTester(unittest.TestCase):
                 max_length=self.max_length,
             )
 
-            self.assertTrue(isinstance(trainer.model, PeftModel))
-            self.assertTrue(trainer.ref_model is None)
+            assert isinstance(trainer.model, PeftModel)
+            assert trainer.ref_model is None
 
             # train the model
             trainer.train()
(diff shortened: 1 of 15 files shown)


younesbelkada (commented):
Hi @akx 
Thanks for your contribution ! 
few tests are failing with:
```bash
FAILED tests/test_core.py::CoreTester::test_masked_whiten - TypeError: type Tensor doesn't define __round__ method
```
I think adding `.item()` on those would fix the tests - could you please have a look? 🙏

younesbelkada (commented):
Sorry for the additional iteration @akx ! Could you also fix
```bash
FAILED tests/test_reward_trainer.py::RewardTrainerTester::test_reward_trainer_margin
```
? 🙏 Thanks !

akx:
Should be good to go again, I gave the failing test a spin locally (which took some editing because the tests don't run OOTB on `mps` devices, but there we go). 👍

younesbelkada:
Thanks very much @akx - just re-ran the CI

younesbelkada (approved):
Huge work ! Thanks !

PR #1302, opened 2024-01-31
Token-level rewards

XanderJC:
Allow passing rewards to step that are List[`torch.FloatTensor`] but each either of shape 1 (i.e. same as existing functionality) or (`response_length`) so that we can give reward to individual tokens. Important for more fine-grained feedback.

### DIFF
--- a/tests/test_ppo_trainer.py
+++ b/tests/test_ppo_trainer.py
@@ -459,7 +459,6 @@ class PPOTrainerTester(unittest.TestCase):
         for query_tensor, response_tensor in dummy_dataloader:
             # define a reward for response
             # (this could be any reward such as human feedback or output from another model)
-            # reward = [torch.tensor([[1.0]]), torch.tensor([[0.0]])]
             reward = [torch.tensor([1.0, 2.0, 3.0]), torch.tensor([[0.0, 1.0]])]
             # train model - this should raise an error
             with self.assertRaises(ValueError):
@@ -575,7 +574,6 @@ class PPOTrainerTester(unittest.TestCase):
 
         dummy_queries = [torch.tensor([1, 2, 3, 4]), torch.tensor([1, 2, 3, 4, 5, 6, 7])]
         dummy_responses = [torch.tensor([5, 6, 7, 8, 9]), torch.tensor([8, 9, 10, 11, 12, 13])]
-        # dummy_scores = torch.Tensor([1, 2])
         dummy_scores = [
             torch.tensor([0, 0, 0, 0, 1], device=ppo_trainer.current_device),
             torch.tensor([0, 0, 0, 0, 0, 2], device=ppo_trainer.current_device),
--- a/tests/test_ppo_trainer.py
+++ b/tests/test_ppo_trainer.py
@@ -151,7 +151,9 @@ class PPOTrainerTester(unittest.TestCase):
     def _init_dummy_dataset(self):
         # encode a query
         query_txt = "This morning I went to the "
-        query_tensor = self.gpt2_tokenizer.encode(query_txt, return_tensors="pt")
+        query_tensor = self.gpt2_tokenizer.encode(query_txt, return_tensors="pt").to(
+            self.gpt2_model.pretrained_model.device
+        )
         assert query_tensor.shape == (1, 7)
         # get model response
         response_tensor = respond_to_batch(self.gpt2_model, query_tensor)
@@ -457,12 +459,18 @@ class PPOTrainerTester(unittest.TestCase):
         for query_tensor, response_tensor in dummy_dataloader:
             # define a reward for response
             # (this could be any reward such as human feedback or output from another model)
-            reward = [torch.tensor([[1.0]]), torch.tensor([[0.0]])]
+            # reward = [torch.tensor([[1.0]]), torch.tensor([[0.0]])]
+            reward = [torch.tensor([1.0, 2.0, 3.0]), torch.tensor([[0.0, 1.0]])]
             # train model - this should raise an error
             with self.assertRaises(ValueError):
                 _ = ppo_trainer.step([q for q in query_tensor], [r for r in response_tensor], reward)
 
             reward = [torch.tensor([1.0]), torch.tensor([0.0])]
+            # train model - this should work
+            _ = ppo_trainer.step([q for q in query_tensor], [r for r in response_tensor], reward)
+
+            # token-level rewards
+            reward = [torch.tensor([1.0] * 7), torch.tensor([0.0] * 7)]
             # train model - this should work
             _ = ppo_trainer.step([q for q in query_tensor], [r for r in response_tensor], reward)
             break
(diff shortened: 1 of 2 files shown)


trl/trainer/ppo_trainer.py
```diff
@@ -142,10 +142,10 @@ class PPOTrainer(BaseTrainer):
 
     def __init__(
         self,
-        config: Optional[PPOConfig] = None,
-        model: Optional[PreTrainedModelWrapper] = None,
+        config: PPOConfig = None,
+        model: PreTrainedModelWrapper = None,
```
akx, line 146: If the default value is `None`, the type must be `Optional[]`; please see #1297 and https://github.com/hauntsaninja/no_implicit_optional.

(Repeated for all instances in this PR.)

XanderJC: Fixed

younesbelkada (commented):
cc @vwxyzjn @lvwerra what do you think?

trl/trainer/ppo_trainer.py
```diff
@@ -605,14 +605,34 @@ def _step_safety_checker(
...
+                        f"Score and response must have the same length if score not scalar- got {score.shape[0]} and {response.shape[0]}"
+                    )
+                else:
+                    scores[i] = score
+                    token_level_score = True
+            else:
+                token_score = torch.zeros_like(response, dtype=float).squeeze().to(self.current_device)
+                token_score[-1] = score
+                scores[i] = token_score
+                token_level_score = False
```
vwxyzjn, line 633: This seems like way too many if-else statements. The end-of-sequence reward is a special case of token-level rewards. Maybe we could just always convert end-of-sequence reward to token-level rewards internally?

tests/test_ppo_trainer.py
```diff
@@ -565,7 +574,10 @@ def test_loss_trainer(self):
 
         dummy_queries = [torch.tensor([1, 2, 3, 4]), torch.tensor([1, 2, 3, 4, 5, 6, 7])]
         dummy_responses = [torch.tensor([5, 6, 7, 8, 9]), torch.tensor([8, 9, 10, 11, 12, 13])]
-        dummy_scores = torch.Tensor([1, 2])
+        dummy_scores = [
+            torch.tensor([0, 0, 0, 0, 1], device=ppo_trainer.current_device),
+            torch.tensor([0, 0, 0, 0, 0, 2], device=ppo_trainer.current_device),
+        ]
```
vwxyzjn, line 580: Is this a breaking change requiring users to always pass in token level rewards?

trl/trainer/ppo_trainer.py
```diff
@@ -821,8 +854,12 @@ def step(
         train_stats["policy/advantages"] = torch.nan_to_num(train_stats["policy/advantages"], WANDB_PADDING)
         train_stats["policy/ratio"] = torch.flatten(train_stats["policy/ratio"]).unsqueeze(0)
 
+        total_scores = torch.tensor([score[score != padding_value].sum() for score in padded_scores]).to(
+            self.current_device
+        )
+
```
vwxyzjn, line 860: We should not only sum it up — the discount factor gamma also comes in. The total_scores should be the discounted episodic return. See https://spinningup.openai.com/en/latest/spinningup/rl_intro.html

image

vwxyzjn (commented):
Thanks for the PR! This PR could be useful for research like https://openai.com/research/improving-mathematical-reasoning-with-process-supervision :) Did a quick pass and left a few comments. The two main high-level things are:

1. keep back compatibility of the API
2. maybe clean up the logic a bit regarding the if / else statements

trl/trainer/ppo_trainer.py
XanderJC: @vwxyzjn happy to change things but let me just quickly explain what I'm doing a bit here in a bit more detail to see if it makes sense first, as it's doing two things which maybe makes the number of if/else statements a bit more palatable

1. lines 611-619 are doing most of the safety checking because to make sure its back-compatitble we need to accept scores that are either scalar e.g. [1.0] or as a vector e.g. [1.0,2.0,3.0]. I think this means we need to do a lot more dim checks as in the first case I could expect either torch.tensor(1.0) or torch.tensor([1.0]) with dim 0,1 respectively from the user. Similarly with the second either torch.tensor([1.0,2.0,3.0]) or torch.tensor([[1.0,2.0,3.0]]) with dim 1,2 respectively. So here we're just making sure we deal with dim 1, i.e.  torch.tensor([1.0]) or torch.tensor([1.0,2.0,3.0]). Theoretically could simplify the code here and just get rid of the logic for turning non dim 1 tensors to dim 1 and throw an error, but it seems helpful to convert - let me know if you think this is a better strategy.
2. lines 621-633 is doing what you suggest which is then to convert end-of-sequence rewards to token-level rewards as we check the length and if it's 1 we turn it into a token-level reward with the score at the last spot, otherwise we just check the token-level reward is the sam size as the response and if it is then everything's good

I added the token_level_score flag as well to inform later manipulations of the score which I think should depend on which one got passed to trainer.step()

tests/test_ppo_trainer.py
XanderJC: @vwxyzjn I don't think this is a breaking change, but only because I'm *assuming* that no one is calling self.compute_rewards() outside of self.step(), as I couldn't see why that would be useful.

I set it up so you can pass either token-level or end-of-sequence rewards to step(), which then all get converted to token-level in _step_safety_checker(), meaning everything that gets sent to compute_rewards() would be always token-level.

trl/trainer/ppo_trainer.py
XanderJC: @vwxyzjn ah yes good point, I've added that calculation

trl/trainer/ppo_trainer.py
```diff
@@ -605,14 +605,34 @@ def _step_safety_checker(
         scores = [tensor.to(self.current_device) for tensor in scores]
         masks = [tensor.to(self.current_device) for tensor in masks] if masks is not None else None
 
-        # squeeze scores if needed
-        for i, score in enumerate(scores):
-            if score.dim() > 1:
+        # format scores to token-level scores if needed
+        for i, (score, response) in enumerate(zip(scores, responses)):
+            # make score 1-dimensional
```
vwxyzjn, line 610: This comment seems outdated.

trl/trainer/ppo_trainer.py
vwxyzjn: Why do we need a `token_level_score`? Maybe we can just assume `score` is token-level. Also the `token_level_score` appears not tracked for each index, so if at `i`, `token_level_score=True`, and `i+1`, `token_level_score=False`. Then isn't this information missing?

trl/trainer/ppo_trainer.py
```diff
@@ -1330,6 +1372,9 @@ def log_stats(
         """
 
         # all gather stats
+
+        # sum to episodic rewards
+        rewards = [reward.sum() for reward in rewards]
```
vwxyzjn, line 1377: This is not the discounted return

vwxyzjn (commented):
Some further comments

PR #1307, opened 2024-02-01
[DPOTrainer] Add multiprocessing for the eval_dataset map

esceptico:
Added the num_proc parameter to the eval_dataset processing (similar to train_dataset)

cc @kashif


kashif:
thanks! having a look

kashif (approved):

PR #1309, opened 2024-02-01
ENH: Run CI only if relevant files are modified

younesbelkada:
cc @lvwerra

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ on:
       - "trl/**.py"
       - "examples/**.py"
       - "scripts/**.py"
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -5,6 +5,13 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+    paths:
+      # Run only when relevant files are modified
+      - "trl/**.py"
+      - "examples/**.py"
+      - "scripts/**.py"
+      - ".github/**yml"
```
younesbelkada, line 13:
```suggestion
      - ".github/**.yml"
```

PR #1312, opened 2024-02-02
Fix typos in docs for Multi Adapter RL (MARL).

elhusseiniali:
Just fixed up a few typos, nothing really significant


younesbelkada (approved):
Thanks for fixing the typos !

Issue #1238, opened 2024-01-17
When reference_free the ref_logratios is zero int, but int type has no attribute .to for set device causes AttributeError


maliozer:
https://github.com/huggingface/trl/blob/baf3c1c2939986b08660cb023da12310ee293da6/trl/trainer/dpo_trainer.py#L846C9-L846C22

reference_free case causes AttributeError

```python
ref_logratios = 0
ref_logratios = ref_logratios.to(self.accelerator.device)

AttributeError: 'int' object has no attribute 'to'
```

younesbelkada:
Indeed, nice catch @maliozer would you be happy to open a PR to fix it?

maliozer:
ok let me check for it

younesbelkada:
Thanks for your contribution and fix @maliozer !

PR #1313, opened 2024-02-02
Fix AttributeError in dpo_trainer for reference_free case in dpo_loss function

maliozer:
Issue: An `AttributeError` was being raised due to trying to call `.to()` on an integer object. This error occurred because `ref_logratios` was initialized as an integer (`0`) instead of a tensor. https://github.com/huggingface/trl/issues/1238

Fix: The fix involved initializing `ref_logratios` as a tensor with PyTorch's `torch.tensor()` function before calling the `.to()` method.  Also, `ref_logratios` is now stored as an instance variable `self.ref_logratios` during initialization to ensure it retains its value throughout the lifecycle of the object.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -133,7 +133,7 @@ class DPOTrainer(Trainer):
             Name of the train target PEFT adapter, when using LoRA with multiple adapters.
         ref_adapter_name (`str`, defaults to `None`):
             Name of the reference PEFT adapter, when using LoRA with multiple adapters.
-        reference_free` (`bool`):
+        reference_free (`bool`):
             If True, we ignore the _provided_ reference model and implicitly use a reference model that assigns equal probability to all responses.
     """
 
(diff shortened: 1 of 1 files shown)


trl/trainer/dpo_trainer.py
```diff
@@ -133,6 +133,8 @@ class DPOTrainer(Trainer):
             Name of the train target PEFT adapter, when using LoRA with multiple adapters.
         ref_adapter_name (`str`, defaults to `None`):
             Name of the reference PEFT adapter, when using LoRA with multiple adapters.
+        reference_free` (`bool`):
```
younesbelkada, line 136:
```suggestion
        reference_free (`bool`):
```

trl/trainer/dpo_trainer.py
```diff
@@ -844,16 +847,15 @@ def dpo_loss(
...
         Returns:
             A tuple of three tensors: (losses, chosen_rewards, rejected_rewards).
             The losses tensor contains the DPO loss for each example in the batch.
             The chosen_rewards and rejected_rewards tensors contain the rewards for the chosen and rejected responses, respectively.
         """
         pi_logratios = policy_chosen_logps - policy_rejected_logps
-        if reference_free:
-            ref_logratios = 0
+        if self.reference_free:
+            ref_logratios = torch.tensor([0], dtype=pi_logratios.dtype)
```
younesbelkada, line 858:
```suggestion
            ref_logratios = torch.tensor([0], dtype=pi_logratios.dtype, device=pi_logrations.device)
```

younesbelkada (approved):
Makes sense to me, thanks! 
what do you think @kashif ?

kashif:
yes makes sense... i never tested it with this flag turned on but yeah looks good

younesbelkada:
Perfect - @maliozer would you be happy to accept the suggestions? Then we can merge I think

PR #1315, opened 2024-02-02
Update README.md to clarify model requirement

### LINKED ISSUE #1257: Can TRL be used to finetune non-transformer torch models?
Can TRL be used to finetune non-transformer torch models like VGG, ResNet etc?

markstur:
Clarify that language models must be transformers models for text. This is a bit redundant with intro description, but attempts to better address a question that does come up (issue 1257).

Fixes #1257 

Cc: @saptab @younesbelkada


younesbelkada (approved):
Makes sense, thanks !

PR #1321, opened 2024-02-06
Fix doc snippet PPOTrainer argument train_dataset -> dataset

j-cb:
Both the argument's name and the value need to be renamed in the code snippet. Otherwise we get both

`NameError: name 'train_dataset' is not defined`
and
`TypeError: PPOTrainer.__init__() got an unexpected keyword argument 'train_dataset'`


younesbelkada (approved):
thanks for fixing the typos!

PR #1325, opened 2024-02-11
Best practice recommendation update for dpo_trainer.mdx

R-seny:
In the document as it is now, the best practice recommendations on merging adaptors into the base model seem neither consistent nor correct.

For example, the documentation links a tweet with a recommendation to merge adaptors into a quantized model, and a script that supposedly illustrates how to apply that recommendation. But the script actually does the opposite of what the tweet recommends, first dequantizing the model. 

There are similar inconsistencies/ambiguities further in that paragraph. For example, saying that using an unquantized model would lead to lower performance (I changed it to "higher memory demand").

Overall, I updated the paragraph to improve consistency and provided links to slightly more evidence-based merging recommendations.

### DIFF
--- a/docs/source/dpo_trainer.mdx
+++ b/docs/source/dpo_trainer.mdx
@@ -167,15 +167,13 @@ You have three main options (plus several variants) for how the reference model 
 
 ### Downsides to merging QLoRA before DPO (approach 2)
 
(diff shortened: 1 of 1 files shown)


lewtun (approved):
Thanks for the fix!

PR #1344, opened 2024-02-21
[WIP] Add SPIN trainer

lewtun:
Implements the Self-Play fIne-tuNing (SPIN) algorithm from: https://arxiv.org/abs/2401.01335

## TODO

- [ ] Validate training works for a few iterations of Qwen-1.5-0.5b
- [x] Fix generation with ZeRo-3 init (hitting annoying `RuntimeError: 'weight' must be 2-D` errors)
- [ ] Add docs
- [ ] Refactor / clean up internals


examples/scripts/spin.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+@dataclass
+class ScriptArguments:
+    dataset_name: str = field(default="HuggingFaceH4/spin-ultrachat-prompts", metadata={"help": "the dataset name"})
+    max_seq_length: int = field(default=512, metadata={"help": "The maximum sequence length for SFT Trainer"})
+    preprocessing_num_workers: int = field(
+        default=12, metadata={"help": "The number of processes to use for the preprocessing."}
+    )
+    do_generate: bool = field(default=False)
```
lewtun, line 87: These args could arguably live in the `SPINConfig`

examples/scripts/dpo.py
```diff
@@ -49,16 +49,21 @@
...
 
+import datasets
 import torch
 from datasets import Dataset, load_dataset
 from transformers import AutoModelForCausalLM, AutoTokenizer, HfArgumentParser, TrainingArguments
 
 from trl import DPOTrainer, ModelConfig, get_kbit_device_map, get_peft_config, get_quantization_config
 
 
+datasets.disable_caching()
```
vwxyzjn, line 64: What is this? Maybe we should do `load_from_cache=False`?

examples/scripts/spin.py
```diff
@@ -0,0 +1,221 @@
...
+        spin_trainer.save_state()
+        spin_trainer.save_model(training_args.output_dir)
+        spin_trainer.push_to_hub()
+
+        print(f"Training completed! Model pushed to {spin_trainer.hub_model_id}")
+
+    if script_args.do_generate:
+        print(f"Generating completions for {len(prompt_train_ds)} training examples")
+        train_completions = spin_trainer.generate(prompt_train_ds, "prompt", generation_config, batch_size=16)
+        test_completions = spin_trainer.generate(prompt_test_ds, "prompt", generation_config, batch_size=16)
```
vwxyzjn, line 204: If the generation only occurs in the end, wdyt about doing a subprocess call on a vllm/tgi script to generate instead?

trl/trainer/utils.py
```diff
@@ -358,6 +358,224 @@ def __call__(self, features: List[Dict[str, Any]]) -> Dict[str, Any]:
...
+        We also create the labels for the chosen/rejected responses, which are of length equal to
+            the sum of the length of the prompt and the chosen/rejected response, with
+            label_pad_token_id  for the prompt tokens.
+        """
+        batch = {}
+
+        if not self.is_encoder_decoder:
+            chosen_tokens = self.tokenizer(chosen, add_special_tokens=False)
+            rejected_tokens = self.tokenizer(rejected, add_special_tokens=False)
+            prompt_tokens = self.tokenizer(prompt, add_special_tokens=False)
```
vwxyzjn, line 420: This feels quite similar to the logic in DPOTrainer. Maybe unify the logic somehow (or use similar terminology). 

https://github.com/huggingface/trl/blob/b32656f7266c02500ea346eda75a0db7a9a1bb75/trl/trainer/dpo_trainer.py#L721-L729

Not necessary but we could also benefit from doing a filediff between SPIN and DPO and try to make the lines of code differences minimal.

image

vwxyzjn (commented):
Great work! Left some initial comments :)

davidberenstein1957:
I wanted to look at implementing this in TRL. @lewtun @vwxyzjn was there a reason this was closed? Should we reopen it and can I continue the work?

Issue #1348, opened 2024-02-22
StackLlaMa 2 dpo train failed: 8-bit model can't train with multiple gpus 


fancyerii:
I am following https://github.com/huggingface/trl/tree/main/examples/research_projects/stack_llama_2/scripts.

I ran with:

```
accelerate launch --config_file 7b.yaml examples/research_projects/stack_llama_2/scripts/dpo_llama2.py     --model_name_or_path="sft/final_checkpoint"     --output_dir="dpo"
```

7b.yaml file:
```
compute_environment: LOCAL_MACHINE
debug: false
distributed_type: MULTI_GPU
downcast_bf16: 'no'
gpu_ids: 0,1,2,3
machine_rank: 0
main_training_function: main
mixed_precision: bf16
num_machines: 1
num_processes: 4
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

error message:
```
Traceback (most recent call last):
  File "/nas/lili/codes/pt/ft/trl/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py", line 213, in <module>
    dpo_trainer.train()
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/ft-zSqjAXBp-py3.9/lib/python3.9/site-packages/transformers/trainer.py", line 1539, in train
    return inner_training_loop(
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/ft-zSqjAXBp-py3.9/lib/python3.9/site-packages/transformers/trainer.py", line 1687, in _inner_training_loop
    model, self.optimizer = self.accelerator.prepare(self.model, self.optimizer)
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/ft-zSqjAXBp-py3.9/lib/python3.9/site-packages/accelerate/accelerator.py", line 1227, in prepare
    result = tuple(
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/ft-zSqjAXBp-py3.9/lib/python3.9/site-packages/accelerate/accelerator.py", line 1228, in <genexpr>
    self._prepare_one(obj, first_pass=True, device_placement=d) for obj, d in zip(args, device_placement)
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/ft-zSqjAXBp-py3.9/lib/python3.9/site-packages/accelerate/accelerator.py", line 1104, in _prepare_one
[...]

PR #1352, opened 2024-02-22
Update dpo_llama2.py to fix 8-bit multi-gpu training bug

fancyerii:
fix "ValueError: You can't train a model that has been loaded in 8-bit precision on a different device than the one you're training on." see https://github.com/huggingface/trl/issues/1348

### DIFF
--- a/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
+++ b/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
@@ -129,6 +129,7 @@ if __name__ == "__main__":
         low_cpu_mem_usage=True,
         torch_dtype=torch.float16,
         load_in_4bit=True,
+        device_map={"": Accelerator().local_process_index}
     )
     model.config.use_cache = False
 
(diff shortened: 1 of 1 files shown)


examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
```diff
@@ -176,6 +177,7 @@ def return_prompt_and_responses(samples) -> Dict[str, str]:
         remove_unused_columns=False,
         run_name="dpo_llama2",
     )
+    training_args.gradient_checkpointing_kwargs = dict(use_reentrant=False)
```
younesbelkada, line 180: Can you make that configurable through `script_args` so that users can disable it if they want?

younesbelkada (commented):
Thanks a lot LGTM with one comment - what do you think?

examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
fancyerii: I have create a new pr https://github.com/huggingface/trl/pull/1353

PR #1353, opened 2024-02-22
fix 8-bit multi-gpu training bug 

fancyerii:
see https://github.com/huggingface/trl/issues/1348

### DIFF
--- a/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
+++ b/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
@@ -42,6 +42,10 @@ class ScriptArguments:
         default=True, metadata={"help": "whether to use gradient checkpointing"}
     )
 
+    gradient_checkpointing_use_reentrant: Optional[bool] = field(
+        default=True, metadata={"help": "whether to use reentrant for gradient checkpointing"}
+    )
+
     lora_alpha: Optional[float] = field(default=16, metadata={"help": "the lora alpha parameter"})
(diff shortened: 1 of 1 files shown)


examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
```diff
@@ -74,6 +75,11 @@ class ScriptArguments:
         },
     )
 
+    device_map_local_process_idx: Optional[bool] = field(
+        default=True, metadata={"help": "whether to device map model to local process index, see "
+                                "https://github.com/huggingface/trl/issues/1348"}
+    )
```
younesbelkada, line 81:
```suggestion
```
I think we can remove that and always create `device_map = "device_map": {"": Accelerator().local_process_index}`

Can you just make `gradient_checkpointing_kwargs` configurable here? 🙏

younesbelkada (commented):
Thanks very much @fancyerii ! I left one comment about making `gradient_checkpointing_kwargs` configurable, I think after that we can ship your PR ! 🚀

examples/research_projects/stack_llama_2/scripts/dpo_llama2.py
fancyerii: updated.

younesbelkada:
Thanks @fancyerii ! 
Could you run the styling checks? `make precommit`

fancyerii:
> Thanks @fancyerii ! Could you run the styling checks? `make precommit`

done. 
btw, when I run `make test`, there are many warnings about "wait for wandb.init"

younesbelkada (approved):
Thanks again !

PR #1357, opened 2024-02-23
set seed in sft/dpo/reward_modeling to make result reproducable

sywangyi:


younesbelkada (approved):
Thanks a lot for this work @sywangyi !

PR #1363, opened 2024-02-25
Fix transformers version checking for Python < 3.8

samuki:
To be consistent with the existing code and maintain compatibility for environments running Python versions earlier than 3.8, this pull request includes a minor adjustment to the `is_transformers_greater_than` function.


younesbelkada (approved):
Makes sense, thanks !

PR #1366, opened 2024-02-27
Add some arguments for support XPU

yuanwu2017:


younesbelkada (commented):
Thanks for the nice refactor ! Can you run the styling checks? `make precommit`

younesbelkada (approved):
Thanks a lot !

Issue #1343, opened 2024-02-21
[FEATURE REQUEST] Allow setting packing=True for training but =False for evaluation in the same trainer


Peter-Devine:
* I would like to train on a very large dataset, but eval on a very small dataset
* I need to set packing=True with the SFT trainer because of the amazing efficiencies it brings.
* However, this means that I get an `ValueError: Error occurred while packing the dataset. Make sure that your dataset has enough samples to at least yield one packed sequence.` error.
* This is due to the fact that the eval dataset is very small.
* I would like to be able to set packing=True for the training dataset and eval_packing=False for the evaluation set.
* This feature is implemented by Axolotl [here](https://github.com/OpenAccess-AI-Collective/axolotl/blob/5a5d47458d9aaf7ead798d15291ba3d9bef785c5/README.md?plain=1#L681), so it should be technically possible.

Would TRL consider implementing such a feature?

Thanks!

younesbelkada:
Hi @Peter-Devine 
Absolutely yes! I'll work on it!

Peter-Devine:
@younesbelkada Not all heroes wear capes - thanks!

younesbelkada:
Hi @Peter-Devine 
#1369 should enable this!

Peter-Devine:
@younesbelkada 
Hey, sorry, just got round to checking this. First, thanks for the quick fix!

Running it, it seems like it doesn't log the loss of each (or any) eval dataset I have, but only logs the runtime, samples per second, and steps per second.

<img width="691" alt="image" src="https://github.com/huggingface/trl/assets/49399312/2849093d-0213-447c-b62f-91d045b7c60b">

Here is the code I used to train:

```python
from dataclasses import dataclass, field

import torch
from datasets import load_dataset, load_from_disk
from tqdm import tqdm
from transformers import TrainingArguments
from trl import SFTTrainer
import os

from unsloth import FastLanguageModel

tqdm.pandas()

# ###############
# Model & Tokenizer
# ###############

model, tokenizer = FastLanguageModel.from_pretrained(
    model_name = model_name_or_path,
    max_seq_length = max_seq_length,
    dtype = dtype,
    load_in_4bit = load_in_4bit,
)
tokenizer.pad_token = tokenizer.eos_token

model = FastLanguageModel.get_peft_model(
    model,
    r = lora_r, # Choose any number > 0 ! Suggested 8, 16, 32, 64, 128
[...]

PR #1369, opened 2024-02-27
FEAT: [`SFTTrainer`] Add  `eval_packing`

younesbelkada:
Fixes: https://github.com/huggingface/trl/issues/1343

Adds the possibility for users to enable/disable packing for the eval_dataset using a ternary logic (if None -> packing else eval_packing)

cc @lvwerra @kashif

### DIFF
--- a/docs/source/sft_trainer.mdx
+++ b/docs/source/sft_trainer.mdx
@@ -271,6 +271,7 @@ trainer.train()
 ```
 
 Note that if you use a packed dataset and if you pass `max_steps` in the training arguments you will probably train your models for more than few epochs, depending on the way you have configured the packed dataset and the training protocol. Double check that you know and understand what you are doing.
+If you don't want to pack your `eval_dataset`, you can pass `eval_packing=False` to the `SFTTrainer` init method.
 
 #### Customize your prompts using packed dataset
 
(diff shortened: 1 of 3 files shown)


kashif (approved):

federicoBetti:
Hi @younesbelkada ,
i'm using eval_packing = False to better compute metrics on the evaluation dataset. 
The problem is when using it together with `batch_eval_metrics`, the Trainer crashes when it tries to delete the tensors of losses, logits, labels, inputs.

```
            if self.args.batch_eval_metrics:
                if self.compute_metrics is not None and logits is not None and labels is not None:
                    is_last_step = self.accelerator.gradient_state.end_of_dataloader
                    if args.include_inputs_for_metrics:
                        metrics = self.compute_metrics(
                            EvalPrediction(predictions=logits, label_ids=labels, inputs=inputs),
                            compute_result=is_last_step,
                        )
                    else:
                        metrics = self.compute_metrics(
                            EvalPrediction(predictions=logits, label_ids=labels),
                            compute_result=is_last_step,
                        )

                del losses, logits, labels, inputs
```

This is due to the fact that, when packing is False, the labels are not kept in the `input` tensor, thus loss and labels after the prediction steps are None. And here the code crashes as you're trying to delete a None object. 
Is there something I'm missing to make it work properly?

PR #1382, opened 2024-02-29
Add CPOTrainer

fe1ixxu:
Hi! This PR wants to add CPOTrainer proposed in the paper [Contrastive Preference Optimization: Pushing the Boundaries of LLM
Performance in Machine Translation](https://arxiv.org/pdf/2401.08417.pdf)

The CPO method is one of the algorithm for building the state-of-the-art LLM-based translation model: [ALMA](https://github.com/fe1ixxu/ALMA)


fe1ixxu:
cc @kashif @lewtun

kashif:
@fe1ixxu how close is the trainer in terms of code to the DPOTrainer? Can one subclass from it?

fe1ixxu:
@kashif Thanks for the quick response! CPO is an approximation of DPO. The key difference between CPOTrainer and DPOTrainer is:
- Remove the need of reference model
- add an extra [NLL loss](https://github.com/fe1ixxu/trl/blob/31fd808adb86e8c044574f27234c9bc18254dbac/trl/trainer/cpo_trainer.py#L774) for preferred data

I'm uncertain whether subclassing CPOTrainer from DPOTrainer is a proper idea, as DPOTrainer introduces numerous features related to reference models that are unnecessary for CPOTrainer.

fe1ixxu:
Hi @kashif CPO docs has been finished now! Thanks!

kashif (approved):

docs/source/cpo_trainer.mdx
```diff
@@ -0,0 +1,102 @@
+# CPO Trainer
+
+Contrastive Preference Optimization (CPO) as introduced in the paper [Contrastive Preference Optimization: Pushing the Boundaries of LLM Performance in Machine Translation](https://arxiv.org/pdf/2401.08417.pdf) by Haoran Xu, Amr Sharaf, Yunmo Chen, Weiting Tan, Lingfeng Shen, Benjamin Van Durme, Kenton Murray, and Young Jin Kim. At a high-level CPO  trains models to
```
lewtun, line 3: nit (hf papers are better for discussion):

```suggestion
Contrastive Preference Optimization (CPO) as introduced in the paper [Contrastive Preference Optimization: Pushing the Boundaries of LLM Performance in Machine Translation](https://huggingface.co/papers/2401.08417) by Haoran Xu, Amr Sharaf, Yunmo Chen, Weiting Tan, Lingfeng Shen, Benjamin Van Durme, Kenton Murray, and Young Jin Kim. At a high-level, CPO  trains models to
```

docs/source/cpo_trainer.mdx
```diff
@@ -0,0 +1,102 @@
...
+avoid generating adequate but not perfect translations in the Machine Translation (MT) task.
```
lewtun, line 4: Does CPO generalise beyond machine translation? My understanding is that Eq. (3) of the paper is a general approximation to the DPO loss, so perhaps we can explain this as follows:

```suggestion
avoid generating adequate, but not perfect translations in Machine Translation (MT) tasks. However, CPO is a general approximation to the DPO loss and can be applied to other domains like chat.
```

trl/trainer/cpo_trainer.py
```diff
@@ -0,0 +1,959 @@
...
+        # Add tags for models that have been loaded with the correct transformers version
+        if hasattr(self.model, "add_model_tags"):
+            self.model.add_model_tags(self._tag_names)
+
+        if not hasattr(self, "accelerator"):
+            raise AttributeError(
+                "Your `Trainer` does not have an `accelerator` object. Consider upgrading `transformers`."
+            )
+
+    def _prepare_deepspeed(self, model: PreTrainedModelWrapper):
```
lewtun, line 305: Since there is no reference model, is this method still needed? We needed it in the DPO trainer to shard the reference model, but `accelerate` should handle the sharding of the single `model` in this trainer

trl/trainer/cpo_trainer.py
```diff
@@ -0,0 +1,959 @@
...
+            The losses tensor contains the CPO loss for each example in the batch.
+            The chosen_rewards and rejected_rewards tensors contain the rewards for the chosen and rejected responses, respectively.
+        """
+        logits = (policy_chosen_logps - policy_rejected_logps).to(self.accelerator.device)
+
+        # The beta is a temperature parameter for the CPO loss, typically something in the range of 0.1 to 0.5.
+        # We ignore the reference model as beta -> 0. The label_smoothing parameter encodes our uncertainty about the labels and
+        # calculates a conservative CPO loss.
+        if self.loss_type == "sigmoid":
+            losses = (
```
lewtun, line 616: I don't quite understand where this equation comes from - is it Eq. (5) from the paper in the limit where `label_smoothing -> 0` or something else? It would be good to include a small comment here which points to which part of the paper we're using

lewtun (approved):
Thank you for this very nice implementation of CPO @fe1ixxu 🔥 ! I left a few small comments and a suggestion to remove a deepspeed function I don't think we need. Apart from that LGTM!

docs/source/cpo_trainer.mdx
fe1ixxu: Thanks for the suggestion! The link from HF sounds good :)

trl/trainer/cpo_trainer.py
fe1ixxu: Thanks for the reminder! I have deleted this function since it won't be used for CPO training.

trl/trainer/cpo_trainer.py
fe1ixxu: This is the loss of L_prefer in Eq. 3 from the paper where `label_smoothing -> 0`. The total loss of CPO is L_prefer + L_nll showing here: https://github.com/fe1ixxu/trl/blob/b3004734132fa9862315e8777e8159b38f6d81d9/trl/trainer/cpo_trainer.py#L728C9-L728C47

lewtun: Ah perfect, that makes sense! I think all we need is a small comment that this reduces to Eq. (3) from the paper when label_smoothing -> 0 and then we're good to go - thank you!

fe1ixxu: Thanks! I have added the comment!

PR #1383, opened 2024-02-29
[KTO] merge eval dataset only if it exists

kashif:
exception if there is no eval dataset


lewtun (approved):
LGTM, but could we also add a unit test to ensure we really solve the issue?

kashif:
thanks! test added

PR #1391, opened 2024-03-01
Log ddpo reward as float to fix numpy conversion during bf16 training

skavulya:
Fixes `TypeError: Got unsupported ScalarType BFloat16` when logging the reward during bf16 finetuning with ddpo


younesbelkada (approved):
Thank you for the fix!

PR #1413, opened 2024-03-10
[KTO] model init when args are given

kashif:
fix model init when args are given


lewtun (approved):
LGTM! Let's see what @younesbelkada thinks about the failing doc build test before merging

PR #1416, opened 2024-03-11
Add support for FSDP+QLoRA and DeepSpeed ZeRO3+QLoRA

pacman100:
### What does this PR do? 
1. `prepare_model_for_kbit_training` and `peft_module_casting_to_bf16` should be disabled when using FSDP+QLoRA or DeepSpeed ZeRO3+QLoRA.

This PR should be merged after Transformers PR https://github.com/huggingface/transformers/pull/29587


pacman100:
cc @younesbelkada

trl/trainer/sft_trainer.py
```diff
@@ -186,7 +186,15 @@ def __init__(
                     inspect.signature(prepare_model_for_kbit_training).parameters
                 )
                 gradient_checkpointing_kwargs = getattr(args, "gradient_checkpointing_kwargs", None) or {}
-                if getattr(model, "is_loaded_in_8bit", False) or getattr(model, "is_loaded_in_4bit", False):
+                is_sharded_qlora = False
+                if getattr(model, "is_loaded_in_4bit", False):
```
younesbelkada, line 190:
```suggestion
                # Below is to support QLoRA + FSDP / DS-Zero3 - one should never call 
                # peft_module_casting_to_bf16 or prepare_model_for_kbit_training when doing 
                # QLoRA + FSDP / DS-Zero3
                if getattr(model, "is_loaded_in_4bit", False):
```

younesbelkada (approved):
Thanks so much !

younesbelkada:
I can't repro the CI failure locally and on main, will fix in a follow up PR!

PR #1419, opened 2024-03-12
FEAT: Add CLIs in TRL ! 

younesbelkada:
# What does the PR do?

This PR introduces a new feature in TRL - CLIs for DPO and SFTTrainer! 

```bash
trl sft --config config.yaml --output_dir test-llama
```

```bash
trl dpo --config config.yaml --output_dir test-llama
```

All arguments that are supported in `ModelConfig` should be supported by the CLI together with arguments from `TrainingArguments` or `transformers`.

Users will need to first call `accelerate config` before running the CLI to use custom accelerate configs


pyproject.toml
```diff
@@ -19,4 +19,4 @@ extend-select = ["E", "F", "I", "W", "UP", "B", "T", "C"]
 
 [tool.ruff.lint.isort]
 lines-after-imports = 2
-known-first-party = ["trl"]
+known-first-party = ["trl"]
```
lvwerra, line 22: what does this do?

setup.py
```diff
@@ -97,6 +97,9 @@
     ],
     url="https://github.com/huggingface/trl",
     packages=find_packages(),
+    entry_points={
+        "console_scripts": ["trl=trl.commands.trl_cli:main"],
```
lvwerra, line 101: maybe renaming `trl_cli` to `cli` would be a bit less redundant?

trl/__init__.py
```diff
@@ -2,45 +2,121 @@
...
+            DDPOSchedulerOutput,
+            DDPOStableDiffusionPipeline,
+            DefaultDDPOStableDiffusionPipeline,
+        )
+        from .trainer import DDPOConfig, DDPOTrainer
+
+else:
+    import sys
+
+    sys.modules[__name__] = _LazyModule(__name__, globals()["__file__"], _import_structure, module_spec=__spec__)
```
lvwerra, line 122: Does the lazy loading have any downsides? looks like a pretty dramatic change

lvwerra: In any case we should test this extensively

trl/commands/sft.py
```diff
@@ -0,0 +1,148 @@
...
+    --gradient_accumulation_steps=16 \
+    --output_dir="sft_openassistant-guanaco" \
+    --logging_steps=1 \
+    --num_train_epochs=3 \
+    --max_steps=-1 \
+    --push_to_hub \
+    --gradient_checkpointing \
+    --use_peft \
+    --lora_r=64 \
+    --lora_alpha=16
```
lvwerra, line 47: that's redundant, right?

trl/commands/sft.py
```diff
@@ -0,0 +1,148 @@
...
+            tokenizer=tokenizer,
+            packing=True,
+            peft_config=get_peft_config(model_config),
+            callbacks=[RichProgressCallback],
+        )
+
+    trainer.train()
+
+    with console.status("[bold green]Training completed ! saving the model ..."):
+        trainer.save_model(training_args.output_dir)
```
lvwerra, line 155:
```suggestion
    with console.status(f"[bold green]Training completed! Saving the model to {training_args.output_dir}"):
        trainer.save_model(training_args.output_dir)
```

and we should also push to hub by default :)

trl/commands/sft.py
```diff
@@ -0,0 +1,148 @@
+# flake8: noqa
```
lvwerra, line 1: So i guess this means we can't use `trl/examples/scripts/sft.py` directly for this?

setup.py
younesbelkada: Makes sense, done !

pyproject.toml
younesbelkada: ohh let me revert that

younesbelkada: done!

trl/commands/sft.py
younesbelkada: yes!

trl/commands/sft.py
younesbelkada: yeah for now but I think we can move the example folders there, let me think a bit

setup.cfg
```diff
@@ -1,2 +1,5 @@
 [metadata]
 license_file = LICENSE
+
+[options.packages.find]
+include = examples/scripts/*.py
```
lvwerra, line 5: can also be removed, no?

trl/commands/cli.py
```diff
@@ -0,0 +1,82 @@
...
+            )
+
+        # Get the required args
+        config = args.config
+
+        # if the configuration is None, create a new `output_dir` variable
+        config_parser = YamlConfigParser(config, [args, training_args, model_config])
+        trl_examples_dir = os.path.dirname(__file__)
+
+        model_name = model_config.model_name_or_path
```
lvwerra, line 59: I wonder if it would be cleaner to just pass all the args as they are to the downstream script rather than parsing them here and then passing them as a string. We could add the logic to update the config with passed args inside the dpo.py and sft.py so they would also immediately profit from being able to be called with a config. wdyt?

lvwerra (commented):
Looks great, mainly left a comment on where we want to handle the config parsing :)

trl/commands/cli.py
younesbelkada: Indeed the approach you suggested is much cleaner!

trl/commands/cli_utils.py
```diff
@@ -0,0 +1,95 @@
...
+            return_remaining_strings=True
+        )
+
+        self.config_parser = YamlConfigParser(parsed_args.config)
+        args, training_args, model_config = self.config_parser.merge_dataclasses(
+            ((parsed_args, parsed_training_args, parsed_model_config))
+        )
+
+        training_args.gradient_checkpointing_kwargs = dict(use_reentrant=args.gradient_checkpointing_use_reentrant)
+        return args, training_args, model_config
```
lvwerra, line 95: can we make it a bit more agnostic such that we don't hardcode the order and the kind of args? e.g. like the `HfArgumentParser` does it. the chat interface for example will only have one dataclass to pass and maybe a future method might require 3 or 4.

trl/commands/config_parser.py
```diff
@@ -0,0 +1,102 @@
...
+# limitations under the License.
+import os
+from copy import deepcopy
+from dataclasses import fields
+from typing import Any, List
+
+import yaml
+
+
+class YamlConfigParser:
```
lvwerra, line 22: I would move this into `cli_utils.py`

trl/commands/cli_utils.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def parse_args_and_config(self):
+        dataclasses = self.parse_args_into_dataclasses(return_remaining_strings=True)
+        # Pop the last element which should be the remaining strings
+        dataclasses = dataclasses[:-1]
+        self.config_parser = None
+
+        for parser_dataclass in dataclasses:
+            if hasattr(parser_dataclass, "config"):
+                self.config_parser = YamlConfigParser(parser_dataclass.config)
```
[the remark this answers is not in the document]
younesbelkada: Makes totally sense - fixed it !

lvwerra (approved):

PR #1424, opened 2024-03-13
Create standard dataset for TRL

vwxyzjn:


younesbelkada (approved):
Thank you for the clean up and standardizing the DPO datasets ! 🚀 This is going to be very helpful for the CLI work ! 

cc @lewtun @lvwerra @kashif FYI

PR #1429, opened 2024-03-14
[Kto] torch_dtype kwargs fix

kashif:
since the kwargs is a dict the torch_dtype might be a string but transformer needs the dtype


younesbelkada (approved):
Thanks ! we could perhaps propagate that logic in sfttrainer & dpotrainer but let's merge it to make the CI happy !

PR #1435, opened 2024-03-17
ORPO trainer

kashif:
ORPO trainer 

[Reference-free Monolithic Preference Optimization with Odds Ratio](https://arxiv.org/abs/2403.07691)

cc @jiwooya1000 

- [x] figure out what to log
- [x] add logging section to the docs


kashif:
cc @philschmid

examples/scripts/orpo.py
```diff
@@ -0,0 +1,139 @@
...
+
+from datasets import Dataset, load_dataset
+from transformers import AutoModelForCausalLM, AutoTokenizer, HfArgumentParser
+
+from trl import ModelConfig, ORPOConfig, ORPOTrainer, get_peft_config
+
+
+@dataclass
+class ScriptArguments:
+    sanity_check: bool = field(default=True, metadata={"help": "only train on 1000 samples"})
```
lewtun, line 66: I recommend removing this altogether

trl/trainer/orpo_config.py
```diff
@@ -0,0 +1,68 @@
...
+    Using [`HfArgumentParser`] we can turn this class into
+    [argparse](https://docs.python.org/3/library/argparse#module-argparse) arguments that can be specified on the
+    command line.
+
+    Parameters:
+        max_length (`int`, defaults to `None`):
+            The maximum length of the sequences in the batch. This argument is required if you want to use the default data collator.
+        max_prompt_length (`int`, defaults to `None`):
+            The maximum length of the prompt. This argument is required if you want to use the default data collator.
+        max_target_length (`int`, defaults to `None`):
```
lewtun, line 34: Since we have a new trainer, should we try to harmonise the names like we did for the KTO trainer?
```suggestion
        max_completion_length (`int`, defaults to `None`):
```

trl/trainer/orpo_trainer.py
```diff
@@ -0,0 +1,942 @@
...
+        optimizers (`Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR]`):
+            The optimizer and scheduler to use for training.
+        preprocess_logits_for_metrics (`Callable[[torch.Tensor, torch.Tensor], torch.Tensor]`):
+            The function to use to preprocess the logits before computing the metrics.
+        peft_config (`Dict`, defaults to `None`):
+            The PEFT configuration to use for training. If you pass a PEFT configuration, the model will be wrapped in a PEFT model.
+        compute_metrics (`Callable[[EvalPrediction], Dict]`, *optional*):
+            The function to use to compute the metrics. Must take a `EvalPrediction` and return
+            a dictionary string to metric values.
+        dataset_num_proc (`Optional[int]`, *optional*):
```
lewtun, line 92: I think this could live in the `ORPOConfig` (ideally we want nearly everything that is not a callable to live in a single config so it can be easily tweaked at the command line)

trl/trainer/orpo_trainer.py
```diff
@@ -0,0 +1,942 @@
...
+
+        if is_encoder_decoder:
+            concatenated_batch["concatenated_input_ids"] = batch["prompt_input_ids"].repeat(2, 1).to(device=device)
+            concatenated_batch["concatenated_attention_mask"] = (
+                batch["prompt_attention_mask"].repeat(2, 1).to(device=device)
+            )
+
+        return concatenated_batch
+
+    def or_loss(
```
lewtun, line 593: I think `or_loss` is a bit cryptic, so would prefer either `orpo_loss` or this:
```suggestion
    def odds_ratio_loss(
```

trl/trainer/orpo_trainer.py
```diff
@@ -0,0 +1,942 @@
...
+        self,
+        policy_chosen_logps: torch.FloatTensor,
+        policy_rejected_logps: torch.FloatTensor,
+    ) -> Tuple[torch.FloatTensor, torch.FloatTensor, torch.FloatTensor]:
+        """Compute ORPO's OR loss for a batch of policy and reference model log probabilities.
```
lewtun, line 598:
```suggestion
        """Compute ORPO's odds ratio (OR) loss for a batch of policy and reference model log probabilities.
```

trl/trainer/orpo_trainer.py
```diff
@@ -0,0 +1,942 @@
...
+        Args:
+            policy_chosen_logps: Log probabilities of the policy model for the chosen responses. Shape: (batch_size,)
+            policy_rejected_logps: Log probabilities of the policy model for the rejected responses. Shape: (batch_size,)
+
+        Returns:
+            A tuple of three tensors: (losses, chosen_rewards, rejected_rewards).
+            The losses tensor contains the ORPO loss for each example in the batch.
+            The chosen_rewards and rejected_rewards tensors contain the rewards for the chosen and rejected responses, respectively.
+        """
+        log_odds = (policy_chosen_logps - policy_rejected_logps) - (
```
lewtun, line 609:
```suggestion
        # Derived from Eqs. (4) and (7) from https://arxiv.org/abs/2403.07691 by using log identities and exp(log(P(y|x)) = P(y|x)
        log_odds = (policy_chosen_logps - policy_rejected_logps) - (
```

trl/trainer/orpo_trainer.py
```diff
@@ -0,0 +1,942 @@
...
+            policy_nll_loss,
+        ) = self.concatenated_forward(model, batch)
+
+        losses, chosen_rewards, rejected_rewards = self.or_loss(policy_chosen_logps, policy_rejected_logps)
+
+        loss = policy_nll_loss - losses.mean()
+        reward_accuracies = (chosen_rewards > rejected_rewards).float()
+
+        prefix = "eval_" if train_eval == "eval" else ""
+        metrics[f"{prefix}rewards/chosen"] = chosen_rewards.mean().cpu()
```
lewtun, line 746: Maybe we can just keep `logps` and `nll_loss` and also log odds ratio to simplify this?

lewtun (commented):
Very cool to see this elegant method get added so quickly @kashif !

I left a few remarks about what to log etc and to harmonise the example script to be less hard-coded wrt Anthropic HH. I'd also like to see some small experiments which show the metrics look sane for the examples. Otherwise it's looking great!

jiwooya1000:
Thank you so much for such a fast implementation of ORPO @kashif😀

Also, about https://github.com/huggingface/trl/pull/1435#discussion_r1528005067, I think it is a great idea to log the log odds ratio as it can help monitor the effect of β(the weighting hyperparam) as [this report](https://wandb.ai/jiwooya1000/PREF/reports/Mistral-ORPO-7B-Training-Log--Vmlldzo3MTE3MzMy?accessToken=dij4qbp6dcrofsanzbgobjsne9el8a2zkly2u5z82rxisd4wiwv1rhp0s2dub11e).

image

Thank you again for the implementation!

kashif:
ok i'll add the logging next

examples/scripts/orpo.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def extract_anthropic_prompt(prompt_and_response):
```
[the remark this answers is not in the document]
vwxyzjn: We now have standard datasets under https://huggingface.co/datasets/trl-internal-testing/hh-rlhf-trl-style (#1424). 

* https://github.com/huggingface/trl/blob/main/examples/datasets/anthropic_hh.py creates the standard dataset (prompt, chosen, rejected) 
* https://github.com/huggingface/trl/blob/main/examples/datasets/tokenize_ds.py is a usage example. 

In this case, maybe you could try:

```python
    ds = load_dataset(args.dataset)
    if args.debug:
        for key in ds:
            ds[key] = ds[key].select(range(50))
    tokenizer = AutoTokenizer.from_pretrained(args.model)
        tokenizer.chat_template = "{% for message in messages %}{{message['role'] + ': ' + message['content'] + '\n\n'}}{% endfor %}{{ eos_token }}"

    def process(row):
        row["chosen"] = tokenizer.apply_chat_template(row["chosen"], tokenize=False)
        row["rejected"] = tokenizer.apply_chat_template(row["rejected"], tokenize=False)
        return row

    ds = ds.map(
        process,
        num_proc=1 if args.debug else multiprocessing.cpu_count(),
        load_from_cache_file=False,
    )
    train_dataset = ds["train"]
    eval_dataset = ds["test"]
    trainer = ORPOTrainer(
        model,
        args=orpo_args,
        train_dataset=train_dataset,
        eval_dataset=eval_dataset,
        tokenizer=tokenizer,
    )
```

here the `args.debug` is doing the same thing as `sanity_check`.

vwxyzjn (commented):
Great work! Added some comments.

examples/scripts/orpo.py
```diff
@@ -0,0 +1,121 @@
...
+
+    ################
+    # Dataset
+    ################
+    ds = load_dataset(args.dataset)
+    if orpo_args.debug:
+        for key in ds:
+            ds[key] = ds[key].select(range(50))
+    if tokenizer.chat_template is None:
+        tokenizer.chat_template = "{% for message in messages %}{{message['role'] + ': ' + message['content'] + '\n\n'}}{% endfor %}{{ eos_token }}"
```
lewtun, line 92: I recommend ChatML as the default
```suggestion
        tokenizer.chat_template = "{% if not add_generation_prompt is defined %}{% set add_generation_prompt = false %}{% endif %}{% for message in messages %}{{'<|im_start|>' + message['role'] + '\n' + message['content'] + '<|im_end|>' + '\n'}}{% endfor %}{% if add_generation_prompt %}{{ '<|im_start|>assistant\n' }}{% endif %}"
```

trl/trainer/orpo_trainer.py
lewtun: @kashif WDYT about this proposal to move the arg to the config?

lewtun (approved):
Thanks for iterating @kashif ! I left one final nit and a question about moving the dataset proc args to the config. Apart form that LGTM 🔥

jiwooya1000:
Just gave a quick try with ORPOTrainer with `facebook/opt-350m` + `argilla/ultrafeedback-binarized-preferences-cleaned`, and it seems to be working well😃 Thank you for your work @kashif!

image

trl/trainer/orpo_config.py
```diff
@@ -0,0 +1,72 @@
...
+        model_init_kwargs (`Optional[Dict]`, *optional*):
+            Dict of Optional kwargs to pass when instantiating the model from a string
+        dataset_num_proc (`Optional[int]`, *optional*):
+            The number of workers to use to tokenize the data. Defaults to None.
+    """
+
+    max_length: Optional[int] = None
+    max_prompt_length: Optional[int] = None
+    max_completion_length: Optional[int] = None
+    max_completion_length: Optional[int] = None
```
alvarobartt, line 59: Hi here, this is duplicated!

```suggestion
    max_completion_length: Optional[int] = None
```

trl/trainer/orpo_trainer.py
```diff
@@ -0,0 +1,945 @@
...
+        train_dataset: Optional[Dataset] = None,
+        eval_dataset: Optional[Union[Dataset, Dict[str, Dataset]]] = None,
+        tokenizer: Optional[PreTrainedTokenizerBase] = None,
+        model_init: Optional[Callable[[], PreTrainedModel]] = None,
+        callbacks: Optional[List[TrainerCallback]] = None,
+        optimizers: Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
+        preprocess_logits_for_metrics: Optional[Callable[[torch.Tensor, torch.Tensor], torch.Tensor]] = None,
+        peft_config: Optional[Dict] = None,
+        compute_metrics: Optional[Callable[[EvalLoopOutput], Dict]] = None,
+    ):
```
alvarobartt, line 110: Wouldn't it be nice to also include here the `ORPO` specific args from the `ORPOConfig` so that users can still provide `TrainingArguments` to `args` within the `ORPOTrainer` while still being able to provide i.e. `beta`? Asking because that implementation would be closer to the current `DPOTrainer` and friends cc @lewtun

trl/trainer/orpo_trainer.py
```diff
@@ -0,0 +1,945 @@
...
+                )
+
+            self.use_dpo_data_collator = True
+        else:
+            self.use_dpo_data_collator = False
+
+        if args.disable_dropout:
+            disable_dropout_in_model(model)
+
+        self.max_length = max_length
```
alvarobartt, line 255: This will fail as `max_length` is not provided as an `arg` within `ORPOTrainer` so `max_length` is never assigned if provided via `ORPOConfig` in `args`

alvarobartt:
```suggestion
        self.max_length = args.max_prompt_length or max_length
```

trl/trainer/orpo_trainer.py
```diff
@@ -0,0 +1,945 @@
...
+        self.generate_during_eval = args.generate_during_eval
+        self.label_pad_token_id = args.label_pad_token_id
+        self.padding_value = args.padding_value if args.padding_value is not None else tokenizer.pad_token_id
+        self.max_prompt_length = max_prompt_length
```
alvarobartt, line 259: Same thing here, this will fail if provided via `args` in the `ORPOConfig`

trl/trainer/orpo_trainer.py
```diff
@@ -0,0 +1,945 @@
...
+                UserWarning,
+            )
+            max_length = 512
+        if args.max_prompt_length is None:
+            warnings.warn(
+                "`max_prompt_length` is not set in the ORPOConfig's init"
+                " it will default to `128` by default, but you should do it yourself in the future.",
+                UserWarning,
+            )
+            max_prompt_length = 128
```
alvarobartt, line 220: Should be updated to either assign the value to `self` here, or to assign the value from `args`

trl/trainer/orpo_trainer.py
```diff
@@ -0,0 +1,945 @@
...
+
+        self.max_length = max_length
+        self.generate_during_eval = args.generate_during_eval
+        self.label_pad_token_id = args.label_pad_token_id
+        self.padding_value = args.padding_value if args.padding_value is not None else tokenizer.pad_token_id
+        self.max_prompt_length = max_prompt_length
+        self.truncation_mode = args.truncation_mode
+        self.tokenizer = tokenizer
+
+        self.beta = args.beta
```
alvarobartt, line 263: `beta` could be an arg within the `ORPOTrainer.__init__` as in `DPOTrainer`

trl/trainer/orpo_trainer.py
alvarobartt: Yes, it works, but those are not exposed as part of the `__init__` of the `ORPOTrainer` as in `DPOTrainer`, see https://github.com/huggingface/trl/blob/423991c204c4b36d4aea6955eeb53140fbbea9bb/trl/trainer/dpo_trainer.py#L144-L179

trl/trainer/orpo_trainer.py
alvarobartt: Something along the following lines should do the work

```suggestion
        if args.max_length is None:
            warnings.warn(
                "`max_length` is not set in the ORPOConfig's init"
                " it will default to `512` by default, but you should do it yourself in the future.",
                UserWarning,
            )
            max_length = 512
        else:
            max_length = args.max_length
        if args.max_prompt_length is None:
            warnings.warn(
                "`max_prompt_length` is not set in the ORPOConfig's init"
                " it will default to `128` by default, but you should do it yourself in the future.",
                UserWarning,
            )
            max_prompt_length = 128
        else:
            max_prompt_length = args.max_prompt_length
```

alvarobartt: I can confirm this works fine when providing those via `args: ORPOConfig` 👍🏻  Running an ORPO fine-tune right now

trl/trainer/orpo_trainer.py
alvarobartt: Oops, edited the suggestion above

trl/trainer/orpo_trainer.py
lewtun: Hey @alvarobartt as the SFT and DPO trainers have evolved, it has become clear that it's better to group all the non-callable arguments into a single config because otherwise you end up with a situation where some parameters are set in the `args` and others are set in the `__init__` which is confusing + harder to maintain.

We've started applying this approach to new trainers like KTO and ORPO and will gradually refactor the others to follow this. Is there some cases where it would be beneficial to preserve the current `__init__` based approach to set certain hyperparameters?

alvarobartt: Oh not at all, it was more in terms of alignment, as it was easier to always assume that one had to instantiate the `TrainingArguments` and then the additional stuff on top of the `...Trainer.__init__`, because otherwise one needs to import the specific configuration dataclass i.e. `ORPOConfig`, but I guess it's just to educate users to use the current approach to avoid `__init__` arg-duplication and to use the `...Config` dataclasses instead of just `TrainingArguments` 👍🏻

alvarobartt: Anyway, even if keeping only `args` there's still an issue with the values for both `max_length` and `max_prompt_length` that when provided the `ORPOTrainer.__init__` breaks, I'll leave the solution unresolved and resolve the rest of the comments, thanks!

trl/trainer/orpo_trainer.py
alvarobartt: Here's the fix @lewtun, feel free to close the rest of my comments here! 👍🏻

alvarobartt (commented):
Maybe it would also be nice to include the `ORPOTrainer` within the `README.md` listing of supported trainers see https://github.com/huggingface/trl?tab=readme-ov-file#highlights

kashif:
@alvarobartt ok yes good idea! adding

PR #1441, opened 2024-03-19
Fix PPOTrainer README example

nikihowe:
The current PPOTrainer example given in the main README does not work for two reasons: 1) pad token, 2) relationship between batch size, minibatch size, and gradient accumulation steps.

This PR addresses these two issues :).

Fixes: https://github.com/huggingface/trl/issues/1442


younesbelkada (approved):
Thanks for fixing the PPOTrainer README example ! 🚀

PR #1456, opened 2024-03-19
Use the standard dataset for DPO CLI

vwxyzjn:
This PR uses the standard dataset created in #1424 and also updates the DPO CLI docs.


younesbelkada (approved):
Thanks ! LGTM once the styling checks pass !

PR #1458, opened 2024-03-20
Fix chat CLI for model revisions

lewtun:
This PR passes the `revision` arg to the tokenizer so one can chat with models at a given branch/revision. I also set the padding token ID to suppress warnings on models that don't have this in their config.


kashif (approved):

younesbelkada (approved):
Makes sense thanks !

Issue #1464, opened 2024-03-20
Tokenizer bos token problem when using KTO trainer


eleluong:
While using the KTO trainer, I observed an issue. Upon further debugging, I identified the code where the bos_token is added to the beginning of the _input_ids. However, some tokenizers lack a bos_token, resulting in None instead of the expected bos_token.

You guys can use below code to check if there is any bos token or not:
```py

if self.tokenizer.bos_token_id != None:
    batch[f"{prefix}prompt_input_ids"] = [self.tokenizer.bos_token_id] + all_tokens["prompt_input_ids"]
    batch[f"{prefix}prompt_attention_mask"] = [1] + all_tokens["prompt_attention_mask"]
    batch[f"{prefix}completion_input_ids"] = (
        [self.tokenizer.bos_token_id]
        + all_tokens["prompt_input_ids"]
        + all_tokens["answer_input_ids"]
        + [self.tokenizer.eos_token_id]
    )
    batch[f"{prefix}completion_attention_mask"] = (
        [1] + all_tokens["prompt_attention_mask"] + all_tokens["answer_attention_mask"] + [1]
    )
else:
    batch[f"{prefix}prompt_input_ids"] = all_tokens["prompt_input_ids"]
    batch[f"{prefix}prompt_attention_mask"] = all_tokens["prompt_attention_mask"]
    batch[f"{prefix}completion_input_ids"] = (
        all_tokens["prompt_input_ids"]
        + all_tokens["answer_input_ids"]
        + [self.tokenizer.eos_token_id]
    )
    batch[f"{prefix}completion_attention_mask"] = (
        all_tokens["prompt_attention_mask"] + all_tokens["answer_attention_mask"] + [1]
    )  
```

codedecde:
Concretely, consider the Qwen 7B model

```python
from transformers import AutoTokenizer
tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen-7B-chat", trust_remote_code=True)
assert tokenizer.bos_token_id is None  # This is true
```

The presence of a BOS token is strictly not necessary for DPO / KTO. Hence the presence should be optional.

PR #1466, opened 2024-03-22
handling models with no bos_token_id

codedecde:
Addressing https://github.com/huggingface/trl/issues/1464, so that DPO and KTO works for models with tokenizers that do not have a bos_token_id

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -655,14 +655,14 @@ class DPOTrainer(Trainer):
                     "last token due to tokenizer merge ops."
                 )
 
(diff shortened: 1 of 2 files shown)


younesbelkada (approved):
I think this is nice! Can you rebase with main branch? cc @kashif @lewtun this should be safe enough no? Wdyt?

kashif (approved):

younesbelkada:
@codedecde we can merge this PR after you rebase your branch with main

chujiezheng:
@codedecde I met the same bugs. We should also check whether the first token in `input_ids` is `bos_token_id`.

PR #1468, opened 2024-03-22
Allow streaming (datasets.IterableDataset)

### LINKED ISSUE #1455: Support streaming + packing in SFTTrainer
Now that the alignment handbook has support for continued pretraining (using SFTTrainer under the hood), it'd be great to account for those cases where massive datasets are used. In such scenarios we want to use packing + streaming. However, in my experience that does not work well.

From my experiments and digging I have uncovered the following. In a typical scenario we first load a dataset with a given text column e.g. `text` and with streaming enabled through load_dataset. The result is a `datasets.IterableDataset`. This is passed to the SFTTrainer. However, the first thing goes wrong here when deciding how to prepare the dataset (packing/non-packing):

https://github.com/huggingface/trl/blob/1705aebebafea53ac7ad3509ce6e5002e0f36240/trl/trainer/sft_trainer.py#L404

This if-statement will evaluate to true because a `datasets.IterableDataset` is [also a `torch.IterableDataset`](https://github.com/huggingface/datasets/blob/ca8409a8bec4508255b9c3e808d0751eb1005260/src/datasets/iterable_dataset.py#L1187-L1188). As such tokenization/packing never happens for "stream" (IterableDataset) datasets and the dataset remains as-is: a dataset with just a text field.

The first suggestion therefore is to check explicitly that this is NOT a `datasets.IterableDataset`, so replace the if-statement by:

```python
    if isinstance(dataset, (torch.utils.data.IterableDataset, torch.utils.data.Dataset, ConstantLengthDataset)) and not isinstance(dataset, datasets.IterableDataset):
        return dataset
```

Unfortunately even after that change we run into undesired behavior. Because packing is enabled, we are preparing a constantlengthdataset. That's fine - it's iter method is a nice generator that will not by itself process the whole massive streaming dataset. However, it seems that that behavior IS triggered here, where the constantlengthdataset is exhausted into a new, regular Dataset (which is not an `IterableDataset` anymore!)

https://github.com/huggingface/trl/blob/1705aebebafea53ac7ad3509ce6e5002e0f36240/trl/trainer/sft_trainer.py#L520-L522

For small datasets that is not an issue, and I imagine that it is useful then to have all data preprocessed in-memory. However, for streaming datasets that is exactly what we do not want! Both in terms of memory and in terms of how long this preprocessing takes. This problem can be resolved by simply returning the constantlengthdataset itself, without transforming it into a regular Dataset.

```python
if isinstance(dataset, IterableDataset):
    return constant_length_iterator
```

If you agree I can make a PR for these two changes. However, it is also possible that I missed something that motivates the reason why the current implementation does not deal well with streaming.

---

To repro the original issue, you can install the alignment handbook ([from this branch](https://github.com/BramVanroy/alignment-handbook/tree/allow_streaming)), use the config file below

```yaml
# Model arguments
model_name_or_path: gpt2
model_revision: main
torch_dtype: bfloat16
use_flash_attention_2: false

# Data training arguments
dataset_mixer:
  uonlp/CulturaX: 1.0
dataset_configs:
  - nl
dataset_splits:
  - train
use_streaming: true
preprocessing_num_workers: 8

# SFT trainer config
bf16: true
do_eval: False
evaluation_strategy: "no"
gradient_accumulation_steps: 8
gradient_checkpointing: true
gradient_checkpointing_kwargs:
  use_reentrant: False
hub_model_id: gpt2-cpt-dutch
learning_rate: 1.0e-05
weight_decay: 0.01
log_level: info
logging_steps: 5
logging_strategy: steps
lr_scheduler_type: cosine
max_seq_length: 480
max_steps: 917130
output_dir: data/gpt2-cpt-dutch
overwrite_output_dir: true
per_device_train_batch_size: 4
remove_unused_columns: true
save_strategy: "steps"
save_steps: 100000
save_total_limit: 3
seed: 42
warmup_ratio: 0.01
```

and launch with

```bash
python scripts/run_cpt.py config.yaml
```

BramVanroy:
The motivation for this PR is given in this issue: https://github.com/huggingface/trl/issues/1455#issue-2195471496

Currently using `IterableDataset`s in the SFTTrainer is not plausible because of two issues:

1. an IterableDataset is a subclass of a torch dataset, and therefore will not be considered when deciding whether to use packing

https://github.com/huggingface/trl/blob/1705aebebafea53ac7ad3509ce6e5002e0f36240/trl/trainer/sft_trainer.py#L404-L405

2. when packing, a given dataset is exhausted and loaded fully in-memory with Dataset.from_generator - which defeats the purpose of having an IterableDataset in the first place

https://github.com/huggingface/trl/blob/1705aebebafea53ac7ad3509ce6e5002e0f36240/trl/trainer/sft_trainer.py#L520-L522

This PR remedies both those issues by explicitly checking whether the dataset is a datasets.IterableDataset in point 1, and also in point 2 in which case the packed ConstantLengthDataset is returned as-is (which is also an IterableDataset). 

These changes should make the SFTTrainer better compatible with streaming datasets. The motivation to improve the situation is because in the alignment handbook we also use the SFTTrainer for continued pretraining where massive datasets (streamed) should be supported.

Note: seems that the epoch calculation does not happen correctly when max_steps is given though. With batch size 2, accum. steps 8 and optim steps 7153 with 16 gpus, I get this strange calculation for num examples and num epochs:

```
***** Running training *****
Num examples = 1,831,168
Num Epochs = 9,223,372,036,854,775,807
Instantaneous batch size per device = 2
Total train batch size (w. parallel, distributed & accumulation) = 256
Gradient Accumulation steps = 8
Total optimization steps = 7,153
```

closes #1455

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ trl sft --model_name_or_path facebook/opt-125m --dataset_name imdb --output_dir 
 **DPO:**
 
 ```bash
-trl dpo --model_name_or_path facebook/opt-125m --dataset_name trl-internal-testing/Anthropic-hh-rlhf-processed --output_dir opt-sft-hh-rlhf 
+trl dpo --model_name_or_path facebook/opt-125m --dataset_name trl-internal-testing/hh-rlhf-trl-style --output_dir opt-sft-hh-rlhf 
 ```
 
 **Chat:**
--- a/commands/run_dpo.sh
+++ b/commands/run_dpo.sh
@@ -3,7 +3,7 @@
 # but defaults to QLoRA + PEFT
 OUTPUT_DIR="test_dpo/"
 MODEL_NAME="HuggingFaceM4/tiny-random-LlamaForCausalLM"
-DATASET_NAME="trl-internal-testing/Anthropic-hh-rlhf-processed"
+DATASET_NAME="trl-internal-testing/hh-rlhf-trl-style"
 MAX_STEPS=5
 BATCH_SIZE=2
 SEQ_LEN=128
--- a/docs/source/clis.mdx
+++ b/docs/source/clis.mdx
@@ -63,29 +63,35 @@ The SFT CLI is based on the `examples/scripts/sft.py` script.
 
 ### Direct Policy Optimization (DPO)
 
-First, follow the basic instructions above and run `trl dpo --output_dir <output_dir> <*args>`. Make sure to process your DPO dataset in the TRL format as follows:
+To use the DPO CLI, you need to have a dataset in the TRL format such as 
 
-1- Make sure to pre-tokenize the dataset using chat templates:
+* TRL's Anthropic HH dataset: https://huggingface.co/datasets/trl-internal-testing/hh-rlhf-trl-style
+* TRL's OpenAI TL;DR summarization dataset: https://huggingface.co/datasets/trl-internal-testing/tldr-preference-trl-style
+
+These datasets always have at least three columns `prompt, chosen, rejected`:
+
+* `prompt` is a list of strings.
+* `chosen` is the chosen response in [chat format](https://huggingface.co/docs/transformers/main/en/chat_templating)
+* `rejected` is the rejected response [chat format](https://huggingface.co/docs/transformers/main/en/chat_templating) 
+
+
+To do a quick start, you can run the following command:
 
 ```bash
-python examples/datasets/tokenize_ds.py --model gpt2 --dataset yourdataset
+trl dpo --model_name_or_path facebook/opt-125m --output_dir trl-hh-rlhf --dataset_name trl-internal-testing/hh-rlhf-trl-style
 ```
 
-You might need to adapt the `examples/datasets/tokenize_ds.py` to use yout chat template
 
-2- Format the dataset into TRL format (you can adapt the `examples/datasets/anthropic_hh.py`):
+The DPO CLI is based on the `examples/scripts/dpo.py` script.
+
+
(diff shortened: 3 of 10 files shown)


younesbelkada (commented):
Thanks a lot @BramVanroy for the detailed work on this ! I think there should be no harm supporting this in SFTTrainer, can you run the styling checks? `make precommit` - then we can merge imo

BramVanroy:
> Thanks a lot @BramVanroy for the detailed work on this ! I think there should be no harm supporting this in SFTTrainer, can you run the styling checks? `make precommit` - then we can merge imo

Do youi have any thoughts on the incorrect reporting of number of epochs?

BramVanroy:
Also, I'm not able to run pre-commit:

```
$ make precommit
pre-commit run --all-files
make: pre-commit: No such file or directory
make: *** [Makefile:18: precommit] Error 127
```

younesbelkada:
@BramVanroy hmmm not sure what we can do here to be honest :/

BramVanroy:
> @BramVanroy hmmm not sure what we can do here to be honest :/

Hm yeah, maybe let's keep it like that for now then.

younesbelkada (approved):
Thanks again !

BramVanroy:
@younesbelkada I probably should have included it in this PR, too, but there is another useful (imo) change that is just three lines of code that could make the SFTTrainer even more flexible. I added it to a separate PR: https://github.com/huggingface/trl/pull/1520

PR #1470, opened 2024-03-22
[KTO] Use batching to speed up data processing

lewtun:
The `KTOTrainer` currently applies all dataset preprocessing at the per-example-level via a method called `tokenize_row()`. This isn't ideal because:

- Rust tokenizers support batching which is typically 3-5x faster
- the method can't be hashed and thus parallelised via multiprocessing in `datasets`.

This PR massively speeds up the dataset preprocessing of `KTOTrainer` by doing the following:

* Rewrites the tokenization step to use batching (best practice for Rust tokenizers). This is now called the `_tokenize` method and is done separately to the subsequent token ID processing
* Instead of tokenizing the train and KL datasets separately, we do this once and then modify _get_kl_dataset()` to reverse the token IDs directly. This saves us ~2x processing time since tokenization is the bottleneck
* Introduces a new `_process_tokens()` method that supports multiprocessing
* Migrates all data processing methods to live outside the trainer
* Adds a new `dataset_num_proc` attribute to the `KTOConfig` to allow users to configure the multiprocessing steps

I've also followed the advice of the `datasets` team to bring out the dataset processing functions out of the trainer: this is needed because `num_proc>1` assumes the method and the parent class can be hashed, which is not the case for the parent trainer class.

I've also added some unit tests for the new logic and verified that the `kto.py` example script produces identical results with this refactor.

Screenshot 2024-03-25 at 10 38 34

With this PR, we can now process the whole Anthropic HH dataset of 300k samples in ~3' vs ~15' :)

## Remarks

- Since these data processing methods appear to be quite specific to KTO, I didn't put them in a separate module. Happy to refactor if we want to do this
- If we agree with the general approach, I can apply this in follow up PRs to the DPO trainer and friends.


trl/trainer/kto_trainer.py
```diff
@@ -326,78 +480,157 @@ def make_inputs_require_grad(module, input, output):
...
                 stopping_strategy="all_exhausted",
             )
-        else:
-            interleaved_eval_dataset = None
+            interleaved_train_dataset = interleaved_train_dataset.shuffle(seed=args.data_seed)
+
+            if eval_dataset is not None:
+                interleaved_eval_dataset = interleave_datasets(
+                    [
+                        eval_dataset.filter(lambda x: x["label"], num_proc=args.dataset_num_proc),
```
lewtun, line 627: I am aware of #1476 which removes the interleaving, so will rebase after that PR is merged to make sure we don't re-introduce this step

kashif (approved):

trl/trainer/kto_trainer.py
```diff
@@ -62,6 +63,163 @@
 if is_deepspeed_available():
     import deepspeed
 
+if TYPE_CHECKING:
+    from transformers import PreTrainedTokenizer
+
+
+def _get_kl_dataset(batch) -> Dict:
```
lewtun, line 70: Denote these as internal methods

lvwerra (approved):
LGTM!

PR #1471, opened 2024-03-22
[peft] Update test_reward_trainer.py to fix tests

kashif:
Since we are requiring peft >= 0.4.0

This check fails when peft_version = "0.10.0"


vwxyzjn (approved):
Thanks @kashif!

Issue #1467, opened 2024-03-22
[Question] desirable_weight and undesirable_weight in KTOTrainer


seanexp:
I found that KTOTrainer [logs warning message](https://github.com/huggingface/trl/blob/main/trl/trainer/kto_trainer.py#L366-L385) if desirable_weight or undesirable_weight are not in ideal bounds.

However, KTOTrainer makes [balanced dataset](https://github.com/huggingface/trl/blob/main/trl/trainer/kto_trainer.py#L387-L400) via `interleave_datasets`.

Given that the training dataset is balanced, the `lower_bound` and `upper_bound` calculation seems a bit awkward.

To my understanding the equation below 
```
des_weight_lower_bound = round((len(undesirable) * self.undesirable_weight / len(desirable)) * 1, 2)
```

should be rewritten into 
```
des_weight_lower_bound = round(self.undesirable_weight, 2)
```

as the ratio of `len(undesirable)` and `len(desirable)` equals 1 in `interleaved_train_dataset`.

Please correct me if I'm wrong. (cc. @kashif @kawine )
Thanks in advance!

kawine:
Thanks for pointing this out @seanexp !

This looks like a bug, since I was under the (incorrect) impression that interleaving datasets with all_exhausted would preserve the relative size of the original datasets, which on second look, it appears to not do. Instead, it duplicates the data in an unpredictable way, which we don't want.

The interleaving is also unnecessary, at least in the current version of the code, so I'll make a PR to remove it.

seanexp:
Thanks! @kawine

younesbelkada:
Hi ! This should be fixed by #1499 I believe, do you think we can close this issue?

seanexp:
@younesbelkada Yes, I'll close this issue

PR #1472, opened 2024-03-22
[KTO]: Fix nan losses and crashing job

claralp:
fixes #1447   
- use `nanmean()` instead of `mean()` for losses to avoid nan losses
- remove obsolete `accelerator.gather` for metrics as the metrics are all collected to cpu anyway and averaged later when logging them.

PR #1476, opened 2024-03-24
[KTO] fix minor bugs with data loading and reporting

kawine:
- remove interleaving in KTOTrainer since it is no longer needed for KL estimation and artificially duplicates data to force an even number of positive and negative examples (see https://github.com/huggingface/trl/issues/1467)
- do not report batch statistics for good/bad data if there are zero good/bad examples in the current batch (currently NaNs are returned, which are sometimes converted to floats before being sent to wandb, leading to reward spikes (see https://twitter.com/_lewtun/status/1771493512262942801)

cc @kashif

### DIFF
--- a/examples/scripts/kto.py
+++ b/examples/scripts/kto.py
@@ -17,34 +17,38 @@ Run the KTO training script with the commands below. In general, the optimal con
 
 # Full training:
 python examples/scripts/kto.py \
-    --model_name_or_path=stabilityai/stablelm-2-zephyr-1_6b \
+    --model_name_or_path=trl-lib/qwen1.5-1.8b-sft \
     --per_device_train_batch_size 16 \
     --num_train_epochs 1 \
-    --learning_rate 2e-5 \
+    --learning_rate 1e-5 \
+    --lr_scheduler_type=cosine \
     --gradient_accumulation_steps 1 \
     --logging_steps 10 \
     --eval_steps 500 \
-    --output_dir="kto-aligned-model" \
-    --warmup_steps 150 \
+    --output_dir=kto-aligned-model \
+    --warmup_ratio 0.1 \
     --report_to wandb \
     --bf16 \
     --logging_first_step
 
-# LoRA:
+# QLoRA:
 python examples/scripts/kto.py \
-    --model_name_or_path=stabilityai/stablelm-2-zephyr-1_6b \
-    --per_device_train_batch_size 16 \
+    --model_name_or_path=trl-lib/qwen1.5-1.8b-sft \
+    --per_device_train_batch_size 8 \
     --num_train_epochs 1 \
-    --learning_rate 2e-4 \
+    --learning_rate 1e-4 \
+    --lr_scheduler_type=cosine \
     --gradient_accumulation_steps 1 \
     --logging_steps 10 \
     --eval_steps 500 \
-    --output_dir="kto-aligned-model-lora" \
-    --warmup_steps 150 \
+    --output_dir=kto-aligned-model-lora \
+    --warmup_ratio 0.1 \
     --report_to wandb \
     --bf16 \
     --logging_first_step \
     --use_peft \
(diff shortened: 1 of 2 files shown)


kashif:
thanks @kawine! could you also have a quick peek at https://github.com/huggingface/trl/pull/1472

tests/test_kto_trainer.py
```diff
@@ -39,6 +40,7 @@ def setUpClass(cls):
         cls.t5_model = AutoModelForSeq2SeqLM.from_pretrained(model_id)
         cls.t5_ref_model = AutoModelForSeq2SeqLM.from_pretrained(model_id)
         cls.t5_tokenizer = AutoTokenizer.from_pretrained(model_id)
+        wandb.init(mode="disabled")
```
lewtun, line 43: I would prefer to replace this with a decorator on the whole class like this:

```
from unittest.mock import patch

@patch.dict(os.environ, {"WANDB_DISABLED": "true"})
```

lewtun (approved):
Thanks for these fixes @kawine ! LGTM with a small suggestion to use a decorator for the WANDB env variable

trl/trainer/kto_trainer.py
```diff
@@ -999,13 +989,25 @@ def get_batch_loss_metrics(
...
+        metrics[f"{prefix}rewards/chosen"] = self.accelerator.gather(mean_chosen_reward).nanmean().item()
+        metrics[f"{prefix}logps/chosen"] = self.accelerator.gather(mean_chosen_logps).nanmean().item()
+        metrics[f"{prefix}rewards/rejected"] = self.accelerator.gather(mean_rejected_reward).nanmean().item()
+        metrics[f"{prefix}logps/rejected"] = self.accelerator.gather(mean_rejected_logps).nanmean().item()
         metrics[f"{prefix}rewards/margins"] = metrics[f"{prefix}rewards/chosen"] - metrics[f"{prefix}rewards/rejected"]
         metrics[f"{prefix}kl"] = kl.item()  # has already been gathered in kto_loss
-        metrics[f"{prefix}logps/chosen"] = self.accelerator.gather(mean_chosen_logps).nanmean().cpu()
-        metrics[f"{prefix}logps/rejected"] = self.accelerator.gather(mean_rejected_logps).nanmean().cpu()
 
+        if np.isnan(metrics[f"{prefix}rewards/chosen"]):
```
claralp, line 1000: you can omit these checks if you use
```
        logs[key] = torch.tensor(metrics).nanmean().item()
```
 in line [1203](https://github.com/huggingface/trl/pull/1476/files#diff-f4e653023a53778d00b43a8202a76dbac88b9a790987970f232bafa93e581299R1203). This ignores the nan values when logging metrics

trl/trainer/kto_trainer.py
```diff
@@ -360,15 +360,15 @@ def make_inputs_require_grad(module, input, output):
...
+        num_undesirable = len(train_dataset) - num_desirable
 
-        if len(desirable) != len(undesirable):
+        if num_desirable != num_undesirable:
             # The lower and upper bounds come from Eq. (8) of https://arxiv.org/abs/2402.01306
-            des_weight_lower_bound = round((len(undesirable) * self.undesirable_weight / len(desirable)) * 1, 2)
-            des_weight_upper_bound = round((len(undesirable) * self.undesirable_weight / len(desirable)) * 1.33, 2)
-            und_weight_lower_bound = round((len(desirable) * self.desirable_weight / len(undesirable)) / 1.33, 2)
-            und_weight_upper_bound = round((len(desirable) * self.desirable_weight / len(undesirable)) / 1, 2)
+            des_weight_lower_bound = round((len(num_undesirable) * self.undesirable_weight / len(num_desirable)) * 1, 2)
```
claralp, line 368: isn´t `num_undesirable ` the int length already? For me `len(num_undesirable)` throws an error

PR #1478, opened 2024-03-24
Add `use_cache=False` in `{ORPO,CPO}Trainer.concatenated_forward`

### LINKED ISSUE #1477: `ORPOTrainer` fails with `flash-attention2`
## Description

The `ORPOTrainer` fails when training using `attn_implementation="flash_attention_2"`, since the cache is being used, and falls back to the default configuration i.e. `padding_side="right"` for the tokenizer in this case.

## Bug in code

Missing `use_cache=False` to prevent the model from using the cache, to avoid issues with Flash Attention 2.

https://github.com/huggingface/trl/blob/2ce8e45bb222db68fe9eb92dc8b2a2f1339a2d75/trl/trainer/orpo_trainer.py#L685-L689

alvarobartt:
## Description

This PR fixes the bug described in #1477, similarly to the former issue within the `DPOTrainer` solved by @younesbelkada  (so kudos to him!).

Adding `use_cache=False` when running the forward pass during training is not needed, because we're not in an inference use case, only computing the logits, and it runs into conflicts with Flash Attention 2, so disabling it has no other impact than fixing the collision bug with FA2.

Closes #1477 

cc @kashif @younesbelkada

### DIFF
--- a/trl/trainer/cpo_trainer.py
+++ b/trl/trainer/cpo_trainer.py
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import inspect
 import random
 import warnings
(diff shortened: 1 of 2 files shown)


kashif (approved):

kashif:
thanks @alvarobartt  can you also add it to the CPO Trainer?

PR #1481, opened 2024-03-25
Update KTO example with good dataset & chat format

lewtun:
This PR removes the custom preprocessing that's applied with the Anthropic dataset in favour of:

* A small, high quality dataset that's already suitable to apply a chat template to
* A small, but good SFT model that has a chat template

The current script runs in about ~2' on a single node. The current set of hparams show margins going up / loss going down which is sufficient for a demo.

Screenshot 2024-03-25 at 16 25 08

### DIFF
--- a/examples/scripts/kto.py
+++ b/examples/scripts/kto.py
@@ -13,49 +13,45 @@
 # limitations under the License.
 
 """
-Run the KTO training script with the following command with some example arguments.
-In general, the optimal configuration for KTO will be similar to that of DPO:
+Run the KTO training script with the commands below. In general, the optimal configuration for KTO will be similar to that of DPO.
 
-# regular:
+# Full training:
 python examples/scripts/kto.py \
-    --model_name_or_path=gpt2 \
+    --model_name_or_path=stabilityai/stablelm-2-zephyr-1_6b \
     --per_device_train_batch_size 16 \
-    --max_steps 1000 \
+    --num_train_epochs 1 \
     --learning_rate 2e-5 \
     --gradient_accumulation_steps 1 \
     --logging_steps 10 \
     --eval_steps 500 \
-    --output_dir="kto_anthropic_hh" \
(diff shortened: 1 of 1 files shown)


kashif (approved):

examples/scripts/kto.py
```diff
@@ -68,85 +64,44 @@ class ScriptArguments:
...
         tokenizer.pad_token = tokenizer.eos_token
 
-    # 2. Load the Anthropic Helpful-Harmless dataset
-    train_dataset = get_hh("train", sanity_check=script_args.sanity_check)
+    # Load the dataset
+    dataset = load_dataset(script_args.dataset_name)
+
+    # Apply chat template
+    def format_dataset(example):
+        example["prompt"] = tokenizer.apply_chat_template(example["prompt"], tokenize=False)
```
lewtun, line 87: This may fail if a user uses a tokenizer that doesn't have a chat template - shall I add a check for this?

examples/scripts/kto.py
```diff
@@ -68,85 +64,44 @@ class ScriptArguments:
...
+        eval_dataset=formatted_dataset["test"],
         tokenizer=tokenizer,
         peft_config=get_peft_config(model_args),
     )
 
-    # 5. train and save the model
+    # Train and push the model to the Hub
     kto_trainer.train()
     kto_trainer.save_model(kto_args.output_dir)
+    kto_trainer.push_to_hub()
```
lewtun, line 107: I think we should put `push_to_hub()` in all our example scripts to track usage on the Hub

examples/scripts/kto.py
kashif: right in that case one would just skip this?

lewtun: Added here https://github.com/huggingface/trl/pull/1481/commits/3c7944842a93ae681a156a93c995bf9e52b56923

PR #1488, opened 2024-03-27
hackey update to ModelConfig to allow lora_target_modules="all-linear"

galtay:
the type hint forces a list which raises a "all-linear" layer not found. forcing a string makes it work. updating the type hint to `Union[str, list[str]]` also raise a parsing error

looks like peft handles this by making the arg a str and using comma delimited values 
https://github.com/huggingface/peft/blob/main/examples/sft/train.py#L30


lewtun (approved):
Thanks for the fix - I was about to open a PR with the same hack :)

LGTM

PR #1491, opened 2024-03-28
[ORPO] use log1p for loss

kashif:
torch.log1p is more accurate for smaller values


lewtun (approved):
LGTM! Does this also prevent NaNs appearing in the log odds?

alvarobartt (approved):

PR #1499, opened 2024-04-01
[KTO] fix interleaving, reporting, hanging bugs

kawine:
- remove interleaving in KTOTrainer since it is no longer needed for KL estimation and artificially duplicates data to force an even number of positive and negative examples (see https://github.com/huggingface/trl/issues/1467)
- do not report statistics for good/bad data if there are zero good/bad examples in the current batch (currently NaNs are returned, which are sometimes converted to floats before being sent to wandb, leading to reward spikes (see https://twitter.com/_lewtun/status/1771493512262942801)
- make same changes as in #1472 to prevent hanging but still aggregate metrics from across batch (instead of only reporting from main process)
- fix arithmetic errors in metric aggregation introduced in #1476 when trying to patch the NaNs issue

cc @kashif @lewtun


kashif:
@kawine are you disabling the WAND on your side because of some local issue?

kawine:
just removed the disabling of wandb

kawine:
@kashif i reverted the changes to the tests -- are we good to merge?

kashif (approved):

PhilipMay:
Hi @kawine and @kashif thanks for this PR and for merging it.

Nevertheless, I find it a bit unusual to take the fixes from an older PR from other contributors and then put them into a much larger PR instead of actively helping out in the original PR. 🙁

See #1472

kashif:
@PhilipMay appologies! yes it's my bad I believe... let me see how to attribute Clara properly.

claralp:
@kashif I am mostly fine with the version implemented here now.  
However, one minor issue still exists with this version:  
When logging metrics after multiple iterations e.g. `mean(mean(rewards/chosen in iteration1), mean(rewards/chosen in iteration2))` is not the overall mean, because the total number of chosen samples differs.  
@kawine fixed this for different number of samples in one batch across multiple devices, but the same applies when averaging across multiple iterations

kawine:
@PhilipMay apologies for the confusion! i had an earlier PR #1476 that had to be closed bc it conflicted with the proposed changes in #1472, but #1472 had some separate issues with logging, which is why i created this PR (#1499). users were actively blocked because the code was hanging, so fixing that asap was high priority

clara's contributions are very much appreciated and i'm happy to defer to you folks on crediting everyone properly

PR #1509, opened 2024-04-06
fix add_special_tokens issue for data with template

### LINKED ISSUE #1400: conversational data for SFTTrainer
For SFTTrainer, if we load the dataset using a conversational form (ChatML format), the function `apply_chat_template` is used (https://github.com/huggingface/trl/blob/v0.7.11/trl/extras/dataset_formatting.py#L55) with `tokenize=False`. Later in SFTTrainer, the data is tokenized again with `add_special_tokens=True`. In tokenizer like LLaMATokenizer, there will be two bos tokens at the very beginning: `<s><s> ...`, which is not intended. Maybe we should modify `dataset_kwargs` at this line https://github.com/huggingface/trl/blob/v0.7.11/trl/trainer/sft_trainer.py#L246 so that `dataset_kwargs['add_special_tokens']=True`?

edixiong:
Fix the add_special_tokens issue as is mentioned in this issue: https://github.com/huggingface/trl/issues/1400

Fixes #1400

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -264,6 +264,12 @@ class SFTTrainer(Trainer):
             # check if dataset has ChatML format or instruction format and is supported
             # if not stays #None
             formatting_func = get_formatting_func_from_dataset(train_dataset, tokenizer)
+            # if a template is detected, we don't need to add special tokens again
+            if formatting_func is not None:
+                if dataset_kwargs is None:
+                    dataset_kwargs = {"add_special_tokens": False}
+                else:
+                    dataset_kwargs["add_special_tokens"] = False
 
         if not packing:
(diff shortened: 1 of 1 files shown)


younesbelkada (commented):
Thanks for the fix ! Can you slightly elaborate on why this should be the right fix? 🙏 
cc @philschmid FYI

edixiong:
Hi @younesbelkada , in `line 267`, we check if there is a template. If true, `dataset_formatting.py` will call `apply_chat_template` and the special token will be added there. In `sft_trainer.py`, the input will be tokenized and the special token will be added again if we do not explicitly set that `add_special_tokens` to False. Another workaround is to warn the user to manually set `add_special_tokens` to False (similar to @philschmid did).

younesbelkada (approved):
Thanks !

PR #1514, opened 2024-04-08
[KTO] fix metric logging

claralp:
When logging metrics after multiple iterations e.g. `mean(mean(rewards/chosen in iteration1), mean(rewards/chosen in iteration2))` is not the overall mean, because the total number of chosen samples often differs in iteration a and 2.  
Relates to #1499


kashif:
thanks @claralp can you kindly run the per-commit to format?

kashif (approved):

PR #1516, opened 2024-04-09
[ORPO] Update NLL loss to use `input_ids` instead

alvarobartt:
## Description

This PR updates how the NLL loss in ORPO is being calculated so as to use the `input_ids` for the chosen response i.e. the whole conversation, instead of only the labels i.e. the ids for the chosen response only.

As mentioned by @nlee-208, calculating the full prompt loss instead of solely the loss on the chosen response leads to better results in MT-Bench, leading then to a more faithful reproduction of the original codebase at https://github.com/xfactlab/orpo.git (see the screenshot shared by Noah below where the `wo-pl`, standing for without prompt loss, fine-tune performs worst on MT-Bench)

image

Kudos to @nlee-208, @jiwooya1000 and @lewtun

### DIFF
--- a/trl/trainer/orpo_trainer.py
+++ b/trl/trainer/orpo_trainer.py
@@ -705,7 +705,11 @@ class ORPOTrainer(Trainer):
             loss = loss_fct(logits, labels)
             return loss
 
-        labels = concatenated_batch["concatenated_input_ids"].clone()
+        if self.is_encoder_decoder:
+            labels = concatenated_batch["concatenated_labels"].clone()
+        else:
+            labels = concatenated_batch["concatenated_input_ids"].clone()
+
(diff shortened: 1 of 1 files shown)


kashif (approved):

kashif:
@alvarobartt so the failing tests are with regards to the enc-decoder models I believe

alvarobartt:
> @alvarobartt so the failing tests are with regards to the enc-decoder models I believe

Sure, let me check those!

lewtun (approved):
Thanks for the fix! LGTM but let's wait until we've figure out why the CI is red on `main` before merging this

kashif (approved):

lewtun:
Chatted with @younesbelkada about the failing slow tests (seems like a timeout issue), so it's fine to merge this

PR #1518, opened 2024-04-10
Adds VLM Training support to SFTTrainer + VSFT script

edbeeching:
- Modifies the SFTTrainer so that it can be used with a VLM SFT dataset
- Adds an example script `vsft.py` to train the llava1.5 model with an instruct dataset

TODO:
- [x] Test PEFT support
- [x] Run full training
- [x] Add example to docs
- [x] add tests

Example usage:
```
python examples/scripts/vsft.py \
    --model_name_or_path="llava-hf/llava-1.5-7b-hf" \
    --report_to="wandb" \
    --learning_rate=1.4e-5 \
    --per_device_train_batch_size=8 \
    --logging_steps=5 \
    --num_train_epochs=1 \
    --push_to_hub \
    --gradient_checkpointing \
    --torch_dtype=float16 \
    --fp16=True \ 
    --dataset_name=HuggingFaceH4/llava-instruct-mix-vsft \
```


examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+# regular:
+python examples/scripts/vsft.py \
+    --model_name_or_path="llava-hf/llava-1.5-7b-hf" \
+    --report_to="wandb" \
+    --learning_rate=1.41e-5 \
```
lewtun, line 20: Maybe pick something a bit less magic?
```suggestion
    --learning_rate=1.00e-5 \
```

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+
+tqdm.pandas()
+
+if TRL_USE_RICH:
+    logging.basicConfig(format=FORMAT, datefmt="[%X]", handlers=[RichHandler()], level=logging.INFO)
+
+
+if __name__ == "__main__":
+    parser = TrlParser((SftScriptArguments, TrainingArguments, ModelConfig))
+    args, training_args, model_config = parser.parse_args_and_config()
```
lewtun, line 93: For gradient checkpointing I think we need to manually add the `use_reentrant=False` flag like we do here: https://github.com/huggingface/trl/blob/57aebe9c36021e679662181468b04ff42432daf3/examples/scripts/reward_modeling.py#L46C5-L46C76

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+    # Force use our print callback
+    if TRL_USE_RICH:
+        training_args.disable_tqdm = True
+        console = Console()
+
+    ################
+    # Model, Tokenizer & Processor
+    ################
+
+    LLAVA_CHAT_TEMPLATE = """{% for message in messages %}{% if message['role'] == 'user' %}USER:{% else %}ASSISTANT:{% endif %}{% for item in message['content'] %}{% if item['type'] == 'text' %}{{ item['text'] }}{% elif item['type'] == 'image' %}<image>{% endif %}{% endfor %}{{'\n'}}{% endfor %}"""
```
lewtun, line 104: Should this be configurable from the command line instead of hardcoded? I think it's also not quite correct since we're missing the system prompt which is quite important for evals. The full template is something like the following, adjusted for your new data schema:

```
"{% for message in messages %}{% if loop.index0 == 0 %}A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful, detailed, and polite answers to the user's questions. USER: {{ message['content'] }} {% elif message['role'] == 'user' %}USER: {{ message['content'] }} {% else %} ASSISTANT: {{ message['content'] }}{{ eos_token }}{% endif %}{% endfor %}{% if add_generation_prompt %}{{ 'ASSISTANT:' }}{% endif %}"
```

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+        if model_config.torch_dtype in ["auto", None]
+        else getattr(torch, model_config.torch_dtype)
+    )
+    quantization_config = get_quantization_config(model_config)
+    model_kwargs = dict(
+        revision=model_config.model_revision,
+        trust_remote_code=model_config.trust_remote_code,
+        attn_implementation=model_config.attn_implementation,
+        torch_dtype=torch_dtype,
+        # use_cache=False if training_args.gradient_checkpointing else True, # llava does not have this option
```
lewtun, line 117: Do you mean llava doesn't have a cache support or that gradient checkpointing isn't supported? If the latter, then we should remove it from the command line args

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+        device_map=get_kbit_device_map() if quantization_config is not None else None,
+        quantization_config=quantization_config,
+    )
+    tokenizer = AutoTokenizer.from_pretrained(model_config.model_name_or_path, use_fast=True)
+    # tokenizer.pad_token = tokenizer.eos_token
```
lewtun, line 122: Delete?

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+    processor = AutoProcessor.from_pretrained(model_config.model_name_or_path)
+    processor.tokenizer = tokenizer
+
+    model = LlavaForConditionalGeneration.from_pretrained(model_config.model_name_or_path, **model_kwargs)
+
+    ################
+    # Create a data collator to encode text and image pairs
+    ################
+
+    class LLavaDataCollator:
```
lewtun, line 133: Should this live in a `collators.py` module? That way we can properly unit test it (which would be good in any case)

trl/trainer/sft_trainer.py
```diff
@@ -396,12 +396,15 @@ def _prepare_dataset(
...
         append_concat_token=True,
         add_special_tokens=True,
+        skip_prepare_dataset=False,
     ):
         if dataset is None:
             raise ValueError("The dataset should not be None")
 
         # check if torch dataset / dataloader and do nothing
-        if isinstance(dataset, (torch.utils.data.IterableDataset, torch.utils.data.Dataset, ConstantLengthDataset)):
+        if skip_prepare_dataset or isinstance(
```
lewtun, line 405: Let's add an integration test for this case under the existing `SFTTrainer` tests, along with an example training a tiny random llava model

lewtun (commented):
Very clean implementation @edbeeching ! Just a few nits and a few unit/integration tests needed and this should be good to merge

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+    --remove_unused_columns=False \
+    --torch_dtype=float16 \
+    --fp16=True \ 
+    --dataset_name=HuggingFaceH4/llava-instruct-mix-vsft \
+    
+# peft:
+python examples/scripts/vsft.py \
+    --model_name_or_path="llava-hf/llava-1.5-7b-hf" \
+    --report_to="wandb" \
+    --learning_rate=1.41e-5 \
```
lewtun, line 38:
```suggestion
    --learning_rate=1.00e-4 \
```

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+    --per_device_train_batch_size=8 \
+    --gradient_accumulation_steps=1 \
+    --output_dir="data/vsft-llava-1.5-7b-hf" \
+    --logging_steps=1 \
+    --num_train_epochs=1 \
+    --max_steps=-1 \
```
lewtun, line 44:
```suggestion
```

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
```
lewtun, line 26:
```suggestion
```

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+"""
+# regular:
+python examples/scripts/vsft.py \
+    --model_name_or_path="llava-hf/llava-1.5-7b-hf" \
+    --report_to="wandb" \
+    --learning_rate=1.41e-5 \
+    --per_device_train_batch_size=8 \
+    --gradient_accumulation_steps=1 \
+    --output_dir="data/vsft-llava-1.5-7b-hf" \
+    --logging_steps=1 \
```
lewtun, line 24: Are you sure we want to log every step? Maybe 
```suggestion
    --logging_steps=5 \
```

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+    
+# peft:
+python examples/scripts/vsft.py \
+    --model_name_or_path="llava-hf/llava-1.5-7b-hf" \
+    --report_to="wandb" \
+    --learning_rate=1.41e-5 \
+    --per_device_train_batch_size=8 \
+    --gradient_accumulation_steps=1 \
+    --output_dir="data/vsft-llava-1.5-7b-hf" \
+    --logging_steps=1 \
```
lewtun, line 42:
```suggestion
    --logging_steps=5 \
```

examples/scripts/vsft.py
```diff
@@ -0,0 +1,198 @@
...
+            peft_config=get_peft_config(model_config),
+            callbacks=[RichProgressCallback] if TRL_USE_RICH else None,
+            data_collator=data_collator,
+            dataset_kwargs={"skip_prepare_dataset": True},
+        )
+
+    trainer.train()
+
+    with save_context:
+        trainer.save_model(training_args.output_dir)
```
lewtun, line 198: After this I think we might need a final `trainer.push_to_hub()`

examples/scripts/vsft.py
pcuenca: Could this (or parts of it) be potentially derived from the tokenizer, if a chat template is defined for the model?

pcuenca: Also, if this is Llava-specific, I'd call the script `vsft_llava.py` (or rename this variable to `CHAT_TEMPLATE`).

examples/scripts/vsft.py
pcuenca: For an initial example I'd keep it simple and self-contained. I think it's valuable to understand everything and easy to tweak when necessary.

examples/scripts/vsft.py
lewtun: > Could this (or parts of it) be potentially derived from the tokenizer, if a chat template is defined for the model?

Indeed, ideally we should use the tokenizer/processor chat template if it exists and then fallback to the default llava chat template if needed.

examples/scripts/vsft.py
edbeeching: Will do, the value is copied from the SFT example: https://github.com/huggingface/trl/blob/57aebe9c36021e679662181468b04ff42432daf3/examples/scripts/sft.py#L20

examples/scripts/vsft.py
edbeeching: It has this option is the base LLM, but the LlavaForConditionalGeneration model does not expose it.

examples/scripts/vsft.py
edbeeching: Thanks, I added the system prompt. 

There are a lot of llava specific things in this script, the model itself, the chat template, etc. I am wondering if it makes sense to make it more general now, or to wait until we can support more VLM models?

pcuenca: In my opinion, I'd keep it llava specific for the release so it's as easy to understand as possible. We could invite the community to open PRs for other examples and then decide whether to write a more general version.

examples/scripts/vsft_llava.py
```diff
@@ -0,0 +1,194 @@
...
+    --push_to_hub \
+    --gradient_checkpointing \
+    --remove_unused_columns=False \
+    --torch_dtype=float16 \
+    --fp16=True \ 
+    --dataset_name=HuggingFaceH4/llava-instruct-mix-vsft \    
+    --use_peft \
+    --lora_r=64 \
+    --lora_alpha=16
+"""
```
lewtun, line 52:
```suggestion

# to evaluate, first install the lmms-eval framework: pip install git+https://github.com/EvolvingLMMs-Lab/lmms-eval.git
# then run:
accelerate launch --num_processes=8 -m lmms_eval \
        --model llava_hf \
        --model_args pretrained=llava-hf/llava-1.5-7b-hf \
        --tasks mmbench \
        --batch_size 1 \
        --output_path ./logs/ \
        --log_sample
"""
```

younesbelkada (approved):
Thanks ! We should be good to go after fixing the merge conflict on main ! 🚀

PR #1520, opened 2024-04-11
Allow pre-tokenized datasets in SFTTrainer

BramVanroy:
This tiny PR makes it so that datasets that are already tokenized (have an `input_ids` column) can be used as-is. This greatly improves flexibility, especially in larger training runs: we can do data preprocessing (tokenization) one one CPU-heavy server and save the tokenized dataset to disk, and transfer it to a GPU server to use it directly as a dataset.

Note: this only works for `datasets` datasets, not for PyTorch datasets.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -401,7 +401,16 @@ class SFTTrainer(Trainer):
         if dataset is None:
             raise ValueError("The dataset should not be None")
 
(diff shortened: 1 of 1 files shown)


BramVanroy:
@younesbelkada Sure! https://github.com/huggingface/trl/pull/1521

younesbelkada:
Thanks @BramVanroy !

PR #1531, opened 2024-04-12
[CPO] fix memory leak due to retained value

### LINKED ISSUE #1526: a bug which leads to "Cuda: Out of memory" in CPOTrainer (cpo_trainer.py), trl 0.8.1 0.8.2, please fix this bug
there is a bug in CPOTrainer. when runing CPOTrainer after runing sevreal steps, the  usage of gpu memory increases and it raises the out-of-memory exception. we found that the exception is caused by missing the "detach" in line 741 of CPOTrainer (cpo_trainer.py), please fix this bug, be careful!

line 741 of CPOTrainer (cpo_trainer.py):  
metrics[f"{prefix}nll_loss"] = policy_nll_loss.cpu().mean()
->
metrics[f"{prefix}nll_loss"] = policy_nll_loss**.detach()**.cpu().mean()

kashif:
fixes #1526


lewtun (approved):
Thanks for the fix! LGTM

PR #1532, opened 2024-04-12
VSFT hotfix - adds gen prompt to template and processor to hub

edbeeching:
Fixes two things that were missing from the vsft script:

- Adds generation prompt to the Chat template
- Adds processor push to hub at end of script

### DIFF
--- a/examples/scripts/vsft_llava.py
+++ b/examples/scripts/vsft_llava.py
@@ -66,6 +66,7 @@ accelerate launch --num_processes=8 -m lmms_eval \
 import logging
 import os
 from contextlib import nullcontext
+from pathlib import Path
 
 TRL_USE_RICH = os.environ.get("TRL_USE_RICH", False)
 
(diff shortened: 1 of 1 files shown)


examples/scripts/vsft_llava.py
```diff
@@ -205,3 +207,5 @@ def __call__(self, examples):
     with save_context:
         trainer.save_model(training_args.output_dir)
         trainer.push_to_hub()
+        if Accelerator().is_main_process:
+            processor.push_to_hub(Path(training_args.output_dir).name)
```
lewtun, line 211: Do we need this `Path` stuff or could we instead do:

```suggestion
            processor.push_to_hub(training_args.hub_model_id)
```

lewtun (approved):
LGTM barring a suggestion to make the push to Hub cleaner

examples/scripts/vsft_llava.py
edbeeching: thanks, cleaner

PR #1534, opened 2024-04-12
Fixed ref model not used in PPO generation

ejmejm:
In the PPOTrainer.generate function, the primary model was being used instead of the reference model to generate reference responses for single samples. This change uses the correct  model and introduces a test to check for this. The new test was failing with the previous codebase.


vwxyzjn:
Nice catch thank you! I am gonna run the tests and merge if it passes.

vwxyzjn (approved):
Thanks for the PR!

Issue #848, opened 2023-10-09
PPO Performance improvement by reducing the number of model calls


freQuensy23-coder:
I wanted to highlight a potential optimization in the PPO trainer to improve training speed. Currently, the forward pass through the model is executed three times with the same inputs during each training step:

 - When generating the response using ppo_trainer.generate() (with the LM head and torch no grad)
 - When evaluating the value head as the critic in ppo_trainer.step() (with value head and torch.no_grad())
 - When calculating the policy gradients without torch.no_grad() to update parameters like LoRA adapters - again using the (with LM head and grad)(https://github.com/huggingface/trl/blob/eda1f36c57e95399bdbfe0a74e1bb0f9af2c9508/trl/trainer/ppo_trainer.py#L760)
Since the model is called three times on the same inputs, the training speed could potentially be improved by optimizing these repeated forward passes. One approach could be to cache the embeddings or model outputs computed in the first pass, and reuse those later in the training step.

I believe optimizing these repeated computations could significantly increase overall training throughput. Please let me know if you have any other suggestions to improve this workflow. I'm happy to open a GitHub issue as well for further discussion.

Thank you for your consideration!
 
So, i think that these model calls can be optimized to increase the speed of work ~ 2 times?

## Code example
```python
for epoch, batch in tqdm(enumerate(ppo_trainer.dataloader)):
    query_tensors = batch["input_ids"]
    # Get response from the policy model
    response = ppo_trainer.generate(query_tensors[0], **generation_kwargs)
        response_tensors.append(response.squeeze()[-gen_len:])
    batch["response"] = [tokenizer.decode(r.squeeze()) for r in response_tensors]

    # Compute sentiment score # noqa
    texts = batch["response"]
    toxicity_inputs = toxicity_tokenizer(texts, padding=True, truncation=True, return_tensors="pt").to(
        ppo_trainer.accelerator.device
    )
    logits = toxicity_model(**toxicity_inputs).logits.float()
    toxicity_labels = (logits[:, 0]).tolist()

    rewards = [torch.tensor(output) for output in toxicity_labels]

    # Run PPO step
    stats = ppo_trainer.step(query_tensors, response_tensors, rewards)
    ppo_trainer.log_stats(stats, batch, rewards)
```

lvwerra:
[...]

PR #1535, opened 2024-04-13
Ability to store generation logits and vals for training

ejmejm:
When using the PPO trainer, a user will generally generate responses with `PPOTrainer.generate()`. The input queries and resulting responses are then passed to `PPOTrainer.step()` to train. At the start of training, the initial logits and values of the these input sequences are again calculated in `PPOTrainer.step()` (line 1129 in this branch). The majority of this computation is a repeat of what was computed during the call to `PPOTrainer.generate()`, which leaves an opportunity to save the computation cost of one forward pass per training batch. This optimization was previous requested in ticket #848.

This change adds the ability to return the values and logits during generation, so that they can be fed back to `PPOTrainer.step()`, and reused to save on computation. Example usage:

```
response_tensors, values_and_logits = ppo_trainer.generate(
    query_tensors,
    return_prompt=False,
    return_values_and_logits=True,
    **generation_kwargs,
)

ppo_trainer.step(query_tensors, response_tensors, reward, values_and_logits=values_and_logits)
```

This is an easy addition to anyone who wants to save compute. The `return_values_and_logits` and `values_and_logits` arguments of these functions are optional, so using these functions without change is also not an issue.

### DIFF
--- a/tests/test_ppo_trainer.py
+++ b/tests/test_ppo_trainer.py
@@ -19,6 +19,7 @@ import tempfile
 import unittest
 from functools import partial
 
+import numpy as np
 import pytest
 import torch
 from huggingface_hub import HfApi, HfFolder, delete_repo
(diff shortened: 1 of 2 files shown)


vwxyzjn:
@ejmejm I believe https://github.com/vwxyzjn/trl/blob/61e39010bd660fbddb9cff6a1f50d347ae375f9e/trl/trainer/ppov2_bandit_rloo_trainer.py#L461-L481 does what you are thinking? Ah I also realize the latest transformer just supported the `output_logits` which previously I needed to do `output_score`.

PR #1536, opened 2024-04-13
Adds reward bootstrapping to PPOTrainer

ejmejm:
The current return calculations for the PPO trainer assumes that every response terminates without truncation (https://github.com/huggingface/trl/blob/main/trl/trainer/ppo_trainer.py#L1159).

```
nextvalues = values[:, t + 1] if t < gen_len - 1 else 0.0
```

When an episode is truncated early, it is technically correct to either drop the episode, or more commonly, bootstrap the reward with the value of the final state (e.g. `last reward <- last reward + gamma * value of final state`). This change adds a `bootstrap_rewards` option to `PPOConfig` that enables bootstrapping when sequences do not end with an EOS token. An alternative way to implement this is to have the user pass in a list of which responses were terminated early, but I felt the option that was simpler for the user would be better.


vwxyzjn:
Related #1540

Issue #1063, opened 2023-12-05
Support PPO with value model, not value head


natolambert:
There's a lot of effort brewing to do thorough comparisons of DPO to PPO and other methods, but to the best of my knowledge many of the best PPO successes in public (and mostly private) use entire value models instead of just value heads. 
Any chance we can have help introducing this? Anyone? 🤗

natolambert:
I think it'll be an additional option / parameter (non-default) then everything needs to be validated again. It's tough, but worth it?

The value network is interesting. Really odd, not sure why it's so bad. 
Maybe the normalization is more important with a value model rather than a value head, given it has less regularization by "More different" parameters?

natolambert:
Related, @vwxyzjn shared this photo from the appendix of the summarize work with me 
![image](https://github.com/huggingface/trl/assets/10695622/40777242-9b8f-443e-8778-31258731bc5f)

vwxyzjn:
Adding a note here:

To use separate value head for PPO, we might not be able to use AutoModelForSequenceClassification without refactoring transformers.

The reason is that AutoModelForSequenceClassification only returns the pooled_logits corresponding to EOS tokens ([modeling_gpt_neox.py#L1214](https://github.com/huggingface/transformers/blob/45b1dfa342f2ec00634c31f635bded6e346c98d2/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L1214)), there isn't any way to get the actual logits corresponding to each token, but for separate value network we need to model the value at each token.

vwxyzjn:
A current workaround is something like 


```

class ScalarModelConfig(PretrainedConfig):
    def __init__(
        self,
        base_model: str = "EleutherAI/pythia-160m",
        base_config: PretrainedConfig = AutoConfig.from_pretrained("EleutherAI/pythia-160m"),
        hidden_size: int = 768,
        bias: float = 0.0,
        **kwargs,
    ):
        super().__init__(**kwargs)
        self.base_model = base_model
        self.base_config = base_config
        self.hidden_size = hidden_size
        self.bias = bias


class ScalarModel(PreTrainedModel):
    config_class = ScalarModelConfig

    def __init__(self, config: ScalarModelConfig):
        super().__init__(config)
        self.config = config
        self.lm_backbone = AutoModel.from_pretrained(
            config.base_model,
            config=self.config.base_config,
            trust_remote_code=True,
        )
        self.scalar_head = layer_init(
[...]

PR #1540, opened 2024-04-15
PPO / Reinforce Trainers

vwxyzjn:
This RP supports the REINFORCE RLOO trainers in https://arxiv.org/pdf/2402.14740.pdf. 

Note that [REINFORCE's loss is a special case of PPO](https://arxiv.org/pdf/2205.09123.pdf), as shown below

image

it matches the REINFORCE loss presented in the Cohere paper (where PPO uses advantages A hat, but REINFORCE uses the RLHF reward R(y, x))
image

We add the following files
* trl/trainer/ppov2_trainer.py
    * a PPO implementation used in our [learning to summarize reproduction work](https://arxiv.org/abs/2403.17031)
* trl/trainer/ppov2_bandit_rloo_trainer.py
    * a PPO variant which implements 1) modeling completion as a joint action and 2) RLOO loss, which does not use a value network
    * I copied this file directly from `ppov2_trainer.py`, so feel free to do a file diff to see the changes(e.g., the following diff shows how the RLOO loss is implemented)
image

Two more examples showing how they work with dummy reward models
* examples/scripts/minimal/ppo.py: [preliminary experiment](https://wandb.ai/costa-huang/huggingface/runs/vm67rxzl?nw=nwusercostahuang) shows RLHF reward goes up, so from an optimization standpoint it works as intended
image

* examples/scripts/minimal/ppo_bandit_rloo.py:[ preliminary experiment](https://wandb.ai/costa-huang/huggingface/runs/astk72ix?nw=nwusercostahuang) shows RLHF reward goes up, so from an optimization standpoint it works as intended; though the KL kind of exploded, so we may need to use a larger beta for stronger regularization.
image


trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+                "try increase the `mini_batch_size` or `gradient_accumulation_steps`"
+            )
+        # note that if mask_sum == 1, then there is a division by zero issue
+        # to avoid it you just need to use a larger minibatch_size
+        bessel_correction = mask_sum / (mask_sum - 1)
+        variance = variance * bessel_correction
+    return variance
+
+
+def masked_whiten(values, mask, shift_mean=True):
```
lapp0, line 163: this doesn't appear to be called anywhere.

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+    """Whiten values with masked values."""
+    mean, var = masked_mean(values, mask), masked_var(values, mask, False)
+    whitened = (values - mean) * torch.rsqrt(var + 1e-8)
+    if not shift_mean:
+        whitened += mean
+    return whitened
+
+
+def get_reward(model, query_responses, tokenizer, context_length):
```
lapp0, line 172: Should we have `get_reward` as a class argument so users can subclass and create custom reward functions easily?

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+        # position_ids=position_ids,
+        return_dict=True,
+        output_hidden_states=True,
+    )
+    reward_logits = model.score(output.hidden_states[-1])
+    sequence_lengths = (
+        first_true_indices(query_responses[:, context_length:] == tokenizer.pad_token_id) - 1 + context_length
+    )
+    # https://github.com/huggingface/transformers/blob/dc68a39c8111217683bf49a4912d0c9018bab33d/src/transformers/models/gpt2/modeling_gpt2.py#L1454
+    return (
```
lapp0, line 189: Could we have a doc string detailing what the return tuple is comprised of?

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+        )
+        policy.generation_config.pad_token_id = None  # generate tokens without truncation / padding
+
+        #########
+        # setup model, optimizer, and others
+        #########
+        if args.truncate_token and args.truncate_token == "eos":
+            args.truncate_token_id = tokenizer.eos_token_id
+        self.model = policy
+        self.create_optimizer_and_scheduler(num_training_steps=args.num_updates)
```
lapp0, line 382: Should `num_training_steps` be a function of `self.train_dataset` and the number of epochs and batch size, so it's automatic and the user doesn't need to calculate this?

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+
+        #########
+        ### trainer specifics
+        #########
+        self.state = TrainerState(
+            is_local_process_zero=self.is_local_process_zero(),
+            is_world_process_zero=self.is_world_process_zero(),
+        )
+        DEFAULT_CALLBACKS = [DefaultFlowCallback]
+        default_callbacks = DEFAULT_CALLBACKS + get_reporting_integration_callbacks(self.args.report_to)
```
lapp0, line 392: Can we move most of this to just being a `super().__init__()` call? This will prevent the need for maintenance which is already handled by the base `Trainer` class.

examples/scripts/minimal/ppo_bandit_rloo_large.py
```diff
@@ -0,0 +1,177 @@
...
+    examples/scripts/minimal/ppo_bandit_rloo_large.py \
+    --noptepochs 1 \
+    --nminibatches 1 \
+    --rloo_k 4 \
+    --learning_rate 3e-6 \
+    --output_dir models/minimal/ppo \
+    --per_device_train_batch_size 1 \
+    --gradient_accumulation_steps 16 \
+    --total_episodes 10000 \
+    --base_model mistralai/Mistral-7B-v0.1 \
```
lapp0, line 40: We probably should use a dataclass subclassing from https://github.com/huggingface/trl/blob/08655727487f8436980d4b538637832d1dcfa2e0/trl/trainer/model_config.py#L13

lapp0 (changes requested):
Thanks so much for this contribution. I've been hoping to experiment REINFORCE w/ transformers for a while now, but didn't have the time to roll my own implementation.

This is a great foundation in terms of functionality. I'll be playing around with it soon.

I think we should reduce the repetition, and use inheritance of existing classes so we can take advantage of the great infrastructure built out by huggingface/transformers and huggingface/trl.

Happy to help if you're interested in collaborating, let me know.

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+        dataloader = DataLoader(
+            self.train_dataset,
+            batch_size=args.local_batch_size,
+            shuffle=True,
+            collate_fn=DataCollatorWithPadding(tokenizer),
+            drop_last=True,  # needed; otherwise the last batch will be of ragged shape
+        )
+
+        device = accelerator.device
+        torch.backends.cudnn.deterministic = True # TODO: check if cudnn matters
```
lewtun, line 435: FYI this can cause dramatic slowdown so should be avoided except for debugging: https://pytorch.org/docs/stable/notes/randomness.html

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+        torch.manual_seed(args.seed)
+        model, optimizer, dataloader = accelerator.prepare(model, optimizer, dataloader)
+        torch.manual_seed(local_seed)  # reset the local seed again
+
+        def repeat_generator():
+            while True:
+                yield from dataloader
+
+        iter_dataloader = iter(repeat_generator())
+        if args.deepspeed2:
```
lewtun, line 448: You should be able to infer this from the `accelerator` directly - see how we did it for DPO

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+    """Whether to offload ref policy and reward model to CPU"""
+    reward_model_path: str = "EleutherAI/pythia-160m"
+    """the path to the reward model"""
+    sft_model_path: str = "EleutherAI/pythia-160m"
+    """the path to the sft model"""
+
+    # ppo args
+    nminibatches: int = 1
+    """the number of minibatches to split a batch into"""
+    noptepochs: int = 4
```
lewtun, line 106: Maybe we can reuse `num_train_epochs` from the trainer? The current name is a bit cryptic IMO

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+                queries = queries.repeat(args.rloo_k, 1)
+                context_length = queries.shape[1]
+                query_responses = []
+                responses = []
+                postprocessed_responses = []
+                logprobs = []
+                ref_logprobs = []
+                scores = []
+                sequence_lengths = []
+                for i in range(0, queries.shape[0], args.local_rollout_forward_batch_size):
```
lewtun, line 494: Here you could use the `unwrap_model_for_generation()` context manager as well

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+                    # del everything and empty cache
+                    # fmt: off
+                    del (
+                        output, logits, new_all_logprobs, new_logprobs,
+                        logprobs_diff, ratio, pg_losses, pg_losses2,
+                        pg_loss, loss, pg_clipfrac, prob_dist, entropy, approxkl,
+                        mb_advantage, mb_responses, mb_query_responses, mb_logprobs,
+                    )
+                    # fmt: on
+                    torch.cuda.empty_cache()
```
lewtun, line 642: Does it also make sense to call `gc.collect()` here?

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+
+# taken from https://github.com/vwxyzjn/direct-preference-optimization/blob/f8b8c0f49dc92a430bae41585f9d467d3618fe2f/utils.py#L99
+def disable_dropout(model: torch.nn.Module):
+    """Disable dropout in a model."""
+    for module in model.modules():
+        if isinstance(module, torch.nn.Dropout):
+            module.p = 0
+
+
+def masked_mean(values, mask, axis=None):
```
lewtun, line 136: If we have functions copied across classes, we could implement a similar parser to what is done in `transformers` where you annotate with `# Copied from: ...` https://github.com/huggingface/transformers/blob/main/utils/check_copies.py

(Not for this PR, but would be valuable later on)

lewtun (commented):
Epic work on adding these new RL trainers @vwxyzjn ! I've left some high level feedback on the RLOO trainer for now and will do a more fine-grained review when we've iterated a bit on the design.

Overall looks super clean

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+                            pg_losses = -mb_advantage * ratio
+                            pg_losses2 = -mb_advantage * torch.clamp(ratio, 1.0 - args.cliprange, 1.0 + args.cliprange)
+                            pg_loss_max = torch.max(pg_losses, pg_losses2)
+                            pg_loss = pg_loss_max.mean()
+                            pg_clipfrac = (pg_losses2 > pg_losses).float().mean()
+                            loss = pg_loss
+                            accelerator.backward(loss)
+                            optimizer.step()
+                            optimizer.zero_grad()
+                            with torch.no_grad():
```
lapp0, line 619: why is `torch.no_grad` necessary here?

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+                                pg_clipfrac = pg_clipfrac
```
lapp0, line 620: unnecessary

vwxyzjn:
Hi @lapp0 thanks for the review! Will look into these comments more closely. I started running some experiments and noticed the KL of RLOO was orders of magnitude higher than that of the new PPO trainer. Not exactly sure the reason but will further investigate.

image

The PPO / Vanilla PG actually seems quite stable now with RLHF reward going up and model gets good scores and reasonable completions. There were some implementation details I found particularly helpful such as truncate at EOS token (i.e., `--truncate_token eos`), and I suspect the same technique could work nicely with RLOO. 

Right now I am a bit focused on a zephyr PPO /Vanilla PG recipe for these couple of days, and will look into RLOO right after.

image

image
image

lapp0 (changes requested):
I have a WIP fork. The main difference is that it lets `transformers.Trainer` set everything up:  batch creation, accelerate / deepspeed, etc. Instead it overrides `training_step`.

https://github.com/lapp0/trl/blob/onpolicy/trl/trainer/rloo_trainer.py

The main behavior difference is that it generates once per batch and runs for `num_train_epochs` rather than generating once per update and running for `num_train_epochs * num_updates`. Have you experimented with updating once per batch, and if so, does this harm stability? Is it important that I retain the ability to update once and run multiple epochs based on the generation with the original model?

It's possible that generating once per batch instead of per update would improve KL now that you mention it.

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+
+
+@dataclass
+class PPOConfig(TrainingArguments):
+    # common args
+    exp_name: str = os.path.basename(__file__)[: -len(".py")]
+    """the name of this experiment"""
+    run_name: Optional[str] = None
+    """a unique name of this run"""
+    deepspeed2: bool = False
```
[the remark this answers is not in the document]
vwxyzjn: Good point! Will explore how this can be done

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+    """Whether to use deepspeed to train the model"""
+    deepspeed3: bool = False
```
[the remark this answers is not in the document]
vwxyzjn: Good point! Will explore how this can be done.

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+    """Whether to use deepspeed to train the model"""
+
+    # various batch sizes
+    world_size: Optional[int] = None
```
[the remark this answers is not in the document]
vwxyzjn: Should be. I still think we can include it in the args though that way we know what the global batch size is.

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+    """The number of processes (GPUs) to use"""
+    num_train_epochs: int = 1
```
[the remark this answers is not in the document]
vwxyzjn: Nope. Removed.

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+    """The number of updates to train"""
+    total_episodes: Optional[int] = 1000000
+    """The total number of episodes in the dataset"""
+    micro_batch_size: Optional[int] = None
+    """The micro batch size across devices (HF's `per_device_train_batch_size` * `world_size`)"""
+    local_batch_size: Optional[int] = None
+    """The batch size per GPU (HF's `per_device_train_batch_size` * `gradient_accumulation_steps`)"""
+    batch_size: Optional[int] = None
+    """The batch size across devices (HF's `per_device_train_batch_size` * `world_size` * `gradient_accumulation_steps`)"""
+    nminibatches: int = 1
```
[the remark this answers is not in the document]
vwxyzjn: Done

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+    """the mini batch size per GPU"""
+    mini_batch_size: Optional[int] = None
+    """the mini batch size across GPUs"""
+    local_eval_batch_size: int = 2
+    """per rank eval batch size"""
+    local_rollout_forward_batch_size: int = 64
+    """per rank no grad forward pass in the rollout phase"""
+
+    # other args
+    base_model: str = "EleutherAI/pythia-160m"
```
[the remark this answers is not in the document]
vwxyzjn: Will look into this.

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+    """the truncate token"""
+    truncate_token_id: Optional[int] = None
+    """the truncation token id"""
+    temperature: float = 0.7
+    """the sampling temperature"""
+    penalty_reward_value: int = -1
+    """the reward value for responses that do not contain `truncate_token_id`"""
+    non_eos_penalty: bool = False
+    """whether to penalize responses that do not contain `truncate_token_id`"""
+    offload: bool = False
```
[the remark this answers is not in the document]
vwxyzjn: TODO:

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: renamed to num_ppo_epochs

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: TODO

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: not in rloo trainer. removed

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: That is a good point. I will think about the design. TODO.

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: TODO

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+        "prescale_gradients": False,
+        "wall_clock_breakdown": False,
+    }
+    model, *_ = deepspeed.initialize(model=model, config=eval_ds_config)
+    model.eval()
+    print("🔥 deepspeed2 is initialized")
+    return model
+
+
+def prepare_deepspeed3(model, accelerator):
```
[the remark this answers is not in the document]
vwxyzjn: TODO

trl/trainer/ppov2_bandit_rloo_trainer.py
```diff
@@ -0,0 +1,704 @@
...
+class PPOTrainer(Trainer):
+    def __init__(
+        self,
+        args: PPOConfig,
+        tokenizer: PreTrainedTokenizer,
+        policy: nn.Module,
+        ref_policy: nn.Module,
+        reward_model: nn.Module,
+        train_dataset: Dataset,
+        train_generation_config: GenerationConfig,
```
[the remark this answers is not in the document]
vwxyzjn: Currently I remove the `train_generation_config` and just let the users specify response length and temperature instead.

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: TODO: I will think about this.

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: `super().__init__()` is exactly what I tried, but I remember running into some arcane errors. Will try it out again. TODO.

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: removed

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: Will check it out. TODO.

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: Will check it out

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: That way `prob_dist = torch.nn.functional.softmax(logits, dim=-1)` is not incurring gradient.

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: Not sure if it's helpful because `torch.cuda.empty_cache()` should already do the job, but I will double check.

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,819 @@
...
+        ) # no need to shuffle eval dataset
+        self.eval_dataloader = accelerator.prepare(self.eval_dataloader)
+
+    def get_train_dataloader(self) -> DataLoader:
+        return self.dataloader
+    
+    def get_eval_dataloader(self) -> DataLoader:
+        return self.eval_dataloader
+
+    def push_to_hub(self, **kwargs):
```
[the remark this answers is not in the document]
vwxyzjn: I am not sure I follow. What does having a simple backup mean?

vwxyzjn (commented):
@lapp0 @lewtun thanks so much for the review! I put some comments down and TODO items.

vwxyzjn:
After some refactoring / bug fixes, the new RLOO also seems much more stable. Will report when having newer results.
image

trl/trainer/ppov2_trainer.py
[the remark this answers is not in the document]
lapp0: Sorry I wasn't clear. What I mean is that we should borrow the code below

```
        backup_model = self.model
        self.model = self.accelerator.unwrap_model(self.model).policy # save only the policy
        super().push_to_hub(**kwargs)
        self.model = backup_model
```

And do something similar for `save_model`, e.g.

```
    def save_model(self, **kwargs):
        backup_model = self.model
        self.model = self.accelerator.unwrap_model(self.model).policy # save only the policy
        super().save_model(**kwargs)
        self.model = backup_model
```

I might be misunderstanding the behavior but it seems because `save_model` calls `push_to_hub` and not the other way around, this is fine.

trl/trainer/ppov2_bandit_rloo_trainer.py
lapp0: I'm working on a changeset that could solve this

- base class which handles `get_reward` https://github.com/huggingface/trl/pull/1586/files#diff-334f2885f88f88edbf319820b62d7ccbecf6e428d6eb8b6313112895af3f75a6
- https://github.com/huggingface/trl/pull/1586/files#diff-85ec70f780f824b490e8cc4b4df9263fa61bf238e681f1cebd6bb18c294223de

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,819 @@
...
+        time_int = broadcast(time_tensor, 0).item()  # avoid different timestamps across processes
+        args.run_name = f"{args.exp_name}__{args.seed}__{time_int}"
+        self.local_seed = args.seed + accelerator.process_index * 100003  # Prime
+        if args.num_sample_generations > 0:
+            self.sample_generations_freq = max(1, args.num_updates // args.num_sample_generations)
+
+        #########
+        # disable dropout
+        #########
+        for module in [policy, ref_policy, value_model, reward_model]:
```
lapp0, line 394: Why are we disabling dropout for the `reward_model` and `ref_policy`?

For the trainable modules, `policy` and `value_model`, is it necessary to disable dropout? If not, I think we should leave disabling dropout to the user when they pass the modules.

vwxyzjn: For training policy, we need dropout absolutely. Otherwise, it causes issues for PPO / REINFORCE. E.g., 


image


```python
import torch
import transformers
torch.set_printoptions(precision=4, sci_mode=False)


def disable_dropout(model: torch.nn.Module) -> None:
    """Disable dropout in a model."""
    for module in model.modules():
        if isinstance(module, torch.nn.Dropout):
            module.p = 0
def forward(model, query_responses, tokenizer):
    attention_mask = query_responses != tokenizer.pad_token_id
    position_ids = attention_mask.cumsum(1) - attention_mask.long()
    input_ids = torch.masked_fill(query_responses, ~attention_mask, 0)
    return model(
        input_ids=input_ids,
        position_ids=position_ids,
        return_dict=True,
        output_hidden_states=True,
    )

def generate(lm_backbone, queries, tokenizer, generation_config):
    """generate in a way that does not affect padding tokens"""
    context_length = queries.shape[1]
    attention_mask = queries != tokenizer.pad_token_id
    input_ids = torch.masked_fill(queries, ~attention_mask, 0)
    output = lm_backbone.generate(
        input_ids=input_ids,
        # position_ids=attention_mask.cumsum(1) - attention_mask.long(), # already handled in generation
        generation_config=generation_config,
        return_dict_in_generate=True,
        output_scores=True
    )
    logits = torch.stack(output.scores, 1)
    return torch.cat((queries, output.sequences[:, context_length:]), dim=1), logits


def main(tokenizer, policy, query, temperature, context_length, generation_config):
    with torch.no_grad():
        query_response, logits = generate(policy, query, tokenizer, generation_config)
        response = query_response[:, context_length:]
        all_logprob = F.log_softmax(logits, dim=-1)
        generation_logprob = torch.gather(all_logprob, 2, response.unsqueeze(-1)).squeeze(-1)
        print(f"{response=}")
        print(f"{generation_logprob=}")

    output = forward(policy, query_response, tokenizer)
    logits = output.logits[:, context_length - 1 : -1]
    logits /= temperature
    all_logprob = F.log_softmax(logits, dim=-1)
    forward_logprob = torch.gather(all_logprob, 2, response.unsqueeze(-1)).squeeze(-1)
    print(f"{forward_logprob=}")

    ppo_ratio = (forward_logprob - generation_logprob).exp()
    print(f"{ppo_ratio=}")



tokenizer = transformers.AutoTokenizer.from_pretrained("gpt2", padding_side="right")
tokenizer.add_special_tokens({"pad_token": "[PAD]"})
pad_id = tokenizer.pad_token_id
policy = transformers.AutoModelForCausalLM.from_pretrained("gpt2")
policy.generation_config.pad_token_id = policy.generation_config.eos_token_id
policy.train()
query = torch.tensor([
    [pad_id, pad_id, 23073],
    [pad_id, pad_id, 234],
])
temperature = 0.7
context_length = query.shape[1]
generation_config = transformers.GenerationConfig(
    max_new_tokens=5,
    min_new_tokens=5,
    temperature=temperature,
    top_k=0.0,
    top_p=1.0,
    do_sample=True,
)
print("=======Before disabling dropout")
main(tokenizer, policy, query, temperature, context_length, generation_config)

disable_dropout(policy)
print("=======After disabling dropout")
main(tokenizer, policy, query, temperature, context_length, generation_config)
```

vwxyzjn:
So if I don't use `unwrap_model_for_generation`, things seem to work even with 1024 generation length, but speed is an issue. GPU utilization is also kind of poor...
image

lapp0:
> So if I don't use `unwrap_model_for_generation`, things seem to work even with 1024 generation length, but speed is an issue. GPU utilization is also kind of poor...

I've run into issues before with alternating between generation and training. They were resolved by using a simple context manager that ensured `model.eval()` was set. Perhaps it would help here?

```
class eval_mode:
    def __init__(self, model):
        self.model = model

    def __enter__(self):
        self.was_training = self.model.training
        if self.was_training:
            self.model.eval()

    def __exit__(self, exc_type, exc_val, exc_tb):
        if self.was_training:
            self.model.train()
```

vwxyzjn:
Btw I figured it out. I just need to do the following. Basically only calling `with unwrap_model_for_generation(model, self.accelerator)` once.

```
                with unwrap_model_for_generation(model, self.accelerator) as unwrapped_model:
                    for i in range(0, queries.shape[0], args.local_rollout_forward_batch_size):
                        query = queries[i : i + args.local_rollout_forward_batch_size]
                        query_response, logits = generate(
                            unwrapped_model.policy,
                            query,
                            tokenizer,
                            generation_config,
                        )
```

instead of 

```
                 for i in range(0, queries.shape[0], args.local_rollout_forward_batch_size):
                   with unwrap_model_for_generation(model, self.accelerator) as unwrapped_model:
                        query = queries[i : i + args.local_rollout_forward_batch_size]
                        query_response, logits = generate(
                            unwrapped_model.policy,
                            query,
                            tokenizer,
                            generation_config,
                        )
```

examples/scripts/minimal/ppo_zephyr.py
```diff
@@ -0,0 +1,142 @@
...
+        padding_side="left",
+        trust_remote_code=True,
+    )
+    tokenizer.add_special_tokens({"pad_token": "[PAD]"})
+    if tokenizer.chat_template is None:
+        # a default chat template to simply concatenate the messages
+        tokenizer.chat_template = (
+            "{% for message in messages %}\n{% if message['role'] == 'user' %}\n{{ '<|user|>\n' + message['content'] + eos_token }}\n{% elif message['role'] == 'system' %}\n{{ '<|system|>\n' + message['content'] + eos_token }}\n{% elif message['role'] == 'assistant' %}\n{{ '<|assistant|>\n'  + message['content'] + eos_token }}\n{% endif %}\n{% if loop.last and add_generation_prompt %}\n{{ '<|assistant|>' }}\n{% endif %}\n{% endfor %}"
+        )
+    value_model: PreTrainedModel = AutoModelForSequenceClassification.from_pretrained(
```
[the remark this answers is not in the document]
vwxyzjn: 1. yes it is possible to use a randomly initialized value model
2. The `ValueHead` as the name suggests is just a value head that uses the hidden states from the policy.

OAI did an ablation study as follows (https://arxiv.org/pdf/2009.01325): 
image

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,825 @@
...
+    def get_train_dataloader(self) -> DataLoader:
+        return self.dataloader
+    
+    def get_eval_dataloader(self) -> DataLoader:
+        return self.eval_dataloader
+
+    def push_to_hub(self, **kwargs):
+        """Modified from `Trainer.save_model` to only save the policy and not the value network."""
+        self.backup_model = self.model
+        self.model = self.accelerator.unwrap_model(self.model).policy # save only the policy
```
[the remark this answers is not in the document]
vwxyzjn: That's a good point. Maybe what we can do is to save `self.model.policy` to a repo and `self.model.value` to a separate repo. This way the users can still call the AutoModel.from_pretrained(...) If you want to experiment with this that will be great!

examples/scripts/minimal/ppo_zephyr.py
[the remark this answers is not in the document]
lapp0: Oh perfect! I knew I'd seen this chart before https://github.com/huggingface/trl/issues/1063

Glad it's being integrated!

trl/trainer/ppov2_trainer.py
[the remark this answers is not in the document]
lapp0: Should be doable, will have a go at it later after getting peft working for ppov2!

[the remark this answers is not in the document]
vwxyzjn: Sounds great!!!

vwxyzjn:
Sharing some updates here @lapp0. vLLM based approaches are 2x faster. I believe now the bottlenecks is no longer the generation but rather the forward passes and backward passes.

image

I found the PPO check points start to get reasonable MT Bench scores. Very exciting!

image

lapp0:
Awesome news! As discussed in the other thread, I was able to get generation down substantially, but it's still about 2x the forward/backward passes. vLLM is super fast, but unfortunately I don't have the extra ~60GB it requires to run a 7B model. 

I think we should consider optional (and recommended in docs) vLLM inference for those with sufficient hardware, and unsloth for those who don't.

trl/trainer/rloo_trainer.py
```diff
@@ -0,0 +1,723 @@
...
+                            unwrapped_model,
+                            query,
+                            tokenizer,
+                            generation_config,
+                        )
+                    response = query_response[:, context_length:]
+
+                    # use the logits during generation directly, instead of using the following
+                    all_logprob = F.log_softmax(logits, dim=-1)
+                    logprob = torch.gather(all_logprob, 2, response.unsqueeze(-1)).squeeze(-1)
```
lapp0, line 479: Why is `ref_logprob` scaled by `args.temperature + 1e-7`, but this `logprob` isn't?

trl/trainer/rloo_trainer.py
```diff
@@ -0,0 +1,723 @@
...
+    attention_mask = queries != tokenizer.pad_token_id
+    input_ids = torch.masked_fill(queries, ~attention_mask, 0)
+    output = lm_backbone.generate(
+        input_ids=input_ids,
+        attention_mask=attention_mask,
+        # position_ids=attention_mask.cumsum(1) - attention_mask.long(), # not needed: already adjusted in generations
+        # https://github.com/huggingface/transformers/blob/ac33aeeeee2a7a89b89c93c2962e6feb90daef0a/src/transformers/models/gpt2/modeling_gpt2.py#L1227-L1250
+        generation_config=generation_config,
+        return_dict_in_generate=True,
+        output_scores=True,
```
lapp0, line 187: Why do we use `output_scores` instead of `output_logits`?

trl/trainer/rloo_trainer.py
vwxyzjn: Great question. It's a limitation of the API, the logits from `generate` is already scaled.

To check it, feel free to print `logprob - ref_logprob`.

trl/trainer/rloo_trainer.py
```diff
@@ -0,0 +1,723 @@
...
+                    response = query_response[:, context_length:]
+
+                    # use the logits during generation directly, instead of using the following
+                    all_logprob = F.log_softmax(logits, dim=-1)
+                    logprob = torch.gather(all_logprob, 2, response.unsqueeze(-1)).squeeze(-1)
+                    del logits, all_logprob
+                    torch.cuda.empty_cache()
+
+                    ref_output = forward(ref_policy, query_response, tokenizer)
+                    ref_logits = ref_output.logits[:, context_length - 1 : -1]
```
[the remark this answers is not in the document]
vwxyzjn: That is correct. Before the optimizer pass, they should be the same.

trl/trainer/rloo_trainer.py
vwxyzjn: Good point! We can know (since this is a new API :) )

lapp0:
Why is loss consistently `0.0` in your wandb log? https://wandb.ai/costa-huang/huggingface/runs/astk72ix?nw=nwusercostahuang

vwxyzjn:
@lapp0 you mean the value loss? That's because in RLOO there is no value network 😂. So `vf_loss_stats` is not updated at all. We should remove it for the RLOO trainer.

image

lapp0:
Oh right...

trl/trainer/rloo_trainer.py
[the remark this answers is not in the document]
lapp0: Shouldn't `logits` from `generate()` always be the same as logits from the later call to

```
logits = output.logits[:, context_length - 1 : -1]
logits /= args.temperature + 1e-7
```

If so, won't `new_logprobs - mb_logprobs` always be 0?

trl/trainer/rloo_trainer.py
lapp0: As I understand it the difference between scores and logits is the application of logitprocessor.

Sampling with temp 1.0, scores != logits
```
>>> inputs = tokenizer("Hello, please tell me if the logits work", return_tensors="pt")
>>> output = base_model.generate(**inputs, output_logits=True, output_scores=True, return_dict_in_generate=True, do_sample=True, temperature=1.0)
>>> >>> torch.max(F.softmax(torch.stack(output.logits), dim=2), dim=2).values
tensor([[0.1969],
        [0.2204],
        [0.2907],
        [0.3935],
        [0.8413],
        [0.9996],
        [0.9149],
        [0.1471],
        [0.3852]])
>>> torch.max(F.softmax(torch.stack(output.scores), dim=2), dim=2).values
tensor([[0.2024],
        [0.2977],
        [0.3351],
        [0.4187],
        [0.8420],
        [0.9997],
        [0.9149],
        [0.1526],
        [0.4085]])
```

No sampling, logits and scores are equal
```
>>> output = base_model.generate(**inputs, output_logits=True, output_scores=True, return_dict_in_generate=True, do_sample=False)
>>> torch.max(F.softmax(torch.stack(output.logits), dim=2), dim=2).values
tensor([[0.1969],
        [0.4525],
        [0.3918],
        [0.9702],
        [0.7385],
        [0.6670],
        [0.1989],
        [0.9851],
        [0.3038]])
>>> torch.max(F.softmax(torch.stack(output.scores), dim=2), dim=2).values
tensor([[0.1969],
        [0.4525],
        [0.3918],
        [0.9702],
        [0.7385],
        [0.6670],
        [0.1989],
        [0.9851],
        [0.3038]])
````

I noticed this when I saw a difference between the `generate()` logits and the `forward()` logits.

File: trl/trainer/rloo_trainer.py (hunk 30 above)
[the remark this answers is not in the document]
Comment (vwxyzjn) [reply trl/trainer/rloo_trainer.py:491]: > If so, won't new_logprobs - mb_logprobs always be 0?

No. When PPO does second epoch or second minibatch, the `ratio = (new_logprobs - mb_logprobs).exp()` no longer is 1. That's when PPO's clipped objective comes in.

**Comment by vwxyzjn** [comment]
To show it works, per Arash's suggestion, I also ran experiments on TL;DR to see if it works. Should have more results in https://wandb.ai/costa-huang/huggingface/reports/ppo-rloo-tldr--Vmlldzo3ODUzNDEx tomorrow.

File: examples/accelerate_configs/deepspeed_zero2.yaml [hunk 31]
```diff
@@ -2,7 +2,6 @@ compute_environment: LOCAL_MACHINE
 debug: false
 deepspeed_config:
   deepspeed_multinode_launcher: standard
-  gradient_accumulation_steps: 1
```
younesbelkada, line 5: why this has been removed?

examples/accelerate_configs/deepspeed_zero3.yaml
```diff
@@ -2,7 +2,6 @@ compute_environment: LOCAL_MACHINE
 debug: false
 deepspeed_config:
   deepspeed_multinode_launcher: standard
-  gradient_accumulation_steps: 1
```
younesbelkada, line 5: same question as above 👀 if that's intended that"s fine though

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,804 @@
...
+    return variance
+
+
+def masked_whiten(values, mask, shift_mean=True):
+    """Whiten values with masked values."""
+    mean, var = masked_mean(values, mask), masked_var(values, mask, False)
+    whitened = (values - mean) * torch.rsqrt(var + 1e-8)
+    if not shift_mean:
+        whitened += mean
+    return whitened
```
younesbelkada, line 156: those look the same as in: https://github.com/huggingface/trl/blob/3b4c24946b7d5580fd354b0e3800fc1047b82a41/trl/core.py#L152 - can't you re-use them from trl.core?

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,804 @@
...
+
+
+def get_reward(model, query_responses, tokenizer, context_length):
```
younesbelkada, line 159: can you move this method to `trl.core`?

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,804 @@
...
+    attention_mask = query_responses != tokenizer.pad_token_id
+    # position_ids = attention_mask.cumsum(1) - attention_mask.long()  # exclusive cumsum
```
younesbelkada, line 161:
```suggestion
```

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,804 @@
...
+    lm_backbone = getattr(model, model.base_model_prefix)
+    input_ids = torch.masked_fill(query_responses, ~attention_mask, 0)
+    output = lm_backbone(
+        input_ids=input_ids,
+        attention_mask=attention_mask,
+        # position_ids=position_ids,
```
younesbelkada, line 167:
```suggestion
```

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,804 @@
...
+
+    def forward(self, **kwargs):
+        output = self.critic_backbone(
+            **kwargs,
+        )
+        logits = self.value_model.score(output.hidden_states[-1])
+        return self.policy(**kwargs), logits
+
+
+def exact_div(a, b):
```
younesbelkada, line 204: Same method as: https://github.com/huggingface/trl/blob/3b4c24946b7d5580fd354b0e3800fc1047b82a41/trl/trainer/utils.py#L583

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,804 @@
...
+
+def generate(lm_backbone, queries, tokenizer, generation_config):
+    """generate in a way that does not affect padding tokens"""
+    context_length = queries.shape[1]
+    attention_mask = queries != tokenizer.pad_token_id
+    input_ids = torch.masked_fill(queries, ~attention_mask, 0)
+    output = lm_backbone.generate(
+        input_ids=input_ids,
+        attention_mask=attention_mask,
+        # position_ids=attention_mask.cumsum(1) - attention_mask.long(), # not needed: already adjusted in generations
```
younesbelkada, line 219:
```suggestion
```

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,804 @@
...
+    return model(
+        input_ids=input_ids,
+        attention_mask=attention_mask,
+        # position_ids=position_ids,
+        return_dict=True,
+        output_hidden_states=True,
+    )
+
+
+def prepare_deepspeed(model, per_device_train_batch_size):
```
younesbelkada, line 262: can you move this method to `trl.trainer.utils` as it can potentially used by other trainers in the future I think 🙏

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,804 @@
...
+        ref_policy: nn.Module,
+        reward_model: nn.Module,
+        train_dataset: Dataset,
+        value_model: Optional[nn.Module] = None,
+        data_collator: Optional[DataCollatorWithPadding] = None,
+        eval_dataset: Optional[Union[Dataset, Dict[str, Dataset]]] = None,
+        # less commonly used
+        optimizers: Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
+        # compute_metrics: Optional[Callable[[EvalPrediction], Dict]] = None,
+        # model_init: Optional[Callable[[torch.nn.Module], None]] = None,
```
younesbelkada, line 312:
```suggestion
```

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,804 @@
...
+                        "zero_optimization.stage3_param_persistence_threshold": 10 * hidden_size,
+                        "zero_optimization.stage3_prefetch_bucket_size": 0,
+                    }
+                )
+    model, *_ = deepspeed.initialize(model=model, config=config_kwargs)
+    model.eval()
+    return model
+
+
+class PPOTrainer(Trainer):
```
younesbelkada, line 297: Hmm shouldn't we use a different name here? Something like `PPOv2Trainer`

trl/trainer/rloo_trainer.py
```diff
@@ -0,0 +1,715 @@
...
+        reward_logits,
+        reward_logits[
+            torch.arange(reward_logits.size(0), device=reward_logits.device),
+            sequence_lengths,
+        ].squeeze(-1),
+        sequence_lengths,
+    )
+
+
+def exact_div(a, b):
```
younesbelkada, line 165: Same comment as for PPOv2

trl/trainer/rloo_trainer.py
```diff
@@ -0,0 +1,715 @@
...
+        attention_mask=attention_mask,
+        # position_ids=position_ids,
+        return_dict=True,
+        output_hidden_states=True,
+    )
+    reward_logits = model.score(output.hidden_states[-1])
+    sequence_lengths = (
+        first_true_indices(query_responses[:, context_length:] == tokenizer.pad_token_id) - 1 + context_length
+    )
+    # https://github.com/huggingface/transformers/blob/dc68a39c8111217683bf49a4912d0c9018bab33d/src/transformers/models/gpt2/modeling_gpt2.py#L1454
```
younesbelkada, line 154: I see this is used both by PPOv2 and here, can you move them in `trainer.utils` ?

trl/trainer/rloo_trainer.py
```diff
@@ -0,0 +1,715 @@
...
+        policy: nn.Module,
+        ref_policy: nn.Module,
+        reward_model: nn.Module,
+        train_dataset: Dataset,
+        data_collator: Optional[DataCollatorWithPadding] = None,
+        eval_dataset: Optional[Union[Dataset, Dict[str, Dataset]]] = None,
+        # less commonly used
+        optimizers: Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
+        # compute_metrics: Optional[Callable[[EvalPrediction], Dict]] = None,
+        # model_init: Optional[Callable[[torch.nn.Module], None]] = None,
```
younesbelkada, line 272:
```suggestion
```

examples/scripts/minimal/ppo.py
younesbelkada, line 1: Should we move all ppo-related minimal scripts under a new `ppo/` dir and rloo under `rloo/` dir ? What do you think?

younesbelkada (commented):
Thanks a lot for this huge work ! I left some comments ! I think the new classes should be also exposed in TRL's main init - LMK wdyt about my suggestions below 🙏

examples/accelerate_configs/deepspeed_zero2.yaml
vwxyzjn: We can specifiy the `gradient_accumulation_steps` via `TrainingArguments` and accelerate will handle that automatically! Hard-coding `gradient_accumulation_steps` here I think causes issues for training if we do `--gradient_accumulation_steps 16`.

younesbelkada: makes sense, thanks for explaining !

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: I added this:

```
        if args.total_episodes is None:  # allow the users to define episodes in terms of epochs.
            args.total_episodes = args.num_train_epochs * self.train_dataset_len
```

examples/scripts/minimal/ppo_bandit_rloo.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+"""
+# run ppo
+python examples/scripts/minimal/ppo_bandit_rloo.py \
+    --learning_rate 3e-6 \
+    --output_dir models/minimal/ppo \
+    --per_device_train_batch_size 64 \
+    --gradient_accumulation_steps 1 \
+    --total_episodes 10000 \
```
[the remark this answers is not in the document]
vwxyzjn: I used 

```
        if args.total_episodes is None:  # allow the users to define episodes in terms of epochs.
            args.total_episodes = args.num_train_epochs * self.train_dataset_len
```

This way people can use it like other trainers with just num_train_epochs, per_device_train_batch_size, and gradient accumulation_steps. Would that address this concern?


What does `max_steps` mean? Does each step correspond to doing a learning pass on `per_device_train_batch_size * gradient_accumulation_steps`?

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,732 @@
...
+    """the name of this experiment"""
+    run_name: Optional[str] = None
+    """a unique name of this run"""
+    deepspeed2: bool = False
+    """Whether to use deepspeed to train the model"""
+    deepspeed3: bool = False
+    """Whether to use deepspeed to train the model"""
+
+    # various batch sizes
+    world_size: Optional[int] = None
```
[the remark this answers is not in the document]
vwxyzjn: Actually yes, the `world_size` is set by `accelerate` later. I put it here for experiment tracking purposes. For example `world_size` helps calculate the `args.batch_size` (which is the effective / global batch size across all processes).

examples/scripts/minimal/ppo.py
```diff
@@ -0,0 +1,108 @@
...
+    --deepspeed3 \
+    --non_eos_penalty \
+"""
+
+
+if __name__ == "__main__":
+    parser = HfArgumentParser((PPOConfig, ModelConfig))
+    config, model_config = parser.parse_args_into_dataclasses()
+    # remove output_dir if exists
+    shutil.rmtree(config.output_dir, ignore_errors=True)
```
lewtun, line 48: FYI you can set `overwrite_output_dir` in `PPOConfig` (via `TrainingArguments`)

trl/models/utils.py
```diff
@@ -150,7 +150,7 @@ def unwrap_model_for_generation(
     if accelerator.state.deepspeed_plugin is not None and accelerator.state.deepspeed_plugin.zero_stage == 3:
         with deepspeed.zero.GatheredParameters(model.parameters()):
             remove_hooks(model)
-            yield model
+            yield accelerator.unwrap_model(model)
```
lewtun, line 153: I believe models wrapped with the `DeepSpeedEngine` can still generate, so I'm curious why this is needed

trl/trainer/ppov2_trainer.py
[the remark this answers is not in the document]
lewtun: I see, but why do we only seem to need this for the RL trainers and not the other ones like `SFTTrainer`? In general, I'd like to avoid exposing this distributed stuff to the user if we can because it might not be clear if they should set the value manually or let `accelerate` handle it for them

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,799 @@
...
+    """wether to run in debug mode"""
+
+    # various batch sizes
+    world_size: Optional[int] = None
+    """The number of processes (GPUs) to use"""
+    num_updates: Optional[int] = None
+    """The number of updates to train"""
+    total_episodes: Optional[int] = None
+    """The total number of episodes in the dataset"""
+    micro_batch_size: Optional[int] = None
```
lewtun, line 55: As with `world_size` is this something that we expect the user to set manually? I wonder if these attributes can be set as a post init or within the trainer init directly if we don't expect users to set it

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,799 @@
...
+
+from trl.models.utils import unwrap_model_for_generation
+from trl.trainer.utils import exact_div, print_rich_table
+
+
+INVALID_LOGPROB = 1.0
+
+
+@dataclass
+class PPOConfig(TrainingArguments):
```
lewtun, line 39: For consistency with our other trainers, can we move this to `ppov2_config.py` ?

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,799 @@
...
+    # various batch sizes
+    world_size: Optional[int] = None
+    """The number of processes (GPUs) to use"""
+    num_updates: Optional[int] = None
+    """The number of updates to train"""
+    total_episodes: Optional[int] = None
+    """The total number of episodes in the dataset"""
+    micro_batch_size: Optional[int] = None
+    """The micro batch size across devices (HF's `per_device_train_batch_size` * `world_size`)"""
+    local_batch_size: Optional[int] = None
```
lewtun, line 57: Same question here as to whether we can set this in the init / post init

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,799 @@
...
+    """The batch size per GPU (HF's `per_device_train_batch_size` * `gradient_accumulation_steps`)"""
+    batch_size: Optional[int] = None
```
lewtun, line 59: And here too re hiding from the user

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,799 @@
...
+    """per rank no grad forward pass in the rollout phase"""
+    num_sample_generations: int = 10
+    """the number of debugging samples generations (i.e., `generate_completions` calls) throughout training"""
+
+    # other config
+    base_model: str = "EleutherAI/pythia-160m"
+    """the name of the pretrained model to use"""
+    response_length: int = 53
+    """the length of the response"""
+    truncate_token: Optional[Literal["eos"]] = None
```
lewtun, line 79: Is this the token to terminate generation on? If so, I would call it this for consistency with the `transformers` lib:

```suggestion
    stop_tokens: Optional[Literal["eos"]] = None
```

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,799 @@
...
+    """the truncate token"""
+    truncate_token_id: Optional[int] = None
+    """the truncation token id"""
+    temperature: float = 0.7
```
lewtun, line 83: Would it make sense to allow users to provide a `GenerationConfig` object here or is temperature the only hyperparameter we expect needs tuning?

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,799 @@
...
+    """the sampling temperature"""
+    penalty_reward_value: int = -1
+    """the reward value for responses that do not contain `truncate_token_id`"""
+    non_eos_penalty: bool = False
+    """whether to penalize responses that do not contain `truncate_token_id`"""
+    offload: bool = False
```
lewtun, line 89: Sorry if I mentioned this before, but this can be handled automatically by `accelerate` right? I.e. the user just passes `accelerate launch --config_file=config_with_offload.yaml` and then we handle the offloading internally

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,799 @@
...
+    lam: float = 0.95
+    """the lambda value for GAE"""
+    whiten_rewards: bool = False
+    """whether to whiten the rewards"""
+    kl_coef: float = 0.05
+    """the KL coefficient"""
+
+
+# taken from https://github.com/vwxyzjn/direct-preference-optimization/blob/f8b8c0f49dc92a430bae41585f9d467d3618fe2f/utils.py#L99
+def disable_dropout(model: torch.nn.Module) -> None:
```
lewtun, line 116: Shall we move this to model utils and add a unit test? https://github.com/huggingface/trl/blob/main/trl/models/utils.py

trl/trainer/ppov2_trainer.py
lewtun: If we use `PPOv2Trainer` do you think it will be fine to later rename it to `PPOTrainer` if we eventually deprecate the old one?

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,799 @@
...
+        self.callbacks = callbacks
+
+        #########
+        # calculate various batch sizes
+        #########
+        if args.total_episodes is None:  # allow the users to define episodes in terms of epochs.
+            args.total_episodes = args.num_train_epochs * self.train_dataset_len
+        accelerator = Accelerator(gradient_accumulation_steps=args.gradient_accumulation_steps)
+        self.accelerator = accelerator
+        args.world_size = accelerator.num_processes
```
lewtun, line 333: Ah I see now how this attribute is being set! Can't we just add it as a trainer attribute instead of including it in the config?

trl/trainer/ppov2_trainer.py
```diff
@@ -0,0 +1,799 @@
...
+                    # fmt: off
+                    del (
+                        output, vpred_temp, logits, new_all_logprobs, new_logprobs, vpred, vpredclipped,
+                        vf_losses1, vf_losses2, vf_loss, vf_clipfrac, logprobs_diff, ratio, pg_losses, pg_losses2, pg_loss_max,
+                        pg_loss, loss, pg_clipfrac, prob_dist, entropy, approxkl, mb_return,
+                        mb_advantage, mb_values, mb_responses, mb_query_responses, mb_logprobs,
+                    )
+                    # fmt: on
+                    torch.cuda.empty_cache()
+                # accelerator.print(
```
lewtun, line 694: Delete this dead code?

trl/trainer/rloo_trainer.py
```diff
@@ -0,0 +1,700 @@
...
+
+from trl.models.utils import unwrap_model_for_generation
+from trl.trainer.utils import exact_div, print_rich_table
+
+
+INVALID_LOGPROB = 1.0
+
+
+@dataclass
+class RLOOConfig(TrainingArguments):
```
lewtun, line 39: Same comment here as the `PPOConfig` - let's move this to `rloo_config.py`

trl/trainer/rloo_trainer.py
```diff
@@ -0,0 +1,700 @@
...
+    # common config
+    exp_name: str = os.path.basename(__file__)[: -len(".py")]
+    """the name of this experiment"""
+    run_name: Optional[str] = None
+    """a unique name of this run"""
+    sanity_check: bool = False
+    """wether to run in debug mode"""
+
+    # various batch sizes
+    world_size: Optional[int] = None
```
lewtun, line 49: Same comment here about hiding this from the user and setting in the trainer itself

lewtun (commented):
Thanks for iterating on this epic PR @vwxyzjn ! Overall it's looking quite close to being finished and I think the main remaining points to address are splitting off the configs into their own modules and seeing if we can hide config variables like `world_size` from the end user

trl/trainer/ppov2_bandit_rloo_trainer.py
vwxyzjn: let's do this possibility in a refactor :)

examples/scripts/minimal/ppo.py
vwxyzjn: I gave it a quick test but it does not seem to remove the output_dir. 

image

A quick search shows that the removing logic seems no longer there.

image

trl/models/utils.py
vwxyzjn: Ah it causes issues for RLOO, having some errors like "DeepSpeedEngine has not attribute generate`, so we still need to unwrap it.

trl/trainer/ppov2_trainer.py
vwxyzjn: IIRC you mentioned that `GenerationConfig` cannot be tuned via CLI or h4's yaml, so my idea was to put it in the arguments. 

For the most part temperature and generation length are the only parameters to tune. Standard literature does not use things like beam search or top k for generation purposes.

trl/trainer/ppov2_trainer.py
vwxyzjn: ah sorry forgot to remove it. I just removed it and in practice DS3 would just work.

vwxyzjn:
Thank you @lewtun @younesbelkada @lapp0 for the review. I have addressed most of the concerns and also added some docs and benchmarks. Let me know if there is anything else needed :D

younesbelkada (approved):
Huge work ! Thanks @vwxyzjn ! Good for me to merge once @lewtun is happy about the latest changes + CI is green ! 🚀

lapp0:
Great work @vwxyzjn! Really impressed with the vLLM integration along with the other components you've introduced here.

I'll be working on a follow-up PR for quantized training using ppo_v2 once [Unsloth's numerical stability issue](https://github.com/unslothai/unsloth/issues/427#issuecomment-2099866374) is resolved, and hopefully incorporate a few structural changes as well, so I don't have any further comments on structure right now. 

Did any of your RLOO runs result in improved benchmarks or at least improved score metrics? I was able to [reproduce improving scores with ppov2](https://huggingface.co/lapp0/ppov2_experiment_v2/tensorboard)  my [refactor of your branch with BnB / peft support](https://github.com/lapp0/trl/tree/ppov2), but I never managed to do the same with RLOO.

vwxyzjn:
@lapp0 Very nice to hear your great results with PPOv2 and peft! I was able to get 1B RLOO good results on tl;dr summarization. See https://moon-ci-docs.huggingface.co/docs/trl/pr_1540/en/rloo_trainer#benchmark-experiments.

lapp0:
Great work @vwxyzjn really exciting research and implementation you have put together. Feel free to ping me on any other PRs.

PR #1554, opened 2024-04-18
initial DPOConfig

kashif:


examples/scripts/dpo.py
```diff
@@ -166,14 +167,14 @@ def process(row):
             model,
             model_ref,
             args=training_args,
-            beta=args.beta,
+            beta=training_args.beta,
```
lewtun, line 170: This arg should now live in `args` right? That is, we should remove all these params from the trainer init

lewtun (commented):
Thanks for the refactor and new config @kashif ! Overall LGTM with some nits about moving all  args from the trainer init to the config in the script example

As with the SFT trainer, it would be good to have a small integration test that checks we have the same trainer if it is initialised in both ways (via `args` and via init) to ensure backwards compat

younesbelkada (approved):
Great work thanks @kashif ! LGTM once @lewtun comments are addressed !

trl/commands/cli_utils.py
```diff
@@ -156,12 +156,8 @@ class SftScriptArguments:
 @dataclass
 class DpoScriptArguments:
```
lewtun, line 157: turbo nit:
```suggestion
class DPOScriptArguments:
```

lewtun (approved):
Thanks for iterating @kashif ! I left one turbo nit and then this can be merged after addressing it :)

trl/commands/cli_utils.py
```diff
@@ -156,12 +156,8 @@ class SftScriptArguments:
...
     dataset_name: str = field(default=None, metadata={"help": "the dataset name"})
-    beta: float = field(default=0.1, metadata={"help": "the beta parameter for DPO loss"})
-    max_length: int = field(default=512, metadata={"help": "max length of each sample"})
-    max_prompt_length: int = field(default=128, metadata={"help": "max length of each sample's prompt"})
-    max_target_length: int = field(
-        default=128, metadata={"help": "Only used for encoder decoder model. Max target of each sample's prompt"}
-    )
+    dataset_train_name: str = field(default="train", metadata={"help": "the name of the training set of the dataset"})
```
lewtun, line 159: nit:
```suggestion
    dataset_train_split: str = field(default="train", metadata={"help": "The dataset split to use for training"})
```

trl/commands/cli_utils.py
```diff
@@ -156,12 +156,8 @@ class SftScriptArguments:
...
+    dataset_test_name: str = field(default="test", metadata={"help": "the name of the training set of the dataset"})
```
lewtun, line 160:
```suggestion
    dataset_test_split: str = field(default="test", metadata={"help": "The dataset split to use for evaluation"})
```

Issue #1564, opened 2024-04-19
TrlParser not respecting config 


philschmid:
Hey, 
i am trying to use the new TrlParser but i am not getting it work. I tried main and 0.8.5 with 
minimal script
```python
from transformers import TrainingArguments
from trl.commands.cli_utils import  TrlParser

if __name__ == "__main__":
    parser = TrlParser((TrainingArguments))
    training_args = parser.parse_args_and_config()
    print(training_args[0].bf16)
    print(training_args[0].fsdp)
    
```
yaml
```yaml
# training parameters
output_dir: "./llama-70b-hf-no-robot"  # Temporary output directory for model checkpoints
bf16: true
fsdp: "full_shard auto_wrap offload"

```

Error: 
```bash
False
[]
```

run with

```bash
python test.py --config config.yaml
```

PR #1565, opened 2024-04-19
CLI: Add warning when ignored params are passed + parse config file if config if passed

younesbelkada:
Fixes https://github.com/huggingface/trl/issues/1564


younesbelkada:
cc @philschmid

philschmid (approved):

PR #1573, opened 2024-04-23
 A pull request for POVIDTrainer 

gzcch:
This RP supports the POVID trainer in [arxiv.org/pdf/2402.11411.pdf](https://arxiv.org/pdf/2402.11411.pdf). Note that POVID's loss is a modified case of  [DPO](https://arxiv.org/pdf/2305.18290.pdf) for  Large Vision-Language models, as shown below:

POVID

We add the `trl/trainer/povid_trainer.py` to support our method, where we mainly modified the `get_batch_metrics` function.


kashif:
@gzcch would it make sense to subclass the `DPOTrainer`?

gzcch:
Yes, it would make sense to subclass the DPOTrainer.

kashif:
@gzcch now that `DPOTrainer` supports VLLMs can you kindly check how to incorporate this loss via the `DPOTrainer`?

PR #1588, opened 2024-04-25
Fix max_completion_length for encoder_decoder models in KTO Trainer

samuki:
For encoder_decoder models the argument, `max_completion_length` is missing in the kto_trainer, leading to an error in [line 205](https://github.com/huggingface/trl/blob/1d0a7ea17b8055a6850970ab59a34709d8ca494d/trl/trainer/kto_trainer.py#L205).

### DIFF
--- a/trl/trainer/kto_trainer.py
+++ b/trl/trainer/kto_trainer.py
@@ -526,6 +526,7 @@ class KTOTrainer(Trainer):
                 "truncation_mode": self.truncation_mode,
                 "label_pad_token_id": self.label_pad_token_id,
                 "max_prompt_length": self.max_prompt_length,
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Thanks !

younesbelkada:
WDYT @kashif ?

kashif (approved):

kashif:
also looks good to me... let me ask @prompteus to have a quick check too as they trained KTO models using T5

prompteus:
Hi, sorry for the delay. I didn't run into this issue in my case because I used the KTO as the loss parameter in the DPO Trainer (the paired variant of KTO), but the changes look alright.

PR #1590, opened 2024-04-26
Set GPU as default device for chat if CUDA is available

lewtun:
This PR tweaks the way we set the device map of the chat CLI so that we shard the model by default on all GPUs if CUDA is available. This should make it a bit more intuitive for users running models on their own devices and forgetting to specify the device explicitly.


lewtun:
Closing in favour of simpler PR

PR #1591, opened 2024-04-26
fixed adding bos and eos token unconditionally

jasonyux:
The original implementation in DPO/CPO will add BOS and EOS token to a tokenized sequence regardless if its already there. This is problematic when using tokenizers (e.g., gemma-2b) whose chat template _already_ adds BOS/EOS token.

The change is therefore to simply check if the start/last token is already BOS/EOS, before adding it.

### DIFF
--- a/trl/trainer/cpo_trainer.py
+++ b/trl/trainer/cpo_trainer.py
@@ -410,23 +410,23 @@ class CPOTrainer(Trainer):
                 )
 
             # add BOS token to head of prompt. Avoid adding if it's already there
-            if tokenizer.bos_token_id != prompt_tokens["prompt_input_ids"][0]:
-                prompt_tokens["prompt_input_ids"] = [tokenizer.bos_token_id] + prompt_tokens["prompt_input_ids"]
+            if self.tokenizer.bos_token_id != prompt_tokens["prompt_input_ids"][0]:
+                prompt_tokens["prompt_input_ids"] = [self.tokenizer.bos_token_id] + prompt_tokens["prompt_input_ids"]
                 prompt_tokens["prompt_attention_mask"] = [1] + prompt_tokens["prompt_attention_mask"]
-            if tokenizer.bos_token_id != chosen_tokens["prompt_input_ids"][0]:
-                chosen_tokens["prompt_input_ids"] = [tokenizer.bos_token_id] + chosen_tokens["prompt_input_ids"]
+            if self.tokenizer.bos_token_id != chosen_tokens["prompt_input_ids"][0]:
+                chosen_tokens["prompt_input_ids"] = [self.tokenizer.bos_token_id] + chosen_tokens["prompt_input_ids"]
                 chosen_tokens["prompt_attention_mask"] = [1] + chosen_tokens["prompt_attention_mask"]
-            if tokenizer.bos_token_id != rejected_tokens["prompt_input_ids"][0]:
-                rejected_tokens["prompt_input_ids"] = [tokenizer.bos_token_id] + rejected_tokens["prompt_input_ids"]
+            if self.tokenizer.bos_token_id != rejected_tokens["prompt_input_ids"][0]:
+                rejected_tokens["prompt_input_ids"] = [self.tokenizer.bos_token_id] + rejected_tokens["prompt_input_ids"]
                 rejected_tokens["prompt_attention_mask"] = [1] + rejected_tokens["prompt_attention_mask"]
 
             # add EOS token to end of answer. Avoid adding if it's already there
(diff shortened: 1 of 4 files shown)


kashif (approved):
yeah this was done without regards to the chat templates so this makes sense!

jasonyux:
I have fixed a minor bug (using `self.tokenizer`) so that it should now pass the code checks. I also added this fix to `ORPO` since it does the same thing.

kashif:
does KTO suffer from this issue too?

jasonyux:
It does seem so. In the above commits I primarily fixed all code that has `def tokenize_row` (i.e. DPO, CPO, and ORPO). I just now skimmed KTO, and I think `def _process_tokens` makes a similar mistake. I will update shortly.

kashif:
thank you!

jasonyux:
I have updated KTO trainer by:
1. determines `max_length` by first checking if EOS/BOS is already there
2. similar to the fix to CPO/DPO/ORPO, check for EOS/BOS token ids before adding them

kashif:
yes kindly run `pre-commit run --all-files` in the root of TRL to fix formatting

kashif:
@jasonyux so the tests are failing...

jasonyux:
it seems that sometimes `prompt_token_id` can be empty, hence it results in indexing error when I check for the first token/last token in the list. I added a condition to first check if the token ids are empty before checking for BOS/EOS token, which should patch this.

Is there a way for me to run the test cases before I commit?

jasonyux:
Thanks, this makes things much easier. I will update here when I make other changes after checking/passing the test cases.

jasonyux:
> Thanks, this makes things much easier. I will update here when I make other changes after checking/passing the test cases.

I just ran `pytest tests/test_*_trainer.py` for DPO, CPO, KTO, and ORPO. **All cases where cleared if I specify `export CUDA_VISIBLE_DEVICES=`** .

When I allowed a GPU, there seems to be pickling error that I don't understand why its there. Is this normal?

```
trl/trainer/dpo_trainer.py:492: in __init__                                                                                                                                                                        
    train_dataset = train_dataset.map(self.tokenize_row, num_proc=self.dataset_num_proc)                                                                                                                           
/usr/local/lib/python3.8/dist-packages/datasets/arrow_dataset.py:592: in wrapper
    out: Union["Dataset", "DatasetDict"] = func(self, *args, **kwargs)
[other error message omitted]
/usr/local/lib/python3.8/dist-packages/dill/_dill.py:412: in save
    StockPickler.save(self, obj, save_persistent_id)
/usr/lib/python3.8/pickle.py:560: in save
    f(self, obj)  # Call unbound method with explicit self
/usr/lib/python3.8/pickle.py:886: in save_tuple
    save(element)
/usr/local/lib/python3.8/dist-packages/dill/_dill.py:412: in save
    StockPickler.save(self, obj, save_persistent_id)
/usr/lib/python3.8/pickle.py:560: in save
    f(self, obj)  # Call unbound method with explicit self
/usr/local/lib/python3.8/dist-packages/dill/_dill.py:1212: in save_module_dict
    StockPickler.save_dict(pickler, obj)
/usr/lib/python3.8/pickle.py:971: in save_dict
    self._batch_setitems(obj.items())
/usr/lib/python3.8/pickle.py:1002: in _batch_setitems
    save(v)
/usr/local/lib/python3.8/dist-packages/dill/_dill.py:412: in save
    StockPickler.save(self, obj, save_persistent_id)
/usr/lib/python3.8/pickle.py:578: in save
    rv = reduce(self.proto)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    def __getstate__(self):
>       raise pickle.PicklingError(
            "Cannot pickle a prepared model with automatic mixed precision, please unwrap the model with `Accelerator.unwrap_model(model)` before pickling it."
        )
E       _pickle.PicklingError: Cannot pickle a prepared model with automatic mixed precision, please unwrap the model with `Accelerator.unwrap_model(model)` before pickling it.

/usr/local/lib/python3.8/dist-packages/accelerate/utils/operations.py:678: PicklingError
```

jasonyux:
Thanks @kashif . It seems that all test cases are cleared now.

Issue #1550, opened 2024-04-18
User turn masking is not robust with the DataCollatorForCompletionOnlyLM


edbeeching:
I have been running some experiments using `packing=False` and the [DataCollatorForCompletionOnlyLM](https://github.com/huggingface/trl/blob/75c1c47fcc922136ca1de249d79d34742ab1783b/trl/trainer/utils.py#L69). 

## Problem:
Perhaps I have been using this incorrectly but I find the masking implementation to be not very robust to adjacent words/symbols.
Some examples using the template from the docs and the llama2 tokenizer:
```
chat_template = f"### Question: {question}\n ### Answer: {answer}"
response_template = " ### Answer:"
```

```
tokenizer.encode(" ### Answer:", add_special_tokens=False)    # [28705,     774, 26307, 28747] 
tokenizer.encode("\n ### Answer:", add_special_tokens=False)  # [28705, 13, 774, 26307, 28747]
tokenizer.encode(".\n ### Answer:", add_special_tokens=False) # [842,   13, 774, 26307, 28747]
```
The collator uses the response_token_ids (first line above) to identify where in the sequence of token ids to start / end masking, but as you can see from the examples above, the encoded token ids change based on the words/symbols around them. Because of this mismatch many samples have their labels all set to -100 and do not contribute to the loss. I have tried a few variations to include the `\n` etc in the template but there are too many edge cases to resolve.

## Proposed solution:
I suggest that we have a new/updated collator and add the option to include masking tokens within the chat template itself. We use them to identify the start/end indices of where to mask, but the mask tokens themselves are not added to the model embedding layer. We would omit mask token ids from the sequence of token ids and adjust the masking indices accordingly, or set the mask token ids to -100.
An example:
```
tokenizer.add_special_tokens({"additional_special_tokens":["<MASK_START>", "<MASK_END>"]})
template = f"<MASK_START>### Question: {question}\n<MASK_END> ### Answer: {answer}"
# use these tokens to identify where to mask 
```

Potential issue: if the mask tokens sits between two symbols that are usually mapped to one index by the tokenizer then there may be a mismatch between tokenization with and without the mask tokens.

lewtun:
cc @younesbelkada who implemented the masking collator I believe and may have some good input on the above!

RonanKMcGovern:
Thanks, yeah that's definitely a more robust suggestion to explicitly define the mask.

PR #1592, opened 2024-04-26
Added DataCollatorForMultiTurnCompletions

AswanthManoj:
This PR adds a new data collator `DataCollatorForMultiTurnCompletions` to the `trl` library. This data collator is designed for multi-turn completion tasks, where the prompts are ignored while only considering assistant completions in multi-turn conversation datasets. 

The existing `DataCollatorForCompletionOnlyLM` is designed for single-turn completion tasks, where the input contains a single prompt or instruction, and the model is expected to generate a single response. However, it does not handle multi-turn conversational scenarios effectively, where the input sequence might contain multiple user prompts and assistant responses interspersed.

The `DataCollatorForMultiTurnCompletions` class inherits from `DataCollatorForLanguageModeling` and extends its functionality to handle multi-turn conversations. It takes two additional arguments:

1. `user_template`: A string or list of token IDs that indicate the start of a user prompt or input.
2. `assistant_template`: A string or list of token IDs that indicate the start of an assistant response.

During the forward pass, the collator identifies the assistant responses based on the provided templates (or token IDs), and masks out the user prompts during the loss calculation. This ensures that the model is trained to generate appropriate responses based on the user input, while not letting the model learn on the entire text.

The new data collator can be used as follows:

```python
    from trl import SFTTrainer, DataCollatorForMultiTurnCompletions
    
    # Load and prepare your dataset
    dataset = ...
    
    model_id = "<your model id>"

    model = AutoModelForCausalLM.from_pretrained(model_id)
    
    # If the dataset text field contains multi-turn conversation in `chatml` format.
    
    user_template = "<|im_start|>user\n"
    assistant_template="<|im_start|>assistant\n"
    
    collator = DataCollatorForMultiTurnCompletions(user_template=user_template, assistant_template=assistant_template, tokenizer=tokenizer, mlm=False)

    trainer = SFTTrainer(
        model,
        train_dataset=dataset,
        dataset_text_field="text",
        data_collator=collator,
    )

    trainer.train()
```

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -208,6 +208,116 @@ class DataCollatorForCompletionOnlyLM(DataCollatorForLanguageModeling):
         return batch
 
 
+class DataCollatorForMultiTurnCompletions(DataCollatorForLanguageModeling):
+    """
+    Data collator used for multi-turn assistant completion tasks. It ensures that the loss is calculated only on the responses
+    made by the assistant, while ignoring the user prompts or or other inputs.
+
+    Args:
+        user_template (Union[str, List[int]]): The template or token IDs that indicate the start of a user prompt or input. typically something like
+            '### Response:\n'. It can also be passed as tokenized ids, which can be useful when using a tokenizer that encodes the response
+            differently if it does not have proper context.
+        assistant_template (Union[str, List[int]]): The template or token IDs that indicate the start of an assistant response. typically something like
+            '### Human:\n'. Useful for assistant-style conversation datasets. It can also be passed as tokenized ids.
+        mlm (bool, optional, defaults to False): Whether or not to use masked language modeling in the underlying
+            `DataCollatorForLanguageModeling` class. Note that this option currently has no effect but is present 
+            for flexibility and backwards-compatibility.
+        ignore_index (int, optional, defaults to -100): The index to use to ignore the non-response tokens.
+
+    Example usage:
+
+    ```python
+    from datasets import load_dataset
+    from transformers import AutoModelForCausalLM, AutoTokenizer
+    from trl import SFTTrainer, DataCollatorForMultiTurnCompletions
+    
+    # Load and prepare your dataset
+    dataset = ...
+    
+    model_id = "<your model id>"
+
+    tokenizer = AutoTokenizer.from_pretrained(model_id)
+    model = AutoModelForCausalLM.from_pretrained(model_id)
+    
+    # If the dataset text field contains multi-turn conversation in `chatml` format.
+    
+    user_template = "<|im_start|>user\n"
+    assistant_template="<|im_start|>assistant\n"
+    
+    collator = DataCollatorForMultiTurnCompletions(user_template=user_template, assistant_template=assistant_template, tokenizer=tokenizer, mlm=False)
+
+    trainer = SFTTrainer(
+        model,
+        train_dataset=dataset,
+        dataset_text_field="text",
+        data_collator=collator,
+    )
+
+    trainer.train()
+
+    ```
+
+    This data collator can be used with a multi-turn conversational dataset, where each instance contains multiple
+    user prompts and assistant responses. The collator will identify the assistant responses based on the provided
+    templates (or token IDs), and mask out the user prompts during the loss calculation. This ensures that the model
+    is trained to generate appropriate responses based on the user input, while ignoring its own previous responses.
+    """
+    def __init__(
+        self,
+        user_template: Union[str, List[int]],
+        assistant_template: Union[str, List[int]],
+        *args,
+        mlm: bool = False,
+        ignore_index: int = -100,
+        **kwargs,
+    ):
+        super().__init__(*args, mlm=mlm, **kwargs)
+
+        self.user_template = user_template
+        if isinstance(user_template, str):
+            self.user_token_ids = self.tokenizer.encode(self.user_template, add_special_tokens=False)
+        else:
+            self.user_token_ids = user_template
+
+        self.assistant_template = assistant_template
+        if isinstance(assistant_template, str):
+            self.assistant_token_ids = self.tokenizer.encode(self.assistant_template, add_special_tokens=False)
+        else:
+            self.assistant_token_ids = assistant_template
+
+        self.ignore_index = ignore_index
+
+    def torch_call(self, examples: List[Union[List[int], Any, Dict[str, Any]]]) -> Dict[str, Any]:
+        batch = super().torch_call(examples)
+        input_ids = batch["input_ids"]
+        labels = input_ids.clone()
+
+        for i in range(input_ids.size(0)):
+            assistant_response_idxs = []
+            for start_idx in np.where(labels[i] == self.assistant_token_ids[0])[0]:
+                if self.assistant_token_ids == labels[i][start_idx:start_idx + len(self.assistant_token_ids)].tolist():
+                    end_idx = start_idx + len(self.assistant_token_ids)
+                    while end_idx < labels.size(1) and labels[i][end_idx] != self.user_token_ids[0]:
+                        end_idx += 1
+                    assistant_response_idxs.append((start_idx, end_idx))
+
+            if len(assistant_response_idxs) == 0:
+                warnings.warn(
+                    f"Could not find response key `{self.assistant_template}` in the "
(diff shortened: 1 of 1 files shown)


edbeeching:
Hi. Thanks for this implementation, we have also been looking at this internally. [Here](https://github.com/huggingface/trl/issues/1550) I describe some of the problems with the current implementation. 

One challenge with this templating approach are the tokenizer's "edge effects" when mapping strings to ID's. For example `"<|im_start|>user\n"` may not be mapped to the same sequence of IDs when a whitespace follows it. Some examples from the issue, with a different tempalte:
```
chat_template = f"### Question: {question}\n ### Answer: {answer}"
response_template = " ### Answer:"
tokenizer.encode(" ### Answer:", add_special_tokens=False)    # [28705,     774, 26307, 28747] 
tokenizer.encode("\n ### Answer:", add_special_tokens=False)  # [28705, 13, 774, 26307, 28747]
tokenizer.encode(".\n ### Answer:", add_special_tokens=False) # [842,   13, 774, 26307, 28747]
```

I think the simplest approach is to use only special tokens such as `<|im_start|>` or create special <mask_start> and <mask_end> tokens that indicate where the masking should be done.  

Let me know what you think and we can iterate on a solution.

AswanthManoj:
Hi there,

Thank you for explaining the tokenizer edge effects and the challenges with the templating approach. The idea of using special tokens like `<mask_start>` and `<mask_end>` is a valid solution, but I think an alternative and simpler approach could address this issue more effectively.

Usually, when we pass the dataset, we use the text column for tokenization. Instead of using the text column, we could directly pass the messages column, which has an array of turns.

For example:
```python
messages = [
    {"role": "user", "content": "Hello there"},
    {"role": "assistant", "content": "Hello user nice to meet you."}
]
```

When using DataCollatorForMultiTurnCompletions and during data preprocessing, we can tokenize each role without adding special tokens and concatenate them together. For masking the completions, we can set them to -100.

```python
roles_to_ignore = ['user', 'tool_response', 'system']
input_ids = [tokenizer.bos_token_id]
for turn in messages:
    tokenized = tokenizer.encode(turn['content'], add_special_tokens=False)
    if turn['role'] in roles_to_ignore:
        input_ids.extend([-100] * len(tokenized))
    else:
        input_ids.extend(tokenized)
```

By tokenizing and masking each message individually, we can avoid tokenizer edge effects and ensure consistent masking. Also using `roles_to_ignore` we will have flexibility to specify which roles we should ignore while calculating the loss.

Let me know if you would like me to elaborate further on this approach or if you have any other suggestions.

AswanthManoj:
Hi there, Thank you for checking out the implementation. I have made a simple solution considering the tokenizer's "edge effects" by splitting the prompt based on the eos_token or any special token that occurs in every turn. This allows us to identify the assistant's turns and others, enabling us to apply the mask appropriately.

```python
class DataCollatorForMultiTurnCompletions(DataCollatorForLanguageModeling):
    """
    Data collator for multi-turn conversational data. It masks the responses in the input sequences with the
    specified `ignore_index` token, except for the instances where the `response_template` string is present.

    Args:
        separator_token (Union[str, int]): The token used to separate different turns in the conversation.
        response_template (Union[str, List[int]]): The string or token IDs representing the response template.
            Responses matching this template will not be masked.
        mlm (bool, optional): Whether to use Masked Language Modeling or not. Defaults to False.
        ignore_index (int, optional): The index to use for masking responses. Defaults to -100.

    Example usage:
        >>> tokenizer = AutoTokenizer.from_pretrained("gpt2")
        >>> collator = DataCollatorForMultiTurnCompletions(
        ...     separator_token="<|eot_id|>",
        ...     response_template="<|start_header_id|>assistant<|end_header_id|>",
        ...     tokenizer=tokenizer,
        ... )
        ... message = [
        ...     dict(role='user', content='hello there'),
        ...     dict(role='assistant', content="hi how are you"),
        ...     dict(role='user', content='i am great what about you?'),
        ...     dict(role='assistant', content='am also fine thank you'),
        ... ]
        >>> prompt = tokenizer.apply_chat_template(message, tokenize=True, add_special_tokens=False)
        >>> tokenizer.decode(prompt)
            <|begin_of_text|><|start_header_id|>user<|end_header_id|>

            hello there<|eot_id|><|start_header_id|>assistant<|end_header_id|>

            hi how are you<|eot_id|><|start_header_id|>user<|end_header_id|>

            i am great what about you?<|eot_id|><|start_header_id|>assistant<|end_header_id|>

            am also fine thank you<|eot_id|>
            
        >>> batch = collator([inputs])
        >>> batch["input_ids"]
        tensor([[128000, 128006, 882, 128007, 271, 15339, 1070, 128009, 
        128006, 78191, 128007, 271, 6151, 1268, 527, 499, 128009, 
        128006, 882, 128007, 271, 72, 1097, 2294, 1148, 922, 499, 30, 128009, 
        128006, 78191, 128007, 271, 309, 1101, 7060, 9901, 499, 128009]]), 
        
        >>> batch["labels"]
        tensor([[-100, -100, -100, -100, -100, -100, -100, -100, 
                128006, 78191, 128007, 271, 6151, 1268, 527, 499, 128009, 
                -100, -100, -100, -100, -100, -100, -100, -100, -100, -100, -100, -100, 
                128006, 78191, 128007, 271, 309, 1101, 7060, 9901, 499, 128009]])
    """
    def __init__(
        self,
        separator_token: Union[str, int],
        response_template: Union[str, List[int]],
        *args,
        mlm: bool = False,
        ignore_index: int = -100,
        **kwargs,
    ):
        
        if isinstance(separator_token, str):
            self.separator_token = separator_token
            separator_token_ids = self.tokenizer.encode(separator_token, add_special_tokens=False)
            if len(separator_token_ids) != 1:
                raise ValueError("The separator token should be a single token.")
        else:
            self.separator_token = self.tokenizer.decode(separator_token)
            
        if isinstance(response_template, str):
            self.response_template = response_template
            self.response_template_ids = self.tokenizer.encode(response_template, add_special_tokens=False)
        else:
            self.response_template = self.tokenizer.decode(response_template)
            self.response_template_ids = response_template
        self.ignore_index = ignore_index

    def torch_call(self, examples: List[Union[List[int], Any, Dict[str, Any]]]) -> Dict[str, Any]:
        batch = super().torch_call(examples)
        input_ids = batch["input_ids"]
        labels = []

        for i in range(input_ids.size(0)):
            decoded_input = self.tokenizer.decode(input_ids[i])
            splits = decoded_input.split(self.separator_token)
            instance_labels = []
            separator_token_is_eos_token = self.separator_token == self.tokenizer.eos_token
            for j, split in enumerate(splits):
                if split:
                    if separator_token_is_eos_token:
                        split = split + self.separator_token
                    else:
                        split = self.separator_token + split
                    if self.response_template in split:
                        instance_labels.extend(self.tokenizer.encode(split, add_special_tokens=False))
                    else:
                        instance_labels.extend(len(self.tokenizer.encode(split, add_special_tokens=False)) * [self.ignore_index])

            labels.append(instance_labels)

        if not any(self.response_template in self.tokenizer.decode(input_ids[i]) for i in range(input_ids.size(0))):
            warnings.warn(
                f"Could not find response template `{self.response_template}` in any instance. "
                f"These instances will be ignored in loss calculation. "
                f"Note, if this happens often, consider increasing the `max_seq_length`."
            )

        if len(input_ids) != len(labels):
            raise ValueError(
                "The lengths of input_ids and labels do not match after processing. "
                "This should not happen and may indicate a bug in the DataCollator."
            )

        return {"input_ids": input_ids, "labels": labels}
```

This modified implementation works by splitting the input sequences based on the provided `separator_token`, which separates different turns in the conversation. For each split:

- If it contains the `response_template`, the entire split is included in the labels without any masking.
- If it does not contain the `response_template`, the split is masked by replacing all tokens with the specified `ignore_index`.

Please check this out and do let me know.

hahuyhoang411:
This is cool. Thanks @AswanthManoj 

For the example, I think you have a wrong variable
`batch = collator([inputs])` -> `batch = collator([prompt])`

seolhokim:
Thanks for your work. @hahuyhoang411 But i got length mismatch error when i used it with `SFTTrainer`. even i tried padding also.

YooSungHyun:
@AswanthManoj This seems like a very valuable and meaningful contribution. Would it be possible for you to continue working on it?

PR #1596, opened 2024-04-28
[CLI] Use auto device map for model load

lewtun:
As a user of the chat CLI, I expect the model to be loaded on GPU by default instead of having to manually specify `--device "cuda"` 

I believe this can be achieved by simply setting `device_map="auto"` which will also work on CPU if the user should require it.

### DIFF
--- a/examples/scripts/chat.py
+++ b/examples/scripts/chat.py
@@ -33,7 +33,7 @@ from rich.markdown import Markdown
 from transformers import AutoModelForCausalLM, AutoTokenizer, TextIteratorStreamer
 
 from trl.commands.cli_utils import ChatArguments, TrlParser, init_zero_verbose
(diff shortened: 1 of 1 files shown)


examples/scripts/chat.py
```diff
@@ -220,7 +220,7 @@ def load_model_and_tokenizer(args):
         trust_remote_code=args.trust_remote_code,
         attn_implementation=args.attn_implementation,
         torch_dtype=torch_dtype,
-        device_map=get_kbit_device_map() if quantization_config is not None else None,
```
lewtun, line 223: I don't think the k-bit device map is actually needed for inference (one can shard and quantize AFAIK)

younesbelkada (approved):
sounds good, thanks !

PR #1599, opened 2024-04-29
[`KTOTrainer`] add BCO (reward shift and underlying distribution matching)

seanexp:
add Binary Classifier Optimization (BCO) loss function from https://arxiv.org/abs/2404.04656

Implemented BCE loss, reward shift and underlying distribution matching.

Also added example script at `examples/scripts/bco.py`

### DIFF
--- a/docs/source/kto_trainer.mdx
+++ b/docs/source/kto_trainer.mdx
@@ -84,6 +84,13 @@ After this one can then call:
 kto_trainer.train()
 ```
 
+## Loss Functions
+
+Given the binary signal data indicating whether a completion is desirable or undesirable for a prompt, we can optimize an implicit reward function that aligns with the key principles of Kahneman-Tversky's prospect theory, such as reference dependence, loss aversion, and diminishing sensitivity.
+
+The [BCO](https://arxiv.org/abs/2404.04656) authors train a binary classifier whose logit serves as a reward so that the classifier maps {prompt, chosen completion} pairs to 1 and {prompt, rejected completion} pairs to 0.
+The `KTOTrainer` can be switched to this loss via the `loss_type="bco"` argument.
+
 ## KTOTrainer
 
 [[autodoc]] KTOTrainer
--- a/examples/scripts/bco.py
+++ b/examples/scripts/bco.py
@@ -1,0 +1,177 @@
+"""
+Run the BCO training script with the commands below. In general, the optimal configuration for BCO will be similar to that of KTO.
+
+# Full training:
+python examples/scripts/bco.py \
+    --model_name_or_path=nnheui/stablelm-2-1_6b-sft-full \
+    --per_device_train_batch_size 16 \
+    --per_device_eval_batch_size 32 \
+    --num_train_epochs 1 \
+    --learning_rate 1e-6 \
+    --gradient_checkpointing \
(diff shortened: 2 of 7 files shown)


kashif (approved):

kashif:
@seanexp  can you kindly run:

```
pre-commit run --all-file
```
in the root folder of TRL to clean up the formatting?

seanexp:
@kashif 

My bad. Applied `pre-commit` at [a9a7a9c](https://github.com/huggingface/trl/pull/1599/commits/a9a7a9c0093d5a6da5b5ab3ae16f8fc999ffb6ed)

kashif:
thanks @seanexp one questions, would it make sense in the example script to have an option to try with different embedding models? or are things hard-coded for nomic-embed?

kashif:
@seanexp  you might need to add scikit-learn dep here: https://github.com/huggingface/trl/blob/main/setup.py#L72

kashif:
oh sorry @seanexp  i thought the scikit-learn is in the tests but its in the main trainer... so we need to add that as a regular dependency! 🙇🏽

seanexp:
@kashif 

I found that adding `scikit-learn` is redundant as it is a required package of `transformers`.

Should I still add `scikit-learn` as a regular dependency?

kashif:
@seanexp yeah lets remove the dep from the test then... you can see here that even though transformers is installed, the tests fail: https://github.com/huggingface/trl/actions/runs/8891165233/job/24412685192#step:5:2843

seanexp:
Seems like the dep in the test is necessary. Without the dep the test fails. Shall we leave as it is? @kashif

kashif:
@seanexp ok lets leave the import in the tests and add a helper here: https://github.com/huggingface/trl/blob/main/trl/import_utils.py for scikit-learn and then in the trainer we can check if its available when BCO is the loss type and if not, we can ask that they install it via `pip instal ...`

kashif:
@seanexp  something like this in the KTOConfig: https://github.com/huggingface/trl/blob/main/trl/trainer/ddpo_config.py#L116-L120

seanexp:
@kashif 

done at [ddbbbdf](https://github.com/huggingface/trl/pull/1599/commits/ddbbbdf7624e715b586cd805ad9bb1b65db22f7c) :)

seanexp:
> thanks @seanexp one questions, would it make sense in the example script to have an option to try with different embedding models? or are things hard-coded for nomic-embed?

Users can try different embedding models. Please note that users have to modify `embedding_func` (and `embed_prompt`) accordingly as embedding models asks different post processing methods (e.g. mean/max pooling, using only subset of features) @kashif

kashif (approved):
LGTM!

seanexp:
@younesbelkada 

hmm... I'll take a look few hours later!

seanexp:
Not calling `super().__post_init__()` was the problem. Fixed at [5c58065](https://github.com/huggingface/trl/pull/1599/commits/5c580655a2bebc3641fef688d5acb105ef083fc7)

The tests now run properly on my machine :)

kashif:
good catch @seanexp 🥇

younesbelkada (approved):
Woah great catch ! Thanks again for adding this nice feature !

PR #1604, opened 2024-04-30
Docs: Fix build main documentation

younesbelkada:
as per title and discussed offline

cc @mishig25 @Wauplin


Wauplin (commented):
You must also make the change in https://github.com/huggingface/trl/blob/main/.github/workflows/build_pr_documentation.yml, otherwise PR docs will continue to fail.

younesbelkada:
Ohh make sense, thanks @Wauplin !

Wauplin (approved):

Issue #1606, opened 2024-04-30
Bug: `tests` are being included in package


jamesbraza:
`pip install trl==0.8.6` is installing `tests/` on top of `trl/`, which is sadly breaking my imports

The request is to exclude `tests` from the published package

younesbelkada:
Thanks @jamesbraza for the fix ! Much appreciated !

PR #1607, opened 2024-04-30
Removing `tests/` from package data

jamesbraza:
Fixes https://github.com/huggingface/trl/issues/1606


younesbelkada (approved):
Makes sense, thanks !

PR #1613, opened 2024-05-02
Update HH dataset on helpful only subset

vwxyzjn:
This PR updates the scripts to use the helpful only subset of the anthropic's HH dataset.  

https://huggingface.co/datasets/trl-internal-testing/hh-rlhf-helpful-base-trl-style


lewtun (approved):
Thanks for the fix! LGTM

PR #1605, opened 2024-04-30
Adds Online DPO

edbeeching:
WIP ignore for now

usage
```
accelerate launch --config_file deepspeed_zero3.yaml examples/scripts/dpo_online.py ----model_name_or_path=HuggingFaceH4/mistral-7b-ift --model_revision=v25.2 --output_dir=data/mistral-7b-odpo --dataset_name=HuggingFaceH4/ultrafeedback_binarized --dataset_train_split=train_gen --dataset_test_split=test_gen --gradient_accumulation_steps=1 --bf16=True --attn_implementation=flash_attention_2 --per_device_train_batch_size=2
```

PR #1617, opened 2024-05-03
Fix ZeRO-3 generation context manager

### LINKED ISSUE #1543: Speed up ZeRO-3 generation with DPO
Hi, a [recent PR](https://github.com/huggingface/trl/pull/1483) brought large improvements (x10) to PPO generation with ZeRO-3.
@lewtun, you mention on the PR that it can be adapted for other trainers. I gave it a quick shot and it seems that naive applying the context manager to trainers like DPO does not work:
```
in remove_hooks
    if model.optimizer is not None and hasattr(
       ^^^^^^^^^^^^^^^^^^^^
    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
AttributeError: 'GPTNeoXForCausalLM' object has no attribute 'optimizer'
```
There seems to be an inconsistency between the base classes. Is there a reason why DPO is based on Trainer from transformers and PPO on BaseTrainer ? What would be the easy way to add this feature to other trainers ? Thanks !

lewtun:
This PR incorporates a fix from @pacman100 to enable the `unwrap_model_for_generation()` context manager to work with other trainers like DPO. 

The basic issue with our original implementation is that we weren't removing all active parameters during the hook removal and this led to DeepSpeed errors like:

```
AssertionError: {'id': 0, 'status': 'AVAILABLE', 'numel': 25755648, 'ds_numel': 25755648, 'shape': (50304, 512), 'ds_shape': (50304, 512), 'requires_grad': True, 'grad_shape': None, 'persist': False, 'active_sub_modules': {182}, 'ds_tensor.shape': torch.Size([3219456])}
```

Tested with this gist: https://gist.github.com/pacman100/a6c89a681f8f76bdf17f5bf6874eb983

cc @edbeeching you should rebase on `main` once this is merged so #1605 will work

Closes #1543

### DIFF
--- a/trl/models/utils.py
+++ b/trl/models/utils.py
@@ -1,3 +1,4 @@
+import itertools
 from contextlib import contextmanager
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Literal, Optional, Tuple, Union
@@ -118,6 +119,9 @@ def remove_hooks(model: "DeepSpeedEngine") -> None:
     elif model.optimizer is not None:
         optimizer_offload = model.optimizer
 
+    for param in iter_params(optimizer_offload.module, recurse=True):
+        param.ds_active_sub_modules.clear()
+
     for hook in optimizer_offload.forward_hooks:
         hook.remove()
     for hook in optimizer_offload.backward_hooks:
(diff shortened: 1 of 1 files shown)


edbeeching (approved):
Thanks, LGTM

younesbelkada (approved):
Thanks a lot !

vwxyzjn (approved):

PR #1624, opened 2024-05-06
[ORPO] Enable batched tokenization & multiprocessing to process large datasets

lewtun:
This PR applies a similar refactoring logic as #1470 to massively speed up the preprocessing of the `ORPOTrainer`. Similar to that PR, the `ORPOTrainer` currently applies all dataset preprocessing at the per-example-level via a method called `tokenize_row()`. This isn't ideal because:

- Rust tokenizers support batching which is typically 3-5x faster
- the method can't be hashed and thus parallelised via multiprocessing in `datasets`.

This PR massively speeds up the dataset preprocessing of `ORPOTrainer` by doing the following:

* Rewrites the tokenization step to use batching (best practice for Rust tokenizers). This is now called the `_tokenize` method and is done separately to the subsequent token ID processing
* Introduces a new `_process_tokens()` method that supports multiprocessing
* Migrates all data processing methods to live outside the trainer
* Adds a new `dataset_num_proc` attribute to the `ORPOConfig` to allow users to configure the multiprocessing steps

I've also added some unit tests for the new logic and verified that the `orpo.py` example script produces near identical results with this refactor.

Screenshot 2024-05-06 at 15 11 10
Screenshot 2024-05-06 at 15 10 59
Screenshot 2024-05-06 at 15 10 53


trl/trainer/orpo_trainer.py
```diff
@@ -59,6 +59,212 @@
...
+
+    for p, m in zip(prompt_input_ids, prompt_attention_mask):
+        if len(p) != len(m):
+            raise ValueError("Prompt input ids and attention mask should have the same length.")
+
+    rejected_input_ids = [f[r:] for f, r in zip(full_input_ids, response_token_ids_start_idx)]
+    rejected_attention_mask = [f[r:] for f, r in zip(prompt_and_rejected_attention_mask, response_token_ids_start_idx)]
+
+    return dict(
+        prompt_input_ids=prompt_input_ids,
```
lewtun, line 151: Note there is one _small_ difference with the new implementation: we do not distinguish between the prompt input IDs of the chosen / rejected pairs, favouring instead to have a single set of IDs per prompt.

In general, I think the processing logic could be simplified significantly, but for now I've focused on ensuring parity with the current version

trl/trainer/orpo_trainer.py
```diff
@@ -269,13 +475,46 @@ def make_inputs_require_grad(module, input, output):
...
+                "tokenizer": self.tokenizer,
+                "max_length": self.max_length,
+                "truncation_mode": self.truncation_mode,
+                "label_pad_token_id": self.label_pad_token_id,
+                "max_prompt_length": self.max_prompt_length,
+            }
+            train_dataset = train_dataset.map(
+                _process_tokens,
+                fn_kwargs=fn_kwargs,
+                num_proc=args.dataset_num_proc,
```
kashif, line 501: so should this be in the `_tokenize` map?

trl/trainer/orpo_trainer.py
```diff
@@ -59,6 +59,212 @@
...
+    in case the prompt + completion responses is/are too long. First
+        we truncate the prompt; if we're still too long, we truncate the completion.
+
+    We also create the labels for the completion responses, which are of length equal to
+        the sum of the length of the prompt and the completion response, with
+        label_pad_token_id  for the prompt tokens.
+    """
+    batch = {}
+
+    if not kwargs["is_encoder_decoder"]:
```
kashif, line 173: i guess we can remove this too

trl/trainer/orpo_trainer.py
```diff
@@ -59,6 +59,239 @@
...
 
 
+def _tokenize(batch: Dict[str, List[Any]], tokenizer: "PreTrainedTokenizerBase") -> Dict[str, List[Any]]:
+    """Tokenize a batch from a binary preference dataset."""
+    prompt_tokenized = tokenizer(batch["prompt"], add_special_tokens=False)
+    prompt_input_ids = prompt_tokenized["input_ids"]
+    prompt_attention_mask = prompt_tokenized["attention_mask"]
+
+    # Process prompt & chosen
+    prompt_and_chosen = [prompt + chosen for prompt, chosen in zip(batch["prompt"], batch["chosen"])]
```
vwxyzjn, line 69: Would it be `prompt + chosen` or just `chosen`? E.g., in ultrafeedback_binarized, `prompt` is `how can i develop a habit of drawing daily`, and `chosen` is `[ { "content": "how can i develop a habit of drawing daily", "role": "user" }, { "content": "Developing a daily habit of drawing can be challenging but with consistent practice and a few tips, it can become an enjoyable and rewarding part of your daily routine. Here are some strategies to help you develop the habit of drawing daily:\n\n1. Set a specific time: Allocate a specific time of the day to draw. It could be in the morning, afternoon, or evening. Make drawing a part of your daily routine.\n2. Set a specific duration: Determine the amount of time you want to spend on drawing each day. It can be as little as 10 minutes or as long as an hour. Be consistent with the duration to help build the habit.\n3. Start small and simple: Don't try to create a masterpiece every day, start with simple and easy-to-do sketches. Focus on improving your skills gradually.\n4. Use a variety of tools and mediums: Experiment with different tools like pencils, pens, markers, and different mediums like paper, canvas, or digital apps to keep your drawing practice interesting and engaging.\n5. Take breaks and rest: Taking breaks and resting after some time of drawing can help you avoid burnout and stay motivated.\n6. Challenge yourself: Set challenges like drawing objects from memory or a specific subject to improve your skills and keep your drawing practice interesting.\n7. Track your progress: Keep a record of your daily drawing practice and track your progress. This can be a source of motivation and help you see how far you've come.\n\nRemember, developing a habit takes time and patience. Stay consistent with your drawing practice, be flexible and open to trying new things, and with time, you'll develop a habit of daily drawing that brings you joy and satisfaction.", "role": "assistant" } ]`

image

PR #1636, opened 2024-05-09
visualize rm prediction

vwxyzjn:
This PR adds a utility to visualize rm prediction every time `evaluate` is called. It helps us better visualize and understand what RM is doing automatically. So, easier vibe tests :D


younesbelkada (approved):
Very cool work ! Thanks so much !

Issue #1631, opened 2024-05-08
Wrong prefix for logs in KTOTrainer


bartoszzuk:
The implementation of [log](https://github.com/huggingface/trl/blob/3b4c24946b7d5580fd354b0e3800fc1047b82a41/trl/trainer/kto_trainer.py#L1518) method in `KTOTrainer` uses wrong prefix `eval/` instead of `eval_` (similar with train metrics).
This results in names of metrics not being properly processed by logging callbacks (I tested on `WandbCallback`). The problem is specifically with function [rewrite_logs](https://github.com/huggingface/transformers/blob/e7a5f45ed1eaa10abed2797e92dde5be10f1be32/src/transformers/integrations/integration_utils.py#L574), which expects correct prefixes. This results in metric names in wandb like `train/train/rewards/margins` or `train/eval/rewards/margins` (see picture below). If you would like I can submit a pull request for this issue since the fix should be pretty trivial?

```python
def rewrite_logs(d):
    new_d = {}
    eval_prefix = "eval_"
    eval_prefix_len = len(eval_prefix)
    test_prefix = "test_"
    test_prefix_len = len(test_prefix)
    for k, v in d.items():
        if k.startswith(eval_prefix):
            new_d["eval/" + k[eval_prefix_len:]] = v
        elif k.startswith(test_prefix):
            new_d["test/" + k[test_prefix_len:]] = v
        else:
            new_d["train/" + k] = v
    return new_d
```

![Screenshot from 2024-05-08 16-23-54](https://github.com/huggingface/trl/assets/57541034/b4f76d33-e874-4397-adb7-75b8af8711af)

kawine:
thank for catching this @bartoszzuk ! please feel free to open a PR

PR #1641, opened 2024-05-13
Fixed wrong logs prefixes in KTOTrainer

bartoszzuk:
This PR fixes the issues with KTOTrainer logs described here #1631 

Fixes: https://github.com/huggingface/trl/issues/1631


younesbelkada (approved):
Thanks for fixing !

PR #1644, opened 2024-05-15
don't cast the trainable lora layers to half precision

pacman100:
### What does this PR do?
1. While the QLoRA repo https://github.com/artidoro/qlora/blob/7f4e95a68dc076bea9b3a413d2b512eca6d004e5/qlora.py#L395-L405 casts the trainable parameters to `bfloat16`(half-precision), this is incorrect and can lead to unstable training when using automatic mixed precision. Hence, this PR avoid casting the trainable params to half precision. This is inline with the changes we are making to fix this issue at https://github.com/artidoro/qlora/blob/7f4e95a68dc076bea9b3a413d2b512eca6d004e5/qlora.py#L395-L405
2. The impact of this PR would be increase in the GPU memory consumption as the trainable params are now in full precision.

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -668,9 +668,7 @@ def peft_module_casting_to_bf16(model):
     from peft.tuners.tuners_utils import BaseTunerLayer
 
     for name, module in model.named_modules():
-        if isinstance(module, BaseTunerLayer):
-            module = module.to(torch.bfloat16)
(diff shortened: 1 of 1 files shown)


pacman100:
cc @lvwerra and @younesbelkada

BenjaminBossan:
> 2\. The impact of this PR would be increase in the GPU memory consumption as the trainable params are now in full precision.

This should be highlighted in the next trl release notes.

Issue #1623, opened 2024-05-06
Learning to generate EOS tokens 


vwxyzjn:
@edbeeching and I noticed sometimes the trained SFT models do not learn to stop generations. In other words, the model never learn to generate EOS tokens.

Upon some digging, I noticed this is mainly an issue with the dataset preprocessing. In particular, if we simply pass a dataset like https://huggingface.co/datasets/timdettmers/openassistant-guanaco to the `SFTTrainer`, the trainer may not postpend the completion with an EOS token.

If we run `for item1, item2 in zip(inputs["input_ids"][1], inputs["attention_mask"][1]): print(item1, item2)` at https://github.com/huggingface/transformers/blob/91d155ea92da372b319a79dd4eef69533ee15170/src/transformers/trainer.py#L3207, with our SFT example we get

```
python examples/scripts/sft.py \
    --model_name_or_path="facebook/opt-350m" \
    --report_to="wandb" \
    --learning_rate=1.41e-5 \
    --per_device_train_batch_size=2 \
    --gradient_accumulation_steps=16 \
    --output_dir="sft_openassistant-guanaco" \
    --logging_steps=1 \
    --num_train_epochs=3 \
    --max_steps=-1 \
    --push_to_hub \
    --gradient_checkpointing \
    --dataset_text_field text
```

<img width="1105" alt="image" src="https://github.com/huggingface/trl/assets/5555347/889e2915-2007-48c1-8153-d201cd9496a9">

Notice how the pad token / eos token corresponds to attention mask = 0. 

## potential solution

This can be resolved if we add an **eos token to the dataset itself**. For example, https://github.com/huggingface/trl/blob/dc012ea683ee92c0a079abb01bccaa52fc15b675/examples/scripts/minimal/sft.py#L57 always adds an EOS token to the tokenized dataset, and as a result we get

```
python examples/scripts/minimal/sft.py \
    --learning_rate 3e-6 \
    --per_device_train_batch_size 1 \
    --per_device_eval_batch_size 1 \
    --gradient_accumulation_steps 32 \
    --learning_rate 5e-05 \
    --logging_steps 10 \
    --evaluation_strategy epoch \
    --max_seq_length 1024 \
    --num_train_epochs 5 \
    --output_dir models/minimal/sft
```

<img width="466" alt="image" src="https://github.com/huggingface/trl/assets/5555347/336389ac-2453-420a-b150-a71e540cf09f">

Notice how the first eos token corresponds to attention mask = 1.

[...]

PR #1646, opened 2024-05-16
Prototype Dataset Processor

vwxyzjn:
This PR attempts to refactor and pull **all tokenization logic** out of the Trainer class. Having a separate tokenization process gives us higher visibility into what's being used in training, providing more clarified logic and reducing bugs. It attempts to do the following things. 

```
# 1. PPO (prompt)
# 2. SFT (prompt + demonstration), there is also packing.
# 3. ✅ RM / DPO (chosen and rejected)
# 4. ✅ Visualization of length distributions?
# 5. ✅ Filter?
#   * Smart truncation?
# 6. ✅ dataset_num_proc
# 7. check EOS token
# 8. dataset mixer?
# 9. ✅ pretty print that show tokenization?
# 10. hashable tokneization?
# 11. inputs / labels / attention_mask
# 12. always set a `tokenizer.pad_token_id`?
```

# why?

Currently, the Trainer is also responsible for tokenization. It causes several issues:
1. **duplicate tokenization steps**. For example, [alignment-handbook](https://github.com/huggingface/alignment-handbook/blob/606d2e954fd17999af40e6fb4f712055ca11b2f0/scripts/run_sft.py#L133-L143) calls apply_chat_template(tokenize=False) for the dataset, followed by SFT/DPO trainer calling tokenized again. To remove duplication, we only needed to go through the dataset once by calling `apply_chat_template(tokenize=True)`
2. **truncation logic happens in various places and is hard to predict**.  SFTTrainer calls it the `max_seq_length`, RewardModeling calls it `max_length`, DPO/KTOTrainers call it `max_length`, `max_prompt_length`, `max_target_length`. There are also different truncation logics. E.g., [(truncate the prompt if prompt + chosen is too long)]
(https://github.com/huggingface/trl/blob/99f2c94b2200927a1dc156f16e012dca11f865e1/trl/trainer/dpo_trainer.py#L797-L799). This causes issue like https://huggingface.slack.com/archives/C04EX6W3QSY/p1715255460198239 as raised by @abhishekkrthakur. 
   * the hard truncation logic seems debatable: if the sequence length is too long, shouldn't we filter them out instead of giving a truncated response? The truncated response could be an incomplete code snippet / summaries (basically bad data). If truncation is really desired, we should do some kind of smart truncation like truncate at the last paragraph, so the sentences are still complete.

3. **learning to generate EOS tokens**. https://github.com/huggingface/trl/issues/1623#issuecomment-2113230864 suggested that EOS tokens always 1) correspond to -100 in the labels and 2) if the dataset contains the EOS token before collating, then the attention mask of EOS token is also 1. It's possible that the model may never learn to generate EOS tokens. 
    * what's a bit unclear to me is how zephyr learns to output EOS tokens, despite all the labels of EOS token are marked with -100 and are being masked out. My suspicion is that the attention_mask=1 plays some roles in it.
5. **`dataset_num_proc` is not uniformly applied**, as a result #1624 is needed. There is also the question of hashable tokenization 
6. Dataset mixer (e.g., in our h4 codebase), that should be more widely available to use in TRL and can be combined with this class.

## The current design

The current design roughly looks like this. Note that we can still put it in `Trainer.__init__` so users don't have to configure it directly.

```python
dataset_config = DatasetConfig(max_token_length=1024, max_prompt_token_lenth=128)
dataset_processor = PreferenceDatasetProcessor(tokenizer=tok, config=dataset_config)
train_dataset = dataset_processor.tokenize(preference_datasets["train"])
stats = dataset_processor.get_token_length_stats(train_dataset)
pprint.pp(stats)
train_dataset = dataset_processor.filter(train_dataset)
stats = dataset_processor.get_token_length_stats(train_dataset)
pprint.pp(stats)
dataset_processor.get_token_length_visualization(train_dataset)
print(tok.decode(train_dataset[0]["chosen"]))
visualize_token(train_dataset[0]["chosen"], tok)
```

image
image


kashif:
very cool! thanks! checking

edbeeching:
> what's a bit unclear to me is how zephyr learns to output EOS tokens, despite all the labels of EOS token are marked with -100 and are being masked out. My suspicion is that the attention_mask=1 plays some roles in it.

I think for zephyr we used packing and there is a concat token=eos that is not masked / ignored.

vwxyzjn:
@edbeeching you are right, because the datacollator is not called when using the packed dataset! The output below confirms it.

image

docs/source/reward_trainer.md
```diff
@@ -0,0 +1,418 @@
...
+    --eval_strategy steps \
+    --max_token_length 1280 \
+    --max_prompt_token_lenth 1024 \
+    --remove_unused_columns False \
+    --num_train_epochs 1 \
+    --eval_steps=300 \
+    --bf16 \
+    --output_dir models/rm/rm_tldr_1b \
+    --push_to_hub \
+    --hub_model_id trl-internal-testing/rm_tldr_1b
```
vwxyzjn, line 343: cc @qgallouedec

lewtun:
Bot begone!

PR #1647, opened 2024-05-17
Visual DPO

qgallouedec:
This PR makes add support for visual models (starting with Idefics2)

Changes:
- In `DPOTrainer` we check whether the model is visual (`self.is_vision_model`). If so, the argument `tokenizer` is the `processor`, and call the processor with the images as input.
- A `pad` function has been added to utils for more general padding support (which also applies to images).
- In `DPOCollatorWithPadding`, we integrate visual keys (`pixel_vlaues`) support.
- An example is added. At the moment we have a dedicated example file (`vdpo.py`), and it also adds visual support to `dpo.py`. But we should eventually chose which makes more sense. 

Test with:
```sh
accelerate launch examples/scripts/dpo.py \
    --dataset_name HuggingFaceH4/rlaif-v_formatted \
    --model_name_or_path HuggingFaceM4/idefics2-8b \
    --output_dir dpo_idefics_rlaif-v \
    --gradient_accumulation_steps 16 \
    --learning_rate 1e-5 \
    --bf16 \
    --torch_dtype bfloat16 \
    --use_peft \
    --lora_target_modules=all-linear
``` 

TODO:
- [ ] Allow number of images != 1 in prompt (test already implemented)


qgallouedec:
```python examples/scripts/vsft_llava.py \
    --dataset_name="HuggingFaceH4/llava-instruct-mix-vsft" \
    --model_name_or_path="llava-hf/llava-1.5-7b-hf" \
    --report_to="wandb" \
    --learning_rate=1.4e-5 \
    --per_device_train_batch_size=2 \
    --logging_steps=5 \
    --num_train_epochs=1 \
    --push_to_hub \
    --gradient_checkpointing \
    --torch_dtype=float16`
```
gives

```
{'loss': 1.7722, 'grad_norm': nan, 'learning_rate': 1.3999459784840019e-05, 'epoch': 0.0}                                                                                                                    
{'loss': 0.0, 'grad_norm': nan, 'learning_rate': 1.3998919569680038e-05, 'epoch': 0.0}                                                                                                                       
{'loss': 0.0, 'grad_norm': nan, 'learning_rate': 1.3998379354520057e-05, 'epoch': 0.0}                                                                                                                       
{'loss': 0.0, 'grad_norm': nan, 'learning_rate': 1.3997839139360076e-05, 'epoch': 0.0} 
```

trl/trainer/utils.py
```diff
@@ -354,13 +399,26 @@ def __call__(self, features: List[Dict[str, Any]]) -> Dict[str, Any]:
                         padding_value = self.label_pad_token_id
                     elif k.endswith("_attention_mask"):
                         padding_value = 0
+                    elif k.endswith("_pixel_values"):
+                        padding_value = 0  # TODO: check if this is correct
```
qgallouedec, line 403: Anyone have a counter example (ie pixel padding value not equal to 0?)

qgallouedec:
Should we able to fine-tune vlm on text-only preference data?

trl/trainer/utils.py
kashif: should we add it to the model or method config?

docs/source/dpo_trainer.mdx
```diff
@@ -70,8 +70,33 @@ dpo_dataset_dict = {
 
 where the `prompt` contains the context inputs, `chosen` contains the corresponding chosen responses and `rejected` contains the corresponding negative (rejected) responses. As can be seen a prompt can have multiple responses and this is reflected in the entries being repeated in the dictionary's value arrays.
 
+`DPOTrainer` can be used to fine-tune visual language models (VLMs). In this case, the dataset must also contain the key `images`. For example, for Idefics2, the processor expects the dataset to have the following format:
```
kashif, line 73:
```suggestion
`DPOTrainer` can be used to fine-tune visual language models (VLMs). In this case, the dataset must also contain the key `images`, and the trainer's `tokenizer` is the VLM's `processor`. For example, for Idefics2, the processor expects the dataset to have the following format:
```

kashif (approved):

qgallouedec:
For the record, I've considered adding support for Llava as well but Llava processor doesn't have `apply_chat_template`. However, `apply_chat_template` will be part of the default methods for all processor from the next transformers release: https://github.com/huggingface/transformers/pull/30744. So I recommend adding llava support later, in another PR.

trl/trainer/utils.py
qgallouedec: idk

qgallouedec:
ok ready to merge imo

fangkuoyu:
> For the record, I've considered adding support for Llava as well but Llava processor doesn't have `apply_chat_template`. However, `apply_chat_template` will be part of the default methods for all processor from the next transformers release: [huggingface/transformers#30744](https://github.com/huggingface/transformers/pull/30744). So I recommend adding llava support later, in another PR.

Has LLaVA been supported?

qgallouedec:
Not yet but I think you can slightly modify dpo visual to make it work with llava (just add the chat template)

fangkuoyu:
> Not yet but I think you can slightly modify dpo visual to make it work with llava (just add the chat template)

@qgallouedec Thanks for the comments!

PR #1653, opened 2024-05-22
[DPO] Add 'robust' loss_type

Abilityguy:
Add the 'Robust DPO' loss function from https://arxiv.org/abs/2403.00409

This PR implements the Robust DPO loss function from the paper. This loss function is an unbiased estimate of the DPO loss and experiments show that it performs better than cDPO and DPO in the presence of noisy preferences in data.


younesbelkada (approved):
Thanks for adding !

kashif (approved):

PR #1654, opened 2024-05-22
Do not upcast adapters when using FSDP+QLoRA

pacman100:
### What does this PR do?
1. When using FSDP, all tensors need to be of the same type when it flattens them. As such, we need to avoid upcasting the adapter parameters.


pacman100:
cc @BenjaimBossan @younesbelkada

younesbelkada (approved):
Thanks a lot !

PR #1659, opened 2024-05-23
Fix inheritance order in PPOv2Config

Nicolinho:
Hi, I get the following error when importing PPOv2Config
`TypeError: non-default argument 'output_dir' follows default argument`

Changing the order of inheritance in the class definition solves this problem.

### DIFF
--- a/trl/trainer/ppov2_config.py
+++ b/trl/trainer/ppov2_config.py
@@ -12,7 +12,7 @@ from ..trainer.utils import (
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


vwxyzjn:
Could you change that for the RLOOConfig as well? We can merge after that. Thanks!

vwxyzjn (approved):

Nicolinho:
I also changed the order in RLOOConfig. Thank you for your work!

vwxyzjn:
@Nicolinho there are some more issues of the refacotr. I will open another PR to fix them.

PR #1660, opened 2024-05-23
Fix wrong argument bug in PPOv2Trainer

Nicolinho:
Hi @vwxyzjn , I noticed that in the PPOv2Trainer at two occurrences a function expects the tokenizer's pad_id as an argument instead of the tokenizer itself.


vwxyzjn:
See #1661

vwxyzjn:
Closed by #1661

PR #1661, opened 2024-05-23
Fix ppov2 test case

vwxyzjn:
Follow up to #1659


vwxyzjn:
Regression report looks great. I did notice RLOO seems to experience more fluctuations. 

image

image

kashif (approved):

vwxyzjn:
Merging the hot fix. Will update the 6.9B results in a separate PR.

PR #1559, opened 2024-04-18
Apply deprecated `evaluation_strategy`

muellerzr:
This PR applies the deprecated `evaluation_strategy` from https://github.com/huggingface/transformers/pull/30190.

TL;DR: `evaluation_strategy` -> `eval_strategy`

(Should go live in v4.41.0, but opening now so it's here when ready!)

PR #1662, opened 2024-05-23
[Tests] update eval_strategy API

kashif:
fix some remaining `eval_strategy` calls missed by #1559 and add some tests for #1653


younesbelkada (approved):
Thanks a lot kashif !

PR #1670, opened 2024-05-27
fix dataset load error

sywangyi:
image
load dataset error if datasets >= 2.19.0.
error is like


sywangyi:
occurred after https://github.com/huggingface/datasets/pull/6714 is merged.

sywangyi:
@younesbelkada

younesbelkada (approved):
Thanks for this prompt fix !

PR #1680, opened 2024-05-29
transforme `tokenize_row` and `build_tokenized_answer` to static method

AIR-hl:
It is worth noting that these two methods are not only in DPOTrainer, but also in ORPOTrainer etc, but I only modified DPOTrainer


younesbelkada (commented):
Thanks a lot @AIR-hl ! After thinking a bit, instead of making the method static, would it makes sense to move it as a separate method within `utils.py` and import that method from there?

AIR-hl:
> Thanks a lot @AIR-hl ! After thinking a bit, instead of making the method static, would it makes sense to move it as a separate method within `utils.py` and import that method from there?

It's my pleasure to contribute to the community :). I think you are correct. After all, there is more than one Trainer using it.

Issue #1689, opened 2024-06-02
Using optimize_device_cache instead of optimize_cuda_cache doesn't work


alexisrozhkov:
`optimize_cuda_cache` in `PPOConfig` is [marked as deprecated](https://github.com/huggingface/trl/blob/151a452d14c8ebccbaf8a033812ceb2dc77f634d/trl/trainer/ppo_config.py#L108), and `optimize_device_cache` is recommended instead.

However, when I use `optimize_device_cache` instead of `optimize_cuda_cache`, `time/ppo/optimize_step` in my runs increases from ~42 to ~230 seconds.

It seems like currently when `optimize_cuda_cache` is not set it overrides whatever value was assigned to `optimize_device_cache` with `False`:
https://github.com/huggingface/trl/blob/151a452d14c8ebccbaf8a033812ceb2dc77f634d/trl/trainer/ppo_config.py#L142-L148

PR #1690, opened 2024-06-02
Fix overriding optimize_device_cache with optimize_cuda_cache in PPOConfig

alexisrozhkov:
Removed the unneeded else branch and added a check to prevent both options being set at the same time

Fixes: #1689


younesbelkada (approved):
Thanks !

PR #1692, opened 2024-06-04
Fix default padding_value in dpo_config.py

mnoukhov:
dpo_config default padding value should be None, not 0, otherwise it by default overrides the padding value of any tokenizer to 0


kashif (approved):

younesbelkada (approved):
Thanks !

PR #1696, opened 2024-06-05
Quick fix on GPT4-eval

vwxyzjn:
The `from vllm import SamplingParams, SingleGPULLM` was really from my own fork. This PR fixes it and makes it work with regular `vllm`.


younesbelkada (approved):
Thank you for fixing this !

PR #1701, opened 2024-06-05
adds AOT

imelnyk:
This code adds Distributional Preference Alignment Via Optimal Transport (AOT) to TRL. 

`self.loss_type == "aot"` - this part is for the cases where for each prompt there is a pair of chosen and rejected responses.

`self.loss_type == "aot_pair"` - here for each prompt the chosen and rejected responses do not need to be paired (e.g., as in kto_pair)

Note that for AOT algorithm to really show its advantage, the actual per-GPU batch size needs to be maximized. 

In our upcoming ArXiv paper we show that, for example, _ibm/merlinite-7b_ (original AlpacaEval LC Win Rate 17.1) can be aligned using our AOT method to get to _ibm/merlinite-7b-AOT_ with AlpacaEval LC Win Rate 31.3


kashif:
thanks @imelnyk can you also kindly add a description of the method and how to use it in the trainer's doc as well as have the tests also run with this loss?

younesbelkada (commented):
Great contribution @imelnyk ! As stated by @kashif can you add few lines in the documentation about this new method?

imelnyk:
Thanks for the reminder, @younesbelkada @kashif. Yes, we are working on it now and we'll update this PR soon. Thank you!

imelnyk:
@younesbelkada @kashif : I pushed the changes (docs + tests)

younesbelkada (approved):
Great contribution ! Thanks a lot for adding this !

kashif (approved):

PR #1686, opened 2024-06-01
intial RPO loss

kashif:
from the paper: https://arxiv.org/pdf/2404.19733

PR #1705, opened 2024-06-06
[RPO] fix nll loss

kashif:
fixes the nll for RPO loss from #1686 and log it when `rpo_alpha` is not `None`

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -1122,7 +1122,7 @@ class DPOTrainer(Trainer):
 
     def concatenated_forward(
         self, model: nn.Module, batch: Dict[str, Union[List, torch.LongTensor]]
-    ) -> Tuple[torch.FloatTensor, torch.FloatTensor, torch.FloatTensor, torch.FloatTensor]:
+    ) -> Tuple[torch.FloatTensor, torch.FloatTensor, torch.FloatTensor, torch.FloatTensor, torch.FloatTensor]:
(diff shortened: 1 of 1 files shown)


kashif:
@winglian can you kindly try with this fix?

trl/trainer/dpo_trainer.py
```diff
@@ -1225,7 +1241,7 @@ def get_batch_loss_metrics(
         reward_accuracies = (chosen_rewards > rejected_rewards).float()
 
         if self.args.rpo_alpha is not None:
-            losses = losses * self.args.rpo_alpha - policy_chosen_logps_avg
+            losses = losses * self.args.rpo_alpha + policy_nll_loss
```
lewtun, line 1244: Perhaps I'm missing something, but isn't the NLL loss normalized by the total sequence length? 


Screenshot 2024-06-06 at 14 54 39

lewtun (commented):
Code LGTM, but I have one question about where we normalize the NLL loss

trl/trainer/dpo_trainer.py
kashif: yes the `CrossEntropyLoss` is reducing by `mean` by default

kashif:
also note @lewtun the v2 of the paper has this loss averaged over the completions (c and y):

Screenshot 2024-06-06 at 17 47 49

lewtun (approved):
LGTM, thanks for the fix!

PR #1707, opened 2024-06-06
SFTTrainer: Fix backward Compatibility issue with `TrainingArguments`

younesbelkada:
# What does this PR do?

Fixes: https://github.com/huggingface/transformers/issues/31280

This PR makes sure users can still pass `TrainingArguments` to `SFTTrainer` without facing errors

cc @vwxyzjn @kashif

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -145,6 +145,8 @@ class SFTTrainer(Trainer):
             output_dir = "tmp_trainer"
             warnings.warn(f"No `SFTConfig` passed, using `output_dir={output_dir}`.")
             args = SFTConfig(output_dir=output_dir)
+        elif args is not None and args.__class__.__name__ == "TrainingArguments":
+            args = SFTConfig(**args.to_dict())
 
         if model_init_kwargs is not None:
             warnings.warn(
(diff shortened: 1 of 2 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -145,6 +145,8 @@ def __init__(
             output_dir = "tmp_trainer"
             warnings.warn(f"No `SFTConfig` passed, using `output_dir={output_dir}`.")
             args = SFTConfig(output_dir=output_dir)
+        elif args is not None and args.__class__.__name__ == "TrainingArguments":
+            args = SFTConfig(**args.to_dict())
```
vwxyzjn, line 149: Nice fix!

vwxyzjn (approved):
Thanks @younesbelkada!

trl/trainer/sft_trainer.py
anhuong: This still fails for me with error: `TypeError: SFTConfig.__init__() got an unexpected keyword argument 'cache_dir'` because TrainingArguments has a field for `cache_dir` but SFTConfig does not

younesbelkada: Hmm that's strange, there is no field `cache_dir` in TrainingArguments: https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py - can you show us how do you get that error?

anhuong: Apologies I see this is mixture of TrainingArguments where my project built a class off of transformers.TrainingArguments that we pass into trl.SFTTrainer and so the additional parameters that we are adding to TrainingArguments is causing the discrepancy issue.

PR #1713, opened 2024-06-06
fix yaml parser for derived config classes

### LINKED ISSUE #1712: YamlConfigParser fails on RewardConfig, DPOConfig etc..
Using the YamlConfigParser with derived classes of `TrainingArguments` throws an error because it assumes all classes that are `isinstance(class, TrainingArguments)` must be `TrainingArguments` and not a derived class

mnoukhov:
fixes #1712
reformatted cli_utils with ruff


mnoukhov:
I can change the examples to use the `TRLParser` as well to show this works

younesbelkada (approved):
Thanks a lot for the prompt fix !

PR #1718, opened 2024-06-09
Fix masking of response tokens

### LINKED ISSUE #1717: Problem handling of response masks
Current handling of `response_masks` inside `batch_forward_pass` function does not take padding into consideration which results with shape unmatch during masking. I think response tokens should not be concatenated with a `torch.zeros(query_length)` and masking operation should be done without slicing.

An example with batch size of 2:
- First sample in the batch has a query size of 10 and response size of 9 (response mask has also size of 9).
- Second sample in the batch has a query size of 10 and response size of 5(response mask has also size of 5).
- With the concatenation, `response_mask_batch[1]` has the size of 15.
- `start` will be 14 for second sample(due to the padding) and `end` will be 19.
Hence, 
The operation 
```
masks[j, start:end] = masks[j, start:end] * response_masks_batch[j][start:end]
```
will yield a `RuntimeError: The size of tensor a (5) must match the size of tensor b (1) at non-singleton dimension 0`
as `response_mask_batch[1][14:19]` is same as `response_mask_batch[1][14:15]` which has length of 1.

Removing the concatenation of the response mask and removing the slicing from the response mask since response mask already has the length of `end - start + 1`, which is equal to length of `masks[j, start:end]` should be a better approach.

mertsayar8:
Current handling of `response_masks` inside `batch_forward_pass` function does not take padding into consideration which results with shape unmatch during masking as stated in #1717. Response mask can directly be used for the masking operation instead of doing preprocessing such as concatenation.

Remove the concatenation of the response mask, remove the slicing from the response mask since response mask already has the length of `end - start`, which is equal to length of `masks[j, start:end]`.

Update the docstring for `masks` parameter under `_step_safety_checker` which is incompatible with the docstring under `step` function.

Fixes #1717

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -614,7 +614,7 @@ class PPOTrainer(BaseTrainer):
             scores (List[`torch.FloatTensor`]):
                 List of tensors containing the scores.
             masks (List[`torch.LongTensor`], *optional*):
-                list of optional tensors containing the masks of shape (`query_length` + `response_length`)
+                list of optional tensors containing the masks of shape (`response_length`)
         Returns:
             `tuple`: The input processed data.
         """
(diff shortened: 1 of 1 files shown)


younesbelkada:
cc @vwxyzjn what do you think?

vwxyzjn:
@mertsayar8 thanks for the fix! I am happy to merge it as is. That said, we recently made a refactor as added a PPOv2Trainer. Feel free to give that a try.

Issue #1651, opened 2024-05-22
KTO finetuning - float division by zero


jetlime:
I am attempting to finetune the [LLama3-8B-Instruct model](https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/) on the [UNSW-NB15 dataset](https://huggingface.co/datasets/Jetlime/NF-UNSW-NB15-v2).

```python
dataset = load_dataset("Jetlime/NF-UNSW-NB15-v2", streaming=False, split="train")

# Model
base_model = "meta-llama/Meta-Llama-3-8B-Instruct"

# QLoRA config
bnb_config = BitsAndBytesConfig(
    load_in_4bit=True,
    bnb_4bit_quant_type="nf4",
    bnb_4bit_compute_dtype=torch_dtype,
    bnb_4bit_use_double_quant=True,
)

# LoRA config
peft_config = LoraConfig(
    r=16,
    lora_alpha=32,
    lora_dropout=0.05,
    bias="none",
    task_type="CAUSAL_LM",
    target_modules=['up_proj', 'down_proj', 'gate_proj', 'k_proj', 'q_proj', 'v_proj', 'o_proj']
)

# Load tokenizer
tokenizer = AutoTokenizer.from_pretrained(base_model, token=HUGGING_FACE_READ_TOKEN)

# Load model
model = AutoModelForCausalLM.from_pretrained(
    base_model,
    quantization_config=bnb_config,
    attn_implementation=attn_implementation,
    token=HUGGING_FACE_READ_TOKEN
)
model, tokenizer = setup_chat_format(model, tokenizer)
model = prepare_model_for_kbit_training(model)

# Use only a small subset of the training set for a first finetuning trial
dataset = dataset.train_test_split(test_size=0.95, seed=123, stratify_by_column="Attack")
dataset_finetuning = dataset["train"]
dataset_finetuning

# Dataset({
#    features: ['input', 'output', 'Attack'],
#    num_rows: 113538
#})

# Creating the dataset columns required by the KTO finetuner
import random
def format_chat_template(row):
    row['prompt'] = row["input"]
    if random.randrange(0,1):
        row["label"] = False
        if row["output"] == 1:
            row["completion"] = '0'
        else:
            row["completion"] = '1'
    else:
        row["label"] = True
        row["completion"] = str(row["output"])
    return row

dataset_finetuning = dataset_finetuning.map(
    format_chat_template, num_proc=os.cpu_count()
)
dataset_finetuning
```

[...]

PR #1724, opened 2024-06-11
ktotrainer: Refuse datasets which contain only one class of labels

jetlime:
Fixes: #1651


younesbelkada (approved):
Very clean work ! Thanks so much for this contribution !

PR #1725, opened 2024-06-11
Adding SimPO to TRL

yumeng5:
Hello,

This PR adds [SimPO](https://arxiv.org/abs/2405.14734) implementations to TRL. 

The official codebase: https://github.com/princeton-nlp/SimPO

Thank you!
Yu


AIR-hl:
> Hello,
> 
> This PR adds [SimPO](https://arxiv.org/abs/2405.14734) implementations to TRL.
> 
> The official codebase: https://github.com/princeton-nlp/SimPO
> 
> Thank you! Yu

@yumeng5 hi! `simpo` has been implemented in `cpotrainer`  but not yet released

xiamengzhou:
@vwxyzjn Hi! Thank you for reaching out. We have carefully considered the integration of CPO and SimPO and have a few concerns:

**Hyperparameter Differences**: Although CPO and SimPO share some hyperparameters, the optimal values can vary significantly. For example, beta tends to be much larger in SimPO compared to other policy optimization algorithms.

**Additional Hyperparameter for SimPO**: SimPO introduces an extra hyperparameter, the gamma_beta ratio, which plays a crucial role in determining the appropriate target reward margin. Unlike CPO, which implements this as a constant gamma, we find that tuning the gamma_beta ratio in SimPO is more effective for reaching an optimal point.

Our experiments show that the optimal values for these hyperparameters, as well as the learning rate, can differ significantly from those used in DPO, SLiC-HF, or CPO, despite the similarities in their formulations. Therefore, we think it's better to develop a separate Trainer for SimPO to better highlight its unique requirements for users. What do you think?

vwxyzjn:
Hi @xiamengzhou, there are excellent reasons, and they make sense. However, I do have reservations about bloating the codebase from the maintenance point of view.

Maybe two possibilities are going forward:


1) better documentation and example: for example, you could add a `simpo.py` in the `examples/scripts` folder and by default, set the hyperparameters that make sense for simpo like here (
https://github.com/yumeng5/trl/blob/223ce737d651a78a9b54ee1a4472fc4e4eb61760/examples/scripts/simpo.py#L19-L32). Most people use examples when playing with the trainers anyway, so you can ensure that people use the right `beta` and so on. Then, in the documentation you can explain that SimPO uses different optimal hyperparameters and recommend the users.

2) do a very light wrapper of CPOTrainer. For example, instead of inheriting from `TrainingArgumetns,` maybe inherit from `CPOConfig,`, and you can name it `SimPOConfig.` That way you can override the default `beta` values. Similarly, you can inherit from CPOTrainer.

xiamengzhou:
@vwxyzjn Thanks for the response! Your suggestions make sense to me and we will revise the PR accordingly :) Will ping you here once we are done.

yumeng5:
Hi @vwxyzjn 

Thanks again for your suggestions. We have modified the PR by doing a wrapper of the CPOTrainer. Would you mind taking a look and letting us know if that looks good to you?

Thanks!

docs/source/simpo_trainer.mdx
```diff
@@ -0,0 +1,90 @@
+# SimPO Trainer
+
+[Simple Preference Optimization](https://arxiv.org/abs/2405.14734) (SimPO) by Yu Meng, Mengzhou Xia, and Danqi Chen proposes a simpler and more effective preference optimization algorithm than DPO without using a reference model. The key designs in SimPO are (1) using length-normalized log likelihood as the implicit reward, and (2) incorporating a target reward margin in the Bradley-Terry ranking objective.
+
+The official code can be found [here](https://github.com/princeton-nlp/SimPO).
```
qgallouedec, line 5:
```suggestion
The official code can be found at [princeton-nlp/SimPO](https://github.com/princeton-nlp/SimPO).
```

docs/source/simpo_trainer.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+## Expected dataset format
+
+The SimPO trainer expects a format identical to the DPO trainer, which should include three entries. These entries should be named as follows:
```
qgallouedec, line 9:
```suggestion
The SimPO trainer expects a format identical to the [DPO trainer](https://huggingface.co/docs/trl/trainer#trl.DPOTrainer), which should include three entries. These entries should be named as follows:

File: docs/source/simpo_trainer.mdx [hunk 3]
```diff
@@ -0,0 +1,90 @@
...
+        "Javascript",
+        "C++",
+        "C++",
+    ],
+}
+```
+where the `prompt` contains the context inputs, `chosen` contains the corresponding chosen responses and `rejected` contains the corresponding negative (rejected) responses. Note that a prompt can have multiple responses and this is reflected in the entries being repeated in the dictionary's value arrays.
+
+## Expected model format
+The SimPO trainer expects a model of `AutoModelForCausalLM`, compared to PPO that expects `AutoModelForCausalLMWithValueHead` for the value function.
```
qgallouedec, line 51:
```suggestion
The SimPO trainer expects a model of `AutoModelForCausalLM`.

File: docs/source/simpo_trainer.mdx [hunk 4]
```diff
@@ -0,0 +1,90 @@
...
+
+## Using the `SimPOTrainer`
+For a detailed example have a look at the `examples/scripts/simpo.py` script. At a high level we need to initialize the `SimPOTrainer` with a `model` we wish to train. **Note that SimPOTrainer eliminates the need to use the reference model, simplifying the optimization process.** The `beta` refers to the hyperparameter `lambda` in eq. (6) of the paper and refers to the weighting of the relative odd ratio loss in the standard cross-entropy loss used for SFT.
+
+```py
+simpo_config = SimPOConfig(
+    beta=2.5, # beta
```
qgallouedec, line 58:
```suggestion
    beta=2.5,
```

docs/source/simpo_trainer.mdx
```diff
@@ -0,0 +1,90 @@
...
+    ],
+}
+```
+where the `prompt` contains the context inputs, `chosen` contains the corresponding chosen responses and `rejected` contains the corresponding negative (rejected) responses. Note that a prompt can have multiple responses and this is reflected in the entries being repeated in the dictionary's value arrays.
+
+## Expected model format
+The SimPO trainer expects a model of `AutoModelForCausalLM`, compared to PPO that expects `AutoModelForCausalLMWithValueHead` for the value function.
+
+## Using the `SimPOTrainer`
+For a detailed example have a look at the `examples/scripts/simpo.py` script. At a high level we need to initialize the `SimPOTrainer` with a `model` we wish to train. **Note that SimPOTrainer eliminates the need to use the reference model, simplifying the optimization process.** The `beta` refers to the hyperparameter `lambda` in eq. (6) of the paper and refers to the weighting of the relative odd ratio loss in the standard cross-entropy loss used for SFT.
```
qgallouedec, line 54:
```suggestion
For a detailed example have a look at the `examples/scripts/simpo.py` script. At a high level we need to initialize the `SimPOTrainer` with a `model` we wish to train. **Note that SimPOTrainer eliminates the need to use the reference model, simplifying the optimization process.**
```

qgallouedec: There isn't lambda in the Eq 6

docs/source/simpo_trainer.mdx
```diff
@@ -0,0 +1,90 @@
...
+)
+
+simpo_trainer = SimPOTrainer(
+    model,
+    args=simpo_config,
+    train_dataset=train_dataset,
+    tokenizer=tokenizer,
+)
+```
+After this one can then call:
```
qgallouedec, line 69:
```suggestion
After this, one can then call:
```

examples/scripts/simpo.py
```diff
@@ -0,0 +1,122 @@
...
+if __name__ == "__main__":
+    parser = HfArgumentParser((ScriptArguments, SimPOConfig, ModelConfig))
+    args, simpo_args, model_config = parser.parse_args_into_dataclasses()
+
+    ################
+    # Model & Tokenizer
+    ################
+    model = AutoModelForCausalLM.from_pretrained(model_config.model_name_or_path)
+    peft_config = get_peft_config(model_config)
+    tokenizer = AutoTokenizer.from_pretrained(model_config.model_name_or_path)
```
qgallouedec, line 81:
```suggestion
    model = AutoModelForCausalLM.from_pretrained(
        model_config.model_name_or_path, trust_remote_code=model_config.trust_remote_code
    )
    tokenizer = AutoTokenizer.from_pretrained(
        model_config.model_name_or_path, trust_remote_code=model_config.trust_remote_code
    )
```

- Peft config is not needed here (redefined in trainer)
- Allow `trust_remote_code` arg

trl/trainer/simpo_trainer.py
```diff
@@ -0,0 +1,441 @@
...
+
+if is_peft_available():
+    from peft import PeftModel, get_peft_model, prepare_model_for_kbit_training
+
+
+if is_wandb_available():
+    import wandb
+
+
+class SimPOTrainer(CPOTrainer):
```
qgallouedec, line 57: I'd prefer `SimPOTrainer` to inherit `Trainer`. It's more in line with our single-file philosophy.

qgallouedec: Ah, I see I'm contradicting https://github.com/huggingface/trl/pull/1725#issuecomment-2189147362 in a way. 

What about this:

```python
    def __init__(
        self,
        model: Optional[Union[PreTrainedModel, nn.Module, str]] = None,
        args: Optional[SimPOConfig] = None,
        data_collator: Optional[DataCollator] = None,
        train_dataset: Optional[Dataset] = None,
        eval_dataset: Optional[Union[Dataset, Dict[str, Dataset]]] = None,
        tokenizer: Optional[PreTrainedTokenizerBase] = None,
        model_init: Optional[Callable[[], PreTrainedModel]] = None,
        callbacks: Optional[List[TrainerCallback]] = None,
        optimizers: Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
        peft_config: Optional[Dict] = None,
        compute_metrics: Optional[Callable[[EvalLoopOutput], Dict]] = None,
    ):
    super().__init__(
        mode=model,
        args=args
        data_collator=data_collator
        train_dataset=train_dataset
        eval_dataset=eval_dataset
        tokenizer=tokenizer
        model_init=model_init
        callbacks=callbacks
        optimizers=optimizers
        preprocess_logits_for_metrics=preprocess_logits_for_metrics
        peft_config=peft_config
        compute_metrics=compute_metrics
    )
    self.gamma_beta_ratio = args.gamma_beta_ratio
    self.sft_weight = args.sft_weight
```

Instead of the full init?

yumeng5:
Hi @qgallouedec

Thanks again for your help with reviewing our PR! Let me know if any further revision is needed!

Thanks,
Yu

trl/trainer/simpo_trainer.py
qgallouedec: Your comment makes sense. At this point, I see two solutions that both seem good to me:
- Rewrite a complete trainer
- Make a very light subclass

I feel we're in the middle of these two solutions here. So I think that even if the error messages are misleading (CPO trainer instead of SimPO), this won't be a problem for the user.

docs/source/simpo_trainer.mdx
```diff
@@ -0,0 +1,90 @@
+# SimPO Trainer
+
+[Simple Preference Optimization](https://arxiv.org/abs/2405.14734) (SimPO) by Yu Meng, Mengzhou Xia, and Danqi Chen proposes a simpler and more effective preference optimization algorithm than DPO without using a reference model. The key designs in SimPO are (1) using length-normalized log likelihood as the implicit reward, and (2) incorporating a target reward margin in the Bradley-Terry ranking objective.
```
qgallouedec, line 3:
```suggestion
[Simple Preference Optimization](https://huggingface.co/papers/2405.14734) (SimPO) by Yu Meng, Mengzhou Xia, and Danqi Chen proposes a simpler and more effective preference optimization algorithm than DPO without using a reference model. The key designs in SimPO are (1) using length-normalized log likelihood as the implicit reward, and (2) incorporating a target reward margin in the Bradley-Terry ranking objective.
```

qgallouedec:
Hey @yumeng5  I made some new remarks, and there are still the previous ones to address. Don't hesitate to ping me if you need help.

yumeng5:
Hi @qgallouedec

Thanks a lot for your comprehensive review! Following your suggestion, I'm updating the SimPO trainer to be a complete one, which also follows the current setup for other trl trainers. I think I have committed all the changes you made previously. Let me know if they look good to you!

Thanks,
Yu

cchenv:
Hi @yumeng5 thanks for adding your implementation to TRL! I just have a few quick questions. 
1. The existing implementation uses CPOTrainer for training SimPO models. However, by using this approach, I was not able to reproduce those model checkpoints or metrics linked on your Princeton Github: https://github.com/princeton-nlp/SimPO/tree/main  Even though I tried to use the same hyperparameters and dataset, the trained models are so much worse, e.g., they produce gibberish quite frequently. I'm wondering if this PR is a better reflection of your original work.
2. In your original implementation, there is a special preprocessing step by keeping only the `model/assistant` message in the `chosen/rejected` field: https://github.com/princeton-nlp/SimPO/blob/main/scripts/run_simpo.py#L89-L99. I also see this is the case for this PR. So I'm wondering if the users need to provide their own preprocessing given a dataset where the `chosen` and `rejected` fields contain both user (i.e., prompt) and assistant messages (e.g., https://huggingface.co/datasets/trl-lib/ultrafeedback_binarized?row=0)?

image

qgallouedec:
I'm closing this PR because no recent activity. Please feel free to reopen it or request a reopening if necessary.

PR #1729, opened 2024-06-13
rloo and ppov2 trainer with trainer callbacks

mnoukhov:
I'm proposing to add trainer callbacks to the rloo and ppov2 trainer 

Added printer callback, saving callback
Correctly set global_step to number of updates
Log `episode` on top of global step

Finally out of draft @vwxyzjn


vwxyzjn (approved):
LGTM.

PR #1731, opened 2024-06-13
Add simplified version of BCO loss

Trangle:
The BCO algorithm increases the complexity of the model and data due to the need to estimate the feedback distribution of samples by passing in embeddings. By observation, this probability can be achieved by adding auxiliary loss constraints. The observation of actual training performance is very effective.


seanexp:
@Trangle 

Looks interesting, could you share the experimental results?

PR #1734, opened 2024-06-14
small KTO fixes

kawine:
- remove the "kto_pair" option from DPOTrainer (this is an older version of KTO has been removed from the paper; only the version in KTOTrainer should be used)
- speed up data processing for calculating suggested lambda weights by avoiding full passes through the dataset

cc @younesbelkada


trl/trainer/dpo_config.py
```diff
@@ -29,8 +29,12 @@ class DPOConfig(TrainingArguments):
             The robust DPO label smoothing parameter from the [cDPO](https://ericmitchell.ai/cdpo.pdf) report and [Robust DPO](https://arxiv.org/abs/2403.00409) paper that should be between 0 and 0.5.
         loss_type (`str`, defaults to `"sigmoid"`):
             The type of DPO loss to use. Either `"sigmoid"` the default DPO loss,`"hinge"` loss from [SLiC](https://arxiv.org/abs/2305.10425) paper, `"ipo"` from [IPO](https://arxiv.org/abs/2310.12036) paper,
+<<<<<<< HEAD
```
younesbelkada, line 32: this needs to be removed

trl/trainer/dpo_config.py
```diff
@@ -29,8 +29,12 @@ class DPOConfig(TrainingArguments):
...
+            `"bco_pair"` from [BCO](https://arxiv.org/abs/2404.04656) paper or `"robust"` from [Robust DPO](https://arxiv.org/abs/2403.00409) paper.
+=======
```
younesbelkada, line 34:
```suggestion
```

trl/trainer/dpo_config.py
```diff
@@ -29,8 +29,12 @@ class DPOConfig(TrainingArguments):
...
             `"kto_pair"` from the HALOs [report](https://github.com/ContextualAI/HALOs/blob/main/assets/report.pdf), `"bco_pair"` from [BCO](https://arxiv.org/abs/2404.04656) paper or `"robust"` from [Robust DPO](https://arxiv.org/abs/2403.00409) paper,
             "aot" and "aot_pair" from alignment via optimal transport
+>>>>>>> f5168fdbaf9cbf6a3f1bdc64dc44b9db3a9ae333
```
younesbelkada, line 37:
```suggestion
```

trl/trainer/dpo_config.py
```diff
@@ -79,7 +83,7 @@ class DPOConfig(TrainingArguments):
     beta: float = 0.1
     label_smoothing: float = 0
     loss_type: Literal[
-        "sigmoid", "hinge", "ipo", "kto_pair", "bco_pair", "sppo_hard", "nca_pair", "robust", "aot", "aot_pair"
+        "sigmoid", "hinge", "ipo", "bco_pair", "sppo_hard", "nca_pair", "robust", "aot", "aot_pair"
     ] = "sigmoid"
```
younesbelkada, line 87: Can you add a logic below that raises an error if a user passes `kto_pair` and tell them that this is not supported anymore?

trl/trainer/cpo_config.py
```diff
@@ -66,7 +66,7 @@ class CPOConfig(TrainingArguments):
 
     beta: float = 0.1
     label_smoothing: float = 0
-    loss_type: Literal["sigmoid", "hinge", "ipo", "kto_pair", "simpo"] = "sigmoid"
+    loss_type: Literal["sigmoid", "hinge", "ipo", "simpo"] = "sigmoid"
```
younesbelkada, line 69: Same here, if you can add an error here and educate users on what to do to replace `kto_pair` which is not supported anymore would be perfect 🙏

trl/trainer/kto_trainer.py
```diff
@@ -639,24 +639,15 @@ def make_inputs_require_grad(module, input, output):
                 # merge the datasets
                 eval_dataset = concatenate_datasets([eval_dataset, eval_kl_dataset], axis=1)
 
-            desirable = train_dataset.filter(
-                lambda x: x["label"], num_proc=args.dataset_num_proc, desc="Filtering desirable examples"
-            )
-            undesirable = train_dataset.filter(
-                lambda x: not x["label"], num_proc=args.dataset_num_proc, desc="Filtering undesirable examples"
-            )
+            num_desirable = max(sum(train_dataset["label"]), 1)
```
younesbelkada, line 642: Nice fix !

younesbelkada (commented):
Thanks a lot for this clean work ! I left one suggestion to ensure users are properly aware on what to do in case they still use `kto_pair`, what do you think?

kawine:
@younesbelkada thanks for taking a look! fixed the merge issue in the files and added some logic to handle `kto_pair` being passed as an argument

younesbelkada (approved):
Thanks a lot !

kashif (approved):
thanks great catch!

PR #1739, opened 2024-06-14
better trl parser with yaml config

### LINKED ISSUE #1733: TRLParser needs changes, overwrites command line arguments with config
TRLParser allows parsing a yaml config and command line arguments. Logically, the command line arguments should override the yaml config so if `config.yaml` contains `gradient_accumulation_steps: 4` and the user calls `python script.py --config config.yaml --gradient_accumulation_steps 1` then we would expect `gradient_accumulation_steps` to be set to `1`.

But currently TRLParser will set it to 4, choosing the config over the user's input.

This is because merging the two dataclasses happens in the YAMLParser which takes in the command line parsed into dataclasses as well as the yaml parsed into a config. But to decide whether to override the dataclass with the config, it checks only if the default value has changed https://github.com/huggingface/trl/blob/main/trl/commands/cli_utils.py#L90

If the user specifies the default value on the command line e.g. `--gradient_accumulation_steps 1` then this is indistinguishable from the dataclass having a default value of `--gradient_accumulation_steps 1`. 

The simplest way to make the config parser and argument parser work together is to manually parse the yaml file and prepend it to the args somehow before calling, like HfArgumentParser does with its config file https://github.com/huggingface/transformers/blob/v4.41.2/src/transformers/hf_argparser.py#L331

The best way, in my opinion, is just to parse the yaml file and then use it to set the defaults of all the dataclasses' with `self.set_defaults` https://github.com/python/cpython/blob/3.12/Lib/argparse.py#L1427

Advice and pointers welcome!

mnoukhov:
- correctly overrides yaml config with command line arguments
do this by changing the defaults to values in the yaml so we get dataclass default < yaml config < command line arguments
 
- adds return_remaining_strings
- when return_remaining_strings is False, raises error if yaml contains extra args that are not in the dataclasses
- removes the need to have a `config` field in one of your dataclasses (I can revert this if this is preferred)
- simpler and cleaner than previous yaml parsing without the need for merging dataclasses 

Fixes #1733

@younesbelkada let me know if you have comments or suggestions!

### DIFF
--- a/trl/commands/cli_utils.py
+++ b/trl/commands/cli_utils.py
@@ -183,7 +183,7 @@ class ChatArguments:
     use_bnb_nested_quant: bool = field(default=False, metadata={"help": "use nested quantization"})
 
 
-class TRLParser(HfArgumentParser):
+class TrlParser(HfArgumentParser):
     def __init__(self, parsers):
         """
         The TRL parser parses a list of parsers (TrainingArguments, trl.ModelConfig, etc.), creates a config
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
This approach looks great to me ! Thanks a lot for the detail report and walk-through of the solution. Tried it locally and this is great to me and simpler than the previous approach
Failing CI is unrelated and trying to solve it in https://github.com/huggingface/trl/pull/1747

PR #1742, opened 2024-06-15
Support for returning past_key_values from the model

idanshen:
The current implementation allows the passing of a KV cache into the forward functions of the Q\Value models but does not get the KV cache of such a call. This limits the ability to use the KV cache with these models.


trl/models/modeling_value_head.py
```diff
@@ -143,6 +143,7 @@ def forward(
         self,
         input_ids=None,
         past_key_values=None,
+        return_past_key_values=False,
```
younesbelkada, line 146: Can you put that at the end (before kwargs)?

trl/models/modeling_value_head.py
```diff
@@ -405,6 +410,7 @@ def forward(
...
```
younesbelkada, line 413: same comment here

younesbelkada (commented):
Thanks for extending the abilities of the value head modeling classes to output ! Just one comment about args ordering

trl/models/modeling_value_head.py
idanshen: Done

trl/models/modeling_value_head.py
idanshen: Done

idanshen:
I change the order of variables, and synced the branch with main

younesbelkada (approved):
Thanks ! Failing tests are unrelated and will be fixed in https://github.com/huggingface/trl/pull/1750

PR #1747, opened 2024-06-17
CI / core: Pin `numpy` to `!=2.0.0` for CI and to users

younesbelkada:
TRL has an optional dependency with DeepSpeed which is right now not compatible with numpy >= 2.0.0

Related: https://github.com/microsoft/DeepSpeed/issues/5671

cc @vwxyzjn


tests/test_best_of_n_sampler.py
```diff
@@ -23,6 +23,7 @@ class BestOfNSamplerTester(unittest.TestCase):
     tokenizer = AutoTokenizer.from_pretrained(ref_model_name)
     tokenizer.pad_token = tokenizer.eos_token
     output_length_sampler = LengthSampler(2, 6)
+    dummy = None
```
younesbelkada, line 26:
```suggestion
```

PR #1748, opened 2024-06-17
`TrlParser`: Add ignore extra args option

younesbelkada:
https://github.com/huggingface/trl/pull/1739 introduced raising errors if users passed extra un-used arguments in the config file. Before that PR we were not raising errors and some users reported that it worked before on their setup (and might be convenient for some use-cases) we should re-enable that through a simple API `ignore_extra_args=True` when creating the `TrlParser`

cc @philschmid @mnoukhov @vwxyzjn

### DIFF
--- a/trl/commands/cli_utils.py
+++ b/trl/commands/cli_utils.py
@@ -199,6 +199,7 @@ class TrlParser(HfArgumentParser):
         """
         super().__init__(parsers)
         self.yaml_parser = YamlConfigParser()
+        self.ignore_extra_args = ignore_extra_args
 
     def post_process_dataclasses(self, dataclasses):
         # Apply additional post-processing in case some arguments needs a special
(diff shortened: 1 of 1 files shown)


philschmid (approved):

trl/commands/cli_utils.py
```diff
@@ -193,6 +193,9 @@ def __init__(self, parsers):
         Args:
             parsers (`List[argparse.ArgumentParser`]):
                 List of parsers.
+            ignore_extra_args (`bool`):
+                Whether to ignore extra arguments passed by the config
+                and not raise errors.
         """
         super().__init__(parsers)
         self.yaml_parser = YamlConfigParser()
```
younesbelkada, line 201:
```suggestion
        self.yaml_parser = YamlConfigParser()
        self.ignore_extra_args = ignore_extra_args
```

vwxyzjn (approved):

PR #1750, opened 2024-06-17
CI / `KTOTrainer`: Remove old tests

younesbelkada:
https://github.com/huggingface/trl/pull/1734 changed the logic for handling non desirable outputs from the KTOTrainer. Therefore we need to remove the corresponding tests to make the CI happy

cc @vwxyzjn @kashif


vwxyzjn:
The DPO trainer tests seem to fail for some reason. Maybe @kashif can take a look?

image

kashif (approved):

Issue #1752, opened 2024-06-18
Question about apply_chat_template in examples


EganGu:
When I looked at the examples I found that the example script for DPO uses `apply_chat_template` for `chosen` and `rejected` but not for `prompt`.
https://github.com/huggingface/trl/blob/d1ed730ab8281b1b0c78d7d61bc0f6603a9ce958/examples/scripts/dpo.py#L150-L152
And it seems that `chosen` is a complete conversation. 
```
[ { "content": "Hi, I want to learn to play horseshoes. Can you teach me?", "role": "user" }, { "content": "I can, but maybe I should begin by telling you that a typical game consists of 2 players and 6 or 8 horseshoes.", "role": "assistant" }, { "content": "Okay. What else is needed to play, and what are the rules?", "role": "user" }, { "content": "A horseshoe is usually made out of metal and is about 3 to 3.5 inches long and around 1 inch thick. The horseshoe should also have a 2 inch by 3 inch flat at the bottom where the rubber meets the metal. We also need two stakes and six horseshoes.", "role": "assistant" } ]
```
I think that using chat_template for the input prompt and only remaining the `assistant` output as `chosen`/`rejected` will be consistent with the inference phase.

AIR-hl:
> When I looked at the examples I found that the example script for DPO uses `apply_chat_template` for `chosen` and `rejected` but not for `prompt`.
> 
> https://github.com/huggingface/trl/blob/d1ed730ab8281b1b0c78d7d61bc0f6603a9ce958/examples/scripts/dpo.py#L150-L152
> 
> 
> And it seems that `chosen` is a complete conversation.
> ```
> [ { "content": "Hi, I want to learn to play horseshoes. Can you teach me?", "role": "user" }, { "content": "I can, but maybe I should begin by telling you that a typical game consists of 2 players and 6 or 8 horseshoes.", "role": "assistant" }, { "content": "Okay. What else is needed to play, and what are the rules?", "role": "user" }, { "content": "A horseshoe is usually made out of metal and is about 3 to 3.5 inches long and around 1 inch thick. The horseshoe should also have a 2 inch by 3 inch flat at the bottom where the rubber meets the metal. We also need two stakes and six horseshoes.", "role": "assistant" } ]
> ```
> 
> I think that using chat_template for the input prompt and only remaining the `assistant` output as `chosen`/`rejected` will be consistent with the inference phase.

[...]

Issue #1541, opened 2024-04-15
Bug in example DPO script in dataloading


sohrabi1:
Since the example [DPO script](https://github.com/huggingface/trl/blob/main/examples/scripts/dpo.py) uses [hh-rlhf dataset](https://huggingface.co/datasets/trl-internal-testing/hh-rlhf-trl-style) in OpenAI messages format, the loading in the script [here](https://github.com/huggingface/trl/blob/e823458a6a793778b959a1c134cd2ee3eaa9a9bd/examples/scripts/dpo.py#L148) seems incorrect:

```
    def process(row):
        row["chosen"] = tokenizer.apply_chat_template(row["chosen"], tokenize=False)
        row["rejected"] = tokenizer.apply_chat_template(row["rejected"], tokenize=False)
        return row
```

since it adds all messages to both chosen and rejected. But it also ignores the prompt template for the prompt.
If my understanding is correct the right process function would be

```
    def process(row):
        # we should extract the final turn of messages to define chosen/rejected responses and keep the rest as prompt
        prompt_messages = row["chosen"][:-1]
        chosen_messages = row["chosen"][-1:]
        rejected_messages = row["rejected"][-1:]

        row["prompt"] = tokenizer.apply_chat_template(prompt_messages, tokenize=False)
        row["chosen"] = tokenizer.apply_chat_template(chosen_messages, tokenize=False)
        row["rejected"] = tokenizer.apply_chat_template(rejected_messages, tokenize=False)
        return row
```
As far as i see only the answer is expected in the chosen / rejected parts in the DPO trainer.

AIR-hl:
> Since the example [DPO script](https://github.com/huggingface/trl/blob/main/examples/scripts/dpo.py) uses [hh-rlhf dataset](https://huggingface.co/datasets/trl-internal-testing/hh-rlhf-trl-style) in OpenAI messages format, the loading in the script [here](https://github.com/huggingface/trl/blob/e823458a6a793778b959a1c134cd2ee3eaa9a9bd/examples/scripts/dpo.py#L148) seems incorrect:
> 
> ```
>     def process(row):
>         row["chosen"] = tokenizer.apply_chat_template(row["chosen"], tokenize=False)
>         row["rejected"] = tokenizer.apply_chat_template(row["rejected"], tokenize=False)
>         return row
> ```
> 
> since it adds all messages to both chosen and rejected. But it also ignores the prompt template for the prompt. If my understanding is correct the right process function would be
> 
> ```
[...]

PR #1753, opened 2024-06-18
change the `process` function in the example of DPO

AIR-hl:
#1752 #1541


younesbelkada:
cc @kashif

kashif (approved):
thanks! nice catch

AIR-hl:
> thanks! nice catch

@kashif Hi! I just modify the codes of dpo example, but there seems to be an error related to sft in checks
image

vwxyzjn:
LGTM.

PR #1757, opened 2024-06-20
New sentiment and descriptiveness dataset

vwxyzjn:
Slightly more organized and uniform. Here, we assume the dataset always has a train and test split.

### DIFF
--- a/examples/datasets/sentiment_descriptiveness.py
+++ b/examples/datasets/sentiment_descriptiveness.py
@@ -1,9 +1,7 @@
-import multiprocessing
 import sys
 from dataclasses import dataclass, field
 from typing import Optional
 
-import numpy as np
 from datasets import Dataset, DatasetDict
 from huggingface_hub import HfApi, hf_hub_download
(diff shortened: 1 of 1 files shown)


kashif (approved):

examples/datasets/sentiment_descriptiveness.py
```diff
@@ -0,0 +1,189 @@
...
+        for i in range(len(row["best"])):
+            best_idx = row["best"][i]
+            chosen_sample = row[f"sample{best_idx}"][i].strip()
+            row["chosen"].append(
+                [
+                    {"role": "user", "content": row["prompt"][i].strip()},
+                    {"role": "assistant", "content": chosen_sample},
+                ]
+            )
+            # find the first rejected sample which is different from the chosen one
```
vwxyzjn, line 129: Note that this is tricky because for some examples we do see some completions being the same.

vwxyzjn:
Thanks @kashif! Merging the PR now :)

Issue #1781, opened 2024-06-27
Conflict in start index under `batched_forward_pass`


mertsayar8:
The code and the comment do not align in:
https://github.com/huggingface/trl/blob/b68ff96f0c74368961e194081e122959cd1f4d4d/trl/trainer/ppo_trainer.py#L1032

In the comment, it is stated that logprobs starts from the second query token, however `start` starts from the last query token in the code. Everything outside `[last_query_token, last_response_token)` are masked out in:
https://github.com/huggingface/trl/blob/b68ff96f0c74368961e194081e122959cd1f4d4d/trl/trainer/ppo_trainer.py#L1037-L1038

This causes a problem especially when using `response_masks` in `step` function. `response_mask` is multiplied with wrong indices in:
https://github.com/huggingface/trl/blob/b68ff96f0c74368961e194081e122959cd1f4d4d/trl/trainer/ppo_trainer.py#L1040
and masking is done wrongly.

I believe `start` should be basicly set to the first response token which is the length of the query. This will make `masks[j, start:end]` and `response_masks_batch[j]` correspond to the same indices and solve this problem.

qgallouedec:
See discussion in #1782

PR #1782, opened 2024-06-27
Fix `start` index under `batched_forward_pass`

mertsayar8:
Start index under `batched_forward_pass` starts from the last query token which does not align with the comment in line 1032. This also causes a problem when handling the response tokens as stated in #1781.

Set `start` to the first response token instead of the last query token.


vwxyzjn:
@mertsayar8 btw we are now recommending to use the PPOv2Trainer :)

qgallouedec:
Hey, thank you for the PR, sorry for the late reply.
I'm not sure about this. Let's consider the simple following example:
The input is `[1, 2, 3, 4, 5, 6, 7]`, and you predict  `[2, 3, 4, 5, 6, 7, 8]`. So you want a mask like `[0, 0, 1, 1, 1, 1, 0]

```python
import torch

query_batch = [torch.tensor([1, 2, 3])]
response_batch = [torch.tensor([4, 5, 6, 7])]
attention_mask = torch.tensor([[1, 1, 1, 1, 1, 1, 1]])

# code from PPOTrainer
masks = torch.zeros_like(attention_mask)
masks[:, :-1] = attention_mask[:, 1:]

for j in range(len(query_batch)):
    # ...
    start = len(query_batch[j])  # logprobs starts from the second query token
    if attention_mask[j, 0] == 0:  # offset left padding
        start += attention_mask[j, :].nonzero()[0]
    end = start + len(response_batch[j])

masks[j, :start] = 0
masks[j, end:] = 0

print(masks)  # tensor([[0, 0, 0, 1, 1, 1, 0]])
```

While we rather expect a mask `tensor([[0, 0, 1, 1, 1, 1, 0]])`

Am I missing something?

PR #1787, opened 2024-06-28
[SFT] add model_init_kwargs to training_args

kashif:
Fix `SFTTrainer` warning by moving `model_init_kwargs` to the `SFTConfig`


vwxyzjn (approved):
LGTM. Are there other scripts with similar `model_kwargs`?

kashif:
@vwxyzjn this is the only example with this warning

PR #1791, opened 2024-07-01
Clean examples

qgallouedec:
- Remove redundant `peft_config`
- Use `image.thumbnail` instead of re-implementing it


kashif (approved):

vwxyzjn (commented):
cpo.py:
```
```

dpo.py
```
peft_config=peft_config,
```

qgallouedec:
Yes, it's because in some cases you need to know if the `peft_config` is None before instantiating the trainer:

```python
peft_config = get_peft_config(model_config)
if peft_config is not None:
    ...

trainer = DPOTrainer(
    ...
    peft_config=peft_config,
)
```

and sometime you don't need peft config until the trainer is instantiated (like in cpo):


```python
trainer = DPOTrainer(
    ...
)
```

For uniformity's sake, we could set the `peft_config` systematically before instantiating the trainer, but I find that a little less readable. But it's a detail and I'm fine with the solution.

vwxyzjn:
I see. Either way make sense to me then. Feel free to choose either one and merge :)

qgallouedec:
I would keep it like this. Thanks for the review. Can a maintainer merge? I can't.

PR #1794, opened 2024-07-01
Bugfix: Preserve token fields when converting TrainingArguments to SFTConfig

noahlt:
## What does this PR do?

This fixes a bug introduced in #1707, where if you create a `transformers.TrainingArguments` and pass it into `SFTTrainer`, you end up with `SFTTrainer.args.hub_token = "<PUSH_TO_HUB_TOKEN>"`, breaking the push-to-Hub functionality.

## Explanation

`TrainingArguments.to_dict()` redacts token fields ([transformers/src/transformers/training_args.py:2417](https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L2417)), so we have to individually copy them over when converting to SFTConfig to avoid breaking push_to_hub functionality.

To be clear, users can work around this bug by directly creating an `SFTConfig` rather than creating a `TrainingArguments` object, but strangely some tutorials mix and match them, so maintaining compatibility seems prudent.

## Other notes

This PR also adds a test. I was able to run the test and verify this satisfies the test line I added, however, I couldn't figure out how to run the full tests without having a `wandb` account configured.

This is my first contribution to this repo, so apologies in advance if I missed something. Happy to revise or make any changes.

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -148,9 +148,7 @@ class SFTTrainer(Trainer):
         elif args is not None and args.__class__.__name__ == "TrainingArguments":
             args_as_dict = args.to_dict()
             # Manually copy token values as TrainingArguments.to_dict() redacts them
-            for k in args_as_dict.keys():
-                if k.endswith("_token"):
-                    args_as_dict[k] = getattr(args, k)
+            args_as_dict.update({k: getattr(args, k) for k in args_as_dict.keys() if k.endswith("_token")})
             args = SFTConfig(**args_as_dict)
 
         if model_init_kwargs is not None:
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -146,7 +146,11 @@ class SFTTrainer(Trainer):
             warnings.warn(f"No `SFTConfig` passed, using `output_dir={output_dir}`.")
             args = SFTConfig(output_dir=output_dir)
         elif args is not None and args.__class__.__name__ == "TrainingArguments":
-            args = SFTConfig(**args.to_dict())
+            args_as_dict = args.to_dict()
+            # Manually copy token values as TrainingArguments.to_dict() redacts them
+            args_as_dict['hub_token'] = args.hub_token
+            args_as_dict['push_to_hub_token'] = args.push_to_hub_token
+            args = SFTConfig(**args_as_dict)
 
         if model_init_kwargs is not None:
             warnings.warn(
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -148,7 +148,9 @@ class SFTTrainer(Trainer):
         elif args is not None and args.__class__.__name__ == "TrainingArguments":
             args_as_dict = args.to_dict()
             # Manually copy token values as TrainingArguments.to_dict() redacts them
-            args_as_dict |= {k: getattr(args, k) for k in args_as_dict.keys() if k.endswith("_token")}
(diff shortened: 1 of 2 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -146,7 +146,11 @@ def __init__(
             warnings.warn(f"No `SFTConfig` passed, using `output_dir={output_dir}`.")
             args = SFTConfig(output_dir=output_dir)
         elif args is not None and args.__class__.__name__ == "TrainingArguments":
-            args = SFTConfig(**args.to_dict())
+            args_as_dict = args.to_dict()
+            # Manually copy token values as TrainingArguments.to_dict() redacts them
+            args_as_dict["hub_token"] = args.hub_token
+            args_as_dict["push_to_hub_token"] = args.push_to_hub_token
```
kashif, line 152:
```suggestion
            args_as_dict = args.to_dict() | {"hub_token": args.hub_token, "push_to_hub_token": args.push_to_hub_token}
```

noahlt: I made this change, thanks!

noahlt:
@kashif I took your suggestion and changed this to a one-line definition:

```py
# Manually copy token values as TrainingArguments.to_dict() redacts them
args_as_dict = args.to_dict() | {"hub_token": args.hub_token, "push_to_hub_token": args.push_to_hub_token}
args = SFTConfig(**args_as_dict)
```

But then I realized I could write it in the general form, which matches [the behavior in TrainingArguments](https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L2417) by copying all values for keys with the `_token` suffix:

```py
args_as_dict = args.to_dict()
# Manually copy token values as TrainingArguments.to_dict() redacts them
args_as_dict |= {k: getattr(args, k) for k in args_as_dict.keys() if k.endswith("_token")}
args = SFTConfig(**args_as_dict)
```

This is an extra line of code but it seems better to avoid hardcoding the keys `hub_token` and `push_to_hub_token`.

Let me know what you think, and thank you for reviewing!

kashif:
yes the new way looks nicer

kashif (approved):

noahlt:
Hm it turns out that `|=` as a dict operation was only added in Python 3.9. This caused tests to fail on this branch, so I rewrote it again as a boring old for-loop. Let me know what you think.

kashif (approved):

noahlt:
This test failure looks like a network error:

> FAILED tests/test_modeling_value_head.py::Seq2SeqValueHeadModelTester::test_dropout_config - requests.exceptions.ReadTimeout: (ReadTimeoutError("HTTPSConnectionPool(host='huggingface.co', port=443): Read timed out. (read timeout=10)"), '(Request ID: 74b6a05e-8c61-4996-9a55-05d9a1a08e33)')

Not sure whether I should re-run tests or what.

noahlt:
Not sure what's going on with the tests — the passing tests say "succeeded 52 minutes ago" ([example](https://github.com/huggingface/trl/actions/runs/9770324540/job/27000396516?pr=1794)), but the cancelled tests all say "cancelled 9 hours ago" (for example, [tests (3.8, ubuntu-latest)](https://github.com/huggingface/trl/actions/runs/9770324540/job/27000395437?pr=1794)), so it appears they did not get retriggered.

noahlt:
Another ReadTimeout network error 😔

@kashif I assume you're manually retriggering these tests, since I can't find any way to retrigger them myself. Thanks for helping me get this merged, and let me know if there's anything I should do.

PR #1803, opened 2024-07-04
KtoTrainer: BCO improvements

claralp:
I recently experimented quite a bit with the BCO loss type in the KTO Trainer.  
This PR includes some changes that helped me to successfully and effectively run BCO on multiple GPUs (on Azure in this case):
- **When using the BCO loss type there is no need for the KL dataset anymore**. By skipping the creation of the KL dataset we save time and memory.
- **Do not assert both desired and undesired data with a per-device mini batch**. With this it was impossible to train large models that only allow a per-device batch size of 1. Also KTO and BCO are supposed to work with unpaired preference data with different amounts of desired or undesired examples. In my experiments BCO proofed to work well  without this assumption.
- **When checkpointing, also save the `RunningMoments` object which is used in BCO to calculate the $\delta$ mean reward value**. If this is not saved, it will corrupt the whole training process when restarting from a checkpoint. 

Please have a look at these and let me know if you can approve the solutions implemented here or if you prefer other solutions to the issues described above

@kashif @lewtun

### DIFF
--- a/trl/trainer/kto_trainer.py
+++ b/trl/trainer/kto_trainer.py
@@ -1269,24 +1269,37 @@ class KTOTrainer(Trainer):
             The chosen_rewards and rejected_rewards tensors contain the rewards for the chosen and rejected responses, respectively.
             The delta value contains the moving average of all implicit rewards.
         """
-        assert policy_chosen_logps.shape[0] > 0, f"no chosen data at {self.accelerator.local_process_index}"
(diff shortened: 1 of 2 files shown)


kashif:
thanks @claralp just looking again do you think it makes sense to split the BCO and KTO trainers for a cleaner implementation?

claralp:
@kashif that would also make sense. But then some shared functions (e.g. `_tokenize`, `_process_tokens`) need to move to a shared place, maybe trainer/utils.py

kashif:
yeah at least for me these if... else make things confusing...

PR #1806, opened 2024-07-05
Consistent use of trust_remote_code

qgallouedec:
Trust remote code should also be used for tokenisers and processor. The easiest/most consistent way to go is to always use `trust_remote_code` from model config.


vwxyzjn (approved):
Nice cleanup! LGTM!

PR #1754, opened 2024-06-18
Issue #1751 Fix

yash-srivastava19:
#1751 mentioned that the TRL CLI is not completely capturing the torch_dtype. I thought the issue was urgent, so quickly patched a hacky fix, which at least initiates the SFT Trainer. 

Original Issue : 
On running the following command :
```
trl sft --model_name_or_path=facebook/opt-125m --dataset_name=imdb  --dataset_text_field=text --max_steps=1 --torch_dtype=bfloat16 --output_dir=./test
```
The error was that trl sft is does not identify it as a string when calling `getattr(torch, model_init_kwargs["torch_dtype"])`.

The fix was made which allows the to not break the pipeline at this stage. Although it is a hacky fix, I'm willing to work on it further :)

The error after that is from the transformers library that isn't able to serialize the dtype object(screenshot attached):

```
...
...
TypeError: Object of type dtype is not JSON serializable
Traceback (most recent call last):
```   

Screenshot 2024-06-18 171826

PR #1807, opened 2024-07-05
Fix `torch_dtype` handling in `{DPO,SFT}Trainer` when provided via CLI

### LINKED ISSUE #1751: [BUG] TRL CLI not capturing `torch_dtype` correctly
## Description

Hi here! 🤗 Apparently the TRL's CLI command `trl sft` is not properly capturing the value provided to the `--torch_dtype` flag as it does not identify it as a string when calling `getattr(torch, model_init_kwargs["torch_dtype"])`.

Most likely this issue happens on the rest of the implementations since when parsing the `torch_dtype` provided by the CLI the conversion to `torch.dtype` happens, and then the `SFTTrainer` in this case, receives the `torch_dtype=torch.bfloat16` instead, and attempts to `getattr(torch, torch.bfloat16)`.

So there's two potential fixes:

* Handling the received type for `torch_dtype` within each `...Trainer` subclass so as to provide it to the `model_init_kwargs` as it without the need of calling `getattr(torch, ...)`
* Respecting the `torch_dtype` as a string and letting each `...Trainer` subclass do the `str` -> `torch.dtype` conversion instead, which is more convenient IMO

## To reproduce

```bash
trl sft --model_name_or_path=facebook/opt-125m --dataset_name=imdb  --dataset_text_field=text --max_steps=1 --torch_dtype=bfloat16 --output_dir=./test
```

alvarobartt:
## Description

This PR fixes an issue with the `torch_dtype` handling when propagated to both `DPOConfig` and `SFTConfig` via the CLI, since their trainers: `DPOTrainer` and `SFTTrainer`, respectively, were only handling the `torch_dtype` value provided via `config.model_init_kwargs["torch_dtype"]` as if it was either `"auto"`, `None`, or a valid `torch.dtype` string.

So on, this PR also adds handling for `torch.dtype` if provided, otherwise it raises a `ValueError`.

Thanks to @yash-srivastava19 for opening a PR aiming to tackle this issue too! #1754

Closes #1751


vwxyzjn (approved):
Thanks @alvarobartt for the very nice PR! The ValueError messages are very nice and so are the test cases. Feel free to merge.

neelsjain:
Did this resolve the `TypeError: Object of type dtype is not JSON serializable` error that was potentially tied this issue? I pulled the latest version of TRL and it seems to still be an issue, but I might be doing something wrong here. Curious to hear if you guys didn't run into the same issue.

kashif:
@neelsjain you get it with which script? your own or the example one?

neelsjain:
I trying to use the Alignment Handbook. I made some changes to handle the arg mismatch between the two repos, but I think what I am seeing is the same as what seen [here] (https://github.com/huggingface/trl/pull/1754#issue-2359688422).

PR #1808, opened 2024-07-05
Fix `TRL_USE_RICH` environment variable handling

### LINKED ISSUE #1800: Update `TRL_USE_RICH` flag for CLI
## Description

Apparently the usage of `rich` when using `trl`'s CLI is currently enforced to `True` (set to `"1"`), meaning that `rich` will always be used.

While this may be a nice enhancement to the CLI display to some, in some cases as logging to an external tracking tool as e.g. GCP's Log Explorer, makes the experience a bit hard since `rich` is not properly handled so no logs are propagated, as the `tqdm` bar is being updated.

So on, how the `TRL_USE_RICH` flag is defined and handled could be improved, so that it can be both enabled and disabled (at the moment it can only be enabled as per https://github.com/huggingface/trl/blob/78f8228874d5cf9c0e68952533cb377202e1eb22/examples/scripts/dpo.py#L58, meaning that when set, no matter what value is is, the condition `if TRL_USE_RICH` will always be true).

## Proposal

- Handle the boolean value for the environment value similarly to https://github.com/python/cpython/blob/48f9d3e3faec5faaa4f7c9849fecd27eae4da213/Lib/distutils/util.py#L308-L321 (already deprecated since Python 3.10, meaning we would need to copy that function into `utils.py`)
- If `TRL_USE_RICH` is not set when running `trl`'s CLI, then set it to `True` by default; otherwise, simply evaluate what's provided using the function mentioned above

alvarobartt:
## Description

This PR fixes how the `TRL_USE_RICH` environment variable value is handled, since at the moment it cannot be disabled.

> [!WARNING]
> I'm not really happy with the current implementation since the `TRL_USE_RICH` environment variable is not used in the core, but it's used in the `examples/scripts` triggered via the CLI, which means that the same function needs to be used more than once and defining that once per file feels weird too. So would love your input here @vwxyzjn on what would be the best approach to ensure that the value is properly handled, but at the same time without adding noise to the codebase 😩 

Closes #1800

### DIFF
--- a/examples/scripts/dpo.py
+++ b/examples/scripts/dpo.py
@@ -55,9 +55,10 @@ import multiprocessing
 import os
 from contextlib import nullcontext
 
-TRL_USE_RICH = os.environ.get("TRL_USE_RICH", False)
-
 from trl.commands.cli_utils import DPOScriptArguments, init_zero_verbose, TrlParser
+from trl.env_utils import strtobool
+
+TRL_USE_RICH = strtobool(os.getenv("TRL_USE_RICH", "0"))
 
 if TRL_USE_RICH:
     init_zero_verbose()
--- a/examples/scripts/sft.py
+++ b/examples/scripts/sft.py
@@ -49,9 +49,10 @@ import logging
 import os
 from contextlib import nullcontext
 
-TRL_USE_RICH = os.environ.get("TRL_USE_RICH", False)
-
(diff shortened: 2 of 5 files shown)


vwxyzjn (approved):
LGTM. 

Regarding the design for the CLI, let's not block this PR and consider possible designs in a separate issue / PR.

kashif (approved):

PR #1809, opened 2024-07-05
Online DPO and Online trainer refactor

vwxyzjn:
This PR supports online DPO per @mnoukhov's PR and also includes refactor on existing online trainers.


trl/trainer/utils.py
```diff
@@ -1199,3 +1235,26 @@ def generate(
...
+        query = queries[i : i + local_rollout_forward_batch_size]
+        query_response, logits = generate(
+            model,
+            query,
+            pad_token_id,
+            generation_config,
+        )
+        query_responses.append(query_response)
+        logitss.append(logits)
+    return torch.cat(query_responses, 0), torch.cat(logitss, 0)
```
vwxyzjn, line 1260: Here I abstracted the `batch_generation` logic into a separate method that the online trainers all use. In the future, we can implement it to support vLLM generation.

mnoukhov:
`online_trainer.py` seems to just be an old copy of RLOOTrainer? Should it be deleted in favour of a future base class?

trl/trainer/online_dpo_trainer.py
```diff
@@ -0,0 +1,566 @@
...
+        )
+        args.micro_batch_size = int(args.per_device_train_batch_size * args.world_size)
+        args.batch_size = int(args.local_batch_size * args.world_size)
+        args.mini_batch_size = exact_div(
+            args.batch_size, args.num_mini_batches, "`batch_size` must be a multiple of `num_mini_batches`"
+        )
+        args.local_mini_batch_size = exact_div(
+            args.local_batch_size, args.num_mini_batches, "`local_batch_size` must be a multiple of `num_mini_batches`"
+        )
+        args.num_updates = args.total_episodes // args.batch_size
```
mnoukhov, line 100: if `batch_size` doesn't exactly divide `total_episodes` then the run doesn't actually go for `total_episodes` and the logic below for `num_train_epochs` (and therefore other things like saving) ends up being wrong. 

I'd suggest changing `num_updates` to something like `num_total_batches` which better reflects its actual meaning and using `exact_div` to make sure it divides correctly. Otherwise, you could support not exact division but this seems like a hassle

trl/trainer/online_dpo_trainer.py
```diff
@@ -0,0 +1,566 @@
...
+
+        #########
+        # setup model, optimizer, and others
+        #########
+        for module in [policy, ref_policy, reward_model]:
+            disable_dropout_in_model(module)
+        if args.stop_token and args.stop_token == "eos":
+            args.stop_token_id = tokenizer.eos_token_id
+        self.model = policy
+        self.create_optimizer_and_scheduler(num_training_steps=args.num_updates)
```
mnoukhov, line 125: followup, this `args.num_updates` should actually be `args.total_num_batches * args.num_mini_batches` whereas currently it is just `args.total_num_batches` I believe

trl/trainer/online_dpo_trainer.py
```diff
@@ -0,0 +1,566 @@
...
+        self.optimizer, self.lr_scheduler = optimizers
+
+        #########
+        # calculate various batch sizes
+        #########
+        if args.total_episodes is None:  # allow the users to define episodes in terms of epochs.
+            args.total_episodes = int(args.num_train_epochs * self.train_dataset_len)
+        accelerator = Accelerator(gradient_accumulation_steps=args.gradient_accumulation_steps)
+        self.accelerator = accelerator
+        args.world_size = accelerator.num_processes
```
mnoukhov, line 88: it feels slightly wrong to make these variables that aren't set by the user as `args` 
- world_size
- local_batch_size
- micro_batch_size
- batch_size
- mini_batch_size
- local_mini_batch_size
- num_updates

maybe change to `self`

trl/trainer/online_dpo_trainer.py
```diff
@@ -0,0 +1,566 @@
...
+            self.sample_generations_freq = max(1, args.num_updates // args.num_sample_generations)
+        self.local_dataloader_batch_size = exact_div(
+            args.local_batch_size,
+            args.num_generation_per_prompt,
+            "`local_batch_size` must be a multiple of `num_generation_per_prompt`",
+        )  # DPO logic: repeats the same prompt args.rloo_k times
+
+        ### DPO stuff
+        self.beta = config.beta
+        self.loss_type = config.loss_type
```
mnoukhov, line 115: change from `config` to `args` for consistency

trl/trainer/online_dpo_trainer.py
```diff
@@ -0,0 +1,566 @@
...
+        args.mini_batch_size = exact_div(
+            args.batch_size, args.num_mini_batches, "`batch_size` must be a multiple of `num_mini_batches`"
+        )
+        args.local_mini_batch_size = exact_div(
+            args.local_batch_size, args.num_mini_batches, "`local_batch_size` must be a multiple of `num_mini_batches`"
+        )
+        args.num_updates = args.total_episodes // args.batch_size
+        time_tensor = torch.tensor(int(time.time()), device=accelerator.device)
+        time_int = broadcast(time_tensor, 0).item()  # avoid different timestamps across processes
+        args.run_name = f"{args.exp_name}__{args.seed}__{time_int}"
```
mnoukhov, line 103: maybe make this `run_name` only if it is None, otherwise keep user-specified `run_name`

trl/trainer/online_dpo_trainer.py
vwxyzjn: That is a good point. However, the main purpose of putting everything in `args` is so that the experiment management (e.g., wandb) could track these. After all, we'd be interested in the global effective batch size `batch_size`.

trl/trainer/online_dpo_trainer.py
vwxyzjn: See above, it's mainly for experiment tracking purposes.

trl/trainer/online_dpo_trainer.py
vwxyzjn: `exact_div` is probably too stringent. What do you think of `math.ceil(args.total_episodes // args.batch_size)`? `num_total_batches` is a nice suggestion.

trl/trainer/online_dpo_trainer.py
vwxyzjn: It's only called at the batch level, so the current impl should be fine. Would you agree?

https://github.com/huggingface/trl/blob/98ad01ddfd1e1b67ec018014b83cba40e0caea66/trl/trainer/ppov2_trainer.py#L273

trl/trainer/online_dpo_trainer.py
mnoukhov: that works but then you need to trim the size of `queries` in the last batch

trl/trainer/online_dpo_trainer.py
mnoukhov: I guess that works. Maybe add a comment that clears it up?

trl/trainer/online_dpo_trainer.py
mnoukhov: That's a fair point. I guess an assertion that those args aren't set would be helpful but not necessary

trl/trainer/online_dpo_trainer.py
mnoukhov: but `run_name` is used as the name for wandb and should be user-settable, otherwise you can't control your own wandb run name

trl/trainer/online_dpo_trainer.py
vwxyzjn: Well, maybe that's too stringent.

trl/trainer/online_dpo_trainer.py
vwxyzjn: Why would you like to control the `run_name`?

trl/trainer/online_dpo_trainer.py
vwxyzjn: added a comment

docs/source/online_dpo_trainer.md
```diff
@@ -0,0 +1,295 @@
...
+    --push_to_hub
+```
+
+1B experiment can be found here:
+
+- [🤗 Model checkpoint](https://huggingface.co/vwxyzjn/ppo_tldr)
+- [🐝 Tracked experiment](https://wandb.ai/huggingface/trl/runs/dd2o3g35)
+
+
+To evaluate, we use vLLM to load the checkpoints and GPT3.5 as a judge model to evaluate the generated TL;DR against the reference TL;DR.
```
qgallouedec, line 228:
```suggestion
To evaluate, we use vLLM to load the checkpoints and GPT4 as a judge model to evaluate the generated TL;DR against the reference TL;DR.
```

qgallouedec (approved):

PR #1822, opened 2024-07-10
Delete unused benchmark.yml workflow

AdnaneKhan:
This workflow file has not been used in over 9 months and in its current iteration has a TOCTOU (if run on forks), so best to remove it.


vwxyzjn (approved):
Thanks for the PR!

PR #1826, opened 2024-07-11
refactor trainer callbacks

kashif:
move trainer callbacks to their own file from utils


vwxyzjn (approved):
@kashif thanks so much for the change!

PR #1838, opened 2024-07-16
on policy files

kevalshah90:
Added SFT training notebook and related files. SFT `Qwen0.5B` (Student) on `Qwen7B` (Teacher) generated completions dataset.


qgallouedec:
Hi, can you explain a bit more what is the need that you're trying to address here? Why do you need a dedicated "onpolicy" directory?

PR #1841, opened 2024-07-17
[SFT] fix neftune_noise_alpha in SFTTrainer

### LINKED ISSUE #1837: Possible bug in SFTTrainer
Hi folks,

I am running into an issue with the SFTTrainer regarding the neftune_noise_alpha parameter.

If I leave this argument empty and let it default to None, there seems to be an error in the code flow on [line 307 of the SFTTrainer]( https://github.com/huggingface/trl/blob/main/trl/trainer/sft_trainer.py#L307). Namely (with added comments):

```
self._trainer_supports_neftune = hasattr(args, "neftune_noise_alpha") # This evaluates to true when the default None is used
if neftune_noise_alpha is not None and self._trainer_supports_neftune: # this gets skipped because of neftune_noise_alpha is None
            args.neftune_noise_alpha = neftune_noise_alpha
            warnings.warn(
                "You passed a `neftune_noise_alpha` argument to the SFTTrainer, the value you passed will override the one in the `SFTConfig`."
            )
            # self.neftune_noise_alpha is done at Trainer level
elif not self._trainer_supports_neftune: # But this never gets executed and thus self.neftune_noise_alpha is never set
            self.neftune_noise_alpha = neftune_noise_alpha
```

Later on in the [train() at line 448](https://github.com/huggingface/trl/blob/main/trl/trainer/sft_trainer.py#L448), the error `AttributeError: 'SFTTrainer' object has no attribute 'neftune_noise_alpha'`

**Possible fix**:
I believe one fix would be to change line 307 from `hasattr(args, "neftune_noise_alpha")` to `hasattr(args, "neftune_noise_alpha") and neftune_noise_alpha is not None`

**Additional info**:
* OS: Ubuntu 22.04.4
* TRL 0.9.6
* Torch 1.13.0+cu116
* Transformers 4.34.1

Thanks and cheers!

kashif:
fixes #1837 by removing it from the SFTTrainer as the super-class has support for it via the TRL requirements for Transformer version


trl/trainer/sft_trainer.py
```diff
@@ -445,14 +442,14 @@ def make_inputs_require_grad(module, input, output):
     @wraps(Trainer.train)
     def train(self, *args, **kwargs):
         # Activate neftune right before training.
-        if self.neftune_noise_alpha is not None and not self._trainer_supports_neftune:
```
qgallouedec, line 448: It's weird, I read: if it doesn't support support netftune, then activate neftune

jjc10: Yes I am a bit confused too. There seems to be some redundant logic when alternating on checks between `neftune_noise_alpha` not being none and `self._trainer_supports_neftune` (which itself depends on `neftune_noise_alpha` earlier). I suppose there was a reason for it but I can't see it.

trl/trainer/sft_trainer.py
```diff
@@ -445,22 +442,15 @@ def make_inputs_require_grad(module, input, output):
...
 
         # After training we make sure to retrieve back the original forward pass method
         # for the embedding layer by removing the forward post hook.
-        if self.neftune_noise_alpha is not None and not self._trainer_supports_neftune:
-            unwrapped_model = unwrap_model(self.model)
-            if is_peft_available() and isinstance(unwrapped_model, PeftModel):
-                embeddings = unwrapped_model.base_model.model.get_input_embeddings()
-            else:
-                embeddings = unwrapped_model.get_input_embeddings()
-
```
jjc10, line 461: I reckon this was dead code that was actually never getting executed since if `netfune_noise_alpha` is set, `self._trainer_supports_netfune` was going to be `True` but assuming the bug is fixed, this code is probably necessary no?

jjc10 (commented):
Thanks for picking this up!

qgallouedec:
@kashif what is the status of this PR?

kashif:
@jjc10  can you kindly test with this fix?

trl/trainer/sft_trainer.py
```diff
@@ -651,7 +655,10 @@ def _trl_activate_neftune(self, model):
...
 
-        embeddings.neftune_noise_alpha = self.neftune_noise_alpha
-        hook_handle = embeddings.register_forward_hook(neftune_post_forward_hook)
-        self.neftune_hook_handle = hook_handle
+        if not hasattr(embeddings, "neftune_noise_alpha"):
+            embeddings.neftune_noise_alpha = self.neftune_noise_alpha
+            hook_handle = embeddings.register_forward_hook(neftune_post_forward_hook)
+            self.neftune_hook_handle = hook_handle
+        else:
+            warnings.warn("NEFTune appears to be already activated. Skipping activation.")
```
lewtun, line 663:
```suggestion
            warnings.warn("NEFTune is already activated. Skipping activation.")
```

lewtun (commented):
Thanks for the fix! I would love to see a regression test that reproduces the original error and shows the changes in this PR pass - WDYT?

qgallouedec (approved):

PR #1856, opened 2024-07-19
Improve judges

qgallouedec:
## Motivation:

Judges aren't always pairwise
Judges may be slow when called with numerous prompts/completions
BaseAPIJudge doesn't provide any simplification

## Solutions

1. Define a new inheritance structure:
- `BaseJudge` for all judges
- `BaseRankJudge(BaseJudge)` for judges that returns ranking
- `BasePairwiseJudge(BaseJudge)` for judges that compare two completions and returns the best index

2. We rename `HuggingFaceJudge` to `HfPairwiseJudge` and `OpenAIJudge` to `OpenAIPairwiseJudge`

3. We use multithreading for api call to speedup inference. Why multithreading instead of asyncio? 

| | Sequential | Multi-threading | asyncio |
|-|-|-|-|
| `InferenceClient` | Slow | Ok | Slow (`AsyncInferenceClient`) |
| `OpenAI` | Slow | Ok | Ok (`AsyncOpenAI`) |

4. Other minor changes in the documentation


trl/trainer/callbacks.py
```diff
@@ -177,7 +177,7 @@ class WinRateCallback(TrainerCallback):
     def __init__(
         self,
         prompts: List[str],
```
edbeeching, line 179: Not from your PR, but I do not think these prompts are ever used. Given the potentially high cost of running this callback, perhaps the user could provide the eval dataset or a subset of it directly, rather than assuming the callback needs to be run on the entire eval dataset.

Example:
```
...
    def __init__(
        self,
        eval_dataset: Dataset,
        judge: BaseJudge,
        trainer: Trainer,
        generation_config: Optional[GenerationConfig] = None,
        batch_size: int = 4,
    ):
        self.generation_config = generation_config
        self.judge = judge
        self.ref_completions = []
        self.trainer = trainer
        self.eval_dataset = eval_dataset
...
```

edbeeching:
Another comment unrelated to this exact PR, but perhaps we can include this change. In the original version of the WinRateCallback, it was possible to use the starting model as the reference directly, without having the following:
```
    if not hasattr(trainer, "ref_model"):
        raise AttributeError("Trainer must have a `ref_model` attribute.")
```
and the `model` rather than `ref_model` was used here:
```
    def on_train_begin(self, args: TrainingArguments, state: TrainerState, control: TrainerControl, **kwargs):
        tokenizer = kwargs["tokenizer"]
        tokenizer.padding_side = "left"
        accelerator = self.trainer.accelerator
        with accelerator.split_between_processes(self.eval_dataset["prompt"], apply_padding=True) as prompts:
            self.ref_completions = self.generate_completions_for_model(self.trainer.ref_model, tokenizer, prompts)
```

This enabled us to use this callback to monitor progress during an SFT, without the need for a ref model.

I believe the result should be the same even in a DPO setting. @qgallouedec are you aware of why this was changed in @lewtun's PR?

edbeeching:
Hey @qgallouedec, regarding asyncio vs multithreading. The original version of judges used threadpools, do you know why this was changed to asyncio in the first place? Perhaps there is a good reason we are not aware of.

qgallouedec:
> why this was changed to asyncio in the first place?

`asyncio` is a more natural choice when handling I/O-bound tasks because it efficiently manages a large number of concurrent operations with lower overhead, avoids the complexities of thread synchronization, and offers better scalability and performance for tasks like API requests.

Also, I wanted to remove `BaseAPIJudge`, so this was the opportunity to question threadpool's choice.

examples/scripts/evals/judge_tldr.py
```diff
@@ -0,0 +1,74 @@
...
+from transformers import HfArgumentParser
+from vllm import LLM, SamplingParams
+
+from trl import HfPairwiseJudge, OpenAIPairwiseJudge
+
+
+"""
+Examples:
+
+python examples/scripts/evals/judge_tldr.py --model_name_or_path vwxyzjn/rloo_tldr--num_examples 1000
```
edbeeching, line 14: missing space here `--model_name_or_path vwxyzjn/rloo_tldr--num_examples 1000`

examples/scripts/evals/judge_tldr.py
```diff
@@ -0,0 +1,74 @@
...
+Model win rate: 31.40%
+
+python examples/scripts/evals/judge_tldr.py --model_name_or_path vwxyzjn/rloo_tldr --judge_model gpt-3.5-turbo-0125 --num_examples 1000
+Model win rate: 51.60%
+
+python examples/scripts/evals/judge_tldr.py --model_name_or_path vwxyzjn/rloo_tldr --judge_model gpt-4o-mini --num_examples 1000
+Model win rate: 51.20%
+
+python examples/scripts/evals/judge_tldr.py --model_name_or_path vwxyzjn/ppo_tldr--num_examples 1000
```
edbeeching, line 23: also missing space here

edbeeching (approved):
Thanks for creating issues for the points I raised. LGTM barring the remaining minor comments and resolving the remaining conflicts.

kashif (approved):

PR #1874, opened 2024-07-25
Skip BigBird save and load test until next transformers version

qgallouedec:
CI fails since https://github.com/huggingface/transformers/commit/6f40a213eb10e38a5f242d0645519d413d32d798

Caused by BigBird Pegasus, which does not support param buffer assignment. Fixed here: https://github.com/huggingface/transformers/pull/32222

Skip the test until we dumb to the next transformers version

### DIFF
--- a/tests/test_modeling_value_head.py
+++ b/tests/test_modeling_value_head.py
@@ -17,7 +17,6 @@ import unittest
 
 import pytest
 import torch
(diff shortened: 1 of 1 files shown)


kashif:
readd when https://github.com/huggingface/transformers/pull/32222 is merged

kashif (approved):

PR #1877, opened 2024-07-25
Support Rank Stabilized LoRA in the ModelConfig/LoraConfig

JohnGiorgi:
Tiny PR to support Rank Stabilized LoRA. To do this we:

- Expose the argument in `ModelConfig`. It is copied straight from https://huggingface.co/docs/peft/en/package_reference/lora#peft.LoraConfig.use_rslora
- Pass it through to `LoraConfig` in `utils.py`

Of course, a user could always do this outside TRL, but my motivation for enabling it this way was that I am working with a slightly modified version of the `sft.py` script, so it would be nice if I could enable Rank Stabilized LoRA with a single flag.


kashif:
@JohnGiorgi  do we need to pin peft to some version for this?

kashif (approved):

JohnGiorgi:
@kashif Good question! Looks like yes... [v0.8.0](https://github.com/huggingface/peft/releases/tag/v0.8.0) based on digging through the release notes

kashif:
yes i believe we will need to pin peft to that for this PR, also could you kindly also add a test?

JohnGiorgi:
@kashif Okay I bumped the peft dependency to >=0.8.0. I also added a test for `get_peft_config` (which looks to be untested), which also serves as a test to ensure `use_rslora` is getting passed from a `ModelConfig` to resulting `LoraConfig`

kashif:
@JohnGiorgi i believe its failing for one of the older python with:
```python
AttributeError: 'str' object has no attribute 'removeprefix'
```

JohnGiorgi:
Fixed!

trl/trainer/model_config.py
```diff
@@ -64,6 +64,17 @@ class ModelConfig:
     lora_task_type: str = field(
         default="CAUSAL_LM", metadata={"help": "The task_type to pass for LoRA (use SEQ_CLS for reward modeling)"}
     )
+    use_rslora: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "When set to True, uses <a href='https://doi.org/10.48550/arXiv.2312.03732'>Rank-Stabilized LoRA</a>"
+                " which sets the adapter scaling factor to `lora_alpha/math.sqrt(r)`, since it"
+                " was proven to work better. Otherwise, it will use the original default"
```
qgallouedec, line 73: > was proven to work better

do you have a ref?

qgallouedec: Ah ok I get it, let me rephrase a bit the documentation.

setup.py
```diff
@@ -71,15 +71,15 @@
 EXTRAS = {
     "test": [
         "parameterized",
+        "peft>=0.8.0"
         "pytest",
         "pytest-xdist",
-        "accelerate",
```
qgallouedec, line 76: already in required

trl/trainer/model_config.py
JohnGiorgi: This was copied from https://huggingface.co/docs/peft/main/en/conceptual_guides/lora verbatim

qgallouedec:
Merged, thank you @JohnGiorgi

JohnGiorgi:
> Merged, thank you @JohnGiorgi

Thank you!

PR #1864, opened 2024-07-23
[online-DPO] online dpo cleanups

kashif:
* renamed some configs 
* removed unused configs
* call reference and reward model with no_grads
* use the dropout flag
* added test

PR #1879, opened 2024-07-26
Small fixes to online dpo example

edbeeching:
#1864 updated configs but missed a couple of changes.


kashif:
ah right.. i thought it was ok to have a different family of reference model, but yeah makes sense to also have the reference model be the original one... approving

kashif (approved):

edbeeching:
> ah right.. i thought it was ok to have a different family of reference model, but yeah makes sense to also have the reference model be the original one... approving

yes also the property `sft_model_path` no longer exists in the config

kashif:
ah i see it exists in the `OnPolicyConfig` and the `OnlineDPOConfig` inherits from it...

PR #1885, opened 2024-07-29
fix dpo_trainer bug for LLMs without bos_token in config

DZ9:
In [dpo_trainer.py](https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L860), the `bos_token` will be  automatically added if the first token in current tokenized sentence is not equal to  `bos_token`. Currently this `bos_token` is read from tokenizer config file, but in some LLMs, like [Qwen2](https://huggingface.co/Qwen/Qwen2-7B-Instruct/blob/main/tokenizer_config.json), `bos_token` is leaving to `None` in config, which results in `None` is added to the input_ids tensor , as shown below:
image

and then the followiing trace will be raised when running dpo:
```
  File "examples/scripts/dpo.py", line 185, in <module>
    trainer.train()
  File "/usr/lib/python3.10/site-packages/transformers/trainer.py", line 1859, in train
    return inner_training_loop(
  File "/usr/lib/python3.10/site-packages/transformers/trainer.py", line 2165, in _inner_training_loop
    for step, inputs in enumerate(epoch_iterator):
  File "/usr/lib/python3.10/site-packages/accelerate/data_loader.py", line 452, in __iter__
  File "/usr/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 631, in __next__
    data = self._next_data()
  File "/usr/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 675, in _next_data
  File "/usr/lib/python3.10/site-packages/torch/utils/data/_utils/fetch.py", line 54, in fetch
    return self.collate_fn(data)
  File "/usr/lib/python3.10/site-packages/trl/trainer/utils.py", line 461, in __call__
    torch.tensor(ex[k], dtype=dtype)
TypeError: 'NoneType' object cannot be interpreted as an integer
  0%|          | 0/25 [00:00<?, ?it/s]  
```
Also, in LLMs like Qwen2, bos_token(`<|endoftext|>`) is not equal to the first template token(`<|im_start|>`). Automatically add this `bos_token` without user awareness will cause unexpected behavior when using the trained dpo model to inference. Actually the `bos_token` should only be added when this value in tokenizer config is not None. 

After this PR, the data is running normally like this:
image

and the model can be trained normally:
image

Test command is:
```
python examples/scripts/dpo.py
  --dataset_name trl-internal-testing/hh-rlhf-helpful-base-trl-style \
  --model_name_or_path Qwen2/Qwen2-0.5B-Instruct \
  --learning_rate 1e-3 \
  --logging_steps 10 \
  --eval_steps 500 \
  --output_dir ./output \
  --warmup_steps 150 \
  --report_to tensorboard \
  --bf16 \
  --logging_first_step \
  --max_length 4096 \
  --sanity_check True
```

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -1163,3 +1163,39 @@ def batch_generation(
         query_responses.append(query_response)
         logitss.append(logits)
     return torch.cat(query_responses, 0), torch.cat(logitss, 0)
+
+
+def add_bos_token_if_needed(
+    bos_token_id: int,
+    prompt_len_input_ids: int,
+    prompt_tokens: Dict[str, List[int]],
+    chosen_prompt_len_input_ids: int,
+    chosen_tokens: Dict[str, List[int]],
+    rejected_prompt_len_input_ids: int,
+    rejected_tokens: Dict[str, List[int]]
+):
+    if bos_token_id is not None:
+        if prompt_len_input_ids == 0 or bos_token_id != prompt_tokens["prompt_input_ids"][0]:
+            prompt_tokens["prompt_input_ids"] = [bos_token_id] + prompt_tokens["prompt_input_ids"]
+            prompt_tokens["prompt_attention_mask"] = [1] + prompt_tokens["prompt_attention_mask"]
+        if chosen_prompt_len_input_ids == 0 or bos_token_id != chosen_tokens["prompt_input_ids"][0]:
+            chosen_tokens["prompt_input_ids"] = [bos_token_id] + chosen_tokens["prompt_input_ids"]
+            chosen_tokens["prompt_attention_mask"] = [1] + chosen_tokens["prompt_attention_mask"]
+        if rejected_prompt_len_input_ids == 0 or bos_token_id != rejected_tokens["prompt_input_ids"][0]:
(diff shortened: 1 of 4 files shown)


kashif (approved):

kashif:
@DZ9 since this snippet is used in a number of other trainers, would it be better to add it as a helper function and then use it in DPO, CPO, ORPO trainers?

DZ9:
@kashif Sure. I've changed the adding of bos and eos token function to utils.py, and applied them in dpo, orpo and cpo trainer.

kashif (approved):

kashif:
thanks @DZ9 can you also run `pre-commit run --all-files` in the root of TRL folder to fix formatting issues

DZ9:
@kashif Absolutely. Done with running the formatting command and commited.

trl/trainer/utils.py
```diff
@@ -1163,3 +1163,37 @@ def batch_generation(
         query_responses.append(query_response)
         logitss.append(logits)
     return torch.cat(query_responses, 0), torch.cat(logitss, 0)
+
+
+def add_bos_token_if_needed(
+    bos_token_id: int,
```
kashif, line 1169:
```suggestion
    bos_token_id: Optional[int],
```

PR #1892, opened 2024-08-01
fix serialization of RunningMoments on multiple GPUs

claralp:
This issue is related to PR #1869  

When training in a distributed setup, the collected count of the RunningMoments class ended up being a tensor, not an integer. This led to errors in json serializtion.  

This fixes the type of xs_count in distributed trainings.  
Previous fix (see https://github.com/huggingface/trl/pull/1869/commits/c22035049d857775ce6ba82763e82b934959064b) caused tests to fail and was therefore reverted by @kashif. Also @seanexp


kashif (approved):

kashif:
thanks @claralp  could you kindly run the pre-commit to fix the formatting?

Issue #1252, opened 2024-01-19
DPOTrainer.tokenize_row is not hashable


normster:
`DPOTrainer.tokenize_row` is not hashable, so the `datasets` library assigns the transformed dataset a "random" fingerprint. This fingerprint relies on the `random` library, but since the random seed is often fixed in training configs this always yields the same fingerprint across different runs even when the underlying dataset is completely different: https://github.com/huggingface/datasets/issues/6604.

The immediate fix should be to change the fingerprinting logic in `datasets` to use a library like `uuid` to guarantee different fingerprints each run. But even after this is fixed, `tokenize_row` not being hashable will trigger re-processing every run, which is slow and wasteful. Is there any way to make it hashable so we can leverage caching?

lvwerra:
Maybe a quick solution could be to change the processing function that's passed to `.map` to a standalone function rather than a class method. cc @kashif

norabelrose:
Bumping this so it's not stale anymore; still a problem

younesbelkada:
cc @kashif what do you think? 🙏

thusinh1969:
Really wasting time too much. Please fix. Takes hours to starf.

Steve

thusinh1969:
BTW. Unsloth run Ok without this problem but single GPU only.

Ghubertang:
I really need this to be fixed. Please. Waiting too long to start a training.

thusinh1969:
Same here...please fix.

Shiguang-Guo:
same problem , and it takes me up to half an hour to process data (whether it is ten thousand or one hundred). I am a little curious about what it is waiting for. Is this the same issue caused by not hashable?

Map (num_proc=32): 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 152583/152583 [32:57<00:00, 77.18 examples/s]
Map (num_proc=32): 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 997/997 [30:33<00:00,  1.84s/ examples]
Map (num_proc=32): 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [30:08<00:00, 18.08s/ examples]

kashif:
refresh! on it currently

Issue #1310, opened 2024-02-01
Error when using multiprocessing in the DPO data preprocessing


imraviagrawal:
dpo_trainer = DPOTrainer(
  File "/usr/local/lib/python3.10/dist-packages/trl/trainer/dpo_trainer.py", line 371, in __init__
    train_dataset = train_dataset.map(self.tokenize_row, num_proc=self.dataset_num_proc)
  File "/usr/local/lib/python3.10/dist-packages/datasets/arrow_dataset.py", line 592, in wrapper
    out: Union["Dataset", "DatasetDict"] = func(self, *args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/datasets/arrow_dataset.py", line 557, in wrapper
    out: Union["Dataset", "DatasetDict"] = func(self, *args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/datasets/arrow_dataset.py", line 3185, in map
    for rank, done, content in iflatmap_unordered(
  File "/usr/local/lib/python3.10/dist-packages/datasets/utils/py_utils.py", line 654, in iflatmap_unordered
    [async_result.get(timeout=0.05) for async_result in async_results]
  File "/usr/local/lib/python3.10/dist-packages/datasets/utils/py_utils.py", line 654, in <listcomp>
    [async_result.get(timeout=0.05) for async_result in async_results]
  File "/usr/local/lib/python3.10/dist-packages/multiprocess/pool.py", line 774, in get
    raise self._value
  File "/usr/local/lib/python3.10/dist-packages/multiprocess/pool.py", line 540, in _handle_tasks
    put(task)
  File "/usr/local/lib/python3.10/dist-packages/multiprocess/connection.py", line 209, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/usr/local/lib/python3.10/dist-packages/multiprocess/reduction.py", line 54, in dumps
    cls(buf, protocol, *args, **kwds).dump(obj)
  File "/usr/local/lib/python3.10/dist-packages/dill/_dill.py", line 418, in dump
    StockPickler.dump(self, obj)
  File "/usr/lib/python3.10/pickle.py", line 487, in dump
    self.save(obj)
  File "/usr/local/lib/python3.10/dist-packages/dill/_dill.py", line 412, in save
    StockPickler.save(self, obj, save_persistent_id)
  File "/usr/lib/python3.10/pickle.py", line 560, in save
[...]

PR #1896, opened 2024-08-03
fix DPOTrainer.tokenize_row is not hashable

GuanghaoYe:
The current implementation of `tokenize_row` in the [TRL repository](https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L789) is a class function, which is not hashable. As a result, the `datasets` library assigns a "random" fingerprint to the transformed dataset, as mentioned in issue #1252. This leads to several issues:

1. Ideally, only the main process should tokenize the data. However, due to the different fingerprints being created, each process ends up tokenizing its own copy.
2. The `datasets.map` function also relies on the fingerprint, causing it to run in a single process regardless of the `dataset_num_proc` setting.
3. Since preprocessing time for large datasets can be very long with a single process, and the default timeout for the NCCL communicator is 30 minutes, it becomes impossible to run `dpo_trainer` with large datasets.

With this PR, these issues are resolved, ensuring proper functionality and efficiency.
Test command is:
```bash
 python examples/scripts/dpo.py \                                                                                              
    --dataset_name=trl-internal-testing/hh-rlhf-helpful-base-trl-style \
    --model_name_or_path=gpt2 \
    --per_device_train_batch_size 4 \
    --learning_rate 1e-3 \
    --logging_steps 10 \
    --eval_steps 500 \
    --output_dir="dpo_anthropic_hh" \
    --warmup_steps 150 \
    --report_to wandb \
    --bf16 \
    --logging_first_step \
    --no_remove_unused_columns \
    --dataset_num_proc 128
```

We run this command for twice, 
the first time, it tokenize the data with 128 processes
Screenshot 2024-08-02 173029
and the second time it directly loads the tokenized data.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -538,7 +538,7 @@ class DPOTrainer(Trainer):
                 truncation_mode=self.truncation_mode,
                 label_pad_token_id=self.label_pad_token_id,
             )
-            tokenize_row_partial = partial(tokenize_row, tokenizer = tokenizer, processor=self.processor if self.is_vision_model else None, args=tokenize_row_config)
+            tokenize_row_partial = partial(tokenize_row, tokenizer = tokenizer, processor=self.processor if self.is_vision_model else None, tokenize_row_config=tokenize_row_config)
             train_dataset = train_dataset.map(tokenize_row_partial, num_proc=self.dataset_num_proc, writer_batch_size=10)
             if eval_dataset is not None:
                 eval_dataset = eval_dataset.map(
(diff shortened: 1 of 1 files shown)


kashif:
thanks @GuanghaoYe also if you check the KTO trainer, it tokenizes the whole batch and also fixes this issue... do you think you can try that strategy for DPO as well?

kashif:
@GuanghaoYe something like #1914 is what i had in mind, what do you think?

qgallouedec:
When merging, please check if this also solves #1310.

kashif:
@GuanghaoYe i will close the PR if that is ok for you

PR #1898, opened 2024-08-03
[Online-DPO] num_generation_per_prompt is fixed

kashif:
* `num_generation_per_prompt` is always 2 for the online DPO case
* removed un-needed `with torch.no_grad():`


qgallouedec:
Important: seems like eos penalty had an error

```diff
- scores = torch.where(contain_eos_token, scores, args.penalty_reward_value)
+ scores = torch.where(contain_eos_token, args.penalty_reward_value, scores)
```

trl/trainer/online_dpo_trainer.py
```diff
@@ -287,14 +288,13 @@ def repeat_generator():
                     del logits, all_logprob
                     torch.cuda.empty_cache()
 
-                    with torch.no_grad():
```
qgallouedec, line 290: it's because ref model is in eval mode that you don't need to use the context?

qgallouedec (approved):

PR #1904, opened 2024-08-06
[RPO] use loss from v3 of paper

### LINKED ISSUE #1868: RPO Loss Inconsistency
In the [RPO paper](https://arxiv.org/abs/2404.19733), the loss is calculated as the sum of the vanilla DPO loss and the NLL loss on the chosen completion, as shown in the image below. Here, the NLL loss is weighted by alpha.

2024-07-24 at 17 10 53

In the implementation of [DPOTrainer](https://github.com/chrisliu298/trl/blob/ea0d3985a04ff9cfccde01d05121dfd04dcb3c14/trl/trainer/dpo_trainer.py#L1387-L1388), the loss is calculated as the DPO loss weighted by alpha, plus the NLL loss:

```python
    losses = losses * self.args.rpo_alpha + policy_nll_loss
```

If I understand correctly, the above formulation is different from the loss function in the official paper. I wonder why the weight alpha is moved to the DPO loss in TRL.

kashif:
fixes #1868


qgallouedec:
https://github.com/huggingface/trl/blob/29bd0046a9e85cc5c4ff80d15322737caa06adfb/docs/source/dpo_trainer.mdx?plain=1#L149

```
SFT loss on the chosen preferences together with a weighted DPO loss
```


to

```
DPO loss together with a weighted SFT loss on the chosen preferences
```

trl/trainer/dpo_config.py
```diff
@@ -89,7 +89,7 @@ class DPOConfig(TrainingArguments):
         ref_model_sync_steps ('int', defaults to 2):
             The tau parameter from the [TR-DPO](https://huggingface.co/papers/2404.09656) paper.
         rpo_alpha ('float', defaults to `None`):
-            The alpha parameter from the [RPO](https://huggingface.co/papers/2404.19733) paper. If None, no weighting is applied and the loss is the same as the DPO loss.
+            The alpha parameter from the [RPO](https://huggingface.co/papers/2404.19733) paper V3. If None, no weighting is applied and the loss is the same as the DPO loss.
```
qgallouedec, line 92:
```suggestion
            The alpha parameter from the [RPO](https://huggingface.co/papers/2404.19733) paper V3. If None, no weighting is applied and the loss is the same as the DPO loss. The paper recommends `rpo_alpha=1.0`.
```

PR #1909, opened 2024-08-07
Various args and test fix

### LINKED ISSUE #1905: Configuration argument overriding by deprecated default values
For a many deprecated args, you need to pass them it in the config and in the trainer, otherwise it will be overwritten by the default value of the trainer. Example with `num_of_sequences` and `SFTTrainer`:

```python
>>> from trl import SFTConfig, SFTTrainer
>>> args = SFTConfig("tmp", num_of_sequences=32)
>>> trainer = SFTTrainer("gpt2", args=args)
/fsx/qgallouedec/trl/trl/trainer/sft_trainer.py:185: UserWarning: You passed a model_id to the SFTTrainer. This will automatically create an `AutoModelForCausalLM` or a `PeftModel` (if you passed a `peft_config`) for you.
  warnings.warn(
/fsx/qgallouedec/trl/trl/trainer/sft_trainer.py:289: UserWarning: You didn't pass a `max_seq_length` argument to the SFTTrainer, this will default to 1024
  warnings.warn(
/fsx/qgallouedec/trl/trl/trainer/sft_trainer.py:352: UserWarning: You passed a `num_of_sequences` argument to the SFTTrainer, the value you passed will override the one in the `SFTConfig`.
  warnings.warn(
>>> trainer.args.num_of_sequences
1024
```

If wrongly report that I've passed a `num_of_sequences` argument to the SFTTrainer. And the value that I've set in `SFTConfig` is overwritten by the trainer's default (1024). The only way to make it work as expected is to use 

```python
args = SFTConfig("tmp", num_of_sequences=32)
trainer = SFTTrainer("gpt2", args=args, num_of_sequences=32)
```

which is not satisfying, since it create a duplicate, and it uses a deprecated arg.

qgallouedec:
Some test doesn't pass locally. Consequently, I propose the following fix/improvement:

## 1. Remove unused `peft_test` marker
```
tests/test_ppo_trainer.py:910
  /fsx/qgallouedec/trl/tests/test_ppo_trainer.py:910: PytestUnknownMarkWarning: Unknown pytest.mark.peft_test - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @mark.peft_test
```

## 2. Don't report to wandb
wandb does not cope well with multiple executions in the same process, and generates errors such as 
```
wandb.sdk.lib.config_util.ConfigError: Attempted to change value of key "model/num_parameters" from 1658681 to 1663801
```
the easiest is not to report during test. Already fixed for reward tester in #1895

## 3. Don't share model across tests

Also partially addressed in #1895. When sharing the model, it creates conflicts between tests, due to model re-training mostly.

```
FAILED tests/test_dpo_trainer.py::DPOTrainerTester::test_dpo_trainer_w_dataset_num_proc - _pickle.PicklingError: Cannot pickle a prepared model with automatic mixed precision, please unwrap the model with `Accelerator.unwrap_model(model)` before pickling it.
```

The solution is to replace `setUpClass` by `setUp` (so that the setup is not shared)

## 4. Fixes configuration argument being overrided by deprecated default values

Fixes #1905 and adds tests.


qgallouedec:
Finally 😮‍💨

kashif (approved):

PR #1439, opened 2024-03-18
Before update the tr_loss, make sure tr_loss_step is in the same device. 

pengwei715:
# What does this PR do?

Fixes: https://github.com/huggingface/trl/issues/958
Fixes: https://github.com/huggingface/trl/issues/1399

It's a very strong assumption that `tr_loss` and `tr_loss_step` are on the same device. `arg.device` may not be as same as the current device. 

For example, the `dpo_trainer` is child class of `trainer`. If you are running a DPO training job with one node multiple GPUs. If you set the `device_map='auto'`. The `tr_loss_step` could be on different device after computing the `dpo_loss`.  

Based on the feedback of @guy1992l, @younesbelkada, @amyeroberts, I have open a PR in the transformers `Trainer` class to do a assert checking. 

https://github.com/huggingface/transformers/pull/29695

This PR is put the `tr_loss_step` to the `self.args.device` before update. 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?
@younesbelkada 
@amyeroberts
@guy1992l

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #1919, opened 2024-08-12
Fix orpo trainer loss device

SunMarc:
# What does this PR do ? 

Same PR as https://github.com/huggingface/trl/pull/1439 but for ORPOTrainer. 
TLDR: It's a very strong assumption that `tr_loss` and `tr_loss_step` are on the same device. `arg.device` may not be as same as the current device.
Fixes: https://github.com/huggingface/transformers/pull/32480

cc @kashif


kashif (approved):

SunMarc:
I'm unable to merge as I don't have write access @kashif. Can you merge it please ?

PR #1943, opened 2024-08-18
[XPO] xpo trainer

kashif:
https://huggingface.co/papers/2405.21046 implementation


trl/trainer/online_dpo_trainer.py
```diff
@@ -162,6 +167,16 @@ def __init__(
...
 
+        self.generation_config = GenerationConfig(
+            max_new_tokens=args.max_new_tokens,
+            min_new_tokens=args.max_new_tokens,
+            temperature=args.temperature,
+            top_k=0,
+            top_p=1.0,
+            do_sample=True,
+            use_cache=False if args.gradient_checkpointing else True,
+        )
```
qgallouedec, line 178: Out of curiosity, why do you need this change?

kashif: i realized we were making this for every step and since its fixed for now... i moved it over to the main __init__

qgallouedec: ok I got it, so that you don't have to define it again in XPOTrainer

kashif: yes that too haha

trl/trainer/xpo_trainer.py
```diff
@@ -0,0 +1,404 @@
...
+        chosen_rewards = chosen_log_ratios * self.args.beta
+        rejected_rewards = rejected_log_ratios * self.args.beta
+        self.stats["rewards/chosen"].append(gather_mean(chosen_rewards.mean()))
+        self.stats["rewards/rejected"].append(gather_mean(rejected_rewards.mean()))
+
+        # Calculate margins correctly
+        if chosen_rewards.numel() > 0 and rejected_rewards.numel() > 0:
+            # Compute average margin
+            margin = chosen_rewards.mean() - rejected_rewards.mean()
+        else:
```
qgallouedec, line 317: is it possible?

kashif: good catch... no its not let me remove it...

qgallouedec:
kl isn't logged in XPO. Is it possible to have it logged?

trl/trainer/xpo_trainer.py
qgallouedec: I'm not saying it's not possible, I just don't understand why we've this `if/else` but not in online dpo.

kashif: ah because i mistakenly thought it could happen that there are no chosen or rejected sampled... but yeah its not correct... we are as in the online-dpo, will always have a set of chosen and rejected

qgallouedec:
> Is it possible to have it logged?

from @kashif 
> In the online-dpo what is the "kl" being logged?

`"kl"` is the approximate KL div from the ref model distribution to the trained model distribution. Approximated as:

https://github.com/huggingface/trl/blob/72f19c3fcecb0d53f356b1821d36073b65148aa0/trl/trainer/online_dpo_trainer.py#L389-L389

Your calculation [`abdf1a5`](https://github.com/huggingface/trl/pull/1943/commits/adbf1a512891134bf4a93133bfdaa9d09a429476) seems correct to me

```python
        # Calculate KL divergence for model and ref data
         kl_model_data = model_logprobs_model_data - ref_logprobs_model_data
         kl_ref_data = model_logprobs_ref_data - ref_logprobs_ref_data
         mean_kl = (kl_model_data.sum(1) + kl_ref_data.sum(1)).mean() / 2
         self.stats["objective/kl"].append(gather_mean(mean_kl))
```

examples/scripts/xpo.py
```diff
@@ -0,0 +1,129 @@
...
+# limitations under the License.
+"""
+Usage:
+
+python examples/scripts/xpo.py \
+    --model_name_or_path EleutherAI/pythia-14m  \
+    --reward_model_path EleutherAI/pythia-14m \
+    --dataset_name trl-lib/tldr \
+    --learning_rate 5.0e-7 \
+    --output_dir pythia-1b-tldr-xpo \
```
qgallouedec, line 23: 1b or 14m? You need to choose ;)

kashif: let me copy from the online dpo example

trl/trainer/xpo_trainer.py
```diff
@@ -0,0 +1,417 @@
...
+            # Remove "kl", "entropy", "non_score_reward", "rlhf_reward", "scores"
+            # Add "loss/dpo", "loss/xpo"
+            "loss/dpo": [],
+            "loss/xpo": [],
+            # Replace "scores" by "model_scores" and "ref_scores"
+            "objective/model_scores": [],
+            "objective/ref_scores": [],
+            "objective/scores_margin": [],
+            "objective/kl": [],
+            "objective/entropy": [],
```
qgallouedec, line 124:
```suggestion
            # Remove "non_score_reward", "rlhf_reward", "scores"
            # Add "loss/dpo", "loss/xpo"
            "loss/dpo": [],
            "loss/xpo": [],
            "objective/kl": [],
            "objective/entropy": [],
            # Replace "scores" by "model_scores" and "ref_scores"
            "objective/model_scores": [],
            "objective/ref_scores": [],
            "objective/scores_margin": [],
```

qgallouedec (approved):
Very nice work @kashif

PR #1945, opened 2024-08-19
Skip token in `push_to_hub`

qgallouedec:


trl/trainer/bco_trainer.py
```diff
@@ -1432,12 +1432,12 @@ def push_to_hub(
         self,
         commit_message: Optional[str] = "End of training",
         blocking: bool = True,
-        token: Optional[str] = None,
+        # token: Optional[str] = None,  # Unlike the parent class, skip the token to mitigate security risks.
```
lewtun, line 1435: Can't we delete altogether instead of having this commented out code?

```suggestion
```

lewtun (commented):
Thanks for the safety improvement! My main question is whether we need to have commented out args - is there a cleaner way?

trl/trainer/bco_trainer.py
qgallouedec: We can. 
I just want to find a way to tell my future self not to re-add it. I'll add to the comment below, let me know if that's ok.

qgallouedec:
Note that thanks to `@wraps(Trainer.push_to_hub)`, the docstring (```"""Overwrite the `push_to_hub` method in order to ..."""```) won't appear in the online doc. It will only be visible in the source code.

trl/trainer/bco_trainer.py
lewtun: This is better, thanks!

lewtun (approved):
Thanks for iterating, LGTM!

PR #1949, opened 2024-08-20
Don't mark issues as stale if nobody answered

qgallouedec:
A lot of issues are closed before being answered. I think unanswered issues shouldn't be closed automatically.


lewtun (approved):
Thanks for catching this - LGTM!

PR #1951, opened 2024-08-20
Log WandB tables on main process

lewtun:
This PR fixes a small bug where tables of model completions were being logged on all processes.


vwxyzjn (approved):
Yikes. Good catch!

kashif (approved):

PR #1952, opened 2024-08-21
Feat: Add support for APO-zero in KTOTrainer

KarelDO:
### Feat: Add support for APO-zero in KTOTrainer

**Now ready to merge**

This PR adds support for the unpaired variant of APO-zero in the KTOTrainer. See the [APO paper](https://www.arxiv.org/abs/2408.06266).

To achieve this, I:
- Added a `loss_type` variable to `KTOConfig` (similar to `DPOConfig`)
- Added the APO loss to `KTOTrainer` (similar to `DPOTrainer`)
- Updated `KTOTrainer` to only calculate the KL when the loss requires this (calculating the KL is expensive, as a result APO-zero runs faster than KTO).

Additionally, I updated the `kto.py` script to be interoperable with any dpo-formatted datasets:
- I added a util (in `data_util.py`) which checks the format of the dataset and turns a dpo-formatted dataset into a kto-formatted dataset.
- I've updated the kto script to also work with ChatML formatted datasets (similar to what happens in `dpo.py`).


KarelDO:
I've confirm the training dynamics of APO-zero-unpaired are as intended.

Compared to a KTO-run, APO-zero produces about a 40% speedup since no KL values need to be calculated.

I will run more downstream evaluations to understand the differences between KTO and APO-zero for unpaired alignment later.

**This PR can now be merged**

qgallouedec:
Thank you @KarelDO, it's a feature we're very happy to see coming to TRL. We'll be reviewing your PR soon for sure. Can you share the elements you have to confirm that everything is working as expected? Maybe curves, trained models, etc?

trl/trainer/kto_config.py
```diff
@@ -60,6 +65,10 @@ class KTOConfig(TrainingArguments):
             Number of processes to use for processing the datasets.
     """
 
+    loss_type: Literal[
+        "kto",
+        "apo_zero_unpaired",
```
qgallouedec, line 70: is "unpaired" really necessary? As far as I understand, there is no such thing as "paired" version for kto, right?

KarelDO: APO-zero does have a paired and unpaired variant, and you could definitely construct a paired variant of KTO.

We can remove "_unpaired" here since the KTOTrainer also implies it, but I thought it would be good for people to actively think about the distinction when selecting a loss.

KarelDO:
Hey @qgallouedec , thanks!

Here's a [wandb report pdf](https://github.com/user-attachments/files/16738076/Comparison.APO-unpaired.pdf) with some training curves. I can't share the full WandB project yet, but I have a print out of the report attached. I've summarized the main take-aways below:

- APO-zero trains considerably faster
- On an RLAIF preference dataset, the training dynamics of KTO and APO-zero seem identical
- on a CLAIR preference dataset, the training dynamics of KTO and APO-zero differ. This is due to a higher KL on this dataset. APO-zero does display the intended training dynamic of smoothly increasing desirable rewards and decrease undesirable rewards without calculating a KL.


TL;DR: different loss functions, they sometimes behave similarly depending on underlying preference dataset, APO-zero trains faster.

trl/data_utils.py
```diff
@@ -0,0 +1,56 @@
...
+    rejected_row = {
+        "prompt": row["prompt"],
+        "completion": row["rejected"],
+        "label": [False] * len(row["chosen"]),
+    }
+    new_rows = {k: chosen_row[k] + rejected_row[k] for k in chosen_row.keys()}
+    return new_rows
+
+
+def maybe_reformat_dpo_to_kto(dataset: DatasetDict, num_proc: int = None):
```
lewtun, line 32: For public methods, would you mind adding a docstring and a unit test please?

trl/trainer/kto_config.py
lewtun: Yes given we have an `apo_zero` loss also in the `DPOTrainer`, It's good to retain the `_unpaired` distinction IMO

Would you mind adding this loss term to the intergration tests here: https://github.com/huggingface/trl/blob/47ab034ca981104ff35802a83db46c0cb5fe5a16/tests/test_kto_trainer.py#L76

You might want to look at the DPO trainer for inspiration: https://github.com/huggingface/trl/blob/47ab034ca981104ff35802a83db46c0cb5fe5a16/tests/test_dpo_trainer.py#L251

lewtun (approved):
Thanks a lot for this nice contribution @KarelDO ! Overall it LGTM once we have some unit / integration tests added.

trl/data_utils.py
KarelDO: Done!

lewtun:
Thanks for iterating! Would you mind fixing the code quality test and then we can merge!

KarelDO:
Thanks @lewtun , should be fixed now!

lewtun:
Ah it seems some of the KTO tests are now failing after rebasing on main - would you mind fixing those 🙏 ?

karel-contextual:
@lewtun we should be good now!

lewtun:
Thanks for iterating!

PR #1954, opened 2024-08-21
initial grpo files

saisurbehera:


lewtun:
Thank you for working on this nifty algorithm @saisurbehera ! I see you're basing your implementation on `PPOTrainer` but we've recently overhauled our RL implementations to be more aligned with the rest of the library, e.g. here's the new PPO version: https://github.com/huggingface/trl/blob/main/trl/trainer/ppov2_trainer.py

Would you mind adapting your implementation to this new API? Since GRPO is somewhat similar to RLOO, you might find it is possible to copy-paste a large part of that code: https://github.com/huggingface/trl/blob/main/trl/trainer/rloo_trainer.py

saisurbehera:
Hello @lewtun , 

I ported the format to the new methodlogy, it was way simpler than the first version. I still have to do some validations and testing.

saisurbehera:
I think we should test it out. Thanks a lot for the change. 

Overall, I do think most of my work is done based on the limits of trl. We need more extensive changes to add PRM and reward model training. 

Sorry for not spending some more time on it, I was busy at work and family stuff.

fzyzcjy:
Hi, thanks for the PR, it would be great to have GRPO and looking forward to it!

rawsh:
@saisurbehera curious if it's ready to test? want to try my hand at hacking in PRM rewards

saisurbehera:
Go ahead

fzyzcjy:
Hi, is there any updates? I would appreciate it if this could be implemented!

saisurbehera:
Let me work over this weekend to verify, Sorry for the delay.

fzyzcjy:
Looking forward to it!

hijkzzz:
REINFORCE++ is better than GRPO: https://github.com/hijkzzz/Awesome-LLM-Strawberry/blob/main/resources/REINFORCE++.pdf

fzyzcjy:
Hi is there any updates?

saisurbehera:
Sorry for the late response. My code works now. The problem is the new model has very high KL divergence compared to the reference model. The scores compared to rloo don't look right. I have to debug as to why. Sorry for the it.

fzyzcjy:
It's OK and looking forward to the fix!

ehartford:
Do you have example training dataset?

PR #1957, opened 2024-08-22
Fix response truncation in examples/notebooks/gpt2-sentiment.ipynb

### LINKED ISSUE #1755: Misuse of gen_len in examples/notebooks/gpt2-sentiment.ipynb
https://github.com/huggingface/trl/blob/a57e75967c2b787f42f4e402ed7ca23cd9bad9a9/examples/notebooks/gpt2-sentiment.ipynb#L383
The intention of this line of code is to separate the new tokens generated by the model(i.e., response) from the prmopt+response.
But since gen_len is used as max_new_tokens，len(response)  <= gen_len.
When **len(response)< gen_len**_(which happens a lot when gen_len is large),_  this line of code doesn't actually work.
Then, query will be repeated in the  spliced  texts, resulting in an incorrect reward
https://github.com/huggingface/trl/blob/a57e75967c2b787f42f4e402ed7ca23cd9bad9a9/examples/notebooks/gpt2-sentiment.ipynb#L387

qgallouedec:
Fixes #1755


lewtun (approved):
Thanks for the fix!

PR #1959, opened 2024-08-22
Support LLaVA-NeXT in Vision SFT

### LINKED ISSUE #1785: Error with SFT of LLaVA-Next
I'm trying to instruction tune llava-next models following the llava_vsft.py examples shared for llava-1.5.
```
python vsft.py \
    --dataset_name="HuggingFaceH4/llava-instruct-mix-vsft" \
    --model_name_or_path="llava-hf/llava-v1.6-mistral-7b-hf" \
    --report_to="tensorboard" \
    --learning_rate=2e-5 \
    --lr_scheduler_type="cosine" \
    --per_device_train_batch_size=8 \
    --logging_steps=1 \
    --num_train_epochs=1 \
    --gradient_checkpointing \
    --torch_dtype=float16 \
    --fp16=True \
    --max_seq_length=4096 \
    --attn_implementation="flash_attention_2"
```
The run keeps failing on a 8xH100 VM with the following error:
```
RuntimeError: Input tensor at index 1 has invalid shape [8, 2785, 32064], but expected [8, 2889, 32064]
```

The full code and error stack trace is available in [this gist](https://gist.github.com/GohioAC/44cc9ac2ee54380ff80aa6ab89fea698).

qgallouedec:
Closes #1785


kashif (approved):

kashif:
do we need to pin the transformer version to the latest?

qgallouedec:
> do we need to pin the transformer version to the latest?

Good point! let me check

kashif:
it would be good to pin the transformer to latest due to another PR as well...

lewtun:
> do we need to pin the transformer version to the latest?

Where do you want to pin the version? I don't think we want to do this in `requirements.txt`, but an alternative would be to specify a min version for this specific model that is checked like we do in `transformers`: https://github.com/huggingface/transformers/blob/273c0afc8f0289e0cb5d18dc76ccc251504298e0/src/transformers/utils/__init__.py#L260

kashif:
@lewtun sorry i meant `>= <latest>`

lewtun:
> @lewtun sorry i meant `>= <latest>` in here: https://github.com/huggingface/trl/blob/main/setup.py#L65

Ah no we don't want to do that because it will lead to a lot of pain for users with conflicting `transformers` dependencies. In my opinion we should use the checker helper (or something similar) for architectures which are version-specific

qgallouedec:
LLaVA-NeXT has been added in v4.40.0~235 but the example is failing before https://github.com/huggingface/transformers/commit/e84012737 (not released yet)

examples/scripts/vsft_llava.py
```diff
@@ -27,6 +27,9 @@
     --use_peft \
     --dataloader_num_workers 32 \
     --lora_target_modules=all-linear
+
+For LLaVA-NeXT, use: (requires transformers>=4.45)
```
qgallouedec, line 31: @lewtun @kashif would this be enough for the version?

lewtun: This is fine for me - if we get more issues we can always add a stronger check in the trainer

lewtun (approved):
LGTM once the tests pass!

PR #1961, opened 2024-08-23
Fix issue with precompute_ref_log_probs not working when rpo_alpha is None

mina-parham:
When using the `precompute_ref_log_probs=True` option, the training process is halted( the training loss sucks at 0.6932 which shows the model is not training). This issue arises due to the following code segment:

```
if (
            "reference_chosen_logps" in batch
            and "reference_rejected_logps" in batch
            and self.args.rpo_alpha is not None
        ):
            reference_chosen_logps = batch["reference_chosen_logps"]
            reference_rejected_logps = batch["reference_rejected_logps"]
```

The condition `self.args.rpo_alpha` is not None prevents the `reference_chosen_logps` and `reference_rejected_logps` from being used in the training process when `rpo_alpha` is set to None. This is problematic because when `precompute_ref_log_probs=True`, we still need to access these log probabilities from the batch regardless of the `rpo_alpha` value.

To resolve this issue, the condition should be adjusted to ensure that `reference_chosen_logps` and `reference_rejected_logps` are accessed when `precompute_ref_log_probs` is set to True, independent of the `rpo_alpha` value.

### DIFF
--- a/tests/test_dpo_trainer.py
+++ b/tests/test_dpo_trainer.py
@@ -163,7 +163,13 @@ class DPOTrainerTester(unittest.TestCase):
                 if param.sum() != 0:
                     assert not torch.allclose(param, new_param, rtol=1e-12, atol=1e-12)
 
-    def test_dpo_trainer_without_providing_ref_model(self):
+    @parameterized.expand(
+        [
+            [None, "Test when rpo_alpha is set to None"],
+            [0.5, "Test when rpo_alpha is set to 0.5"],
+        ]
+    )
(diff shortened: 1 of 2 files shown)


kashif (approved):
good catch

kashif:
@mina-parham would it be possible to add a test for this?

mina-parham:
> @mina-parham would it be possible to add a test for this?

yes of course, I'm on it :-)

mina-parham:
I added the test for this.

Issue #1968, opened 2024-08-26
Unnecessary memory cached during DPO reference logp calculation


SeungyounShin:
Unnecessary memory cached during DPO reference logp calculation.

qgallouedec:
Closed by #1969

PR #1969, opened 2024-08-26
Optimize DPO log probability calculation by retaining necessary cache, saving up to 30GB of memory (#1968)

SeungyounShin:
# What does this PR do?

This PR addresses an issue where unnecessary cache clearing was performed during the log probability calculation in the training loop. The original code included calls to `torch.cuda.empty_cache()` and `self.accelerator.free_memory()` to avoid out-of-memory (OOM) errors, but these were not necessary and may have introduced performance overhead. By removing these calls, this PR optimizes the memory management during training.

Fixes # (issue)

## Motivation and Context

This change improves the efficiency of the log probability calculation during training by eliminating redundant memory management operations. This can lead to better performance and more efficient use of GPU memory.

## Before submitting

- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs).

## Who can review?

Anyone in the community is welcome to review this PR once the tests have passed. Feel free to tag members or contributors who may be interested in reviewing.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -648,6 +648,10 @@ class DPOTrainer(Trainer):
                 reference_chosen_logps.append(reference_chosen_logp.cpu())
                 reference_rejected_logps.append(reference_rejected_logp.cpu())
 
+                # Unnecessary cache clearing to avoid OOM
+                torch.cuda.empty_cache()
+                self.accelerator.free_memory()
+
             all_reference_chosen_logps = torch.cat(reference_chosen_logps).float().numpy()
             all_reference_rejected_logps = torch.cat(reference_rejected_logps).float().numpy()
 


trl/trainer/dpo_trainer.py
```diff
@@ -648,6 +648,10 @@ def get_train_dataloader(self) -> DataLoader:
                 reference_chosen_logps.append(reference_chosen_logp.cpu())
                 reference_rejected_logps.append(reference_rejected_logp.cpu())
 
+                # Unnecessary cache clearing to avoid OOM
```
SeungyounShin, line 651: Without this line, GPU memory usage can increase up to 45GB when running a LLaMA-8B model on an 8K context dataset:

Screenshot 1

Simply adding this line reduces and stabilizes memory consumption, keeping it under control:

Screenshot 2

SeungyounShin: It seems that this cache occupies memory during training, leading to reduced available memory and, consequently, limiting the context size that can be used for training.

kashif (approved):
nice catch!

PR #1990, opened 2024-08-28
Relax numpy upper bound and bump deepspeed version

hvaara:
DeepSpeed `v0.14.4` and later supports current numpy (confirmed in https://github.com/microsoft/DeepSpeed/issues/5671).

This should allow us to get rid of the upper bound on numpy after a version bump of DeepSpeed.

See https://github.com/huggingface/trl/pull/1747 for original pin.


hvaara:
Running tests locally passed

```
$ make test
212 passed, 180 skipped, 592 warnings in 480.39s (0:08:00)
```

I'm not sure if this actually tests anything related to DeepSpeed with `numpy>=2.0.0`. Will the DeepSpeed integration be tested in CI?

setup.py
```diff
@@ -63,7 +63,7 @@
 REQUIRED_PKGS = [
     "torch>=1.4.0",
     "transformers>=4.31.0",
-    "numpy>=1.18.2,<2.0.0",
+    "numpy>=1.18.2",
```
lewtun, line 66: To be safe, let's add the upper bound for the next major release:
```suggestion
    "numpy>=1.18.2,<3.0.0",
```

lewtun (approved):
Thanks for fixing the deps @hvaara ! LGTM with a small change

setup.py
hvaara: Are you sure you really want this? There hasn't been any any known issues with this version yet. By this logic we should add upper bound pins on all the other deps too?

I'll make the change if you really want it. Just want to make sure you do first.

lewtun: I took a look at our other libs and indeed we don't upper bound `numpy`, so let's keep it as is. Thanks!

hvaara:
Thanks for the review! 🤗

PR #1999, opened 2024-08-30
feat: start working on revisions instead of preferences

KarelDO:
**Don't land this PR yet**. I would initially like some feedback on the idea of this PR, and if this is a welcome change in TRL.

This PR seeks to adds _revisions_ to TRL. Revisions create preference pairs by minimally revising a completion, unlike Judges who rank a set of completions.

To do this, I've added a `tmp/trainer/revisions.py` file with all the revision logic.

see CLAIR paper: https://www.arxiv.org/abs/2408.06266

TODO:

- [ ] add tests
- [ ] add max retries in OpenAIReviser
- [ ] handle errors in the chat completion parsing

### DIFF
--- a/examples/notebooks/revisions.ipynb
+++ b/examples/notebooks/revisions.ipynb
@@ -1,0 +1,58 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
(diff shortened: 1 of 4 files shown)


qgallouedec:
I'm closing because there hasn't been any activity for a while. This PR may be reopened if there is renewed interest in it.

PR #2001, opened 2024-08-31
Integrate OrpoTrainer with PyTorchXLA for faster step time on TPUs

wenxindongwork:
OrpoTrainer currently runs very slow on TPU because the code is not integrated with TorchXLA. Currently there are too many dynamic shapes and data device transfer in the code which trigger graph recompilation and slow down step time. This PR makes changes to improve step time of OrpoTrainer on TPUs by more than 300x. Tested on Llama3-8b, the current step time is 2s using Lora on all linear modules, compared to 10mins which we started with.  

The changes should not impact performance on other backends since we have guarded the changes with is_torch_xla_available.


trl/trainer/orpo_trainer.py
```diff
@@ -533,7 +536,17 @@ def tokenize_row(self, feature, model: Optional[Union[PreTrainedModel, nn.Module
...
+        if is_torch_xla_available():
+            #Pad the sequences to global max_length
+            for k in batch:
+                if "labels" in k or self.is_encoder_decoder:
+                    pad_value = self.label_pad_token_id
+                elif k.endswith("_input_ids"):
+                    pad_value = self.padding_value
+                elif k.endswith("_attention_mask"):
+                    pad_value = 0
+                batch[k] = pad_list_to_length(batch[k], self.max_length, pad_value=pad_value)
```
qgallouedec, line 549:
```suggestion
                batch[k] = batch[k] + [pad_value] * (self.max_length - len(batch[k]))
```

trl/trainer/orpo_trainer.py
```diff
@@ -533,7 +536,17 @@ def tokenize_row(self, feature, model: Optional[Union[PreTrainedModel, nn.Module
                 batch["chosen_decoder_input_ids"] = model.prepare_decoder_input_ids_from_labels(
                     labels=torch.tensor(batch["chosen_labels"])
                 )
-
+        
+        if is_torch_xla_available():
```
qgallouedec, line 540: Why do you need this only when `is_torch_xla_available`?

trl/trainer/orpo_trainer.py
```diff
@@ -35,7 +35,7 @@
 from transformers import AutoModelForCausalLM, DataCollator, PreTrainedModel, PreTrainedTokenizerBase, Trainer
 from transformers.trainer_callback import TrainerCallback
 from transformers.trainer_utils import EvalLoopOutput
-from transformers.utils import is_torch_fx_proxy
+from transformers.utils import is_torch_fx_proxy, is_torch_xla_available
```
qgallouedec, line 38: This has been added with `transformers` v4.39. We should probably set this version as the new minimal requirement.

trl/trainer/orpo_trainer.py
```diff
@@ -659,7 +672,7 @@ def get_batch_logps(
         loss_mask = labels != label_pad_token_id
 
         # dummy token; we'll ignore the losses on these tokens later
-        labels[labels == label_pad_token_id] = 0
+        labels = torch.where(labels == label_pad_token_id, 0, labels)
```
qgallouedec, line 675: Is it a necessary change? Personal opinion, I find it a bit less intuitive to read

qgallouedec:
Thank you very much for this addition @wenxindongwork! Unfortunately we can't test with GitHub CI but I'm relying on you for the fact that it works and run faster.
Can you just address the question/comment? then we're good to merge.

trl/trainer/orpo_trainer.py
wenxindongwork: Pytorch XLA doesn't support dynamic shape compilation, so we are padding all sequences to the global batch size. This may not be a problem for GPU, so I kept the original algorithm which supports padding to the longest sequence length in the batch.

trl/trainer/orpo_trainer.py
wenxindongwork: this for catching this!

wenxindongwork:
addressed comments, thanks for the quick review!

trl/trainer/orpo_trainer.py
wenxindongwork: thanks for the suggestion!

lewtun:
Hello @wenxindongwork can you please fix the code quality issues with `make precommit` 🙏 ?

wenxindongwork:
should work now!

qgallouedec:
Can you also set the min transformers version in `setup.py` as well?

wenxindongwork:
just did, thanks for pointing this out!

PR #2009, opened 2024-09-03
[ci] pin numpy to < 2 on windows

kashif:
# What does this PR do?
Fix windows CI failures by pinning numpy version < 2


lewtun (approved):
Thanks for the fix! LGTM

PR #2036, opened 2024-09-08
Remove deprecated args in trainers

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/dpo_trainer.py
```diff
@@ -389,74 +388,22 @@ class DPOTrainer(Trainer):
 
     _tag_names = ["trl", "dpo"]
 
-    @_deprecate_arguments(
-        version="1.0.0",
```
lewtun, line 393: Out of curiosity, is the reason to remove these deprecated args due to the fact we've had this warning for several releases?

I wonder if it's better to pin the `version` to 2-3 future versions to give people a concrete warning?

lewtun (commented):
Overall I'm happy to remove these args, but since it's technically a breaking change I wonder if we need to specify the specific version we'll do this in (currently set to `v1.0.0` which is too far away I guess)

Also it would be good to have a small PR description as I wasn't quite sure what the motivation of this PR was :)

trl/trainer/dpo_trainer.py
qgallouedec: For context, these arguments have been deprecated in #1554, in [v0.9.3](https://github.com/huggingface/trl/releases/tag/v0.9.3).

Sounds good, let's wait a bit.

Issue #2014, opened 2024-09-03
Can model be `None`?


qgallouedec:
I see we're using `is_encoder_decoder` and the config with the following docstring.

    If no model is provided, we need to know if the model_init returns an encoder-decoder.

I can't understand in what situation a trainer can be instantiated without a model. If there is indeed no such situation, we should remove this argument.

qgallouedec:
It seems that we can indeed not specify any model when we use `model_init` arg in trainers (transferred to the trainer in the init). Not sure if it could work with try's trainer though. We should probably add a test for this at some point.

PR #2041, opened 2024-09-09
PEFT support for Online DPO

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


qgallouedec:
```sh
python examples/scripts/dpo_online.py \
    --model_name_or_path meta-llama/Meta-Llama-3.1-8B-Instruct  \
    --reward_model_path RLHFlow/ArmoRM-Llama3-8B-v0.1 \
    --dataset_name qgallouedec/ultrafeedback-prompt \
    --learning_rate 5.0e-7 \
    --output_dir llama-3.1-8b-ultrafeedback-online-dpo \
    --per_device_train_batch_size 4 \
    --gradient_accumulation_steps 32 \
    --num_train_epochs 3 \
    --max_new_tokens 53 \
    --warmup_ratio 0.1 \
    --missing_eos_penalty 1.0 \
    --push_to_hub
```

kashif (approved):

kashif:
can you kindly add the above peft/lora usage command in the script?

trl/trainer/online_dpo_trainer.py
```diff
@@ -92,23 +97,22 @@ class OnlineDPOTrainer(Trainer):
 
     def __init__(
         self,
-        model: Union[PreTrainedModel, nn.Module] = None,
-        ref_model: Union[PreTrainedModel, nn.Module] = None,
-        reward_model: Optional[nn.Module] = None,
+        model: Union[PreTrainedModel, nn.Module, None] = None,
```
edbeeching, line 100: In what circumstances could the model be `None`?

trl/trainer/online_dpo_trainer.py
```diff
@@ -129,15 +133,40 @@ def __init__(
...
 
+        # Handle the ref_model
+        # Usually, the user wants the ref model to be the initial version of the model. When using PEFT, it's easy to
+        # get the ref model, as it's just the model with a disabled adaptater. When not using PEFT, we need to create
+        # the ref model from the model by copying it and disable the gradients and set it in evaluation mode.
+        if ref_model is None:  # No ref model provided, the most common case
+            if peft_config is None:
+                self.ref_model = create_reference_model(model)  # copy, disable gradients, set eval mode
+            else:
+                self.ref_model = None  # we don't need a ref model here, we can just disable the adaptater.
```
edbeeching, line 160: typo, `adapter`

trl/trainer/online_dpo_trainer.py
kashif: could be none when a pretrained model string id is given?

trl/trainer/online_dpo_trainer.py
```diff
@@ -129,15 +133,40 @@ def __init__(
...
+        else:  # rare case, the user provided a ref model
+            self.ref_model = create_reference_model(ref_model)  # copy, disable gradients, set eval mode
+
+        # Disable the gradient and set the reward model in eval mode
+        if self.reward_model is not None:
+            self.reward_model.eval()
+            for param in self.reward_model.parameters():
+                param.requires_grad = False
```
edbeeching, line 168: I am not sure why this is needed, as surely it is called with a `no_grad` context?

trl/trainer/online_dpo_trainer.py
```diff
@@ -326,8 +356,11 @@ def training_step(self, model: nn.Module, inputs: Dict[str, Union[torch.Tensor,
         del output, logits, all_logprobs  # free memory
 
         # Same for the reference model
-        with torch.no_grad():
+        if self.ref_model is not None:
             ref_output = self.ref_model(prompt_completion_ids, attention_mask=prompt_completion_mask)
```
edbeeching, line 360: is `with torch.no_grad():` needed here?

edbeeching: In fact, I would wrap lines 358 -> 373 in the same no grad context. As all the ref model and reward model stuff does not require grads.

qgallouedec:
Particular attention for reviewing this one. I didn't use the code from DPO mostly because I don't understand it. So I'm afraid to have missed something.

1. Why do you need to merge and unload?
https://github.com/huggingface/trl/blob/e51a5ac9856ca230d3e302278c1426ed69dc612f/trl/trainer/dpo_trainer.py#L533-L535

2. I don't support k-bit training yet
https://github.com/huggingface/trl/blob/e51a5ac9856ca230d3e302278c1426ed69dc612f/trl/trainer/dpo_trainer.py#L544-L556

3. Can we drop it for new trainer?
https://github.com/huggingface/trl/blob/e51a5ac9856ca230d3e302278c1426ed69dc612f/trl/trainer/dpo_trainer.py#L556-L565

4. Why do we need to cast? And why only when loaded in 4-bits. Btw why always using the `getattr`? We know that this arg exists in `TrainingArguments`
https://github.com/huggingface/trl/blob/e51a5ac9856ca230d3e302278c1426ed69dc612f/trl/trainer/dpo_trainer.py#L570-L574

Overall I've choose to go for a code that I understand, even if it implies introducing bug that have been fixed in the past for other trainers.

trl/trainer/online_dpo_trainer.py
qgallouedec: See https://github.com/huggingface/trl/issues/2014

trl/trainer/online_dpo_trainer.py
qgallouedec: thanks

trl/trainer/online_dpo_trainer.py
qgallouedec: eval is needed.

this

```python
for param in self.reward_model.parameters():
    param.requires_grad = False
...
model(...)
```

is the same as 

```python
with torch.no_grad():
    model(...)
```

do you think the later makes more sense?

trl/trainer/online_dpo_trainer.py
qgallouedec: I don't think it's needed because `self.ref_model = create_reference_model(ref_model)` disable the gradients

trl/trainer/online_dpo_trainer.py
qgallouedec: > https://github.com/huggingface/trl/pull/2041#discussion_r1758452282

So `model` would be a string, not a `None`. Anyway passing a string is not support in Online DPO (do we need it?)

kashif: ah yes you are right.. i was thinking in the script not the trainer... i don't think we need it

qgallouedec: let's make the model a required arg then

trl/trainer/online_dpo_trainer.py
qgallouedec: Let's opt for the second option, it is preferable to limit the modification of input objects

trl/trainer/online_dpo_trainer.py
qgallouedec: New:

```diff
  if ref_model is None:  # No ref model provided, the most common case
      ...
  else:  # rare case, the user provided a ref model
-     self.ref_model = create_reference_model(ref_model)  # copy, disable gradients, set eval mode
+     self.ref_model = ref_model
+     self.ref_model.eval()
```

and global `torch.no_grad` context, as suggested

examples/scripts/dpo_online.py
```diff
@@ -21,13 +21,23 @@
...
-    --per_device_train_batch_size 4 \
-    --gradient_accumulation_steps 32 \
-    --num_train_epochs 3 \
-    --max_new_tokens 53 \
+    --per_device_train_batch_size 8 \
+    --gradient_accumulation_steps 16 \
+    --warmup_ratio 0.1 \
+    --missing_eos_penalty 1.0
+
+With PEFT:
```
lewtun, line 29: PEFT is rather broad, so let's be precise here:

```suggestion
With LoRA:
```

examples/scripts/dpo_online.py
```diff
@@ -21,13 +21,23 @@
...
+    --reward_model_path trl-lib/pythia-1b-deduped-tldr-rm \
+    --dataset_name trl-lib/tldr \
+    --learning_rate 5.0e-6 \
+    --output_dir pythia-1b-tldr-online-dpo \
+    --per_device_train_batch_size 16 \
+    --gradient_accumulation_steps 8 \
     --warmup_ratio 0.1 \
     --missing_eos_penalty 1.0 \
-    --push_to_hub
+    --use_peft
```
lewtun, line 40: WDYT about adding the `lora_r` and `lora_alpha` args like we do in our other examples?

examples/scripts/dpo_online.py
```diff
@@ -69,40 +82,40 @@
     model = AutoModelForCausalLM.from_pretrained(
         model_config.model_name_or_path, trust_remote_code=model_config.trust_remote_code, **model_kwargs
     )
-    ref_model = AutoModelForCausalLM.from_pretrained(
-        model_config.model_name_or_path, trust_remote_code=model_config.trust_remote_code, **model_kwargs
-    )
+
     reward_model = AutoModelForSequenceClassification.from_pretrained(
-        training_args.reward_model_path, num_labels=1, trust_remote_code=model_config.trust_remote_code
+        training_args.reward_model_path,
```
lewtun, line 87: If I'm not mistaken, one needs to specify the number of labels to 1 because the default will be 2. 

```suggestion
        training_args.reward_model_path,
        num_labels=1,
```

examples/scripts/dpo_online.py
```diff
@@ -69,40 +82,40 @@
...
     tokenizer = AutoTokenizer.from_pretrained(
         model_config.model_name_or_path,
         padding_side="left",
         trust_remote_code=model_config.trust_remote_code,
+        **model_kwargs,
     )
     if tokenizer.chat_template is None:
         tokenizer.chat_template = SIMPLE_QUERY_CHAT_TEMPLATE
+    if tokenizer.pad_token_id is None:
+        tokenizer.pad_token = tokenizer.eos_token
```
lewtun, line 101: I think you need to set the ID as `pad_token` is just the string repn
```suggestion
        tokenizer.pad_token_id = tokenizer.eos_token_id
```

lewtun (approved):
Thanks for adding PEFT support! Overall LGTM - have you run an experiment on e.g. TLDR to see if it looks OK?

examples/scripts/dpo_online.py
qgallouedec: Really? Because we literally have 68 occurence of this in the codebase 😅

examples/scripts/dpo_online.py
qgallouedec: This is strange that it runs without it...

examples/scripts/dpo_online.py
qgallouedec: Unless the default values are not suited for the method (for example, the learning rate can't be left as the default here), I'm in favour of having as few arguments as possible. It makes things simpler.

lewtun:
> 1. Why do you need to merge and unload?
This is needed if we have e.g. an SFT LoRA that needs to be merged into the base model, before finally inserting an adapter for DPO. See [here](https://huggingface.co/docs/trl/dpo_trainer#reference-model-considerations-with-peft) for some considerations.

> 2. I don't support k-bit training yet
Sure, let's leave this as follow-up PR

> 3. Can we drop it for new trainer?
Fine with me. We can add it later if people open an issue.

> 4. Why do we need to cast? And why only when loaded in 4-bits. Btw why always using the `getattr`? We know that this arg exists in `TrainingArguments`
Good question, this was added by Younes but maybe @BenjaminBossan can help clarify here why PEFT models need upcasting in 4-bit?

examples/scripts/dpo_online.py
lewtun: Ah maybe I'm wrong then. A good check would be to see if `pad_token_id` is also set to `eos_token_id` when we just set the string repns

BenjaminBossan:
> > 7\. Why do we need to cast? And why only when loaded in 4-bits. Btw why always using the `getattr`? We know that this arg exists in `TrainingArguments`
> 
> Good question, this was added by Younes but maybe @BenjaminBossan can help clarify here why PEFT models need upcasting in 4-bit?

The original addition of this function was from #1110 (which in turn references [this repo](https://github.com/artidoro/qlora/blob/7f4e95a68dc076bea9b3a413d2b512eca6d004e5/qlora.py#L396-L405)). I don't know the exact context, but I think the reason is to avoid indiscriminately casting all layers to bf16 -- specifically, layer norm stays in float32. This is probably based on some empirical finding that this is better for training, but after skimming the QLoRA paper, I could not find any mention of that, so I'm unsure.

examples/scripts/dpo_online.py
qgallouedec: Both work!

```python

print(tokenizer.pad_token_id) # None
print(tokenizer.pad_token_id) # 50256

print(tokenizer.pad_token) # None
tokenizer.pad_token_id = tokenizer.eos_token_id
print(tokenizer.pad_token) # <|endoftext|>
```

examples/scripts/dpo_online.py
lewtun: ok!

qgallouedec:
Thanks a lot @BenjaminBossan and @lewtun. I'll further investigate in a dedicated branch.

Issue #2046, opened 2024-09-09
PPOv2Trainer & RLOOTrainer - Add Safety Check that `policy` object != `ref_policy` object


RylanSchaeffer:
### Feature request

Please throw an error or at least raise a warning if `policy` and `ref_policy` are the same object.

### Motivation

I just discovered I was making a very naive mistake. I was creating a `policy` model and passing it to `PPOv2Trainer` and `RLOOTrainer` as both `policy` and `ref_policy` since I wanted the same model to be used as the policy and the reference policy like so:

```
    ppo_trainer = PPOv2Trainer(
        config=ppo_trainer_config,
        tokenizer=policy_model_tokenizer,
        policy=policy_model,
        ref_policy=policy_model,
        reward_model=proxy_reward_model,
        value_model=value_model,
        train_dataset=train_dataset,
        eval_dataset=eval_dataset
    )
```

This was a dumb mistake that just burned quite a bit of time and compute.

I would be very grateful if someone added a sanity check to the constructors of `PPOv2Trainer` and `RLOOTrainer` that checks whether `policy` and `ref_policy` are the same object, and if so, throw an error or at least raise a warning that the programmer probably intends to use two different copies of the same language model rather than the exact same language model.

**SEPARATELY & BIZARRELY**, even if the policy model is the reference policy model, the KL divergence is non-zero and fluctuates during all of training!!!

### Your contribution

I would be happy to implement this. Just let me know if you would prefer an error or a warning.

qgallouedec:
Good idea. I think we should implement it for all trainers. Error makes sense, as there is no way a user would want to do that.
However, handling ref_model/model is pretty tricky currently, maybe wait until #2047 is solved?

> SEPARATELY & BIZARRELY, even if the policy model is the reference policy model, the KL divergence is non-zero and fluctuates during all of training!!!

Do you think it's related to the numerical noise? Or maybe switching between eval/train?

qgallouedec:
closed via #2057

PR #2057, opened 2024-09-11
Added error check to RLOO, PPOv2, OnlineDPO that `ref_policy` and `policy` have different identities

RylanSchaeffer:
As discussed in [Issue 2046](https://github.com/huggingface/trl/issues/2046), we add an error check to make sure that the `ref_policy` and `policy` have different identities.

This is just a quick and dirty demonstration to show what I have in mind. Feedback is very welcome!

### DIFF
--- a/trl/trainer/online_dpo_trainer.py
+++ b/trl/trainer/online_dpo_trainer.py
@@ -107,6 +107,10 @@ class OnlineDPOTrainer(Trainer):
         optimizers: Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
         preprocess_logits_for_metrics: Optional[Callable[[torch.Tensor, torch.Tensor], torch.Tensor]] = None,
     ) -> None:
+
+        if ref_policy is policy:
+            raise ValueError("`policy` and `ref_policy` are the same Python object but should not be. You probably want two copies of the same model.")
+
         self.ref_model = ref_model
 
         if reward_model is not None and judge is not None:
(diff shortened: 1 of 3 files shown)


RylanSchaeffer:
As a quick reminder, `is` is a way of checking whether two Python objects have the same identity: https://stackoverflow.com/a/3647560/4570472

```
x = [1, 2, 3]
y = x
z = [1, 2, 3]
y == x  # True
y is x    # True
z == x  # True
z is x  # False
```

trl/trainer/online_dpo_trainer.py
```diff
@@ -107,6 +107,10 @@ def __init__(
         optimizers: Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
         preprocess_logits_for_metrics: Optional[Callable[[torch.Tensor, torch.Tensor], torch.Tensor]] = None,
     ) -> None:
+
+        if ref_policy is policy:
```
lewtun, line 111: I think we want this here:

```suggestion
        if ref_model is model:
```

lewtun (commented):
Thanks a lot for adding this sanity check @RylanSchaeffer ! If I understand correctly, this scenario can occur for any trainer where there is a reference model - would you like to expand your check to the other trainers as well?

qgallouedec:
Useful check, thanks!

qgallouedec (approved):

qgallouedec:
Failing test not related to this PR

qgallouedec:
Thanks @RylanSchaeffer!

PR #2058, opened 2024-09-11
Mask loss in gkd when generating from the student

gaetanlop:
# What does this PR do?

There was a weird behavior with `min_new_tokens`  set to `max_new_tokens`, I removed it and updated the loss function calculation to avoid computing the loss on padding tokens generated during student generation.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@kashif


kashif (approved):

lewtun (commented):
Thanks a lot for catching the `min_new_tokens` quirk @gaetanlop - we should probably apply this to our other online trainers @qgallouedec @kashif 

Regarding the masked loss, do you know if it makes a difference performance-wise? In any case, the PR logic LGTM and happy to merge once the tests are updated and the CI is green :)

kashif:
i can fix up the other trainers, good catch!

tests/test_gkd_trainer.py
```diff
@@ -103,21 +107,24 @@ def test_generate_on_policy_outputs(self):
...
         # Check shapes
         batch_size = len(prompts)
         self.assertEqual(new_input_ids.shape[0], batch_size)
         self.assertEqual(new_attention_mask.shape[0], batch_size)
+        self.assertEqual(new_labels.shape[0], batch_size)
 
         # Check types
         self.assertIsInstance(new_input_ids, torch.Tensor)
         self.assertIsInstance(new_attention_mask, torch.Tensor)
+        self.assertEqual(new_labels.shape[0], torch.Tensor)
```
kashif, line 123:
```suggestion
        self.assertIsInstance(new_labels, torch.Tensor)
```

kashif (commented):
in the tests perhaps set the `max_length=` to `max_new_tokens=`

trl/trainer/gkd_trainer.py
```diff
@@ -190,12 +198,15 @@ def generate_on_policy_outputs(model, inputs, generation_config, pad_token_id=No
         generated_tokens = generated_outputs.sequences
         # Calculate new attention mask
         new_attention_mask = torch.ones_like(generated_tokens)
+        new_labels = generated_tokens.clone()
 
+        new_labels[new_labels == eos_token_id] = -100
```
kashif, line 203: i also don't think we should mask out the eos token

trl/trainer/gkd_trainer.py
```diff
@@ -177,7 +185,7 @@ def compute_loss(self, model, inputs, return_outputs=False):
         return (loss, outputs_student) if return_outputs else loss
 
     @staticmethod
-    def generate_on_policy_outputs(model, inputs, generation_config, pad_token_id=None):
+    def generate_on_policy_outputs(model, inputs, generation_config, pad_token_id=None, eos_token_id=None):
```
kashif, line 188:
```suggestion
    def generate_on_policy_outputs(model, inputs, generation_config, pad_token_id=None):
```

trl/trainer/gkd_trainer.py
```diff
@@ -207,11 +218,16 @@ def training_step(self, model: nn.Module, inputs: Dict[str, Union[torch.Tensor,
...
         if random.random() <= self.lmbda:
             with unwrap_model_for_generation(model, self.accelerator) as unwrapped_model:
-                new_input_ids, new_attention_mask = self.generate_on_policy_outputs(
-                    unwrapped_model, inputs, self.generation_config, self.tokenizer.pad_token_id
+                new_input_ids, new_attention_mask, new_labels = self.generate_on_policy_outputs(
+                    unwrapped_model,
+                    inputs,
+                    self.generation_config,
+                    self.tokenizer.pad_token_id,
+                    self.tokenizer.eos_token_id,
```
kashif, line 226:
```suggestion
```

kashif:
also you can add the following to the docstrings:
```
labels: Tensor of shape (batch_size, sequence_length) with -100 for padding tokens to ignore when computing loss
```

gaetanlop:
Hey, thanks for the review. 

@lewton I have no idea if it makes a difference performance-wise, but since those tokens are usually masked out from the loss during SFT, I would assume that there is nothing to learn there even with soft labels.

@kashif Are you sure that we do not need to mask the `eos_token_id`? It might be better to shift the logits and the labels as usually done in SFT. Currently,  if the `pad_token_id ` is different from the `eos_token_id` we are not masking the loss for the token that follows the eos_token_id which in my opinion does not give any useful information to the model.

Great work on this `GKDTrainer` btw, it is really useful!

gaetanlop:
Yeah I completely agree, its trained with soft labels from the teacher. The logits at the position of the `eos_token_id` are the one to predict the next token which should be a pad_token. That's why I believe we could mask the `eos_token`.  Am I misunderstanding something? Anyways, I don't think this is an important issue.

I think there is a problem when doing `student_logits = outputs_student.logits[:, prompt_lengths:, :]` as it is in reality skipping the logits predicted for the first token of the student answer. I think it should be prompt_lengths-1

kashif:
ok if you clean up the PR without eos i can look at the lengths and fix it...

gaetanlop:
When you call forward on an autoregressive transformer model, the logits at each position correspond to the predicted probability distribution for the next token, not the input token itself.

Let's take the following example:
input_ids = BOS P1 P2 1 2 3 EOS, where P means prompt
If we want to skip the prompt, the loss needs to be computed on 4 tokens at those positions: [X X O O O O X]

Logits at position P2 correspond to the predicted probability for token 1. When slicing with prompt_length, we are skipping the loss computation for the first token of the student response.

Do we agree?

kashif:
i think you are right i see now...

gaetanlop:
@kashif @lewtun I have made the required modifications to mask the loss on the right places (computes the loss on all the tokens generated by the student except the padding tokens). 

As mentioned above the previous implementation was skipping the first generated token in the loss computation. 

The loss masking is similar to the one in the `OnlineDPOTrainer`.

kashif (approved):

tests/test_gkd_trainer.py
```diff
@@ -103,21 +107,24 @@ def test_generate_on_policy_outputs(self):
...
         # Check shapes
         batch_size = len(prompts)
         self.assertEqual(new_input_ids.shape[0], batch_size)
         self.assertEqual(new_attention_mask.shape[0], batch_size)
+        self.assertEqual(new_labels.shape[0], batch_size)
 
         # Check types
         self.assertIsInstance(new_input_ids, torch.Tensor)
         self.assertIsInstance(new_attention_mask, torch.Tensor)
+        self.assertEqual(new_labels, torch.Tensor)
```
kashif, line 123:
```suggestion
        self.assertIsInstance(new_labels, torch.Tensor)
```

PR #2064, opened 2024-09-13
Use `transformers` utilities when possible

qgallouedec:
# What does this PR do?

A lot of utilities are redefined when we can just rely on `transformers`. I've removed most on them.
Most of https://github.com/huggingface/trl/blob/5c195eb51347ac4df7427eff4f1d636e860d13bf/tests/testing_utils.py

- `is_bitsandbytes_available`
- `is_liger_available`
- `is_peft_available`
- `is_pil_available`
- `is_wandb_available`
- `is_xpu_available`
- `require_peft`
- `require_bitsandbytes`
- `require_diffusers`
- `requires_pil`
- `require_wandb`
- `require_torch_multi_gpu`
- `require_torch_gpu`
- `require_torch_multi_xpu`
- `require_liger_kernel`
- `require_non_cpu`
- `require_multi_accelerator`

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


kashif:
@qgallouedec  if we are using the helpers from transformers we might need to update the minimal transformer version perhaps to something new?

qgallouedec:
Good point. In 4.39, we miss:

- `is_torch_mlu_available` (added in 4.40)
- `is_liger_kernel_available` (added in 4.45)

trl/import_utils.py
```diff
@@ -14,28 +14,38 @@
...
 
 if sys.version_info < (3, 8):
     _is_python_greater_3_8 = False
 else:
     _is_python_greater_3_8 = True
 
-
-def is_peft_available() -> bool:
-    return find_spec("peft") is not None
+_diffusers_available = _is_package_available("diffusers")
```
lewtun, line 29: Why don't we just call `_is_package_available()` within the methods below instead of defining these private variables?

lewtun (approved):
Thanks for the QoL improvement! I wonder if putting a lower bound on `transformers` leads to downstream issues in users who have a pinned, lower version of the lib? Otherwise LGTM

trl/import_utils.py
qgallouedec: This is how it's done in `transformers`.  I've opted for consistency.

https://github.com/huggingface/transformers/blob/main/src/transformers/utils/import_utils.py#L97

I don't know why they do that. Especially since it involves something ugly like this:

https://github.com/huggingface/trl/blob/f8811b6b4af81479909381b96ca9d879225efb9e/tests/test_no_peft.py#L78

qgallouedec:
> Thanks for the QoL improvement! I wonder if putting a lower bound on transformers leads to downstream issues in users who have a pinned, lower version of the lib? Otherwise LGTM

pip will probably outputs something like:

trl 0.11 requires transformers>=4.40, but you have transformers 4.35 which is incompatible.

trl/import_utils.py
lewtun: I see, ok let's follow their approach then - could you just add a comment which points to their convention for future us / maintainers?

qgallouedec: See #2073

PR #2070, opened 2024-09-15
[KTO] learning rate recomentations for kto

kashif:
# What does this PR do?

recommendation for the KTO learning rate to be small. See: https://x.com/ethayarajh/status/1834780782126153763

### DIFF
--- a/docs/source/kto_trainer.mdx
+++ b/docs/source/kto_trainer.mdx
@@ -64,7 +64,7 @@ The `beta` refers to the hyperparameter of the implicit reward, and the dataset 
 The `desirable_weight` and `undesirable_weight` refer to the weights placed on the losses for desirable/positive and undesirable/negative examples.
 By default, they are both 1. However, if you have more of one or the other, then you should upweight the less common type such that the ratio of (`desirable_weight` \\(\times\\) number of positives) to (`undesirable_weight` \\(\times\\) number of negatives) is in the range 1:1 to 4:3.
 
-It is also recommended to set the `learning_rate` to a smaller value, such as `5e-7` instead of larger values.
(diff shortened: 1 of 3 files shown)


qgallouedec:
What do you think about overriding the default lr of `TrainingArgument` in this case?  It seems to be more convenient for the user.

cc @kawine @corbt

```python
@dataclass
class KTOConfig(TrainingArguments):
    ...
    learning_rate: float = 5e-7
```

docs/source/kto_trainer.mdx
```diff
@@ -64,11 +64,14 @@ The `beta` refers to the hyperparameter of the implicit reward, and the dataset
 The `desirable_weight` and `undesirable_weight` refer to the weights placed on the losses for desirable/positive and undesirable/negative examples.
 By default, they are both 1. However, if you have more of one or the other, then you should upweight the less common type such that the ratio of (`desirable_weight` \\(\times\\) number of positives) to (`undesirable_weight` \\(\times\\) number of negatives) is in the range 1:1 to 4:3.
 
+It is also recommended to set the `learning_rate` to a smaller value, such as `5e-7` instead of larger values.
```
kawine, line 67: Can we expand on this a bit? e.g., 

"It is strongly recommended you use a learning rate between 5e-7 and 5e-6 with an effective batch size between 8 and 32, for both LoRA and full finetuning. Even if you are working with a small dataset, we do not recommend using a learning rate outside this range; instead, using smaller batch sizes and/or more training epochs will give you better results."

kawine:
thanks for taking the initiative @qgallouedec ! I just had one suggestion re: the docs.

docs/source/kto_trainer.mdx
kashif: done

kashif:
@qgallouedec  done

qgallouedec:
Thanks, for the record, we should extend this to all our configs, when necessary

qgallouedec (approved):

PR #2074, opened 2024-09-16
Use wrapped model for reference completions in `WinRateCallback` and set default `freq` to `eval_steps` in LogCompletionsCallback`

lewtun:
# What does this PR do?

While training some models with `GKDTrainer` I ran into an issue with ZeRO-3 and `WinRateCallback` where the initial model provided by the callback is _unwrapped_ and this incompatible with `unwrap_model_for_generation()`. I also realised that using `eval_steps` is far more intuitive for both `WinRateCallback` and `LogCompletionsCallback` because we usually reserve logging for fast computations like the loss, and `LogCompletionsCallback` produces a major bottleneck in the logging if we generate every `logging_steps`.

This PR fixes both issues

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/callbacks.py
+++ b/trl/trainer/callbacks.py
@@ -255,12 +255,12 @@ class WinRateCallback(TrainerCallback):
         accelerator = self.trainer.accelerator
         # Use the reference model if available, otherwise use the initial model
         model = getattr(self.trainer, "ref_model", None)
-        # At this point, there are two cases where `ref_model` is None: 
-        # 1. The method doesn't require a reference model. 
-        # 2. The method uses a reference model, but `ref_model` is set to None. 
-        #    This occurs when using PEFT, where the reference model can be obtained by simply disabling the model's adapter. 
-        #    In theory, we should disable the adapter here, but since it's zero-initialized at the start of training, 
-        #    the model behaves identically with or without the adapter. 
+        # At this point, there are two cases where `ref_model` is None:
+        # 1. The method doesn't require a reference model.
+        # 2. The method uses a reference model, but `ref_model` is set to None.
+        #    This occurs when using PEFT, where the reference model can be obtained by simply disabling the model's adapter.
+        #    In theory, we should disable the adapter here, but since it's zero-initialized at the start of training,
+        #    the model behaves identically with or without the adapter.
         #    Therefore, there's no need to explicitly disable it at this point.
         if model is None:
             model = self.trainer.model_wrapped
--- a/trl/trainer/callbacks.py
+++ b/trl/trainer/callbacks.py
@@ -253,7 +253,8 @@ class WinRateCallback(TrainerCallback):
         tokenizer = kwargs["tokenizer"]
         tokenizer.padding_side = "left"
         accelerator = self.trainer.accelerator
-        model = getattr(self.trainer, "ref_model", kwargs["model"])  # get the ref model if any, else use the model
+        # Use the reference model if available, otherwise use the initial model
+        model = getattr(self.trainer, "ref_model", self.trainer.model_wrapped)
         with accelerator.split_between_processes(self.eval_dataset["prompt"]) as prompts:
             self.ref_completions = _generate_completions(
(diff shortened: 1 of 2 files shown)


trl/trainer/callbacks.py
```diff
@@ -253,7 +253,8 @@ def on_train_begin(self, args: TrainingArguments, state: TrainerState, control:
         tokenizer = kwargs["tokenizer"]
         tokenizer.padding_side = "left"
         accelerator = self.trainer.accelerator
-        model = getattr(self.trainer, "ref_model", kwargs["model"])  # get the ref model if any, else use the model
+        # Use the reference model if available, otherwise use the initial model
+        model = getattr(self.trainer, "ref_model", self.trainer.model_wrapped)
```
qgallouedec, line 257: Does it work in the peft case? ie when ref_model is None?

lewtun: Oh good catch - let me see if I can write a unit test for this case

kashif (approved):

trl/trainer/callbacks.py
lewtun: Test added here with some minor cleanup to reuse the same expected winrates across tests: https://github.com/huggingface/trl/pull/2074/commits/592a0880dcd800e8cf2f42e3916f0a969f3f84fd

lewtun: OK to merge @qgallouedec or would you like some final changes?

trl/trainer/callbacks.py
```diff
@@ -253,7 +253,10 @@ def on_train_begin(self, args: TrainingArguments, state: TrainerState, control:
         tokenizer = kwargs["tokenizer"]
         tokenizer.padding_side = "left"
         accelerator = self.trainer.accelerator
-        model = getattr(self.trainer, "ref_model", kwargs["model"])  # get the ref model if any, else use the model
+        # Use the reference model if available, otherwise use the initial model
+        model = getattr(self.trainer, "ref_model", None)
+        if model is None:
+            model = self.trainer.model_wrapped
```
qgallouedec, line 259: The code handles **three cases** regarding the `ref_model` attribute, although it may initially seem like only two cases are involved:

1. **`ref_model` is an attribute and is a model**: for instance, when using **DPO without PEFT**. The `ref_model` is initialized as a regular model.

2. **`ref_model` is an attribute and is `None`**: for instance, when using **DPO with PEFT**. In such case, the ref model is implicitly defined as the model with the adapter disabled.  

3. **`ref_model` is an attribute and is `None`**: for instance, when using **CPO** (no `ref_model` for this method)

- **Cases 1 and 3** are trivial.
- **Case 2** deserves a comment: in theory, we should take the model (which is a PEFT model), and disable its adapter, to get the ref model. 

```python
if case_2:
    with model.disable_adapter():
        self.ref_completions = _generate_completions(...)
```

**However**, since we're at the start of training, the adapter is at zero, so the model with and without are the same.

IMO it's worth adding a comment here.

qgallouedec:
```suggestion
        # Use the reference model if available, otherwise use the initial model
        model = getattr(self.trainer, "ref_model", None)
        # At this point, there are two cases where `ref_model` is None: 
        # 1. The method doesn't require a reference model. 
        # 2. The method uses a reference model, but `ref_model` is set to None. 
        #    This occurs when using PEFT, where the reference model can be obtained by simply disabling the model's adapter. 
        #    In theory, we should disable the adapter here, but since it's zero-initialized at the start of training, 
        #    the model behaves identically with or without the adapter. 
        #    Therefore, there's no need to explicitly disable it at this point.
        if model is None:
            model = self.trainer.model_wrapped
```

qgallouedec (approved):
LGTM, just added a suggestion for a comment

trl/trainer/callbacks.py
lewtun: Great comment!

PR #2077, opened 2024-09-18
[WIP] Fix `logits/chosen` and `logits/rejected` metrics in `kto_trainer`.

PhilipMay:
The calculation of the `logits/chosen` and `logits/rejected` metrics in `kto_trainer` seem to be wrong. A `nansum()` followed by `nanmean()` applied to the `policy_rejected_logits` is wrong.

Our fix is to apply `nansum()` followed by an other `nansum()` and then devide the result by `count/chosen` or `count/rejected`.

### DIFF
--- a/trl/trainer/kto_trainer.py
+++ b/trl/trainer/kto_trainer.py
@@ -1201,13 +1201,17 @@ class KTOTrainer(Trainer):
         if all_num_chosen > 0:
             metrics["rewards/chosen_sum"] = self.accelerator.gather(chosen_rewards.nansum()).nansum().item()
             metrics["logps/chosen_sum"] = self.accelerator.gather(policy_chosen_logps.nansum()).nansum().item()
-            metrics["logits/chosen"] = self.accelerator.gather(policy_chosen_logits.nansum()).nansum().item() / all_num_chosen
+            metrics["logits/chosen"] = (
+                self.accelerator.gather(policy_chosen_logits.nansum()).nansum().item() / all_num_chosen#
+            )
             metrics["count/chosen"] = all_num_chosen
 
         if all_num_rejected > 0:
             metrics["rewards/rejected_sum"] = self.accelerator.gather(rejected_rewards.nansum()).nansum().item()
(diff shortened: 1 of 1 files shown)


kashif (approved):

PhilipMay:
Tagging @MAOJIASONG as the original author and @claralp and asking for a review.

qgallouedec:
Now I understand:

Let's take the following example:

```
              process 1   process 2
logits        [0, 1, 2]   [3, 4]
sum               3         7
gather              [3, 7]
mean                  5
```

Your proposed fix

```
              process 1   process 2
logits        [0, 1, 2]   [3, 4]
sum               3         7
gather              [3, 7]
sum                   10
/all_num_chosen        2
```

Would the following work?

```
              process 1   process 2
logits        [0, 1, 2]   [3, 4]
gather          [0, 1, 2, 3, 4]
sum                   10
/all_num_chosen        2
```

qgallouedec (approved):

MAOJIASONG:
> Now I understand:
> 
> Let's take the following example:
> 
> ```
>               process 1   process 2
> logits        [0, 1, 2]   [3, 4]
> sum               3         7
> gather              [3, 7]
> mean                  5
> ```
> 
> Your proposed fix
> 
> ```
>               process 1   process 2
> logits        [0, 1, 2]   [3, 4]
> sum               3         7
> gather              [3, 7]
> sum                   10
> /all_num_chosen        2
> ```
> 
> Would the following work?
> 
> ```
>               process 1   process 2
> logits        [0, 1, 2]   [3, 4]
> gather          [0, 1, 2, 3, 4]
> sum                   10
> /all_num_chosen        2
> ```
> 
> Probably, but too memory intensive, right?

May I ask what's the difference between the first one, and the proposed second one? why is it better to use `all_num_chosen` for division? @qgallouedec

qgallouedec:
Because otherwise the result is wrong (5)

PhilipMay:
> Probably, but too memory intensive, right?

@qgallouedec I am not sure about this to be honest.

MAOJIASONG:
> Because otherwise the result is wrong (5)

Thx, I misinterpreted the example. Thanks for point it out.

PR #2078, opened 2024-09-18
Default `dataset_text_field` to `"text"`

qgallouedec:
# What does this PR do?

Most datasets use `“text”` for the text field. Moreover, this is the column we document in trl as [standard language modeling] (https://huggingface.co/docs/trl/en/dataset_formats#overview-of-the-dataset-formats-and-types).
Consequently, it makes sense to use this column by default for `dataset_text_field`.
On the other hand, it allows us to drop `# needs dummy text field` as here:

https://github.com/huggingface/trl/blob/78249d9de46486a7fdb99c441ce0f52b9b0e1980/examples/scripts/sft_vlm.py#L57

### Implications

- The `dataset_text_field` can no longer be `None` in the `SFTConfig`. We then use the following logic in `SFTTrainer`: if `formatting_func` is provided, use it, otherwise, use `dataset_text_field`.
- Things are a little different in `ConstantLengthDataset` where `formatting_func` and `dataset_text_field` are `None` by default, and only one of them can be supplied.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


kashif (approved):

edbeeching (approved):

trl/trainer/sft_trainer.py
```diff
@@ -445,9 +445,9 @@ def _prepare_dataset(
         dataset,
         tokenizer,
         packing,
-        dataset_text_field,
+        dataset_text_field: str,
```
lewtun, line 448: not super important, but maybe we should type the rest of the function signature for consistency?

qgallouedec: Agree in general, but I didn't want to spend to much time on it since I plan to refactor SFT data preprocessing, to be more aligned with what's done in other trainers (which is clearer IMO)

PR #2108, opened 2024-09-24
[online-dpo] allow parse-args as list of floats

kashif:
# What does this PR do?

fixes #2106

PR #2142, opened 2024-09-30
[GKD] Set custom EOS tokens in generation config

lewtun:
# What does this PR do?

This PR exposes aligns the `eos_token_id` attribute from `transformers.GenerationConfig` in GKD to support distillation of Llama 3 models which use special tokens `<|eot_id|>` and `<|eom_id|>` instead of the EOS token `<|endoftext|>` to indicate the end of a turn or message. We do this by checking the model's `generation_config` attribute and including the `eos_token_id` values in the GKD trainer's `GenerationConfig`.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/examples/scripts/gkd.py
+++ b/examples/scripts/gkd.py
@@ -46,7 +46,7 @@ python examples/scripts/gkd.py \
 
 from accelerate import PartialState
 from datasets import load_dataset
-from transformers import AutoTokenizer, GenerationConfig
+from transformers import AutoTokenizer
 
 from trl import (
     GKDConfig,
@@ -93,6 +93,7 @@ if __name__ == "__main__":
 
     tokenizer = AutoTokenizer.from_pretrained(
         model_config.model_name_or_path,
+        revision=model_config.model_revision,
         trust_remote_code=model_config.trust_remote_code,
         padding_side="left",
     )
@@ -124,10 +125,7 @@ if __name__ == "__main__":
         tokenizer=tokenizer,
         peft_config=get_peft_config(model_config),
     )
-    generation_config = GenerationConfig(
-        max_new_tokens=training_args.max_new_tokens, do_sample=True, temperature=training_args.temperature
-    )
-    completions_callback = LogCompletionsCallback(trainer, generation_config, num_prompts=8)
+    completions_callback = LogCompletionsCallback(trainer, trainer.generation_config, num_prompts=8)
     trainer.add_callback(completions_callback)
     trainer.train()
 
--- a/trl/trainer/gkd_config.py
+++ b/trl/trainer/gkd_config.py
@@ -33,7 +33,7 @@ class GKDConfig(SFTConfig):
             beta is `0.0`, the loss is the KL divergence. When beta is `1.0`, the loss is the Inverse KL Divergence.
         max_new_tokens (`int`, *optional*, defaults to `128`):
             Maximum number of tokens to generate per completion.
-        eos_token_id (`Union[int, List[int]]`, *optional*):
+        eos_token_id (`Union[int, List[int]]`, *optional*, defaults to `None`):
             The id of the *end-of-sequence* token. Optionally, use a list to set multiple *end-of-sequence* tokens.
         teacher_model_name_or_path (`Optional[str]`, *optional*, defaults to `None`):
             Model name or path of the teacher model. If `None`, the teacher model will be the same as the model
--- a/trl/trainer/gkd_config.py
+++ b/trl/trainer/gkd_config.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from dataclasses import dataclass
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Optional
 
 from .sft_config import SFTConfig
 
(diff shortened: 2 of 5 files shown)


trl/trainer/gkd_config.py
```diff
@@ -33,6 +33,8 @@ class GKDConfig(SFTConfig):
             beta is `0.0`, the loss is the KL divergence. When beta is `1.0`, the loss is the Inverse KL Divergence.
         max_new_tokens (`int`, *optional*, defaults to `128`):
             Maximum number of tokens to generate per completion.
+        eos_token_id (`Union[int, List[int]]`, *optional*):
```
lewtun, line 36: In `transformers` the supported types are `int` and `List[int]` but the HF arg parser only accepts `Union[X, NoneType]`. Should I refactor the arg parser or just abuse the types a bit here?

kashif (approved):

trl/trainer/gkd_config.py
edbeeching: Hmm I have never tried, `Optional[Union[int, List[int]]]`  is not possible? Otherwise just use Optional[List[int]]? and for the case with 1 token use a list of size 1?

Also unless am referring to the wrong model, these tokens are exposed in the [llama3 gen config](https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct/blob/main/generation_config.json) ?

trl/trainer/gkd_trainer.py
```diff
@@ -100,6 +100,7 @@ def __init__(
             do_sample=True,
             top_k=0,
             use_cache=False if args.gradient_checkpointing else True,
+            eos_token_id=args.eos_token_id,
```
edbeeching, line 103: this can't be `eos_token_id=model_config.eos_token_id` ?

or better yet, `self.generation_config = model.generation_config`
then override temperature etc.

trl/trainer/gkd_config.py
qgallouedec: You can do the as https://github.com/huggingface/trl/pull/2108

qgallouedec:
```suggestion
        eos_token_id (`Union[int, List[int]]`, *optional*, default to `None`):
```

trl/trainer/utils.py
```diff
@@ -1380,9 +1380,9 @@ def generate_model_card(
     tags: Union[str, List[str], None],
     wandb_url: Optional[str],
     trainer_name: str,
-    trainer_citation: Optional[str],
-    paper_title: Optional[str],
-    paper_id: Optional[str],
+    trainer_citation: Optional[str] = None,
+    paper_title: Optional[str] = None,
+    paper_id: Optional[str] = None,
```
qgallouedec, line 1385: Please also update the doc with ```defaults to `None`)```.

qgallouedec:
The generation arguments are part of the training arguments. So I think it's more natural to include them in training args.

qgallouedec:
Can you please also update `tests/test_trainers_args.py`?

edbeeching:
> The generation arguments are part of the training arguments. So I think it's more natural to include them in training args.

Are you referring to my suggestion to use the model's gen config as the default then override with the user config if required?

qgallouedec:
No sorry, I was answering to

> Should we expose the full GenerationConfig to users for all online trainers? If so, should it be part of the config classes or a separate arg in the trainer?

trl/trainer/gkd_config.py
lewtun: Ah nice, thanks for the pointer @qgallouedec

trl/trainer/gkd_trainer.py
lewtun: Ah good point, that could well work since I see these IDs are provided in the model config directly: https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct/blob/0e9e39f249a16976918f6564b8830bc894c89659/config.json#L8

trl/trainer/gkd_trainer.py
```diff
@@ -100,6 +100,7 @@ def __init__(
             do_sample=True,
             top_k=0,
             use_cache=False if args.gradient_checkpointing else True,
+            pad_token_id=self.tokenizer.pad_token_id,
```
lewtun, line 103: Needed to avoid a huge stream of warnings from `transformers`

examples/scripts/gkd.py
```diff
@@ -124,10 +125,7 @@
         tokenizer=tokenizer,
         peft_config=get_peft_config(model_config),
     )
-    generation_config = GenerationConfig(
-        max_new_tokens=training_args.max_new_tokens, do_sample=True, temperature=training_args.temperature
-    )
-    completions_callback = LogCompletionsCallback(trainer, generation_config, num_prompts=8)
+    completions_callback = LogCompletionsCallback(trainer, trainer.generation_config, num_prompts=8)
```
lewtun, line 128: I realised this is simpler and what one wants to ensure alignment across training / logging

examples/scripts/gkd.py
```diff
@@ -93,6 +93,7 @@
 
     tokenizer = AutoTokenizer.from_pretrained(
         model_config.model_name_or_path,
+        revision=model_config.model_revision,
```
lewtun, line 96: Fix for cases where you need to load from a revision

lewtun:
@edbeeching @qgallouedec I found a cleaner approach that doesn't require one to expose this arg to the end user and simply reads it directly from the model's `generation_config.json`. 

Let me know what you think!

examples/scripts/gkd.py
qgallouedec: Good point, I'll extend this to all examples

edbeeching (approved):
Thanks for the updates, LGTM

qgallouedec (approved):
LGTM!

Issue #2110, opened 2024-09-24
[Reward Modelling] Add support for process / stepwise supervision


lewtun:
### Feature request

Process / stepwise supervision is a powerful way to train better reward models, where instead of optimising on the whole completion, one optimises each step (e.g. in CoT a score is assigned to each step). It would be good to support this in TRL as PRMs will likely become popular now that OpenAI's `o1` model seems to have used them. 

This will likely require a new `StepwiseRewardTrainer`, but it might be possible to include in the existing `RewardTrainer` if the changes aren't too drastic (preferred).

* Reference papers: [[1]](https://arxiv.org/abs/2305.20050), [[2]](https://arxiv.org/abs/2406.06592)
* Dataset to test with: https://github.com/openai/prm800k

### Motivation

With online methods coming back in fashion, the need for good reward models is increasing. Having support for stepwise reward modelling would help bridge the gap with proprietary models. 

### Your contribution

Open to the community :)

world2025:
@qgallouedec can i use data format like [openbookqa](https://huggingface.co/datasets/allenai/openbookqa)，one prompt has 4-9 response like paper in InstructGPT,thanks

PR #2148, opened 2024-10-01
🪜 Stepwise supervision dataset type

qgallouedec:
This PR aims to create a standard for stepwise data

Related: #2110

Chosen name: *Stepwise supervision*
Columns: `"prompt"`, `"completions"`, `"labels"`
Structure:
```python
standard_example = {
    "prompt": "Beautiful is better than",
    "completions": [", let me think...", " ugly."],
    "labels": [False, True],
}
```

## TODO

- [x] Update dataset format doc
- [x] Update internal testing dataset (zen)
- [x] Create a script for having our getting started script (from [PRM800K](https://huggingface.co/datasets/tasksource/PRM800K)?)


trl/data_utils.py
```diff
@@ -60,6 +60,49 @@ def is_conversational(example: Dict[str, Any]) -> bool:
...
+
+    merged_messages = [messages[0]]  # Initialize with the first message
+
+    for i in range(1, len(messages)):
+        current_message = messages[i]
+        previous_message = merged_messages[-1]
+
+        if current_message["role"] == previous_message["role"]:
+            # Merge the content of the current message into the previous one
+            previous_message["content"] += "\n" + current_message["content"]
```
gaetanlop, line 98: Joining with `"\n"` might lead to problems when trying to split back the different steps or find the end of each step if one of the messages already contain a new line character.

trl/data_utils.py
```diff
@@ -155,6 +204,7 @@ def maybe_apply_chat_template(
                 - Preference dataset: `"prompt"`, `"chosen"`, and `"rejected"`.
                 - Preference dataset with implicit prompt: `"chosen"` and `"rejected"`.
                 - Unpaired preference dataset: `"prompt"`, `"completion"`, and `"label"`.
+                - [Name to find]: `"prompt"`, `"completion"`, and `"labels"`.
```
gaetanlop, line 207: What do you think of `Stepwise preference dataset`?

trl/data_utils.py
gaetanlop: One solution for this would be to use a `step_separator` that the user woud need to set. We could verify that none of the `messages` contain this `step_separator`.

qgallouedec: Why would you need to "split back"?

trl/data_utils.py
qgallouedec: stepwise: ok
preference: not really, as preference means that you've chosen and rejected. Here it's more "unpaired preference". 
stepwise unpaired preference dataset?

gaetanlop: Yes, good idea

trl/data_utils.py
gaetanlop: Sorry, it is not really splitting back the steps but finding the last token of each step (for the `Process-supervised Reward Model` training). That will cause problems if the new line character is already contained in one of the steps.

We need to find the last token of each steps as PRMs are trained "to predict the correctness of each step after the last token in each step" (https://arxiv.org/pdf/2305.20050)

qgallouedec:
The main intention was to implement _stepwise process supervision_. The doc isn't probably clear enough. Feel free to recommend/directly change anything that made you think it was _multi-step preferences_.

trl/data_utils.py
lewtun: I agree with @gaetanlop - we should only concatenate the messages and leave it up to the user to insert the appropriate delimiters like `\n`

trl/data_utils.py
lewtun: As mentioned above, I am not sure if "stepwise unpaired preference" is confusing since (a) process supervision isn't the same as preference learning and (b) there exists multi-step preference learning and people (like me!) might conflate the two

lewtun (commented):
Thanks for working on a standard format for stepwise supervision @qgallouedec ! The main comment I have is whether this should be grouped with preference learning or not and whether we should bother with the conversational format for now since I've never seen this used in practice (everything is single-turn thus far)

docs/source/dataset_formats.mdx
```diff
@@ -77,6 +77,20 @@ This guide provides an overview of the dataset formats and types supported by ea
  "label": False}</code></pre>
     </td>
   </tr>
+  </tr>
+    <td>Stepwise unpaired preference</td>
+    <td>
+      <pre><code>{"prompt": "Blue light",
+ "completions": [" scatters more in the atmosphere,", " so the sky is green."],
+ "labels": [True, False]}</code></pre>
```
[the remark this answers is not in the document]
lewtun: Note that labels can also be floats, but I wasn't sure where to best document that (maybe the trainer?)

trl/data_utils.py
qgallouedec: I don't think we need to "find the last token of each step" since the steps are already separated in the dataset. During the training step, you'll get a batch with the prompt, and the completions steps. Example (pseudo code):

```python
# in the train step
import torch
from itertools import accumulate

# Batch data
batch = {
    "prompt": torch.tensor([1, 2, 3, 4]),
    "completions": [torch.tensor([5, 6, 7]), torch.tensor([8, 9, 10, 11])]
}

# Get the separator indexes
completion_lengths = torch.tensor([len(step) for step in batch["completions"]])
indexes = torch.tensor([len(batch["prompt"]) + i - 1 for i in accumulate(completion_lengths)])  # [6, 10]

# Concatenate the completions
completion = torch.cat(batch["completions"])  # [5, 6, 7, 8, 9, 10, 11]

# Concatenate prompt and completion
input = torch.cat([batch["prompt"], completion])  # [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]

# Get the output from the reward model
output = reward_model(input)
scores = output[indexes]  # [0.345, 0.897]
```

docs/source/dataset_formats.mdx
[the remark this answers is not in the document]
qgallouedec: idk either. Maybe in a note in the dedicated section of the doc? Let's leave this conversation open for now.

trl/data_utils.py
qgallouedec: As a consequence of my previous comment, we don't use the `_merge_same_role_messages` function. I'm having this deleted accordingly.

qgallouedec: Done in b1d9f2d7b383b9c2bcf728beae2f4d686c6f3b6a

docs/source/dataset_formats.mdx
```diff
@@ -192,7 +232,32 @@ Some preference datasets can be found with [the tag `dpo` on Hugging Face Hub](h
...
+A stepwise unpaired preference dataset is similar to an [unpaired preference](#unpaired-preference) dataset but includes multiple steps of completions, each with its own label. This structure is useful for tasks that need detailed, step-by-step labeling, such as reasoning tasks. By evaluating each step separately and providing targeted labels, this approach helps identify precisely where the reasoning is correct and where errors occur, allowing for targeted feedback on each part of the reasoning process.
+
+```python
+# Standard format
+stepwise_unpaired_preference_example = {
+    "prompt": "Blue light",
+    "completions": [" scatters more in the atmosphere,", " so the sky is green."],
+    "labels": [True, False]
+}
+# Conversational format
```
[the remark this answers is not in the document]
qgallouedec: removed in fe98b5b71328e20731bcb922566a4b724137aa17

docs/source/dataset_formats.mdx
```diff
@@ -338,14 +403,15 @@ This section provides example code to help you convert between different dataset
...
-| Unpaired preference             | [🔗](#from-unpaired-preference-to-language-modeling-dataset)             | [🔗](#from-unpaired-preference-to-prompt-completion-dataset)             | [🔗](#from-unpaired-preference-to-prompt-only-dataset)             | N/A                                                       | N/A                                                       | N/A                                                                       |
+| From \ To                       | Language modeling                                                       | Prompt-completion                                                       | Prompt-only                                                       | Preference with implicit prompt                           | Preference                                                | Unpaired preference                                                       | Stepwise unpaired preference |
```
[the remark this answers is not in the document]
qgallouedec: Done in fe98b5b71328e20731bcb922566a4b724137aa17

lewtun (approved):
Thanks for iterating @qgallouedec - this looks great now and the PRM dataset is a very useful starting point for the PRM PR @gaetanlop

PR #2153, opened 2024-10-02
minor KTO setting changes + KL batch size

kawine:
# What does this PR do?
 
- change the default learning rate from 5e-7 to 1e-6; for models 7B and up (which is what most people use w/ or w/o LoRA), i this works better with the default beta=0.1 than an lr=5e-7, which can lead to slow learning
- add explanations in the docs on choosing the learning rate
- (important) change the batch size of the KL data to be the per-step batch size, not the effective batch size; the current implementation leads to worse results when gradient accumulation is used
- allow dropout to be used, like in DPOTrainer

cc @kashif @qgallouedec

### DIFF
--- a/docs/source/kto_trainer.mdx
+++ b/docs/source/kto_trainer.mdx
@@ -5,6 +5,7 @@ For a full example have a look at  [`examples/scripts/kto.py`].
 
 Depending on how good your base model is, you may or may not need to do SFT before KTO.
 This is different from standard RLHF and DPO, which always require SFT.
+You can also train with imbalanced data (more chosen than rejected examples, or vice-versa), but you will need to adjust hyperparameters accordingly (see below).
 
 ## Expected dataset format
 
@@ -49,7 +50,8 @@ kto_dataset_dict = {
 ```
 
 where the `prompt` contains the context inputs, `completion` contains the corresponding responses and `label` contains the corresponding flag that indicates if the generated completion is desired (`True`) or undesired (`False`).
-A prompt can have multiple responses and this is reflected in the entries being repeated in the dictionary's value arrays. It is required that the dataset contains at least one desirable and one undesirable completion.
+A prompt can have multiple responses and this is reflected in the entries being repeated in the dictionary's value arrays. 
+In theory, the dataset must contain at least one desirable and one undesirable completion; however, some people have success running KTO on _only_ desirable or undesirable data (in the latter case, it is best to use a conservative learning rate).
 
 
 ## Expected model format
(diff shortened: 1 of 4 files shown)


kashif (approved):

kashif:
in the `kto_config.py` docstrings can you kindly add:

```py
disable_dropout (`bool`, *optional*, defaults to `True`):
            Whether to disable dropout in the model.
```

kawine:
> in the `kto_config.py` docstrings can you kindly add:
> 
> ```python
> disable_dropout (`bool`, *optional*, defaults to `True`):
>             Whether to disable dropout in the model.
> ```

done!

trl/trainer/kto_trainer.py
```diff
@@ -73,7 +73,11 @@
 
 
 def _get_kl_dataset(batch: Dict[str, List[Any]]) -> Dict[str, List[Any]]:
-    """Creates mismatched pairs of prompts and completions for the KL dataset by adding a +1 offset to the order of completions."""
+    """
+    Creates mismatched pairs of prompts and completions for the KL dataset by adding a +1 offset to the order of completions.
+    For best results, mismatched outputs y' used to estimate the KL term for a batch should be the same set as the matched
```
kashif, line 78:
```suggestion
    For best results, the mismatched outputs y' used to estimate the KL term for a batch should be the same set as the matched
```

docs/source/kto_trainer.mdx
```diff
@@ -61,13 +63,17 @@ The KTO trainer expects a model of `AutoModelForCausalLM`, compared to PPO that
...
 By default, they are both 1. However, if you have more of one or the other, then you should upweight the less common type such that the ratio of (`desirable_weight` \\(\times\\) number of positives) to (`undesirable_weight` \\(\times\\) number of negatives) is in the range 1:1 to 4:3.
 
 <Tip>
-It is strongly recommended you use a learning rate between `5e-7` and `5e-6` with an effective batch size between `8` and `32`, for both LoRA and full finetuning. Even if you are working with a small dataset, we do not recommend using a learning rate outside this range; instead, using smaller batch sizes and/or more training epochs will give you better results.
+Every choice of `beta` has a maximum learning rate it will tolerate before learning degenerates. For the default `beta = 0.1', this learning rate is `1e-6` for most models. The lower beta is, the lower your learning rate should be. In general, we strongly recommend a learning rate between `5e-7` and `5e-6`. Even if you are working with a small dataset, we do not recommend using a learning rate outside this range; instead, use more epochs.
```
kashif, line 72:
```suggestion
Every choice of `beta` has a maximum learning rate it will tolerate before learning degenerates. For the default `beta = 0.1', this learning rate is `1e-6` for most models. The lower the beta is, the lower your learning rate should be. In general, we strongly recommend a learning rate between `5e-7` and `5e-6`. Even if you are working with a small dataset, we do not recommend using a learning rate outside this range; instead, use more epochs. 
```

kashif (approved):

PR #2155, opened 2024-10-02
[CGPO] Calibrated reward

gaetanlop:
# What does this PR do?

Adds a `get_calibrated_reward` function as introduced in the CGPO paper of Meta. Please refer to equation 5 in section 4.1.1 for more information on this (https://arxiv.org/pdf/2409.20370).

This PR should be part of a set of PRs to incorporate CGPO in trl.

Fixes # (issue)

#2156 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@kashif @lewtun

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -169,3 +169,6 @@ class TestGenerateModelCard(unittest.TestCase):
         assert "my_model" in card_text
         assert 'pipeline("text-generation", model="username/my_hub_model", device="cuda")' in card_text
         assert "My Trainer" in card_text
+
+class TestGetReward(unittest.TestCase):
+    pass
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -15,12 +15,18 @@
 import unittest
 
 import torch
-from transformers import AutoTokenizer
+from transformers import AutoTokenizer, AutoModelForSequenceClassification
 from transformers.testing_utils import require_peft
 from transformers.utils import is_peft_available
 
 from trl.trainer.model_config import ModelConfig
-from trl.trainer.utils import decode_and_strip_padding, generate_model_card, get_peft_config, pad
+from trl.trainer.utils import (
+    decode_and_strip_padding,
+    generate_model_card,
+    get_peft_config,
+    pad,
+    get_calibrated_reward,
+)
 
 
 if is_peft_available():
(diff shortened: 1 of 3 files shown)


kashif:
the way i understood the calibrated reward was that the scores from a reward model might not be comparable across different completions given a prompt, and thus for some completion baseline or ground-truth completion `ā` for a prompt `s` and completion `a` the calibrated reward should be: `Rcalib(s, a) = σ(reward_model(s, a) − reward_model(s, ā))`

gaetanlop:
Thanks for looking at it @kashif. Your code and mine are exactly the same except for the `missing_eos_penalty` part that I did not put in the function to be consistent with your function `get_reward` (we can handle the `missing_eos_penalty` part in the potential `CGPOTrainer` as done in the `OnlineDPOTrainer`). Apart from that, we have the same implementation. You are assuming that the `ground_truth_completion` do not contain the prompt, while I am assuming it contains it. 

Also, I am computing the reward for both `a` and `ā` in a single forward pass by concatenating them, while you are computing it separately for `a` and `ā`. 

Example using your naming conventions and assuming both `query_responses` and `baselines_responses` contain the prompt:
```python
batch_size = query_responses.shape[0]
concatenated_responses = torch.cat(
        (query_responses, baseline_responses),
        dim=0,
    )

reward_logits, final_rewards, sequence_lengths = get_reward(
    model, concatenated_responses, pad_token_id, context_length
)

generated_scores, ground_truth_scores = final_rewards.split(batch_size, dim=0)

final_rewards = F.sigmoid(generated_scores-ground_truth_scores)
```

For the returns, I am also returning all the `calibrated_logits`, and the `sequence_lengths` alongside what you are returning (final calibrated reward) to be consistent with the `get_reward` function of trl.



My implementation lacked a sigmoid function for the reward_logits thought. 

Am I correct?

kashif:
ah right right! you are right!

kashif:
so the reason i have the stuff split is because of padding... when i join the two different completions i have to pad them together while its slightly easier to pad each completion... and then i was scared if by contacting the memory needs might be too much for largish reward models... but yes makes sense

gaetanlop:
Yes @kashif 100% agree, there are pros and cons for both methods. It also depends on the distributed training strategy you are using to train the model.

In any case, I checked the `trl` code base and it seems you adopted this concatenation method in the `OnlineDPOTrainer` https://github.com/huggingface/trl/blob/78249d9de46486a7fdb99c441ce0f52b9b0e1980/trl/trainer/online_dpo_trainer.py#L416-L427 

I think we should keep it this way. Wdyt?

gaetanlop:
Closing in favor of #2190

PR #2158, opened 2024-10-03
Remove graph breaks for torch.compile() in padding free branch in DataCollatorForCompletionOnlyLM

Abhishek-TAMU:
# What does this PR do?

This PR adds `cu_seq_lens_q`, `cu_seq_lens_k`, `max_length_k`, `max_length_q` to the batch in `DataCollatorForCompletionOnlyLM`. This, together with a PR in transformers (link to be added), removes graph breaks in padding-free tuning, allowing for maximum performance to be obtained.
Specifically, these parameters should be generated here (this PR change), outside of the transformers loop, as they incur a cpu-gpu sync that is unavoidable. Otherwise, this cpu-gpu sync happens [here](https://github.com/huggingface/transformers/blob/d7950bff82b18c823193d17d72188c5e46d06c83/src/transformers/modeling_flash_attention_utils.py#L272), inside the attention call which causes graph breaks and hence the transformers PR removes this call to remove all graph breaks when `torch_compile` flag is turned on in Training [arguments](https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.TrainingArguments.torch_compile) to use in SFTTrainer.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


qgallouedec:
Hi, thanks for the PR.
Can you provide the link of the PR in transformers?

qgallouedec:
Could you provide a simple test to:

1. Confirm that it is a case of non-functioning.
2. Verify that this addition resolves it.

It might also be helpful to add a few comments, as these lines are unclear without context.

Abhishek-TAMU:
Thank you @qgallouedec for the review. This is the related transformers [PR](https://github.com/huggingface/transformers/pull/33932) which is approved and merged.

I added 2 test cases. One where Tuning fails `with Padding` and another where doesn't fail `without padding`.

Abhishek-TAMU:
@kashif @qgallouedec Could you possibly review this PR ? Thank you!

tests/test_sft_trainer.py
```diff
@@ -654,6 +654,50 @@ def test_data_collator_completion_lm_with_multiple_text(self):
...
+    def test_data_collator_completion_lm_without_padding(self):
+        os.environ["CUDA_VISIBLE_DEVICES"]="0"
+        model_id = "trl-internal-testing/tiny-random-LlamaForCausalLM"
+        torch_dtype = getattr(torch, "bfloat16", None)
+        model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch_dtype, attn_implementation="flash_attention_2")
+        tokenizer = AutoTokenizer.from_pretrained(model_id, use_fast=True)
+
+        formatted_dataset = lambda example: {
+            "output": f"### prompt:\n{example['prompt'].strip()}\n\n### completion:\n{example['completion'].strip()}{tokenizer.eos_token}"
+        }
```
qgallouedec, line 666: is dataset formatting required here, or can we drop it?

Abhishek-TAMU: dataset formatting is required because the `SFTTrainer` and `DataCollatorForCompletionOnlyLM` expect the dataset to have a specific format—a single text field that combines both the prompt and the completion in a way the model can understand. This function includes both the prompt and completion, ensuring the data collator can correctly identify where the completion starts using the `response_template`.

tests/test_sft_trainer.py
```diff
@@ -654,6 +654,50 @@ def test_data_collator_completion_lm_with_multiple_text(self):
             result_text = tokenizer.decode(batch["input_ids"][i, last_pad_idx + 1 :])
             self.assertEqual(result_text, "I have not been masked correctly.")
 
+    def test_data_collator_completion_lm_without_padding(self):
+        os.environ["CUDA_VISIBLE_DEVICES"]="0"
```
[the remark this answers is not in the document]
Abhishek-TAMU: Due to usage of `flash_attention_2` it would work only on GPU.

Abhishek-TAMU:
Thank you @qgallouedec for looking into this. Sharing you the code which would produce graph break.
Using latest release version of `transformers` (which doesn't have https://github.com/huggingface/transformers/pull/33932 changes) and latest `trl` including changes from this PR.

If this change https://github.com/huggingface/transformers/pull/33932 is used in transformers then Graph break could be avoided. 

```
import os, tempfile, torch
from transformers import (
    AutoModelForCausalLM,
    AutoTokenizer)
from trl.trainer import DataCollatorForCompletionOnlyLM

standard_prompt_completion_dataset = load_dataset(
    "trl-internal-testing/zen", "standard_prompt_completion"
)

os.environ["CUDA_VISIBLE_DEVICES"]="0"
os.environ["CUDA_HOME"]="/home/tuning/.local/cuda-12.1"
model_id = "trl-internal-testing/tiny-random-LlamaForCausalLM"
torch_dtype = getattr(torch, "bfloat16", None)
model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch_dtype, attn_implementation="flash_attention_2")
tokenizer = AutoTokenizer.from_pretrained(model_id, use_fast=True)

formatted_dataset = lambda example: {
"output": f"### prompt:\n{example['prompt'].strip()}\n\n### completion:\n{example['completion'].strip()}{tokenizer.eos_token}"
}

train_dataset = standard_prompt_completion_dataset["train"].map(formatted_dataset)

response_template = "### completion:\n"
data_collator = DataCollatorForCompletionOnlyLM(response_template, tokenizer=tokenizer, padding_free=True)

    training_args = SFTConfig(
        output_dir=tmp_dir,
        dataloader_drop_last=True,
        max_steps=2,
        save_steps=2,
        learning_rate=1e-5,
        dataset_text_field="output",
        torch_compile=True,
        torch_compile_backend="inductor",
        torch_compile_mode="default"
    )

    trainer = SFTTrainer(
        model=model,
        tokenizer=tokenizer,
        train_dataset=train_dataset,
        data_collator=data_collator,
        args=training_args,
    )

    # with assertRaises(Exception):
    trainer.train()
    del os.environ["CUDA_VISIBLE_DEVICES"]
```

Abhishek-TAMU:
Hi @qgallouedec, were you able to reproduce the initial error with this MRE ?

ArthurZucker:
The loss error was fixed in transformers

tests/test_data_collator_completion_only.py
```diff
@@ -114,7 +114,7 @@ def test_padding_free(self):
         inst1 = "### System: You are a helpful assistant.\n\n### User: How much is 2+2?\n\n### Assistant: 2+2 equals 4"
         inst2 = "### System: You are a honest and helpful assistant.\n\n### User: What is the answer of 22x22?\n\n### Assistant: 22x22 equals 484"
 
-        response_template = "\n### Assistant:"
+        response_template = "\n\n### Assistant:"
```
qgallouedec, line 117: otherwise the template isn't found (`\n\n` is jointly tokenized)

qgallouedec (approved):

qgallouedec:
Thanks and sorry for the delay

PR #2174, opened 2024-10-04
🕊️ Migration `PPOv2` -> `PPO`

qgallouedec:
# What does this PR do?

Follows #2016

Fixes # (issue)

It basically:
- Deletes `PPO` (including trainer, config, tests, examples, but not all, see my [comment](https://github.com/huggingface/trl/pull/2174#issuecomment-2399843454))
- Renames `PPOv2` to `PPO` (in trainer, config, examples, tests, doc, etc.)
- Creates an alias `PPOv2` for `PPO`

I've done my best, but it may break some code. To put things in perspective, PPO isn't as popular as it used to be, at the time of writing, the last PPO model created on the hub was two weeks ago.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


qgallouedec:
While working on this, I noticed that many of our documents and notebooks are based on PPO and may be outdated, relying on older datasets, models, references, etc. I believe updating all of our examples and guides falls outside the scope of this PR.

I recommend the following:
- For now, keep the existing outdated documentation, guides, and notebooks as they are.
- Begin a separate effort to rewrite these materials in future PR(s), primarily starting from scratch.

docs/source/detoxifying_a_lm.mdx
```diff
@@ -98,19 +98,15 @@ model = AutoModelForCausalLM.from_pretrained("EleutherAI/gpt-j-6B", torch_dtype=
 
 and the optimizer will take care of computing the gradients in `bfloat16` precision. Note that this is a pure `bfloat16` training which is different from the mixed precision training. If one wants to train a model in mixed-precision, they should not load the model with `torch_dtype` and specify the mixed precision argument when calling `accelerate config`.
 
-- Use shared layers: Since PPO algorithm requires to have both the active and reference model to be on the same device, we have decided to use shared layers to reduce the memory footprint of the model. This can be achieved by just speifying `num_shared_layers` argument when creating a `PPOTrainer`:
+- Use shared layers: Since PPO algorithm requires to have both the active and reference model to be on the same device, we have decided to use shared layers to reduce the memory footprint of the model. This can be achieved by just speifying `num_shared_layers` argument when calling `create_reference_model` function. For example, if you want to share the first 6 layers of the model, you can do it like this:
```
edbeeching, line 101:
```suggestion
- Use shared layers: Since PPO algorithm requires to have both the active and reference model to be on the same device, we have decided to use shared layers to reduce the memory footprint of the model. This can be achieved by specifying `num_shared_layers` argument when calling `create_reference_model` function. For example, if you want to share the first 6 layers of the model, you can do it like this:
```

docs/source/customization.mdx
```diff
@@ -46,171 +46,178 @@ else:
...
-tokenizer = GPT2Tokenizer.from_pretrained('gpt2')
-
-# 2. define config
-ppo_config = {'batch_size': 1, 'learning_rate':1e-5}
-config = PPOConfig(**ppo_config)
-
-
-# 2. Create optimizer
-optimizer = torch.optim.SGD(model.parameters(), lr=config.learning_rate)
+By default, the `DPOTrainer` creates a `torch.optim.Adam` optimizer. You can create and define a different optimizer and pass it to `DPOTrainer`:
```
lewtun, line 51: nit:
```suggestion
By default, the `DPOTrainer` creates a `torch.optim.AdamW` optimizer. You can create and define a different optimizer and pass it to `DPOTrainer` as follows:
```

Since we subclass `transformers.Trainer` nowadays, I'd be tempted to remove this part altogether since it's the same logic in `transformers`

trl/trainer/ppov2_config.py
```diff
@@ -12,51 +12,17 @@
...
-    reward_model_path: str = "EleutherAI/pythia-160m"
-    num_ppo_epochs: int = 4
-    whiten_rewards: bool = False
-    kl_coef: float = 0.05
-    cliprange: float = 0.2
-    vf_coef: float = 0.1
-    cliprange_value: float = 0.2
-    gamma: float = 1.0
-    lam: float = 0.95
+# Define an alias for PPOv2Trainer that raises a warning
```
edbeeching, line 20:
```suggestion
# Define an alias for PPOv2Config that raises a warning
```

docs/source/customization.mdx
```diff
@@ -46,171 +46,178 @@ else:
...
+    model=model,
+    args=training_args,
+    train_dataset=dataset,
+    tokenizer=tokenizer,
+    optimizers=(optimizer, None),
+)
+trainer.train()
 ```
 
+### Use 8-bit optimizer
```
lewtun, line 76: Same here, I think this can be removed in favour of this: https://huggingface.co/docs/transformers/v4.45.2/en/perf_train_gpu_one#optimizer-choice

trl/trainer/ppo_config.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 The HuggingFace Team. All rights reserved.
```
qgallouedec, line 1: This one is just the renaming of ppov2_config. the diff view makes no sense

tests/test_ppo_trainer.py
```diff
@@ -11,1271 +11,53 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
qgallouedec, line 11: This one is just the tests/test_ppov2_trainer.py renamed. The diff view makes no sense

docs/source/customization.mdx
```diff
@@ -46,171 +46,178 @@ else:
 Consult the 🤗 Accelerate [documentation](https://huggingface.co/docs/accelerate/usage_guides/deepspeed) for more information about the DeepSpeed plugin.
 
 
-## Use different optimizers
+## Use different optimizers and schedulers
```
lewtun, line 49: Lots of this doc come from the time when we just had `PPOTrainer` which wasn't a subclass of `transformers`. I think we should remove anything that can be achieved on the `transformers` side, perhaps with an opening comment like:

"Every trainer in TRL is a subclass of the `Trainer` from 🤗 Transformers. This means you can change the optimizer or learning rate scheduler by simply changing the corresponding attributes of the TRL trainer configuration class. See the 🤗 Transformers [documentation](https://huggingface.co/docs/transformers/perf_train_gpu_one#optimizer-choice) for more details."

Note we can also nuke the deepspeed doc above since that's a relic from the old PPO days

docs/source/customization.mdx
```diff
@@ -46,171 +46,178 @@ else:
...
 ```
 
 ## Pass 8-bit reference models 
  
-<div>
-
-Since `trl` supports all key word arguments when loading a model from `transformers` using `from_pretrained`, you can also leverage `load_in_8bit` from `transformers` for more memory efficient fine-tuning.
+Since `trl` supports all keyword arguments when loading a model from `transformers` using `from_pretrained`, you can also leverage `load_in_8bit` from `transformers` for more memory efficient fine-tuning.
 
 Read more about 8-bit model loading in `transformers` [here](https://huggingface.co/docs/transformers/perf_infer_gpu_one#bitsandbytes-integration-for-int8-mixedprecision-matrix-decomposition).
```
lewtun, line 194:
```suggestion
Read more about 8-bit model loading in `transformers` [here](https://huggingface.co/docs/transformers/v4.45.2/en/peft#load-in-8bit-or-4bit).
```

docs/source/customization.mdx
```diff
@@ -46,171 +46,178 @@ else:
...
-model = AutoModelForCausalLMWithValueHead.from_pretrained('bigscience/bloom-560m')
-ref_model = AutoModelForCausalLMWithValueHead.from_pretrained('bigscience/bloom-560m', device_map="auto", load_in_8bit=True)
-tokenizer = AutoTokenizer.from_pretrained('bigscience/bloom-560m')
-
-# 2. initialize trainer
-ppo_config = {'batch_size': 1}
-config = PPOConfig(**ppo_config)
-ppo_trainer = PPOTrainer(config, model, ref_model, tokenizer)
+from datasets import load_dataset
+from transformers import AutoModelForCausalLM, AutoTokenizer
```
lewtun, line 198:
```suggestion
from transformers import AutoModelForCausalLM, AutoTokenizer, BitsAndBytesConfig
```

docs/source/customization.mdx
```diff
@@ -46,171 +46,178 @@ else:
...
+from trl import DPOConfig, DPOTrainer
+
+model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen2.5-0.5B-Instruct")
+ref_model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen2.5-0.5B-Instruct", load_in_8bit=True)
```
lewtun, line 202: I think this is the preferred API now:

```suggestion
quantization_config = BitsAndBytesConfig(load_in_8bit=True)
ref_model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen2.5-0.5B-Instruct", quantization_config= quantization_config)
```

docs/source/example_overview.md
```diff
@@ -44,8 +44,6 @@ Then, it is encouraged to launch jobs with `accelerate launch`!
...
 | [`examples/scripts/orpo.py`](https://github.com/huggingface/trl/blob/main/examples/scripts/orpo.py)                           | This script shows how to use the [`ORPOTrainer`] to fine-tune a model to increase helpfulness and harmlessness using the [Anthropic/hh-rlhf](https://huggingface.co/datasets/Anthropic/hh-rlhf) dataset.                                                                                                                                                                                                                                                          |
-| [`examples/scripts/ppo_multi_adapter.py`](https://github.com/huggingface/trl/blob/main/examples/scripts/ppo_multi_adapter.py) | This script shows how to use the [`PPOTrainer`] to train a single base model with multiple adapters. Requires you to run the example script with the reward model training beforehand.                                                                                                                                                                                                                                                                            |
```
edbeeching, line 47: Do we not want to link to `scripts/ppo/ppo.py` & `scripts/ppo/ppo_tldr.py` ?

docs/source/customization.mdx
lewtun, line 1: Although we use the `DPOTrainer` in these examples, should we start the doc with a small comment that they customisation applies to most (all?) trainers?

edbeeching (approved):
LGTM apart from a comment about the examples and 1 change

docs/source/detoxifying_a_lm.mdx
```diff
@@ -98,19 +98,15 @@ model = AutoModelForCausalLM.from_pretrained("EleutherAI/gpt-j-6B", torch_dtype=
 
 and the optimizer will take care of computing the gradients in `bfloat16` precision. Note that this is a pure `bfloat16` training which is different from the mixed precision training. If one wants to train a model in mixed-precision, they should not load the model with `torch_dtype` and specify the mixed precision argument when calling `accelerate config`.
 
-- Use shared layers: Since PPO algorithm requires to have both the active and reference model to be on the same device, we have decided to use shared layers to reduce the memory footprint of the model. This can be achieved by just speifying `num_shared_layers` argument when creating a `PPOTrainer`:
+- Use shared layers: Since PPO algorithm requires to have both the active and reference model to be on the same device, we have decided to use shared layers to reduce the memory footprint of the model. This can be achieved by specifying `num_shared_layers` argument when calling `create_reference_model` function. For example, if you want to share the first 6 layers of the model, you can do it like this:
```
lewtun, line 101: nit:

```suggestion
- Use shared layers: Since PPO algorithm requires to have both the active and reference model to be on the same device, we have decided to use shared layers to reduce the memory footprint of the model. This can be achieved by specifying `num_shared_layers` argument when calling the `create_reference_model()` function. For example, if you want to share the first 6 layers of the model, you can do it like this:
```

trl/trainer/ppo_config.py
```diff
@@ -41,199 +28,35 @@ class PPOConfig:
     command line.
 
     Parameters:
-        exp_name (`str`, *optional*, defaults to `os.path.basename(__file__)[: -len(".py")]`):
+        exp_name (`str`, *optional*, defaults to `os.path.basename(__file__)[:-3]`):
```
lewtun, line 31: I've always been a bit unsure why we need this extra parameter in our configs. Should we remove it (maybe in a follow-up PR)?

trl/trainer/ppo_config.py
```diff
@@ -41,199 +28,35 @@ class PPOConfig:
...
-    steps: int = 20000
-    learning_rate: float = 1.41e-5
-    adap_kl_ctrl: bool = True
-    init_kl_coef: float = 0.2
-    kl_penalty: Literal["kl", "abs", "mse", "full"] = "kl"
-    target: float = 6.0
-    horizon: float = 10000.0
-    gamma: float = 1.0
-    lam: float = 0.95
+    exp_name: str = os.path.basename(__file__)[: -len(".py")]
```
lewtun, line 53: Perhaps it's a bug in the GitHub diff, but this looks repeated from above

lewtun (approved):
Thanks for the migration @qgallouedec ! Overall LGTM with some nits about the docs since we can take this opportunity to nuke a bunch of outdated stuff

docs/source/example_overview.md
qgallouedec: done in fd77f55

docs/source/customization.mdx
```diff
@@ -46,171 +46,178 @@ else:
...
-# 1. load a pretrained model
-model = AutoModelForCausalLMWithValueHead.from_pretrained('gpt2')
-ref_model = AutoModelForCausalLMWithValueHead.from_pretrained('gpt2')
-tokenizer = GPT2Tokenizer.from_pretrained('gpt2')
+### Use LION optimizer
 
-# 2. define config
-ppo_config = {'batch_size': 1, 'learning_rate':1e-5}
-config = PPOConfig(**ppo_config)
+You can use the new [LION optimizer from Google](https://huggingface.co/papers/2302.06675) as well, first take the source code of the optimizer definition [here](https://github.com/lucidrains/lion-pytorch/blob/main/lion_pytorch/lion_pytorch.py), and copy it so that you can import the optimizer. Make sure to initialize the optimizer by considering the trainable parameters only for a more memory efficient training:
```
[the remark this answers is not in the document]
qgallouedec: done in f43896a

docs/source/customization.mdx
qgallouedec: Done in 9bfab24

docs/source/customization.mdx
qgallouedec: Agree with you. Let's apply these recommendation in other PRs (see https://github.com/huggingface/trl/pull/2174#issuecomment-2399843454)

docs/source/customization.mdx
qgallouedec:
```suggestion
Read more about 8-bit model loading in `transformers` [here](https://huggingface.co/docs/transformers/en/peft#load-in-8bit-or-4bit).
```

docs/source/customization.mdx
qgallouedec: Done in 59515ca

sygi:
Would it be possible to keep the "old" PPO code around until the new trainer achieves feature parity (in particular peft and arbitrary reward support) with the old one?

Void-025:
I would also like to have the old PPO code kept around, I was previously using the old ppotrainer.step() function to  provide my own reward values for each query/response pair, but there doesn't seem to be an equivalent in PPOv2, unless I'm missing some other way of doing this?

qgallouedec:
You should use trl==0.11

corbt:
Are there guides/documentation on how to use the new PPO trainer? I'm particularly interested in how to combine it with a custom reward function (instead of an explicit reward model). Also, I'm curious about any new benefits from this change—what are we gaining with it?

Thanks!

dhedey:
Just want to link this issue : https://github.com/huggingface/trl/issues/2718 to this PR.

It looks like `AutoModelForCausalLMWithValueHead` is no longer supported as a PPO (value) model? Because `value_model.base_model_prefix` errors out. Is that expected?

Is the only supported value model and reward model `AutoModelForSequenceClassification` with `num_labels=1` as per https://github.com/huggingface/trl/blob/main/examples/scripts/ppo/ppo_tldr.py and https://github.com/huggingface/trl/blob/main/examples/scripts/ppo/ppo.py?

Are there any docs where the value model is trained alongside the policy, and isn't the same as the reward model? My understanding is that a value model should probably be an `AutoModelForTokenClassification`, as it aught to predict a "reward-to-go" for each token; rather than for the whole sequence?

LukasNel:
this is a pretty obnoxious change! it should be way more clearly shown on the documentation and should have stayed renamed to PPOv2 as it is a breaking change from PPO and is not at all compatible. I wanted to run something like:
for batch in dataloader:
    thing = generate()
    reward = reward_function(thing)
    trainer.step(reward)
    
 And the previous PPO seemed like a great fit, but now I'm unsure what the actual way to do this would be. I don't want to use one of the Trainer ones since they all seem to want to control generation and I have a custom generation function I run.

Issue #2054, opened 2024-09-11
[DPOTrainer] Tokenizer calculation fail during Q+A concat


GokulNC:
### System Info

trl 0.10.1

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

I am performing DPO on an instruction tuned model.

Since TRL does not apply chat template by default, I preprocess the data to add prompt template to each `prompt`.

This is the `chat_template` from my tokenizer:
```
{% set loop_messages = messages %}{% for message in loop_messages %}{% set content = ' [INST] ' + message['content'] | trim + ' [/INST] ' %}{% if message['role'] != 'user' %}{% set content = message['content'] | trim + eos_token %}{% endif %}{% if loop.index0 == 0 %}{% set content = bos_token + content %}{% endif %}{{ content }}{% endfor %}
```

For instance, if the prompt text is `Choose the right option.`, after apply prompt template, it becomes:
`<s> [INST] Choose the right option. [/INST] `
(Note the final space)

When tokenized, it looks like this: (length: 10)
`[1, 67476, 4101, 31558, 4773, 4809, 67937, 67476, 4102, 67476]`

Now, since DPOTrainer concats the prompt + response directly as seen here:
https://github.com/huggingface/trl/blob/e2966c8d99104b6dda4f239d19a19ba36a7a6edc/trl/trainer/dpo_trainer.py#L313

It results in a string like this:
`<s> [INST] Choose the right option. [/INST] D`

which when tokenized, results in: (length 10)
`[1, 67476, 4101, 31558, 4773, 4809, 67937, 67476, 4102, 4884]`

Now the issue is that, in the DPOTrainer code, it results in empty list for the following line when computing answer tokens:
https://github.com/huggingface/trl/blob/e2966c8d99104b6dda4f239d19a19ba36a7a6edc/trl/trainer/dpo_trainer.py#L316-L317

When I trained a model without noticing this issue, it resulted in returning no response. That's when I started debugging this issue, and dug deeper to find this.

Now note that, the solution to this problem is not merely `strip`ing the prompt before concatenating. Because depending on whatever was the last few characters in the prompt/question, when a new answer string is appended, it might result in a different token at the intersection (if such a token exists in the vocabulary).

### Expected behavior

The `answer_input_ids` in the above code snippet should never become empty.

[...]

Issue #1960, opened 2024-08-23
llama3 tokenizer compatibility of DPOTrainer


hash2430:
I see annotations in build_tokenized_answer of DPOTrainer regarding llama tokenizer compatibility.
        """
        Llama tokenizer does satisfy `enc(a + b) = enc(a) + enc(b)`.
        It does ensure `enc(a + b) = enc(a) + enc(a + b)[len(enc(a)):]`.
        Reference:
            https://github.com/EleutherAI/lm-evaluation-harness/pull/531#issuecomment-1595586257
        """
I'm confused because it says on top that enc(a+b)=enc(a)+enc(b) is guarantted and also it says

"        # On some tokenizers, like Llama-2 tokenizer, there are occasions where tokens
        # can be merged together when tokenizing prompt+answer. This could result
        # on the last token from the prompt being different when tokenized on its own
        # vs when done as prompt+answer.
"
I might be misunderstood but they seem conflicted.
However I believe the contributors must have covered any possible issues with llama2 tokenizers.
My question is, is this also the case with llama3?
Because I don't see any comments regarding llama3 tokenizer.

qgallouedec:
I understand your question. This will change with #2209, as we will tokenize prompt and completion separately.
Feel free to take a look at the corresponding section in the PR description.

PR #1914, opened 2024-08-10
[DPO] tokenize and process DPO data via batches

kashif:
tokenize outside of the class and in batches 

fixes #1310 #1896 #1964

duplicate of PR #1896

PR #2209, opened 2024-10-09
Refactor DPO data processing

qgallouedec:
# What does this PR do?

This PR refactors DPO data processing to be less intricate and easier to understand and maintain. 

This main changes are the method `tokenize_row` and its equivalent for vision data `process_row`. 

This PR also solves a couple of issues:
- Wrong reported logits
- Wrong truncation logic

It also modifies the following:
- Make `max_length`, `max_prompt_length` and `max_completion_length` optional
- When merging prompt and completion alter tokenization

Regression tests are presented in my next message.

### Wrong truncation logic

Below is a simplified version of the current truncation logic (from `main`) used to handle sequences exceeding the allowed `max_length`:

```python
max_length = 10  # example value
max_prompt_length = 5  # example value

# if the combined sequence is too long, truncate the prompt
if len(prompt) + len(completion) > max_length:
    prompt = prompt[-max_prompt_length:]

# if it's still too long, truncate the completion from the end
if len(prompt) + len(completion) > max_length:
    completion = completion[: max_length - max_prompt_length]
```

#### Issues with this logic

1. **Unnecessary prompt truncation**: The prompt gets truncated more than necessary.
2. **Non-intuitive logic**: It's not immediately clear how truncation is applied.
3. **Lack of `max_completion_length` handling**: While `max_prompt_length` is applied, there's no equivalent handling for completions, making the logic even less intuitive.

#### Example

Consider the following:

```python
prompt = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
completion = [11, 12, 13, 14, 15]
```

Applying the original truncation logic gives:

```python
>>> prompt
[6, 7, 8, 9, 10]
>>> completion
[11, 12, 13, 14, 15]
>>> len(prompt) + len(completion)
10
```

In this case, the total length ends up being 10, even though `max_length` allows for up to 15 tokens. This means we could have preserved more of the prompt.

#### Proposed Solution

Here's a more flexible and intuitive approach (simplified for clarity):

```python
max_length = 10  # example value
max_prompt_length = None  # example value
max_completion_length = 8  # example value

# Truncate prompt if max_prompt_length is defined
if max_prompt_length is not None:
    prompt = prompt[-max_prompt_length:]

# Truncate completion if max_completion_length is defined
if max_completion_length is not None:
    completion = completion[:max_completion_length]

# Ensure combined prompt and completion don't exceed max_length
if len(prompt) + len(completion) > max_length:
    completion = completion[:max_length - len(prompt)]
```

#### Benefits

1. **Retains more of the prompt**: This logic avoids unnecessarily truncating the prompt.
2. **More intuitive**: It's clearer when each truncation step happens, and the flow is easier to follow.
3. **Supports flexible lengths**: Allows `None` for `max_length`, `max_prompt_length`, and `max_completion_length`, offering more control over how each part is truncated.

#### Result with the new logic

```python
>>> prompt, completion
([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], [11, 12, 13, 14, 15])
>>> len(prompt) + len(completion)
15
```

In this example, the prompt and completion are preserved within the allowed `max_length` without over-truncation.

### Wrong reported logits and log probs

The reported logits and log probs didn't take into account the padding. Eg:

```python
chosen_logits = logits.mean()
```

The new reporting first applies the mask before computing the mean:

```python
chosen_logits = logits[loss_mask].mean()
```

### When merging prompt and completion alter tokenization

Related #1960 and #2054

(very rare but worth mentioning)

There might be a difference between tokenizing the full concatenated text versus tokenizing the prompt and completion separately, and then merging their token lists.

Consider the following example:

```python
prompt = "Lorem."
completion = "\n\nipsum"
```

Tokenizing the combined prompt and completion produces different results than tokenizing each part separately and then concatenating their token lists:

```python
>>> tokenizer(prompt + completion)["input_ids"]
[32783, 382, 573, 1242]
>>> tokenizer(prompt)["input_ids"]
[32783, 13]
>>> tokenizer(completion)["input_ids"]
[271, 573, 1242]
```

This difference arises because we have a specific token for the string `".\n\n"`:

```python
>>> tokenizer.decode([382])
'.\n\n'
>>> tokenizer.decode([13, 271])
'.\n\n'
```

In the first tokenization method, the `".\n\n"` is treated as a single token, whereas in the second method, the token is split between the prompt and the completion.

In this example, the `main` code moves the merging token (`382`) to the completion:

```python
>>> prompt_ids = [32783]
>>> completion_ids = [382, 573, 1242]
>>> tokenizer.decode(prompt_ids)
'Lorem'
>>> tokenizer.decode(completion_ids)
'.\n\nipsum'
```

While this approach is functional, it slightly alters the original text since the final period is no longer part of the prompt. To avoid this, we can split the tokens without altering the data:

```python
>>> prompt_ids = [32783, 13]
>>> completion_ids = [271, 573, 1242]
>>> tokenizer.decode(prompt_ids)
'Lorem.'
>>> tokenizer.decode(completion_ids)
'\n\nipsum'
```

This way, the period remains in the prompt, preserving the original structure of the text. Note that this scenario is quite rare, it typically does not occur in conversational datasets.

## Before submitting

      Pull Request section?
      to it if that's the case.

## Who can review?


qgallouedec:
Regression tests for:

- [ ] Ecoder-decoder

kashif (approved):

tests/test_dpo_trainer.py
```diff
@@ -1139,7 +979,7 @@ def test_vdpo_trainer(self, model_id):
                 output_dir=tmp_dir,
                 per_device_train_batch_size=2,
                 max_length=512,
-                max_prompt_length=128,
```
qgallouedec, line 1142: Otherwise it truncates the image tokens

trl/trainer/dpo_trainer.py
```diff
@@ -815,27 +607,28 @@ def make_inputs_require_grad(module, input, output):
...
-                "model": model if self.is_encoder_decoder else None,
+                "processing_class": processing_class,
+                "max_prompt_length": args.max_prompt_length,
+                "max_completion_length": args.max_completion_length,
+                # for enc-dec, we add the special tokens ([bos_token] + prompt + [eos_token]; completion + [eos_token])
+                "add_special_tokens": self.is_encoder_decoder,
             }
             train_dataset = train_dataset.map(
-                _tokenize,
+                self.tokenize_row if not self.is_vision_model else self.process_row,
```
lewtun, line 617: very clean separation of concerns!

trl/trainer/dpo_trainer.py
```diff
@@ -815,27 +607,28 @@ def make_inputs_require_grad(module, input, output):
...
                 fn_kwargs=fn_kwargs,
-                batched=True,
                 num_proc=self.dataset_num_proc,
                 writer_batch_size=10,
                 desc="Tokenizing train dataset",
+                load_from_cache_file=False,
```
lewtun, line 622: Is this to prevent the issue we've discussed previously about how changing a data mixture / number of samples doesn't recompute the processing?

trl/trainer/dpo_trainer.py
```diff
@@ -815,27 +607,28 @@ def make_inputs_require_grad(module, input, output):
...
+                "max_prompt_length": args.max_prompt_length,
+                "max_completion_length": args.max_completion_length,
+                # for enc-dec, we add the special tokens ([bos_token] + prompt + [eos_token]; completion + [eos_token])
+                "add_special_tokens": self.is_encoder_decoder,
             }
             train_dataset = train_dataset.map(
-                _tokenize,
+                self.tokenize_row if not self.is_vision_model else self.process_row,
                 fn_kwargs=fn_kwargs,
-                batched=True,
```
lewtun, line 826: Does this have any significant impact on preprocessing? I recall @kashif looked into this because the preprocessing was really slow for large preference datasets like Anthropic's HHH

trl/trainer/dpo_trainer.py
```diff
@@ -1052,121 +933,116 @@ def null_ref_context(self):
             if self.ref_adapter_name:
                 self.model.set_adapter(self.model_adapter_name or "default")
 
-    def compute_reference_log_probs(self, padded_batch: Dict) -> Dict:
+    def compute_reference_log_probs(self, batch: Dict[str, torch.LongTensor]) -> Dict:
```
lewtun, line 936: turbo nit: since we're already renaming variables, maybe we can align the method too?
```suggestion
    def compute_ref_log_probs(self, batch: Dict[str, torch.LongTensor]) -> Dict:
```

trl/trainer/dpo_trainer.py
```diff
@@ -1284,190 +1162,148 @@ def dpo_loss(
...
+            attention_mask = torch.cat((prompt_attention_mask, completion_attention_mask), dim=1)
+            # Mask the prompt but not the completion for the loss
+            loss_mask = torch.cat(
+                (torch.zeros_like(prompt_attention_mask), completion_attention_mask),
+                dim=1,
+            )
 
-        if self.loss_type == "ipo":
-            all_logps = all_logps / size_completion
+            # Flush left to reduce the memory usage
```
lewtun, line 1232: nice optimisation!

trl/trainer/dpo_trainer.py
```diff
@@ -1284,190 +1162,148 @@ def dpo_loss(
...
+
+        output = {}
 
-        policy_weights = None
         if self.use_weighting:
-            chosen_weights = all_weights[:len_chosen]
-            rejected_weights = all_weights[len_chosen:]
-            policy_weights = torch.clamp(torch.exp(chosen_weights + rejected_weights), max=1)
+            with torch.no_grad():
+                # Eq (2) of the WPO paper: https://huggingface.co/papers/2406.11827
```
lewtun, line 1276: Is this weighting applied for vanilla DPO as well or is it excluded since the loss doesn't depend on the `policy_weights` attribute?

lewtun (commented):
Excellent refactor @qgallouedec - I really like how clean the new code is to read! I've left a few nits and questions about the weights loss from WPO, but otherwise this looks great. 

Regarding the difference in the chosen/rejected rewards of your regression tests, have you looked at the impact on downstream evals like IFEval / AlpacaEval / MixEval? I can run those for you if you have the checkpoints handy and then we can be pretty sure it's fine

trl/trainer/dpo_trainer.py
```diff
@@ -76,266 +77,73 @@
...
+    Args:
+        pad_token_id (`int`):
+            Token ID to use for padding.
+        return_tensors (`str`, *optional*, defaults to `"pt"`):
+            Type of Tensor to return. Only `"pt"` is currently supported.
+
+    Examples:
+    ```python
+    >>> from trl import PreferenceCollator
+    >>> collator = PreferenceCollator(pad_token_id=0, collation_strategy="split")
```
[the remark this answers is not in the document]
qgallouedec: This is an option I deleted but forgot to remove from the doc, thanks!
```suggestion
    >>> collator = PreferenceCollator(pad_token_id=0)
```

trl/trainer/dpo_trainer.py
qgallouedec: Arf no, it was more convenient to have it when debugging, good catch
```suggestion
```

trl/trainer/dpo_trainer.py
```diff
@@ -815,27 +607,27 @@ def make_inputs_require_grad(module, input, output):
...
             if eval_dataset is not None:
                 eval_dataset = eval_dataset.map(
-                    _tokenize,
+                    self.tokenize_row if not self.is_vision_model else self.process_row,
                     fn_kwargs=fn_kwargs,
-                    batched=True,
                     num_proc=self.dataset_num_proc,
                     writer_batch_size=10,
                     desc="Tokenizing eval dataset",
+                    load_from_cache_file=False,
```
qgallouedec, line 630:
```suggestion
```

trl/trainer/dpo_trainer.py
qgallouedec: done in e509549

trl/trainer/dpo_trainer.py
```diff
@@ -1478,67 +1314,42 @@ def get_batch_loss_metrics(
...
         reward_accuracies = (chosen_rewards > rejected_rewards).float()
 
         if self.args.rpo_alpha is not None:
-            # RPO loss from V3 of the paper:
-            losses = losses + policy_nll_loss * self.args.rpo_alpha
+            losses = losses + self.args.rpo_alpha * model_output["nll_loss"]  # RPO loss from V3 of the paper
 
         if self.use_weighting:
-            losses = losses * policy_weights
+            losses = losses * model_output["policy_weights"]
```
qgallouedec, line 1333: Answering https://github.com/huggingface/trl/pull/2209/files/9824f1670bb08845ab7bb850b5b14671bf843da5#r1804420747

I'm not 100% sure to understand your question. When the weighting is disabled, the loss won't be multiplied by the weights here. It's not even in `model_output`.
Please tell me if I've missed something.

trl/trainer/dpo_trainer.py
qgallouedec: In fact, with batch processing, the problem is just shifted, as this function takes a list of examples instead of a single example, but processes it with a for loop.
The advantage of using `batched=False` is that you can make more intensive use of multiprocessing, which should be able to solve the slowness problem.

But I didn't know it was for that reason that we moved to `batched=True` (it wasn't mentioned in #1914). I'll double-check anyway to verify what I said.

qgallouedec: |          | `dataset_num_proc = None` | `dataset_num_proc = 32` |
| -------- | ------------------------- | ----------------------- |
| `main`   | 04:59                     | 00:38                   |
| PR #2209 | 02:02                     | 00:17                   |

It's worth noting that the comparison isn't entirely fair, because in the case of the 2209, we don't compute the attention mask and the labels in the preprocessing function.

qgallouedec:
> Regarding the difference in the chosen/rejected rewards of your regression tests, have you looked at the impact on downstream evals like IFEval / AlpacaEval / MixEval? I can run those for you if you have the checkpoints handy and then we can be pretty sure it's fine

Nice idea, I'll send you the checkpoints!

trl/trainer/dpo_trainer.py
lewtun: this looks great. let's avoid batching and just use multiprocessing as you have

trl/trainer/dpo_trainer.py
lewtun: arf sorry i was being dumb - i missed the `self.use_weighting` line above and thought we were computing `policy_weights` by default. ignore my question :)

qgallouedec:
@lewtun

Here is one: https://huggingface.co/qgallouedec/Qwen2.5-7B-DPO-2209
Trained with 

```
accelerate launch --config_file=examples/accelerate_configs/deepspeed_zero2.yaml examples/scripts/dpo.py \
    --dataset_name trl-lib/ultrafeedback_binarized \
    --model_name_or_path Qwen/Qwen2.5-7B-Instruct \
    --learning_rate 5.0e-7 \
    --num_train_epochs 1 \
    --gradient_checkpointing \
    --logging_steps 10 \
    --eval_strategy steps \
    --eval_steps 100 \
    --output_dir Qwen2.5-7B-DPO-2209 \
    --gradient_checkpointing \
    --max_length 512 \
    --use_peft \
    --bf16 \
    --push_to_hub
```

qgallouedec:
## IFEval

The new implementation seems to improve results

| Model                                                                         | inst_level_loose_acc | inst_level_strict_acc | prompt_level_loose_acc | prompt_level_strict_acc |
| ----------------------------------------------------------------------------- | -------------------- | --------------------- | ---------------------- | ----------------------- |
| [Qwen2.5-7B](https://huggingface.co/Qwen/Qwen2.5-7B-Instruct)                 | 0.7122               | 0.6631                | 0.6026 ± 0.0211        | 0.5416 ± 0.0214         |
| [Qwen2.5-7B-DPO-main](https://huggingface.co/qgallouedec/Qwen2.5-7B-DPO-main) | 0.7182               | 0.6751                | 0.6155 ± 0.0209        | 0.5693 ± 0.0213         |
| [Qwen2.5-7B-DPO-2209](https://huggingface.co/qgallouedec/Qwen2.5-7B-DPO-2209) | 0.7326               | 0.6775                | 0.6303 ± 0.0208        | 0.5656 ± 0.0213         |

kashif (approved):

PR #2220, opened 2024-10-11
Add Sequence-Level KD

mst272:
# What does this PR do?

In the original paper, they compared Sequence-Level KD 、Supervised KD and GKD (On-policy). In trl GKDTrainer, Supervised KD and GKD have been implemented.  So i add Sequence-Level KD in GKDTrainer, control whether to perform Sequence-Level KD through a seq_kd parameter.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


kashif:
thanks @mst272 can you also kindly add these options to the docs-strings and the documentation of `GKDTrainer`?

mst272:
hi@kashif,I've added these to the docs-strings and the documentation

trl/trainer/gkd_config.py
```diff
@@ -41,6 +41,9 @@ class GKDConfig(SFTConfig):
             from a string.
         disable_dropout (`bool`, *optional*, defaults to `True`):
             Whether or not to disable dropouts in `model`.
+        seq_kd (`bool`, *optional*, defaults to `False`):
+            Seq_kd parameter that controls whether to perform Sequence-Level KD(can be viewed as supervised FT
```
kashif, line 45:
```suggestion
            Seq_kd parameter that controls whether to perform Sequence-Level KD (can be viewed as supervised FT
```

docs/source/gkd_trainer.md
```diff
@@ -19,8 +19,9 @@ This post-training method was contributed by [Kashif Rasul](https://huggingface.
 
 ## Usage tips
 
-The GKD Trainer is a wrapper around the [`SFTTrainer`] class that takes in a teacher model argument. It needs two parameters to be set via the [`GKDConfig`] namely:
+The GKD Trainer is a wrapper around the [`SFTTrainer`] class that takes in a teacher model argument. It needs three parameters to be set via the [`GKDConfig`] namely:
 * `lmbda`:  controls the student data fraction, i.e., the proportion of on-policy student-generated outputs. When `lmbda=0.0`, the loss reduces to supervised JSD where the student is trained with the token-level probabilities of the teacher. When `lmbda=1.0`, the loss reduces to on-policy JSD, where the student generates output sequences and token-specific feedback on these sequences from the teacher. For values in between [0, 1] it is random between the two based on the `lmbda` value for each batch.
+* `seq_kd`:  controls whether to perform Sequence-Level KD(can be viewed as supervised FT on teacher-generated out). When `seq_kd=True` and `lmbda=0.0`, the loss reduces to supervised JSD, where the teacher generates output sequences and the student receive token-specific feedback on these sequences from the teacher.
```
kashif, line 24:
```suggestion
* `seq_kd`:  controls whether to perform Sequence-Level KD (can be viewed as supervised FT on teacher-generated out). When `seq_kd=True` and `lmbda=0.0`, the loss reduces to supervised JSD, where the teacher generates output sequences and the student receives token-specific feedback on these sequences from the teacher. 
```

trl/trainer/gkd_config.py
```diff
@@ -41,6 +41,9 @@ class GKDConfig(SFTConfig):
...
+            on teacher-generated out).
```
kashif, line 46:
```suggestion
            on teacher-generated output).
```

docs/source/gkd_trainer.md
```diff
@@ -19,8 +19,9 @@ This post-training method was contributed by [Kashif Rasul](https://huggingface.
 
 ## Usage tips
 
-The GKD Trainer is a wrapper around the [`SFTTrainer`] class that takes in a teacher model argument. It needs two parameters to be set via the [`GKDConfig`] namely:
+The GKD Trainer is a wrapper around the [`SFTTrainer`] class that takes in a teacher model argument. It needs three parameters to be set via the [`GKDConfig`] namely:
```
qgallouedec, line 22:
```suggestion
The [`GKDTrainer`] is a wrapper around the [`SFTTrainer`] class that takes in a teacher model argument. It needs three parameters to be set via the [`GKDConfig`] namely:
```

qgallouedec (approved):
LGTM, I leave the last word to @kashif

kashif (approved):

moussaKam:
Hi there, 

I don't really understand how this PR adds seq_kd. To my understanding the seq_kd computes the standard `cross_entropy` between student logits and the output generated by the teacher. 

In this PR we are simply generating the teacher output then computing the same `generalized_jsd_loss`. Am I missing something ?

kashif:
@moussaKam So recall there are 2 other parameters apart from the `seq_kd` flag, namely the `lmbda` so set that to zero and then the `beta` that interpolates between the forward and reverse KL. SeqKD then would be `lmbda=0`, and `beta=0` if i am not mistaken... its an interesting question and might need exploring, what happens with other values for these 2 hyper-parameters...

kashif:
@moussaKam also note that in this case, the KL-div is the same as the CE and a constant term, i.e. the entropy of the target which we assume does not change

moussaKam:
@kashif  thanks for you're reply, yes according to the definition from the paper 

> Sequence-Level KD (Kim & Rush, 2016). SeqKD maximizes the likelihood of high probability sequences generated by the teacher, and can be viewed as supervised FT on teacher-generated output.

I understand that we compute the cross-entropy in the case of seq_kd. This is what we do in standard sft no?

moussaKam:
@kashif another point, if `seq_kd` is put to `True` we are computing the teacher inference twice. [here](https://github.com/huggingface/trl/blob/74e20cbbbcbac7ac8d426df09eda5f310c637def/trl/trainer/gkd_trainer.py#L286) and [here](https://github.com/huggingface/trl/blob/74e20cbbbcbac7ac8d426df09eda5f310c637def/trl/trainer/gkd_trainer.py#L227). Do we really need that?

kashif:
@moussaKam so in the first we generate completions and in the 2nd we calculate the logits of the completions... i suppose we could do that once, and then keep track of it with a bunch of if-else but opted for some cleaner logic here that could work for any of the different hyperparams... any ideas on how to make it a bit more dry?

kashif:
@moussaKam Mind you there is an orthogonal abstraction i have been working on where instead of the logits (which are assumed to come from the same vocab. size for both the student and teacher) we allow for the student-teacher to have different vocabs: see https://github.com/huggingface/trl/pull/2263 and I would welcome any thoughts if this should be a separate class?

moussaKam:
@kashif, we don't need to compute the logits, we generate the output with the teacher which becomes the new labels, then we run the forward of the student and compute the cross entropy using just the teacher output tokens. 

I can implement it in the afternoon if it sounds good for you.

PR #2251, opened 2024-10-18
Adjust padding in batch generation

gaetanlop:
# What does this PR do?

Fix a bug in the `batch_generation` function where we try to concatenate tensors that can be of different dimensions (`sequence_length`) when generating with `mini_batches`.

## Before submitting

## Who can review?

@kashif @qgallouedec

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -14,234 +14,243 @@
 
 import unittest
 
+import torch
 from datasets import load_dataset
 from transformers import AutoModelForCausalLM, AutoTokenizer, GenerationConfig
+from transformers.testing_utils import require_peft
+from transformers.utils import is_peft_available
 
+from trl.trainer.model_config import ModelConfig
 from trl.trainer.utils import (
+    DataCollatorForChatML,
     batch_generation,
+    decode_and_strip_padding,
+    generate_model_card,
+    get_peft_config,
+    pad,
 )
 
 
-# if is_peft_available():
-#     from peft import LoraConfig
+if is_peft_available():
+    from peft import LoraConfig
 
 
-# class TestPad(unittest.TestCase):
-#     def test_pad_1_dim_left(self):
-#         x = torch.tensor([1, 2, 3])
-#         y = torch.tensor([4, 5])
-#         output = pad((x, y), padding_value=0, padding_side="left")
-#         expected = torch.tensor([[1, 2, 3], [0, 4, 5]])
-#         self.assertTrue(torch.equal(output, expected))
+class TestPad(unittest.TestCase):
+    def test_pad_1_dim_left(self):
+        x = torch.tensor([1, 2, 3])
+        y = torch.tensor([4, 5])
+        output = pad((x, y), padding_value=0, padding_side="left")
+        expected = torch.tensor([[1, 2, 3], [0, 4, 5]])
+        self.assertTrue(torch.equal(output, expected))
 
-#     def test_pad_1_dim_right(self):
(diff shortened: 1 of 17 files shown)


tests/test_utils.py
```diff
@@ -250,3 +251,60 @@ def test_data_collator_for_chatml(self):
...
+
+        self.assertGreater(max_length, context_length)
+        self.assertEqual(query_responses.shape, (bs, max_length))
+        self.assertEqual(logits.shape, (bs, max_length, self.model.config.vocab_size))
+
+    def test_single_batch_generation(self):
+        batch = [
+            self.tokenizer.apply_chat_template(example[:-1], add_generation_prompt=True, tokenize=False)
+            for example in self.examples
+        ]
```
kashif, line 298:
```suggestion
       single_example = self.examples[0]
        batch = [
            self.tokenizer.apply_chat_template(single_example[:-1], add_generation_prompt=True, tokenize=False)
        ]
```

kashif:
@gaetanlop  can we use the `pad` helpers?
```
padded_query_responses = pad(query_responses, padding_value=pad_token_id)
padded_logitss = pad(logitss, padding_value=float('-inf'))
```

trl/trainer/utils.py
```diff
@@ -1284,6 +1284,15 @@ def batch_generation(
...
         logitss.append(logits)
+
+    # padding tensors
+    max_length = max(query_response.shape[1] for query_response in query_responses)
+    query_responses = [
+        torch.nn.functional.pad(query_response, (0, max_length - query_response.shape[1]), value=pad_token_id)
+        for query_response in query_responses
+    ]
+    logitss = [torch.nn.functional.pad(logits, (0, 0, 0, max_length - logits.shape[1]), value=0) for logits in logitss]
+
```
kashif, line 1295:
```suggestion
    # padding tensors
    padded_query_responses = pad(query_responses, padding_value=pad_token_id, padding_side="right")
    padded_logitss = pad(logitss, padding_value=0, padding_side="right")

```

gaetanlop:
@kashif, the `pad` function expects the tensor to have no leading dimension corresponding to the batch size.

Here is an example `query_responses`:
```python
query_responses = [
  torch.randint(vocab_size, (bs, seq_length1)), 
  torch.randint(vocab_size, (bs, seq_length2)), 
  torch.randint(vocab_size, (remaining_samples, seq_length3))
]
```

Using the `pad` function as it is would require the following change before passing the `query_responses` to the `pad` function: 

```python
query_responses=[query_reps[i] for query_reps in query_responses for i in range(query_reps.size(0))]
```

We can also change the pad function? What do you prefer?

kashif (approved):

trl/trainer/utils.py
```diff
@@ -1284,6 +1284,11 @@ def batch_generation(
         )
         query_responses.append(query_response)
         logitss.append(logits)
+
+    # padding tensors
+    padded_query_responses = pad(query_responses, padding_value=pad_token_id, padding_side="right")
+    padded_logitss = pad(logitss, padding_value=0, padding_side="right")
+
     return torch.cat(query_responses, 0), torch.cat(logitss, 0)
```
kashif, line 1292:
```suggestion
    return padded_query_responses, padded_logitss
```

gaetanlop:
This won't work @kashif, it still requires reshaping the tensors

kashif:
ah damn! my bad sorry!

gaetanlop:
This should be fixed now

PR #2286, opened 2024-10-26
🖨️ Fix error text in BCO and KTO tokenizing function

PhilipMay:
The error text is saying:

> "Prompt input ids and answer input ids should have the same length."

This is very misleading. The code does not compare prompt and answer in this case.


PhilipMay:
I don't think the CI problems have anything to do with the changes in this PR...

qgallouedec (approved):

qgallouedec:
Thanks @PhilipMay! Do you mind updating your branch? I don't have the writing rights on your branch.

Issue #2293, opened 2024-10-29
DataCollatorForCompletionOnlyLM not working


SwayamInSync:
### System Info

```
Collecting environment information...
PyTorch version: 2.5.0+cu124
Is debug build: False
CUDA used to build PyTorch: 12.4
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.6 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.31

Python version: 3.10.15 | packaged by conda-forge | (main, Oct 16 2024, 01:24:24) [GCC 13.3.0] (64-bit runtime)
Is CUDA available: True
CUDA runtime version: 12.4.99
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: 
GPU 0: NVIDIA A100 80GB PCIe
GPU 1: NVIDIA A100 80GB PCIe
GPU 2: NVIDIA A100 80GB PCIe
GPU 3: NVIDIA A100 80GB PCIe

Nvidia driver version: 560.35.03
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Byte Order:                           Little Endian
Address sizes:                        48 bits physical, 48 bits virtual
CPU(s):                               96
On-line CPU(s) list:                  0-95
Thread(s) per core:                   1
Core(s) per socket:                   48
Socket(s):                            2
NUMA node(s):                         4
Vendor ID:                            AuthenticAMD
CPU family:                           25
Model:                                1
Model name:                           AMD EPYC 7V13 64-Core Processor
Stepping:                             1
CPU MHz:                              2445.440
BogoMIPS:                             4890.88
Virtualization type:                  full
L1d cache:                            3 MiB
L1i cache:                            3 MiB
L2 cache:                             48 MiB
L3 cache:                             384 MiB
NUMA node0 CPU(s):                    0-23
NUMA node1 CPU(s):                    24-47
NUMA node2 CPU(s):                    48-71
NUMA node3 CPU(s):                    72-95
Vulnerability Gather data sampling:   Not affected
Vulnerability Itlb multihit:          Not affected
Vulnerability L1tf:                   Not affected
Vulnerability Mds:                    Not affected
[...]

PR #2295, opened 2024-10-29
[DataCollator] Add warning for out-of-vocabulary token IDs

SwayamInSync:
# What does this PR do?

This PR adds a patch for #2293 adding a validation check in `DataCollatorForCompletionOnlyLM` to warn users when token IDs exceed the tokenizer's vocabulary size. This is a common issue that can occur when:
1. Using incorrect pad tokens (especially with LLaMA models)
2. Modifying tokenizer vocabulary without updating the model
3. Adding custom tokens without proper model resizing

The warning provides possible fixes, making debugging easier for users.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@qgallouedec

### DIFF
--- a/tests/test_data_collator_completion_only.py
+++ b/tests/test_data_collator_completion_only.py
@@ -142,3 +142,25 @@ class DataCollatorForCompletionOnlyLMTester(unittest.TestCase):
         self.assertTrue((input_ids_remove_pad == batch_paddingfree["input_ids"]).all())
         self.assertTrue((expected_position_ids == batch_paddingfree["position_ids"]).all())
         self.assertTrue((expected_labels == batch_paddingfree["labels"]).all())
+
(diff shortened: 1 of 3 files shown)


qgallouedec:
I understand the motivation behind this proposal, but I feel that warning would make more sense directly in the tokenizer. At least, that's where I'd look for it. What's more, this data collator is only used in specific cases. Having it in a collator (which is initially designed to collate data, not process it) seems strange to me.

SwayamInSync:
Hey @qgallouedec ,
Thanks and yeah that totally make sense, I can instead try to put this in tokenizer or as a check before starting training with Trainer.
Feel free to close it here

qgallouedec:
Good. Please link this PR if you push this elsewhere

PR #2303, opened 2024-10-31
Specify min versions 

qgallouedec:
# What does this PR do?

This PR adds a min version for `datasets` and `accelerate`. It adds a test to ensure that these min version are indeed supported.

I've found the specified min version by dichotomy.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


.github/workflows/tests.yml
```diff
@@ -110,6 +110,38 @@ jobs:
...
+          cache: "pip"
+          cache-dependency-path: |
+              setup.py
+              requirements.txt
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install accelerate==0.34.0
+          python -m pip install datasets==2.21.0
+          python -m pip install transformers==4.46.0
```
lewtun, line 135: For my understanding, if we bump the min version in `setup.py` we also need to bump it here as well right?

lewtun (approved):
Nice work enforcing the min versions - LGTM!

.github/workflows/tests.yml
qgallouedec: That's right. 
Ideally, we would be able to pip install minimal versions via, eg, `pip install --prefer_min .`, but it's not possible yet, see https://github.com/pypa/pip/issues/8085

PR #2333, opened 2024-11-07
Bump liger-kernel to 0.4.0

ByronHsu:
This fixes HF grad acc and other fixes/features/optimizations


kashif (approved):

qgallouedec:
As far as I understand, the grad accum think is only an issue with SFT right?

kashif:
right i think its more about the updated kernels

ByronHsu:
Yes grad accum is only used for sft. Besides grad accum, we also have other fixes.

qgallouedec:
I approve, as this is an important issue affecting the most widely used trainer. (Thanks for solving it!)

For the record, generally speaking, I won’t raise the minimum version requirement unless a new feature from the dependency is needed in our codebase.

qgallouedec (approved):

PR #2345, opened 2024-11-11
🔮 Inference mode in `GeometricMixtureWrapper.forward`

kashif:
# What does this PR do?

When calling the GeometricMixture generation mixer, the model is left in eval mode.


qgallouedec:
Wsh do you need the model to be un eval mode?

Can we use the inference mode in forward instead?

kashif:
@ qgallouedec using inference mode so there should be no unexpected behaviour

qgallouedec (approved):

PR #2380, opened 2024-11-21
🕹️ CLI refactor

qgallouedec:
# What does this PR do?

## Historically

The example scripts were initially placed outside the package directory (`.examples/scripts`) because they were not intended for runtime use. However, with the introduction of the TRL CLI in #1419, these scripts became executable at runtime through the CLI. Unfortunately, the scripts could not be directly imported from the package. To address this, we implemented a workaround in `setuptools` by [creating a symbolic link](https://github.com/huggingface/trl/blob/672c96546d9cae7a6d0afba381b189bb3cb2e8b5/Makefile#L13) from `trl/trl/commands` to `examples/scripts`.

## Motivations for this refactor

- Recurrent issues with the simlink not properly created: https://github.com/huggingface/trl/issues/1716, https://github.com/huggingface/trl/issues/1829 https://github.com/huggingface/trl/pull/2369#issue-2671794993
- Populating the CLI with new command isn't flexible https://github.com/huggingface/trl/issues/2101 https://github.com/huggingface/trl/pull/2157
- We [implement our own parser](https://github.com/huggingface/trl/blob/672c96546d9cae7a6d0afba381b189bb3cb2e8b5/trl/commands/cli.py#L128) based on `sys.argv` to have our own subparsing logic, while `argparse` [have everything we need for subparsing](https://docs.python.org/3/library/argparse.html#sub-commands)
- We don't have the usual parsing associated features like the help etc:
    ```
    $ trl dpo --help
    The following values were not passed to `accelerate launch` and had defaults used instead:
            `--num_processes` was set to a value of `1`
            `--num_machines` was set to a value of `1`
            `--mixed_precision` was set to a value of `'no'`
            `--dynamo_backend` was set to a value of `'no'`
    To avoid this warning pass in values for each of the problematic parameters or run `accelerate config`.
    /fsx/qgallouedec/miniconda3/envs/trl/bin/python: can't open file '/fsx/qgallouedec/trl/trl/commands/scripts/dpo.py': [Errno 2] No such file or directory
    Traceback (most recent call last):
      File "/fsx/qgallouedec/miniconda3/envs/trl/bin/accelerate", line 8, in <module>
        sys.exit(main())
                ^^^^^^
      File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.11/site-packages/accelerate/commands/accelerate_cli.py", line 48, in main
        args.func(args)
[...]

PR #2412, opened 2024-11-28
📑 Refactor `TrlParser`

qgallouedec:
# What does this PR do?

First step part of #2380

- Test `TrlParser`
- Deprecate `YamlConfigParser` (not used outside `TrlParser`, only get the piece of code and add it to `TrlParser`)
- Update `TrlParser` documentation
- deprecate `post_process_dataclasses` (no-op since we removed `XXXScriptArgument`. We will find another way of passing `gradient_checkpointing_use_reentrant`
- Remove config arg from `ScriptConfig` (not compatible with `TrlParser`)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/tests/test_cli_utils.py
+++ b/tests/test_cli_utils.py
@@ -78,7 +78,7 @@ class TestTrlParser(unittest.TestCase):
 
         args = ["--arg1", "3", "--config", "config.yaml"]  # override arg1 default with 3
 
-        # Simulate the config being loaded and environment variables being set
+        # Simulate the config being loaded and arguments being passed
         result_args = parser.parse_args_and_config(args)
 
         # Check the parsed arguments
--- a/tests/test_cli_utils.py
+++ b/tests/test_cli_utils.py
@@ -76,7 +76,7 @@ class TestTrlParser(unittest.TestCase):
 
         parser = TrlParser(dataclass_types=[MyDataclass])
 
-        args = ["--arg1", "3", "--config", "config.yaml"]  # don't set arg1 to test default value
+        args = ["--arg1", "3", "--config", "config.yaml"]  # override arg1 default with 3
 
         # Simulate the config being loaded and environment variables being set
         result_args = parser.parse_args_and_config(args)
--- a/tests/test_cli_utils.py
+++ b/tests/test_cli_utils.py
@@ -142,3 +142,20 @@ class TestTrlParser(unittest.TestCase):
         self.assertEqual(result_args[0].arg1, 2)
         self.assertEqual(result_args[0].arg2, "value")
         self.assertEqual(result_args[1], ["remaining"])
+
+    @patch("yaml.safe_load")
+    def test_parse_args_and_config_with_remaining_strings_in_config_and_args(self, mock_yaml_load):
+        mock_yaml_load.return_value = {"remaining_string_in_config": "abc"}
+
+        parser = TrlParser(dataclass_types=[MyDataclass])
+
+        args = ["--arg1", "2", "--remaining_string_in_args", "def", "--config", "config.yaml"]
+
+        # Simulate the config being loaded and arguments being passed
+        result_args = parser.parse_args_and_config(args, return_remaining_strings=True)
+
+        # Check that the arguments are parsed as is
+        self.assertEqual(len(result_args), 2)
+        self.assertIsInstance(result_args[0], MyDataclass)
+        self.assertEqual(result_args[0].arg1, 2)
+        self.assertEqual(result_args[1], ["--remaining_string_in_config", "abc", "--remaining_string_in_args", "def"])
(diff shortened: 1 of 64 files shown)


trl/commands/cli_utils.py
```diff
@@ -152,88 +163,126 @@ class ChatArguments:
...
+    def __init__(
+        self,
+        dataclass_types: Union[DataClassType, Iterable[DataClassType]],
+        ignore_extra_args: Optional[bool] = None,
+        **kwargs,
+    ):
+        super().__init__(dataclass_types=dataclass_types, **kwargs)
+        if ignore_extra_args is not None:
+            warnings.warn(
+                "The `ignore_extra_args` parameter is deprecated and will be removed in version 0.14. "
```
lewtun, line 219: Is this deprecated to enforce strict typing between a dataclass and the runtime config? I'm mostly wondering if this was included by design in the original parser for some specific use cases. In any case, let's deprecate and see if we get reports between now and the deprecation

trl/commands/cli_utils.py
```diff
@@ -152,88 +163,126 @@ class ChatArguments:
...
+        if "--config" in args:
+            # Get the config file path from
+            config_index = args.index("--config")
+            args.pop(config_index)  # remove the --config flag
+            config_path = args.pop(config_index)  # get the path to the config file
+            with open(config_path) as yaml_file:
+                config = yaml.safe_load(yaml_file)
+
+            # Set the environment variables specified in the config file
+            if "env" in config:
```
lewtun, line 264: This is a cool feature!

tests/test_cli_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    def test_init_with_config_field(self):
+        """Test initialization with a 'config' field in the dataclass (should raise ValueError)."""
+        with self.assertRaises(ValueError) as context:
+            TrlParser(dataclass_types=[InvalidDataclass])
+        self.assertTrue("has a field named 'config'" in str(context.exception))
+
+    @patch("builtins.open", mock_open(read_data="env:\n  VAR1: value1\n  VAR2: value2"))
+    @patch("yaml.safe_load")
+    @patch("os.environ", new_callable=dict)  # Mock os.environ as a dictionary
+    def test_parse_args_and_config_with_valid_config(self, mock_environ, mock_yaml_load):
```
lewtun, line 48: Can we add a test that also checks an existing arg in the config can be overwritten at runtime? I realise this test is quite close to that, but it would be good to be sure vs the creation of a new arg

lewtun (approved):
Nice harmonisation with the parser @qgallouedec ! LGTM with an ask to include 1 more test

trl/commands/cli_utils.py
qgallouedec: Good point! done in d378afa

tests/test_cli_utils.py
qgallouedec: Good point! done in d378afa

trl/commands/cli_utils.py
qgallouedec: Initially I deprecated it because I didn't understand what it was doing. It seemed to be no-op. But now I understand: When you have an argument in `config.yaml` that is not passed (belongs to no dataclass, then it returns a value error. Unless you set this arg to True:

```yaml
# config.yml
output_dir: "output_dir"
extra_arg: "This is not a recognized argument"
```

```python
from trl import TrlParser
from transformers import TrainingArguments

parser = TrlParser((TrainingArguments,), ignore_extra_args=False)
args = parser.parse_args_and_config()  # raise a ValueError

parser = TrlParser((TrainingArguments,), ignore_extra_args=True)
args = parser.parse_args_and_config()
print(args)
# (TrainingArguments(_n_gpu=1, ...), Namespace(extra_arg='This is not a recognized argument'))
```

Finally, it's mostly the same as `return_remaining_strings` in `parse_args_and_config` but for the config file. What I'll do:

Keep this arg and make it functional; deprecate it in favour of `return_remaining_strings`. Something like

```python
from trl import TrlParser
from transformers import TrainingArguments

parser = TrlParser((TrainingArguments,))
args = parser.parse_args_and_config()  # raise a ValueError

parser = TrlParser((TrainingArguments,), ignore_extra_args=True)
args = parser.parse_args_and_config()
print(args)
# Warning, this is depracated, use `return_remaining_strings`` instead
# (TrainingArguments(_n_gpu=1, ...), Namespace(extra_arg='This is not a recognized argument'))

parser = TrlParser((TrainingArguments,))
args = parser.parse_args_and_config(return_remaining_strings=True) 
print(args)
# (TrainingArguments(_n_gpu=1, ...), Namespace(extra_arg='This is not a recognized argument'))
```

qgallouedec: Done in 4f87b75

Issue #2387, opened 2024-11-23
eos_token config in PPOTrainer


kechunFIVE:
### Feature request

It appears that the  generation config in PPOTrainer does not set an eos_token, resulting in each generation process continuing until it reaches the maximum length before stopping, which is quite time-consuming.

### Motivation

If the `eos_token` is set, it will significantly reduce the time spent during the generation phase.

### Your contribution

i'm sorry

qgallouedec:
Thanks for reporting it. Would you like to open a PR to fix it?

dawidm:
Seems that current code is inspired by https://iclr-blogposts.github.io/2024/blog/the-n-implementation-details-of-rlhf-with-ppo/, section *General implementation details*, 4.2. Authors tried to recreate results from early OpenAI work, but they say:
> Note that in a more recent codebase https://github.com/openai/summarize-from-feedback, OpenAI does stop sampling when encountering EOS token ([summarize_from_feedback/utils/experiment_helpers.py#L19](https://github.com/openai/summarize-from-feedback/blob/8af822a428c93432aa80ffbe5b065a8f93895669/summarize_from_feedback/utils/experiment_helpers.py#L19)). However in this work we aim to do a 1:1 replication, so we align the setting that could keep sampling even eos_token is encountered

Reading `PPOTrainer` implementation, it seems that `stop_token`/`stop_token_id` arguments should control when to stop generation: 
* When `stop_token` is set to `eos` value, everything that is after EOS is padded and therefore ignored in calculations (using masks). In that case, keeping generating after all sequences have EOS seems just wasting of time and resources. I've checked that between stopping on EOS and generating until max length, loss values are almost exact and minimal differences of the order of 1e-8 are caused by masked calculations.
* When `stop_token_id` is set, sequences get padded after this token, so generation should stop there, not model's EOS.
* The case when `stop_token != None` and `stop_token != 'eos'` are not supported, so they should probably raise an exception.

Benjoyo:
The option to control the stop token should be added to all online trainer configs imho

haoxiongliu:
This is a fatal issue for training current r1-like long-cot model with extending contexts such as in the implementation of [deepscaleR](https://github.com/agentica-project/deepscaler).

haoxiongliu:
Can we also fix grpo? Or is there a convenient way to fix all online algorithms at onece?

PR #2420, opened 2024-11-30
added eos token for ppotrainer

dame-cell:
# What does this PR do?

Fixes  [2387](https://github.com/huggingface/trl/issues/2387)

- not sure if this works correctly ,might be missing some thing 
- I added  the use `use_model_eos_token` in the `ppo_conifg.py` to add  kind of a way for the user to make configuration , 
do we like this or  just make it default with no configuration ? 
- second do i need to write documentation since i added some new stuff to the config? 
 
## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@qgallouedec


trl/trainer/ppo_trainer.py
```diff
@@ -136,10 +136,16 @@ def __init__(
...
-            None  # disable `pad_token_id` and `eos_token_id` because we just want to
-        )
-        self.model.generation_config.pad_token_id = None  # generate tokens without truncation / padding
+        # Configure generation settings based on config
+        if hasattr(self.processing_class, "eos_token_id") and self.args.use_model_eos_token:
+            # Use the tokenizer's EOS token if available
+            self.model.generation_config.eos_token_id = self.processing_class.eos_token_id
+        elif hasattr(self.model.config, "eos_token_id") and self.args.use_model_eos_token:
+            # Fallback to model config's EOS token
+            self.model.generation_config.eos_token_id = self.model.config.eos_token_id
```
qgallouedec, line 145: Maybe this line is enough? Why would we need two options?

dame-cell: I remove this line 
```python 
elif hasattr(self.model.config, "eos_token_id") and self.args.use_model_eos_token:
    # Fallback to model config's EOS token
    self.model.generation_config.eos_token_id = self.model.config.eos_token_id
```
and kept the one which uses the tokenizer 

I tested it locally and it seems to be doing find the training was done perfectly with and without the eos_token_id
Do i need to like write custom tests or documentation ?

Superskyyy:
Since the other PR is merged this one can probably be closed?

Issue #2445, opened 2024-12-06
TypeError: DPOTrainer.log() takes 2 positional arguments but 3 were given


gp-1108:
### System Info

Hi, 
I was trying to run the example script from hf to finetune through DPO but I am encountering a logging error whenever I try to run the sample script.
More specificaly the error is ` TypeError: DPOTrainer.log() takes 2 positional arguments but 3 were given`.

**THE OUTPUT:**
```
The following values were not passed to `accelerate launch` and had defaults used instead:
   `--num_machines` was set to a value of `1`
   `--mixed_precision` was set to a value of `'no'`
   `--dynamo_backend` was set to a value of `'no'`
 To avoid this warning pass in values for each of the problematic parameters or run `accelerate config`.
 wandb: WARNING The `run_name` is currently set to the same value as `TrainingArguments.output_dir`. If this was not intended, please specify a different run  name by setting the `TrainingArguments.run_name` parameter.
 wandb: Using wandb-core as the SDK backend.  Please refer to https://wandb.me/wandb-core for more information.
 2024/12/06 12:23:16 INFO server is running addr=127.0.0.1:40681
 2024/12/06 12:23:16 INFO Will exit if parent process dies. ppid=736908
 2024/12/06 12:23:16 INFO connection: ManageConnectionData: new connection created id=127.0.0.1:43492
 wandb: Currently logged in as: <my user>. Use `wandb login --relogin` to force relogin
 2024/12/06 12:23:16 INFO handleInformInit: received streamId=m5gz74o1 id=127.0.0.1:43492
 2024/12/06 12:23:16 INFO handleInformInit: stream started streamId=m5gz74o1 id=127.0.0.1:43492
 wandb: Tracking run with wandb version 0.19.0
 wandb: Run data is saved locally in /<abs path>/NLP_DPO-Finetuning/llama3.1_dpo/wandb/run-20241206_122316-m5gz74o1
 wandb: Run `wandb offline` to turn off syncing.
 wandb: Syncing run Qwen2-0.5B-DPO
 wandb: ⭐️ View project at https://wandb.ai/<user info>/huggingface
 wandb: 🚀 View run at https://wandb.ai/<user info>/huggingface/runs/<user info>
 ^M  0%|          | 0/186405 [00:00<?, ?it/s]Could not estimate the number of tokens of the input, floating-point operations will not be computed
 ^M  0%|          | 1/186405 [00:01<82:07:22,  1.59s/it]Traceback (most recent call last):
[...]

PR #2447, opened 2024-12-06
Pin transformers version <4.47

kashif:
# What does this PR do?
pins the transformers version to <4.47

fixes [#2445](https://github.com/huggingface/trl/issues/2445)


lewtun (approved):
LGTM, don't forget to add a tag after merging this PR

anhuong:
Does this also need to be updated in the [requirements.txt](https://github.com/huggingface/trl/blob/main/requirements.txt#L4)? as it still shows `transformers>=4.46.0`

kashif:
@anhuong I believe the `requirements.txt` is used by the CI and the issue is fixed in the main branch...

PR #2462, opened 2024-12-11
☄️ Add support for Comet  experiment management SDK integration

yaricom:
# What does this PR do?

This PR adds integration with Comet experiment management SDK.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/integration_utils.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2024 The HuggingFace Team. All rights reserved.
```
qgallouedec, line 1: I think it's fine having this function directly in `utils.py`.

yaricom: Done

qgallouedec:
Looks good, feel free to mark it ready for review when it's ready :)

yaricom:
@qgallouedec Hi, Quentin! I can see that there are some trainer implementations that do logging of tabular data as `wandb.Table` using `Trainer.log()` method rather than using corresponding method of WandB API. 

For example:

```Python
class DPOTrainer(Trainer):
     ......

    def evaluation_loop(...)

          .....

           self.log(
                {
                    "game_log": wandb.Table(
                        columns=["Prompt", "Policy", "Ref Model"],
                        rows=[
                            [prompt, pol[len(prompt) :], ref[len(prompt) :]]
                            for prompt, pol, ref in zip(
                                random_batch["prompt"], policy_output_decoded, ref_output_decoded
                            )
                        ],
                    )
                }
            )
```

We are not sure on the best way to update this part of code in order to support other integrations like Comet, for example. 

What do you think if I change the mentioned code block to log table using WandB API instead of `Trainer.log()`? 

Something like:

```Python
            if "wandb" in self.args.report_to:
                import wandb

                if wandb.run is not None:
                    wandb.log({"game_log": wandb.Table(
                            columns=["Prompt", "Policy", "Ref Model"],
                            rows=[
                                [prompt, pol[len(prompt):], ref[len(prompt):]]
                                for prompt, pol, ref in zip(
                                    random_batch["prompt"], policy_output_decoded, ref_output_decoded
                                )
                            ],
                        )}
                    )
```

This will greatly simplify adding other integrations.

yaricom:
@qgallouedec Thank you for quick response. I noticed that `LogCompletionsCallback` is a subclass of `WandbCallback`, which requires the `wandb` module to be present; otherwise, an exception is raised.

It seems a bit out of place to leave this inheritance unchanged and simply add Comet integration to this callback. There could be situations where Comet is installed, but WandB is either not installed or not initialized (e.g., missing API key).

What do you think?

qgallouedec:
> It is possible to change `LogCompletionsCallback` inheritance to use `TrainerCallback` as superclass and then implement both integrations: wandb and Comet.
> 
> What do you think?



Yes, I think your solution makes sense. Would you like to make it as part of this PR?

yaricom:
I think it would be better to have another PR for `LogCompletionsCallback` changes to keep things more granular.

trl/templates/lm_model_card.md
```diff
@@ -20,7 +20,8 @@ print(output["generated_text"])
 
 ## Training procedure
 
-{% if wandb_url %}[<img src="https://raw.githubusercontent.com/wandb/assets/main/wandb-github-badge-28.svg" alt="Visualize in Weights & Biases" width="150" height="24"/>]({{ wandb_url }}){% endif %}
+{% if wandb_url %}[<img src="https://raw.githubusercontent.com/wandb/assets/main/wandb-github-badge-28.svg" alt="Visualize in Weights & Biases" width="150" height="24"/>]({{ wandb_url }}){% endif %} 
+{% if comet_url %}[<img src="https://raw.githubusercontent.com/comet-ml/comet-examples/master/logo/comet_badge.png" alt="Visualize in Comet" width="135" height="20"/>]({{ comet_url }}){% endif %}
```
qgallouedec, line 24: Can you make an svg instead? Or at least increase the quality?

trl/trainer/utils.py
```diff
@@ -1492,3 +1497,37 @@ def generate_model_card(
         tokenizers_version=version("tokenizers"),
     )
     return card
+
+
+def get_comet_experiment_url() -> Optional[str]:
+    """If Comet integration is enabled returns the URL of the current Comet experiment
```
qgallouedec, line 1503:
```suggestion
    """
    If Comet integration is enabled, return the URL of the current Comet experiment; otherwise, return `None`.
```

trl/templates/lm_model_card.md
yaricom: I'll ask GFX designer to fix this.

trl/trainer/utils.py
```diff
@@ -1492,3 +1497,37 @@ def generate_model_card(
...
+    Args:
+        name (`str`):
+            Table name.
+        table (`pd.DataFrame`):
+            The Pandas DataFrame containing the table to log.
+    """
+    if not is_comet_available():
+        return
+
+    import comet_ml
```
qgallouedec, line 1530: Instead, I'd return an error here

```python
if is_comet_available():
    import comet_ml

[...]

def log_table_to_comet_experiment(name: str, table: pd.DataFrame):
    if not is_comet_available():
        raise ModuleNotFoundError("...")
```

qgallouedec:
LGTM, waiting https://github.com/huggingface/trl/pull/2462#discussion_r1884306215 to be addressed then I'll approve & merge. Thanks!

trl/trainer/utils.py
yaricom: Done

qgallouedec (approved):
Thanks a lot!

PR #2474, opened 2024-12-13
🧑‍🤝‍🧑 Proper metrics gathering across ranks before logging

### LINKED ISSUE #2468: DPOTrainer log metrics are not gathered and meaned across ranks
### Feature request

synchronize and average metrics across ranks.

### Motivation

current metrics reported are only numbers on rank 0.

```python
        metrics[f"{prefix}rewards/chosen"] = chosen_rewards.mean().cpu()
        metrics[f"{prefix}rewards/rejected"] = rejected_rewards.mean().cpu()
        metrics[f"{prefix}rewards/accuracies"] = reward_accuracies.mean().cpu()
        metrics[f"{prefix}rewards/margins"] = (chosen_rewards - rejected_rewards).mean().cpu()
        metrics[f"{prefix}logps/chosen"] = model_output["chosen_logps"].detach().mean().cpu()
        metrics[f"{prefix}logps/rejected"] = model_output["rejected_logps"].detach().mean().cpu()
        metrics[f"{prefix}logits/chosen"] = model_output["mean_chosen_logits"].detach().cpu()
        metrics[f"{prefix}logits/rejected"] = model_output["mean_rejected_logits"].detach().cpu()
```

all of these aren't synced.

### Your contribution

current log function looks like:

```python
    def log(self, logs: Dict[str, float]) -> None:
        """
        Log `logs` on the various objects watching training, including stored metrics.

        Args:
            logs (`Dict[str, float]`):
                The values to log.
        """
        # logs either has 'loss' or 'eval_loss'
        train_eval = "train" if "loss" in logs else "eval"
        # Add averaged stored metrics to logs
        for key, metrics in self._stored_metrics[train_eval].items():
            logs[key] = torch.tensor(metrics).mean().item()
        del self._stored_metrics[train_eval]
        return super().log(logs)
  ```
  
  it would have this feature if it looks like:
  
  ```python
    def log(self, logs: Dict[str, float]) -> None:
        """
        Log `logs` on the various objects watching training, including stored metrics.

        Args:
            logs (`Dict[str, float]`):
                The values to log.
        """
        # logs either has 'loss' or 'eval_loss'
        train_eval = "train" if "loss" in logs else "eval"
        # Add averaged stored metrics to logs
        for key, metrics in self._stored_metrics[train_eval].items():
            if isinstance(metrics[0], torch.Tensor):
                gathered = self._nested_gather([m.cuda() for m in metrics])
                metrics = [g.mean() for g in gathered]
            meaned = torch.tensor(metrics).mean()
            logs[key] = meaned.item()
        del self._stored_metrics[train_eval]
        return super().log(logs)
```

I'm happy to submit a pr.

zhc7:
according to https://github.com/huggingface/trl/issues/2468

# What does this PR do?

Fixes #2468

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/dpo_trainer.py
```diff
@@ -1424,7 +1424,11 @@ def log(self, logs: dict[str, float], start_time: Optional[float] = None) -> Non
         train_eval = "train" if "loss" in logs else "eval"
         # Add averaged stored metrics to logs
         for key, metrics in self._stored_metrics[train_eval].items():
-            logs[key] = torch.tensor(metrics).mean().item()
+            if isinstance(metrics[0], torch.Tensor):
+                gathered = self._nested_gather([m.cuda() for m in metrics])
```
qgallouedec, line 1428: do you need `.cuda()`?

zhc7: > do you need `.cuda()`?

metrics are moved cpu before, some backends (e.g. nccl) does not support gathering tensors on cpu. but I admit `.cuda` here loses some genrality. maybe `.to(self.accelerator.device)` is better?

zhc7: > maybe `self.accelerator.gather(metrics).mean().item()` would be simpler?

I agree `self.accelerator.gather` is better. but `metrics` in the loop is a `list[torch.Tensor]` or `list[float]`, so `gather` actually returns a `list[torch.Tensor]`. So I think I change it into:

```python
            if isinstance(metrics[0], torch.Tensor):
                gathered = self.accelerator.gather([m.to(self.accelerator.device) for m in metrics])
                metrics = [g.mean() for g in gathered]
            meaned = torch.tensor(metrics).mean()
            logs[key] = meaned.item()
```

I know creating a new tensor on metrics seems a little weird, but that is how it was originally written. I don't why but I don't why to break anything so I left it there.

qgallouedec:
I realized that the problem was in fact not only confined to DPO but also to other trainers. So I thought twice about the approach, and I think the right solution is actually to gather at the time these metrics are calculated. It's the most logical and it's what's done in the other trainers.
Thanks for spotting  on the issue, I'll wait until the CI is green and merge.

kashif:
@zhc7 accelerate has a dedicated helper for metrics: https://huggingface.co/docs/accelerate/en/package_reference/accelerator#accelerate.Accelerator.gather_for_metrics which is what is recommended to be used

qgallouedec:
Valid point! thanks @kashif. I generalised `gather_for_metrics` in 17383f9

qgallouedec (approved):

PR #2482, opened 2024-12-15
🐯 [Liger] add native liger-kernel ORPO loss

kashif:
# What does this PR do?

Adds support for Liger ORPO loss kernel to the ORPO Trainer natively.


qgallouedec:
2 questions/remarks:

- can you run benchmark so that we can (1) quantify the improvement and (2) check that results with and without liger are the same
- we could have an additional tag for the hub when a model is trained with liger

tests/test_orpo_trainer.py
```diff
@@ -148,3 +148,47 @@ def test_orpo_trainer_with_lora(self, config_name):
...
+                per_device_train_batch_size=2,
+                max_steps=3,
+                remove_unused_columns=False,
+                gradient_accumulation_steps=1,
+                learning_rate=9e-1,
+                eval_strategy="steps",
+                logging_steps=1,
+                beta=0.1,
+                report_to="none",
+                use_liger_loss=True,  # Enable Liger loss
```
qgallouedec, line 167:
```suggestion
                output_dir=tmp_dir,
                eval_strategy="steps",
                report_to="none",
                use_liger_loss=True,  # Enable Liger loss
```

trl/trainer/orpo_trainer.py
```diff
@@ -357,6 +356,22 @@ def make_inputs_require_grad(module, input, output):
                 "Your `Trainer` does not have an `accelerator` object. Consider upgrading `transformers`."
             )
 
+        # Import Liger loss if enabled
```
qgallouedec, line 359: IMO, passing `self.args.use_liger_loss` without liger installed should raised an error.

qgallouedec:
I think we should bump liger version to v0.5 (it doesn't include the loss before), see https://github.com/linkedin/Liger-Kernel/releases/tag/v0.5.0

trl/trainer/orpo_trainer.py
```diff
@@ -357,6 +356,26 @@ def make_inputs_require_grad(module, input, output):
...
+                    "You set `use_liger_loss=True` but the liger kernel is not available. "
+                    "Please install liger-kernel first: `pip install liger-kernel`"
+                )
+            try:
+                from liger_kernel.chunked_loss import LigerFusedLinearORPOLoss
+
+                self.orpo_loss_fn = LigerFusedLinearORPOLoss(ignore_index=self.label_pad_token_id, beta=self.beta)
+                self._using_liger = True
+            except ImportError:
+                raise ImportError(
```
qgallouedec, line 372: If we bump the version, we shouldn't need this anymore

trl/trainer/orpo_trainer.py
```diff
@@ -68,7 +68,6 @@
 if is_peft_available():
     from peft import PeftModel, get_peft_model, prepare_model_for_kbit_training
```
qgallouedec, line 70:
```suggestion
if is_liger_kernel_available():
    from liger_kernel.chunked_loss import LigerFusedLinearORPOLoss
```

even if it is not always used, I think it's more readable when all imports are on top. 
+ If there’s an issue with a dependency, it will be detected as soon as the script runs

trl/trainer/orpo_trainer.py
```diff
@@ -756,51 +767,79 @@ def concatenated_forward(
             concatenated_batch["concatenated_input_ids"],
```
SumanthRH, line 767: I don't think this will get the full benefits of the liger kernel. This will perform full model forward pass including the last layer, and will materialize the full logits: https://github.com/huggingface/transformers/blob/eed11f34abe558c76590366a21316318e5f820ed/src/transformers/models/llama/modeling_llama.py#L1179 

Using `output_hidden_states` and then using Liger's loss based on the last hidden state + lm head is not enough. We need to do the forward pass only until the lm head. 

For Llama, we can do the forward pass on just the `LlamaModel` submodule (`model.model`) : 

```
# tiny random llama repr
LlamaForCausalLM(
  (model): LlamaModel(
    (embed_tokens): Embedding(32000, 16, padding_idx=31999)
    (layers): ModuleList(
      (0-1): 2 x LlamaDecoderLayer(
        (self_attn): LlamaSdpaAttention(
          (q_proj): Linear(in_features=16, out_features=16, bias=False)
          (k_proj): Linear(in_features=16, out_features=16, bias=False)
          (v_proj): Linear(in_features=16, out_features=16, bias=False)
          (o_proj): Linear(in_features=16, out_features=16, bias=False)
          (rotary_emb): LlamaRotaryEmbedding()
        )
        (mlp): LlamaMLP(
          (gate_proj): Linear(in_features=16, out_features=64, bias=False)
          (up_proj): Linear(in_features=16, out_features=64, bias=False)
          (down_proj): Linear(in_features=64, out_features=16, bias=False)
          (act_fn): SiLU()
        )
        (input_layernorm): LlamaRMSNorm((16,), eps=1e-06)
        (post_attention_layernorm): LlamaRMSNorm((16,), eps=1e-06)
      )
    )
    (norm): LlamaRMSNorm((16,), eps=1e-06)
    (rotary_emb): LlamaRotaryEmbedding()
  )
  (lm_head): Linear(in_features=16, out_features=32000, bias=False)
)
```

I believe the implementation might end up being model specific because we have to get the attribute name for this common backbone first (I'm not sure if all HF models have the backbone under ".model" attribute). For example, see impl here https://github.com/linkedin/Liger-Kernel/blob/0bb6c72d03f0fc570b9e954c42b723ab7e0c315f/src/liger_kernel/transformers/trainer/orpo_trainer.py#L93-L110

SumanthRH (commented):
Thanks for this PR! I stumbled upon this and wanted to highlight an important point (maybe this change is in progress already, in which case, great!)

trl/trainer/orpo_trainer.py
kashif: thanks @SumanthRH can you check the current implementation of how i was thinking about skipping the lm_head layer?

trl/trainer/orpo_trainer.py
```diff
@@ -752,55 +763,87 @@ def concatenated_forward(
...
-            # Flatten the tokens
-            loss_fct = nn.CrossEntropyLoss()
-            logits = logits.view(-1, logits.shape[-1])
-            labels = labels.view(-1)
-            # Enable model parallelism
-            labels = labels.to(logits.device)
-            loss = loss_fct(logits, labels)
-            return loss
+        if self.args.use_liger_loss:
+            # skip the lm head and get the last hidden state
```
SumanthRH, line 767: Nice! 

I guess we don't have much of an option beyond using a config parameter for now. 

Given that we run forward pass on a submodule, it would be very nice to have some validation so that there are no unexpected failures etc with different distributed training settings.  But in this case, I feel there might be compatibility issues with FSDP given the limitation from the docs: https://pytorch.org/docs/stable/fsdp.html 

"FSDP does not support running the forward pass of a submodule that is contained in an FSDP instance. This is because the submodule’s parameters will be sharded, but the submodule itself is not an FSDP instance, so its forward pass will not all-gather the full parameters appropriately."

(might be fixed by just making the base model attribute an FSDP instance as well, coz why not)

Beyond that this looks fine! I have a couple nits (don't matter that much): 
- Does `model.get_decoder()` work all the time btw for AutoModelForCausalLM instances? Was wondering if that's a cleaner solution for getting the base model attribute. But I think some base model classes have some further wrapping over the actual decoder (to format outputs, etc) https://github.com/huggingface/transformers/blob/a7f5479b45a8040392af80bf1107a2bdd796931c/src/transformers/models/opt/modeling_opt.py#L1044 
- Maybe the config is `base_model_attribute_name` since its the attribute name of the base model in the CausalLM object?

kashif: thanks @SumanthRH yes you are right `get_decoder()` will work

Yes next is to verify the distributed training cases

tests/test_orpo_trainer.py
```diff
@@ -148,3 +148,36 @@ def test_orpo_trainer_with_lora(self, config_name):
                     # check the params have changed - ignore 0 biases
                     if param.sum() != 0:
                         self.assertFalse(torch.equal(param, new_param))
+
+    @require_liger_kernel
+    def test_orpo_trainer_with_liger(self):
+        """Test ORPO trainer with Liger loss enabled."""
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            training_args = ORPOConfig(
```
austin362667, line 156: You might consider setting beta to a larger value here. By default, beta is set to 0.1, while alpha is 1. Adjusting beta could increase the weight of the preference loss relative to the NLL loss.

Reference:
[LinkedIn Liger Kernel - fused_linear_preference.py](https://github.com/linkedin/Liger-Kernel/blob/main/src/liger_kernel/chunked_loss/fused_linear_preference.py)

Parameter Details:

alpha (float): Weight for the NLL loss (default: 1).
beta (float): Weight for the preference loss (default: 0.1).

kashif: so the plot was with the same parameters with and without the liger_loss flag... rather than in general... so trying to figure out why there is a difference between the two settings...

kashif:
waiting on https://github.com/linkedin/Liger-Kernel/pull/486

kashif:
waiting on https://github.com/huggingface/trl/pull/2502

qgallouedec:
@kashif can you share the curves once it's ready?

kashif:
needs: https://github.com/linkedin/Liger-Kernel/pull/503

qgallouedec (approved):
lgtm!

shimizust:
Hi @kashif @qgallouedec thanks for the effort here! I was wondering what is the status? Can this be merged? Looks like one test failed, but not sure if it was a real issue

qgallouedec:
closing as stale

PR #2502, opened 2024-12-19
[ORPO] fix orpo chosen-nll loss

kashif:
# What does this PR do?
Calculate the ORPO chosen nll loss with respect to the chosen completion only rather than the whole prompt+compeletion.

Also return the shifted logits when the model is decoder only

### DIFF
--- a/trl/trainer/orpo_trainer.py
+++ b/trl/trainer/orpo_trainer.py
@@ -774,18 +774,12 @@ class ORPOTrainer(Trainer):
             loss = loss_fct(logits, labels)
             return loss
 
-        if self.is_encoder_decoder:
-            labels = concatenated_batch["concatenated_labels"].clone()
-        else:
-            labels = concatenated_batch["concatenated_input_ids"].clone()
-            attention_mask = concatenated_batch["concatenated_attention_mask"]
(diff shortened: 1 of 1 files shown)


trl/trainer/orpo_trainer.py
```diff
@@ -774,18 +774,12 @@ def cross_entropy_loss(logits, labels):
...
             return loss
 
-        if self.is_encoder_decoder:
-            labels = concatenated_batch["concatenated_labels"].clone()
-        else:
-            labels = concatenated_batch["concatenated_input_ids"].clone()
-            attention_mask = concatenated_batch["concatenated_attention_mask"]
-            labels = torch.where(attention_mask == 1, labels, self.label_pad_token_id)
-
+        labels = concatenated_batch["concatenated_labels"].clone()
```
qgallouedec, line 777: Yes, checked together, if you do 

```python
labels = concatenated_batch["concatenated_input_ids"].clone()
attention_mask = concatenated_batch["concatenated_attention_mask"]
labels = torch.where(attention_mask == 1, labels, self.label_pad_token_id)
```


you don't ignore the prompt.

qgallouedec (approved):

PR #2527, opened 2024-12-28
↩️ Revert ORPO loss changes

kashif:
# What does this PR do?
revert changes from #2502


qgallouedec:
Can you tell me again why it's needed?

kashif:
i was mistaken... in orpo the nll loss is over the prompt + completion

qgallouedec:
Maye we can add a comment here so that we don't revert the reversion in the future ;)

PR #2540, opened 2025-01-03
custom reward function support for ppo trainer

### LINKED ISSUE #2518: [question] best way to have my own reward model which is backed by rules
hi there,
i see that in recent versions from 0.11, especially in PPO, there is a big change that only `reward_model` of `nn.Module` is accepted for scoring. in contrast to [previous versions](https://huggingface.co/docs/trl/v0.11.4/en/ppo_trainer) that I can use my own reward module to get the scores on the fly. 

thus, my question is, if i have my own reward module, which is backed by rules instead of a LM, what is the recommended practice to integrate it into the current version ? 

what comes into my mind are two manners:

-  perhaps to change the function in `get_reward` https://github.com/huggingface/trl/blob/main/trl/trainer/utils.py#L1051 ?
-  Use my own reward module to score samples and train a module as the reward LM and then pass it into the ppo_trainer ?

My question also applied to `OnlineDPOTrainer`.

Thanks.

August-murr:
# What does this PR do?

Fixes #2518 

Adding support for a custom reward function for the PPO trainer. 

### How it works

Write a custom function that takes a list of texts as input, representing a batch of responses, and outputs a list of scores.

```python
def custom_reward_function(texts: list) -> list:
    """
    Custom reward function that applies a given reward logic to each item in the list.
    
    Args:
        items (list): List of items to evaluate.
    
    Returns:
        list: List of rewards based on the provided reward logic.
    """
    rewards = [reward_logic(item) for item in items]
```

I will add more documentation and explanations later after running several tests to make sure the implementation is functional. 
## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@qgallouedec

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -19,7 +19,7 @@ import textwrap
 import time
 from collections import defaultdict
 from contextlib import contextmanager, nullcontext
-from typing import Optional, Union
+from typing import Callable, Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -114,7 +114,7 @@ class PPOTrainer(Trainer):
         ],
         model: nn.Module,
         ref_model: Optional[nn.Module],
-        reward_model: nn.Module,
+        reward_model: Union[nn.Module, Callable],
         train_dataset: Dataset,
         value_model: Optional[nn.Module] = None,
         data_collator: Optional[DataCollatorWithPadding] = None,
@@ -218,7 +218,7 @@ class PPOTrainer(Trainer):
         # setup model, optimizer, and others
         #########
         for module in [self.policy_model, self.ref_model, self.value_model, self.reward_model]:
-            if module is not None:
+            if isinstance(module, nn.Module):
                 disable_dropout_in_model(module)
         if args.stop_token and args.stop_token == "eos":
             args.stop_token_id = processing_class.eos_token_id
@@ -285,9 +285,10 @@ class PPOTrainer(Trainer):
         self.eval_dataloader = accelerator.prepare(self.eval_dataloader)
 
         if self.is_deepspeed_enabled:
-            self.reward_model = prepare_deepspeed(
-                self.reward_model, args.per_device_train_batch_size, args.fp16, args.bf16
-            )
+            if isinstance(self.reward_model, nn.Module):
+                self.reward_model = prepare_deepspeed(
+                    self.reward_model, args.per_device_train_batch_size, args.fp16, args.bf16
+                )
 
             if self.ref_model is None:
                 if not self.is_peft_model:
@@ -302,7 +303,8 @@ class PPOTrainer(Trainer):
                     raise ValueError("No reference model and model is not a Peft model.")
             else:
                 self.ref_model = self.ref_model.to(self.accelerator.device)
-            self.reward_model = self.reward_model.to(self.accelerator.device)
+                if isinstance(self.reward_model, nn.Module):
+                    self.reward_model = self.reward_model.to(self.accelerator.device)
 
     def get_train_dataloader(self) -> DataLoader:
         return self.dataloader
@@ -457,11 +459,19 @@ class PPOTrainer(Trainer):
                     sequence_length = first_true_indices(postprocessed_response == processing_class.pad_token_id) - 1
                     unwrapped_value_model = accelerator.unwrap_model(model).value_model
                     full_value, _, _ = get_reward(
-                        unwrapped_value_model, query_response, processing_class.pad_token_id, context_length
+                        unwrapped_value_model,
+                        processing_class,
+                        query_response,
+                        processing_class.pad_token_id,
+                        context_length,
                     )
                     value = full_value[:, context_length - 1 : -1].squeeze(-1)
                     _, score, _ = get_reward(
-                        reward_model, postprocessed_query_response, processing_class.pad_token_id, context_length
+                        reward_model,
+                        processing_class,
+                        postprocessed_query_response,
+                        processing_class.pad_token_id,
+                        context_length,
                     )
 
                     responses.append(response)
@@ -713,7 +723,11 @@ class PPOTrainer(Trainer):
 
                     postprocessed_query_response = torch.cat((query, postprocessed_response), 1)
                     _, score, _ = get_reward(
-                        self.reward_model, postprocessed_query_response, processing_class.pad_token_id, context_length
+                        self.reward_model,
+                        processing_class,
+                        postprocessed_query_response,
+                        processing_class.pad_token_id,
+                        context_length,
                     )
                     table["score"].extend(self.accelerator.gather(score).float().cpu().numpy())
 
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -62,6 +62,7 @@ from .utils import (
     generate_model_card,
     get_comet_experiment_url,
     get_reward,
+    get_reward_custom,
     log_table_to_comet_experiment,
     peft_module_casting_to_bf16,
     prepare_deepspeed,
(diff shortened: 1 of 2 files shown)


August-murr:
### Question:  
Does this theoretically work? I'm asking because I haven't read the PPO papers. When the PPO trainer is training, it outputs: `query`, `model_response`, and `score`, with the score being the tensor logits from the reward model. I have tested this branch and the changes, and it looks normal and functional. 

For example, let's say the custom reward function is based on the count of a specific word, like "Good":
```python
def reward_function(texts):
    rewards = [text.count("good") for text in texts]
    return rewards
```
and the printed output is just the count of the word good in the text and it looks normal since it's in the same format.

But is there more to it? theoretically?

trl/trainer/ppo_trainer.py
```diff
@@ -713,7 +723,11 @@ def generate_completions(self, sampling: bool = False):
 
                     postprocessed_query_response = torch.cat((query, postprocessed_response), 1)
                     _, score, _ = get_reward(
-                        self.reward_model, postprocessed_query_response, processing_class.pad_token_id, context_length
+                        self.reward_model,
+                        processing_class,
+                        postprocessed_query_response,
+                        processing_class.pad_token_id,
+                        context_length,
```
qgallouedec, line 730: Can we move the `if isinstance(model, torch.nn.Module):` here? I would allow not to introduce breaking change in `get_reward`

qgallouedec:
Can you add a test as well?

August-murr:
> Can you add a test as well?

I'll take that as a yes.

Yes I will add the test and the docs later, maybe a blogpost or something to show how it works if I don't run out of resources.

trl/trainer/ppo_trainer.py
August-murr: You need to clarify what you mean.

qgallouedec: sorry, it wasn't clear:

something like this instead:

```python
if isinstance(model, torch.nn.Module):
    full_value, _, _ = get_reward(
        unwrapped_value_model, query_response, processing_class.pad_token_id, context_length
    )
else:
    full_value = ...
```

doing such we don't need to introduce a breaking change in `get_reward`.

August-murr: I mean I changed get_reward to work either way with both a callable and an nn.Module.
So you want to add `if isinstance(model, torch.nn.Module)` there and keep `get_reward` as it is without change?

Superskyyy:
Thanks for the contribution! We look forward to this flexibility added!

August-murr:
@qgallouedec 
just pushed https://github.com/huggingface/trl/pull/2540/commits/11484e724804f0fe4395b1863696f2019a458bbc to do the same thing without modifying `get_reward` 
is it better?

trl/trainer/utils.py
```diff
@@ -1097,6 +1097,14 @@ def get_reward(
         sequence_lengths,
     )
 
+def get_reward_custom(model: Callable, processor: PreTrainedTokenizerBase, query_responses: torch.Tensor) -> torch.Tensor:
+    """
+    This function ensures that the custom reward function produces the correct output structure for integration with the trainer script.
+    """
+    texts = processor.batch_decode(query_responses)
```
qgallouedec, line 1104: Should we skip special tokens here?

qgallouedec:
Yes it looks better imo!

trl/trainer/utils.py
August-murr: You, good point.
another concern:
Currently, the `postprocessed_query_response` includes both the prompt and the generated response, which are then scored by the reward model(or custom function). Should the reward model only score the generated response, or should it score both the prompt and the response together?

qgallouedec: > should it score both the prompt and the response together

Yes, and the reason is quite intuitive. For example, consider a prompt like `2+3=` with the generated response being `6`. If the reward model only has access to the generated response (`6`), how could it determine whether the calculation is correct without knowing the prompt?

trl/trainer/ppo_trainer.py
```diff
@@ -457,12 +460,20 @@ def repeat_generator():
                     sequence_length = first_true_indices(postprocessed_response == processing_class.pad_token_id) - 1
                     unwrapped_value_model = accelerator.unwrap_model(model).value_model
                     full_value, _, _ = get_reward(
-                        unwrapped_value_model, query_response, processing_class.pad_token_id, context_length
+                        unwrapped_value_model,
+                        query_response,
+                        processing_class.pad_token_id,
+                        context_length,
```
qgallouedec, line 466:
```suggestion
                        unwrapped_value_model, query_response, processing_class.pad_token_id, context_length
```

qgallouedec:
You could also add a test, with, eg, this reward fund

```python
def reward_func(text):
    return float(len(text))
```

Superskyyy:
Hi, does this same change work for the RLOO trainer?

August-murr:
> Hi, does this same change work for the RLOO trainer? (aka do the same change to the rloo trainer)

We will apply the same approach to RLOO after conducting some tests on this.

August-murr:
@qgallouedec 
When using a custom reward function, what will be the role of the `value_model`? Do I need to modify anything related to the `value_model`?

correct me if I'm wrong, but the `score` should represent the reward given by the custom reward function to generated text by the model
then what will be the `full_value` when using custom reward functions?
https://github.com/huggingface/trl/blob/88514d51e39eda05317b2506c0fa23374bebe2db/trl/trainer/ppo_trainer.py#L459-L461

Superskyyy:
I tested the branch locally and seems like it works fine.

August-murr:
> I tested the branch locally and seems like it works fine.

wut?🤔

Could you share the code you're using that works for you? 

I'm currently considering how the value model would be replaced or used, what the `full_value` should be or how it should be calculated for the trainer to function correctly. 
It seems like it currently "_shouldn't_" work.

qgallouedec:
> Do I need to modify anything related to the value_model?

No, I don't think you need to modify anything related to the value function here.

> then what will be the `full_value` when using custom reward functions?

In fact, it's the same with any reward function. The value model is trained to estimates the value of a state (= token), i.e. the expectation of the discounted future rewards. It seems to me that the current implementation is sufficient.

August-murr:
> > Do I need to modify anything related to the value_model?
> 
> No, I don't think you need to modify anything related to the value function here.
> 
> > then what will be the `full_value` when using custom reward functions?
> 
> In fact, it's the same with any reward function. The value model is trained to estimates the value of a state (= token), i.e. the expectation of the discounted future rewards. It seems to me that the current implementation is sufficient.

When using it, what should I set as the `value_model`? It's says `[optional]`, but it it breaks at `PolicyAndValueWrapper` if I don't provide a `value_model`. I also can't use the reward function as the `value_model`.

Superskyyy:
> > > Do I need to modify anything related to the value_model?
> > 
> > 
> > No, I don't think you need to modify anything related to the value function here.
> > > then what will be the `full_value` when using custom reward functions?
> > 
> > 
> > In fact, it's the same with any reward function. The value model is trained to estimates the value of a state (= token), i.e. the expectation of the discounted future rewards. It seems to me that the current implementation is sufficient.
> 
> @qgallouedec When using it, what should I set as the `value_model`? It's says `[optional]`,
> 
> https://github.com/huggingface/trl/blob/d9f056862f2bd1514fad068f49f6bc05d8494d4a/trl/trainer/ppo_trainer.py#L119
> 
> 
> but it it breaks at `PolicyAndValueWrapper` if I don't provide a `value_model`. I also can't use the reward function as the `value_model`.

Oh i get what you mean, @August-murr @qgallouedec I think typically the value model is initialized same as the policy model trained, but when we dont have the policy model what do we specify as the value model?

Superskyyy:
In the meanwhile I implemented the same for RLOO and can confirm it works (apparently)

musab-mk:
What @Superskyyy said is a real concern. We still need Value model. And probably trained just like a reward model? So in that case, using a custom function might still require another trained classifier. 

Would using the SFT pre-trained base as Value model work?

Superskyyy:
> What @Superskyyy said is a real concern. We still need Value model. And probably trained just like a reward model?
> 
> Would using the SFT pre-trained base as Value model work? Anyone has experimented with this?

In theory it should. There are two ways of initiating a value model, either from policy or the trained reward.

PR #2556, opened 2025-01-10
Add generation caching in TextEnvironment and fix bugs in TextEnvironment

konrad-gerlach:
This PR mainly affects the TextEnvironment class and adds caching in between generation calls, in order to not have to recompute all previous activations when generating the next segment. This is mainly intended for use cases where many tool calls are performed sequentially and thus the activations for the (possibly quite large) system prompt would have to be calculated at each step. For stability, caching is optional.

Bug fixes:
This issue also addresses two bugs I encountered:
1. max_length checking in TextEnvironment class threw an error, as it assumed batching was present, when no batching existed.
I fixed the bug and also added a check at generation time to ensure, that the padded inputs also do not exceed max length.
2. The StringStoppingCriteria did not take generated eos tokens into account, which I have now fixed.

RE testing:
I only made sure, that the tests in tests/test_environments.py were completing.
Using `make test` some tests were failing and the tests were taking a long time to run. However, the only tests, which call TextEnvironment seem to be in test_environments.py, so the rest should be unaffected as far as I know. Nevertheless, I would be grateful, if somebody else could run all the tests before merging. I suspect, that my environment may not be ideally configured. Is testing automated via a CI?


konrad-gerlach:
I would be very grateful for a review by:
@lvwerra
@vwxyzjn
@younesbelkada
@qgallouedec
or any others, that feel up to the task.

konrad-gerlach:
I was unable to execute the pre-commit hook, so I manually ran the linter.

trl/environment/base_environment.py
```diff
@@ -241,6 +253,7 @@ def __init__(
             max_tool_response (Optional[int]): The maximum number of characters to allow in a tool response.
             max_length (Optional[int]): The maximum number of tokens to allow in an episode.
             generation_kwargs (Optional[dict]): A dictionary of keyword arguments to pass to the model's generate method.
+            use_cache (bool): Whether or not to cache past_key_values between segments. When using caching, TextEnvironment is not suited for training use, i.e. backpropagation through the generated graph. Use with Trainers is of course possible. Furthermore, caching requires, that there be no calculation dependencies between examples at inference time. When using BatchNorm, the model should thus be in eval mode.
```
qgallouedec, line 256:
```suggestion
            use_cache (bool): Whether to cache past_key_values between segments. When using caching, [`TextEnvironment`] is not suited for training use, i.e. backpropagation through the generated graph. Use with Trainers is of course possible. Furthermore, caching requires, that there be no calculation dependencies between examples at inference time. When using `BatchNorm`, the model should thus be in eval mode.
```

qgallouedec:
Thanks for the PR!
Let's see what's the CI outputs.

konrad-gerlach:
Just to be sure, as I'm unfamiliar with their implementation: The trl Trainers like PPO should not try to back propagate through the generated tokens, right? I'm 99% sure that's impossible...

konrad-gerlach:
The CI failing for Python 3.9 seems unrelated to this PR.

qgallouedec:
> The trl Trainers like PPO should not try to back propagate through the generated tokens, right?

Yes that's correct. The backprop is done on the output of a forward pass

konrad-gerlach:
@qgallouedec Could you run the precommit to fix the linting issues? I haven't gotten it to work.

konrad-gerlach:
I'm still working on adding some more tests and cleaning up the code a bit.

qgallouedec:
Ok, ping me when it's ready, I'll run the precommits and merge

konrad-gerlach:
For future reference: Setting position_ids for generation did not appear necessary, as it seems to be handled here: https://github.com/huggingface/transformers/blob/241c04d36867259cdf11dbb4e9d9a60f9cb65ebc/src/transformers/generation/utils.py#L409-L416

konrad-gerlach:
cache_position also appears to be handled automatically here https://github.com/huggingface/transformers/blob/241c04d36867259cdf11dbb4e9d9a60f9cb65ebc/src/transformers/generation/utils.py#L799-L806 and here https://github.com/huggingface/transformers/blob/241c04d36867259cdf11dbb4e9d9a60f9cb65ebc/src/transformers/generation/utils.py#L1560

Looking at the transformers generation code, it seems like there are currently issues with torch compile in transformers generate (mentioned in comment), see https://github.com/huggingface/transformers/blob/2e752ead46a8845e8a160d2043c1336447895690/src/transformers/generation/utils.py#L1582 -> I think, I will include a warning in the docs.

konrad-gerlach:
In the previous commits, I also fixed what I believe to be an off-by-one error in the StringStoppingCriteria: an inaccurate number for generated_tokens.

konrad-gerlach:
I noticed, that GPT2 seems to only support the legacy cache format, so I am adding support for this,

konrad-gerlach:
> In the previous commits, I also fixed what I believe to be an off-by-one error in the StringStoppingCriteria: an inaccurate number for generated_tokens.

It appears, that it was still not fixed. Working on a solution and on testing StringStoppingCriteria

konrad-gerlach:
As the tests did not include a encoder-decoder architecture,  I did not test for it either. I think, that this is out of scope for this Pull Request. Where this was of concern in _generate_batched, I mirrored the implementation already provided.

konrad-gerlach:
I added a TODO in the code, that the batch_size parameter should be exposed to the user. I think this is out of scope for this PR.

konrad-gerlach:
Should there be a general warning for TextEnvironment, that support for Encoder Decoder is not (automatically) tested?

konrad-gerlach:
@qgallouedec I am ready for review. Please also read the comments/ questions above and do the pre-commit hooks (as discussed). I hope you like the code :)

konrad-gerlach:
A notable behavioural change of this PR is, that if input_ids exceeds TextEnvironment.max_length we truncate all sequences instead of allowing the model to error. This technically breaks the isolation of the samples within the batch, as a history might be truncated due to another history being too long, but I think this is acceptable behaviour for now. Future work might remove the longest sequence, truncate its history and continue generating using the remaining histories. This would require considerable work in my opinion.

konrad-gerlach:
@qgallouedec Could you please review or respond some other way? That would be most helpful.

tests/test_environments.py
```diff
@@ -276,3 +341,340 @@ def test_text_environment_run(self, mock_generate):
...
+            tools=[DummyTool()],
+            reward_fn=lambda x: torch.tensor(1),
+            prompt="I am a prompt!\n",
+        )
+        self.assertTrue(env._same_is_none("", ""))
+        self.assertTrue(env._same_is_none(None, None))
+        self.assertFalse(env._same_is_none("", None))
+        self.assertFalse(env._same_is_none(None, ""))
+        self.assertTrue(env._same_is_none(None))
+        self.assertTrue(env._same_is_none(""))
```
qgallouedec, line 445: We usually don't test private methods

qgallouedec:
Thanks for the PR! I don't have enough time to review it in great details unfortunately, but overall it look good to me. I'll run the pre-commits and merge

tests/test_environments.py
konrad-gerlach: Do you want me to remove a test, that's already written? I can of course, although for the other private methods I think the tests are sensible.

konrad-gerlach:
It seems as though Qwen2 removed support for the legacy cache (https://github.com/huggingface/transformers/compare/v4.47.1..v4.48.2 ,see src/transformers/models/qwen2/modeling_qwen2.py old lines 832-851). I am therefore removing the parametrization for the failing tests. There is still test_cache_class_support, testing that the cache is passed in legacy format.

konrad-gerlach:
It seems to me, that the current StringStoppingCriteria is probably incompatible with beamsearch (and probably was before as well), so I may look into that in a future issue/PR. I have not tested this theory yet.

konrad-gerlach:
@qgallouedec Thanks for the review! I have addressed the private method tests. Tests should be working now, too. Please rerun the CI, run the pre-commits and review/merge.

konrad-gerlach:
@qgallouedec Code quality issue should be fixed.

konrad-gerlach:
@qgallouedec Should be fixed, please rerun.

konrad-gerlach:
@qgallouedec Could you please run the CI?

konrad-gerlach:
@qgallouedec I saw, that you merged main into this branch. Could you please also review, so that we can merge and close this PR?

qgallouedec:
Closing as TextEnvironment is deprecated

PR #2558, opened 2025-01-10
🏎️ vllm for Online DPO

qgallouedec:
# What does this PR do?

Use vLLM for generation. 2.2x faster 🚀🚀

Demo:

```python
from trl import OnlineDPOConfig, OnlineDPOTrainer, PairRMJudge

model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen2-0.5B-Instruct")
tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen2-0.5B-Instruct")
judge = PairRMJudge()
train_dataset = load_dataset("trl-lib/ultrafeedback-prompt", split="train")

training_args = OnlineDPOConfig(output_dir="Qwen2-0.5B-OnlineDPO-vllm", logging_steps=10, use_vllm=True, gradient_accumulation_steps=8)
trainer = OnlineDPOTrainer(
    model=model, judge=judge, args=training_args, processing_class=tokenizer, train_dataset=train_dataset
)

trainer.train()
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/online_dpo_config.py
+++ b/trl/trainer/online_dpo_config.py
@@ -37,6 +37,10 @@ class OnlineDPOConfig(TrainingArguments):
             Name of the judge to use. Either `judge` or `reward_model_path` must be set, but not both.
         max_new_tokens (`int`, *optional*, defaults to `64`):
             Maximum number of tokens to generate per completion.
+        max_length (`int`, *optional*, defaults to `256`):
+            Maximum total length of the sequence (prompt + completion) used to compute log probabilities. If the
+            sequence exceeds this limit, the leftmost tokens will be truncated to preserve as much of the completion as
+            possible.
         temperature (`float`, *optional*, defaults to `0.9`):
             Temperature for sampling. The higher the temperature, the more random the completions.
         missing_eos_penalty (`float` or `None`, *optional*, defaults to `None`):
@@ -84,6 +88,14 @@ class OnlineDPOConfig(TrainingArguments):
     max_new_tokens: int = field(
         default=64,
         metadata={"help": "Maximum number of tokens to generate per completion."},
+    )
+    max_length: int = field(
+        default=256,
+        metadata={
+            "help": "Maximum total length of the sequence (prompt + completion) used to compute log probabilities. If "
+            "the sequence exceeds this limit, the leftmost tokens will be truncated to preserve as much of the "
+            "completion as possible."
+        },
     )
     temperature: float = field(
         default=0.9,
--- a/trl/trainer/online_dpo_config.py
+++ b/trl/trainer/online_dpo_config.py
@@ -58,6 +58,8 @@ class OnlineDPOConfig(TrainingArguments):
             Number of processes to use for processing the dataset.
         disable_dropout (`bool`, *optional*, defaults to `True`):
             Whether to disable dropout in the model and reference model.
+        use_vllm (`bool`, *optional*, defaults to `False`):
+            Whether to use the vLLM for generating completions. Requires vLLM to be installed (`pip install vllm`).
     """
 
     learning_rate: float = field(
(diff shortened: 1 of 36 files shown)


qgallouedec:
we get different results with vllm. probably linked to sampling param. investigating

qgallouedec:
CI fails because in the latest transformers version release yesterday, transformers uses a python 3.10+ syntax (`timeout: float | None = None`). I'm not sure why it fails only for the cli test, but I think we can safely ignore it.

trl/trainer/online_dpo_config.py
```diff
@@ -37,6 +37,10 @@ class OnlineDPOConfig(TrainingArguments):
             Name of the judge to use. Either `judge` or `reward_model_path` must be set, but not both.
         max_new_tokens (`int`, *optional*, defaults to `64`):
             Maximum number of tokens to generate per completion.
+        max_length (`int`, *optional*, defaults to `256`):
+            Maximum total length of the sequence (prompt + completion) used to compute log probabilities. If the
+            sequence exceeds this limit, the leftmost tokens will be truncated to preserve as much of the completion as
+            possible.
```
qgallouedec, line 43: To avoid OOM for long prompts

trl/trainer/online_dpo_trainer.py
```diff
@@ -241,14 +248,36 @@ def __init__(
...
+                top_k=50,
+                top_p=1.0,
+                detokenize=False,  # to avoid vllm to decode (we don't need it)
+            )
+            # vLLM dynamically adjusts the size of the key-value cache based on available GPU memory at instanciation.
+            # A larger cache size improves speed, so we would expect gpu_memory_utilization=1.
+            # However, at this stage, the optimizer's weights are not yet loaded onto the GPU; they will be loaded
+            # after the first optimizer step and remain in GPU memory throughout training. So we must reserve enough
+            # space for them. Setting gpu_memory_utilization to 0.6 seems to work well in practice.
+            self.llm = LLM(model=model.name_or_path, gpu_memory_utilization=0.55, dtype=torch.float32)
```
qgallouedec, line 270: `dtype` and `gpu_memory_utilization`  are hardcoded but we can still make them as arg in the future.
See here why using `torch.float32` is important.

qgallouedec:
Surprisingly, the precision of the generator model seems to have a pretty high impact on the results:

Screenshot 2025-01-12 at 15 31 31

When you keep the default precision (bfloat16), the results seem to be significantly worst. It may be related to higher noise in generation? Making the temperature vary might help to confirm this intuition. In the meantime, I've hard-coded float32 as model precision. It hurts a bit the speed (you've less space for KV cache) but it's still way faster than without vLLM.

kashif (approved):

edbeeching:
Awesome addition. If I understand correctly there are now two instances of the model, so this will only work in with DDP and small models where memory capacity is not an issue?

qgallouedec:
> there are now two instances of the model

Correct: one trained model and one generation model. Note that the weights of these two models are always the same. I wonder if we could share them instead of duplicating them? @hmellor any idea?

> small models where memory capacity is not an issue

Yes, parallelism should make it possible to relax this constraint.

hmellor:
https://github.com/vllm-project/vllm/pull/10353 is probably the closest we can currently get to directly accessing vLLM's model, but this PR is expected to be superseded by work done in the V1 engine re-architecture that is ongoing (planned beta-release any day now, on by default by end of January).

https://github.com/vllm-project/vllm/pull/12084 introduces some RLHF features to vLLM, but it follows the OpenRLHF model where training and inference processes live on different GPUs.

skandermoalla:
@qgallouedec great feature! How do you see this being used with distributed training? Did you test with DDP or ZeRO-3?
And would this work with LoRA?

LuisVasquezBSC:
> > there are now two instances of the model
> 
> Correct: one trained model and one generation model. Note that the weights of these two models are always the same. I wonder if we could share them instead of duplicating them? @hmellor any idea?
> 
> > small models where memory capacity is not an issue
> 
> Yes, parallelism (not supported yet) should make it possible to relax this constraint.

If your reference model is not the same as the model being trained, you can have a much bigger reference model running on VLLM. See [this paper](https://arxiv.org/abs/2407.13709)

> Our experiments reveal an interesting finding: stronger reference models can indeed offer more benefits than the SFT model, but only when they are compatible with the model being fine-tuned. Specifically, this compatibility likely stems from the intrinsic similarities of models, such as model architectures or pre-training corpora.

So, with this PR, you could have a 70B reference model on VLLM and a 13B model in training!

PR #2565, opened 2025-01-13
👨‍👨‍👧‍👧 GRPO

qgallouedec:
# What does this PR do?

```python
from peft import LoraConfig

# Load the dataset

training_args = GRPOConfig(
    output_dir="Qwen2-0.5B-GRPO",
    learning_rate=1e-5,
    logging_steps=10,
    gradient_accumulation_steps=16,
    max_completion_length=128,
)
trainer = GRPOTrainer(
    reward_model="weqweasdas/RM-Gemma-2B",
    args=training_args,
    train_dataset=dataset,
    peft_config=LoraConfig(task_type="CAUSAL_LM"),
)

trainer.train()
```

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


qgallouedec:
From the paper:

$$\mathcal{J}_{\text{GRPO}}(\theta) =
\frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|} 
 \left[\min \left(
\frac{\pi_\theta(o_{i,t} | q, o_{i,<t})}{\pi_{\theta_{\text{old}}}(o_{i,t} | q, o_{i,<t})} \hat{A}_{i,t}, 
\text{clip}\left(
\frac{\pi_\theta(o_{i,t} | q, o_{i,<t})}{\pi_{\theta_{\text{old}}}(o_{i,t} | q, o_{i,<t})}, 1 - \epsilon, 1 + \epsilon
\right) \hat{A}_{i,t}
\right)
- \beta \mathbb{D}_{\text{KL}}\left[\pi_\theta \| \pi_{\text{ref}}\right]
\right]$$

where:
- $G$ is the number of generations per prompt
- $o_i$ is the $i$-th generation of the prompt and $|o_i|$ is the number of tokens in $o_i$
- $q$ is the prompt
- $\pi_\theta$ is the policy model
- $\pi_{\theta_{\text{old}}}$ is the policy model before the update
- $\pi_{\text{ref}}$ is the reference policy
- $\hat{A}_{i,t}$ is the advantage estimate for the $t$-th token in the $i$-th generation (see under)
- $\epsilon$ and $\beta$ are hyperparameters

In Section 4.2 we can read:

> The policy model only has a single update following each exploration stage.

It implies that $\pi_{\theta_{\text{old}}} = \pi_\theta$. Consequently,

$$\mathcal{J}_{\text{GRPO}}(\theta) = \frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|}  \left[\hat{A}_{i,t}- \beta \mathbb{D}_{\text{KL}}\left[\pi_\theta \| \pi_{\text{ref}}\right]\right]$$
$$\mathcal{J}_{\text{GRPO}}(\theta) = \frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|} \hat{A}_{i,t}- \beta \frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|}\mathbb{D}_{\text{KL}}\left[\pi_\theta \| \pi_{\text{ref}}\right]$$

from 4.1.2, $\hat{A}_{i,t}$ is defined as

$$\hat{A}_{i,t} = \tilde{r}_i = \frac{r_i - \text{mean}(\mathbf{r})}{\text{std}(\mathbf{r})},$$

where $\mathbf{r} = \{r_1, r_2, \ldots, r_{G}\}$. It implies that $\hat{A}_{i,t}$ doesn't depend on $t$. Let's use the notation $\tilde{r}_i$ instead for clarity. So

$$\mathcal{J}_{\text{GRPO}}(\theta) = \frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|} \tilde{r}_i- \beta \frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|}\mathbb{D}_{\text{KL}}\left[\pi_\theta \| \pi_{\text{ref}}\right]$$
$$\mathcal{J}_{\text{GRPO}}(\theta) = \frac{1}{G} \sum_{i=1}^G \tilde{r}_i- \beta \frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|}\mathbb{D}_{\text{KL}}\left[\pi_\theta \| \pi_{\text{ref}}\right]$$

we know that $\{\tilde{r}_1, \tilde{r}_2, \ldots, \tilde{r}_{G}\}$ is normalized, so the mean is 0. It comes

$$\mathcal{J}_{\text{GRPO}}(\theta) = - \beta \frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|}\mathbb{D}_{\text{KL}}\left[\pi_\theta \| \pi_{\text{ref}}\right]$$

As a result, the GRPO objective just minimizes the KL divergence between the policy model and the reference policy.

qgallouedec:
@ZhihongShao if you can help by any chance

qgallouedec:
I have the answer to the question above.

The math is correct; if you look at the loss, it is indeed equal to the KL value.

However, in terms of differentiation, we cannot remove 

```math
\frac{\pi_\theta(o_{i,t} | q, o_{i,< t})}{\pi_{\theta_{\text{old}}}(o_{i,t} | q, o_{i,< t})}
```

justifying that it equals 1.

Let’s denote the stop-gradient operator as \(\left[\cdot\right]_\cancel{\nabla}\).

We must retain the term 

```math
\frac{\pi_\theta(o_{i,t} | q, o_{i,< t})}{\pi_{\theta_{\text{old}}}(o_{i,t} | q, o_{i,< t})}
```

in the equation, rewriting it as 

```math
\frac{\pi_\theta(o_{i,t} | q, o_{i,< t})}{\pi_{\theta_{\text{old}}}(o_{i,t} | q, o_{i,< t})} = \frac{\pi_\theta(o_{i,t} | q, o_{i,< t})}{\left[\pi_\theta(o_{i,t} | q, o_{i,< t})\right]_\cancel{\nabla}}.
```

Finally, the objective is written as 

```math
\mathcal{J}_{\text{GRPO}}(\theta) = \frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|} \left[ \frac{\pi_\theta(o_{i,t} | q, o_{i,< t})}{\left[\pi_\theta(o_{i,t} | q, o_{i,< t})\right]_\cancel{\nabla}} \hat{A}_{i,t} - \beta \mathbb{D}_{\text{KL}}\left[\pi_\theta \| \pi_{\text{ref}}\right] \right].
```

In the end, the value remains equal to the KL divergence as initially stated. However, when implemented in this way, the gradient can propagate through the equation, allowing the policy to update effectively.

Thanks @edbeeching for helping me with this!!!

tests/test_grpo_trainer.py
```diff
@@ -0,0 +1,147 @@
...
+        # the reward has an impact.
+        with torch.no_grad():
+            reward_model.score.weight *= 1000
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            training_args = GRPOConfig(
+                output_dir=tmp_dir,
+                learning_rate=0.1,  # increase the learning rate to speed up the test
+                per_device_train_batch_size=2,  # reduce the batch size to reduce memory usage
+                num_generations=3,  # reduce the number of generations to reduce memory usage
```
qgallouedec, line 127:
```suggestion
                num_generations=5,  # reduce the number of generations to reduce memory usage
```

trl/trainer/grpo_trainer.py
```diff
@@ -0,0 +1,351 @@
...
+        mean_grouped_rewards = rewards.view(-1, self.num_generations).mean(dim=1)
+        std_grouped_rewards = rewards.view(-1, self.num_generations).std(dim=1)
+
+        # Normalize the rewards to compute the advantages
+        mean_grouped_rewards = mean_grouped_rewards.repeat_interleave(self.num_generations, dim=0)
+        std_grouped_rewards = std_grouped_rewards.repeat_interleave(self.num_generations, dim=0)
+        advantages = (rewards - mean_grouped_rewards) / (std_grouped_rewards + 1e-4)
+
+        # x - x.detach() allows for preserving gradients from x
+        advatages = torch.exp(per_token_logps - per_token_logps.detach()) * advantages.unsqueeze(1)
```
edbeeching, line 272: typo `advatages`

edbeeching (approved):
Clean implementation, I had to check the KL term as it if different from PPO. But it appears to match the paper.

.github/workflows/build_pr_documentation.yml
```diff
@@ -9,7 +9,7 @@ concurrency:
 
 jobs:
   build:
-    uses: huggingface/doc-builder/.github/workflows/build_pr_documentation.yml@main
+    uses: huggingface/doc-builder/.github/workflows/build_pr_documentation.yml@e4fcf608695cf4bddb8c7f4f72aa15fa14110a94
```
qgallouedec, line 12: temporary pinning to a previous version, the latest one doesn't work

SonuDixit:
Hi Team,
1. Initially KL loss is zero(its minimum), since the ref and initial model are the same. So, the backward_gradient = 0 from this loss

2. If we update only once, the first loss term (= sum of normalised advantage for all generations) is zero. In this case, the gradient of loss=prob*(sum_of_advantages) with respect to the prob is zero. Hence, all the backward gradients should be zero for SGD(without weight decay) optimisation. 

From 1,2 above, the backward gradient is zero, if we update policy only once. The network may not update if we use simple optimiser like SGD.

XiaofengZHOU:
> Hi Team,
> 
> 1. Initially KL loss is zero(its minimum), since the ref and initial model are the same. So, the backward_gradient = 0 from this loss
> 2. If we update only once, the first loss term (= sum of advantage for all generations) is zero. In this case, the gradient of loss(=prob*(sum_of_advantages)) with respect to the prob is zero. Hence, all the backward gradients should be zero for SGD(without weight decay) optimisation.
> 
> From 1,2 above, the backward gradient is zero, if we update policy only once. The network may not update if we use simple optimiser like SGD.

I have the same question

zachluo:
@SonuDixit 
1. Initially KL loss is zero(its minimum), since the ref and initial model are the same. But after that, the initial policy model will be updated (let's assume the loss is not zero) and the reference model (initialized from SFT model) will not be updated at all. 
2. the first loss term is not zero as pi / pi_old = 1 or torch.exp(per_token_logps - per_token_logps.detach()) = 1. Also, advantage is not zero too so that per_token_loss = -torch.exp(per_token_logps - per_token_logps.detach()) * advantages.unsqueeze(1) will not become zero.

zachluo:
> I have the answer to the question above.
> 
> The math is correct; if you look at the loss, it is indeed equal to the KL value.
> 
> However, in terms of differentiation, we cannot remove
> 
> 𝜋
> 𝜃
> (
> 𝑜
> 𝑖
> ,
> 𝑡
> |
> 𝑞
> ,
> 𝑜
> 𝑖
> ,
> <
> 𝑡
> )
> 𝜋
> 𝜃
> old
> (
> 𝑜
> 𝑖
> ,
> 𝑡
> |
> 𝑞
> ,
> 𝑜
> 𝑖
> ,
> <
> 𝑡
> )
> justifying that it equals 1.
> 
> Let’s denote the stop-gradient operator as [ ⋅ ] ∇ .
> 
> We must retain the term
> 
> 𝜋
> 𝜃
> (
> 𝑜
> 𝑖
> ,
> 𝑡
> |
> 𝑞
> ,
> 𝑜
> 𝑖
> ,
> <
> 𝑡
> )
> 𝜋
> 𝜃
> old
> (
> 𝑜
> 𝑖
> ,
> 𝑡
> |
> 𝑞
> ,
> 𝑜
> 𝑖
> ,
> <
> 𝑡
> )
> in the equation, rewriting it as
> 
> 𝜋
> 𝜃
> (
> 𝑜
> 𝑖
> ,
> 𝑡
> |
> 𝑞
> ,
> 𝑜
> 𝑖
> ,
> <
> 𝑡
> )
> 𝜋
> 𝜃
> old
> (
> 𝑜
> 𝑖
> ,
> 𝑡
> |
> 𝑞
> ,
> 𝑜
> 𝑖
> ,
> <
> 𝑡
> )
> =
> 𝜋
> 𝜃
> (
> 𝑜
> 𝑖
> ,
> 𝑡
> |
> 𝑞
> ,
> 𝑜
> 𝑖
> ,
> <
> 𝑡
> )
> [
> 𝜋
> 𝜃
> (
> 𝑜
> 𝑖
> ,
> 𝑡
> |
> 𝑞
> ,
> 𝑜
> 𝑖
> ,
> <
> 𝑡
> )
> ]
> ∇
> .
> Finally, the objective is written as
> 
> 𝐽
> GRPO
> (
> 𝜃
> )
> =
> 1
> 𝐺
> ∑
> 𝑖
> =
> 1
> 𝐺
> 1
> |
> 𝑜
> 𝑖
> |
> ∑
> 𝑡
> =
> 1
> |
> 𝑜
> 𝑖
> |
> [
> 𝜋
> 𝜃
> (
> 𝑜
> 𝑖
> ,
> 𝑡
> |
> 𝑞
> ,
> 𝑜
> 𝑖
> ,
> <
> 𝑡
> )
> [
> 𝜋
> 𝜃
> (
> 𝑜
> 𝑖
> ,
> 𝑡
> |
> 𝑞
> ,
> 𝑜
> 𝑖
> ,
> <
> 𝑡
> )
> ]
> ∇
> 𝐴
> ^
> 𝑖
> ,
> 𝑡
> −
> 𝛽
> 𝐷
> KL
> [
> 𝜋
> 𝜃
> ∥
> 𝜋
> ref
> ]
> ]
> .
> In the end, the value remains equal to the KL divergence as initially stated. However, when implemented in this way, the gradient can propagate through the equation, allowing the policy to update effectively.
> 
> Thanks @edbeeching for helping me with this!!!

Should we implement the case that \mu > 1 to enable more flexible training?

zachluo:
> The policy model only has a single update following each exploration stage.

Are we going to implement the case that the policy model has multiple updates following each exploration stage?

PR #2567, opened 2025-01-13
WIP: Base Online Trainer

mnoukhov:
# What does this PR do?

Make `RLOOTrainer` and `PPOTrainer` follow the same structure as `OnlineDPOTrainer` to leverage HF Trainer utilities e.g. checkpointing. As discussed with @qgallouedec 

Done
- Create `BaseOnlineTrainer` class for all online RLHF methods
- Restructure `OnlineDPOTrainer` to use `BaseOnlineTrainer`

To Do
- Create `RLOOv2Trainer` that inherits from `BaseOnlineTrainer`
- Create `PPOv2Trainer` that inherits from `BaseOnlineTrainer`

## Before submitting
- [x ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),Pull Request section?
- [ ] Did you write any new necessary tests? No new tests necessary

## Who can review?
@qgallouedec @kashif


mnoukhov:
Currently, changing RLOO/PPO to use `BaseOnlineTrainer` would remove the functionality of generating K minibatches from your model and then doing `K` updates on the generated data. This is not present in the OnlineDPO code but it is a standard practice in RLHF.

To solve this, in either this PR or a future PR, I propose moving the LLM generation logic into [`Trainer.get_batch_samples`](https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L2480) 
- `get_batch_samples` would take prompt inputs and return prompts, completions, and maybe logprobs of completions under the generating model
- separates generation logic from training logic, allowing for cleaner code / code reuse
- by simply modifying `num_batch_samples` to be `num_batch_samples * K` allows for generating `K` minibatches from a model and `Trainer` logic should handle everything

Looking for feedback!

qgallouedec:
Hi @mnoukhov,  

### Base Online Trainer

I understand the motivation behind this PR. However, from a philosophical standpoint, I believe that, except in rare cases, duplicated code should be preferred over inheritance in this context.

We've attempted to implement a base trainer several times in the past—whether as a general solution or specifically for online methods—and found that it tends to add more complexity than it resolves (harder to maintain, develop, debug). Even recently, with NashMD and XPO inheriting from Online DPO, I wonder if avoiding inheritance might have resulted in a simpler and clearer design.

Therefore, unless the differences between two trainers are truly minimal—such as requiring the override of just one method—I would advise against taking this approach.


### RLOO refactoring

Absolutely agree on this point. I wonder if we can skip the "v2" stage. RLOO was added recently, so I believe we can afford to break it (important: TRL is still in alpha).

For users, however, we should ensure there's a clear migration guide to help them transition smoothly.

mnoukhov:
Closing in favour of #2724

PR #2606, opened 2025-01-22
🏆 Custom reward function for GRPO

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #2612, opened 2025-01-23
🍭 Custom reward function for RLOO

August-murr:
Similar to #2540 

Since RLOO doesn't use a value model, it's much simpler.

### DIFF
--- a/trl/trainer/rloo_trainer.py
+++ b/trl/trainer/rloo_trainer.py
@@ -18,7 +18,7 @@ import os
 import textwrap
 import time
 from collections import defaultdict
-from typing import Optional, Union
+from typing import Callable, Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -54,6 +54,7 @@ from ..trainer.utils import (
     first_true_indices,
     forward,
     get_reward,
+    get_reward_custom,
     prepare_deepspeed,
     print_rich_table,
     truncate_response,
@@ -79,7 +80,7 @@ class RLOOTrainer(Trainer):
         ],
         policy: nn.Module,
         ref_policy: nn.Module,
-        reward_model: nn.Module,
+        reward_model: Union[nn.Module, Callable],
         train_dataset: Dataset,
         data_collator: Optional[DataCollatorWithPadding] = None,
         eval_dataset: Optional[Union[Dataset, dict[str, Dataset]]] = None,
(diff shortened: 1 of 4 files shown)


trl/trainer/utils.py
```diff
@@ -1198,6 +1198,18 @@ def get_reward(
...
 
+def get_reward_custom(
+    model: Callable, processor: PreTrainedTokenizerBase, query_responses: torch.Tensor
+) -> torch.Tensor:
+    """
+    This function ensures that the custom reward function produces the correct output structure for integration with the trainer script.
+    """
+    texts = processor.batch_decode(query_responses, skip_special_tokens=True)
+    rewards = model(texts)
+    rewards = torch.tensor(rewards, dtype=torch.float).to(query_responses.device)
```
qgallouedec, line 1209: I'm wondering, since it's only three lines and used once, maybe you can put it directly in RLOO?

trl/trainer/rloo_trainer.py
```diff
@@ -79,7 +80,7 @@ def __init__(
         ],
         policy: nn.Module,
         ref_policy: nn.Module,
-        reward_model: nn.Module,
+        reward_model: Union[nn.Module, Callable],
```
qgallouedec, line 83:
```suggestion
        reward_model: Union[nn.Module, Callable[[list[str]], list[float]],
```

qgallouedec:
Nice!

Can be done in a follow-up PR, but it might make sense to rename `reward_model` to `reward_func`. See https://github.com/huggingface/trl/pull/2606#discussion_r1925849928 for more context.

trl/trainer/utils.py
August-murr: but it could be used in other trainers that use `get_reward` like ppo or online dpo

August-murr:
> Nice!
> 
> Can be done in a follow-up PR, but it might make sense to rename `reward_model` to `reward_func`. See [#2606 (comment)](https://github.com/huggingface/trl/pull/2606#discussion_r1925849928) for more context.

yea I think it's better to do it for all trainers in a separate PR

qgallouedec (changes requested):
We just need a unit test and we're good to merge

trl/trainer/rloo_trainer.py
```diff
@@ -350,9 +353,17 @@ def repeat_generator():
...
-                    _, score, _ = get_reward(
-                        reward_model, postprocessed_query_response, processing_class.pad_token_id, context_length
-                    )
+
+                    if isinstance(reward_model, nn.Module):
+                        _, score, _ = get_reward(
+                            reward_model, postprocessed_query_response, processing_class.pad_token_id, context_length
+                        )
+                    else:
+                        score = torch.tensor(
```
qgallouedec, line 362: need to add `dtype=torch.float` I think

qgallouedec (approved):
Nice! Feel free to merge when the CI is green (ignore dev dep that will fail)

PR #2646, opened 2025-01-24
Add special token to PRM vocabulary if not present

plaguss:
# What does this PR do?

It checks if the `step_separator` token is present in the vocabulary of the model's tokenizer, and adds if it that's not the case. By default it uses `<step_token>` similar to Qwen's PRM from [The Lessons of Developing Process Reward Models in Mathematical Reasoning](https://huggingface.co/papers/2501.07301) `<extra_0>`.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/utils.py
```diff
@@ -1647,3 +1647,22 @@ def flush_left(mask: torch.Tensor, *tensors: torch.Tensor) -> tuple[torch.Tensor
         return mask
     else:
         return mask, *tensors
+
+
+def is_token_in_vocab(tokenizer: PreTrainedTokenizerBase, token: str) -> bool:
+    # Use repr() to handle special characters, and remove the extra quotes
+    # Try direct vocab check
+    if token in tokenizer.get_vocab():
```
qgallouedec, line 1655: does `get_vocab` implied loading a big vocab file? Probably better to do it only once

examples/scripts/prm.py
```diff
@@ -89,6 +90,16 @@
...
+    # Check if the step separator is in the vocabulary, if it's not, add it
+    if not is_token_in_vocab(tokenizer, training_args.step_token):
+        tokenizer.add_special_tokens({"additional_special_tokens": [training_args.step_separator]})
+        model.resize_token_embeddings(
+            len(tokenizer),
+            pad_to_multiple_of=(
+                training_args.resize_to_multiple_of if training_args.resize_to_multiple_of is not None else None
+            ),
+        )
+
```
qgallouedec, line 102: WDYT of having it inside the PRM trainer?

trl/trainer/utils.py
plaguss: it's just building a dict internally, the info is already loaded from a JSON file on initialization, but anyway makes sense to do it only once

examples/scripts/prm.py
plaguss: makes sense, will update the code

PR #2600, opened 2025-01-21
⚡ vLLM for fast generation in GRPO

qgallouedec:
```python
from datasets import load_dataset
from trl import GRPOConfig, GRPOTrainer
import random

def random_reward(completions, **kwargs):
    return [random.random() for _ in completions]


def main():
    # Load the dataset
    dataset = load_dataset("trl-lib/ultrafeedback-prompt", split="train[:5%]")

    training_args = GRPOConfig(
        output_dir="Qwen2-0.5B-GRPO",
        logging_steps=2,
        per_device_train_batch_size=4,
        gradient_accumulation_steps=16,
        max_prompt_length=64,
        max_completion_length=32,
        num_generations=4,
        num_train_epochs=1,
        use_vllm=True,
        vllm_device=2,
    )
    trainer = GRPOTrainer(
        model="Qwen/Qwen2-0.5B-Instruct",
        reward_funcs=random_reward,
        args=training_args,
        train_dataset=dataset,
    )
    trainer.train()

if __name__ == "__main__":
    main()
```

```
accelerate launch --num_processes 2 train_grpo.py
```

PR #2658, opened 2025-01-25
vLLM serve

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/vllm_utils.py
+++ b/trl/vllm_utils.py
@@ -1,0 +1,61 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import io
+
+import requests
+import torch
+
+
+class VLLMClient:
+    """
+    Client for interacting with a VLLM server.
+
+    Args:
+        url (str): The URL of the VLLM server.
+
+    Run the server with:
+    ```bash
+    trl vllm-serve
+    ```
+
+    Example:
+    ```python
+    >>> client = VLLMClient("http://127.0.0.1:5000")
+    >>> data = {"prompts": ["The closest planet to the Sun is", "The capital of France is"]}
+    >>> response = client.generate(data["prompts"])
+    >>> print(response)
+    {'completions': [' Mercury.', ' Paris.']}
+    ```
+    """
+
+    def __init__(self, url):
+        self.url = url
+        self.buffer = io.BytesIO()
+
+    def generate(self, prompts: list[str]) -> dict[str, list[str]]:
+        data = {"prompts": prompts}
+        response = requests.post(self.url + "/generate", json=data)
+        return response.json()
+
+    def chat(self, prompts: list[list[dict[str, str]]]) -> dict[str, list[str]]:
+        data = {"prompts": prompts}
+        response = requests.post(self.url + "/chat", json=data)
+        return response.json()
+
+    def load_weights(self, state_dict) -> None:
+        torch.save(state_dict, self.buffer)
+        self.buffer.seek(0)
+        response = requests.post(self.url + "/load_weights", data=self.buffer.read())
+        return response.json()
--- a/trl/cli.py
+++ b/trl/cli.py
@@ -25,6 +25,7 @@ from .scripts.grpo import make_parser as make_grpo_parser
 from .scripts.kto import make_parser as make_kto_parser
 from .scripts.sft import make_parser as make_sft_parser
 from .scripts.utils import TrlParser
+from .scripts.vllm_serve import vllm_serve
 
 
 def main():
@@ -40,6 +41,7 @@ def main():
     make_grpo_parser(subparsers)
     make_kto_parser(subparsers)
     make_sft_parser(subparsers)
+    subparsers.add_parser("vllm-serve", help="Start a Flask app to serve the VLLM model")
 
     # Parse the arguments
     args = parser.parse_args()
@@ -86,6 +88,9 @@ def main():
         # Feed the args to the launch command
         args.training_script_args = sys.argv[2:]  # remove "trl" and "sft"
         launch_command(args)  # launch training
+
+    elif args.command == "vllm-serve":
+        vllm_serve()
 
 
 if __name__ == "__main__":
--- a/trl/scripts/vllm_serve.py
+++ b/trl/scripts/vllm_serve.py
@@ -1,0 +1,133 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import io
+
+import torch
+from flask import Flask, jsonify, request
+from vllm import LLM
+
+
+def vllm_serve():
+    llm = LLM(model="Qwen/Qwen2.5-0.5B")
+
+    # Create Flask app
+    app = Flask(__name__)
+
+    @app.route("/generate", methods=["POST"])
+    def generate():
+        """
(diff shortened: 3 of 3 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -316,9 +324,43 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+            all_prompt_inputs = self.accelerator.gather(prompt_inputs["input_ids"])
+
+            if self.accelerator.is_main_process:
+                if self.state.global_step != self._last_loaded_step:
+                    self.vllm_client.load_weights(model.state_dict())
+                    self._last_loaded_step = self.state.global_step
+
+                # Get completions from vLLM for all prompts
+                completions_data = self.vllm_client.chat(all_prompts)
+                completions = completions_data["completions"]
```
qgallouedec, line 340: @kashif here we need an if/else, depending on whether the data is conversational. But I'm wondering if we can have this if/else directly in the client side, and use a unique `generation`. In other words:

what we should do here:

```python
if is_conversational:
    completions_data = self.vllm_client.chat(all_prompts)
    ...
else:
   completions_data = self.vllm_client.generate(all_prompts)
   ...
```

OR, modify the client, so that here it's handling both cases smoothly:

```python
completions_data = self.vllm_client.generate(all_prompts)  # would work for both cons and not conv
```

wdyt? IMO the later is the best.

setup.py
```diff
@@ -91,7 +91,7 @@
     "quantization": ["bitsandbytes"],
     "scikit": ["scikit-learn"],
     "test": ["parameterized", "pytest-cov", "pytest-rerunfailures", "pytest-xdist", "pytest"],
-    "vllm": ["vllm; sys_platform != 'win32'"],  # vllm is not available on Windows
+    "vllm": ["flask", "vllm; sys_platform != 'win32'"],  # vllm is not available on Windows
```
lewtun, line 94: Not super important, but I personally prefer [`fastapi`](https://fastapi.tiangolo.com) for HTTP servers since it's both more performant and has lots of nice features like [request validation](https://fastapi.tiangolo.com/tutorial/query-params-str-validations/)

trl/trainer/grpo_trainer.py
```diff
@@ -316,9 +325,35 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+                if self.accelerator.is_main_process:
+                    self.vllm_client.load_weights(self.accelerator.unwrap_model(model).state_dict())
+                self._last_loaded_step = self.state.global_step
+                self.accelerator.wait_for_everyone() # Wait for main process to finish loading weights
+
+            # Set up sampling parameters
+            sampling_params = {"n": self.args.num_generations, "temperature": self.args.temperature}
+
+            # Get completions from vLLM for all prompts
+            with self.accelerator.main_process_first():  # Why it doesn't work without this?
```
lewtun, line 341: Presumably the answer to your question is that without this you get a race condition with the other processes? What kind of error do you get?

trl/trainer/grpo_trainer.py
```diff
@@ -316,9 +325,35 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+            prompt_inputs_repeated = torch.repeat_interleave(prompt_inputs["input_ids"], self.num_generations, dim=0)
+            prompt_completion_ids = torch.cat([prompt_inputs_repeated, completion_ids], dim=1)
+        else:
+            # Regular generation path
+            with unwrap_model_for_generation(model, self.accelerator) as unwrapped_model:
+                prompt_completion_ids = unwrapped_model.generate(
+                    **prompt_inputs, generation_config=self.generation_config
+                )
+
+        # print(f"Generate time: {time.time() - before_generate}")
```
lewtun, line 356:
```suggestion
```

trl/vllm_utils.py
```diff
@@ -0,0 +1,256 @@
...
+if is_flask_available():
+    from flask import Flask, jsonify, request
+
+if is_vllm_available():
+    from vllm import LLM, SamplingParams
+
+
+class VLLMServer:
+    r"""
+    A vLLM server that exposes a REST API for generating completions and chatting with a vLLM model.
```
lewtun, line 35: turbo nit:
```suggestion
    A vLLM server that exposes a REST API for generating completions from a model.
```

setup.py
lewtun: Also, should we put a lower bound on `vllm` since there's lots of important fixes in the latest version

trl/vllm_utils.py
```diff
@@ -0,0 +1,256 @@
...
+                sampling_params = SamplingParams(**data["sampling_params"]) if "sampling_params" in data else None
+
+                # Determine return type (default to 'text')
+                return_type = data.get("return_type", "text")
+                if return_type not in {"text", "tokens"}:
+                    return jsonify({"error": f"Invalid return_type '{return_type}'. Must be 'text' or 'tokens'."}), 400
+
+                # Generate completions
+                if is_conversational({"prompt": prompts[0]}):
+                    outputs = self.llm.chat(prompts, sampling_params=sampling_params)
```
lewtun, line 132: For my understanding, this case will be passing `messages` right?

trl/vllm_utils.py
```diff
@@ -0,0 +1,256 @@
...
+
+            except Exception as e:
+                return jsonify({"status": "error", "message": str(e)}), 400
+
+    def run(self):
+        """Run the Flask server."""
+        self.app.run(host=self.host, port=self.port)
+
+
+class VLLMClient:
```
lewtun, line 173: Did you have to roll your own server / client because the existing [one](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html?ref=blog.mozilla.ai) in `vllm` doesn't support model loading etc?

lewtun (commented):
Great feature! Overall looks really good - ideally it would be nice to have some GPU unit tests of the server / client, but don't worry about that if you want to go fast for now

trl/trainer/grpo_trainer.py
lewtun: An alternative would be to use `apply_chat_template` and pass that directly to `generate()`. It's worth pointing out there's some bugs with how BOS tokens are handled on these endpoints: https://github.com/vllm-project/vllm/issues/9519

trl/trainer/grpo_trainer.py
kashif: so there is no error I believe but it could be that the completions come from an "older" model, so we first get the main processor to update the server's model weights and then use it...

qgallouedec: For context, I run it with two processes.
I don't get an error, here, but for some reason, the one the two processes get an empty list for answer from the server.

lewtun: Can you share the full commands you're running (both the `trl vlm-serve` and the subsequent train command?

qgallouedec:
```si
trl vllm-serve
```

```python
import random

def random_reward(completions, **kwargs):
    return [random.random() for _ in completions]


def main():
    # Load the dataset
    dataset = load_dataset("trl-lib/ultrafeedback-prompt", split="train[:5%]")

    training_args = GRPOConfig(
        output_dir="Qwen2-0.5B-GRPO",
        logging_steps=2,
        gradient_accumulation_steps=16,
        max_prompt_length=64,
        max_completion_length=32,
        num_generations=4,
        num_train_epochs=1,
        use_vllm=True,
        vllm_url="http://26.0.160.225:5123"
    )
    trainer = GRPOTrainer(
        reward_funcs=random_reward,
        args=training_args,
        train_dataset=dataset,
    )
    trainer.train()

if __name__ == "__main__":
    main()
```

```sh
accelerate launch train.py
```

trl/vllm_utils.py
qgallouedec: Exactly

trl/trainer/grpo_trainer.py
qgallouedec: That's a good point! Let's try not to rely on vllm to apply chat template then

trl/vllm_utils.py
qgallouedec: yes, prompts are messages here

SeungyounShin:
Currently, the 1.5B model fits onto a single GPU without issues, but starting with the 7B model, it fails due to OOM errors. This makes it impossible to load the model entirely (for generation).

It seems like a good approach would be to separate GPUs for **generation** (e.g., using `vllm`) and **training** (e.g., using `deepspeed` or `transformers` models).

Would it be okay if I proceed to implement and commit this change?  
For reference, OpenRLHF handles this separation similarly:  
[train/openrlhf/cli/train_ppo_ray.py#L72](https://github.com/hkust-nlp/simpleRL-reason/blob/17fef38a45c9b59cc5d77230edec27f067e0d9dd/train/openrlhf/cli/train_ppo_ray.py#L72)

I believe we need to divide training and generation across different devices. For example:

| **Role**       | **GPU** |
|-----------------|---------|
| Actor       | GPU0    |
| Actor       | GPU1    |
| Reference   | GPU2    |
| Generation  | GPU3    |

This setup ensures a clearer separation of responsibilities across GPUs, minimizing memory issues during operation.

qgallouedec:
closed in favour of #2600, saving/loading weights was too slow

PR #2677, opened 2025-01-28
concatenate_inputs in dpo_trainer improved!

shirinyamani:
# What does this PR do?

This PR is focused on improving the concatanate_inputs method in trl [dpo_trainer](https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py) in a sense that it provides more clarity and simplicity for the purpose of concatenating the prompt+completion (chosen/rejected) which further will be used in forwarding the model. 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@qgallouedec

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -913,6 +913,95 @@ class DPOTrainer(Trainer):
         )
 
         return output
+    
+    @staticmethod
+    def concatenated_inputs(
+        batch: dict[str, Union[list, torch.LongTensor]], 
+        padding_value: int, 
+        is_encoder_decoder: bool = False, 
+        label_pad_token_id: int = -100, 
+        device: Optional[torch.device] = None
+    ) -> dict[str, torch.LongTensor]:
+        """
+        Concatenate the `chosen` and `rejected` inputs from the batch into a single tensor for both the prompt
+        and completion sequences.
+
+        Args:
+            batch (`dict[str, Union[list, torch.LongTensor]]`):
+                A batch of input data. The batch must contain the following keys:
+
+                - `"prompt_input_ids"`: Tensor of shape `(batch_size, prompt_length)` representing the prompt input IDs.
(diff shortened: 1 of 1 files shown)


trl/trainer/dpo_trainer.py
```diff
@@ -877,39 +883,55 @@ def concatenated_inputs(
...
+                new_concatenated_key = k.replace("chosen", "concatenated")
+                output[new_concatenated_key] = pad_to_length(batch[k], max_length, pad_value=pad_value)
+
+            #concat the rejected response;    
+            elif k.startswith("rejected") and isinstance(batch[k], torch.Tensor):
+                if is_encoder_decoder:
+                    pad_value = label_pad_token_id
+                elif k.endswith("_input_ids"):
+                    pad_value = padding_value
+                elif k.endswith("_attention_mask"):
```
qgallouedec, line 914: If it starts by `"rejected"`, and ends by `"attention_mask"`, then it's `"rejected_attention_mask"`, no? Why not explicitly using this key?

shirinyamani: Why should we define a separate key for it? Observing that the third condition `(elif k.endswith("_attention_mask"))` in the ` elif k.startswith("rejected")` loop handles keys that end with `_attention_mask`, which includes `rejected_attention_mask`. 
Because I believe defining an explicit `rejected_attention_mask` is identical to the third `elif k.endswith("_attention_mask")`, because either way (whether defined _explicitly_ OR within the `elif`) the `pad_value` should be `0` when we have `rejected_attention_mask`, WDYT?  

Also we can even go one level more abstract than this version which would be;

```python
            # Handle rejected keys and concatenate them
            elif k.startswith("rejected") and isinstance(batch[k], torch.Tensor):
                pad_value = (
                    label_pad_token_id if is_encoder_decoder else
                    padding_value if k.endswith("_input_ids") else 
                    0
                )
```


But tbh, it's not quite explanatory!

Issue #2684, opened 2025-01-29
Support iterative GRPO


howardzhou:
### Feature request

Hi, 

The [GRPO paper](https://arxiv.org/pdf/2402.03300) also mentioned an interactive version of GRPO which allows for periodic update of reference model (see Algorithm 1). This has shown good performance for cold-start models, see Figure 6. For DeepSeek-R1-Zero, since it's not SFT'ed, they very likely used this iterative version of GRPO. 

Currently, in GRPO trainer, the reference model cannot be updated: https://github.com/huggingface/trl/blob/main/trl/trainer/grpo_trainer.py#L200-L209
```
        # Reference model
        if is_deepspeed_zero3_enabled():
            self.ref_model = AutoModelForCausalLM.from_pretrained(model_id, **model_init_kwargs)
        elif peft_config is None:
            # If PEFT configuration is not provided, create a reference model based on the initial model.
            self.ref_model = create_reference_model(model)
        else:
            # If PEFT is used, the reference model is not needed since the adapter can be disabled
            # to revert to the initial model.
            self.ref_model = None
```

Would be nice to support periodically updating of reference model in Trainer, e.g. after each epoch or certain steps. 

### Motivation

Might be important for reproducing DeepSeek-R1-Zero and DeepSeek-R1

### Your contribution

ideas: 
- may possibly change in `compute_loss` function, but not seems to be an elegant way. 
- may need to overwrite the train function in base trainer and adding that option

qgallouedec:
What about using a callback? That the trainer would internally add.

shirinyamani:
@qgallouedec  Do you mean `callbacks: Optional[list[TrainerCallback]] ` ?

qgallouedec:
Not really. Like an arg in the `GRPOConfig`, let's say `sync_ref_steps`. And having in the init of `GRPO.__init__` something like

```python
if args.sync_ref_steps is not None:
    sync_ref_callback = SyncRefCallback(args.sync_ref_steps)
    self.add_callback(sync_ref_callback)
```

howardzhou:
> What about using a callback? That the trainer would internally add.

Good idea! Something like that?
```
class UpdateRefModelCallback(TrainerCallback):
    def on_epoch_end(self, args, state, control, trainer, **kwargs):
        trainer.ref_model.load_state_dict(trainer.model.state_dict())
```

qgallouedec:
Note that we already have such callback in trl, I think it makes sense to reuse it

[...]

PR #2700, opened 2025-01-30
🔁 🦈 Support iterative GRPO 

shirinyamani:
# What does this PR do?

Following the thread of this [issue#2684](https://github.com/huggingface/trl/issues/2684) and based on [Deepseek](https://arxiv.org/pdf/2402.03300) paper we came to conclude that we need to add a feature which every once in a while (`ref_model_sync_steps`) can iteratively update the reference model. 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@qgallouedec 


tests/test_grpo_trainer.py
```diff
@@ -58,6 +58,9 @@ def test_training(self, config_name):
                 reward_funcs="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
                 args=training_args,
                 train_dataset=dataset,
+                sync_ref_model=True,
+                ref_model_mixup_alpha=0.5,
+                ref_model_sync_steps=1,
```
qgallouedec, line 63:
```suggestion
                ref_model_sync_steps=2,
```

maybe better?

qgallouedec:
Nice! Can you try locally with Multi GPU / DeepSpeed ZeRO 1/2/3? If you don't have the hardware, I can do it.

qgallouedec:
In the DeepSeek-R1 paper, I think they sync the ref after each epoch, no?

shirinyamani:
@qgallouedec 
I do not have access to multi-gpu atm unfortunately! 
I can request access but it might take long time for them to assign gpu to me!

for the update, I thiiiink they do the update after one complete iteration, but I am not sure because I think this way there might be a conflict, because the default `ref_model_sync_steps` is 64 , meaning the update of the `ref_model` will happen after these many steps, but one epoc will be alot more than this probably! 
Maybe I am misunderstanding? 

Screenshot 2025-01-30 at 10 38 58 AM

shirinyamani:
Note that this algorithm and the update discussion is from the [DeepSeekMath](https://arxiv.org/pdf/2402.03300) paper where they discussed the grpo math.

qgallouedec:
Don't bother with multi gpu, I'm go a test myself

I think we understand similarly. I'm wondering what the user would expect. 
This soft update as implemented gives probably better results. But it doesn't match the paper.

Let me make some tests. I'll come back to you.

shirinyamani:
@qgallouedec Did you get to test this by any chance ? 🤔

qgallouedec:
Not yet, will do asap

qgallouedec:
Actually I don't have time to test unfortunately, but I think it's really worth:
1. adding a param `ref_model_sync_epochs` to allow user reproduce precisely the method describe in the paper. Ideally, allow this value to be both int and and float in (0, 1)
2. run some experiments to check if
  -  the current default values make sense
  - does it give significantly different results

Do you want to handle 1. @shirinyamani?

In the meantime I'll merge this one.

tests/test_grpo_trainer.py
```diff
@@ -58,6 +58,9 @@ def test_training(self, config_name):
                 reward_funcs="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
                 args=training_args,
                 train_dataset=dataset,
+                sync_ref_model=True,
+                ref_model_mixup_alpha=0.5,
+                ref_model_sync_steps=2,
```
qgallouedec, line 63:
```suggestion
```

shirinyamani:
> Actually I don't have time to test unfortunately, but I think it's really worth:
> 
> 1. adding a param `ref_model_sync_epochs` to allow user reproduce precisely the method describe in the paper. Ideally, allow this value to be both int and and float in (0, 1)

> Do you want to handle 1. @shirinyamani?

Sure, for brainstorming purposes let's break down our options; 

**Goal:** is to have a param like  `0 < ref_model_sync_epochs < 1` that allow user to once in a while after X number of epochs (can be 0.2 of epoch) to update the ref_model ?

**How to build:**

**Option 1**): to override the current `on_step_end` method in the `SyncRefModelCallback` class to reflect what we want; 
now it is like this; 

```python 

class SyncRefModelCallback(TrainerCallback):
    def __init__(
        self,
        ref_model: Union[PreTrainedModel, torch.nn.Module],
        accelerator: Optional[Accelerator],
    ):
        self.accelerator = accelerator
        self.ref_model = ref_model

    @staticmethod
    def _sync_target_model(model, target_model, alpha):
        for target_param, copy_param in zip(target_model.parameters(), model.parameters()):
            target_param.data.mul_(1.0 - alpha).add_(copy_param.data, alpha=alpha)

    @staticmethod
    def sync_target_model(model, target_model, alpha):
        deepspeed_plugin = AcceleratorState().deepspeed_plugin
        if deepspeed_plugin is not None and deepspeed_plugin.zero_stage == 3:
            with deepspeed.zero.GatheredParameters(
                list(model.parameters()) + list(target_model.parameters()), modifier_rank=0
            ):
                if deepspeed.comm.get_rank() == 0:
                    SyncRefModelCallback._sync_target_model(model, target_model, alpha)
        else:
            SyncRefModelCallback._sync_target_model(model, target_model, alpha)

    def on_step_end(self, args, state, control, **kwargs):
        model: PreTrainedModel = kwargs["model"]

        if self.ref_model is not None and state.global_step % args.ref_model_sync_steps == 0:
            if self.accelerator:
                model = self.accelerator.unwrap_model(model)
            self.sync_target_model(model, self.ref_model, args.ref_model_mixup_alpha)

```

with the changes would be sth like;

```python
    def on_step_end(self, args, state, control, **kwargs):
        model: PreTrainedModel = kwargs["model"]

        # Calculate total steps per epoch
        steps_per_epoch = state.max_steps // args.num_train_epochs

        # Determine if we should sync based on ref_model_sync_epochs
        if isinstance(self.ref_model_sync_epochs, int):
            # Sync based on integer number of epochs
            should_sync = state.global_step % (self.ref_model_sync_epochs * steps_per_epoch) == 0
        elif isinstance(self.ref_model_sync_epochs, float):
            # Sync based on fraction of total epochs
            should_sync = (state.global_step / steps_per_epoch) % (self.ref_model_sync_epochs * args.num_train_epochs) == 0
        else:
            raise ValueError("ref_model_sync_epochs must be an int or a float")

        if self.ref_model is not None and should_sync:
            if self.accelerator:
                model = self.accelerator.unwrap_model(model)
            self.sync_target_model(model, self.ref_model, args.ref_model_mixup_alpha)
```

This might work if my understanding of `on_step_end` is correct! Since there was no doc string, my understanding from this method is that its like a check if the reference model should be synchronized based on ref_model_sync_steps and performs the synchronization if necessary.


**Option 2:)** is to add the `ref_model_sync_epochs` locally to the grpo loss and backward update ?

shirinyamani:
One more Question for you; @qgallouedec 

If the ref_model is getting updated so frequently, would it be same as not having ref_model at all ?

PR #2728, opened 2025-02-01
↔️  GRPO: Set max_model_len when initializing vLLM instance

mirceapricop:
# What does this PR do?

By default, the vLLM model will be set up to support the max context of the input model.

However, during training we know we will only observe at most max_prompt_length + max_completion_length tokens, so we can use that to have a reduced memory footprint.

This is especially relevant when running on limited hardware as the improvement in memory can be significant.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -131,29 +131,25 @@ class GRPOConfig(TrainingArguments):
             "(`pip install vllm`)."
         },
     )
+    vllm_init_kwargs: Optional[dict] = field(
+        default_factory=lambda: {
+            "device": "auto",
+            "gpu_memory_utilization": 0.9,
+        },
+        metadata={
+            "help": "Keyword arguments for `vllm.LLM.__init__` when `use_vllm` is true"
+        },
+    ) 
     vllm_device: Optional[str] = field(
         default="auto",
         metadata={
-            "help": "Device where vLLM generation will run, e.g. 'cuda:1'. If set to 'auto' (default), the system "
-            "will automatically select the next available GPU after the last one used for training. This assumes "
-            "that training has not already occupied all available GPUs."
+            "help": "Deprecated. Set `device` in `vllm_init_kwargs` instead."
         },
     )
     vllm_gpu_memory_utilization: float = field(
         default=0.9,
         metadata={
-            "help": "Ratio (between 0 and 1) of GPU memory to reserve for the model weights, activations, and KV "
-            "cache on the device dedicated to generation powered by vLLM. Higher values will increase the KV cache "
-            "size and thus improve the model's throughput. However, if the value is too high, it may cause "
-            "out-of-memory (OOM) errors during initialization."
-        },
-    )
-    vllm_max_model_len: Optional[int] = field(
-        default=None,
-        metadata={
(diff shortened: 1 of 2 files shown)


qgallouedec:
We could do that, but memory limitation shouldn't come from generation. Or maybe are you using the same device from training and generation?

mirceapricop:
Indeed, my use case is specifically running on a single consumer GPU. It might be wishful thinking, but with this patch I am able to run a training loop for a 1.5B model.

qgallouedec:
Have you tried to reduce the vllm gpu memory usage?

mirceapricop:
Yes, in fact that's what led me to this change. Lowering it reduces the space left for a KV cache, and vLLM  prints:

```
The model has a long context length (131072). This may cause OOM errors during the initial memory profiling phase, or result in low performance due to small KV cache space. Consider setting --max-model-len to a smaller value.
```

So my understanding is that with this, a smaller KV cache is more efficiently utilized.

Are there any downsides to setting this? We could make it opt-in through an arg if you think it can have negative implications.

qgallouedec:
Now it makes sense. Can you add an arg in the config instead?

qgallouedec:
`vllm_max_model_len`

mirceapricop:
Works for me, done.

qgallouedec (approved):
Perfect

qunash:
I think it would be cleaner to add a `vllm_init_kwargs` dict that would be passed directly to the vLLM engine initialization. It would replace the individual `vllm_device` and `vllm_gpu_memory_utilization` parameters with a more flexible approach that:

1. Makes it easier to support all vLLM engine parameters without adding new fields to the config
2. Follows the same pattern as `model_init_kwargs` that's already in the codebase

```Python
vllm_init_kwargs: Optional[dict] = field(
        default_factory=lambda: {
            "device": "auto",
            "gpu_memory_utilization": 0.9,
        },
        metadata={
        },
    )
```

mirceapricop:
That sounds good to me, but I'm not sure if it can be done in a backwards compatible manner (or if it's acceptable to make backwards-incompatible flag changes). @qgallouedec thoughts?

sroecker:
> I think it would be cleaner to add a `vllm_init_kwargs` dict that would be passed directly to the vLLM engine initialization.

This would be the best solution for single GPU (poor) training as people might want to tune other parameters as well as seen in this thread: https://x.com/robertshaw21/status/1885781591961571455

mirceapricop:
Now updated with `vllm_init_kwargs` and marking previous vllm init args as deprecated (though still using them to stay backward compatible)

qgallouedec:
In fact, I'm in favor of explicitly stating the parameters for two main reasons:

-  not doing so means implicitly assuming that any combination of parameters is compatible with GRPO, which is not the case
- although less flexible, it's easier for the user to know which arguments are available, without having to refer to the vLLM doc

As for backwards compatibility, GRPO is a new trainer and the lib is still in alpha, so there's no real need to ensure that.

This can be discussed again in the future if this lack of flexibility is really a problem. But adding parameters one-by-one should be good for now

mirceapricop:
Now reverted to just adding a single new arg

qgallouedec (approved):
Thanks, merging when CI is green :)

Issue #2608, opened 2025-01-22
GRPO questions


natolambert:
Hey friends! I have some questions on the GRPO implementation, happy to discuss.

1. It looks like you apply the KL distance in the advantages, while the DeepSeekMath paper says “Also note that, instead of adding KLpenalty in the reward, GRPO regularizes by directly adding the KL divergence between the trained policy and the reference policy to the loss, avoiding complicating the calculation of 𝐴ˆ”
2. Did any thought go into making this a sum of loss and not mean? We aren’t sure https://github.com/huggingface/trl/blob/fe4b5efe4e23f4331ba9c5b0c8bd92dc8302c287/trl/trainer/grpo_trainer.py#L286
3. I didn’t see the PPO clipping logic in policy gradient loss, coming soon?

qgallouedec:
Hey @natolambert!!


> It looks like you apply the KL distance in the advantages, while the DeepSeekMath paper says “Also note that, instead of adding KLpenalty in the reward, GRPO regularizes by directly adding the KL divergence between the trained policy and the reference policy to the loss, avoiding complicating the calculation of 𝐴ˆ”

Here's how I understand it:
In PPO, the KL div is subtracted from the reward (_adding KL penalty in the reward_)

https://github.com/huggingface/trl/blob/949db2357e62d2f0a34decfc5e87eeeea0c6d72c/trl/trainer/ppo_trainer.py#L509-L515

Here, in GRPO, the advantage is computed without the KL term. It's just the output of the reward function, normalised per group:

https://github.com/huggingface/trl/blob/949db2357e62d2f0a34decfc5e87eeeea0c6d72c/trl/trainer/grpo_trainer.py#L274-L281

and later you subtract the KL term

https://github.com/huggingface/trl/blob/949db2357e62d2f0a34decfc5e87eeeea0c6d72c/trl/trainer/grpo_trainer.py#L285

I don't know how complicated it would have been to integrate KL into the reward. It would probably look something like

```python
# Subtract KL
rewards  = rewards - self.beta * per_token_kl  # <- THIS IS ADDED

# Compute grouped-wise rewards
mean_grouped_rewards = rewards.view(-1, self.num_generations).mean(dim=1)
std_grouped_rewards = rewards.view(-1, self.num_generations).std(dim=1)

# Normalize the rewards to compute the advantages
mean_grouped_rewards = mean_grouped_rewards.repeat_interleave(self.num_generations, dim=0)
[...]

PR #2739, opened 2025-02-02
feat: Add cliprange to GRPO loss

joey00072:
Updates the loss calculation to use a clipped version of the policy ratio, similar to PPO's clipping mechanism.

# What does this PR do?
Adds clipping objective  to GRPO loss
Screenshot 2025-02-02 at 6 05 26 PM
https://arxiv.org/pdf/2402.03300 eq 3 page 12
 While the current loss carries the spirit of the ppo trust region objective, it is not the same as described in the paper.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec @lewtun

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -109,6 +109,12 @@ class GRPOConfig(TrainingArguments):
         default=8,
         metadata={"help": "Number of generations to sample."},
     )
+    
+    cliprange: float = field(
+        default=0.2,
+        metadata={"help": "Clip range."},
+    )
+    
     temperature: Optional[float] = field(
         default=0.9,
         metadata={"help": "Temperature for sampling. The higher the temperature, the more random the completions."},
@@ -144,7 +150,7 @@ class GRPOConfig(TrainingArguments):
             "out-of-memory (OOM) errors during initialization."
         },
     )
-
+    
     # Parameters that control the training
     learning_rate: float = field(
         default=1e-6,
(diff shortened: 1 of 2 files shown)


qgallouedec:
We are only doing one update so there is no need to clip

joey00072:
https://x.com/shxf0072/status/1886407737023865012

I was trying to run in gpu limited env,
where we do some 128 generation and 128 train step,

its become wrong then

ppl look up hf for reference implementation, 
i don't think making this assumption are good

anyways can we atleast add this link https://github.com/huggingface/trl/issues/2608
in comment above loss

qgallouedec:
It's not wrong since you can't do multiple optimization steps with the current implementation.
We may allow this in the future, and if so, we'll add the clipping.
Feel free to suggest a comment if you feel that clarity is missing here. Thanks!

winglian:
@qgallouedec Would you consider moving the per_token_loss logic to it's own method? would at least make it easier to extend for those wanting to do so without having to duplicate the entire compute_loss block.

tyler-romero:
+1 for allowing multiple optimization steps per rollout, similar to how the Deepseekmath paper describes GRPO and similar to how PPO is implemented in practice.

qgallouedec:
> @qgallouedec Would you consider moving the per_token_loss logic to it's own method? would at least make it easier to extend for those wanting to do so without having to duplicate the entire compute_loss block.

Yes, done in https://github.com/huggingface/trl/pull/2762

aiyinyuedejustin:
page 13 bro

joey00072:
closing this https://github.com/huggingface/trl/pull/2899/,
thanks @qgallouedec 
(sorry forgot about this😅)

PR #2743, opened 2025-02-02
Raise UserWarning in RewardTraining if PEFT target_modules="all-linear"

JohnGiorgi:
# What does this PR do?

I ran into a nasty edge case that I don't think is documented anywhere when using the `RewardTrainer` to train a reward model. If a `peft_config` is provided and `target_modules="all-linear"` ([a wild card that means: adapt all linear layers _except_ for the output](https://huggingface.co/docs/peft/main/en/package_reference/lora#peft.LoraConfig.target_modules)) then the output layer of the reward model, which is often newly initialized and is used to score the chosen/rejected completions, will go un-adapted (and therefore un-trained). Performance will be impacted as you might expect, e.g. here's two runs of mine with `target_modules="all-linear"` and `target_modules=None` (the default):

This PR simply raises a `UserWarning` in `RewardModel` in this case. You could almost argue that raising an error would be warranted, but I wonder if there is some scenario in which the output layer is already trained, and a user wants to just adapt some intermediate layers with LoRA.

## Before submitting

      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/reward_trainer.py
+++ b/trl/trainer/reward_trainer.py
@@ -25,6 +25,7 @@ import torch.nn as nn
 from accelerate import PartialState
 from accelerate.utils import gather_object
 from datasets import Dataset
+from peft.utils import INCLUDE_LINEAR_LAYERS_SHORTHAND
 from transformers import (
     BaseImageProcessor,
     DataCollator,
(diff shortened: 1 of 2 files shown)


trl/trainer/reward_trainer.py
```diff
@@ -159,6 +160,19 @@ def __init__(
 
                     model = prepare_model_for_kbit_training(model, **prepare_model_kwargs)
 
+                # Warn if the user passes "all-linear" for the target_modules
+                target_modules = (
+                    peft_config.get("target_modules", None)
+                    if isinstance(peft_config, dict)
+                    else peft_config.target_modules
+                )
```
JohnGiorgi, line 168: I am handling both cases of `peft_config` being `dict` and being a `PeftConfig`object. 

It is type-hinted as a `dict` in the `RewardModel` docstring, but I can see a `PeftConfig` object is passed to `RewardTrainer` [here](https://github.com/huggingface/trl/blob/bbdd6db17c49db813695d0a8bc0da7bf6b1bb88e/examples/scripts/reward_modeling.py#L119).

JohnGiorgi:
Closing because I wasn't able to reproduce this on closer inspection... I can also see `model.score.parameters()` is updated after training even with `target_modules=all-linear`...

PR #2708, opened 2025-01-31
⚠️ Fix Attention Masking in GRPO

andyl98:
# What does this PR do?

Fix a small yet dangerous issue -- not adding `attention_mask` when computing logits.

Will add more notes.

## Details
```python
from transformers import AutoTokenizer, AutoModelForCausalLM, GenerationConfig

model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen2.5-Coder-0.5B", device_map="auto")
tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen2.5-Coder-0.5B")

prompts = [
    "hello, 1 + 1 = ?",
    "hello, 12781297598217521580 + 120481209481290482 = ?",
]

prompt_inputs = tokenizer(
    prompts, return_tensors="pt", padding=True, padding_side="left", add_special_tokens=False
).to(model.device)

generation_config = GenerationConfig(
    max_new_tokens=100,
    do_sample=True,
    temperature=0.2,
    num_return_sequences=4,
    pad_token_id=tokenizer.pad_token_id,
)

prompt_completion_ids = model.generate(
    **prompt_inputs, generation_config=generation_config
)

prompt_length = prompt_inputs["input_ids"].size(1) #45
completion_ids = prompt_completion_ids[:, prompt_length:]
num_logits_to_keep = completion_ids.size(1) #100
```

```python
> input_ids

{'input_ids': tensor([[151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643,
         151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643,
         151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643,
         151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643,
          14990,     11,    220,     16,    488,    220,     16,    284,    937],
        [ 14990,     11,    220,     16,     17,     22,     23,     16,     17,
             24,     22,     20,     24,     23,     17,     16,     22,     20,
             17,     16,     20,     23,     15,    488,    220,     16,     17,
             15,     19,     23,     16,     17,     15,     24,     19,     23,
             16,     17,     24,     15,     19,     23,     17,    284,    937]],
       device='mps:0'), 'attention_mask': tensor([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
[...]

PR #2747, opened 2025-02-03
Fix device placement for GRPO attention mask in compute_loss

tgaddair:
Follow-up to #2708. Seems the attention_mask in the prompt_inputs can be CPU even when the rest of training is on GPU.

cc @qgallouedec


ctjlewis:
good catch, cc @qgallouedec

qgallouedec:
Surprising, since the device placement should be handled by 

```python
prompt_inputs = super()._prepare_inputs(prompt_inputs)
```

can you provide a code to reproduce?

PR #2749, opened 2025-02-03
Diff from upstream of #2747 [do not merge]

whitead:
Just putting this on a branch so we can pull it in. We can close and pull from upstream once they merge


whitead:
wrong upstream! Sorry

PR #2763, opened 2025-02-04
Remote GRPO ref model

edbeeching:
Adds an option to use a remote reference model, hosted on another node. The user can provide the url.
I was originally going to use vllm serve but you would have to decode the ids and encode etc. This seems simpler, but may not be as robust. Speed is not an issue though as it is just a forward pass and not autoregressive generation.
Usage:
```shell
# launch the app on on node
python trl/models/remote_model_app.py --model_name deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B

# then provide the url to the training script
python trl/models/remote_model_app.py ref_model_url: http://1.2.3.4:8000
```
Still to do:
- Multi-GPU training, as the calls to the remote model may conflict with each other
- Large batches
- add some better error handling / restarts
- current implemention assumes the model + activations fit on 1 GPU, which is fine for up to 14B but we will need a different approach for larger models.
- Create a PR on the open-r1 repo with slurm script and recipe

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -78,7 +78,7 @@ class GRPOConfig(TrainingArguments):
             Number of updates steps to accumulate the gradients for, before performing a backward/update pass.
         beta (`float`, *optional*, defaults to `0.04`):
             KL coefficient.
-        
+
         > Parameters that control remote models
         ref_model_url: str
     """
@@ -178,4 +178,6 @@ class GRPOConfig(TrainingArguments):
         metadata={"help": "KL coefficient."},
     )
 
-    ref_model_url: Optional[str] = field(default=None, metadata={"help": "URL of the reference model, if you are using a TRL RemoteModel"})
+    ref_model_url: Optional[str] = field(
+        default=None, metadata={"help": "URL of the reference model, if you are using a TRL RemoteModel"}
+    )
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -78,6 +78,9 @@ class GRPOConfig(TrainingArguments):
             Number of updates steps to accumulate the gradients for, before performing a backward/update pass.
         beta (`float`, *optional*, defaults to `0.04`):
             KL coefficient.
+        
+        > Parameters that control remote models
+        ref_model_url: str
     """
 
     # Parameters that control the model and reference model
@@ -174,3 +177,5 @@ class GRPOConfig(TrainingArguments):
         default=0.04,
         metadata={"help": "KL coefficient."},
     )
+
+    ref_model_url: Optional[str] = field(default=None, metadata={"help": "URL of the reference model, if you are using a TRL RemoteModel"})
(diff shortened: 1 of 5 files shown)


shirinyamani:
if the ref_model exists on another machine (node), I don't fully understand how fsdp can take place ? Just to improve my knowledge I appreciate explanation on how the sync can happen without conflict ?

Superskyyy:
This would potentially conflict with PR #2684 though, maybe need a note on doc.

edbeeching:
@shirinyamani 
The ref model is fixed, so there is no need to sync weights to it. It could be wrapped with FSDP, although this implementation does not expose that option, we assume it is small enough to fit on one GPU. 

Only the model being optimized is sharded in this setting, the ref model is running on another node in order to free memory on the node being used for optimization.

@Superskyyy , good point. Yes I do not think that iterative GRPO is compatible with this option.

trl/trainer/grpo_config.py
```diff
@@ -78,6 +78,9 @@ class GRPOConfig(TrainingArguments):
             Number of updates steps to accumulate the gradients for, before performing a backward/update pass.
         beta (`float`, *optional*, defaults to `0.04`):
             KL coefficient.
+
+        > Parameters that control remote models
+        ref_model_url: str
```
qgallouedec, line 83:
```suggestion
        ref_model_url: str
            .... Using a remote ref model isn't compatible with ref model syncing.
```

Superskyyy:
> @shirinyamani 
> 
> The ref model is fixed, so there is no need to sync weights to it. It could be wrapped with FSDP, although this implementation does not expose that option, we assume it is small enough to fit on one GPU. 
> 
> 
> 
> Only the model being optimized is sharded in this setting, the ref model is running on another node in order to free memory on the node being used for optimization.
> 
> 
> 
> @Superskyyy , good point. Yes I do not think that iterative GRPO is compatible with this option. 
> 
> 

When a more distributed backend is built into the lib it can be solved naturally.

edbeeching:
@Superskyyy 
Just pasting a comment from our internal slack below, I should be able to include interactive GRPO in this PR using a remove SDLang endpoint:

> I am going refactor it to work with an SDLang endpoint. The reasoning is that SDLang endpoints have two nice features, [working directly with the token_ids](https://docs.sglang.ai/backend/native_api.html#Skip-Tokenizer-and-Detokenizer) and the capacity to [reload a new model weights from disk](https://docs.sglang.ai/backend/native_api.html#Update-Weights-From-Disk).
> The weight reload is useful in two settings, iterative GRPO, which is already available in TRL by this [PR](https://github.com/huggingface/trl/pull/2700), and also async GRPO, which may be required in the future to scale up the training.

Superskyyy:
> @Superskyyy Just pasting a comment from our internal slack below, I should be able to include interactive GRPO in this PR using a remote SDLang endpoint:
> 
> > I am going refactor it to work with an SDLang endpoint. The reasoning is that SDLang endpoints have two nice features, [working directly with the token_ids](https://docs.sglang.ai/backend/native_api.html#Skip-Tokenizer-and-Detokenizer) and the capacity to [reload a new model weights from disk](https://docs.sglang.ai/backend/native_api.html#Update-Weights-From-Disk).
> > The weight reload is useful in two settings, iterative GRPO, which is already available in TRL by this [PR](https://github.com/huggingface/trl/pull/2700), and also async GRPO, which may be required in the future to scale up the training.

Thanks! I'm planning some further decoupling and efficiency gains, once this is merged I will try to add something on top of it.

qgallouedec:
Closed because no recent updates, to be reopened if we want to push further in this direction.

edbeeching:
> Closed because no recent updates, to be reopened if we want to push further in this direction.

Yes, I think given that we are mostly running with `beta=0`, this optimization is not that useful.

PR #2772, opened 2025-02-05
📠 Log completions for GRPO

qgallouedec:
@lewtun 
Screenshot 2025-02-05 at 15 46 14


trl/trainer/grpo_trainer.py
```diff
@@ -536,6 +537,15 @@ def _prepare_inputs(self, inputs: dict[str, Union[torch.Tensor, Any]]) -> dict[s
...
         self._metrics["reward_std"].append(self.accelerator.gather_for_metrics(std_grouped_rewards).mean().item())
 
+        if (
+            self.accelerator.is_main_process
+            and self.log_completions
+            and self.state.global_step % self.args.logging_steps == 0
+        ):
+            for prompt, completion, reward in zip(prompts, completions, rewards):
+                text = f"---\n\033[31m{prompt}\033[34m{completion}\033[0m\n--- \033[32mReward: {reward}\033[0m ---"
+                self.accelerator.print(text)
```
lewtun, line 547: Does this get logged to WandB? What I had in mind was something like Costa did for PPO here: https://github.com/huggingface/trl/blob/af4ad47035529164799be10f3fe558ee642a9880/trl/trainer/ppo_trainer.py#L733

Since GRPO produces N completions per prompt, a variant would be to explode the DataFrame column-wise, so that one has columns like ``(prompt, completion_0, reward_0, completion_1, reward_1, ..., completion_{N-1}, reward_{N-1}`

WDYT?

qgallouedec: Working with W&B tables is frustrating. They often don’t update properly (see [this](https://github.com/wandb/wandb/issues/2981)), generate excessive artifacts, and make it unclear whether step visualization is possible. The only somewhat reliable approach seems to be logging the entire table once at the end of training, as shown [here](https://docs.wandb.ai/tutorials/tables/#2-run-training-and-log-test-predictions), but that’s not very useful in practice. And wandb Provide very little support. That's why I avoid it usually.

I'll push this solution, you'll tell me what you think.

qgallouedec:
This is what I get:

https://wandb.ai/huggingface/huggingface/runs/eipsql4s

Here the table has 24 rows, when it's suppose to have (50 steps / 2 logging steps * 8 completions) = 200. 🤔

lewtun:
> Here the table has 24 rows, when it's suppose to have (50 steps / 2 logging steps * 8 completions) = 200. 🤔

Let me test it out - might be a DDP issue

lewtun:
Here is a run that looks correct to me: https://wandb.ai/huggingface/huggingface/runs/y377jshc?nw=nwuserlewtun

Run on 3+1 GPUs with the code below

```python
import random
"""Usage (on 8 x H100s):

pip install vllm==0.7.0 --extra-index-url https://download.pytorch.org/whl/cu121
pip install -e '.[dev]'

# DDP
accelerate launch --config_file examples/accelerate_configs/multi_gpu.yaml --num_processes 3 scratch/grpo_demo.py

# ZeRO-2
accelerate launch --config_file examples/accelerate_configs/deepspeed_zero2.yaml --num_processes 7 scratch/grpo_demo.py

# ZeRO-3
accelerate launch --config_file examples/accelerate_configs/deepspeed_zero3.yaml --num_processes 7 scratch/grpo_demo.py

# FSDP
accelerate launch --config_file examples/accelerate_configs/fsdp.yaml --num_processes 7 scratch/grpo_demo.py

"""

def random_reward(completions, **kwargs):
    return [random.random() for _ in completions]


def main():
    # Load the dataset
    dataset = load_dataset("trl-lib/ultrafeedback-prompt", split="train[:5%]")

    training_args = GRPOConfig(
        output_dir="Qwen2-0.5B-GRPO",
        gradient_checkpointing=False,
        max_prompt_length=64,
        max_completion_length=32,
        num_generations=3,
        num_train_epochs=1,
        use_vllm=True,
        vllm_device="auto",
        vllm_gpu_memory_utilization=0.7,
        bf16=True,
        log_completions=True,
        logging_strategy="steps",
        logging_steps=2,
        report_to="wandb",
        max_steps=4
    )
    trainer = GRPOTrainer(
        reward_funcs=random_reward,
        args=training_args,
        train_dataset=dataset,
    )
    trainer.train()

if __name__ == "__main__":
    main()
```

trl/trainer/grpo_trainer.py
```diff
@@ -535,6 +536,13 @@ def _prepare_inputs(self, inputs: dict[str, Union[torch.Tensor, Any]]) -> dict[s
 
         # Decode the generated completions
         completions = self.processing_class.batch_decode(completion_ids, skip_special_tokens=True)
+
+        # For logging
+        table = defaultdict(list)
+        table["step"] = [str(self.state.global_step)] * len(gather_object(prompts_text))
```
lewtun, line 542: Useful for knowing which completions belong to which step

qgallouedec:
Looks good!

PR #2777, opened 2025-02-06
fix reward function in GRPO example

### LINKED ISSUE #2771: Question about reward function in GRPO example
https://huggingface.co/docs/trl/main/en/grpo_trainer

Am I misunderstanding something about the reward function in the documentation example? 

Looking at the current example:
```python
    return [abs(20 - len(completion)) for completion in completions]
```

I think this reward function might be working opposite to what we want. Since GRPO tries to maximize the reward, wouldn't this make the model generate text that's far from 20 characters? 

Shouldn't we use something like:
```python
```

or 
```python
    return [1 / (1 + abs(20 - len(completion))) for completion in completions]
```

so that completions closer to 20 characters get higher rewards?

Would appreciate if someone could clarify this

junuMoon:
# What does this PR do?

Fixes #2771 

Change reward function to return normalized values between 0 and 1, where completions closer to target length (20) get higher rewards.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec

### DIFF
--- a/docs/source/grpo_trainer.md
+++ b/docs/source/grpo_trainer.md
@@ -36,7 +36,7 @@ dataset = load_dataset("trl-lib/tldr", split="train")
 
 # Define the reward function, which rewards completions that are close to 20 characters
 def reward_len(completions, **kwargs):
-    return [abs(20 - len(completion)) for completion in completions]
+    return [1 / (1 + abs(20 - len(completion))) for completion in completions]
 
 training_args = GRPOConfig(output_dir="Qwen2-0.5B-GRPO", logging_steps=10)
 trainer = GRPOTrainer(


docs/source/grpo_trainer.md
```diff
@@ -36,7 +36,7 @@ dataset = load_dataset("trl-lib/tldr", split="train")
 
 # Define the reward function, which rewards completions that are close to 20 characters
 def reward_len(completions, **kwargs):
-    return [abs(20 - len(completion)) for completion in completions]
+    return [1 / (1 + abs(20 - len(completion))) for completion in completions]
```
qgallouedec, line 39: Maybe just -abs ? Simples to understand

qgallouedec (approved):

Issue #2809, opened 2025-02-09
[Question] Proper data format for GRPO Agent Training


August-murr:
At first I wanted to use smolagents Agents to train with GRPO but then I realized the enormous amounts of prompts used to guide the Agent aren't useful for training an agent and will complicate everything since the Agent learns to do everything on it's own. 

so instead I want to proceed with a simple format like the 
```json
[
  {
    "role": "user",
    "content": "Can you help me calculate the factorial of 5 ?"
  },
  {
    "role": "assistant",
    "content": "Sure, I'll write a recursive function for that."
  },
  {
    "role": "assistant",
    "content": "Here's the code:"
  },
  {
    "role": "tool",
    "content": "```python\n# Recursive factorial function\n\ndef factorial(n):\n    if n <= 1:\n        return 1\n    return n * factorial(n-1)\n\nprint(factorial(5))\n```"
  },
  {
    "role": "tool",
    "content": "Output: 120"
  },
  {
    "role": "assistant",
    "content": "The code runs correctly and returns 120. Would you like me to explain how it works?"
  }
]
```
and this raises some questions:

1. Would this multi-turn format (including multiple tool messages—one for code and one for its output) work with the current GRPO implementation?

2. Is there a recommended or “correct” format for such agentic data that includes tool usage for GRPO?
 
3. Are there any special considerations for the tool responses? For instance, should we exclude those tokens from KL divergence computations (as @xiangjjj mentioned in #2723 ) or handle them differently with attention masks?

qgallouedec:
> 1. Would this multi-turn format (including multiple tool messages—one for code and one for its output) work with the current GRPO implementation?

I think it _should_ work.

> 2. Is there a recommended or “correct” format for such agentic data that includes tool usage for GRPO?

Not that I am aware of. As long as the format is supported by the chat template of the model.

> 3. Are there any special considerations for the tool responses? For instance, should we exclude those tokens from KL divergence computations (as [@xiangjjj](https://github.com/xiangjjj) mentioned in [Training Agents with GRPO #2723](https://github.com/huggingface/trl/issues/2723) ) or handle them differently with attention masks?

My hunch is that, indeed, these tokens should be masked for the loss calculation (not just the KL part, also the advantage part). But that it's probably OK not to mask them. My feeling is that even if these tokens won't be generated by the model, it may still help to train the model to predict them, or at least it won't hurt the learning process.

qgallouedec:
[...]

Issue #2723, opened 2025-01-31
[Project] Training Agents with GRPO


August-murr:
Let's discuss how to train agents using GRPO. 

Here, I will link sub-issues related to various problems, features, or questions that need resolution for implementing this idea.

accupham:
So hear me out. I think the current style of function/tool calling as popularized by OpenAI does not lend itself to the expressiveness and fluidity of thought that R1-like model capitalize and excel at. The feedback loop is too long.

Instead, I think the best way is to express function calls as mixed code and language. Let the user decide the calling convention, and inject the results inline.

Here's a example of what I mean. This is QwQ hooked up directly to a python interpreter to do neurosymbolic reasoning. Every line entered is directly fed into the REPL, and results fed back immediately to the LLM.

```
# Okay, let's tackle this step by step. The problem says:
# If the equations x^2 - a = p * y and x^2 - b = p * y have integer solutions,
# then x^2 - ab = p * y also has integer solutions. p is prime, and a, b are not divisible by p.

from sympy import symbols, Eq, solveset, S

# Let's define the symbols we need. I'll start with x, y, p, a, and b.
x, y, p, a, b = symbols('x y p a b', integer=True)

# Alright, I'll write out the first two equations.
eq1 = Eq(x**2 - a, p * y)  # This is x^2 - a = p * y
eq2 = Eq(x**2 - b, p * y)  # This is x^2 - b = p * y

# The claim is that if these two equations have solutions,
# then x^2 - ab = p * y should also have a solution. I'll write that equation too.
eq3 = Eq(x**2 - a * b, p * y)  # This is x^2 - ab = p * y

# Okay, let's check if we can find integer solutions for the first two equations.
# I'm going to use `solveset` to see if x has solutions in the integers for eq1 and eq2.
sol_x1 = solveset(eq1, x, domain=S.Integers)  # Solve eq1 for x
sol_x1  # Let's see what we get here.
# >>> ConditionSet(x, Eq(x**2 - a, p*y), Integers)

# Hmm, that's a generic solution for x in terms of the parameters.
# Makes sense because we don't have concrete values for p, a, or y yet.

sol_x2 = solveset(eq2, x, domain=S.Integers)  # Solve eq2 for x
sol_x2  # Alright, let me look at this one too.
# >>> ConditionSet(x, Eq(x**2 - b, p*y), Integers)

# Okay, so I have the solution sets for x in eq1 and eq2.
[...]

PR #2810, opened 2025-02-09
GRPO Environments for custom multi-step rollouts (vLLM-only)

willccbb:
# What does this PR do?

Adds a protocol under `trl/environments` for an `Environment` object which wraps vLLM's `.generate(...)` to allow for custom rollout logic, and an optional `env` field to the Trainer for passing such an object. 

A simple example usage is included below, others in this repo: [willccbb/verifiers](https://github.com/willccbb/verifiers)

Given the breadth of different agentic tasks people are interested in, I think an implementation of multi-step behavior should be as open-ended and customizable as possible, rather than having everything flow through explicit tool use or a predefined format. Here, the only requirement for an Environment is that it mirrors the behavior of calling `llm.generate()` + extracting `token_ids`. I have found that it's more practical to pass message dicts to Environments rather than preformatted text, hence the addition of `gather_object(prompts)`.  

I agree with Quentin's comment [here](https://github.com/huggingface/trl/issues/2809#issuecomment-2646323768) that while masking of tool outputs/environment responses/messages from other users/agents is maybe desirable in some cases, it is probably not necessary for initial experimentation. Future iterations could perhaps extend the `Environment` definition to allow for mask outputs.

Usage looks like:
```python
doublecheck_env = DoubleCheckEnv() # assistant message -> hardcoded user "Are you sure?" message -> second assistant message
...
trainer = GRPOTrainer(
    model=model_name,
    processing_class=tokenizer,
    reward_funcs=reward_funcs,
    env=doublecheck_env, # Optional, defaults to `None`
    args=training_args,
    train_dataset=dataset
)
trainer.train()
```

Example implementation of such an environment:
```python
from typing import List, Callable, Dict, Any, Sequence, Tuple
from vllm import LLM, SamplingParams, RequestOutput

class DoubleCheckEnv:

    def step(self,
             states: List[Dict[str, Any]],
             llm: LLM,
             sampling_params: SamplingParams) -> Tuple[List[Dict[str, Any]], List[RequestOutput]]:
        
        outputs = llm.chat([state["messages"] for state in states], sampling_params=sampling_params) # type: ignore
        for i, state in enumerate(states):
            state["messages"].append({'role': 'assistant', 'content': outputs[i].outputs[0].text})
            state["messages"].append({'role': 'user', 'content': 'Are you sure?'})
            state["prompt_tokens"] = len(outputs[i].prompt_token_ids)

        outputs = llm.chat([state["messages"] for state in states], sampling_params=sampling_params) # type: ignore

        for i, state in enumerate(states):
            state["messages"].append({'role': 'assistant', 'content': outputs[i].outputs[0].text})
            state["completed"] = True
        return states, outputs

    def generate(self, prompts: List[List[Dict[str, Any]]], llm: LLM, sampling_params: SamplingParams) -> List[Sequence[int]]:
        all_completed = False
        states = [{"messages": m, "completed": False, "prompt_tokens": -1} for m in prompts]
        outputs = [None] * len(prompts)
        while not all_completed:
            states, outputs = self.step(states, llm, sampling_params)
            all_completed = all(state["completed"] for state in states)
        all_ids = [list(output.prompt_token_ids) + list(output.outputs[0].token_ids) for output in outputs]
        completion_ids = [all_ids[i][states[i]["prompt_tokens"]:] for i in range(len(outputs))]
        return completion_ids
```

Fixes # (issue)
- https://github.com/huggingface/trl/issues/2723
- https://github.com/huggingface/trl/discussions/2704
- https://github.com/huggingface/trl/issues/2809

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/grpo_trainer.py
```diff
@@ -371,15 +375,13 @@ def data_collator(features):  # No data collation is needed in GRPO
...
-                        # Automatic Prefix Caching caches the KV cache of existing queries, so that a new query can
-                        # directly reuse the KV cache if it shares the same prefix with one of the existing queries.
-                        # This is particularly useful here because we generate completions from the same prompts.
                         enable_prefix_caching=True,
                         max_model_len=self.args.vllm_max_model_len,
                     )
                 self.sampling_params = SamplingParams(
                     temperature=args.temperature,
                     max_tokens=self.max_completion_length,
+                    skip_special_tokens=False,
```
ywang96, line 384: FYI - on vLLM we have another parameter `spaces_between_special_tokens`(set to `True` by default) that adds spaces between special tokens in the output when `skip_special_tokens=False`.

willccbb: Gotcha, thanks for heads up. Will do some testing and make sure everything looks OK at token-level, maybe don't need that line.

willccbb: Reverted that line. If users want different SamplingParams settings they can specify that at the Environment level anyway.

qgallouedec (commented):
Would it be possible to wrap the generate method instead? It would allow keep grpo as is

willccbb:
I'm not quite sure what you mean by wrapping the generate method here. Many parts of the codebase touch the LLM object, wrapping it in another object would require changing the access in each place. Overriding generate entails a high amount of complexity on the user end, as most applications will either want to use the true generate (or chat, which calls generate) method. Adding an if/else Environment route like in this PR is the simplest approach I could think of which allows users to directly use the LLM object as-is in their rollouts, while still allowing the Trainer to reference LLM normally throughout. Note that this enables many requested features--tool use, sampling strategies, agentic interactions--to be encapsulated within Environments, avoiding further complexity down the road.

If you have something specific in mind that you could illustrate with a short snippet I'm happy to try.

accupham:
Seems like the wrapper idea could be implemented internally. The user interface could remain the same with the environments idea.

Also, I think there might be a bug in the `DoubleCheckEnv.generate` example. I think this is the correct implementation:

```python
# original
completion_ids = [all_ids[states[i]["prompt_tokens"]:] for i, output in enumerate(outputs)]
# fixed
completion_ids = [all_ids[i][states[i]["prompt_tokens"]:] for i in range(len(prompts))]
```

docs/source/grpo_trainer.md
accupham, line 1: Examples are best way to learn how to use libraries... I took your example and added illustrative comments to make it easier to understand what is going on and how a user might implement their own.

```python
class DoubleCheckEnv:
    """
    Example Environment that:
      1) Sends an initial user prompt to the LLM.
      2) Appends the assistant's reply and a follow-up user query: "Are you sure?".
      3) Sends everything again to the LLM for a final response.
      4) Returns just the completion tokens for each prompt.
    """

    def step(
        self,
        states: List[Dict[str, Any]],
        llm: LLM,
        sampling_params: SamplingParams
    ) -> Tuple[List[Dict[str, Any]], List[RequestOutput]]:
        # First LLM call for each state's messages
        outputs = llm.chat([s["messages"] for s in states], sampling_params=sampling_params)
        for i, state in enumerate(states):
            state["messages"].append({
                "role": "assistant", 
                "content": outputs[i].outputs[0].text
            })
            state["messages"].append({
                "role": "user", 
                "content": "Are you sure?"
            })
            # Track prompt_tokens to later slice out the completion part
            state["prompt_tokens"] = len(outputs[i].prompt_token_ids)

        # Second LLM call after "Are you sure?" is appended
        outputs = llm.chat([s["messages"] for s in states], sampling_params=sampling_params)
        for i, state in enumerate(states):
            state["messages"].append({
                "role": "assistant", 
                "content": outputs[i].outputs[0].text
            })
            state["completed"] = True

        return states, outputs

    def generate(
        self,
        prompts: List[List[Dict[str, Any]]],
        llm: LLM,
        sampling_params: SamplingParams
    ) -> List[Sequence[int]]:
        # Setup conversation states
        states = [{"messages": p, "completed": False, "prompt_tokens": -1} for p in prompts]
        outputs = [None] * len(prompts)

        # Keep stepping until each conversation is marked complete
        while not all(s["completed"] for s in states):
            states, outputs = self.step(states, llm, sampling_params)

        # Gather prompt+completion IDs, then slice out the prompt portion
        all_ids = [
            list(o.prompt_token_ids) + list(o.outputs[0].token_ids) 
            for o in outputs
        ]
        completion_ids = [
            all_ids[i][states[i]["prompt_tokens"]:] 
            for i in range(len(prompts))
        ]
        return completion_ids

```

willccbb:
@accupham thanks, you may be right, will double check + test shortly

@qgallouedec I would strongly prefer a solution which doesn't require overriding generate for the LLM class directly, as this significantly complicates and limits what can be done with samplers/environments. I think the proposed Environment protocol is about as minimal as possible for accomplishing this.

For example, many environments will want to make multiple calls to generate(...) or LLM.chat(...) (which calls generate), and this is not possible if the method has to be overridden--see the DoubleCheck example in the PR. Users would have to essentially reimplement and maintain full copies of the logic inside of vLLM's generate method without being able to call it directly.

I do understand the goal of avoiding many changes to TRL, but in my opinion this is a very small one which unlocks a large number of use desired cases.

docs/source/grpo_trainer.md
willccbb: Thank you!! Will add that + docstring to the PR

xiangjjj (changes requested):
**Issue with padding and EOS token in multi-step rollouts**
Consider a scenario where the rollout is resulted from possibly multiple llm steps interleaved with tool use, we would have a completion sequence that contains multiple EOS tokens: e.g., `completion_ids = [[1, 2, EOS, 1, 2, EOS], [1, 2, 3, EOS]]`. After padding we obtain `completion_ids = [[1, 2, EOS, 1, 2, EOS], [1, 2, 3, EOS, PAD, PAD]]`. The current approach masks everything after the first EOS token, where `attention_mask=[[1,1,1,0,0,0], [1,1,1,1,0,0,]]`, which would ignore all the rollouts after the first step. I think we should instead mask everything after the last EOS token. Here's an example implementation:
```python
        eos_int = is_eos.int()
        flipped = eos_int.flip(dims=[1])
        first_eos_in_reversed = flipped.argmax(dim=1)
        last_eos_index = eos_int.size(1) - first_eos_in_reversed - 1
        sequence_indices = torch.arange(is_eos.size(1), device=device).expand(is_eos.size(0), -1)
        completion_mask = (sequence_indices <= last_eos_index.unsqueeze(1)).int()
```

Please let me know if this makes sense. Thank you!

willccbb:
Ah good catch, yeah I see what you mean... I think this could just be handled at the user/Environment level by setting a different EOS token for the tokenizer + manually appending it to completion_ids, though would be good to flag somewhere.

My personal preference would be to eventually allow for user-level computation of rewards + masks in addition to completion_ids, though this would require more significant code changes.

xiangjjj:
Thanks for the prompt response. Do you think my proposed change can be incorporated to this pull request as it is backwards compatible with the single-step rollout implementation.

I like the idea of allowing the user to have fine-grained control. For now, the environment implementation only returns `completion_ids` making it difficult to have a more structured way to organize the data and potentially some supplementary information could get lost. Do you have thoughts about how to make it more flexible?

willccbb:
Yes, will test + add shortly.

willccbb:
@xiangjjj One complication is that many base model tokenizers have pad_token_id = eos_token_id, so when padding a batch, the "last EOS token" will be the last token in the pad sequence. Trying out a couple workarounds.

xiangjjj:
Ah, I see! That is tricky. Thank for this!

willccbb:
Simplest solution I think is to move the masking logic into the respective vllm/transformers generate routes. vLLM now masks based on completion_ids length rather than the position of the first EOS token.

xiangjjj:
> Simplest solution I think is to move the masking logic into the respective vllm/transformers generate routes. vLLM now masks based on completion_ids length rather than the position of the first EOS token.

Sure, it makes sense and should resolve the masking issue! Thanks for fixing this!

trl/trainer/grpo_trainer.py
```diff
@@ -536,10 +542,16 @@ def _prepare_inputs(self, inputs: dict[str, Union[torch.Tensor, Any]]) -> dict[s
                 self._last_loaded_step = self.state.global_step
 
             # Generate completions using vLLM: gather all prompts and use them in a single call in the main process
+            all_prompts = gather_object(prompts)
             all_prompts_text = gather_object(prompts_text)
             if self.accelerator.is_main_process:
-                outputs = self.llm.generate(all_prompts_text, sampling_params=self.sampling_params, use_tqdm=False)
-                completion_ids = [out.token_ids for completions in outputs for out in completions.outputs]
+                if self.env is not None:
+                    completion_ids = self.env.generate(
```
xiangjjj, line 549: I was wondering if we could enhance the output structure to include additional metadata alongside the current completion_ids. In multi-step rollouts, having step-wise details available directly would be really useful for determining the final rewards. Currently, parsing completion_ids to reconstruct this information feels a bit cumbersome. Would it be possible to return a dict that encapsulates both the tokens and the extra metadata?

willccbb: I'll defer to @qgallouedec on that one, could be added with pretty minimal changes (happy to do so), but there are also easy enough workarounds for what you're describing (computing rewards at generation time and caching them in a data structure accessed by your reward functions). My first priority with this PR is getting basic support for Environments enabled.

xiangjjj: Do you have suggestions on how to cache the information? I'm new to `trl` and have not done that before.

trl/trainer/grpo_trainer.py
xiangjjj, line 1: @qgallouedec @willccbb Regarding [logits_to_keep](https://github.com/huggingface/trl/pull/2810/files#diff-964e6fd373aa93037604064cb2b822d7f8e2735e33f791065acf2c4c3552d393R598), what do you think is the best strategy to filter out tool observation tokens? My concern is that when we invoke a web search tool—which might return thousands of tokens—those tokens could overwhelm the policy tokens during gradient updates. If this extraneous information turns out to be noisy, it might adversely affect the policy gradient learning. We can leave this out of the current PR, but I’d appreciate your thoughts on ideas/directions to resolve this as I'm new to `trl`.

willccbb: For now I'd suggest having deterministic processing of tool call results to avoid excessive outputs, that'll cause problems whether or not we mask tokens. In some experiments on multi-step code tool use, I found limiting allowed printouts to 500 chars per step worked reasonably well.

Not clear to me that naive masking of tool outputs "makes sense" for GRPO algorithmically, and for now it probably is fine to treat tool call results as just part of the LLM output. Especially you are letting your model "reason" for many tokens per step, it should not be a major problem I think.

xiangjjj: I think that applying policy gradient directly to tool call outputs is not theoretically sound—especially in outcome reward methods. Note that I'm not suggesting "naive masking of tool outputs." I propose that we exclude the tool output tokens from log-likelihood computation that leads to the policy gradient and KL loss functions. While one might argue this is an empirical question depending on the setup of the system, I’m interested in learning about what kind of design in grpo would enable more flexible loss masking.

xiangjjj: I’m fine with keeping this PR as is, but I’m interested in exploring a design that allows for more flexible loss masking in the future.

willccbb: Yeah, I see what you mean. I think this is a very interesting research question, but hard to say definitively right now. Including them as normal response tokens is essentially forcing the model to "model" the tool calls directly, which feels reasonable to me. 

Papers exploring this issue are pretty recent:
- https://arxiv.org/pdf/2412.16145 
- https://arxiv.org/pdf/2410.09302

If interest in multi-step RL continues to grow, I would imagine having more specialized trainers in the future could make sense. For now this is just a way to get something that runs using GRPO, whether or not it is the most principled.

vladrad:
All I would love to help out with this as I am working on it myself.

vladrad:
@willccbb I reach out via your email on your profile! let me know if you want to Colab on this as I am continuing to work on it.

qgallouedec:
I still don't understand why wrapping would limit what you can do. For example for the double call:

```python
def wrapper_decorator(generate_func):
    def generate_wrapper(*args, **kwargs):
        ...  # stuff before
        result1 = generate_func(*args, **kwargs)
        result1 = generate_func(*args, **kwargs)
        ...  # stuff after
        return result
    return generate_wrapper

trainer.llm.model.generate = wrapper_decorator(trainer.llm.model.generate)
```

Taking the env paradigm, I think it should work as is with the main branch with the something like:


```python
env = MyEnv(...)

def wrapper_decorator(generate_func):
    def generate_wrapper(*args, **kwargs):
        prompts = args[0]
        return env.generate(prompts, self, *args, **kwargs)
    return generate_wrapper

trainer.llm.model.generate = wrapper_decorator(trainer.llm.model.generate)
```

I might be missing something though

vladrad:
@qgallouedec
I tried something similar before without success, but I'll give it another go.

I've been experimenting with combining chat and completion responses during training. The idea is to score each response based on its format and content. If a mistake is detected, another LLM—one that provides the correct answer—is consulted. This secondary LLM offers a brief hint to guide the correction, and then the response is re-generated.

For example, this dataset snippet:

```
<think>
Oh, the user wants me to call a tool.
</think>
</answer>
I am going to call X...
</answer>
<tool>
<function name=read_file>...</function>
</tool>
```

If a mistake is found (say, the tool call should be within the `<answer>` tags), the correction process would work like this:
```
<think>
Oh, I forgot the format requires tool calls to be within the answer 
```

Now I go back with the hint and try to get a competition:
```
tags. This will provide the correct format the user requested.
</think>
</answer>
I am going to call X...
<tool>
<function name=read_file>...</function>
</tool>
</answer>
```

My goal is to see I can get auto correction via hints scored on top of it. 

I have made a really hacky overfitted solution where I was training in epoch runs like this, creating the dataset and then going back to round 2,3,4 GRPO training... slowly guiding it to the right answer. Now I think I need to work on getting an actual solution that's how I ended up here. Happy to help/code things up and test. 

Thanks all!

willccbb:
@qgallouedec biggest problem is that `LLM.chat` relies on `LLM.generate`, and many (most?) multi-step interaction protocols will want to use `LLM.chat`. If we override `generate` as you propose, any call to `chat` inside of our wrapper will result in recursive blowup. We also have to keep all of our logic contained within a single wrapper function, and we also can't easily maintain global state within/across rollouts (for things like precomputing/caching rewards to be retrieved by reward functions, which can be objects with access to the Env state).

It also is just much nicer to be able to have access to the `SamplingParams` and `LLM` objects directly, as this is how people typically develop agent applications on top of vLLM. The added complexity to the trainer by allowing an Env object is pretty minor, but it unlocks quite a bit from the user perspective. Other libraries which have already built these kinds of environments (TextArena, reasoning-gym, etc.) are way easier to adapt if we can just "use the model like a normal LLM" rather than having to rewrite all of the chat parsing logic again for every application.

trl/trainer/grpo_trainer.py
Some-random: Does that mean in the current implementation we're still calculating KLD and policy gradient on environment responses? I thought we're not doing that based on my understanding of `DoubleCheckEnv` but now I'm confused...

amitlevy:
Any update? The PR is falling behind the GRPO changes on main

willccbb:
Working on a refactor now which I think should allow directly using the main TRL branch. Approach is to extend `GRPOTrainer` to a class `GRPOEnvTrainer` which only needs to override the `_generate_and_score_completions` function, which I think is already a reasonable encapsulation of the minimum logic needed to implement custom rollout strategies. Once that's tested + pushed to the verifiers repo I'll probably just close this PR.

willccbb:
Working now on the `dev` branch of [verifiers](https://github.com/willccbb/verifiers/tree/dev), will clean up some things + merge to main shortly. Closing this PR, will maybe revisit later, but overloading the trainer seems to be the best method for supporting these kinds of features for now.

PR #2811, opened 2025-02-10
✨ Add vLLM guided decoding support to GRPO Trainer

kldzj:
# What does this PR do?

Adds the ability to pass vLLM's `GuidedDecodingParams` through to the `llm.generate` call.

Example:

```py
from vllm.sampling_params import GuidedDecodingParams

training_args = GRPOConfig(
    use_vllm = True,
    vllm_guided_decoding_params = GuidedDecodingParams(
        backend="outlines",
        regex="<reasoning>\n.*\n</reasoning>\n<answer>\n.*\n</answer>",
    ),
    # ...
)
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


qgallouedec:
Thanks for contributing @kldzj! For the record, can you explain briefly what is the motivation behind using GuidedDecodingParams?

kldzj:
@qgallouedec When using the GRPO trainer, we likely want the model to respond in a specific format, in the example above we enforce the `<reasoning>\n...\n</reasoning>\n<answer>\n...\n</answer>` format right away, without spending many training steps for the model to learn the correct format through our reward functions.

Let me know if there's any problem or flaw in my logic with this.

qgallouedec:
C'est très interessant.
Regarding the implementation, it a bit annoying because `GuidedDecodingParams` isn't json serializable so it causes error. A fair alternative is to only do like this instead

```python
@dataclass
 class GRPOConfig(TrainingArguments):
    ...
    vllm_guided_decoding_regex: Optional[str] = None
```

and 

```python
if args.vllm_guided_decoding_regex is not None:
    guided_decoding = GuidedDecodingParams(backend="outlines", regex= args.vllm_guided_decoding_regex)
```

it's less flexible but explicitly exposes the regex and probably easier for the user.

kldzj:
@qgallouedec Made the suggested change. :)

PR #2725, opened 2025-01-31
⚡ Fix GRPO PEFT

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #2818, opened 2025-02-10
👨‍👩‍👧 GRPO + PEFT + vLLM

winglian:
# What does this PR do?

unlocks PEFT + GRPO + vllm without the complexity of shipping lora weights to vllm via the REST API. This implementation simply merges the lora weights into the base model and ships that to vllm using the existing python API.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -125,7 +125,7 @@ class GRPOTrainerTester(unittest.TestCase):
 
             self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
 
-            # Check the peft params have changed and the base model params have not changed
+            # Check that the peft params have changed and the base model params have not changed
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 if n in base_param_names:  # We expect the base model params to be the same
@@ -168,7 +168,7 @@ class GRPOTrainerTester(unittest.TestCase):
 
             self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
 
-            # Check the params have changed
+            # Check that the params have changed
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
@@ -203,7 +203,7 @@ class GRPOTrainerTester(unittest.TestCase):
 
             self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
 
-            # Check the params have changed
+            # Check that the params have changed
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
@@ -239,7 +239,7 @@ class GRPOTrainerTester(unittest.TestCase):
 
             self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
 
-            # Check the params have changed
+            # Check that the params have changed
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
@@ -278,7 +278,7 @@ class GRPOTrainerTester(unittest.TestCase):
 
             self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
 
-            # Check the params have changed
+            # Check that the params have changed
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
@@ -356,7 +356,7 @@ class GRPOTrainerTester(unittest.TestCase):
 
             self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
 
-            # Check the params have changed
+            # Check that the params have changed
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
@@ -395,7 +395,7 @@ class GRPOTrainerTester(unittest.TestCase):
 
             self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
 
-            # Check the params have changed
+            # Check that the params have changed
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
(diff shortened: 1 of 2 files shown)


trl/trainer/utils.py
```diff
@@ -920,6 +922,66 @@ def get_peft_config(model_args: ModelConfig) -> "Optional[PeftConfig]":
...
+    state_dict = {}
+    key_list = [key for key, _ in model.named_modules() if model.prefix not in key]
+    for key in key_list:
+        try:
+            _, target, _ = _get_submodules(model, key)
+        except AttributeError:
+            continue
+        with onload_layer(target):
+            weight_key = key.replace(base_model_prefix, "") + ".weight"
+            bias_key = key.replace(base_model_prefix, "") + ".bias"
```
winglian, line 955: I know this is pretty janky, so would love feedback on making it better.

trl/trainer/utils.py
```diff
@@ -920,6 +922,66 @@ def get_peft_config(model_args: ModelConfig) -> "Optional[PeftConfig]":
...
+            elif hasattr(target, "weight"):
+                if any(
+                        skip in key
+                        for skip in [
+                            ".original_module",
+                            ".modules_to_save",
+                            ".base_layer",
+                        ]
+                ):
+                    continue
```
winglian, line 978: same for this

winglian:
@qgallouedec The downside there is that you're limited to the lora support in vllm, which means no DoRA support. This approach almost any peft adapter type could be used. While LoRA does converge pretty quickly too compared to full parameter training, dora seems to be more performant.

Screenshot 2025-02-10 at 9 25 06 AM

qgallouedec:
This seems quite reasonable, thank you for the clear explanation.

qgallouedec:
Another pointer that could be useful:

> It is possible to call `model.merge_adapte`r (optionally with `adapter_names` argument), then `model.state_dict()`, then `model.unmerge_adapter`.
> The `state_dict` may require some clean up though, depending on what you need to do with it (I couldn't infer that from the PR).
> By clean up, I mean: After `merge_and_unload` the model looks like the base model. But `merge_adapter` keeps the LoRA structure, with the wrapped base model, LoRA weights etc. still being present in the `state_dict`.

From @BenjaminBossan

winglian:
I tried

```
                unwrapped_model.merge_and_unload()
                state_dict = unwrapped_model.base_model.model.state_dict()
                unwrapped_model.unmerge_adapter()
```

but the state dict results still has the prefix of `base_model.model.`

BenjaminBossan (commented):
Thanks a lot for this PR.

To elaborate on the quote by Quentin, the steps would be:

1. Call `model.merge_adapter()`.
2. Get the `state_dict` of the merged model.
3. Clean up the `state_dict`: Since the base weights already contain the merged LoRA weights, we can remove all LoRA weights
4. Call `model.unmerge_adapter()` if we need to restore the previous state (note that `unmerge_adapter` unmerges _all_ adapters, so if some were already merged before step 1, they need to be re-merged, but it's probably not relevant here)

Here is a small demonstration in code:

```python
from peft import LoraConfig, get_peft_model

model_id = "meta-llama/Llama-3.2-1B"
model = AutoModelForCausalLM.from_pretrained(model_id)
config = LoraConfig()
model = get_peft_model(model, config)
model.merge_adapter()
sd = model.state_dict()
new_sd = {k.removeprefix("base_model.model.").replace(".base_layer", ""): v for k, v in sd.items() if model.prefix not in k}
model.unmerge_adapter()
```

qgallouedec:
I've added the suggested modification to this branch: https://github.com/huggingface/trl/pull/2725 it seems to work...!

BenjaminBossan:
> I've added the suggested modification to this branch: #2725 it seems to work...! EDIT: DORA included

Nice, I added a comment there. Hopefully, one of these branches can be merged soon :)

winglian:
I re-did this PR to account for the other changes, and also updated the test to use lora.

trl/trainer/grpo_trainer.py
```diff
@@ -491,21 +491,20 @@ def _move_model_to_vllm(self):
...
-            elif is_peft_available() and isinstance(unwrapped_model, PeftModel):
+                unwrapped_model = unwrapped_model._orig_mod
+            if is_peft_model(unwrapped_model):
                 unwrapped_model.merge_adapter()
                 state_dict = unwrapped_model.state_dict()
-                unwrapped_model.unmerge_adapter()
                 state_dict = {
-                    k.removeprefix("base_model.model.").replace(".base_layer", ""): v
+                    k.removeprefix("base_model.model.")
+                    .removeprefix("base_model.model.")
```
BenjaminBossan, line 500: Duplicate, you can remove the 2nd line.

BenjaminBossan (commented):
Not sure if this PR is still required after #2725 has been merged, but I did a quick review just in case.

qgallouedec:
thanks for the followup @BenjaminBossan !

tests/test_grpo_trainer.py
```diff
@@ -416,9 +416,10 @@ def test_training_vllm(self):
                 report_to="none",
                 use_vllm=True,
                 vllm_device="cuda:0",  # will raise a warning, but allows this test to work with only one GPU
+                vllm_gpu_memory_utilization=0.5,  # reduce since because we use the same device for training and vllm
             )
             trainer = GRPOTrainer(
-                model="trl-internal-testing/small-Qwen2ForCausalLM-2.5",
+                model="Qwen/Qwen2.5-0.5B-Instruct",  # Tiny is too small for vLLM
```
qgallouedec, line 422:
```suggestion
                model="Qwen/Qwen2.5-0.5B-Instruct",  # tiny is too small for vLLM
```

trl/trainer/grpo_trainer.py
```diff
@@ -249,7 +249,7 @@ def __init__(
         # Reference model
         if is_deepspeed_zero3_enabled():
             self.ref_model = AutoModelForCausalLM.from_pretrained(model_id, **model_init_kwargs)
-        elif peft_config is None:
+        elif not is_peft_model(model):
```
qgallouedec, line 252: allows to support model that is already wrapped by peft

qgallouedec (approved):

zaddy6:
I noticed training without LORA leads to better performance, here is an example without LORA it starts to max the rewards at 1k steps, with Lora it doesnt learn 
image

winglian:
> I noticed training without LORA leads to better performance, here is an example without LORA it starts to max the rewards at 1k steps, with Lora it doesnt learn 
> 

What rank and dataset? It learns pretty quickly with rank 64 o. The gsm8k dataset

zaddy6:
> > I noticed training without LORA leads to better performance, here is an example without LORA it starts to max the rewards at 1k steps, with Lora it doesnt learn
> 
> What rank and dataset? It learns pretty quickly with rank 64 o. The gsm8k dataset

alpha also 64?

wusijie123:
Total steps changed when using this code
image
why the Total optimization steps = Num examples * Num Epochs / Gradient Accumulation steps?
in previews version,Total optimization steps = Num examples * Num Epochs / Total train batch size

By the way,even using lora , most of the time I am waiting for VLLM to generate results. Is there any way to speed up the generation？

PR #2776, opened 2025-02-05
📉 Optimize GRPO memory usage by redefining `per_device_batch_size` as generations per device

qgallouedec:
### Background
In the current implementation of GRPO, each device samples `N` prompts. For these `N` prompts, `G` generations are produced, and the loss is computed over them. However, this approach is highly memory-intensive, especially when using `G ≥ 8`, often leading to OOM errors, even with a batch size of just 1 prompt per device.

### Proposed change 
This PR introduces a more flexible approach:  
- Instead of defining `per_device_batch_size` as the number of prompts per device, it now represents the **number of generations per device**.  
- This allows for much greater flexibility in choosing the number of generations (`G`) and the batch size per device.  
- The only constraint is that the global batch size (`num_processes * per_device_batch_size`) must be divisible by `G`.

Note that these settings should be equivalent:
```python
num_generations = ...  # eg, 8
num_prompts_per_device = ...  # eg, 1
# main
GRPOConfig(num_generations=num_generations, per_device_batch_size=num_prompts_per_device, ...)
# this PR
GRPOConfig(num_generations=num_generations, per_device_batch_size=num_generations*num_prompts_per_device, ...)
```

` (new setting) and ``per_device_batch_size==1` (old setting)

### Benefits
- Reduces memory, making it feasible to use higher `G` or completion length without hitting OOM
- Provides more control over batch size and generation count per device
- Ensures scalability while maintaining the integrity of GRPO's training logic

Before: 

image

Now:

image

PR #2824, opened 2025-02-10
🫘 Add `set_seed()` call in GRPO to ensure unique seed for each process

qgallouedec:
Fixes the issue described in https://github.com/huggingface/trl/pull/2776#issuecomment-2649053292


plaguss (approved):

mark-mcl:
this seems to also affect sampling, tried patching this and I get a different prompt on each device now

kiddj:
thanks! does using a different seed for each process affect training?

qgallouedec:
True @kiddj!! Fixed in [95fcfeb](https://github.com/huggingface/trl/pull/2824/commits/95fcfeb003d8ba0ab339faeabcb41365786cee58)

lewtun (approved):
Nice fix, and could possibly benefit from a regression test to check completions are indeed unique across devices (ignore if you want to go fast though!)

PR #2837, opened 2025-02-12
fix: grpo loss fix

pointerhacker:
# What does this PR do?

This PR modifies how GRPO calculates the policy ratio in its loss function. Specifically, it changes the implementation to use an explicit reference policy's log probabilities instead of a detached version of the current policy's log probabilities. This change better represents the ratio between new and old policies in the GRPO algorithm.

The core change is:
```diff
- per_token_loss = torch.exp(per_token_logps - per_token_logps.detach()) * advantages.unsqueeze(1)
+ per_token_loss = torch.exp(per_token_logps - ref_per_token_logps.detach()) * advantages.unsqueeze(1)
```

Key improvements:
- More accurate representation of the probability ratio between new and old policies
- Better training stability by using a true reference point
- Proper implementation of policy update mechanism in GRPO

## Before submitting
- [ ] This PR fixes a typo or improves the docs (not applicable)
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request)
      to it if that's the case.
- [x] Did you write any new necessary tests? (Added tests to verify correct loss calculation and gradient flow)

## Implementation Details
- The `ref_per_token_logps` tensor represents log probabilities from the reference (old) policy
- Both tensors maintain the same shape and dimensions
- The detach() operation ensures gradients don't flow through the reference policy
- The change preserves the original advantage weighting mechanism

## Testing
The following tests have been added/verified:
- Loss calculation produces expected values
- Gradient flow is correct (only through current policy)
- Works correctly with various batch sizes and sequence lengths

## Who can review?
Anyone in the community is free to review the PR once the tests have passed. Particularly interested in feedback from those familiar with GRPO implementation in TRL.

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -689,7 +689,7 @@ class GRPOTrainer(Trainer):
 
         # x - x.detach() allows for preserving gradients from x
         advantages = inputs["advantages"]
-        per_token_loss = torch.exp(per_token_logps - per_token_logps.detach()) * advantages.unsqueeze(1)
+        per_token_loss = torch.exp(per_token_logps - ref_per_token_logps.detach()) * advantages.unsqueeze(1)
         per_token_loss = -(per_token_loss - self.beta * per_token_kl)
         loss = ((per_token_loss * completion_mask).sum(dim=1) / completion_mask.sum(dim=1)).mean()
 


qgallouedec:
That's right, answer here: https://github.com/huggingface/trl/pull/2565#issuecomment-2595837761 :)

pointerhacker:
Thank you for your reply.

pointerhacker:
As a result, the GRPO objective just minimizes the KL divergence between the policy model and the reference policy.

pointerhacker:
> That's right, answer here: [#2565 (comment)](https://github.com/huggingface/trl/pull/2565#issuecomment-2595837761) :)

Based on what you said that 'As a result, the GRPO objective just minimizes the KL divergence between the policy model and the reference policy,' I have another question: How can this approach achieve preference alignment effects?

PR #2757, opened 2025-02-03
🔂 Use vLLM prefix caching for speedup

winglian:
# What does this PR do?
Since we're rolling out the same prompt several times, this is a quick win.

Screenshot 2025-02-03 at 6 48 25 PM

https://docs.vllm.ai/en/v0.6.4/automatic_prefix_caching/apc.html 

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs).
- [ ] Did you write any new necessary tests?

## Who can review?
@qgallouedec @kashif 
Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #2855, opened 2025-02-13
Optimize vllm num_generations

edbeeching:
# What does this PR do?

Adds an optimization of vllm batching by using `n=num_generations` in the vllm `SamplingParameters`.

@qgallouedec I looking at some examples in the debugger and the ordering of the gathered prompts always seems to be always `[num_generations] * [num_prompts]` but I am concerned there could be edge cases where this is not the case, what do you think?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -539,6 +539,9 @@ class GRPOTrainer(Trainer):
             # Generate completions using vLLM: gather all prompts and use them in a single call in the main process
             all_prompts_text = gather_object(prompts_text)
             if self.accelerator.is_main_process:
+                # Since 'prompts' contains 'num_generations' duplicates, we first take unique prompts, and generate 
+                # num_generations outputs for each one. This is faster than generating outputs for each duplicate 
+                # prompt individually.
                 ordered_set_of_prompts = list(dict.fromkeys(all_prompts_text))
                 all_outputs = self.llm.generate(
                     ordered_set_of_prompts, sampling_params=self.sampling_params, use_tqdm=False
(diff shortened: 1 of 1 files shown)


qgallouedec:
It should always be the case indeed.
The built-in `set` isn't ordered right?
Is vllm faster when you pass 1 prompt with n=N outputs, than with N times the same prompt for n=1?

edbeeching:
> It should always be the case indeed. The built-in `set` isn't ordered right? Is vllm faster when you pass 1 prompt with n=N outputs, than with N times the same prompt for n=1?

Unfortunately, `set` is not ordered. Yes vllm can share the prefill for the n generations so it is faster, I profiled around 1.5x faster with the changes in this PR at 2k `max_completion_length`.

qgallouedec:
Nice!! I am surprised, I expected a smaller speedup given that the prefix should already be reused since https://github.com/huggingface/trl/pull/2757.

We should probably do the same with tranformers generation in a future PR, if it makes sense.

Anyway, can you just add comment somewhere to explain why we do this?

winglian:
My guess is it's an easier optimization for vllm to understand that single prompt has multiple generations than sending the same prompt multiple times from the https://github.com/huggingface/trl/pull/2776 refactor.

edbeeching:
@qgallouedec, without diving into the codebase of vllm, I would assume that the prefix cache is only used to compare a new batch of prompts with previously processed prompts. The system prompt is shared across all prompts, so this is cached and reused for all batches, whereas a new batch of prompts would first all need to have their prefill calculated and entered into the cache before vllm could identify that there are `num_generations` of the prompts are exactly the same. Hence you get some improvement when you move from `prompt*num_generations` to `n` generations for each prompt.

Let me know if you would like me to clarify.

qgallouedec:
Thanks Ed! Actually I meant adding a comment in the code to concisely explain why we merge the prompts. Something like `Since 'prompts' contains 'num_generations' duplicates, we first take unique prompts, generate outputs for them, and then replicate the results. This is faster than generating outputs for each duplicate prompt individually.`.

qgallouedec (approved):
Nice! Feel free to merge

PR #2866, opened 2025-02-15
[GRPO] Disable prefix cache for models with sliding window

XZ-X:
# What does this PR do?

The PR disables prefix cache in vLLM for models with sliding window.
Currently, vLLM does not support sliding window & prefix cache.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -400,7 +400,8 @@ class GRPOTrainer(Trainer):
                 profiling_patch = patch(
                     "vllm.worker.worker.Worker._assert_memory_footprint_increased_during_profiling", return_value=None
                 )
-                # for now, vLLM does not support `enable_prefix_caching` with a model that has sliding window
+                # vLLM does not support `enable_prefix_caching` with a model that has sliding window,
+                # see https://github.com/vllm-project/vllm/issues/3355
                 has_sliding_window = hasattr(self.model.config, "sliding_window")
                 with world_size_patch, profiling_patch:
                     self.llm = LLM(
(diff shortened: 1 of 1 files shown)


XZ-X:
> thanks! do you have maybe a pointer, a PR, or an issue in the vllm repo?

Here's an issue in the vllm repo: https://github.com/vllm-project/vllm/issues/3355

The issue was discussing a related problem on sliding window and prefix cache. But the discussion indeed implies vLLM does not support prefix cache with sliding window at the moment.

Btw, an alternative change is to disable the sliding window, but it will affect model's context length.

trl/trainer/grpo_trainer.py
```diff
@@ -400,6 +400,8 @@ def data_collator(features):  # No data collation is needed in GRPO
                 profiling_patch = patch(
                     "vllm.worker.worker.Worker._assert_memory_footprint_increased_during_profiling", return_value=None
                 )
+                # for now, vLLM does not support `enable_prefix_caching` with a model that has sliding window
```
qgallouedec, line 403:
```suggestion
                # vLLM does not support `enable_prefix_caching` with a model that has sliding window, see https://github.com/vllm-project/vllm/issues/3355
```

XZ-X: I updated the comments. Thank you for the suggestion!

qgallouedec:
In fact, the cases where prefix caching can be applied seem to be a little trickier. Having a parameter seems to be a better option. See #2900

XZ-X:
That makes sense. Let's make it a cli parameter. Maybe consider adding the case (where the model uses sliding window) to the doc where users may want to set prefix caching to `false`?

PR #2867, opened 2025-02-15
BCOTrainer version upgrade fixes

claralp:
# What does this PR do?

This PR fixes to some issues that I identified with the BCOTrainer on the latest version upgrade:
- remove extra shuffles, super class already handles data shuffling. These destroy training reproducibility
- fix serialization of UDM classifier, not all parameters were saved previously
- accelerate.gather not hanging on empty lists anymore
- convert delta to tensor before logsigmoid, this led to numeric errors
- do not update delta value in eval mode

Let me know if you have any questions

## Before submitting
      Pull Request section?
      to it if that's the case.
      [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs). - changes do not affect docs
- [x] Did you write any new necessary tests? - tests exist already

## Who can review?

@kashif @qgallouedec

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,6 +89,7 @@ EXTRAS = {
     "peft": ["peft>=0.8.0"],
     "quantization": ["bitsandbytes"],
     "scikit": ["scikit-learn"],
+    "bco": ["scikit-learn", "joblib"],
     "test": ["parameterized", "pytest-cov", "pytest-rerunfailures", "pytest-xdist", "pytest"],
     # vllm is not available on Windows
     # vllm 0.7.3 causes hanging while gathering. temporary pinning the version until the issue is resolved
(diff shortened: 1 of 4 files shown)


claralp:
@kashif yes, but could use pickle as alternative if you prefer. Just BCO with UDM needs this

PhilipMay:
Hey @kashif and @claralp - what about this PR. We should fix the missing dependency somehow. :-)

PhilipMay:
@claralp since we do not get a response here my suggestion would be to just add joblib to the dependencies so the CI is green. A green CI hopefully then catches enough attention and some progress.

claralp:
@kashif since I got no reply, I just added the job lib library now everywhere were sklearn is required as well

kashif:
yes we can add the requirements to a `EXTRAS` in the setup.py? perhaps a `bco`?

kashif (approved):

Issue #2703, opened 2025-01-30
GRPO: Why does loss start at 0 for first K steps and then increase over time?


arnavgarg1:
### Reproduction

Hi all! 

I've been trying to train a variety of models using GRPO, but I noticed that the train/loss metric remains 0 or close to 0 throughout training even after a large number of steps (>200). My mean rewards also don't change significantly during this same period. This seems unexpected and might indicate that the optimization step isn't working as intended.

However, on the official docs page, I see that that the image of the learning curves shared suggests that loss can remain close to 0 for long periods of time and actually increases instead of decreasing: https://huggingface.co/docs/trl/main/en/grpo_trainer#trl.GRPOConfig. Similarly, in @philschmid's blogpost from earlier today (https://www.philschmid.de/mini-deepseek-r1), I noticed a similar trend where loss stayed 0 for nearly 200 steps before increasing. 

This makes it seem like it is the expected behavior, but I'm having a hard time understanding it.

I had a few questions that I am hoping someone is able to help me understand:
1. Is there an issue with the loss computation in GRPO?
2. Should loss values be expected to remain near zero in this setting? What does an increasing loss suggest?
3. Could this be related to specific hyperparameters or gradient updates not being applied correctly?

Would appreciate any insights or guidance on debugging this! Thanks!

### System Info

- OS: Linux
- Transformers 4.48.1
- TRL: Main Branch

### Checklist

- [x] I have checked that my issue isn't already filed (see [open issues](https://github.com/huggingface/trl/issues?q=is%3Aissue))
- [x] I have included my system information
- [x] Any code provided is minimal, complete, and reproducible ([more on MREs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks))
- [x] Any code provided is properly formatted in code blocks, (no screenshot, [more on code blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks))
- [x] Any traceback provided is complete

qgallouedec:
Intereting question!

The answer is in the math. If you calculate the value of the loss (ignore the gradient), you'll see that it's equal to $\beta \mathrm{KL}$. That's why it starts at 0 and that's why it's incresing.

arnavgarg1:
Thanks for the prompt response @qgallouedec! 

Does this mean that the loss itself is not a reliable indicator of training progression and we should primarily rely on KL and reward trends instead?

[...]

PR #2869, opened 2025-02-15
[GRPO] Reduce steps where loss starts to remain at 0, accelerate training

zhangsheng377:
#2703 

Due to the standardization of `advantages`, the `advantages` is summed in the loss calculation to make it 0, i.e. loss=β KL. But in the first step, since the actor model and ref model are the same, KL=0, which means the first loss must be 0.
And we believe that the main reason why GRPO can start training with an initial loss of 0, besides still having gradients, is due to the calculation error of the GPU.

So, our acceleration approach is to add small perturbations to the loss to accelerate the model changes caused by computational errors in the first few steps; After the formal training begins, the added disturbance is too small and can be ignored, or equivalently dropped out, to improve the model's generalization.
Specifically, we choose to add perturbations after the reward is standardized, because the key to the first step having a loss of 0 is that summing up the standardized rewards will result in it being exactly 0. 
﻿
After adding perturbations, the loss obtains a perturbation with a mean of 0 and a variance of 1e-4. Because we have observed that the initial KL is larger than 1e-4, which means that after the first few steps of acceleration, the disturbance of 1e-4 can be quickly ignored.
﻿
Verification result:
image

After the above improvements, after about 15 steps of training, the loss starts training;
When there is no improvement, the loss needs to exceed 100 steps to reach the same level.

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -639,6 +639,10 @@ class GRPOTrainer(Trainer):
         std_grouped_rewards = std_grouped_rewards.repeat_interleave(self.num_generations, dim=0)
         advantages = (rewards - mean_grouped_rewards) / (std_grouped_rewards + 1e-4)
 
+        epsilon = 1e-4
+        noise = torch.randn_like(advantages) * epsilon
+        advantages = advantages + noise
+
         # Slice to keep only the local part of the data
         process_slice = slice(
             self.accelerator.process_index * len(prompts),


XZ-X:
It's an interesting and inspiring finding! But I am not sure I fully understand one technical point in the analysis: a loss with the value zero does not mean the model is not learning?

There might be some negative advantages, meaning that reducing the coefficient for those advantages may further reduce the loss to below zero. That means the optimization would try to minimize the probability on those negative advantages.

Specifically, I agree with you that the formula `torch.exp(per_token_logps - per_token_logps.detach()) * advantages.unsqueeze(1)` always evaluates to `advantages.unsqueeze(1)`. From my understanding, the gradients is not zero though (due to the `detach` operation)?

qgallouedec:
@XZ-X is right, the model doesn't learn thanks to gpu error. It learns because even if the loss is zero, its gradient isn't zero.

zhangsheng377:
> It's an interesting and inspiring finding! But I am not sure I fully understand one technical point in the analysis: a loss with the value zero does not mean the model is not learning?
> 
> There might be some negative advantages, meaning that reducing the coefficient for those advantages may further reduce the loss to below zero. That means the optimization would try to minimize the probability on those negative advantages.
> 
> Specifically, I agree with you that the formula `torch.exp(per_token_logps - per_token_logps.detach()) * advantages.unsqueeze(1)` always evaluates to `advantages.unsqueeze(1)`. From my understanding, the gradients is not zero though (due to the `detach` operation)?

Well, yes, the gradients is not zero when loss is 0. So in this way, if I want to speed up its startup process, I just need to multiply the advantages by a factor? In order to avoid affecting the actual training, do I need to limit the number of steps that can take effect? 
Wait, can I just turn off the warmup?

qgallouedec:
Are the proposed changes backed up by a study and results that describe and demonstrate the added value of the associated trick?

PR #2873, opened 2025-02-15
🧶 [GRPO][vLLM + LoRA] Move unmerge of PEFT model after weight loading

### LINKED ISSUE #2856: [GRPO] `use_peft: true` together with `use_vllm: true` not working as intended
I’ve come across a potential issue and was wondering if anyone else has experienced it. When `use_peft: true` is set, the behavior differs depending on `use_vllm`. Here is a list of combinations I tried.

- if `use_peft: true` and `use_vllm: false` then ✅
- if `use_peft: true` and `use_vllm: true` then ❌
- if `use_peft: false` and `use_vllm: true` then ✅
- if `use_peft: false` and `use_vllm: false` then ✅

Where:
- ✅ the reward curve displays the expected pattern.
- ❌ the reward curve remains mostly flat.

Any insights into what could cause this behavior?

---

Reference issue: #2725

---

UPDATE: I also added an MRE and plots below in the comments for clarification.

XZ-X:
# What does this PR do?

The `state_dict` obtained from the PEFT model seems to be a shallow copy of model weights.
Therefore, if we `unmerge` the model before generation, the weights are "rolled-back" to the base weights without the adapter.

I move the unmerge operation to the end of generation.

Fixes #2856  (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -486,16 +486,12 @@ class GRPOTrainer(Trainer):
         logits = logits[:, -logits_to_keep:]
         return selective_log_softmax(logits, input_ids)  #  compute logprobs for the input tokens
 
-    def _move_model_to_vllm(self):
+    def _move_model_to_vllm(self, unwrapped_model):
         with unwrap_model_for_generation(
             self.model, self.accelerator, gather_deepspeed3_params=self.args.ds3_gather_for_generation
         ) as unwrapped_model:
-            if is_compiled_module(unwrapped_model):
-                unwrapped_model = unwrapped_model._orig_mod
             if is_peft_model(unwrapped_model):
-                unwrapped_model.merge_adapter()
-                state_dict = unwrapped_model.state_dict()
-                unwrapped_model.unmerge_adapter()
+                state_dict = unwrapped_model.state_dict()                
                 # Remove base_model and base_layer prefixes
                 state_dict = {
                     k.removeprefix("base_model.model.").replace(".base_layer", ""): v for k, v in state_dict.items()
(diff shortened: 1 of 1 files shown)


matt23654:
This could be a stupid question, but is it guaranteed that in a multigpu/multinode situation no other process/worker will be doing anything with the model weights during this window where the adaptors are merged? Come to think of it, is it a good idea to use merge/unmerge adaptor at all given all the possible configurations with multiple GPUS, multiple nodes, etc?

XZ-X:
> This could be a stupid question, but is it guaranteed that in a multigpu/multinode situation no other process/worker will be doing anything with the model weights during this window where the adaptors are merged? Come to think of it, is it a good idea to use merge/unmerge adaptor at all given all the possible configurations with multiple GPUS, multiple nodes, etc?

Thanks for noticing that!

Correct me if I'm wrong. From my understanding, the sync between processes happens at `self.optimizer.step()` in the trainer. That is, before the model weights are updated, all processes will sync with each other. Therefore, the model weights will change in a synchronized way.

XZ-X:
Changed.

qgallouedec:
@AndreiCComan thanks for your work in #2856! Can you try if the patch still gives better results?

qgallouedec:
Thanks! I'm running some experiments to validate

AndreiCComan:
> @AndreiCComan thanks for your work in #2856! Can you try if the patch still gives better results?

Thanks @qgallouedec for the help and @XZ-X for the speedy update!

@qgallouedec here are the plots with "patched" (my inefficient implementation), @XZ-X's "fix-vllm-peft-v1", and @XZ-X's "fix-vllm-peft-v2":




"patched" and "fix-vllm-peft-v2" now seem to be fairly close to each other. Note: I changed my instance in the meantime. I guess a slight difference between the two is justified.

qgallouedec:
Seems to work 🥳

Thanks for the work guys

W B Chart 17_02_2025, 18_53_25

trl/trainer/grpo_trainer.py
```diff
@@ -510,9 +509,11 @@ def _move_model_to_vllm(self):
...
             else:
                 state_dict = unwrapped_model.state_dict()
-        if self.accelerator.is_main_process:
-            llm_model = self.llm.llm_engine.model_executor.driver_worker.model_runner.model
-            llm_model.load_weights(state_dict.items())
+            if self.accelerator.is_main_process:
+                llm_model = self.llm.llm_engine.model_executor.driver_worker.model_runner.model
+                llm_model.load_weights(state_dict.items())
+            if is_peft_model(unwrapped_model):
+                unwrapped_model.unmerge_adapter()
```
qgallouedec, line 516:
```suggestion
            # Unmerge the adapter to restore the model to its original state.
            # This must be done after loading weights to ensure they correspond to the merged state.
            if is_peft_model(unwrapped_model):
                unwrapped_model.unmerge_adapter()
```

qgallouedec (approved):

trl/trainer/grpo_trainer.py
XZ-X: Thanks for the suggestion. I added the comments.

PR #2881, opened 2025-02-17
[GRPO] Fix loss normalization

edbeeching:
# What does this PR do?

The current GRPO implementation uses `per-sequence normalization`, this PR corrects this to be `global normalization`

Details:
In Causal Language Modelling, we typically use `global normalization` to scale the loss, so that each unmasked token's loss provides the same contribution to the total loss. Example from transformers codebase: https://github.com/huggingface/transformers/blob/fae0f3dde83b7a54441f7a5bb0fc45d354fe81ce/src/transformers/loss/loss_utils.py#L24-L29


qgallouedec (approved):

BramVanroy:
As seen on Twitter, some discussion about this change: https://x.com/danielhanchen/status/1900844864134410695

gameofdimension:
should we make it configurable?

PR #2889, opened 2025-02-18
🪪 Adds profiling decorators for GRPOTrainer

edbeeching:
# What does this PR do?

Adds a profiling decorator and wraps some key methods in the GRPO trainer. If you are reporting to wandb you now have a profiling section with some timings:
image


qgallouedec:
Cool!!

qgallouedec (approved):

PR #2902, opened 2025-02-19
Remove CUDA synchronization in mean_token_accuracy

cyyever:
# What does this PR do?

mean_token_accuracy computation invokes ```item()``` on token counts, which triggers CUDA to CPU synchronization. That synchronization becomes a minor performance bottleneck in LLM fine-tuning, as indicated by the following profiling snapshot from v0.15.1:

image

That bottleneck has been fixed in this PR by accumulating the correct and total token counts in tensors. ```item()``` calls are delayed until ```trainer.log()```. 
The effects of the change are indicated by another profiling that the bottleneck disappears:

image

Because the metrics are cleared immediately after logging, this  change should be safe and backwards-compatible.

## Before submitting
 
      Pull Request section?

## Who can review?

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -478,17 +478,27 @@ class SFTTrainer(Trainer):
             correct_tokens = correct_predictions.sum()
 
             # Gather the correct_tokens and total_tokens across all processes
-            correct_tokens = self.accelerator.gather_for_metrics(correct_tokens)
-            total_tokens = self.accelerator.gather_for_metrics(total_tokens)
+            correct_token_sum = self.accelerator.gather_for_metrics(correct_tokens).detach().sum()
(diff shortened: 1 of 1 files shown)


qgallouedec:
I don't understand the profiling actually. Where do you get that this line is the bottleneck?
Thank for contributing!

cyyever:
@qgallouedec It is the sixth line in the first picture.

qgallouedec:
This?

```
dequantize_4bit (bitsandbytes/functional. py:1380)
```

qgallouedec:
The comparison is not very clear to me tbh, do you have clearer results, like two training (one with main, one with your branch) where we can see the speedup in term of steps/sec?

cyyever:
@qgallouedec Of course, I will provide a comparison ASAP.

PR #2901, opened 2025-02-19
GRPOTrainer adds support for OpenAI API-compatible servers to models that generate samples

ZYM66:
# What does this PR do?
The original generative model is loaded on a single GPU, which becomes very slow when the generation length reaches 4096 or more. Therefore, I suggest loading the generation model externally to leverage multi-GPU support.

## Motivation Behind This Feature:

The motivation for this feature stems from performance limitations when using the original vllm generative model, which currently loads on a single GPU. This becomes a bottleneck, especially when the generation length reaches 4096 tokens or more, significantly slowing down the process. By utilizing multiple GPUs, we can distribute the workload more efficiently and drastically improve performance, especially for long-generation tasks.

This feature is crucial for my project, as faster generation times are essential. I believe it could also benefit the broader community by enhancing the scalability and efficiency of model inference on multi-GPU setups.

## Requested Feature:

The feature I am requesting involves loading the generative model outside of the current single-GPU setup in order to leverage multi-GPU capabilities. This would improve performance for tasks involving large generation lengths, making the library more efficient and scalable for demanding use cases.

## Code Snippet:
GRPOTrainer  `__init__`
```python
elif self.args.use_openai_compatible_server:
    api_endpoint = args.api_endpoint
    api_key = args.api_key

    openai_serving_client = openai.OpenAI(base_url=api_endpoint, api_key=api_key, )
    # set the openai logger to ERROR level to avoid mess log information
    logging.getLogger("openai").setLevel(logging.ERROR)
    logging.getLogger("httpx").setLevel(logging.ERROR)
    self.ref_model_name = args.ref_model_name

    self.ref_llm = partial(openai_serving_client.chat.completions.create,
            model=args.ref_model_name,
            max_tokens=self.max_completion_length,
            temperature=args.temperature,
    )
```

GRPOTrainer `_prepare_inputs`
```python
elif self.args.use_openai_compatible_server:
    completions = []
    # don't use any chattemplate, because the server have load it.
    for prompt in prompts:
        # request server
        response = self.ref_llm(messages=prompt)
        completion_text = response.choices[0].message.content
        completion_tokens = self.processing_class.encode(completion_text, add_special_tokens=False)
        completions.append(completion_tokens)

    completion_ids = completions
    completion_ids = [torch.tensor(ids, device=device) for ids in completion_ids]
    completion_ids = pad(completion_ids, padding_value=self.processing_class.pad_token_id)
    prompt_completion_ids = torch.cat([prompt_ids, completion_ids], dim=1)

[...]

Issue #2887, opened 2025-02-18
Bottleneck in GRPO training


ZYM66:
### Feature request

The default code only deploys the reference model on a single GPU, which is slow and inefficient, as other processes have to wait for VLLM completion.

![Image](https://github.com/user-attachments/assets/98ed5d9d-45ec-48e4-8556-903ef11a6afc)

GPU7 deployed the reference model

source code below:
```python
if self.accelerator.is_main_process:
    vllm_device = self.args.vllm_device
    if vllm_device == "auto":
        if torch.cuda.device_count() == 1:
            vllm_device = "cuda:0"  # particular case when training with onyl 1 GPU: share it
        else:
            vllm_device = f"cuda:{self.accelerator.num_processes}"  # take the next GPU idx
    # Check that the requested device is available
    if vllm_device.split(":")[0] == "cuda" and int(vllm_device.split(":")[1]) >= torch.cuda.device_count():
        raise ValueError(
            f"The requested device for vllm ({vllm_device}) is not available. You are likely using vLLM "
            "without restricting the number of GPUs for training. Set the `--num_processes` argument to a "
            "value lower than the number of GPUs available on your machine—typically, reducing it by one "
            f"is sufficient. In your case: `--num_processes {torch.cuda.device_count() - 1}`."
        )
    # Check that the requested device is not also used for training
    if vllm_device in {f"cuda:{idx}" for idx in range(self.accelerator.num_processes)}:
        warnings.warn(
            f"The requested device {vllm_device} is also being used for training. For higher throughput "
            "and to avoid out-of-memory errors, it is recommended to use a dedicated device for vLLM. "
            "If this is intentional, you may ignore this warning but should adjust "
            "`vllm_gpu_memory_utilization` accordingly."
        )
    # vLLM is not compatible with accelerate. So we need to patch it to make sure we can (1) place the vLLM
    # model on the desired device (world_size_patch) and (2) avoid a test that is not designed for our
    # setting (profiling_patch).
    world_size_patch = patch("torch.distributed.get_world_size", return_value=1)
    profiling_patch = patch(
        "vllm.worker.worker.Worker._assert_memory_footprint_increased_during_profiling", return_value=None
    )
    with world_size_patch, profiling_patch:
        self.llm = LLM(
            model=model.name_or_path,
            device=vllm_device,
            gpu_memory_utilization=self.args.vllm_gpu_memory_utilization,
            dtype=self.args.vllm_dtype,
[...]

PR #2929, opened 2025-02-22
Supporting multi-vLLM inference for GRPO

ghrua:
# What does this PR do?

The current GRPO only uses 1 GPU for inference and $N-1$ GPUs for training, where $N$ is the world size. This PR allows users to leverage $K$ GPUs for inference and $N-K$ GPUs for training.

Method:
1. Apply for the resources with $N$ GPUs
2. initializing the distributed training  with $N-K$ processes  on GPU 0, 1, ... $N-K-1$.
3. Initializing the $K$ vLLM objects on 0 to $K-1$ training processes. However, the model parameters of the $K$ vLLM  objects will be placed on GPU $N-K$, $N-K+1$, ..., $N-1$.
4. Using the a vLLM object  to do the real-time inference for a subset of the prompts in each process.
5. Gathering the decoded data from 0 to $K-1$ processes. 

Constraints:
1. The number of vLLM instances should be less or equal to the number of training processes, i.e., K <= N-K. For example, for an environment with 8 GPUs, we can have at most 4 GPUs for inference. Other configs are also allowed, e.g., 1 inference 7 training, 2 inference 6 training, etc.

I tested with 4 A100-80 GPU. The script is from [open-r1](https://github.com/huggingface/open-r1). I train the Qwen-math-1.5B model with `G=num_generation=32` and  `per_device_train_batch_size=32`. Under this setting, I find that the config (2 infer, 2 train) is ~12% faster than (1 infer, 2 train). Probably scalling up the batch size and `num_generation` would introduce higher gains.

## Before submitting
      Pull Request section?
      https://github.com/huggingface/trl/issues/2887
      https://github.com/huggingface/trl/pull/2901

## Who can review?
@qgallouedec @edbeeching

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -781,6 +781,7 @@ class GRPOTrainerTester(unittest.TestCase):
                 use_vllm=True,
                 vllm_device="cuda:0",  # will raise a warning, but allows this test to work with only one GPU
                 vllm_guided_decoding_regex=r"<reasoning>\n.*\n</reasoning>\n<answer>\n.*\n</answer>",
+                vllm_gpu_memory_utilization=0.4,  # reduce the memory utilization rate to reduce memory usage
             )
             trainer = GRPOTrainer(
                 model="Qwen/Qwen2.5-0.5B-Instruct",  # tiny model is too small for vLLM
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -176,6 +176,14 @@ class GRPOConfig(TrainingArguments):
             "(`pip install vllm`)."
         },
     )
+    vllm_worker_num: Optional[int] = field(
+        default=1,
+        metadata={
+            "help": "The number of vllm works used for inference. Notably, this number should be less or equal to "
+            "the number of process used for distributed training (i.e., `--num_processes`). In addition, please "
+            "ensure that `vllm_worker_num + num_processes <= world_size`."
+        },
+    )
     vllm_device: Optional[str] = field(
         default="auto",
         metadata={
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -62,6 +62,10 @@ class GRPOConfig(TrainingArguments):
         use_vllm (`bool`, *optional*, defaults to `False`):
             Whether to use vLLM for generating completions. If set to `True`, ensure that a GPU is kept unused for
             training, as vLLM will require one for generation. vLLM must be installed (`pip install vllm`).
+            ensure that `vllm_worker_num + num_processes <= world_size`.
+        vllm_worker_num (`int`, *optional*, defaults to `1`):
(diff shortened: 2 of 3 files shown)


qgallouedec:
Thanks a lot @ghrua! Huge work! I'll test it asap. It definitely makes sense :)

ghrua:
Hey @qgallouedec! Thanks for reviewing my commit. 

I am sorry that I haven't tried vLLM 0.7.3. Let me do a quick check. Please wait a moment.

qgallouedec:
So it works with this version?
I'm asking because currently, main branch hangs at some point when you use vLLM 0.7.3, and I'm curious to know if your PR solves it

ghrua:
> So it works with this version? I'm asking because currently, main branch hangs at some point when you use vLLM 0.7.3, and I'm curious to know if your PR solves it

Yes, my commit works well with vLLM 0.7.3. I re-run the code to double check it. 

Two parts may be helpful:
1. I use a context manager to control the device that each vLLM can access during inference, because the vLLM may mis-use the devices of distributed training processes: https://github.com/huggingface/trl/blob/79e9af0c7e179f9be5ee1a0f0e88b3e4ac82cb70/trl/trainer/grpo_trainer.py#L670
2. Three additional patches are used for a smoother initialization of vLLM (though not very elegant 😅): https://github.com/huggingface/trl/blob/79e9af0c7e179f9be5ee1a0f0e88b3e4ac82cb70/trl/trainer/grpo_trainer.py#L474

qgallouedec (commented):
Ok now I understand better what you are trying to do.
Why do you need $K$ processes to control $K$ instances of vLLM? Couldn't you do everything in the main process?

ghrua:
> Ok now I understand better what you are trying to do.
> Why do you need $K$ processes to control $K$ instances of vLLM? Couldn't you do everything in the main process?

That was my initial plan... I tried to initialise a vllm_list for K vllm instances in the main process and use the ThreadExecutor to call the inference in parallel. However, it will meet lots of errors (I forget the details, but seems to be the inconsistency between q and kv cache.) I didn't check the detailed code of vLLM, but those errors seems come from the conflict of some shared static variables. And I find this issue: https://github.com/vllm-project/vllm/issues/1676. Please let me know if I misunderstood anything.


Afterwards, I change my strategy by initialising K instances in K processes.

SeiunSky0131:
> This PR allows users to leverage K GPUs for inference and N − K GPUs for training.

Hi, that's a great work! However, when I test this code on my server equipped with 8 NVIDIA A100 80G, with 6 GPUs(GPU 0-5) for training and 2 GPUs(GPU 6-7) for vLLM engine. However, I find that GPU 7's utilization is always 0 throughout the training process.(See the blue line of GPU7 in the following figure)
截屏2025-03-10 10 15 57

It seems that no generation tasks are allocated to GPU 7. Could you check whether this situation applies to your code?

skepsun:
It seems not to solve the problem that vllm hangs when we want to use large models (for example, 32b) which could only be placed on multiple cards.

qgallouedec:
Closed via #3094

PR #2939, opened 2025-02-23
✋ Prevent applying the chat template to tokenized datasets

DanFosing:
Fixes:
- SFTTrainer sometimes tried applying chat template to tokenized datasets, not sure why was it happening (there might be some errors in maybe_apply_chat_template and maybe_convert_to_chatml that caused it to happen?), this fix should prevent the code from even considering to do that if the dataset is already tokenized


kashif:
@DanFosing which version of TRL are you using?

DanFosing:
I experienced this issue with both v0.15.1 and with the alpha version downloaded using:
`pip install git+https://github.com/huggingface/trl.git`

DanFosing:
Oh and I forgot to mention, max_seq_length didn't seem to work for me for some reason, the warning says it will be deprecated in v.0.20.0 but are you sure it wasn't deprecated already? (that's why I added a comment there in the code, but it's not related to the main fix)

kashif:
@DanFosing  ok so kindly remove the `max_seq_length` from the `sft_config.py` and move the chat template logic inside the already defined `if not is_processed:`  where then it makes sense... instead of a new `if not is_processed:` block

kashif:
we can fix the warning and say: removed in version `0.16.0`

qgallouedec:
`maybe_apply_chat_template` applies the chat template if needed. hence "maybe". Are encountering a bug? If so what's the traceback?

kashif:
i don't think there is a bug, but please correct me @DanFosing if I am mistaken, the issue is that it's doing this extra work when it's not needed.

qgallouedec:
> Oh and I forgot to mention, max_seq_length didn't seem to work for me for some reason

WDYM "didn't seem to work"? Same question, is an exception raised? if so what's the traceback?
Have you tried to pull the very last commits? Could be related to #2947

qgallouedec:
> i don't think there is a bug, but please correct me @DanFosing if I am mistaken, the issue is that it's doing this extra work when it's not needed.

for clarification, the only extra work done is iterating through the dataset:

https://github.com/huggingface/trl/blob/5c0591319646c171e8ea213d1692058f4bf68ead/trl/data_utils.py#L218-L221

which is usually very fast

qgallouedec:
That being said, I'm ok to add the `if not is_processed:` to avoid extra logging/iteration

qgallouedec (approved):

kashif (approved):

PR #2890, opened 2025-02-18
🪂 Don't gather logits in SFT to avoid hanging

qgallouedec:
# What does this PR do?

Fixes #2879 

I'm not sure why it does always hang, but it seems that at some point in the training, always the same, it can hang while trying to gather the logits. The fix consist in gathering only the number of correct tokens and the total number of tokens.

The only way was to copy the content of `compute_token_accuracy` in the `SFTTrainer`, which I think should be ok as it is only used once in the codebase.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #2874, opened 2025-02-16
[SFT] fix check for AutoLigerKernelForCausalLM

kashif:
# What does this PR do?
Covers the edge case when an AutoLigerKernelForCausalLM is passed to the SFTTrainer and use_liger is the default

fixes https://github.com/linkedin/Liger-Kernel/issues/568

PR #2940, opened 2025-02-23
🐯 Fix LigerKernel for SFTTrainer

lewtun:
# What does this PR do?

Fixes an issue where `use_liger=True` threw an error in the loss computation because the fix in #2874 was partially reverted in [this line](https://github.com/huggingface/trl/pull/2890/files#diff-67e157adfcd37d677fba66f610e3dfb56238cc550f221e8683fcfa0556e0f7caR180) of #2890. Without this change, one gets errors on the loss computation

```
[rank0]:     shift_logits = outputs.logits[..., :-1, :].contiguous()
[rank0]:                    ~~~~~~~~~~~~~~^^^^^^^^^^^^^
[rank0]: TypeError: 'NoneType' object is not subscriptable
```

Command to test:

```shell
trl sft --model_name_or_path Qwen/Qwen2.5-0.5B     --dataset_name trl-lib/Capybara     --output_dir Qwen2.5-0.5B-SFT --use_liger true
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -173,7 +173,6 @@ class SFTTrainer(Trainer):
             )
         if isinstance(model, str):
             model = self._create_model_from_path(model, args)
-        self.use_liger = is_liger_kernel_available() and isinstance(model, AutoLigerKernelForCausalLM)
 
         # PEFT configuration and model wrapping
         if peft_config is not None:
@@ -462,7 +461,7 @@ class SFTTrainer(Trainer):
         )
 
         # Compute token accuracy if we have labels and if the model is not using Liger (no logits)
-        if "labels" in inputs and not self.use_liger:
+        if "labels" in inputs and not self.args.use_liger:
             shift_logits = outputs.logits[..., :-1, :].contiguous()
             shift_labels = inputs["labels"][..., 1:].contiguous()
 


trl/trainer/sft_trainer.py
```diff
@@ -173,7 +173,6 @@ def __init__(
             )
         if isinstance(model, str):
             model = self._create_model_from_path(model, args)
-        self.use_liger = is_liger_kernel_available() and isinstance(model, AutoLigerKernelForCausalLM)
```
qgallouedec, line 176: What if the model used is already a liger model (and args.use_liger = False)?

lewtun: Good point, but as [recommended](https://github.com/linkedin/Liger-Kernel/issues/568#issuecomment-2661294559) by the Liger maintainer, we shouldn't be passing Liger models at the init (just patched via the config)

Should we deprecate passing the Liger model to the trainer or would you prefer an alternative?

qgallouedec: It seems that this command doesn't achieve what it's supposed to:

```python
>>> from liger_kernel.transformers import AutoLigerKernelForCausalLM
>>> model =  AutoLigerKernelForCausalLM.from_pretrained("Qwen/Qwen2.5-0.5B")
Applied Liger kernels to Qwen2
Sliding Window Attention is enabled but not implemented for `sdpa`; unexpected results may be encountered.
>>> isinstance(model, AutoLigerKernelForCausalLM)
False
```

I don't know of an easy way to test whether a model is liger (perhaps @ByronHsu does?).

Anyway, with your change you can still pass a liger model to the trainer. But you'll need to specify `use_liger=True`. Which sounds good to me.

qgallouedec (approved):

trl/trainer/sft_trainer.py
BenasdTW: This line doesn't work when a Liger Model is converted to PEFT before passing into the trainer. It does not respect args.use_liger either.

BenasdTW: > Good point, but as [recommended](https://github.com/linkedin/Liger-Kernel/issues/568#issuecomment-2661294559) by the Liger maintainer, we shouldn't be passing Liger models at the init (just patched via the config)
> 
> Should we deprecate passing the Liger model to the trainer or would you prefer an alternative?

Deprecating it might cause some problem. LoRA+ requires the model instance to be created before hand (to create optimizer). The flag `use_liger` does not convert PEFT wrapped model to liger model.

```python
model = get_peft_model(model, lora_config)
optimizer = create_loraplus_optimizer(
    model=model,
    optimizer_cls=torch.optim.AdamW,
    lr=lr,
    eps=eps,
    betas=betas,
    weight_decay=weight_decay,
    loraplus_lr_ratio=loraplus_lr_ratio,
)
```

BenasdTW: Also, many VLMs can't be loaded with `AutoLigerKernelForCausalLM`. For example, monkey patching `apply_liger_kernel_to_qwen2_vl()` is required for Qwen2-VL

edbeeching:
@kashif @qgallouedec @lewtun ,

I think Liger models are now supported nativately in in transformers if the `use_liger_kernel==True` flag is set, perhaps we can drop the support for this in the `SFTTrainer` and use the native transformers implementation?

kashif:
i think so too... we will need to pin the transformer version but yes should be a better soluton

qgallouedec:
Thanks @ed!


> i think so too... we will need to pin the transformer version but yes should be a better soluton

After checking, it seems like `use_liger_kernel` exists for at least 4.46, which is the min version in TRL. So we shouldn't need to bump transformers

https://github.com/huggingface/transformers/blob/052e652d6d53c2b26ffde87e039b723949a53493/src/transformers/training_args.py#L1521

qgallouedec:
it has been introduced in 4.45

lewtun:
Thanks for the pointer to `transformers`! Just to confirm, the current PR is fine to merge since the model init is taken care of in this line? https://github.com/huggingface/trl/pull/2940/files#r1967113819

Issue #2962, opened 2025-02-26
`DeepSpeedZeRoOffload` is incompatible with `DeepSpeed>=0.16.4`


jamesbraza:
### Reproduction

https://github.com/deepspeedai/DeepSpeed/pull/6847 released in `deepspeed==0.16.4` renamed `DeepSpeedZeRoOffload._register_hooks_recursively`, which `trl` uses here: https://github.com/huggingface/trl/blob/v0.15.2/trl/models/utils.py#L174

```none
9: [rank75]:   File "/home/james/code/molr1-two/.venv/lib/python3.12/site-packages/trl/models/utils.py", line 173, in add_hooks
9: [rank75]:     optimizer_offload._register_hooks_recursively(optimizer_offload.module)
9: [rank75]:     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
9: [rank75]: AttributeError: 'DeepSpeedZeRoOffload' object has no attribute '_register_hooks_recursively'
```

### System Info

- Platform: Linux-5.15.0-112-generic-x86_64-with-glibc2.35
- Python version: 3.12.9
- PyTorch version: 2.5.1+cu124
- CUDA device(s): NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3
- Transformers version: 4.50.0.dev0
- Accelerate version: 1.3.0
- Accelerate config: not found
- Datasets version: 3.2.0
- HF Hub version: 0.28.1
- TRL version: 0.15.0.dev0
- bitsandbytes version: not installed
- DeepSpeed version: 0.15.4
- Diffusers version: not installed
- Liger-Kernel version: 0.5.2
- LLM-Blender version: not installed
- OpenAI version: 1.61.1
- PEFT version: 0.14.0

### Checklist

- [x] I have checked that my issue isn't already filed (see [open issues](https://github.com/huggingface/trl/issues?q=is%3Aissue))
- [x] I have included my system information
- [x] Any code provided is minimal, complete, and reproducible ([more on MREs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks))
[...]

Issue #2953, opened 2025-02-25
[GRPO] Gradient mode issue when num_iterations > 1


willccbb:
When setting `num_iterations > 1` I get a runtime error at the first backward pass due to inference-only tensors being used. Runs fine when `num_iterations = 1`.

Error:
```
[rank0]:     return inner()
[rank0]:            ^^^^^^^
[rank0]:   File "/workspace/projects/finetuning/.venv/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1790, in inner
[rank0]:     result = forward_call(*args, **kwargs)
[rank0]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/projects/finetuning/.venv/lib/python3.11/site-packages/transformers/models/qwen2/modeling_qwen2.py", line 256, in forward
[rank0]:     hidden_states = self.input_layernorm(hidden_states)
[rank0]:                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/projects/finetuning/.venv/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
[rank0]:     return self._call_impl(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/projects/finetuning/.venv/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1844, in _call_impl
[rank0]:     return inner()
[rank0]:            ^^^^^^^
[rank0]:   File "/workspace/projects/finetuning/.venv/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1790, in inner
[rank0]:     result = forward_call(*args, **kwargs)
[rank0]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/workspace/projects/finetuning/.venv/lib/python3.11/site-packages/transformers/models/qwen2/modeling_qwen2.py", line 222, in forward
[rank0]:     return self.weight * hidden_states.to(input_dtype)
[rank0]:            ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[rank0]: RuntimeError: Inference tensors cannot be saved for backward. To work around you can make a clone to get a normal tensor and use it in autograd.
```

Args:
```python
...
[...]

PR #2871, opened 2025-02-15
🫔 [GRPO] Pass wrapped model to `unwrap_model_for_generation` for DeepSpeed Stage-3 compatibility

kiddj:
# What does this PR do?

Fixes DeepSpeed Stage-3 compatibility by passing the wrapped model (`self.model_wrapped`) to `unwrap-model_for_generation` instead of `self.model`.

Previously, `unwrap_model_for_generation(model)` was called with a model passed via function like `compute_loss`, which avoided any DeepSpeed Stage-3 conflicts. In the current implementation, the model is obtained directly from `self`, causing a crash under Stage-3.

This fix ensures consistency by always passing the wrapped model instead of `self.model`, preventing the Stage-3 compatibility issue.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

@qgallouedec

PR #2963, opened 2025-02-26
🚀  Supporting `deepspeed>=0.16.4`'s rename

jamesbraza:
Fixes https://github.com/huggingface/trl/issues/2962

### DIFF
--- a/trl/models/utils.py
+++ b/trl/models/utils.py
@@ -141,6 +141,8 @@ def remove_hooks(model: "DeepSpeedEngine") -> None:
         optimizer_offload = model.optimizer.parameter_offload
     elif model.optimizer is not None:
         optimizer_offload = model.optimizer
+    else:
+        raise NotImplementedError(f"Didn't yet handle optimizer {model.optimizer}.")
 
     for param in iter_params(optimizer_offload.module, recurse=True):
         param.ds_active_sub_modules.clear()
@@ -170,6 +172,8 @@ def add_hooks(model: "DeepSpeedEngine") -> None:
         optimizer_offload = model.optimizer.parameter_offload
     elif model.optimizer is not None:
         optimizer_offload = model.optimizer
+    else:
+        raise NotImplementedError(f"Didn't yet handle optimizer {model.optimizer}.")
     optimizer_offload._register_hooks_recursively(optimizer_offload.module)
 
 
--- a/trl/models/utils.py
+++ b/trl/models/utils.py
@@ -19,6 +19,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Literal, Optional, Union
 
 from accelerate.utils import is_deepspeed_available
+from packaging import version
 from transformers import PreTrainedModel, PreTrainedTokenizer
 
 from .modeling_value_head import AutoModelForCausalLMWithValueHead, AutoModelForSeq2SeqLMWithValueHead
@@ -174,7 +175,7 @@ def add_hooks(model: "DeepSpeedEngine") -> None:
         optimizer_offload = model.optimizer
     else:
         raise NotImplementedError(f"Didn't yet handle optimizer {model.optimizer}.")
-    if deepspeed.__version__.split(".") >= ("0", "16", "4"):
+    if version.parse(deepspeed.__version__) >= version.parse("0.16.4"):
         # Account for renaming in https://github.com/deepspeedai/DeepSpeed/pull/6847
         optimizer_offload._register_deepspeed_module(optimizer_offload.module)
     else:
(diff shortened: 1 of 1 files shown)


yeruoforever:
TypeError: '>=' not supported between instances of 'list' and 'tuple'

trl/models/utils.py
```diff
@@ -170,7 +172,13 @@ def add_hooks(model: "DeepSpeedEngine") -> None:
         optimizer_offload = model.optimizer.parameter_offload
     elif model.optimizer is not None:
         optimizer_offload = model.optimizer
-    optimizer_offload._register_hooks_recursively(optimizer_offload.module)
+    else:
+        raise NotImplementedError(f"Didn't yet handle optimizer {model.optimizer}.")
```
qgallouedec, line 176: In what case does it occur?

trl/models/utils.py
```diff
@@ -170,7 +172,13 @@ def add_hooks(model: "DeepSpeedEngine") -> None:
...
+    if deepspeed.__version__.split(".") >= ("0", "16", "4"):
```
qgallouedec, line 177:
```suggestion
    if version.parse(deepspeed) >= version.parse("0.16.4"):
```

qgallouedec: you'll need to import version as well

qgallouedec:
Thanks a lot @jamesbraza

qgallouedec:
by the way, is this change needed as well? https://github.com/huggingface/trl/pull/2871

trl/models/utils.py
jamesbraza: Oh yeah basically I was getting IDE (PyCharm) warnings: "Local variable 'optimizer_offload' might be referenced before assignment"

> In what case does it occur?

It could happen if `model.optimizer is None`. That being said, this `else` branch definitely doesn't happen in practice, otherwise someone would have already reported this logic bug, as it'd lead to a `NameError`.

Feel free to ask me to revert this change, it just nice-ifies this failure branch if hit

trl/models/utils.py
jamesbraza: Just implemented this, thank you for the suggestion!

jamesbraza:
> by the way, is this change needed as well? #2871

Yes, testing it now. Clearly I didn't test this PR previously, as @yeruoforever reported it had a `TypeError` haha, my bad.

jamesbraza:
Hi @qgallouedec I have completed my validations, this PR is ready for merge. I had to also pull in:

- https://github.com/huggingface/trl/pull/2871 (thanks for sharing it)
- I hit https://github.com/huggingface/trl/issues/2953, and to fix it, I changed [this line](https://github.com/huggingface/trl/blob/019fc6dbaa03b888f9d5c1845f0f690da8ed310c/trl/trainer/grpo_trainer.py#L752) to `with torch.no_grad()` instead of `with torch.inference_mode()`

qgallouedec:
Thanks, but I can't see the above mentionned changes. Did you pushed them?

jamesbraza:
> Thanks, but I can't see the above mentionned changes. Did you pushed them?

The other ones I mentioned weren't related to this PR, so they're not here. They are all here: https://github.com/Future-House/trl/tree/working-grpo-2025-02-27

trl/models/utils.py
```diff
@@ -141,6 +142,8 @@ def remove_hooks(model: "DeepSpeedEngine") -> None:
         optimizer_offload = model.optimizer.parameter_offload
     elif model.optimizer is not None:
         optimizer_offload = model.optimizer
+    else:
+        raise NotImplementedError(f"Didn't yet handle optimizer {model.optimizer}.")
```
qgallouedec, line 146:
```suggestion
    else:
        raise RuntimeError("The model optimizer is None, which is not supported.")
```

trl/models/utils.py
qgallouedec:
```suggestion
    else:
        raise RuntimeError("The model optimizer is None, which is not supported.")
```

qgallouedec:
Thanks! Just commit the suggestions and we are good to merge.
Usually [allowing maintainer to edit the PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) makes things easier for us.

jamesbraza:
> Usually [allowing maintainer to edit the PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) makes things easier for us.

Yeah I always do, but for some reason that checkbox isn't present in this PR's right panel :/

screenshot of right panel

Regardless, PR is ready for review again

qgallouedec:
Thanks!

qgallouedec (approved):

PR #2989, opened 2025-02-28
🎲 Add support for additional generation kwargs in GRPO Trainer

nopepper:
- Extend GRPOConfig to support additional generation kwargs
- Update GRPOTrainer to incorporate additional generation parameters
- Add tests for training with additional generation kwargs for both standard and vLLM modes

# What does this PR do?

Adds an extra `additional_generation_kwargs` to `GRPOConfig`, which get passed to the `GenerationConfig` (or `SamplingParams`) in the GRPOTrainer `__init__` method.

This is useful because it allows more flexibility in training, especially for non-standard cases where different assumptions hold.

Also adds two new test cases and some missing `vllm_gpu_memory_utilization=0.5` in places to avoid OOM errors.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -56,6 +56,9 @@ class GRPOConfig(TrainingArguments):
             improving generation speed. However, disabling this option allows training models that exceed the VRAM
             capacity of a single GPU, albeit at the cost of slower generation. Disabling this option is not compatible
             with vLLM generation.
+        additional_generation_kwargs (`dict[str, Any]` or `None`, *optional*, defaults to `None`):
+            Additional keyword arguments for `SamplingParams` when using vLLM, or `GenerationConfig` otherwise.
+            Useful kwargs include `presence_penalty`, `top_p`, `top_k`, `repetition_penalty`, `length_penalty`, `min_p`, etc.
 
         > Parameters that control generation acceleration powered by vLLM
 
@@ -167,6 +170,13 @@ class GRPOConfig(TrainingArguments):
             "generation, improving generation speed. However, disabling this option allows training models that "
             "exceed the VRAM capacity of a single GPU, albeit at the cost of slower generation. Disabling this option "
             "is not compatible with vLLM generation."
+        },
+    )
+    additional_generation_kwargs: Optional[dict] = field(
+        default=None,
+        metadata={
+            "help": "Additional keyword arguments for `SamplingParams` when using vLLM, or `GenerationConfig` otherwise. "
+            "Useful kwargs include `presence_penalty`, `top_p`, `top_k`, `repetition_penalty`, `length_penalty`, `min_p`, etc."
         },
     )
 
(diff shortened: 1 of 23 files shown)


qgallouedec:
Thanks for the suggestion. In fact it's already been suggested in #2728, and I think this solution should actually be avoided: https://github.com/huggingface/trl/pull/2728#issuecomment-2635166424

nopepper:
@qgallouedec Interesting, thanks for the links, I hadn't seen the discussion. I was working on a project that really benefited from custom sampling kwargs (seq2seq task where target significantly overlaps input) and thought it would be useful for others. Would it be possible to add the generation arguments by name then to GRPOConfig?

qgallouedec:
Yes I think it makes sense. I'm not sure that they all share the same default in vLLM and transformers though. You'll need to check

nopepper:
I added them to a new section called `Parameters that control generation` together with `temperature`. We get around the diverging default values problem by just making them optional and not passing any argument that's `None` :D

trl/trainer/grpo_config.py
```diff
@@ -57,6 +55,21 @@ class GRPOConfig(TrainingArguments):
...
+            Temperature for sampling. The higher the temperature, the more random the completions.
+        top_p (`float` or `None`, *optional*, defaults to `None`):
+            Top-p value for sampling. If set to `None`, the default value will be used.
+        top_k (`int` or `None`, *optional*, defaults to `None`):
+            Top-k value for sampling. If set to `None`, the default value will be used.
+        min_p (`float` or `None`, *optional*, defaults to `None`):
+            Min-p value for sampling. If set to `None`, the default value will be used.
+        repetition_penalty (`float` or `None`, *optional*, defaults to `None`):
+            Repetition penalty for sampling. If set to `None`, the default value will be used.
+        length_penalty (`float` or `None`, *optional*, defaults to `None`):
```
qgallouedec, line 70: is this one part of `SamplingParams`?

trl/trainer/grpo_config.py
```diff
@@ -57,6 +55,21 @@ class GRPOConfig(TrainingArguments):
             capacity of a single GPU, albeit at the cost of slower generation. Disabling this option is not compatible
             with vLLM generation.
 
+        > Parameters that control generation
+
+        temperature (`float` or `None`, *optional*, defaults to `0.9`):
+            Temperature for sampling. The higher the temperature, the more random the completions.
+        top_p (`float` or `None`, *optional*, defaults to `None`):
+            Top-p value for sampling. If set to `None`, the default value will be used.
```
qgallouedec, line 63: let's expose the default values from `SamplingParams` instead

nopepper: For top_p that works, but I think `GenerationConfig` uses different defaults (also `top_k = -1` by default for vLLM, which I don't think is valid for HF generation)

qgallouedec:
Defaults values

|                    | transformers       | vLLM           |
| ------------------ | ------------------ | -------------- |
| top_p              | 1.0                | 1.0            |
| top_k              | 50 (supports None) | -1 (means all) |
| min_p              | None               | 0.0            |
| repetition_penalty | 1.0                | 1.0            |

trl/trainer/grpo_config.py
qgallouedec: it should be good now

qgallouedec (approved):

PR #3001, opened 2025-03-03
✌️Remove double compute of sum in SFTTrainer

lexasub:
remove double compute of sum (sft_trainer.py) (optimize finetuning)


lexasub:
before
image

after 
image

qgallouedec (approved):

PR #3009, opened 2025-03-04
⚡ Pack 300 times faster, truncate 100 times faster

mariosasko:
# What does this PR do?

Adds fast packing/truncation logic that operates directly on PyArrow arrays to avoid expensive Python-to-PyArrow and PyArrow-to-Python conversions. This makes these steps almost instantaneous regardless of the input dataset's size. 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/data_utils.py
+++ b/trl/data_utils.py
@@ -536,14 +536,13 @@ def truncate_dataset(
     dataset: DatasetType, max_length: int, map_kwargs: Optional[dict[str, Any]] = None
 ) -> DatasetType:
     r"""
-    Truncate sequences in a dataset to a specifed maxium length.
+    Truncate sequences in a dataset to a specifed `max_length.
 
     Args:
         dataset (`Dataset` or `DatasetDict`):
             Dataset to truncate.
         seq_length (`int`):
-            Maximum sequence length to truncate to. All sequences longer than this will be cut
-            to this length.
+            Maximum sequence length to truncate to.
         map_kwargs (`dict` or `None`, *optional*, defaults to `None`):
             Additional keyword arguments to pass to the dataset's map method when truncating examples.
 
--- a/trl/data_utils.py
+++ b/trl/data_utils.py
@@ -511,11 +511,13 @@ def pack_dataset(dataset: DatasetType, seq_length: int, map_kwargs: Optional[dic
                 if pyarrow.types.is_list(column.type) or pyarrow.types.is_large_list(column.type):
                     if isinstance(column, pa.ChunkedArray):
                         column = column.combine_chunks()
-                    num_elements = len(column.values)
-                    dtype = column.offsets.type.to_pandas_dtype()  # np.int32 or np.int64
+                    offsets, values = column.offsets, column.values
+                    values = values[offsets[0].as_py() : offsets[-1].as_py()]
+                    num_elements = len(values)
+                    dtype = offsets.type.to_pandas_dtype()  # np.int32 or np.int64
                     offsets = np.arange(0, num_elements, seq_length, dtype=dtype)
                     offsets = np.concatenate((offsets, [num_elements]))
-                    column = type(column).from_arrays(offsets, column.values)
+                    column = type(column).from_arrays(offsets, values)
                 packed_columns.append(column)
             return pa.Table.from_arrays(packed_columns, names=examples.column_names)
 
--- a/trl/data_utils.py
+++ b/trl/data_utils.py
@@ -513,9 +513,8 @@ def pack_dataset(dataset: DatasetType, seq_length: int, map_kwargs: Optional[dic
                         column = column.combine_chunks()
                     num_elements = len(column.values)
                     dtype = column.offsets.type.to_pandas_dtype()  # np.int32 or np.int64
-                    offsets = np.arange(0, num_elements + 1, seq_length, dtype=dtype)
-                    if offsets[-1] != num_elements:
-                        offsets = np.concatenate((offsets, [num_elements]))
+                    offsets = np.arange(0, num_elements, seq_length, dtype=dtype)
+                    offsets = np.concatenate((offsets, [num_elements]))
                     column = type(column).from_arrays(offsets, column.values)
                 packed_columns.append(column)
             return pa.Table.from_arrays(packed_columns, names=examples.column_names)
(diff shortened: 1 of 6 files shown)


trl/data_utils.py
```diff
@@ -466,6 +471,132 @@ def pack_examples(examples: dict[str, list[list]], seq_length: int) -> dict[str,
...
+    >>> dataset = Dataset.from_dict(examples)
+    >>> packed_dataset = pack_dataset(dataset, seq_length=4)
+    >>> packed_dataset[:]
+    {'input_ids': [[1, 2, 3, 4], [5, 6, 7]],
+     'attention_mask': [[1, 1, 0, 1], [1, 1, 1]]}
+    ```
+    """
+    if map_kwargs is None:
+        map_kwargs = {}
+    if isinstance(dataset, Dataset):
```
lhoestq, line 506: I guess it should also work for DatasetDict ?

trl/data_utils.py
```diff
@@ -466,6 +471,132 @@ def pack_examples(examples: dict[str, list[list]], seq_length: int) -> dict[str,
...
+            for column in examples.columns:
+                if pyarrow.types.is_list(column.type) or pyarrow.types.is_large_list(column.type):
+                    if isinstance(column, pa.ChunkedArray):
+                        column = column.combine_chunks()
+                    num_elements = len(column.values)
+                    dtype = column.offsets.type.to_pandas_dtype()  # np.int32 or np.int64
+                    offsets = np.arange(0, num_elements + 1, seq_length, dtype=dtype)
+                    if offsets[-1] != num_elements:
+                        offsets = np.concatenate([offsets, [num_elements]])
+                    column = type(column).from_arrays(offsets, column.values)
```
lhoestq, line 519: Is there a `pyarrow.compute` function you can use here instead to simplify this ?

The functions from `pyarrow.compute` generally do a copy of the data, but since `.combine_chunks()` also copies the data it might not affect performance that much

trl/data_utils.py
```diff
@@ -466,6 +471,132 @@ def pack_examples(examples: dict[str, list[list]], seq_length: int) -> dict[str,
...
+    else:
+        dataset = dataset.map(
+            functools.partial(pack_examples, seq_length=seq_length),
+            batched=True,
+            **map_kwargs,
+        )
+    return dataset
+
+
+def truncate_dataset(
```
lhoestq, line 535: same comments for truncate

lhoestq (commented):
Hi Mario ! :) just added a few comments but overall lgtm

trl/data_utils.py
qgallouedec: No need to make it work with DatasetDict here, we apply a potentially different preprocessing depending on the split, see https://github.com/huggingface/trl/blob/e3244d2d096ff1e2e248c931d06d39e165e20623/trl/trainer/sft_trainer.py#L185-L200

trl/data_utils.py
mariosasko: The compute API is very limited, so I think this is the best we can do :)

The only way to avoid the copy would be by packing the individual `.chunks` (without crossing the boundaries), but this would then make the operation dependent on the underlying chunking and lead to more padding afterward, so I think the copy is justified here.

trl/data_utils.py
mariosasko: `pyarrow.compute.list_slice` reuses the `.values` buffers and only modifies the `.offsets`, so no unnecessary copies here.

trl/data_utils.py
mariosasko: Made the change to handle that, but I don't remember in what scenario we receive sliced arrays as input, so without a test case.

lhoestq (approved):
LGTM !

trl/data_utils.py
lhoestq: lgtm ! I think it can happen when training on a small portion of the train set, like `--dataset_train_split "train[:100]"`

qgallouedec:
## Benchmark packing

```python
import timeit
import numpy as np
from datasets import Dataset
from trl.data_utils import pack_examples, pack_dataset

# Create a larger dataset with sequence lengths following a gamma distribution
num_samples = 10_000

# Generate sequence lengths following a gamma distribution
seq_lengths = np.random.gamma(shape=5, scale=20, size=num_samples) # mean will be 100
seq_lengths = np.clip(seq_lengths, 10, None).astype(int)  # Clip to [1, max_seq_len]

# Generate input sequences with random lengths based on gamma distribution
examples = {
    "input_ids": [list(range(length)) for length in seq_lengths],
    "attention_mask": [[1] * length for length in seq_lengths],
}

max_length = 128  # Set a fixed packing length

# Benchmark pack_dataset
time_pack_dataset = timeit.timeit(
    lambda: pack_dataset(dataset, max_length), number=10
)

# Benchmark dataset.map with pack_examples
time_pack_examples = timeit.timeit(
    lambda: dataset.map(pack_examples, batched=True, fn_kwargs={"seq_length": max_length}), number=10
)

print(f"pack_dataset time: {time_pack_dataset:.4f} seconds")
print(f"dataset.map(pack_examples) time: {time_pack_examples:.4f} seconds")
```

```
pack_dataset time: 0.0667 seconds
dataset.map(pack_examples) time: 19.3734 seconds
Speedup: 290.46x
```

qgallouedec:
## Benchmark truncate

```python
import timeit
import numpy as np
from datasets import Dataset
from trl.data_utils import truncate_dataset


def truncate_examples(example, max_length):
    return {key: example[key][:max_length] for key in ["input_ids", "attention_mask"]}


# Create a larger dataset with sequence lengths following a gamma distribution
num_samples = 10_000

# Generate sequence lengths following a gamma distribution
seq_lengths = np.random.gamma(shape=5, scale=20, size=num_samples)  # mean will be 100
seq_lengths = np.clip(seq_lengths, 10, None).astype(int)  # Clip to [10, inf)

# Generate input sequences with random lengths based on gamma distribution
examples = {
    "input_ids": [list(range(length)) for length in seq_lengths],
    "attention_mask": [[1] * length for length in seq_lengths],
}

max_length = 128  # Set a fixed truncateing length

# Benchmark truncate_dataset
time_truncate_dataset = timeit.timeit(lambda: truncate_dataset(dataset, max_length), number=10)

# Benchmark dataset.map with truncate_examples
time_truncate_examples = timeit.timeit(
    lambda: dataset.map(truncate_examples, batched=True, fn_kwargs={"max_length": max_length}), number=10
)

print(f"truncate_dataset time: {time_truncate_dataset:.4f} seconds")
print(f"dataset.map(truncate_examples) time: {time_truncate_examples:.4f} seconds")
print(f"Speedup: {time_truncate_examples / time_truncate_dataset:.2f}x")
```

```
truncate_dataset time: 0.0611 seconds
dataset.map(truncate_examples) time: 6.3807 seconds
Speedup: 104.47x
```

qgallouedec (approved):
Thanks @mariosasko @lhoestq 🔥🔥

PR #3043, opened 2025-03-10
Fixing JSD loss computation in GKDTrainer as per definition

abhigoyal1997:
This PR fixes the JSD loss computation in GKDTrainer.

The issue was in following code:
https://github.com/huggingface/trl/blob/e3244d2d096ff1e2e248c931d06d39e165e20623/trl/trainer/gkd_trainer.py#L192-L204

This code snippet from the main branch computes the mixture distribution as per `beta*student + (1-beta)*teacher`; however, the definition is `beta*teacher + (1-beta)*student`. This small change fixes the loss equation, but for `beta = 0` and `beta = 1`, the loss becomes 0 (which is, by definition, correct). Therefore, for the correct behaviour of the loss, we need to handle these edge cases separately and compute KL(teacher || student) when `beta = 0` and KL(student || teacher) when `beta=1`.

This also needed a change in the JSD loss tests since the expected loss was calculated wrongly for these edge cases.

### DIFF
--- a/tests/test_gkd_trainer.py
+++ b/tests/test_gkd_trainer.py
@@ -147,14 +147,14 @@ class TestGeneralizedJSDLoss(unittest.TestCase):
         # Case 1: beta = 1 (should be equivalent to KL(student || teacher))
         loss_beta_1 = GKDTrainer.generalized_jsd_loss(student_logits, teacher_logits, beta=1)
         expected_loss_beta_1 = F.kl_div(
-            F.log_softmax(student_logits, dim=-1), F.softmax(teacher_logits, dim=-1), reduction="batchmean"
+            F.log_softmax(teacher_logits, dim=-1), F.softmax(student_logits, dim=-1), reduction="batchmean"
         )
         self.assertAlmostEqual(loss_beta_1.item(), expected_loss_beta_1.item(), places=5)
 
         # Case 2: beta = 0 (should be equivalent to KL(teacher || student))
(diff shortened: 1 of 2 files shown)


kashif:
thanks @abhigoyal1997 having a look now

kashif:
@abhigoyal1997 is the issue that the `beta` is not the same as the beta in paper? Also, note that the ` F.kl_div` takes inputs q and p to calculate KL(p||q) which can cause confusion too?

kashif:
Screenshot 2025-03-13 at 09 23 46

You can see when beta=0, the loss is the KL(student || teacher) which is `F.kl_div(teacher, student)` in TRL and when beta=1, the loss is KL(teacher || student) which is `F.kl_div(student, teacher)` in TRL so that seems fine to me.

kashif (approved):

abhigoyal1997:
> The paper has: Screenshot 2025-03-13 at 09 23 46
> 
> In TRL it is implemented as:
> 
> D J S D ( β ) ( P | Q ) = β K L ( P | β Q + ( 1 − β ) P ) + ( 1 − β ) K L ( Q | β Q + ( 1 − β ) P )
> 
> You can see when beta=0, the loss is the KL(student || teacher) which is `F.kl_div(teacher, student)` in TRL and when beta=1, the loss is KL(teacher || student) which is `F.kl_div(student, teacher)` so there is a difference in the original vs. the TRL formulation

Hi Kashif, yes this was the problem. The mixture distribution was calculated with the wrong weights.
Thanks for reviewing and approving!

Issue #3047, opened 2025-03-10
`sft_trainer` incompatible with `accelerator.gather_for_metrics`


jamesbraza:
### Reproduction

When using `trl.SFTTrainer` as of current `main` at https://github.com/huggingface/trl/tree/e3244d2d096ff1e2e248c931d06d39e165e20623, I get this error

```none
246 [rank0]:   File "/home/user/code/repo/scripts/sft_train.py", line 415, in main
247 [rank0]:     trainer.train()
248 [rank0]:   File "/home/user/code/repo/.venv/lib/python3.12/site-packages/transformers/trainer.py", line 2245, in train
249 [rank0]:     return inner_training_loop(
250 [rank0]:            ^^^^^^^^^^^^^^^^^^^^
251 [rank0]:   File "/home/user/code/repo/.venv/lib/python3.12/site-packages/transformers/trainer.py", line 2556, in _inner_training_loop
252 [rank0]:     tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
253 [rank0]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
254 [rank0]:   File "/home/user/code/repo/.venv/lib/python3.12/site-packages/transformers/trainer.py", line 3706, in training_step
255 [rank0]:     loss = self.compute_loss(model, inputs, num_items_in_batch=num_items_in_batch)
256 [rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
257 [rank0]:   File "/home/user/code/repo/.venv/lib/python3.12/site-packages/trl/trainer/sft_trainer.py", line 468, in compute_loss
258 [rank0]:     self._total_train_tokens += self.accelerator.gather_for_metrics(inputs["attention_mask"].sum().item())
259 [rank0]:                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
260 [rank0]:   File "/home/user/code/repo/.venv/lib/python3.12/site-packages/accelerate/accelerator.py", line 2581, in gather_for_metrics
261 [rank0]:     data = gather_object(input_data)
262 [rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3048, opened 2025-03-10
💠 Fixing `SFTTrainer.compute_loss` crash with `accelerate`

jamesbraza:
Closes https://github.com/huggingface/trl/issues/3047

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -464,9 +464,7 @@ class SFTTrainer(Trainer):
             model, inputs, return_outputs=True, num_items_in_batch=num_items_in_batch
         )
         if mode == "train":
-            self._total_train_tokens += (
-                self.accelerator.gather_for_metrics(inputs["attention_mask"].sum()).sum().item()
-            )
+            self._total_train_tokens += self.accelerator.gather_for_metrics(inputs["attention_mask"]).sum().item()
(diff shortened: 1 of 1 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -464,7 +464,9 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
             model, inputs, return_outputs=True, num_items_in_batch=num_items_in_batch
         )
         if mode == "train":
-            self._total_train_tokens += self.accelerator.gather_for_metrics(inputs["attention_mask"].sum().item())
+            self._total_train_tokens += (
+                self.accelerator.gather_for_metrics(inputs["attention_mask"].sum()).sum().item()
+            )
```
qgallouedec, line 469:
```suggestion
            self._total_train_tokens += self.accelerator.gather_for_metrics(inputs["attention_mask"]).sum().item()
```

A bit simpler

qgallouedec:
Thanks for fixing it. Can you just apply the suggestion?

trl/trainer/sft_trainer.py
jamesbraza: Nice thinking, applied

qgallouedec (approved):
Thanks!

PR #3070, opened 2025-03-13
💎 Gemma 3 SFT example on Codeforces dataset

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/examples/scripts/sft_gemma3.py
+++ b/examples/scripts/sft_gemma3.py
@@ -18,9 +18,11 @@ Train Gemma-3 on the Codeforces COTS dataset.
 accelerate launch --config_file examples/accelerate_configs/deepspeed_zero3.yaml examples/scripts/sft_gemma3.py
 """
 
-from trl import SFTConfig, SFTTrainer
 from datasets import load_dataset
 from transformers import AutoModelForImageTextToText
+
+from trl import SFTConfig, SFTTrainer
+
 
 def main():
     # Load dataset
@@ -54,6 +56,7 @@ def main():
 
     # Push to hub
     trainer.push_to_hub(dataset_name="open-r1/codeforces-cots")
+
 
 if __name__ == "__main__":
     main()
--- a/examples/scripts/sft_gemma3.py
+++ b/examples/scripts/sft_gemma3.py
@@ -12,11 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from trl import SFTConfig, SFTTrainer
-from datasets import load_dataset
-from transformers import AutoProcessor, AutoModelForImageTextToText
-from trl.trainer.utils import pad
 import torch
+from datasets import load_dataset
+from transformers import AutoModelForImageTextToText, AutoProcessor
+
+from trl import SFTConfig, SFTTrainer
+from trl.trainer.utils import pad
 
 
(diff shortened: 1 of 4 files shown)


burtenshaw:
@qgallouedec Amazing work. thanks! a few questions:

- Is padding like this necessary for all trainers?
- do we need to patch other tokens like this:

```python
processor.pad_token_id = processor.tokenizer.pad_token_id
processor.bos_token_id = processor.tokenizer.bos_token_id
processor.eos_token_id = processor.tokenizer.eos_token_id
```

- did you see how unsloth is dealing with missing token id's: https://huggingface.co/unsloth/gemma-3-4b-it/commit/90fe72f525abc73ff7283c23e6ceccea5d4273bb . Do you think we should open a PR for changes on the hub repo?

qgallouedec:
> Is padding like this necessary for all trainers?

usually tokenizers have a pad method. Here, the gemma processor doesn't. But maybe we shouldn't use the processor and directly load the tokenizer? Checking

examples/scripts/sft_gemma3.py
```diff
@@ -0,0 +1,67 @@
...
+
+if __name__ == "__main__":
+    # Load dataset
+    train_dataset = load_dataset("open-r1/codeforces-cots", split="train[:100]")
+    train_dataset = train_dataset.remove_columns("prompt")
+
+    # Load model and processor
+    model_id = "google/gemma-3-4b-it"
+    processor = AutoProcessor.from_pretrained(model_id)
+    model = AutoModelForImageTextToText.from_pretrained(model_id, attn_implementation="eager")
```
kashif, line 31:
```suggestion
    model = AutoModelForImageTextToText.from_pretrained(model_id, attn_implementation="sdpa")
```

qgallouedec: @kashif I get this warning:

```
It is strongly recommended to train Gemma3 models with the `eager` attention implementation instead of `sdpa`. Use `eager` with `AutoModelForCausalLM.from_pretrained('<path-to-checkpoint>', attn_implementation='eager')`.
```

trl/trainer/sft_trainer.py
```diff
@@ -422,7 +422,7 @@ def concat_prompt_completion(example):
                     map_kwargs["desc"] = f"Tokenizing {dataset_name} dataset"
 
                 def tokenize(example, processing_class, dataset_text_field):
-                    return processing_class(example[dataset_text_field])
+                    return processing_class(text=example[dataset_text_field])
```
kashif, line 425: ah yes good catch! i forgot this one

examples/scripts/sft_gemma3.py
qgallouedec: sdpa gives me OOM currently. Let's keep eager until sdpa is fixed?

kashif: ok...

trl/trainer/sft_trainer.py
kashif: yes I believe so... since the Gemma3 processor has images as the first argument and text as the 2nd and its being called without any arguments..

kashif (approved):

Issue #2917, opened 2025-02-20
GRPO from VLM models?


dipta007:
### Feature request

Is it possible to have GRPO for VLM models soon?
Or does anyone know any alternative library for that?
Thanks!

### Motivation

It will help to create multimodal R1 using the same GRPO trainer.

### Your contribution

N/A

Syazvinski:
+1

MohamedAliRashad:
@qgallouedec

JamesBowerXanda:
This would be really useful.

@dipta007 you could try out https://github.com/Deep-Agent/R1-V. They seem to have done something like this.

qgallouedec:
for the record, GRPO + VLM has been supported for quite some time now, sorry for the late reply!

Issue #2734, opened 2025-02-02
GRPO Trainer supports VLMs


sunildkumar:
### Feature request

The existing GRPO Trainer is pretty rigid in its structure, only accepting an untokenized dataset, and handling tokenization internally. This is problematic for VLMs that often need to do additional work to construct the input (e.g. `Qwen2VL` and `Qwen2.5VL` both require some additional work - generally in `collate_fn` - in inject image data). 

There's probably other aspects of the design that need to be updated as well, but I'm not sure yet? 

Thanks for your consideration!  

### Motivation

GRPO on VLMs would be very cool!

### Your contribution

I'm going to try modifying the TRL source code to support GRPO-ing a Qwen2.5VL, we'll see how that goes.

korbinian-hoermann:
Might be helpful: https://github.com/EvolvingLMMs-Lab/open-r1-multimodal?tab=readme-ov-file

sunildkumar:
Thanks @korbinian-hoermann! That's a great starting point. The lack of strong positive results makes me a bit concerned though, but that might just be an artifact of their data - it seems like they are tough problems? I'm working on trying to get it working on a more constrained and contrived problem, which might give us some more signal on this. Anyways, thanks again!

sunildkumar:
It has been a while, but I just released a package that makes it possible to train visual reasoning models on whatever task you want: https://github.com/groundlight/r1_vlm. I'm currently thinking about if and how to integrate back into the TRL repo. Our code is pretty opinionated, which might make it hard to integrate. But it works pretty well :)

qgallouedec:
Hey @sunildkumar, sorry for the very slow reply here; we're catching up on the backlog.

Good news: this is now supported out of the box. `GRPOTrainer` auto-detects a `ProcessorMixin` and switches to a VLM path; you just add an `image`/`images` column to your dataset and image-to-tensor conversion is handled via the model's image processor. Models like Qwen2.5-VL, Gemma-3, SmolVLM2 and LLaVA-Next are supported. See the "Vision-Language Model (VLM) Training" section of the GRPO docs https://huggingface.co/docs/trl/main/en/grpo_trainer#vision-language-model-vlm-training and the ready-to-run example at https://github.com/huggingface/trl/blob/e1b2e21975994e676d9a676e3ab005e9bee38e40/examples/scripts/grpo_vlm.py. 

Really cool to see your r1_vlm package too! thanks for pushing on this early.

PR #3072, opened 2025-03-13
[GRPO] add vlm training capabilities to the trainer

CompN3rd:
# What does this PR do?

This is an attempt at addressing #2917 . 
An associated unittest has been added and less "toy-examply" trainings seem to maximize rewards as well, but I don't have final models yet.

Issues regarding 4-bit quantization and applying liger kernel to vlm models still remain (marked with TODO in comments) and I would appreciate input on how to tackle/further debug them.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/grpo_trainer.py
```diff
@@ -363,18 +374,16 @@ def __init__(
...
                 # The reward model computes the reward for the latest non-padded token in the input sequence.
                 # So it's important to set the pad token ID to the padding token ID of the processing class.
                 reward_func.config.pad_token_id = reward_processing_class.pad_token_id
                 reward_processing_classes[i] = reward_processing_class
         self.reward_processing_classes = reward_processing_classes
 
-        # Data collator
-        def data_collator(features):  # No data collation is needed in GRPO
-            return features
-
```
qgallouedec, line 377: is there a reason for deleting it?

trl/trainer/grpo_trainer.py
```diff
@@ -571,13 +600,28 @@ def new_group_context():
...
+        for attribute in attributes:
+            try:
+                obj = getattr(obj, attribute)
+            except AttributeError:
+                return default
+        return obj
+
+    @staticmethod
+    def _get_from_processor_or_tokenizer(processor, key):
+        return getattr(processor, key, None) or getattr(getattr(processor, "tokenizer", {}), key, None)
```
qgallouedec, line 616: how is it different from

```python
if hasattr(processor_class, "tokenizer"):
    pad_token_id = processor_class.tokenizer.pad_token_id
```?

**Comment by qgallouedec** [comment]
Thanks! Have you tried to fine-tune a VLM with the trainer? Do you have results to share?

File: trl/trainer/grpo_trainer.py (hunk 2 above)
Comment (CompN3rd) [reply trl/trainer/grpo_trainer.py:616]: It's not really. The iterated `getattr` is simply to not throw an Attribute error when the queried id is neither in the processor or tokenizer member, but just return `None`.
I found it to be marginally more readable to compare against `None` than surrounding the callsites with try except.

As to why not directly writing `.pad_token_id`, because different keys are queried in different places like `eos_token`

File: trl/trainer/grpo_trainer.py (hunk 1 above)
Comment (CompN3rd) [reply trl/trainer/grpo_trainer.py:377]: Ah, yes I made that an argument of `__init__` again, because I am not sure if there is an agreed upon contract about the layout of columns in multi-modal datasets, so I had the feeling that custom collation might be necessary to e.g. convert video frames to images.

Maybe it's not really needed, since many transformations are probably possible with an a priori mapping of a dataset, but then again, I had the feeling that this choice might be best left to the library user.


As a side-note: How to generally support more than image data; like processors, that take `audios=...` or `videos=...` or whatever else might be embedded and passed to a model without hardcoding many fields or overwriting basically the whole `_generate_and_score_completions` is a big question and holdup for "true" multimodality

**Comment by CompN3rd** [comment]
Well, actual fine tuning is still in progress and riddled with oom issues and the quantization bug referenced in the unittest as well  and the full training script relies on a private dataset, but I can at least give a bit more information.

The training task I am currently looking at is fine-tuning a vlm (the language model part) doing image captioning to maximize a clip cosine similarity score.

So the reward module looks like this
```
@dataclass
class CLIPRewardModelOutput(ModelOutput):
    logits: torch.FloatTensor
    """The reward logits for the Trainer."""


class CLIPRewardModel(CLIPModel):
    """Inherits from CLIPModel (i.e. PreTrainedModel), such that the forward computation gives the rl reward,
    but type-based training logic (accelerator.prepare) is still possible"""

    def forward(self, *, input_ids, attention_mask, pixel_values) -> torch.Tensor:
        """Mainly copy-paste from CLIPModel.forward up to the logit and loss computation"""

        vision_outputs = self.vision_model(
            pixel_values=pixel_values,
            output_attentions=False,
            output_hidden_states=False,
            interpolate_pos_encoding=False,
            return_dict=True,
        )

        text_outputs = self.text_model(
            input_ids=input_ids,
            position_ids=None,
            output_attentions=False,
            output_hidden_states=False,
            return_dict=True,
        )

        image_embeds = vision_outputs[1]
        image_embeds = self.visual_projection(image_embeds)

        text_embeds = text_outputs[1]
        text_embeds = self.text_projection(text_embeds)

        # normalized features
        image_embeds = image_embeds / _get_vector_norm(image_embeds)
        text_embeds = text_embeds / _get_vector_norm(text_embeds)

        # cosine similarity of the vectors as reward logits
        return CLIPRewardModelOutput(logits=cosine_similarity(image_embeds, text_embeds, dim=-1).unsqueeze(-1))```

CompN3rd:
Then we have a small modification to the trainer via subclassing (which is why I proposed to split off the relevant code section into it's own member function)

```
class GRPOVlmClipTrainer(GRPOTrainer):
    def _prepare_inputs_for_reward_module(
        self,
        *,
        inputs: dict[str, torch.Tensor | Any],
        reward_processing_class: PreTrainedTokenizerBase,
        prompts: list[str],
        completions: list[str],
        images=None,
    ) -> dict[str, torch.Tensor | Any]:
        # disregard prompts, only prepare completions (captions) and images
        reward_inputs = reward_processing_class(
            images=images,
            text=completions,
            return_tensors="pt",
            padding=True,
            padding_side="right",
            add_special_tokens=True,
            truncation=True,
            max_length=77,
        )
        reward_inputs = super(GRPOTrainer, self)._prepare_inputs(reward_inputs)

        return reward_inputs
```

CompN3rd:
Finally this leads to reward curves like this, which seem to indicate that it generally optimizes in the right direction.

W B Chart 14 3 2025, 09_31_05

Do note however that there still was a bug in reward input preparation, which is hopefully, why the curves seem so unsteady, but new ones will take some more hours to generate still.

MohamedAliRashad:
@CompN3rd If you can give me a simple guide on how to use your PR i can help you with testing

CompN3rd:
> @CompN3rd If you can give me a simple guide on how to use your PR i can help you with testing

Sure, if you want to get started with a semi-realistic example, I'd suggest starting with the setup from the unittest, which should be able to run on a 24Gb gpu (`test_gpo_trainer.py` l.900-987)

```python
@require_flash_attn
    @require_bitsandbytes
    @require_peft
    @require_torch_accelerator
    def test_vlm_training(self):
        model_name = "HuggingFaceTB/SmolVLM-Instruct"
        .....
``` 

Biggest question there is why 8 bit quantization works, but 4 bit quantization breaks the test (or whether that is somehow expected behavior), so any input in that regard would be valuable.

Other than that if you have access to more vram gpus you could rewrite the test to work without quantization or replace the model with an even smaller one.

CompN3rd:
@MohamedAliRashad If I understand this correctly, it is probably  because the `processor` in your case returns already tokenized `input_ids` and probably `pixel_values` or whatever fields are associated with image/video processing.

That is not the type of data the `GRPOTrainer` expects (even before this current pr). In the internal preprocessing function of the trainer, it accesses `prompt` of the input dictionary (and this pr adds `image`, which is expected to be a raw numpy or pil image, not a base64 string).
Then it internally calls the processor and goes from there.

Tldr.: Your data preprocessing probably interferes with the input data preparation done in the trainer class.

nph4rd:
@CompN3rd - so how would one preprocess the data or tell the trainer how to process it? For example, as far as I understand, Qwen2.5-VL uses qwen-vl-util's `process_vision_info`. 

Based on your changes, what would be the best approach to use that during the input preparation?

MohamedAliRashad:
@CompN3rd I changed the preprocessing to be closer to what you have in the test file and it worked wanderfully.
I made full finetuning for qwen 2.5 vl 3B and it worked on an 80 GB GPU

nph4rd:
@MohamedAliRashad - do you mind sharing the setup/coda you used for that?

nph4rd:
I just tested the following with [this dummy dataset](agentsea/vqa-test-formatted) using 4 A100 80GB:

```python
import torch
from transformers import Qwen2_5_VLForConditionalGeneration, AutoProcessor
from qwen_vl_utils import process_vision_info
import copy

model_id = "Qwen/Qwen2.5-VL-3B-Instruct"
model = Qwen2_5_VLForConditionalGeneration.from_pretrained(
    model_id,
    torch_dtype=torch.bfloat16,
    attn_implementation="flash_attention_2",
    use_cache=False,
)
processor = AutoProcessor.from_pretrained(model_id, padding_side="left")

dataset = load_dataset("agentsea/vqa-test-formatted", split="train")
dataset = dataset.remove_columns(["completion"])

def preprocess_vision_info(examples):
    examples_copy = copy.deepcopy(examples)
    batch_size = len(examples["prompt"])
    examples["image"] = []
    for i in range(batch_size):
        prompt_data = examples_copy["prompt"][i]
        image_data = examples_copy["image"][i]
        for message in prompt_data:
            for content in message["content"]:
                if isinstance(content, dict) and content.get("type") == "image":
                    content["image"] = image_data
        processed_images, _ = process_vision_info(prompt_data)
        examples["image"].extend(processed_images)
    return examples

dataset = dataset.with_transform(preprocess_vision_info)


training_args = GRPOConfig(
    output_dir="Qwen2.5-VL-3B-GRPO",
    logging_steps=1,
    use_vllm=True,
    bf16=True,
    gradient_checkpointing=True,
    num_generations=3,
    max_prompt_length=None,
    vllm_device="cuda:3",
)


trainer = GRPOTrainer(
    model=model,
    processing_class=processor,
    reward_funcs=reward_len,
    args=training_args,
    train_dataset=dataset,
)
trainer.train()
```

However I'm encountering this error:

```
ValueError: Attempted to assign 5185 + 5185 + 5185 = 15555 multimodal tokens to 31107 placeholders
```

Upon further inspection I found that the code works if I make the following change in [this line specifically](https://github.com/CompN3rd/trl/blob/328ef463a776a00d02decc8bf7e5f8cfbe215c03/trl/trainer/grpo_trainer.py#L823
):

from:

```python
            prompt_inputs = self.processing_class(
                text=prompts_text,
                images=images,
                return_tensors="pt",
                padding=True,
                padding_side="left",
                add_special_tokens=False,
            )
```

to:

```python
            prompt_inputs = self.processing_class(
                text=prompts_text.copy(), # send a copy instead
                images=images,
                return_tensors="pt",
                padding=True,
                padding_side="left",
                add_special_tokens=False,
            )
```

What is happening is that the processor class is mutating the input [here](https://github.com/huggingface/transformers/blob/42c489f2ae738a3b690bb90aab274f02ff024795/src/transformers/models/qwen2_5_vl/processing_qwen2_5_vl.py#L156C21-L156C25). So, vLLM complains because it's receiving the modified  `prompts_text`. You can test the side-effect in [this script](https://gist.github.com/nph4rd/f003323ac4c8940f779f44a24b815ff7).

I don't think this is an issue that should be handled by neither TRL nor vLLM. I think it should be handled at the source of the proccessor's code.

I can see that the [SmolVLM processor class](https://github.com/huggingface/transformers/blob/main/src/transformers/models/smolvlm/processing_smolvlm.py) doesn't have this kind of side-effect. But Qwen2.5-VL's does, so I do wonder how @MohamedAliRashad made it work. I presume it's because `prompts_text` is a string and not a list when using 1 GPU with `num_generations=1`?

MohamedAliRashad:
@nph4rd The error you are seeing is because of your context size limit.
Qwen (unlike other models) doesn't give a fixed number of tokens for images of different shapes, The number of tokens change based on the size of the input image, if i am not mistaken every `28x28` pixels are one token for them.

What you need to do is to resize your images to be in a smaller size than your acceptable context window and also i didn't use `process_vision_info` and it worked fine with me, so you may consider removing it and send the pil images as it is.

CompN3rd:
> @nph4rd The error you are seeing is because of your context size limit. Qwen (unlike other models) doesn't give a fixed number of tokens for images of different shapes, The number of tokens change based on the size of the input image, if i am not mistaken every `28x28` pixels are one token for them.
> 
> What you need to do is to resize your images to be in a smaller size than your acceptable context window and also i didn't use `process_vision_info` and it worked fine with me, so you may consider removing it and send the pil images as it is.

Can second that observation, also note this option in `GRPOConfig`, which has a default value of 512, which for my tested image resolutions was way too small:


```python            
training_args = GRPOConfig(
                ...,
                max_prompt_length=None,  # needs the full length to not cut off image tokens
                )
```

CompN3rd:
@qgallouedec Let me know if there are refactoring or api changes necessary to make this ready for merging. Would be happy to make those adjustments.

CompN3rd:
@nph4rd Thanks for testing it out. I concur with @MohamedAliRashad observations, that I could produce such errors mostly by having too small of a context window.

As for the `log_completions` error, I had a version, where not all processes participated in the communication, which obviously failed. 
As of yet, I have a test with 2gpus locally, which worked well as well as a cloud test, but that was only one A40 GPU. 
Both produced images in weights and biases, but I admit, I haven't made multi-node tests.

trl/trainer/grpo_trainer.py
```diff
@@ -699,11 +803,19 @@ def _generate_and_score_completions(
...
                 ordered_set_of_prompts = all_prompts_text[:: self.num_generations]
+                ordered_set_of_images = all_images[:: self.num_generations] if all_images is not None else None
+                # cf. https://docs.vllm.ai/en/stable/serving/multimodal_inputs.html#image
+                if ordered_set_of_images is not None:
+                    ordered_set_of_prompts = [
+                        {"multi_modal_data": {"image": image}, "prompt": prompt}
+                        for image, prompt in zip(ordered_set_of_images, ordered_set_of_prompts)
+                    ]
                 with profiling_context(self, "vLLM.generate"):
                     completion_ids = self.vllm_client.generate(
```
nph4rd, line 820: The changes in https://github.com/huggingface/trl/commit/f713f614e9df6b1f581c38724cb6b92da8edac62 imply that it's now not possible to send the multi_modal_data without adapting the new vllm server. I think the following would have to change:

- https://github.com/huggingface/trl/blob/26d86757a7c7e24e397ea44f57ecce6031dfac01/trl/extras/vllm_client.py#L132
- https://github.com/huggingface/trl/blob/26d86757a7c7e24e397ea44f57ecce6031dfac01/trl/scripts/vllm_serve.py#L316

trl/trainer/grpo_trainer.py
```diff
@@ -674,22 +721,79 @@ def _prepare_inputs(self, inputs: dict[str, Union[torch.Tensor, Any]]) -> dict[s
...
+        reward_inputs = super()._prepare_inputs(reward_inputs)
+
+        return reward_inputs
+
     def _generate_and_score_completions(
         self, inputs: dict[str, Union[torch.Tensor, Any]]
     ) -> dict[str, Union[torch.Tensor, Any]]:
         device = self.accelerator.device
         prompts = [x["prompt"] for x in inputs]
+        images = [x["image"] for x in inputs if "image" in x]
```
nph4rd, line 764: I noticed the DPO trainer already supports VLMs but expects the column to be called "images", as a list of PIL images:

https://github.com/huggingface/trl/blob/26d86757a7c7e24e397ea44f57ecce6031dfac01/trl/trainer/dpo_trainer.py#L642

I think this should be "images" too, for consistency, and to support the case of multiple images per prompt.

Also, note that for that case the vLLM server would have to be able to receive the `limit_mm_per_prompt` argument.

nph4rd (commented):
@CompN3rd - left some comments regarding supporting multiple images per prompt and updating to the changes in https://github.com/huggingface/trl/commit/f713f614e9df6b1f581c38724cb6b92da8edac62

Happy to help, with changes and/or testing.

sunildkumar:
Eagerly awaiting this (https://github.com/huggingface/trl/issues/2734 - 2 months and counting)! @CompN3rd  Let me know if  and how I can help. I've been training VLMs with GRPO for a while now, just not on TRL `main`.

sunildkumar:
@qgallouedec – I hope you don’t mind the tag. It’s been a couple of weeks since @CompN3rd has engaged with this PR, and I really appreciate the work that’s been done so far. I’d love to help move it forward if that’s appropriate.

I’m not entirely sure what the etiquette is in cases like this—would it be okay to open a follow-up PR branching off of this one, or would you recommend waiting longer? Apologies if this is a naive question, and thank you in advance for any guidance.

qgallouedec:
Thanks again for your work on this, and sorry for the slow response, be sure we're doing our best. It's a valuable feature and makes a lot of sense to include. That said, it requires thorough review, testing, and documentation before merging, and at the moment we don’t have the capacity to give it the attention it needs. I’ll make sure to revisit it as soon as I can.

In the meantime, keeping the PR open is a great idea. It allows the community to test it, report any issues, and benefit from the feature.

And to your question — yes, feel free to open a follow-up PR based on this one. That’s totally fine and actually very helpful. No need to wait.

sunildkumar:
@qgallouedec - totally understood. Thanks for your advice!

Benjoyo:
Anyone actively working on GRPO support for VLMs still? 🙏

chaodreaming:
How long will it take to support grpo?

chaodreaming:
About how long will it take to be able to support grpo, a lot of people are very excited about this, thank you very much for your contribution!

mccatec:
image
Noticed this🙏🏻️

chaodreaming:
> 图像
> 从 的 x 🙏 中注意到这一点

Where did you see this news, hopefully give me a link, I'll be keeping an eye on the progress

mccatec:
> > 图像
> > 从 的 x 🙏 中注意到这一点
> 
> Where did you see this news, hopefully give me a link, I'll be keeping an eye on the progress

https://x.com/QGallouedec/status/1919806234821026141

nph4rd:
hey in case you want to try grpo+vlm while it's not supported here, i wrote this up based on TRL's code and @CompN3rd's PR

https://github.com/nph4rd/grpo_vlm

chaodreaming:
> 嘿，如果您想尝试 GRPO+VLM 而这里不支持它，我根据 TRL 的代码和 PR 编写了这个
> 
> https://github.com/nph4rd/grpo_vlm

You should mention a pr to become a contributor, it's a good resume!

sergiopaniego:
Hi! 👋 

I've created an [example notebook](https://colab.research.google.com/drive/1xr_3vEFfuCg2qg9OCrylIsweDh-MnF5r?usp=sharing) with the current PR's content for fine tuning a VLM using GRPO. It seems to be working, so maybe it's time to revive this PR, update it, and consider merging it soon!

docs/source/grpo_trainer.md
```diff
@@ -529,6 +529,144 @@ and the reward will be computed as the sum of the rewards from each function, or
...
+- **Memory Usage**: VLMs require more memory due to image processing. Use quantization and LoRA for large models
+- **vLLM Support**: VLM models work with vLLM v0.6.1+ in both server and colocate modes
+- **Batch Size**: Start with smaller batch sizes due to increased memory requirements from images
+- **Generation Length**: Consider longer max completion lengths for detailed image descriptions
+
+### Dataset Format
+
+VLM datasets should include:
+- **`prompt`**: Text prompt formatted through the processor's chat template
+- **`image`**: PIL Image or numpy array representing the image
```
sergiopaniego: We could add some details about the image support since currently only 1 image is supported per prompt

```suggestion
- **`image`**: PIL Image or numpy array representing the image (only one image per prompt is supported)
```

trl/trainer/grpo_trainer.py
```diff
@@ -876,12 +1037,24 @@ def _get_per_token_logps_and_entropies(
...
+            logits = model(**model_inputs).logits
+
+            # For VLMs that don't support logits_to_keep, we need to slice the logits manually
+            if "logits_to_keep" not in self.model_kwarg_keys:
+                # Keep only the last logits_to_keep + 1 logits for efficiency
+                logits = logits[:, -(logits_to_keep + 1) :, :]
+
             logits = logits[:, :-1, :]  # (B, L-1, V), exclude the last logit: it corresponds to the next token pred
             # Divide logits by sampling temperature.
             # See https://huggingface.co/blog/the_n_implementation_details_of_rlhf_with_ppo#policy-training-implementation-details
```
avishaiElmakies: Hi @kashif.
I believe there is a bug here. from what i understand to calculate logps with a multi-modal model, you need to pass the modality as well, which this implementation seems to miss this. 
I explain more about this bug in the other PR: https://github.com/huggingface/trl/pull/3460#issuecomment-2953991156.

avishaiElmakies (commented):
Hi @kashif. 
The PR looks good. I do think there is a small bug that needs to be fixed.
I think that it would also be great if this PR could work with other modalities as well e.g. audio and videos. The changes shouldn't be too hard. Some of them were already made in the other PR. maybe @hellopahe can help with this.

docs/source/grpo_trainer.md
```diff
@@ -529,6 +529,144 @@ and the reward will be computed as the sum of the rewards from each function, or
...
+    use_vllm=True,  # Enable vLLM acceleration
+    vllm_mode="colocate",  # Use colocate mode
+    vllm_gpu_memory_utilization=0.3,
+    report_to="wandb",
+)
+
+# Create trainer
+trainer = GRPOTrainer(
+    model=model,
+    processing_class=processor,  # VLM processor
```
qgallouedec: Is it needed?

docs/source/grpo_trainer.md
```diff
@@ -529,6 +529,144 @@ and the reward will be computed as the sum of the rewards from each function, or
...
+
+# Start training
+trainer.train()
+```
+
+### VLM-Specific Configuration
+
+When training VLMs, consider these configuration options:
+
+- **`max_prompt_length=None`**: Set to `None` to avoid truncating image tokens
```
qgallouedec: Can we think of a way around this?

docs/source/grpo_trainer.md
```diff
@@ -529,6 +529,144 @@ and the reward will be computed as the sum of the rewards from each function, or
...
+- **Memory Optimization**: Supports 4-bit quantization with LoRA for memory-efficient VLM training
+- **Enhanced Data Collation**: Proper handling of VLM-specific data structures including `pixel_values`, `pixel_attention_mask`, and `image_sizes`
+
+### Supported VLM Models
+
+The implementation has been tested with popular VLM architectures including:
+- **Qwen2-VL**: `Qwen/Qwen2-VL-2B-Instruct`
+- **SmolVLM**: `HuggingFaceTB/SmolVLM-Instruct`
+
+### Example Usage
```
qgallouedec: A link to the example in the repo is probably enough?

kashif:
would you like to edit the function here as a "suggestion" and then i can commit it here directly and you will have a commit here

tests/test_grpo_trainer.py
```diff
@@ -11,20 +11,35 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import os
```
```suggestion

import os
```

tests/test_grpo_trainer.py
```diff
@@ -758,24 +772,57 @@ def test_training_vllm(self):
...
-                train_dataset=dataset,
-            )
-
-            previous_trainable_params = {n: param.clone() for n, param in trainer.model.named_parameters()}
-
-            trainer.train()
 
-            self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
+            # Set up environment for vLLM
+            import os
```
qgallouedec: Already imported

tests/test_grpo_trainer.py
```diff
@@ -1297,3 +1440,591 @@ def reward_func(completions, **kwargs):
...
+                yield {
+                    "prompt": prompt,
+                    "image": np.random.uniform(low=0.0, high=255.0, size=(64, 64, 3)).astype(
+                        np.uint8
+                    ),  # Much smaller images
+                }
+
+        dataset = Dataset.from_generator(
+            data_gen, gen_kwargs={"num_samples": 4}, features=Features(image=Image(), prompt=Value(dtype="string"))
+        )
```
qgallouedec: It could be nice to have this in trl-internal-testing instead

qgallouedec (commented):
Some comments, good work so far!

avishaiElmakies:
> would you like to edit the function here as a "suggestion" and then i can commit it here directly and you will have a commit here

Unfortunately, this change is bigger than just a function change since you need to propagate the relevant data through the trainer and make sure it is being used during every use of this function. While it is not hard. it does take some time to make sure everything is ok.

docs/source/grpo_trainer.md
sergiopaniego: I agree 😄

docs/source/grpo_trainer.md
sergiopaniego: Currently, it is. Otherwise, it'd try to init the tokenizer, which will cause errors.

https://github.com/huggingface/trl/blob/97085539a3c8078f5e378cf5fdfc2a1603c5f82f/trl/trainer/grpo_trainer.py#L457

Maybe could be updated with something similar as in the `dpo_trainer.py` with `self.is_vision_model`

kashif:
@ avishaiElmakies i have fixed the log-prob calculations that now take the visual context into account

kashif:
@avishaiElmakies I have now fixed the log-prob calculation so that it takes the visual context into account

avishaiElmakies:
@kashif thanks! Looks good!
Is there a plan to add other modalities to GRPO as well?

sergiopaniego:
> @kashif thanks! Looks good! Is there a plan to add other modalities to GRPO as well?

Sure!! But we can keep only vision for this one and then add the other modalities in another one (even the [one](https://github.com/huggingface/trl/pull/3460) already created). This way, we can merge sooner!

avishaiElmakies:
Ok, thanks!

sergiopaniego:
A working example with the latest iteration for the review :)

[PR recipe](https://github.com/huggingface/cookbook/pull/312) in Cookbook  || To see the notebook directly [here](https://app.reviewnb.com/huggingface/cookbook/pull/312/) 

Fine tuned model in the notebook: [sergiopaniego/Qwen2.5-VL-3B-Instruct-Thinking](https://huggingface.co/sergiopaniego/Qwen2.5-VL-3B-Instruct-Thinking) || [Tensorboard](https://huggingface.co/sergiopaniego/Qwen2.5-VL-3B-Instruct-Thinking/tensorboard)

docs/source/grpo_trainer.md
```diff
@@ -524,6 +524,68 @@ and the reward will be computed as the sum of the rewards from each function, or
...
+
+```bash
+accelerate launch
+    --config_file=examples/accelerate_configs/deepspeed_zero3.yaml \
+    examples/scripts/grpo_vlm.py \
+    --dataset_name lmms-lab/multimodal-open-r1-8k-verified \
+    --model_name_or_path Qwen/Qwen2-VL-2B-Instruct \
+    --per_device_train_batch_size 8 \
+    --output_dir grpo-Qwen2-VL-2B-Instruct \
+    --bf16 true \
```
qgallouedec:
```suggestion
```

default value

docs/source/grpo_trainer.md
```diff
@@ -524,6 +524,68 @@ and the reward will be computed as the sum of the rewards from each function, or
...
+
+We have an example script [examples/scripts/grpo_vlm.py](https://github.com/huggingface/trl/blob/main/examples/scripts/grpo_vlm.py) that you can use to train a VLM model. For example, to train a VLM model on the [multimodal-open-r1-8k-verified](https://huggingface.co/datasets/lmms-lab/multimodal-open-r1-8k-verified) dataset, you can run the following command:
+
+```bash
+accelerate launch
+    --config_file=examples/accelerate_configs/deepspeed_zero3.yaml \
+    examples/scripts/grpo_vlm.py \
+    --dataset_name lmms-lab/multimodal-open-r1-8k-verified \
+    --model_name_or_path Qwen/Qwen2-VL-2B-Instruct \
+    --per_device_train_batch_size 8 \
```
qgallouedec:
```suggestion
```

default value

trl/trainer/grpo_trainer.py
```diff
@@ -221,14 +227,47 @@ def split_tensor_dict(
...
-    return [
-        {
-            key: tensor[i * chunk_size : (i + 1) * chunk_size] if tensor is not None else None
-            for key, tensor in tensor_dict.items()
-        }
-        for i in range(num_chunks)
-    ]
+
+    # Get batch size from first tensor, handling BatchFeature objects
+    if hasattr(first_tensor, "data") and isinstance(first_tensor.data, dict):
```
qgallouedec: can we use `isinstance(...BatchFeature)` here? more explicit

trl/trainer/grpo_trainer.py
```diff
@@ -395,14 +465,125 @@ def reward_func(completions, **kwargs):
...
+    @staticmethod
+    def _get_nested_attr(obj, attr, default=None):
+        """Retrieve nested attribute from an object."""
+        attributes = attr.split(".")
+        for attribute in attributes:
+            try:
+                obj = getattr(obj, attribute)
+            except AttributeError:
+                return default
+        return obj
```
qgallouedec: can you make this a function instead of a static private method? It would make things easier if at some point we want to move it to utils.
And add a small test.

Also, more concise:

```python
def get_nested_attr(obj, attr, default=None):
    try:
        return reduce(getattr, attr.split("."), obj)
    except AttributeError:
        return default
```

trl/trainer/grpo_trainer.py
```diff
@@ -395,14 +465,125 @@ def reward_func(completions, **kwargs):
...
+
+    @staticmethod
+    def _get_from_processor_or_tokenizer(processor, key):
+        """Get an attribute from processor or its tokenizer (if it has one)."""
+        if hasattr(processor, key):
+            return getattr(processor, key)
```
```suggestion
            return processor.key
```

trl/trainer/grpo_trainer.py
```diff
@@ -395,14 +465,125 @@ def reward_func(completions, **kwargs):
...
+        elif hasattr(processor, "tokenizer") and hasattr(processor.tokenizer, key):
+            return getattr(processor.tokenizer, key)
```
```suggestion
            return processor.tokenizer.key
```

trl/trainer/grpo_trainer.py
```diff
@@ -395,14 +465,125 @@ def reward_func(completions, **kwargs):
...
+        attributes = attr.split(".")
+        for attribute in attributes:
+            try:
+                obj = getattr(obj, attribute)
+            except AttributeError:
+                return default
+        return obj
+
+    @staticmethod
+    def _get_from_processor_or_tokenizer(processor, key):
```
qgallouedec: Same here, I'd turn this into a function, easier if in the future we want to move it to utils and use it elsewhere

trl/trainer/grpo_trainer.py
```diff
@@ -453,10 +642,18 @@ def __init__(
...
+        self.is_vision_model = model.config.model_type in MODEL_FOR_VISION_2_SEQ_MAPPING_NAMES.keys()
         if processing_class is None:
-            processing_class = AutoTokenizer.from_pretrained(model.config._name_or_path, padding_side="left")
-        if processing_class.pad_token is None:
-            processing_class.pad_token = processing_class.eos_token
+            if self.is_vision_model:
+                processing_class = AutoProcessor.from_pretrained(model.config._name_or_path, padding_side="left")
+            else:
+                processing_class = AutoTokenizer.from_pretrained(model.config._name_or_path, padding_side="left")
+
```
qgallouedec: duplicated line

trl/trainer/grpo_trainer.py
```diff
@@ -453,10 +642,18 @@ def __init__(
...
+        # Handle pad token for processors or tokenizers
+        pad_token = self._get_from_processor_or_tokenizer(processing_class, "pad_token")
+        if pad_token is None:
+            eos_token = self._get_from_processor_or_tokenizer(processing_class, "eos_token")
+            processing_class.pad_token = eos_token
```
qgallouedec: I'm wondering if for image processor, this should be 


```python 
processing_class.tokenizer.pad_token = eos_token
```

instead

trl/trainer/grpo_trainer.py
```diff
@@ -497,11 +694,18 @@ def __init__(
             if isinstance(reward_func, PreTrainedModel):
                 if reward_processing_class is None:
                     reward_processing_class = AutoTokenizer.from_pretrained(reward_func.config._name_or_path)
-                if reward_processing_class.pad_token_id is None:
-                    reward_processing_class.pad_token = reward_processing_class.eos_token
+
+                # Handle pad token for reward processing class
+                rc_pad_token_id = self._get_from_processor_or_tokenizer(reward_processing_class, "pad_token_id")
```
qgallouedec: do we want to support _visual reward models_? Does it even exist?

trl/trainer/grpo_trainer.py
```diff
@@ -748,13 +952,43 @@ def __init__(
                         reward_func, evaluation_mode=True, device_placement=True
                     )
 
+        # Ensure signature columns are set for VLM support and custom reward functions
+        self._set_signature_columns_if_needed()
+
```
qgallouedec: Are you sure that this is needed? I think it's called directly in the Trainer when needed

trl/trainer/grpo_trainer.py
```diff
@@ -748,13 +952,43 @@ def __init__(
...
         if self._signature_columns is None:
-            self._signature_columns = ["prompt"]
+            # Base signature columns
+            signature_columns = ["prompt"]
+
+            # Check if this is a VLM setup (either VLM model OR VLM processor)
+            is_vlm_processor = hasattr(self.processing_class, "tokenizer") and hasattr(
+                self.processing_class, "image_processor"
+            )
+            if self.is_vision_model or is_vlm_processor:
```
qgallouedec: can they be different?

trl/trainer/grpo_trainer.py
kashif: we do not need it so i am removing it

trl/trainer/grpo_trainer.py
kashif: no one is `AutoProcessor` and the other `AutoTokenizer`

tests/test_grpo_trainer.py
```diff
@@ -127,95 +200,61 @@ def test_sampler_no_shuffle(self):
...
         dataset = ["a", "b", "c", "d", "e", "f", "g"]
         sampler = RepeatSampler(dataset, mini_repeat_count=1, batch_size=2, repeat_count=2)
-        # Should output something like [4, 3, 4, 3, 0, 1, 0, 1, 2, 6, 2, 6]
         sampled = list(sampler)
-        # Check that the length is doubled
-        assert len(sampled) == 2 * (
-            len(dataset) - 1
-        )  # one element is dropped, because it's not enough to form a batch
-        # Check that the sampled indexes are a subset of the dataset indexes
+        assert len(sampled) == 2 * (len(dataset) - 1)
```
kashif:
```suggestion
        # one element is dropped, because it's not enough to form a batch
        assert len(sampled) == 2 * (len(dataset) - 1)
```

tests/slow/test_grpo_slow.py
```diff
@@ -188,3 +206,357 @@ def test_training_with_transformers_paged(self, model_name):
...
+        )
+        model = AutoModelForImageTextToText.from_pretrained(
+            model_name,
+            attn_implementation="flash_attention_2",
+            torch_dtype="bfloat16",
+            device_map=get_kbit_device_map(),
+            quantization_config=quantization_config,
+        )
+
+        processor = AutoProcessor.from_pretrained(model_name)
```
sergiopaniego:
```suggestion
```
Duplicated

tests/slow/test_grpo_slow.py
```diff
@@ -188,3 +206,357 @@ def test_training_with_transformers_paged(self, model_name):
...
+        - Minimal batch size (1) with gradient accumulation
+        - Small images (64x64 instead of 224x224)
+        - Short sequences (max_completion_length=8)
+        - Only 4 training samples
+        - Only 1 training step
+        - Gradient checkpointing and bfloat16
+        """
+
+        def data_gen(num_samples):
+            processor = AutoProcessor.from_pretrained(model_name)
```
sergiopaniego:
```suggestion
            processor = AutoProcessor.from_pretrained(model_name, use_fast=True, padding_side="left")
```

As the [docs](https://moon-ci-docs.huggingface.co/docs/trl/pr_3072/en/grpo_trainer#trl.GRPOTrainer.processing_class) say, `padding_side` should be `left`.

tests/slow/test_grpo_slow.py
```diff
@@ -188,3 +206,357 @@ def test_training_with_transformers_paged(self, model_name):
...
+                vllm_mode="colocate",  # Use colocate mode to avoid server dependency
+                vllm_gpu_memory_utilization=0.05,  # Use minimal GPU memory (5%)
+                gradient_checkpointing=True,  # Enable gradient checkpointing to save memory
+                bf16=True,  # Use bfloat16 to reduce memory
+                report_to="none",
+                logging_strategy="no",
+            )
+
+            # Create a VLM processor
+            processor = AutoProcessor.from_pretrained("HuggingFaceTB/SmolVLM-Instruct")
```
sergiopaniego:
```suggestion
            processor = AutoProcessor.from_pretrained("HuggingFaceTB/SmolVLM-Instruct", use_fast=True, padding_side="left")
```

Same as above

docs/source/grpo_trainer.md
```diff
@@ -524,6 +526,67 @@ and the reward will be computed as the sum of the rewards from each function, or
...
+- **Automatic VLM Detection**: GRPO automatically detects VLM processors (those with both `tokenizer` and `image_processor` attributes) and configures appropriate data handling
+- **Image Processing**: Supports various image formats and automatically processes images through the model's image processor
+- **vLLM Integration**: Full compatibility with vLLM acceleration in both `server` and `colocate` modes for VLMs
+- **Memory Optimization**: Supports 4-bit quantization with LoRA for memory-efficient VLM training
+- **Enhanced Data Collation**: Proper handling of VLM-specific data structures including `pixel_values`, `pixel_attention_mask`, and `image_sizes`
+
+### Supported VLM Models
+
+The implementation has been tested with popular VLM architectures including:
+- **Qwen2.5-VL**: `Qwen/Qwen2.5-VL-3B-Instruct`
```
qgallouedec: Nice
I'll create tiny models, it will be easier for testing

maks-ivanov: would anyone kindly share the setup for running test script with `Qwen2.5-VL-3B-Instruct` please? I tried the parameters from the example script docstring, getting cuda oom with 8 x A100 40GB which feels wrong

docs/source/grpo_trainer.md
sanchit97: Triggered an OOM error in the middle of training

docs/source/grpo_trainer.md
sergiopaniego: For that model, minimal tested config (completed successful train run):
* Example script, 1 A100 80GB.
* [Notebook/recipe](https://github.com/huggingface/trl/pull/3072#issuecomment-3062854034), 1 A100 40GB (Colab Pro).

trl/trainer/grpo_trainer.py
```diff
@@ -402,7 +491,7 @@ def __init__(
         args: Optional[GRPOConfig] = None,
         train_dataset: Optional[Union[Dataset, IterableDataset]] = None,
         eval_dataset: Optional[Union[Dataset, IterableDataset, dict[str, Union[Dataset, IterableDataset]]]] = None,
-        processing_class: Optional[PreTrainedTokenizerBase] = None,
+        processing_class: Optional[Union[PreTrainedTokenizerBase, AutoProcessor, ProcessorMixin]] = None,
```
sergiopaniego: In the rest of the trainers with VLM support, instead of `AutoProcessor`, it is `BaseImageProcessor`. I guess that it should be the same here, although I'm not sure about the reasoning tbh.

SFT: 
https://github.com/huggingface/trl/blob/c8347b428798913d7fa4b9516a7b5104f0e1d675/trl/trainer/sft_trainer.py#L356

DPO: 
https://github.com/huggingface/trl/blob/c8347b428798913d7fa4b9516a7b5104f0e1d675/trl/trainer/dpo_trainer.py#L259

trl/trainer/grpo_trainer.py
```diff
@@ -752,7 +864,30 @@ def _set_signature_columns_if_needed(self):
...
+                        func_sig = inspect.signature(func)
+                        params = func_sig.parameters
+                        # Add any parameter that's not one of the expected defaults
+                        for param_name in params:
+                            if param_name not in ["prompts", "completions", "kwargs", "trainer_state"]:
+                                if param_name not in signature_columns:
+                                    signature_columns.append(param_name)
+                    except (ValueError, TypeError):
+                        # If we can't inspect the function, just skip it
+                        pass
```
qgallouedec: I don't get why all of this is needed. I'll try with something simpler, and revert if I some point it makes sense for me

tests/slow/test_grpo_slow.py
```diff
@@ -188,3 +206,357 @@ def test_training_with_transformers_paged(self, model_name):
...
+                for key, original_value in original_env.items():
+                    if original_value is None:
+                        os.environ.pop(key, None)
+                    else:
+                        os.environ[key] = original_value
+
+                release_memory(model, trainer)
+
+    @require_vllm
+    def test_training_vllm(self):
```
qgallouedec, line 504: is it meant to be in this PR?

tests/test_grpo_trainer.py
```diff
@@ -1384,3 +1422,142 @@ def test_prepare_input_called_with_correct_data(self):
...
+        self.assertTrue(torch.equal(reconstructed_prompt_ids, batch_data["prompt_ids"]))
+        self.assertTrue(torch.equal(reconstructed_visual_input_ids, visual_inputs["input_ids"]))
+
+    def test_vlm_mode_includes_shuffling(self):
+        trainer = GRPOTrainer.__new__(GRPOTrainer)
+        trainer.is_vision_model = True
+        trainer.args = GRPOConfig(per_device_train_batch_size=2, steps_per_generation=2, num_generations=2)
+        trainer.num_iterations = 1
+        trainer._step = 0
+        trainer._buffered_inputs = None
```
qgallouedec: wtf is that?

trl/trainer/grpo_trainer.py
```diff
@@ -1663,3 +2068,104 @@ def create_model_card(
...
+        elif isinstance(inputs, list):
+            # Check if any sample in the batch has visual data
+            if not inputs:
+                return False
+            return any(
+                any(key in sample for key in ["image", "pixel_values", "pixel_attention_mask"]) for sample in inputs
+            )
+        return False
+
+    def _split_vlm_batch_safely(
```
qgallouedec: why do you need a sperate function for this AND modifying split_tensor_dict?

trl/trainer/utils.py
```diff
@@ -1807,6 +1813,123 @@ def entropy_from_logits(logits, chunk_size: int = 1) -> torch.Tensor:
     return per_token_entropies
 
 
+def validate_and_preprocess_images(images: list, processing_class=None) -> list:
```
qgallouedec: At this stage, having a function that tries to handle all formats really doesn't seem like a good idea. There will always be a case you haven't considered, and that requires a great deal of maintenance to have this in the codebase. I think it's up to the user to comply with the supported format, even if it means pre-processing the dataset.

trl/trainer/grpo_trainer.py
kashif: we dont need it, my thinking was that  there might be some extra columns in the data for vlms that might need passing to the reward model but you are right we don't need this and I'll remove it

tests/slow/test_grpo_slow.py
kashif: i moved this into the slow test from the normal tests... not technically part of this pr

qgallouedec:
FYI I'm refactoring a bit to simplify the whole approach. I'll push my updated version tomorrow on this branch

trl/trainer/grpo_trainer.py
```diff
@@ -550,12 +561,8 @@ def __init__(
             model = self._enable_gradient_checkpointing(model, args)
 
         # Processing class
-        self.is_vision_model = model.config.model_type in MODEL_FOR_VISION_2_SEQ_MAPPING_NAMES.keys()
```
kashif: we still need the vision model flag

trl/trainer/grpo_trainer.py
```diff
@@ -1125,10 +1230,26 @@ def _generate_and_score_completions(
         mode = "train" if self.model.training else "eval"
 
         prompts = [x["prompt"] for x in inputs]
+
+        # We don't yet support visual reward models/function, so we keep a copy of the original text-only prompts for
+        # later use in the reward computation. If images are present, we insert {"type": "image"} as required by the
+        # VLM chat template.
+        original_prompts = copy.deepcopy(prompts)
+
+        if "image" in inputs[0]:
```
```suggestion
        kwargs = {}
        if "image" in inputs[0]:
```

trl/trainer/grpo_trainer.py
```diff
@@ -1278,17 +1432,17 @@ def _generate_and_score_completions(
                 torch.no_grad(),
                 FSDP.summon_full_params(self.model_wrapped, recurse=False) if self.is_fsdp_enabled else nullcontext(),
             ):
+                prompt_inputs["input_ids"], prompt_inputs["attention_mask"] = prompt_ids, prompt_mask
                 prompt_completion_ids = unwrapped_model.generate(
-                    prompt_ids, attention_mask=prompt_mask, generation_config=self.generation_config
+                    **prompt_inputs, generation_config=self.generation_config
```
kashif, line 1519: FYI, because of this, the `fake_generate` in the test now expects the argument to have `input_ids` rather than previously, where it didn't matter

qgallouedec (approved):

trl/scripts/vllm_serve.py
```diff
@@ -13,17 +13,22 @@
...
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
+from io import BytesIO
 from itertools import chain
 from multiprocessing import Pipe, Process
 from multiprocessing.connection import Connection
 from typing import Optional
 
 import torch
+from fastapi import FastAPI
```
kashif:
```suggestion
```
as we define it further down

kashif (approved):

trl/scripts/vllm_serve.py
```diff
@@ -13,17 +13,21 @@
...
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
+from io import BytesIO
 from itertools import chain
 from multiprocessing import Pipe, Process
 from multiprocessing.connection import Connection
 from typing import Optional
 
 import torch
+from pydantic import BaseModel
```
kashif:
```suggestion
```

also defined below

tests/test_grpo_trainer.py
```diff
@@ -948,7 +1082,7 @@ def test_training_vllm_and_peft(self):
             )
             lora_config = LoraConfig(
                 target_modules="all-linear",
-                # test with non-default modules as it add extra keys in state_dict tht we need to handle
+                # test with non-default modules as it add extra keys in state_dict that we need to handle
```
sergiopaniego:
```suggestion
                # test with non-default modules as it adds extra keys in state_dict that we need to handle
```

scripts/generate_zen_image_dataset.py
```diff
@@ -0,0 +1,704 @@
+# Copyright 2020-2025 The HuggingFace Team. All rights reserved.
```
edbeeching, line 1: What is the purpose of this script and does it need be included?

trl/trainer/grpo_trainer.py
```diff
@@ -290,6 +300,38 @@ def identity(x):
     return x
 
 
+def split_pixel_values_by_grid(batch: dict[str, torch.Tensor]) -> dict[str, Union[torch.Tensor, list[torch.Tensor]]]:
```
edbeeching, line 303: It might be time to move all these functions into a separate `grpo_utils.py` file? As the `grpo_trainer` file is getting bloated. But we can leave this to a follow up PR. cc @qgallouedec

edbeeching: Also can tests be added for the split and unsplit functions

Revist:
Hi guys, thank you for the great work!

I am trying to use this PR with "llava-hf/llava-v1.6-mistral-7b-hf" however get an error "jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'text'". Is this caused by wrong dataset format or a bug in the PR?

```
accelerate launch   --config_file=examples/accelerate_configs/deepspeed_zero3.yaml   examples/scripts/grpo_vlm.py   --model_name_or_path llava-hf/llava-v1.6-mistral-7b-hf   --output_dir grpo-Qwen2.5-VL-3B-Instruct_test   --learning_rate 1e-5   --gradient_checkpointing   --torch_dtype bfloat16   --max_prompt_length 2048   --max_completion_length 1024   --use_vllm   --vllm_mode colocate   --use_peft   --lora_target_modules "q_proj", "v_proj"   --log_completions
```

trl/trainer/grpo_trainer.py
```diff
@@ -1125,10 +1295,30 @@ def _generate_and_score_completions(
...
+        original_prompts = copy.deepcopy(prompts)
+
+        kwargs = {}
+        has_images = "image" in inputs[0]
+        if has_images:
+            images = [example.get("image") for example in inputs]
+            kwargs = {"images": [[img] for img in images]}
+            for prompt in prompts:
+                if isinstance(prompt, list) and len(prompt) > 0 and isinstance(prompt[0], dict):
+                    prompt[0]["content"] = [{"type": "image"}, {"type": "text", "text": prompt[0]["content"]}]
```
sergiopaniego:
```suggestion
            for prompt in prompts:
                if isinstance(prompt, list):
                    for message in prompt:
                        if isinstance(message, dict) and message.get("role") == "user":
                            if isinstance(message.get("content"), str):
                                message["content"] = [{"type": "image"}, {"type": "text", "text": message["content"]}]
                            break
```

sergiopaniego: The original code adds the image as the first element of the prompt without verifying whether the message belongs to the "user" role. Could lead to incorrect formatting if the first message isn't from the user.

The updated version ensures the image is only prepended to the content of the first "user" message.

trl/trainer/grpo_trainer.py
qgallouedec: > Also can tests be added for the split and unsplit functions

Done in https://github.com/huggingface/trl/pull/3072/commits/b78af712e548c2738890fda925699ffb095973e1

qgallouedec: > It might be time to move all these functions into a separate `grpo_utils.py` file? As the `grpo_trainer` file is getting bloated. But we can leave this to a follow up PR. cc @qgallouedec

As long as they are used only once in the codebase, I prefer having them here. However, when PPO and RLOO are refactored (or RSTrainer implemented), we will move everything into a shared utils file.

scripts/generate_zen_image_dataset.py
qgallouedec: Here, the idea is only to generate the test dataset. I find it quite practical to keep these scripts somewhere, similar to the script for generating the tiny models. This script doesn't need to be there per se. Think of them as useful scripts for developers, a bit like https://github.com/huggingface/transformers/tree/main/utils

EauDeData:
Thanks you for this incredibly important merge. I think the whole DL community is happy today :)

From the initial PR I thought we did not have documentation, yet it passed the documentation check on merge; where can we find a clearer example of how to use it? I cannot find the updated docs...

Again, sincere thanks for your work.

kashif:
thanks @EauDeData you can find the docs here: https://huggingface.co/docs/trl/main/en/grpo_trainer#vision-language-model-vlm-training

ghubnerr:
Hi everyone! Sorry to bring this back - I noticed that the VLM support has a very strict format required for the element spec, where it expects a dict containing the `"prompt"` and `"image"` keys. This removes the control of where the user wants to insert the `<start_of_image>` tag, for example. I can tell that this decision was made because the `maybe_apply_chat_template` function returns a string, processed by the `processor_class`'s apply chat template method (with `tokenize=False`,  making images not compatible with that format. 

With the `AutoProcessor`, one can actually return tensors using the `apply_chat_template` method, which lets you control the image placement better. An example with `Gemma3Processor`:

```py
prompt = "This <img> is the image."
```

```py
messages = [
    {"role": "system", "content": [{"type": "text", "text": system_prompt}]},
    {"role": "user", "content": [
        {"type": "text", "text": prompt}
        {"type": "image", "image": Image.open(BytesIO(image_bytes)),
    ]},
```

Which could be called like this:

```py
formatted_mm_tokens = processor.apply_chat_template(
    conversation=messages,
    add_generation_prompt=True,
    do_pan_and_scan=True,
    tokenize=True,  # <-- Do tokenize
)
```

If this format is provided, this could potentially replace these [lines](https://github.com/huggingface/trl/blob/eee9ec94efbbadb3652aa428827b052b58f36ac7/trl/trainer/grpo_trainer.py#L1321C9-L1330C10) inside an if statement.

The `AutoProcessor` automatically identifies an image in the content and creates a multi-modal token array. It would be really great to have this sort of control. I'd be happy to work on this later -- I'm currently in the middle of an internship, but will be available soon.

qgallouedec:
Hi, thanks for reporting, contributions are very welcome to fix this

MohamedAliRashad:
Is this PR in a working state right or it had breaking changes ?

PR #2405, opened 2024-11-28
🔬 SFT simplification

qgallouedec:
# What does this PR do?

## Context

This PR is the first iteration of SFT refactoring: tests have deliberately been left unmodified, to ensure continuity of support for the user and, if possible, no breaking changes.

Other PRs will follow to further simplify the SFT (to be discussed on a case-by-case basis). These modifications would require test modifications and breaking changes. For example:
- drop support for `ConstantLengthDataset` in favor of the `packing` argument alone. 
- don't allow `TrainingArguments` and impose `SFTConfig`.
- better structure tests
- full support of  training on completions only (does not require a response template but use the data directly)
- ...

## What's in this PR

This PR refactor and simplifies SFT. 

### Add full conversational/standard dataset support

Part of #2071

### Refactor packing

Drop `ConstantLengthDataset` in favour of a simpler approach (see [`pack_examples`](https://github.com/huggingface/trl/blob/73702f7d441c2c88a78fd2b49d8292ce175431a8/trl/data_utils.py#L402))

## Current status

Testing. The goal is to gather code example from anywhere and ensure there're still working.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #3091, opened 2025-03-14
⛔ Add EOS token to processed input in SFT

qgallouedec:
Learn to generate EOS.

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1100,7 +1100,7 @@ class SFTTrainerTester(unittest.TestCase):
                 eval_dataset=self.conversational_lm_dataset["test"],
             )
 
-            self.assertEqual(len(trainer.train_dataset["input_ids"]), 46)  # w/ this dataset, we end up with 46 seqs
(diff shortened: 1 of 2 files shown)


kashif (approved):
for the gemma3 generation issue?

qgallouedec:
No, actually there haven't been recurrent complains that SFT can't learn to generate EOS. I'm pretty sure #2405 re-introduced the bug reported in #1623

skandermoalla:
@qgallouedec I've faced this multiple times. I think it's just because of the bad practice in the examples everywhere setting the pad token to the eos token. Then the SFT preprocessing masks everything that's a pad token (=eos token), including the real eos token in the chat template.

skandermoalla:
Personally, I don't think these forced patches are a good design. I understand that you want the Trainers to work out of the box, but users should still make sure they have a chat template that adds an eos properly. In case someone doesn't want an eos they can't do that anymore now.
(Same for the DPOTrainer btw, I think it adds an extra eos token somewhere.)

HwangYej1:
i got this TypeError: 'Qwen2TokenizerFast' object is not subscriptabl, after change this code

PR #3100, opened 2025-03-17
Update sft trainer

### LINKED ISSUE #2466: Probably a more reasonable method of `packing`
### Feature request

Hi! Here i m again : ) , according to #1850 i noticed that the `packing` method of `SFTTrainer` truncates overly long data directly and roughly, it's obviously that this will cause a lot of text information to be destroyed. 

When browsing the codes of [LLaMA-Factory](https://github.com/hiyouga/LLaMA-Factory), i find they use a different elegant way for packing. In short, they use the _greedy knapsack algorithm_ to fill each packing sequence as much as possible, maximizing its utilization. 

Maybe `trl` can be modified accordingly in the future. What's ur idea?

### Motivation

https://github.com/hiyouga/LLaMA-Factory/blob/main/src/llamafactory/data/processors/supervised.py#L130

### Your contribution

Not yet

Ishan-Kumar2:
# What does this PR do?

Fixes #2466 

## Before submitting
      Pull Request section?
      to it if that's the case.
- [] Did you write any new necessary tests?

## Who can review?
@qgallouedec @shirinyamani @AIR-hl 

Hi everyone, I am new to TRL, have attempted a fix at the issue 2466 here. 
The greedy knapsack tries to fit as many examples as possible without overshooting, the remaining ones are chunked as before by breaking it up into smaller example size (<= max_seq_len). Also I havent made this the default packing and kept it under a flag. 
Please let me know if this is close to what was expected, I can add  tests then, thanks!

### DIFF
--- a/trl/data_utils.py
+++ b/trl/data_utils.py
@@ -515,10 +515,10 @@ def maybe_convert_to_chatml(example: dict[str, list]) -> dict[str, list]:
     return example
 
 
-
 def find_largest_leq_k(ind_len_tuples, k):
     valid = [(i, t[0]) for i, t in enumerate(ind_len_tuples) if t[0] <= k]
     return max(valid, key=lambda x: x[1])[0] if valid else -1
+
 
 def greedy_knapsack(numbers: list[int], capacity: int) -> list[list]:
     """
@@ -573,7 +573,7 @@ def pack_examples_smarter(examples: dict[str, list[list]], seq_length: int) -> d
     """
     len_inputs = [(len(v), ind) for ind, v in enumerate(next(iter(examples.values())))]
     knapsacks, larger_knapsacks = greedy_knapsack(len_inputs, seq_length)
-    
+
     updated_examples = {k: [] for k in examples}
     for knapsack in knapsacks:
         temp_batch = {k: [] for k in examples}
@@ -586,7 +586,7 @@ def pack_examples_smarter(examples: dict[str, list[list]], seq_length: int) -> d
     for knapsack in larger_knapsacks:
         list_index = knapsack[0]
         length = len(next(iter(examples.values()))[list_index])
-        num_chunks = (length + seq_length - 1) // seq_length 
+        num_chunks = (length + seq_length - 1) // seq_length
 
         for chunk_idx in range(num_chunks):
             temp_batch = {k: [] for k in examples}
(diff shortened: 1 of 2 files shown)


trl/data_utils.py
```diff
@@ -513,3 +513,88 @@ def maybe_convert_to_chatml(example: dict[str, list]) -> dict[str, list]:
...
+                break
+            remaining_capacity -= numbers[index][0]  # update the remaining capacity
+            current_knapsack.append(numbers.pop(index)[1])  # add the number to knapsack
+
+        knapsacks.append(current_knapsack)
+
+    return knapsacks, [[tup[1]] for tup in numbers]
+
+
+def pack_examples_smarter(examples: dict[str, list[list]], seq_length: int) -> dict[str, list[list]]:
```
qgallouedec, line 547: wdyt of
```suggestion
def pack_examples_knapsack(examples: dict[str, list[list]], seq_length: int) -> dict[str, list[list]]:
```


or even

```suggestion
def pack_examples_knapsack_greedy(examples: dict[str, list[list]], seq_length: int) -> dict[str, list[list]]:
```

trl/trainer/sft_trainer.py
```diff
@@ -358,6 +364,7 @@ def _prepare_dataset(
         packing: bool,
         formatting_func: Optional[Callable[[dict], str]],
         dataset_name: str,
+        pack_smart: bool = False,
```
qgallouedec, line 367: maybe this is better:

```suggestion
        packing: Union[bool, str],
        formatting_func: Optional[Callable[[dict], str]],
        dataset_name: str,
```

and use `if packing == "knapsack":`

qgallouedec (commented):
Thanks for your contribution, and sorry for the late reply. I've made some comments. Can you please also:
- include a test for `pack_example_smater`
- modify the packing argument in `SFTConfig` to allow for this packing method

trl/data_utils.py
Ishan-Kumar2: fixed!

trl/trainer/sft_trainer.py
Ishan-Kumar2: fixed!

Ishan-Kumar2:
Hi @qgallouedec sorry for the delay, I had my final exams :)
I have made the changes, Please take a look!

qgallouedec:
I think we've arrived at something very satisfactory with FFD see #3521, so I'm closing this PR. Thanks for all your hard work!

PR #3118, opened 2025-03-19
🎬 Clip higher

shirinyamani:
# What does this PR do?

According to [DAPO](https://arxiv.org/abs/2503.14476) paper training detail, setting the upper_bound epsilon will result in higher entropy for the generation. Therefore, we added this feature to the default GRPO trainer!

Screenshot 2025-03-19 at 4 48 55 PM

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -111,8 +111,10 @@ class GRPOConfig(TrainingArguments):
             speed, but may be numerically unstable for long training runs.
         num_iterations (`int`, *optional*, defaults to `1`):
             Number of iterations per batch (denoted as μ in the algorithm).
-        epsilon (`float`, *optional*, defaults to `0.2`):
-            Epsilon value for clipping.
+        epsilon_low (`float`, *optional*, defaults to `0.2`):
+            Lower-bound epsilon value for clipping.
+        epsilon_high (`float`, *optional*, defaults to `0.28`):
+            Upper-bound epsilon value for clipping.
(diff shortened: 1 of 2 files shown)


trl/trainer/grpo_config.py
```diff
@@ -111,8 +111,10 @@ class GRPOConfig(TrainingArguments):
             speed, but may be numerically unstable for long training runs.
         num_iterations (`int`, *optional*, defaults to `1`):
             Number of iterations per batch (denoted as μ in the algorithm).
-        epsilon (`float`, *optional*, defaults to `0.2`):
-            Epsilon value for clipping.
+        epsilon_low (`float`, *optional*, defaults to `0.2`):
+            Lower-bound epsilon value for clipping.
+        epsilon_high (`float`, *optional*, defaults to `0.28`):
+            Upper-bound epsilon value for clipping.
```
qgallouedec, line 117: This change isn't backward compatible


What about this instead:

```suggestion
        epsilon (`float`, *optional*, defaults to `0.2`):
            Epsilon value for clipping.
        epsilon_high (`float`, *optional*, defaults to `None`):
            Upper-bound epsilon value for clipping. If not specified, the upper bound is the same as the lower bound specified in argument `epsilon`. Paper []() recommends `0.28`.
```

shirinyamani: then we need to keep the `epsilon` in the main grpo trainer ?

qgallouedec: I guess

qgallouedec (approved):

PR #3122, opened 2025-03-20
improvement(utils.py): simplify repeating completion string

tpoisonooo:
# What does this PR do?

The old logs are too verbose:
- especially when the LLM is repeating itself
- the reward is also hard to debug because only the total score is printed, and it's not clear which function might have calculated incorrectly

Here is new logging:
image

save my screen and easy to debug reward funcs 2333~

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -916,10 +916,10 @@ class GRPOTrainer(Trainer):
         if self.log_completions and self.state.global_step % self.args.logging_steps == 0:
             prompts_to_log = gather_object(prompts_text)
             completions_to_log = gather_object(completions_text)
-            rewards_to_log = rewards.tolist()
 
             if self.accelerator.is_main_process:
                 if is_rich_available():
+                    rewards_to_log = (rewards_per_func * self.reward_weights.to(device).unsqueeze(0)).to('cpu')
                     print_prompt_completions_sample(
                         prompts_to_log,
                         completions_to_log,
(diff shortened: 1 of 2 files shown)


tpoisonooo:
Please review.

qgallouedec:
What's the motivation? It seems that this can lead to the loss of important information.

tpoisonooo:
> What's the motivation? It seems that this can lead to the loss of important information.

## Motivation

Ignore repetitive information on the screen to speed up the debugging of open-r1.

## Reason

open-r1 starts training based on a cold-start or base model, which often leads to self-repetition. This causes the logs to occupy a large portion of the screen, and in `tmux`, one has to scroll through thousands of lines to see other logs.

I currently believe that:

- The self-repetitive completions of LLMs do not carry useful information, so I skip it
- Considering the execution speed of the code, directly calculating the frequency of each word can quickly filter out this content.

tpoisonooo:
If there are any better suggestions for improvement, I can make the adjustments.

qgallouedec:
Would something fix the issue? https://github.com/huggingface/trl/pull/3149

tpoisonooo:
> Would something fix the issue? #3149

Emmm, PR #3149 would cut `len(logs)`,  while this is for **high-efficiency debugging** , PR 3149 would not fix the issue.

qgallouedec:
I don't think this change is really necessary. I'm closing this for now, but if anyone in the community thinks this change is necessary, please report it here.

PR #3141, opened 2025-03-23
📖 Update `formatting_func`'s instances in `sft_trainer.md`.

YeFD:
Update implementation of `formatting_func` instances in `sft_trainer.md`.

# What does this PR do?

The `formatting_func` has been updated to return a processed string instead of a list after SFT simplification (#2405 ).
However, in the SFTTrainer's documentation, there are two instances of `formatting_func` that still use the old implementation to return a list:
https://github.com/huggingface/trl/blob/a0a53171cce596642c5fbb9e7b88b8f26a44ecad/docs/source/sft_trainer.md?plain=1#L75-L80
and
https://github.com/huggingface/trl/blob/a0a53171cce596642c5fbb9e7b88b8f26a44ecad/docs/source/sft_trainer.md?plain=1#L248-L253

The incorrect implementation will raise `IndexError: string index out of range`.

Therefore, I update these two examples to align with the correct instance in same documentation:
https://github.com/huggingface/trl/blob/a0a53171cce596642c5fbb9e7b88b8f26a44ecad/docs/source/sft_trainer.md?plain=1#L291-L293

## Before submitting
- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [x] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs).
- [x] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #3147, opened 2025-03-24
Fix: Compatibility for `formatting_func` returning a list

YeFD:
# What does this PR do?

When I defined `formatting_func` to return a list of processed strings following the documentation of SFTTrainer, an `IndexError` was raised.

```python
File ~/.conda/envs/rrag/lib/python3.12/site-packages/trl/trainer/sft_trainer.py:413, in SFTTrainer._prepare_dataset(self, dataset, processing_class, args, packing, formatting_func, dataset_name)
    410 if isinstance(dataset, Dataset):  # `IterableDataset.map` does not support `desc`
    411     map_kwargs["desc"] = f"Applying formatting function to {dataset_name} dataset"
--> 413 batched = isinstance(formatting_func(next(iter(dataset))), list)
    415 def _func(example):
    416     return {"text": formatting_func(example)}

Cell In[17], line 4, in formatting_prompts_func(example)
      2 output_texts = []
      3 for i in range(len(example['instruction'])):
----> 4     text = f"### Question: {example['instruction'][i]}\n ### Answer: {example['output'][i]}"
      5     output_texts.append(text)
      6 return output_texts

IndexError: string index out of range
```
I initially thought it was a documentation error and submitted a PR (https://github.com/huggingface/trl/pull/3141).

However, after reviewing the code, I found that it was a bug in the batched judgement. The issue was located in the following line:
https://github.com/huggingface/trl/blob/dee37342a8505beb5ad7e0dd01071c8b9ac584ba/trl/trainer/sft_trainer.py#L413 

After fixing the bug, the code is now compatible with `formatting_func` returning either a processed string or a list.

```python
def formatting_func(example):
    output_texts = []
    for i in range(len(example['instruction'])):
        text = f"### Question: {example['instruction'][i]}\n ### Answer: {example['output'][i]}"
        output_texts.append(text)
    return output_texts

batched = isinstance(formatting_func(dataset), list) # True

def formatting_func(example):
    text = f"### Question: {example['instruction']}\n ### Answer: {example['output']}"
    return text

batched = isinstance(formatting_func(dataset), list) # False
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -410,7 +410,7 @@ class SFTTrainer(Trainer):
                 if isinstance(dataset, Dataset):  # `IterableDataset.map` does not support `desc`
                     map_kwargs["desc"] = f"Applying formatting function to {dataset_name} dataset"
 
-                batched = isinstance(formatting_func(next(iter(dataset))), list)
+                batched = isinstance(formatting_func(dataset), list)
 
                 def _func(example):
                     return {"text": formatting_func(example)}


qgallouedec:
Applying the function to the entire dataset just to check if it is batched doesn’t seem reasonable. In fact, `formatting_func` is quite cumbersome, and I would advise against using it. Instead, we can simply preprocess the dataset before passing it to the trainer.  

That said, determining whether the function is batched is trickier than it seems. Also, I realize that we never document or test support for a non-batched function. The simplest approach would be to assume that the function is always batched, right?

qgallouedec:
```python
            if formatting_func is not None and not is_processed:
                if isinstance(dataset, Dataset):  # `IterableDataset.map` does not support `desc`
                    map_kwargs["desc"] = f"Applying formatting function to {dataset_name} dataset"

                def _func(example):
                    return {"text": formatting_func(example)}

                dataset = dataset.map(_func, batched=True, **map_kwargs)
```

YeFD:
> Applying the function to the entire dataset just to check if it is batched doesn’t seem reasonable. In fact, `formatting_func` is quite cumbersome, and I would advise against using it. Instead, we can simply preprocess the dataset before passing it to the trainer.
> 
> That said, determining whether the function is batched is trickier than it seems. Also, I realize that we never document or test support for a non-batched function. The simplest approach would be to assume that the function is always batched, right?

In fact, there is an instance of `formatting_func` using a non-batched implementation:
https://github.com/huggingface/trl/blob/a0a53171cce596642c5fbb9e7b88b8f26a44ecad/docs/source/sft_trainer.md?plain=1#L291-L293

If only the batched function is supported, we should update the documentation.

qgallouedec:
Indeed! Let's update the doc then

YeFD:
@qgallouedec Please review the code.

YeFD:
The test script is still using a non-batched `formatting_func`. Should I modify the test script to only test batched `formatting_func`?

qgallouedec:
I revisited this issue and explored some open codebases. Unfortunately, both forms are widely used.  

Here’s my suggestion:  

```python
try:
    dataset = dataset.map(_func, batched=False, **map_kwargs)
except Exception as e:
    warnings.warn(
        f"Failed to apply the formatting function due to the following error: {e}. This may be "
        "because the function is designed for batched input. Please update it to process one example "
        "at a time (i.e., accept and return a single example). For now, we will attempt to apply the "
        "function in batched mode, but note that batched formatting is deprecated and will be removed "
        "in version 0.21.",
        DeprecationWarning,
    )
    dataset = dataset.map(_func, batched=True, **map_kwargs)
```

Apologies for the change of direction and any extra work this may cause.

YeFD:
Thank you for updating the code and submitting the PR. I really appreciate your help during my break!

PR #2233, opened 2024-10-15
DPO support `remove_unused_columns`

qgallouedec:
## What does this PR do?

Currently,

1. removing unused column isn't supported
2. when not explicitly passing `--no_remove_unused_columns`, the following warning appears:

```
$ accelerate launch examples/scripts/dpo.py --output_dir Qwen2-0.5B-DPO --dataset_name trl-lib/lm-human-preferences-sentiment --model_name_or_path Qwen/Qwen2-0.5B-Instruct --logging_steps 10 --report_to none
/fsx/qgallouedec/trl/trl/trainer/dpo_trainer.py:720: UserWarning: When using DPODataCollatorWithPadding, you should set `remove_unused_columns=False` in your TrainingArguments. We have set it for you, but you should do it yourself in the future.
  warnings.warn(
```

This PR adds support for removing unused columns, which resolves the warning.

### Does it remove unused columns?

|                                     | Before | After |
| ----------------------------------- | ------ | ----- |
| `--remove_unused_columns` (default) | **No** | Yes   |
| `--no_remove_unused_columns`        | No     | No    |

### Does the training run successfully?

|                                     | Before                   | After |
| ----------------------------------- | ------------------------ | ----- |
| `--remove_unused_columns` (default) | Yes, but raises a warning | Yes   |
| `--no_remove_unused_columns`        | Yes                      | Yes   |

## Context

Trying to split #2209 into smaller pieces

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #3160, opened 2025-03-25
🏃 Fix and make CI faster

qgallouedec:
This PR introduces the following changes affecting the CI

### Unless there is a good reason, use default config values for greater readability
E.g in `test_dpo_trainer_with_weighting`

```diff
  training_args = DPOConfig(
      output_dir=tmp_dir,
-    per_device_train_batch_size=2,
-     max_steps=3,
-     remove_unused_columns=False,
-     gradient_accumulation_steps=1,
      learning_rate=9e-1,
-     eval_strategy="steps",
-     beta=0.1,
-     loss_type="sigmoid",
-     precompute_ref_log_probs=False,
      use_weighting=True,
      report_to="none",
  )
```

### For slow tests, use a `slow` marker

instead of a dedicated subfolder:

```diff
+ @pytest.mark.slow
  def my_slow_test(self): 
```

### Don't expose the import_utils to the root of the lib

These are mainly for internal use:

```diff
- from trl import is_diffusers_available
+ from trl.import_utils import is_diffusers_available
```

### For low priority tests, use a `low_priority` marker

eg, for alignprop trainer

```diff
+ @pytest.mark.low_priority
  def my_low_priority_test(self): 
```

### Refactor BCO tests

The tests are mostly identical as before, but cleaned and optimised. Other refactors to follow


tests/slow/test_sft_slow.py
```diff
@@ -422,8 +422,6 @@ def test_sft_trainer_with_liger(self, model_name, packing):
                 eval_dataset=self.eval_dataset,
             )
 
-            # check that the components of the trainer.model are monkey patched:
-            self.assertTrue(any("Liger" in type(module).__name__ for module in trainer.model.model.modules()))
```
qgallouedec, line 426: This assertion doesn't pass anymore. The liger has been moved to transformers anyway

tests/test_ppo_trainer.py
```diff
@@ -170,55 +176,3 @@ def test_peft_training(self):
 
             self.assertTrue(critic_weights_updated, "Critic weights were not updated during training")
             self.assertTrue(policy_weights_updated, "Policy LoRA weights were not updated during training")
-
-    def test_with_num_train_epochs(self):
```
qgallouedec, line 174: seems like a duplicated test

trl/__init__.py
```diff
@@ -36,15 +36,6 @@
...
     "extras": ["BestOfNSampler"],
-    "import_utils": [
-        "is_deepspeed_available",
-        "is_diffusers_available",
-        "is_llm_blender_available",
-        "is_mergekit_available",
-        "is_rich_available",
-        "is_unsloth_available",
-        "is_vllm_available",
-    ],
```
qgallouedec, line 47: Having that at the root is not a good thing, IMO.

tests/test_bco_trainer.py
qgallouedec, line 1: Heavy refactor of this one. To keep the review easy I'll refactor the others in follow-up PRs

qgallouedec, line 1: Refactor.

Before : 1min08
Now: 44 sec

Makefile
```diff
@@ -6,14 +6,14 @@ ACCELERATE_CONFIG_PATH = `pwd`/examples/accelerate_configs
 COMMAND_FILES_PATH = `pwd`/commands
 
 test:
-	pytest -n auto --dist=loadfile -s -v --reruns 5 --reruns-delay 1 --only-rerun '(OSError|Timeout|HTTPError.*502|HTTPError.*504||not less than or equal to 0.01)' ./tests/
```
qgallouedec, line 9: `--dist=loadfile` isn't necessary, and probably slows test when used

trl/trainer/sft_trainer.py
```diff
@@ -228,6 +228,19 @@ def __init__(
         if processing_class is None:
             processing_class = AutoTokenizer.from_pretrained(model_id)
 
+        # Model
```
qgallouedec, line 231: This part should be before the collator part

tests/test_bco_trainer.py
```diff
@@ -20,76 +20,83 @@
 from accelerate import Accelerator
 from datasets import load_dataset
 from parameterized import parameterized
-from transformers import AutoModel, AutoModelForCausalLM, AutoModelForSeq2SeqLM, AutoTokenizer
```
lewtun, line 23: Just so I understand, we're removing testing for seq2seq models in this PR right? This is fine with me since no significant LLM with that arch has been released for several years, but maybe good to mention in the PR description so users can find the PR we dropped testing for this if needed.

tests/test_bco_trainer.py
```diff
@@ -20,76 +20,83 @@
...
             training_args = BCOConfig(
                 output_dir=tmp_dir,
-                per_device_train_batch_size=2,
-                max_steps=3,
-                gradient_accumulation_steps=1,
-                learning_rate=9e-1,
-                eval_strategy="steps" if eval_dataset else "no",
-                beta=0.1,
-                precompute_ref_log_probs=pre_compute,
+                remove_unused_columns=False,  # warning raised if not set to False
```
lewtun, line 51: Not for this PR, but should we then set this as the default value in `BCOConfig`?

lewtun (approved):
Nice clean up! LGTM with a question on whether we can speed things up even further by using `setUpClass()` to init the model / tokenizer once per set of trainer tests

qgallouedec:
> Nice clean up! LGTM with a question on whether we can speed things up even further by using setUpClass() to init the model / tokenizer once per set of trainer tests

The issue with using `setUpClass` is that the objects are shared between tests cases, and it can interfer with test. Initially we used it but we've changed to use `setUp` instead. See here: https://github.com/huggingface/trl/pull/1895

tests/test_sft_trainer.py
```diff
@@ -772,70 +633,10 @@ def test_data_collator_chat_completion_lm_with_multiple_text(self):
...
-            trainer.train()
-
-            self.assertIsNotNone(trainer.state.log_history[(-1)]["train_loss"])
-            self.assertIsNotNone(trainer.state.log_history[0]["eval_loss"])
-
-            # make sure the trainer did 5 steps
-            self.assertIn("model.safetensors", os.listdir(tmp_dir + "/checkpoint-5"))
-
-    def test_sft_trainer_infinite_with_model_epochs(self):
+    def test_with_model_neftune(self):
```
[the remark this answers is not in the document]
qgallouedec: agree, let's remove it in a follow-up PR

tests/test_bco_trainer.py
```diff
@@ -20,76 +20,83 @@
...
-            ("qwen", False, True, "standard_unpaired_preference"),
-            ("qwen", False, False, "standard_unpaired_preference"),
-            ("qwen", True, True, "conversational_unpaired_preference"),
-        ]
-    )
+    @parameterized.expand([("standard_unpaired_preference"), ("conversational_unpaired_preference")])
     @require_sklearn
-    def test_bco_trainer(self, name, pre_compute, eval_dataset, config_name):
+    def test_train(self, config_name):
+        model_id = "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5"
```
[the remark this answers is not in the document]
qgallouedec: > I guess one caveat here is that training might mutate the model

yes, see https://github.com/huggingface/trl/pull/3160#issuecomment-2786368239

tests/test_bco_trainer.py
qgallouedec: Or even better, make BCO support this arg. Like we did for DPO in https://github.com/huggingface/trl/pull/2233

PR #3174, opened 2025-03-28
🏃 Migrate CI to self-hosted runners

qgallouedec:
This PR:

1. Migration of our CI to a self-hosted runner. This allows us to test on more relevant hardware, since it has 1 GPU. (instead of CPU only)

2. Give up support for Windows. Reasons:
  - Windows is the recurring source of problems in our CI and takes up a lot of our time.
  - According to pypistats it's 3% of TRL usage.
  - Most optional dependencies don't support Windows anyway.
Note that this doesn't mean that TRL no longer works with Windows, it's just that we no longer test it and no longer officially support it.

🏎️💨 2.5x faster


.github/workflows/tests.yml
```diff
@@ -24,9 +24,6 @@ jobs:
     if: github.event.pull_request.draft == false
     steps:
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-          submodules: recursive
```
qgallouedec, line 29: we don't need it

.github/workflows/tests.yml
```diff
@@ -40,129 +37,207 @@ jobs:
...
-        os: ['ubuntu-latest', 'windows-latest']
       fail-fast: false
-    runs-on: ${{ matrix.os }}
+    runs-on:
+      group: aws-g4dn-2xlarge
+    container:
+      image: nvidia/cuda:12.8.1-cudnn-devel-ubuntu24.04
+    defaults:
+      run:
+        shell: bash
```
qgallouedec, line 47: `bash` is needed otherwise it uses `sh` which doesn't have `source`

.github/workflows/tests.yml
```diff
@@ -40,129 +37,207 @@ jobs:
...
+        uses: actions/checkout@v4
+
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-          cache: "pip"
-          cache-dependency-path: |
-              setup.py
-              requirements.txt
```
qgallouedec, line 56: not compatible with the runner (I don't think it's very useful anyway)

.github/workflows/tests.yml
```diff
@@ -40,129 +37,207 @@ jobs:
...
+
+      - name: Install Make and Git
+        run: |
+          apt-get update && apt-get install -y make git curl
+
+      - name: Install uv
```
qgallouedec, line 62: Much faster installation with uv

tests/test_grpo_trainer.py
```diff
@@ -666,38 +665,6 @@ def test_training_vllm(self):
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
 
-    @unittest.skipIf(sys.platform.startswith("win"), "Skipping on Windows")  # compiling seems to be broken on Windows
-    def test_training_torch_compile(self):
```
qgallouedec, line 670: this test was veeeery slow

kashif (approved):

qgallouedec:
I'll merge because for now the CI is broken, but I'm still interested in your remarks if any

PR #3175, opened 2025-03-29
📎 Fix is_clipped to compute the effective clip_ratio

pandong2011:
# What does this PR do?

Below, let me briefly explain my motives and reasons, the code mainly uses the part marked in the figure.
image

if (coef_1 < (1 - self.epsilon_low) and advantage > 0) or (coef_1 > (1 + self.epsilon_high) and advantage < 0)
then coef_1 * advantage < (1 - self.epsilon_low) * advantage
         coef_1 * advantage < (1 + self.epsilon_high) * advantage
min(per_token_loss1, per_token_loss2) = per_token_loss1

**From the perspective of loss, it is actually not clipped.**
`clip_ratio = (is_clipped * completion_mask).sum() / completion_mask.sum()`
**And this code also calculates an effective clip_ratio**       
**original code:** `is_clipped = (coef_1 < (1 - self.epsilon_low)) | (coef_1 > (1 + self.epsilon_high))`
**Therefore, we must combine the sign of the advantage with the original code to calculate is_clipped, thereby computing an effective clip_ratio**

**As follows**
`is_clipped = (coef_1 < (1 - self.epsilon_low) and advantage < 0) or (coef_1 > (1 + self.epsilon_high) and advantage > 0)`
if (coef_1 < (1 - self.epsilon_low) and advantage < 0) or (coef_1 > (1 + self.epsilon_high) and advantage > 0)
then coef_1 * advantage > (1 - self.epsilon_low) * advantage
         coef_1 * advantage > (1 + self.epsilon_high) * advantage
min(per_token_loss1, per_token_loss2) = per_token_loss2

**Therefore, by modifying is_clipped as described above, the effective clip_ratio can be correctly calculated**

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -967,11 +967,11 @@ class GRPOTrainer(Trainer):
             mean_kl = (per_token_kl * completion_mask).sum() / completion_mask.sum()
             self._metrics[mode]["kl"].append(self.accelerator.gather_for_metrics(mean_kl).mean().item())
 
-        # Compute the effective clip_radio
+        # Compute the clip ratio
         is_clipped = (
             ((coef_1 < 1 - self.epsilon_low) & (advantages.unsqueeze(1) < 0)) |
             ((coef_1 > 1 + self.epsilon_high) & (advantages.unsqueeze(1) > 0))
-        ).float()
+        )
         clip_ratio = (is_clipped * completion_mask).sum() / completion_mask.sum()
         self._metrics[mode]["clip_ratio"].append(self.accelerator.gather_for_metrics(clip_ratio).mean().item())
         return loss
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -967,7 +967,11 @@ class GRPOTrainer(Trainer):
             mean_kl = (per_token_kl * completion_mask).sum() / completion_mask.sum()
             self._metrics[mode]["kl"].append(self.accelerator.gather_for_metrics(mean_kl).mean().item())
 
-        is_clipped = (coef_1 < (1 - self.epsilon_low)) | (coef_1 > (1 + self.epsilon_high))
+        # Compute the effective clip_radio
+        is_clipped = (
+            ((coef_1 < 1 - self.epsilon_low) & (advantages.unsqueeze(1) < 0)) |
+            ((coef_1 > 1 + self.epsilon_high) & (advantages.unsqueeze(1) > 0))
+        ).float()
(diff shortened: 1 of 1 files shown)


pandong2011:
@qgallouedec

trl/trainer/grpo_trainer.py
```diff
@@ -967,7 +967,11 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
             mean_kl = (per_token_kl * completion_mask).sum() / completion_mask.sum()
             self._metrics[mode]["kl"].append(self.accelerator.gather_for_metrics(mean_kl).mean().item())
 
-        is_clipped = (coef_1 < (1 - self.epsilon_low)) | (coef_1 > (1 + self.epsilon_high))
+        # Compute the effective clip_radio
+        is_clipped = (
+            ((coef_1 < 1 - self.epsilon_low) & (advantages.unsqueeze(1) < 0)) |
+            ((coef_1 > 1 + self.epsilon_high) & (advantages.unsqueeze(1) > 0))
+        ).float()
```
qgallouedec, line 974: In the end, it's equivalent to this, right?

```suggestion
        # Compute the effective clip_ratio
        is_clipped = per_token_loss == -per_token_loss2
```

qgallouedec (commented):
Thanks @pandong2011! A small remark

qgallouedec:
Ah yes good point

trl/trainer/grpo_trainer.py
qgallouedec, line 974:
```suggestion
        # Compute the clip ratio
        is_clipped = (
            ((coef_1 < 1 - self.epsilon_low) & (advantages.unsqueeze(1) < 0)) |
            ((coef_1 > 1 + self.epsilon_high) & (advantages.unsqueeze(1) > 0))
        )
```

pandong2011:
image
@qgallouedec Haha, I really understand why you've submitted so much code for this project. I don't understand why, when you actually don't understand the code, you would still make those modifications that you think are correct, and even publicly replace my code with yours. I've already made so many explanations above. Could you please respect my hard work? Also, could you please respect the open-source spirit? If your modifications are correct or meaningful, then I have nothing to say. If you think there's nothing else to modify and you want to change the comments, then I suggest you don't do it, because this is the calculated effective clipping rate, which reflects the actual impact of clipping on the loss. Finally, I suggest that you make this project closed-source and then claim that you've done the whole project. You can definitely tinker with this project all by yourself. There's no need to try every means to submit more code, as it's meaningless. Is it for your KPI?

qgallouedec:
what is clip_radio?

qgallouedec (approved):

PR #3182, opened 2025-03-30
(AsyncLLMEngine) Change VLLM Server from Sync to Async

binary-husky:
1. Change VLLM Server from Sync to Async
    - if is_async=True:
        client first call `generate` (non-blocking), then after a while call `get_future` (with identical arguments) to get result
    - if is_async=False:
        client automatically call `get_future` inside `generate`, blocking further execution before the generation is complete

<div align="center">

</div>

2. Speed up grpo_trainer 1.5x faster by submitting N=`gradient_accumulation_steps` batches, so that training and vllm generation can run in parallel! 
However, I have to admit that this piece of code is not elegant enough, remove them if they disqualifies.

<div align="center">

</div>

3. I leave some room by adding a `RolloutEngine` in `trl.scripts.vllm_serve`, for more sophisticated vllm inference functionality, trying to support `lm_generate > MCP tool_call > lm_generate > another MCP tool_call > ...`, but not complete yet.

<div align="center">

</div>

4. add vllm_server_nccl_port in config (previously cannot change default)

<div align="center">

</div>

### DIFF
--- a/trl/extras/vllm_client.py
+++ b/trl/extras/vllm_client.py
@@ -15,6 +15,7 @@
 import atexit
 import logging
 import time
+import uuid
 from typing import Optional
 
 import torch
@@ -80,7 +81,7 @@ class VLLMClient:
     """
 
     def __init__(
-        self, host: str = "0.0.0.0", server_port: int = 8000, group_port: int = 51216, connection_timeout: float = 0.0
+        self, host: str = "0.0.0.0", server_port: int = 8000, group_port: int = 51216, connection_timeout: float = 0.0, without_nccl: bool = False
     ):
         if not is_requests_available():
             raise ImportError("requests is not installed. Please install it with `pip install requests`.")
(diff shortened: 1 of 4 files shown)


binary-husky:
oh, there is another detail worth mentioning:

I add a `version` param, `self.version += 1` whenever `update_model_params` is called.

at server side, I add some lines to ensure there are no on-going generation with some async sleep logic
<div align="center">

</div>

fabianlim:
@binary-husky the speedups you posted look great, though I have a question on how you parallelize the computation. THe picture shows a data dependency between roll outs and model training (and vllm update). 
- are you saying that within gradient accumulation steps the rollouts do not change?
- the `completion_ids` are futures, are you saying the will return enough rollouts for you to complete the grad accum step? 

In other words, this achieve parallization within grad accum steps, and works only if the grad accum > 1?

image

binary-husky:
@fabianlim Yes, works only if the grad acc step > 1.

image

tcapelle:
We should make this happen! The new Mistral reasoning model uses a pipeline like this one.

tcapelle:
Shouldn't populating `_buffered_inputs` asynchronously be enough?

qgallouedec:
Recent update, we will work on a full asynchronous train, the integration will be simpler. Closing this one, thank you though!

PR #3185, opened 2025-03-31
⏯️ Fix logging when resuming from checkpoint GRPO

qgallouedec:
First training, (orange) from 0 to 100 timestep
Second training (purple), resume from timestep 40: curves overlap.
Screenshot 2025-03-30 at 20 24 01
Screenshot 2025-03-30 at 20 24 06


trl/trainer/grpo_trainer.py
```diff
@@ -481,7 +480,7 @@ def data_collator(features):  # No data collation is needed in GRPO
             # vLLM specific sampling arguments
             self.guided_decoding_regex = args.vllm_guided_decoding_regex
 
-            self._last_loaded_step = 0  # tag to avoid useless loading during grad accumulation
+            self._last_loaded_step = -1  # tag to avoid useless loading during grad accumulation
```
qgallouedec, line 483: Make sure that for the very first generation, the model loaded is the same as that of the model running on the server.

kashif (approved):

PR #3196, opened 2025-03-31
Support for Models With Pre-Finetuned LoRA Adapters in GRPO: Add use_peft_as_reference Flag

### LINKED ISSUE #3194: GRPO with pre-finetuned LoRA model as reference
### Feature request

When using the GRPO tuner If you want to GRPO tune a LoRA model which already has fine-tuned LoRA weights the reference model will use the base model and not the SFT LoRA weight leading to divergence since there is an inherent and dubious KL divergence.

### Motivation

 I propose that we add a new argument `use_peft_as_reference` which forces the reference model to copy the full base model and LoRA weights. This will rectify the unwanted difference between the reference and the policy and stabilize the KL divergence for already finetuned LoRA models. Similarly if one performs Qbit training they may want the reference model to be quantized as well for KL divergence stability. 

### Your contribution

This is a similar change to @tchang1997's code and #3108

grpo_trainer.py line 314
`# Reference model
        self.beta = args.beta
        if self.beta == 0.0:
            self.ref_model = None
        elif is_deepspeed_zero3_enabled():
        elif is_peft_model(model) and not args.use_peft_as_reference:
            # If PEFT is used, the reference model is not needed since the adapter can be disabled
            # to revert to the initial model.
            self.ref_model = None
        else:
            # If PEFT configuration is not provided, create a reference model based on the initial model.
            self.ref_model = create_reference_model(model)
`

LoganVegnaSHOP:
# What does this PR do?

This PR introduces a new flag, use_peft_as_reference, to the GRPO configuration and trainer. When using the GRPO tuner with PEFT models (e.g. LoRA or quantized models), the trainer by default creates the reference model by disabling the adapter. This behavior, however, is undesirable when you are working with a model that has already been fine-tuned using LoRA weights—the reference model should mirror the full model (including the adapter) to avoid unwanted divergence.
With the new use_peft_as_reference flag, if set to True, the reference model is created directly using the full PEFT model (via create_reference_model), thereby retaining the fine-tuned adapter weights (or quantization) in the reference. This adjustment ensures a closer match between the policy and reference models, which translates into more stable KL divergence during training.
Additionally, the PR adds tests to verify that when the flag is enabled, the ref_model is properly set.

Fixes #3194

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -923,7 +923,9 @@ class GRPOTrainerTester(unittest.TestCase):
         training_args = GRPOConfig(
             output_dir=tempfile.mkdtemp(),
             learning_rate=0.1,
-            per_device_train_batch_size=1,
+            per_device_train_batch_size=3,
+            num_generations=3,
+            max_completion_length=32,
             use_peft_as_reference=True,
             report_to="none",
         )
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -919,9 +919,10 @@ class GRPOTrainerTester(unittest.TestCase):
     def test_peft_use_as_reference_flag_true(self):
         # Test that for a PEFT model with use_peft_as_reference=True, ref_model is set
         model = AutoModelForCausalLM.from_pretrained("trl-internal-testing/tiny-Qwen2ForCausalLM-2.5")
-        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train").select(range(5))
-        training_args = GRPOConfig(
-            output_dir=tempfile.mkdtemp(),
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            training_args = GRPOConfig(
+                output_dir=tmp_dir,
             learning_rate=0.1,
             per_device_train_batch_size=3,
             num_generations=3,
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -28,7 +28,7 @@ from trl.trainer.grpo_trainer import RepeatRandomSampler
 
 
 if is_peft_available():
-    from peft import LoraConfig, PeftModel
+    from peft import LoraConfig, PeftModel, get_peft_model
 
 
 class RepeatRandomSamplerTester(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


qgallouedec:
Thanks @LoganVegnaSHOP. I wonder in what situation one would want this flag to be set `False`?

tests/test_grpo_trainer.py
```diff
@@ -914,3 +914,26 @@ def test_training_no_scale_rewards(self):
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
+
+    @require_peft
+    def test_peft_use_as_reference_flag_true(self):
+        # Test that for a PEFT model with use_peft_as_reference=True, ref_model is set
+        model = AutoModelForCausalLM.from_pretrained("trl-internal-testing/tiny-Qwen2ForCausalLM-2.5")
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train").select(range(5))
```
qgallouedec, line 922:
```suggestion
        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
```

tests/test_grpo_trainer.py
```diff
@@ -914,3 +914,26 @@ def test_training_no_scale_rewards(self):
...
+        training_args = GRPOConfig(
+            output_dir=tempfile.mkdtemp(),
```
qgallouedec, line 924:
```suggestion
        with tempfile.TemporaryDirectory() as tmp_dir:
        training_args = GRPOConfig(
            output_dir=tmp_dir,
```

tests/test_grpo_trainer.py
```diff
@@ -914,3 +914,26 @@ def test_training_no_scale_rewards(self):
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
+
+    @require_peft
+    def test_peft_use_as_reference_flag_true(self):
```
qgallouedec, line 919: To confirm my understanding of this test: here the model passed isn't a peft model, right?

LoganVegnaSHOP:
> Thanks @LoganVegnaSHOP. I wonder in what situation one would want this flag to be set `False`?

If you are initializing brand new LoRA adapters it would be more efficient to set the flag to false since the ref model would then be slightly smaller when excluding the lora weights. @qgallouedec

tests/test_grpo_trainer.py
LoganVegnaSHOP: Nonetheless, because the intent of the test isn't clear I will construct a PEFT model in the train function instead to simulate the real world use case of this arg

qgallouedec:
closing, see https://github.com/huggingface/trl/issues/3194#issuecomment-5111888363

PR #3199, opened 2025-03-31
📢 Improve GRPO trainer error message for invalid num_generations

AliBakly:
This PR fixes a confusing error message when users try to use `num_generations=1` with the GRPO trainer.

**Current Message:**
`ValueError: The global train batch size (1 x 4) must be evenly divisible by the number of generations per prompt (1). Given the current train batch size, the valid values for the number of generations are: [2, 4].`
This is mathematically incorrect since 1 is a valid divisor of any number, and confused me in the beginning. The real issue is that GRPO requires at least 2 generations to function properly.

**Changes**
Just raise a ValueError if `num_generations<2` with a more informative message.


qgallouedec (approved):
Looks good to me :)

PR #3176, opened 2025-03-29
❤️‍🩹 [CI] fix transformers dev CI failure

kashif:
# What does this PR do?
For:
https://github.com/huggingface/trl/pull/3174

PR #3206, opened 2025-04-01
📉 Add `learning_rate` argument to `_maybe_log_save_evaluate`

qgallouedec:
> [!IMPORTANT]
> This PR is on `release v0.16` branch!

This PR addresses a breaking change introduced in the `transformers` dev branch: https://github.com/huggingface/transformers/pull/36973. This was fixed in the `trl` main branch with #3176, but this fix is not part of TRL 0.16, which could cause issues once transformers new release goes live. To prevent this, we will release v0.16.1, which includes a fix.
By releasing this patch ahead of the upcoming `transformers` release, users will avoid potential bugs caused by the breaking change in `transformers` dev branch. This ensures that `trl` remains compatible with the new `transformers` version as soon as it's released.

This breaking case was reported thanks to this CI: https://github.com/huggingface/trl/actions/workflows/tests_latest.yml

### DIFF
--- a/trl/trainer/online_dpo_trainer.py
+++ b/trl/trainer/online_dpo_trainer.py
@@ -695,7 +695,9 @@ class OnlineDPOTrainer(Trainer):
 
     # Same as Trainer._maybe_log_save_evaluate but log our metrics
     # start_time defaults to None to allow compatibility with transformers<=4.46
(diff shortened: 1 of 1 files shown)


lewtun (approved):
LGTM!

Issue #3201, opened 2025-04-01
Vllm sampling parameters cannot be set


shaipranesh2:
### Feature request

some of the Vllm parameters like temperature, top_k can be set now in the trl package, but other parameters like stop_word_ids cannot be set. It would be great to have this feature to set up all the sampling parameters, as it varies from model to model.

### Motivation

I was trying to set up trl-vllm with qwen-2.5, and despite being a instruct-finetuned model, it was giving me an output of random tokens, and only with stop_toke_ids set up  correctly, it gave me the proper output

### Your contribution

A PR changing the fields in trl/extras/vllm_client.py will fix this issue

shaipranesh2:
This feature will allow models like qwen-2.5 to be run correctly

qgallouedec:
Thanks, feel free to open a PR :)

PR #3210, opened 2025-04-02
Adding sampling parameters for vllm generation

shaipranesh2:
# What does this PR do?
Fixes issue #3201 . Added support in GRPO config files to set additional parameters of vllm sampling.

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -163,6 +163,12 @@ class GRPOConfig(TrainingArguments):
             "help": "Maximum length of the prompt. If the prompt is longer than this value, it will be truncated left."
         },
     )
+    min_prompt_length: Optional[int] = field(
+        default=0,
+        metadata={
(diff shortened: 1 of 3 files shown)


shaipranesh2:
@qgallouedec This PR is ready for review :)

trl/trainer/grpo_config.py
qgallouedec, line 1: For parameters specific to vllm, I would name it with a `vllm` prefix. Ideally some are shared by transformers and vllm (like temperature), in such case, they don't need a prefix

qgallouedec (commented):
Thanks @shaipranesh2! I've added a few comments

kanishkg:
Can this instead just use the SamplingParams object from vllm instead? Or something more flexible?
Since the flags for sampling in vllm will keep changing.

PS: Was about to create a related issue and found this pr :)

qgallouedec:
closing as we now have `generation_kwargs`

PR #3217, opened 2025-04-02
📚 Accumulate completions for logging

lewtun:
# What does this PR do?

This PR fixes a bug where the `rich` and `wandb` tables were not logging the complete set of unique prompts for `gradient_accumulation_steps > 1`. Previously what happened is that the tables we logged for each gradient accumulation step, which is confusing when inspecting the logs for unique prompts.

Here's two runs with/without the fix for:

* `main`: https://wandb.ai/huggingface/huggingface/runs/j3vwwb5t/workspace?nw=nwuserlewtun
* `this_pr`: https://wandb.ai/huggingface/huggingface/runs/78q63hev/overview

These runs have:

* `gradient_accumulation_steps=4`
* `num_generations=16`
* `per_device_train_batch_size=4`
* `num_gpus = 8`

so we expect `num_unique_prompts=8`:

```
num_unique_prompts = num_gpus * per_device_train_batch_size * gradient_accumulation_steps / num_generations = 8
```

We see that without the fix, `wandb` only displays 2 unique prompts in the table (i.e. once per gradient accumulation step). With the fix, the prompts are grouped together per step, which incidentally makes the table scrolling faster in the UI as we have fewer steps to query.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -136,6 +136,9 @@ class GRPOConfig(TrainingArguments):
             installed, it prints the sample. If `wandb` logging is enabled, it logs it to `wandb`.
         num_completions_to_print (`int` or `None`, *optional*, defaults to `None`):
             Number of completions to print with `rich`. If `None`, all completions are logged.
+        wandb_log_unique_prompts (`bool`, *optional*, defaults to `False`):
+            Whether to log unique prompts in wandb. If `True`, only unique prompts are logged. If `False`, all
+            prompts are logged.
     """
 
     # Parameters that control the model and reference model
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -904,36 +904,63 @@ class GRPOTrainer(Trainer):
         self._metrics[mode]["reward"].append(mean_grouped_rewards.mean().item())
         self._metrics[mode]["reward_std"].append(std_grouped_rewards.mean().item())
 
+        # Log completions when we complete a full gradient accumulation cycle
+        # For logging across the gradient accumulation steps, we need to accumulate the data
+        is_last_step_in_grad_accum = (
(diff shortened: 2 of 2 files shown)


trl/trainer/grpo_config.py
```diff
@@ -136,6 +136,9 @@ class GRPOConfig(TrainingArguments):
             installed, it prints the sample. If `wandb` logging is enabled, it logs it to `wandb`.
         num_completions_to_print (`int` or `None`, *optional*, defaults to `None`):
             Number of completions to print with `rich`. If `None`, all completions are logged.
+        wandb_log_unique_prompts (`bool`, *optional*, defaults to `False`):
```
lewtun, line 139: Missed from #3191

trl/trainer/grpo_trainer.py
```diff
@@ -904,36 +904,63 @@ def _generate_and_score_completions(
...
+        is_last_step_in_grad_accum = (
+            self._step % self.args.gradient_accumulation_steps == self.args.gradient_accumulation_steps - 1
+        )
+        should_log_completions = (
+            self.log_completions
+            and self.state.global_step % self.args.logging_steps == 0
+            and is_last_step_in_grad_accum
+        )
+
+        # Collect data for logging throughout the accumulation steps
```
qgallouedec, line 918: I think `self.control.should_log` achieves the same goal. Let me check

qgallouedec: So yes it achieves the same goal but this flag is only update after the `loss.backward` so we can't use it here. I'm starting to think that it would make more sense to have all this code in `self.log`. I'm checking if it's possible.

qgallouedec: So yes It's possible, I'll do it in a follow-up PR

qgallouedec (approved):

PR #3228, opened 2025-04-03
remove unused optimizer config in prepare_deepspeed

liguohao96:
# What does this PR do?

remove optimizer config in `prepare_deepspeed` in `model/utils.py`, to prevent potential OOM due to deepspeed creating  (never used) optimizer

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/models/utils.py
+++ b/trl/models/utils.py
@@ -246,6 +246,9 @@ def prepare_deepspeed(model, accelerator):
                 }
             )
 
+    # remove optimizer related config, tell DeepSpeed do NOT create optimizer (prevent from potential OOM)
+    config_kwargs.pop("optimizer", None)
+
     # If ZeRO-3 is used, we shard both the active and reference model.
     # Otherwise, we assume the reference model fits in memory and is initialized on each device with ZeRO
     # disabled (stage 0)


liguohao96:
TL;DR: this patch address my own OOM, but others may not facing such OOM

This patch was to to address OOM issue when GRPOTrainer initialize ref_model with DeepSpeed.
Lots of time was spent to pin point such OOM. Finally find that DeepSpeed creates a optimizer for ref_model which was never used.
After reading DeepSpeed's doc and code, we could simple prevent such behavor with removing `optimizer` related config.

However, in the following days, I can not reproduce such OOM on another machine while writing testcase, maybe it is a hardware related bug?

qgallouedec:
Thanks for this finding!

> Lots of time was spent to pin point such OOM. Finally find that DeepSpeed creates an optimizer for ref_model which was never used.

Interesting, but does your modification also prevents DeepSpeed to create an optimizer for the trained model?

liguohao96:
Currently, there are many `prepare_deepspeed` (within different files) in TRL.


The first one is where this patch modifies, 
> https://github.com/huggingface/trl/blob/1755e03f6fcb39311f60059cb0db8cfe0edf1e27/trl/models/utils.py#L225
which is only used in GRPOTrainer with `self.ref_model = prepare_deepspeed(self.ref_model, self.accelerator)`. 
The modified `config_kwargs` is deep-copied, so this patch will not affect creating deepspeed optimizer for other models.

The second one is in here and used in many places
> https://github.com/huggingface/trl/blob/1755e03f6fcb39311f60059cb0db8cfe0edf1e27/trl/trainer/utils.py#L1240-L1242

The third type of `prepare_deepspeed` is defined inside the trainer like this one 
> https://github.com/huggingface/trl/blob/1755e03f6fcb39311f60059cb0db8cfe0edf1e27/trl/trainer/dpo_trainer.py#L679

Maybe we should put all these variants inside one file with different name?
Or like other trainers, make `_prepare_deepspeed` as a local function of GRPOTrainer?

qgallouedec:
Yes totally agree. We should merge all these function into one. Let's do this in a follow-up PR.

As for this PR, I've profiled it with different model sizes and for ZeRO stage 1 2 and 3, and I don't see any difference.

In all plots, green is for this PR, and red for main

### 1.5B ZeRO Stage 1
1 5B ZeRO1

### 1.5B ZeRO Stage 2
1 5B ZeRO2

### 1.5B ZeRO Stage 3
1 5B ZeRO3

### 7B ZeRO Stage 3
1 5B ZeRO1-2



```python
# 3228.py

dataset = load_dataset("trl-lib/tldr", split="train[:423]")

    return [len(set(c)) for c in completions]

trainer = GRPOTrainer(
    model="Qwen/Qwen2-7B-Instruct",
    train_dataset=dataset,
    args=GRPOConfig(bf16=True)
)
trainer.train()
```

```
accelerate launch --config_file examples/accelerate_configs/deepspeed_zero2.yaml sandbox/3228.py
```

liguohao96:
> Yes totally agree. We should merge all these function into one. Let's do this in a follow-up PR.
> 
> As for this PR, I've profiled it with different model sizes and for ZeRO stage 1 2 and 3, and I don't see any difference.
> 
> In all plots, green is for this PR, and red for main
> ...

Thanks for your great efforts @qgallouedec . This may happens under some rare condition. Feel free to close this PR.

qgallouedec:
Ok I close for now, if you find other info I'm interested.

PR #3188, opened 2025-03-31
[GRPO] Improve completion length logging

edbeeching:
# What does this PR do?

- Adds min, mean and max logging of completion lengths
- Also logs min, mean, max completion logs for sequences that terminate in EOS
- Adds clipped completion ratio from [SimpleRL-Zero](https://arxiv.org/pdf/2503.18892)
Example:
image

PR #3229, opened 2025-04-04
overlong filtering

shirinyamani:
# What does this PR do?

In RL training, we typically set a maximum length for generation, with overlong samples truncated accordingly. [DAPO](https://huggingface.co/papers/2503.14476) paper found that improper reward shaping for truncated samples can introduce reward noise and significantly disrupt the training process. To investigate the impact of this reward noise, [DAPO](https://huggingface.co/papers/2503.14476) paper first applied an Overlong Filtering strategy which masks the loss of truncated samples. They find that this approach significantly stabilizes training and enhances
performance. 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/docs/source/grpo_trainer.md
+++ b/docs/source/grpo_trainer.md
@@ -126,7 +126,7 @@ $$
 \text{clip}\left( \frac{\pi_\theta(o_{i,t} \mid q, o_{i,< t})}{\pi_{\theta_{\text{old}}}(o_{i,t} \mid q, o_{i,< t})}, 1 - \epsilon, 1 + \epsilon \right)
 $$
 A higher value means more tokens were affected by clipping, limiting how much the policy can change.
-- `mask_truncated_completions`: The completions that were truncated due to exceeding the maximum completion length. Therefore, excluded from the loss calculation.
+- `mask_truncated_completions`: The completions that were truncated due to exceeding the maximum completion length. Therefore, excluded from the loss calculation. This feature was introduced as a metric of training stability in the (DAPO paper)[https://huggingface.co/papers/2503.14476]
 - `epsilon_high`: The upper bound of the clipping range. If set to higher value (e.g. 0.28), the policy can have more exploration ability.
 ## Customization
 
--- a/docs/source/grpo_trainer.md
+++ b/docs/source/grpo_trainer.md
@@ -126,7 +126,8 @@ $$
 \text{clip}\left( \frac{\pi_\theta(o_{i,t} \mid q, o_{i,< t})}{\pi_{\theta_{\text{old}}}(o_{i,t} \mid q, o_{i,< t})}, 1 - \epsilon, 1 + \epsilon \right)
 $$
 A higher value means more tokens were affected by clipping, limiting how much the policy can change.
-
+- `mask_truncated_completions`: The completions that were truncated due to exceeding the maximum completion length. Therefore, excluded from the loss calculation.
+- `epsilon_high`: The upper bound of the clipping range. If set to higher value (e.g. 0.28), the policy can have more exploration ability.
 ## Customization
 
 ### Speed up training with vLLM-powered generation
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -117,7 +117,7 @@ class GRPOConfig(TrainingArguments):
             difficulty bias.
         mask_truncated_completions (`bool`, *optional*, defaults to `False`):
             Whether to mask the loss for completions that are truncated due to exceeding the maximum completion length.
-            When enabled, truncated completions are excluded from loss calculation, preventing them from being
+            When enabled, truncated completions are excluded from the loss calculation, preventing them from being
             incorrectly penalized and introducing noise during training.
         sync_ref_model (`bool`, *optional*, defaults to `False`):
             Whether to synchronize the reference model with the active model every `ref_model_sync_steps` steps, using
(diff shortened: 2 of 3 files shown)


qgallouedec (commented):
Looks good!!
A few comments/questions 

- ideally we would like some benchmarks to validate the results presented in the DAPO paper. But IMO it's ok to merge before we have these though
- can you add a unittest for this setting?
- what happens if all completions are truncated?
- should we exclude the samples from the reward calculation as well? So that it does bias the reward normalization?
- in the doc could you mention that this principle was added in the DAPO paper?
- could you add the new metrics to the "logged metrics" section of the GRPO doc

edbeeching (commented):
Thanks for the implementation, I will launch some benchmarks with it this morning.

I agree it makes sense to exclude truncated sequences from the advantage calculation as well. But perhaps that can be split into another PR?

Another comment, I think the implementation could be simplified by just directly masking the `completion_mask` in the  `_generate_and_score_completions` method, rather than having an additional `truncated_sampled` tensor. This would mean `compute_loss` will be exactly the same and the implementation should be compatible with the liger PR.

trl/trainer/grpo_config.py
```diff
@@ -115,6 +115,10 @@ class GRPOConfig(TrainingArguments):
             applied. The [Dr. GRPO](https://github.com/sail-sg/understand-r1-zero/blob/main/understand-r1-zero.pdf)
             paper recommends not scaling the rewards, as scaling by the standard deviation introduces a question-level
             difficulty bias.
+        mask_truncated_samples (`bool`, *optional*, defaults to `False`):
```
lewtun, line 118: turbo nit: since we use "completions" elsewhere in the config, let's align it here too
```suggestion
        mask_truncated_completions (`bool`, *optional*, defaults to `False`):
```

trl/trainer/grpo_config.py
```diff
@@ -115,6 +115,10 @@ class GRPOConfig(TrainingArguments):
...
+            Whether to mask the loss for samples that were truncated due to exceeding the maximum completion length.
+            When enabled, truncated completions are excluded from loss calculation, preventing them from being
```
lewtun, line 120:
```suggestion
            When enabled, truncated completions are excluded from the loss calculation, preventing them from being
```

trl/trainer/grpo_config.py
```diff
@@ -297,6 +301,14 @@ class GRPOConfig(TrainingArguments):
             "deviation introduces a question-level difficulty bias."
         },
     )
+    mask_truncated_samples: bool = field(
```
lewtun, line 304: Ditto here

trl/trainer/grpo_trainer.py
```diff
@@ -751,6 +752,15 @@ def _generate_and_score_completions(
         eos_idx[is_eos.any(dim=1)] = is_eos.int().argmax(dim=1)[is_eos.any(dim=1)]
         sequence_indices = torch.arange(is_eos.size(1), device=device).expand(is_eos.size(0), -1)
         completion_mask = (sequence_indices <= eos_idx.unsqueeze(1)).int()
+        # Identify truncated samples (completions that reached max_completion_length without generating an EOS token)
+        truncated_samples = torch.zeros(is_eos.size(0), dtype=torch.bool, device=device)
+        if self.mask_truncated_samples:
+            # A sample is truncated if it has no EOS token and its length equals max_completion_length
+            truncated_samples = ~is_eos.any(dim=1) & (completion_ids.size(1) == self.max_completion_length)
```
lewtun, line 759: Is the second requirement necessary? I'm trying to think of a case where a completion would have no EOS token, yet be shorter than `max_completion_length`

trl/trainer/grpo_trainer.py
```diff
@@ -751,6 +752,15 @@ def _generate_and_score_completions(
...
+            # Log the % of truncated samples 
+            truncated_pct = truncated_samples.float().mean().item() * 100
```
lewtun, line 761: If I'm not mistaken, we already have this metric as `completions/clipped_ratio` from #3188

trl/trainer/grpo_trainer.py
qgallouedec: I don't think the second requirement is necessary either

docs/source/grpo_trainer.md
```diff
@@ -126,7 +126,8 @@ $$
 \text{clip}\left( \frac{\pi_\theta(o_{i,t} \mid q, o_{i,< t})}{\pi_{\theta_{\text{old}}}(o_{i,t} \mid q, o_{i,< t})}, 1 - \epsilon, 1 + \epsilon \right)
 $$
 A higher value means more tokens were affected by clipping, limiting how much the policy can change.
-
+- `mask_truncated_completions`: The completions that were truncated due to exceeding the maximum completion length. Therefore, excluded from the loss calculation. This feature was introduced as a metric of training stability in the (DAPO paper)[https://huggingface.co/papers/2503.14476]
+- `epsilon_high`: The upper bound of the clipping range. If set to higher value (e.g. 0.28), the policy can have more exploration ability.
```
qgallouedec, line 130: is it logged somewhere?

docs/source/grpo_trainer.md
```diff
@@ -126,7 +126,8 @@ $$
 \text{clip}\left( \frac{\pi_\theta(o_{i,t} \mid q, o_{i,< t})}{\pi_{\theta_{\text{old}}}(o_{i,t} \mid q, o_{i,< t})}, 1 - \epsilon, 1 + \epsilon \right)
 $$
 A higher value means more tokens were affected by clipping, limiting how much the policy can change.
-
+- `mask_truncated_completions`: The completions that were truncated due to exceeding the maximum completion length. Therefore, excluded from the loss calculation. This feature was introduced as a metric of training stability in the (DAPO paper)[https://huggingface.co/papers/2503.14476]
```
qgallouedec, line 129: I think you don't log it anymore

PR #3231, opened 2025-04-04
[GRPO] Adds an option to scale the loss by a constant factor

edbeeching:
# What does this PR do?
 This PR adds an option to scale the loss by a constant factor, equal to the maximum possible tokens in a batch.

The reasoning behind this PR is that I believe that the current normalization scheme, implemented in #2881, is not invariant to the ordering of samples across devices / gradient accumulation steps. Which may cause instabilities in training.

### Toy Example

Consider a `DDP=2` setting with a `per_device_train_batch_size=4`. For this example, assume that the loss per token is 1.
With global normalization: 
image

Here each token has an equal contribution the loss, but only inside the current device. The loss is not comparable to a setting where all the batch is on a single device, for example consider a `DDP=1` setting with `per_device_train_batch_size=4`

image

One potential solution would be to gather the number of unmasked tokens across all devices and use this for normalization. But the same issue would also occur across gradient accumulation steps.

### Proposed solution
Calculate a constant factor `max_tokens_norm = per_device_train_batch_size * (max_prompt_length +max_completion_length)` and always normalize the loss by this constant factor. 

image

The learning rate will probably need to be increased to get comparable results with our other baselines.

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -101,6 +101,8 @@ class GRPOConfig(TrainingArguments):
             speed, but may be numerically unstable for long training runs.
         num_iterations (`int`, *optional*, defaults to `1`):
             Number of iterations per batch (denoted as μ in the algorithm).
+        use_max_tokens_norm (`bool`, *optional*, defaults to `False`):
+            Whether to use the max tokens norm. If `True`, the loss is normalized by a consant, the maximum possible number of tokens
         epsilon (`float`, *optional*, defaults to `0.2`):
             Epsilon value for clipping.
         epsilon_high (`float` or `None`, *optional*, defaults to `None`):
@@ -274,6 +276,13 @@ class GRPOConfig(TrainingArguments):
     num_iterations: int = field(
         default=1,
         metadata={"help": "Number of iterations per batch (denoted as μ in the algorithm)."},
+    )
+    use_max_tokens_norm: bool = field(
+        default=False,
+        metadata={
+            "help": "Whether to use the max tokens norm. If `True`, the loss is normalized by a constant, the maximum "
+            "possible number of tokens."
+        },
     )
     epsilon: float = field(
         default=0.2,
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -409,7 +409,7 @@ class GRPOTrainer(Trainer):
         self.use_vllm = args.use_vllm
         self.use_liger_loss = args.use_liger_loss
 
-        self.use_max_tokens_norm = self.args.use_max_tokens_norm
+        self.use_max_tokens_norm = args.use_max_tokens_norm
         if self.use_max_tokens_norm:
             if self.use_liger_loss:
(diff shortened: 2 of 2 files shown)


trl/trainer/grpo_config.py
```diff
@@ -101,6 +101,8 @@ class GRPOConfig(TrainingArguments):
             speed, but may be numerically unstable for long training runs.
         num_iterations (`int`, *optional*, defaults to `1`):
             Number of iterations per batch (denoted as μ in the algorithm).
+        use_max_tokens_norm (`bool`, *optional*, defaults to `False`):
+            Whether to use the max tokens norm. If `True`, the loss is normalized by a consant, the maximum possible number of tokens
```
lewtun, line 105: Suggestion to clarify what we mean by "maximum possible"
```suggestion
            Whether to use the max tokens norm. If `True`, the loss is normalized by a constant factor that is determined by the total number of prompt and completions tokens in a batch.
```

trl/trainer/grpo_config.py
```diff
@@ -275,6 +277,13 @@ class GRPOConfig(TrainingArguments):
         default=1,
         metadata={"help": "Number of iterations per batch (denoted as μ in the algorithm)."},
     )
+    use_max_tokens_norm: bool = field(
+        default=False,
+        metadata={
+            "help": "Whether to use the max tokens norm. If `True`, the loss is normalized by a constant, the maximum "
```
lewtun, line 283: Ditto here if you agree with the change above

trl/trainer/grpo_trainer.py
```diff
@@ -1072,7 +1081,11 @@ def _compute_loss(self, model, inputs):
         per_token_loss = -torch.min(per_token_loss1, per_token_loss2)
         if self.beta != 0.0:
             per_token_loss = per_token_loss + self.beta * per_token_kl
-        loss = (per_token_loss * completion_mask).sum() / completion_mask.sum()
+
+        if self.use_max_tokens_norm:
+            loss = (per_token_loss * completion_mask).sum() / self.max_tokens_norm
```
lewtun, line 1086: I'm not sure how easy it is to unit test this, but would it make sense to do it so that we're sure the loss is being computed as your diagrams show?

E.g. an integration test would be to check that specifying the config params gives the expected scaling for some dummy inputs

lewtun (approved):
LGTM with a nit and suggestion on whether we should unit test the scaling

trl/trainer/grpo_config.py
```diff
@@ -101,6 +101,8 @@ class GRPOConfig(TrainingArguments):
             speed, but may be numerically unstable for long training runs.
         num_iterations (`int`, *optional*, defaults to `1`):
             Number of iterations per batch (denoted as μ in the algorithm).
+        use_max_tokens_norm (`bool`, *optional*, defaults to `False`):
```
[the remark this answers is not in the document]
qgallouedec: Currently we use a modified version of DAPO where we normalize per local batch (and not per group).

Gm-URXOagAABJH3

It the above figure, we use something between BNPO (hard to implement with grad accum) and DAPO

edbeeching:
closing in favor of #3256

PR #3234, opened 2025-04-04
👎 [GRPO] Adds option to disable dropout

edbeeching:
# What does this PR do?
Adds an option to disable dropout. 

The RLOOTrainer disables dropout in policy, ref_model and reward model. This PR adds the option to disable dropout to the GRPOTrainer, which may improve training stability. 

[The N+ Implementation Details of RLHF with PPO: A Case Study on TL;DR Summarization](https://arxiv.org/pdf/2403.17031)
Provides more insight about why this option may improve training stability:

> We disable the dropout layers during training, similar to the settings in Ziegler et al. (2019); Huang
et al. (2024). This is important for PPO training, especially because with dropout activated, the log
probabilities of tokens will not be reproducible, making calculating the KL penalty unreliable while
also causing the ratios of the PPO to be not 1s during the first epoch, causing PPO optimization
problems. For consistency, we also disable dropout for SFT and RM training.

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -38,6 +38,9 @@ class GRPOConfig(TrainingArguments):
         model_init_kwargs (`str`, `dict[str, Any]` or `None`, *optional*, defaults to `None`):
             Keyword arguments for [`~transformers.AutoModelForCausalLM.from_pretrained`], used when the `model`
             argument of the [`GRPOTrainer`] is provided as a string.
+        disable_dropout (`bool`, *optional*, defaults to `False`):
+            Whether to disable dropout in the model. This is useful for training with a reference model, as it
+            prevents the model from generating different logprobs for the same input.
 
         > Parameters that control the data preprocessing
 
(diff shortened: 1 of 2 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -359,6 +366,10 @@ def __init__(
                 reward_funcs[i] = AutoModelForSequenceClassification.from_pretrained(
                     reward_func, num_labels=1, **model_init_kwargs
                 )
+                if args.disable_dropout:
+                    if isinstance(reward_funcs[i], nn.Module):
```
lewtun, line 370: I think `AutoModelForSequenceClassification` is loaded in eval model by default, so technically we don't need this here (happy to keep it though if we want to be safe)

lewtun (approved):
Nice feature - LGTM.

trl/trainer/grpo_config.py
```diff
@@ -101,6 +101,9 @@ class GRPOConfig(TrainingArguments):
             speed, but may be numerically unstable for long training runs.
         num_iterations (`int`, *optional*, defaults to `1`):
             Number of iterations per batch (denoted as μ in the algorithm).
+        disable_dropout (`bool`, *optional*, defaults to `False`):
```
qgallouedec, line 104: In the other trainers this is set to `True`, maybe we should do the same here?

trl/trainer/grpo_trainer.py
```diff
@@ -347,6 +348,12 @@ def __init__(
             # If PEFT configuration is not provided, create a reference model based on the initial model.
             self.ref_model = create_reference_model(model)
 
+        if args.disable_dropout:
+            if isinstance(model, nn.Module):
+                disable_dropout_in_model(model)
+            if self.ref_model is not None and isinstance(self.ref_model, nn.Module):
+                disable_dropout_in_model(self.ref_model)
+
```
qgallouedec, line 356: we only support `PreTrainedModel`, which are `nn.Module` (what else could it be?)

```suggestion
        if args.disable_dropout:
            disable_dropout_in_model(model)
            if self.ref_model is not None:
                disable_dropout_in_model(self.ref_model)

```

qgallouedec (approved):
LGTM, in the future we could use a default to True (let's see if it improve stability)

PR #3240, opened 2025-04-05
🔭 Add support for better KL estimator (k3) in PPOTrainer

AMindToThink:
[In this blog post](http://joschu.net/blog/kl-approx.html) by John Schulman, he describes three estimators for KL divergence: k1, k2, and k3. He finds that k3 appears to be strictly better than k1 and k2 for both theoretical and empirical reasons. 

Previously the PPOTrainer was using k1 to estimate KL divergence. This pull request adds support, through an argument in PPOConfig, `kl_estimator`, to decide between "k1" and "k3". 

k3 is unbiased and has less variance than k1.
Additionally, k1 is sometimes negative, which is an undesirable property to have when estimating KL divergence, a quantity which is always positive. k3 is always positive.

I did not add support for k2 because the PPOTrainer was already using that for logging purposes.

Notes: 
- I wrote descriptions of the new variable `kl_estimator` in the scripts, and if I'm understanding correctly, [[autodoc]] will add those descriptions to the documentation.
- I tested by running test_ppo_trainer.py with both default and `kl_estimator='k3'`, but did not feel the need to clutter the tests by writing another one for this one small change.
- For historical reasons, I set "k1" to be the default, so everyone's scripts will run the same way as before after the update. We should consider setting "k3" to be the default because it is lower variance, unbiased, and trivial to compute (it just takes an exp and two subtractions).

## Before submitting
      Pull Request section?
      to it if that's the case.


trl/trainer/ppo_config.py
```diff
@@ -43,6 +43,8 @@ class PPOConfig(OnPolicyConfig):
             Whether to whiten the rewards.
         kl_coef (`float`, *optional*, defaults to `0.05`):
             KL coefficient.
+        kl_estimator (Literal["k1", "k3"], *optional*, defaults to `k1`):
```
qgallouedec, line 46:
```suggestion
        kl_estimator (Literal["k1", "k3"], *optional*, defaults to `"k1"`):
```

trl/trainer/ppo_config.py
```diff
@@ -43,6 +43,8 @@ class PPOConfig(OnPolicyConfig):
...
+            Which estimator for KL-Divergence to use from [Approximating KL Divergence](http://joschu.net/blog/kl-approx.html). Defaults to "k1", a straightforward, unbiased estimator. Can be set to "k3", an unbiased estimator with lower variance which "appears to be a strictly better estimator." Cannot be set to "k2", as it is used for logging purposes.
```
qgallouedec, line 47:
```suggestion
            Which estimator for KL-Divergence to use from [Approximating KL Divergence](http://joschu.net/blog/kl-approx.html).
            Defaults to "k1", a straightforward, unbiased estimator. Can be set to "k3", an unbiased estimator with
            lower variance which "appears to be a strictly better estimator". Cannot be set to "k2", as it is used for
            logging purposes.
```

trl/trainer/ppo_config.py
```diff
@@ -87,6 +89,12 @@ class PPOConfig(OnPolicyConfig):
         default=0.05,
         metadata={"help": "KL coefficient."},
     )
+    kl_estimator: Literal["k1", "k3"] = field(
+        default="k1",
+        metadata={
+            "help": "Which estimator for KL-Divergence to use - either 'k1' (straightforward, unbiased) or 'k3' (lower variance, unbiased, appears to be a strictly better estimator)."
```
qgallouedec, line 98:
```suggestion
            "help": "Which estimator for KL-Divergence to use from Approximating KL Divergence. Defaults to 'k1', a "
            "straightforward, unbiased estimator. Can be set to 'k3', an unbiased estimator with lower variance which "
            "'appears to be a strictly better estimator'. Cannot be set to 'k2', as it is used for logging purposes."
```

AMindToThink: Do you also want to include the link http://joschu.net/blog/kl-approx.html?

trl/trainer/ppo_trainer.py
```diff
@@ -142,6 +142,11 @@ def __init__(
         else:
             self.policy_model.generation_config.eos_token_id = self.stop_token_id = args.stop_token_id  # None or int
 
+        # Check that the kl estimator is valid
+        assert self.args.kl_estimator in {"k1", "k3"}, (
+            "kl_estimator must be either 'k1' (straightforward, unbiased) or 'k3' (lower variance, unbiased, appears to be a strictly better estimator). See [Approximating KL Divergence](http://joschu.net/blog/kl-approx.html) for details."
+        )
```
qgallouedec, line 148:
```suggestion
        if self.args.kl_estimator not in {"k1", "k3"}:
            raise ValueError(
                "kl_estimator must be either 'k1' (straightforward, unbiased) or 'k3' (lower variance, unbiased, "
                "appears to be a strictly better estimator). See "
                "[Approximating KL Divergence](http://joschu.net/blog/kl-approx.html) for details."
            )
```

qgallouedec (approved):
Nice, thanks @AMindToThink!
A few format comments, otherwise we're good to merge

trl/trainer/ppo_config.py
```diff
@@ -43,6 +43,11 @@ class PPOConfig(OnPolicyConfig):
             Whether to whiten the rewards.
         kl_coef (`float`, *optional*, defaults to `0.05`):
             KL coefficient.
+        kl_estimator (Literal["k1", "k3"], *optional*, defaults to `"k1"`):
```
qgallouedec, line 46:
```suggestion
        kl_estimator (`Literal["k1", "k3"]`, *optional*, defaults to `"k1"`):
```

qgallouedec:
Waiting for the CI to pass

AMindToThink:
Pleasure working with you, @qgallouedec . Expect more PRs from me in the future!

qgallouedec:
> For historical reasons, I set "k1" to be the default, so everyone's scripts will run the same way as before after the update. We should consider setting "k3" to be the default because it is lower variance, unbiased, and trivial to compute (it just takes an exp and two subtractions).

If you can provide results showing that in the general case, k3 is a better choice, we can change the default value (in a follow up PR).

AMindToThink:
How would you go about showing that for the general case? 
For one, what experiments would I do? I could find three projects which use PPOTrainer, and rerun the code with both estimators, but that is hardly showing that kx is better in the general case.
And how would I evaluate results? KL is the measure of coherence, and I wouldn't just be able to use one of the KL estimators themselves because that would be unfair to the other (unless both estimators agreed that the kx estimator results in lower KL divergence)?

qgallouedec:
What's important here is to know whether using k3 makes it possible to converge faster and/or towards better models.
My advice would be to take 2 or 3 different configurations (datasets/models) and run the training, then evaluate the checkpoints (the type of evaluation depends on the dataset).

AMindToThink:
@qgallouedec 
I notice that other scripts, such as DPOTrainer use [cap_exp](https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L62) instead of exp as a protection against exponentiating extreme values. Should we use that here, too?

I would think it should be unnecessary, because PPOTrainer is on-policy, so logprobs cannot be huge (since that would imply sampling a really unlikely generation/token), but I thought I'd check in.

qgallouedec:
I'd say, unless we encounter this issue, we can keep the code as is

PR #3260, opened 2025-04-08
[🐯+GRPO] Support FSDP + Fix bug when using LigerGRPO with DDP

shivam15s:
# What does this PR do?
This PR aims to do two things:
1. The recent integration of LigerGRPO had a bug: when using DDP and performing a forward pass through a submodule of the unwrapped model, the necessary hooks weren't registered correctly. This caused the model weights across GPUs to fall out of sync. To fix this, the PR introduces a Forward Redirection mechanism—a workaround that ensures hooks are properly registered (compatible with both DDP and FSDP) and enables the custom forward pass required by Liger.
2. Add support for FSDP to GRPO Trainer. We leverage summon_full_params to make model.generate work with FSDP. 

Experiment Script: https://gist.github.com/shivam15s/08a9bccd0d72dd0d29bdb912cb9885be

DDP: Liger (blue) v Non-liger (black)
image

FSDP: Liger (green) v Non-liger (Purple)
image

**Known Limitations with FSDP (can add support in subsequent PR(s))** 
1. sync_ref_model not supported currently
2. create_reference_model not supported currently

**Benchmarking:**
Dist Strategy: DDP
7 policy workers, 1 vllm worker (8 h100)
image

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


shivam15s:
Python 3.9 test is failing because of a backward incompatible initialization. Created a PR in liger to fix this

hjh0119:
Awesome work! Could you share any memory footprint comparisons (with vs. without Liger Loss enabled)?

shivam15s:
Hi @hjh0119 
I was able to achieve about a 20% reduction in peak memory usage.

**Peak Memory:**  
• Without Liger: 71.1 GB  
• With Liger: 56.7 GB
here's my code: https://gist.github.com/shivam15s/08a9bccd0d72dd0d29bdb912cb9885be

kashif (approved):

kashif (approved):

trl/trainer/grpo_trainer.py
```diff
@@ -407,7 +410,7 @@ def __init__(
         if self.beta == 0.0:
             # If beta is 0.0, the reference model is not needed
             self.ref_model = None
-        elif is_deepspeed_zero3_enabled():
+        elif is_deepspeed_zero3_enabled() or args.fsdp_config is not None:
```
LeonEricsson, line 413: `args.fsdp_config` defaults to

`{'min_num_params': 0, 'xla': False, 'xla_fsdp_v2': False, 'xla_fsdp_grad_ckpt': False}`

when fsdp is not enabled. Probably want `args.fsdp` instead?

kashif: checking

LeonEricsson: ran with

```

def main():
    # Load dataset
    train_dataset = load_dataset("trl-lib/tldr", split="train[:128]")

    # Train model
    training_args = GRPOConfig(
        output_dir=f"{model_id}-codeforces-SFT",
        logging_steps=10,
        bf16=True,
        # use_liger_loss=True,
        max_prompt_length=250,
        max_completion_length=250,
        per_device_eval_batch_size=2,
        num_generations=2,
        num_train_epochs=1,
        do_eval=True,
        optim="paged_adamw_8bit",
        max_steps=10,
        report_to="none",
    )

    print(training_args.fsdp_config)

    trainer = GRPOTrainer(
        args=training_args,
        train_dataset=train_dataset,
        eval_dataset=train_dataset,
        reward_funcs=reward_len,
    )

    trainer.train()


if __name__ == "__main__":
    main()

Comment (lewtun) [inline trl/trainer/grpo_trainer.py:413]: Good catch @LeonEricsson ! Since we inherit from `transformers.Trainer` I believe this should work:
```suggestion
        elif is_deepspeed_zero3_enabled() or self.is_fsdp_enabled is not None:
```

lewtun (approved):
Thanks for the nice PR @shivam15s ! LGTM with a small change on how we determine if FSP is enable for the ref model

kashif:
@LeonEricsson  I have integrated the commit from @jglaser here

trl/trainer/grpo_trainer.py
LeonEricsson: but `Trainer` isn't initialized until [later](https://github.com/huggingface/trl/blob/9fbf125451e8e45547f4c4174573e16f0e35354d/trl/trainer/grpo_trainer.py#L551)

kashif: yes i am getting the same error too

trl/trainer/grpo_trainer.py
```diff
@@ -886,35 +837,78 @@ def _move_model_to_vllm(self):
...
-
-                # Unmerge adapters while parameters are still gathered
-                self.model.unmerge_adapter()
-                # Parameters will automatically be repartitioned when exiting the context
+                if self.is_fsdp_enabled:
+                    # Update vLLM weights while parameters are gathered
+                    # For PEFT with FSDP we need to use the memory efficient post-order traversal
+                    self.model.merge_adapter()
+                    post_order_fsdp_processing(self.model)
+                    self.model.unmerge_adapter()
```
nph4rd, line 884: Is it really necessary to use the `post_order_fsdp_processing` function here? isn't  `gather_if_zero3(list(self.model.parameters()))` gathering ALL the parameters already?

LeonEricsson: `gather_if_zero3` is `nullcontext` when using FSDP. You can't use ZeRO-3 and FSDP at the same time.

nph4rd: That makes sense!

thepowerfuldeez:
Are there plans to support FSDP2?

PR #3266, opened 2025-04-09
🐯 `is_liger_kernel_available` with min version

qgallouedec:
From @lewtun 

> FYI I just install `trl@main` and got this from trying to run GRPO:
> 
> ```
>   File "/fsx/lewis/git/hf/dev/open-r1/openr1/bin/trl", line 4, in <module>
>     from trl.cli import main
>   File "/fsx/lewis/git/hf/dev/open-r1/openr1/lib/python3.11/site-packages/trl/cli.py", line 24, in <module>
>     from .scripts.grpo import make_parser as make_grpo_parser
>   File "/fsx/lewis/git/hf/dev/open-r1/openr1/lib/python3.11/site-packages/trl/scripts/grpo.py", line 22, in <module>
>     from trl import GRPOConfig, GRPOTrainer, ModelConfig, ScriptArguments, TrlParser, get_peft_config
>   File "<frozen importlib._bootstrap>", line 1229, in _handle_fromlist
>   File "/fsx/lewis/git/hf/dev/open-r1/openr1/lib/python3.11/site-packages/trl/import_utils.py", line 127, in __getattr__
>     value = getattr(module, name)
>             ^^^^^^^^^^^^^^^^^^^^^
>   File "/fsx/lewis/git/hf/dev/open-r1/openr1/lib/python3.11/site-packages/trl/import_utils.py", line 126, in __getattr__
>     module = self._get_module(self._class_to_module[name])
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/fsx/lewis/git/hf/dev/open-r1/openr1/lib/python3.11/site-packages/trl/import_utils.py", line 138, in _get_module
>     raise RuntimeError(
> RuntimeError: Failed to import trl.trainer.grpo_trainer because of the following error (look up to see its traceback):
> cannot import name 'LigerFusedLinearGRPOLoss' from 'liger_kernel.chunked_loss' (/fsx/lewis/git/hf/dev/open-r1/openr1/lib/python3.11/site-packages/liger_kernel/chunked_loss/__init__.py)
> ```
> 
> The solution is to bump liger-kernel like [here](https://github.com/huggingface/trl/blob/1d7b8c4f706a0df9fd4d3e8d0d576e8dd5dedb28/setup.py#L84C16-L84C35) but it's odd that doing a new install of trl didn't catch this (I had an older version of liger installed)

From me:

> I think it's because you did `pip install trl` and not `pip install trl[liger]`. Maybe the solution to avoid this is to also check the version in `is_liger_available` (returns `True` only if the min version is installed) , and more generally in all `is_xxx_availalbe`.

---

I suggest the following solution. Now when you run `is_liger_kernel_available` without the min version, it returns `False`.

1. It's a shame that you have to specify the min version in two places (`setup.py` and `import_utils.py`), but I can't think of any other easy and readable way to do it.

2. This implementation actually matches that of transformers, which, as everyone knows, is the source of all truth 🙇: https://github.com/huggingface/transformers/blob/main/src/transformers/utils/import_utils.py

3. We could do this with all our optional dependencies. But it's a bit time-consuming, so I'll leave it to follow-up PRs.

### DIFF
--- a/trl/import_utils.py
+++ b/trl/import_utils.py
@@ -28,7 +28,7 @@ LIGER_KERNEL_MIN_VERSION = "0.5.6"
 _deepspeed_available = _is_package_available("deepspeed")
 _diffusers_available = _is_package_available("diffusers")
 _fastapi_available = _is_package_available("fastapi")
-_is_liger_kernel_available, _liger_kernel_version = _is_package_available("liger-kernel", return_version=True)
+_is_liger_kernel_available, _liger_kernel_version = _is_package_available("liger_kernel", return_version=True)
 _llm_blender_available = _is_package_available("llm_blender")
 _mergekit_available = _is_package_available("mergekit")
 _pydantic_available = _is_package_available("pydantic")
--- a/trl/import_utils.py
+++ b/trl/import_utils.py
@@ -18,13 +18,17 @@ from itertools import chain
 from types import ModuleType
 from typing import Any
 
+from packaging import version
 from transformers.utils.import_utils import _is_package_available
 
+
+LIGER_KERNEL_MIN_VERSION = "0.5.6"
 
 # Use same as transformers.utils.import_utils
 _deepspeed_available = _is_package_available("deepspeed")
 _diffusers_available = _is_package_available("diffusers")
 _fastapi_available = _is_package_available("fastapi")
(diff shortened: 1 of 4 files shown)


kashif (approved):

lewtun (approved):
Thanks for the fix, LGTM!

PR #3280, opened 2025-04-11
add vllm support for token ids as input

wybryan:
# What does this PR do?

This PR adds support for vLLM server & client to accept token ids as input. 

The vLLM engine does support token ids as input as opposed to text under the hood, this PR makes the feature available to end users. The rationale behind is that in certain training user cases user wants precise control of what exact input tokens are used for vLLM, therefore adding this capability to let user have full control of input tokens before sending to vLLM. 

## Who can review?

### DIFF
--- a/trl/extras/vllm_client.py
+++ b/trl/extras/vllm_client.py
@@ -15,7 +15,7 @@
 import atexit
 import logging
 import time
-from typing import Optional
+from typing import Optional, Union
 
 import torch
 from torch import nn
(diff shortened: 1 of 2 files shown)


qgallouedec:
Hi, so sorry for the late review. We are trying to mimic as much as possible de vLLM server. Do you know if it supports it? If so, how?

wybryan:
The vLLM engine natively supports input token ids instead of input strings. My PR just expose this feature from the wrapper in TRL. > Hi, so sorry for the late review. We are trying to mimic as much as possible de vLLM server. Do you know if it supports it? If so, how?

qgallouedec:
After several months of stabilization, we have decided to use `generate` for the standard format and `chat` for the conversational format. Since the vLLM server is not intended to be part of the public API, we are designing its features solely to meet the needs of the online trainers who use it. Therefore, we will not support the transfer of credentials to the server.

Issue #3017, opened 2025-03-06
GRPO split generations into multiple training batches


JamesBowerXanda:
### Feature request

In the GRPO training it would be useful if you could split the generations into smaller batches for the gradient calculations similar to how we split batches into multiple gradient calculations with gradient_accumulation_steps.

I am imagining the config to work something like this:

```
per_device_train_batch_size = 4
num_generations = 8
gradient_accumulation_steps = 4
```

with the condition that `per_device_train_batch_size * gradient_accumulation_steps` is a multiple of `num_generations`. 

### Motivation

In the GRPO algorithm the loss calculation (ignoring the KL part) is an estimation of an expectation under the current models distribution. This will have very high variance if we are limiting the sample size (number of generations) to small numbers giving us a poor estimation of the expectation and therefore making training less stable.

Currently `per_device_train_batch_size` must be a multiple of `num_generations` which can severely limit how large you can make it before hitting OOM particularly when in resource constrained environments working with long context windows. This seems like an unnecessary restriction since nothing in the algorithm stops us from splitting the gradient calculation of a generation batch into multiple smaller batches.

### Your contribution

I don't think I would be able to create the PR myself unfortunately.

JamesBowerXanda:
Ah yes, sorry I forgot about number of devices. Though this doesn't change much right, we just amend my statement to

`num_devices * per_device_train_batch_size * gradient_accumulation_steps ` must be a multiple of `num_generations`.

Is it complicated because currently the prepare_inputs method does both the generation and score calculation then the inputs are passed straight to the compute_loss  method by the Trainer superclass?

I can see how it could cause more issues than it is worth having to fiddle with the core pipeline just for one trainer. I just thought I would bring it because I noticed how much smoother the training seemed when I was able to up the number of generations using smaller models and this seemed to be the big bottleneck to that.

JamesBowerXanda:
Ok, I understand, thanks for your prompt responses.

Unfortunately I am most interested in using this on my personal gpu so I am not using multiple gpu clusters.

Thanks for your time, I am happy for the issue to be closed since it is not deemed feasible.

qgallouedec:
With 1 GPU, the best you can do is to set `num_generations=per_device_train_batch_size`, and set the `gradient_accumulation_steps` depending on the desired effective batch size. Example:

```
per_device_train_batch_size = 8
num_generations = 8
gradient_accumulation_steps = 16
```

To have an effective batch size of 128

[...]

PR #3283, opened 2025-04-12
☝️ [GRPO] Generate once per effective batch

qgallouedec:
# What does this PR do?

#### Summary

This PR optimizes the generation process in GRPO when using gradient accumulation by ensuring that generation happens only once per effective batch, rather than once per micro-batch. This significantly improves performance—particularly with large gradient accumulation values—by leveraging the efficiency of vLLM with large batch sizes.

#### Context

Currently, when using GRPO with gradient accumulation, generation is triggered at every micro-batch. For instance, with a gradient accumulation factor of 16, generation is launched 16 times before a single network update. This is inefficient, especially given that vLLM performs better with larger batches.

#### What this PR changes

This PR modifies the logic so that generation occurs only once per *effective batch* (i.e., after all gradient accumulation steps). To enable this:
- The dataloader is adjusted to repeat the same batch `GAS` times (where `GAS` is the number of gradient accumulation steps).
- Generation is performed only during the first micro-batch.
- The actual sampled batch is only a slice of the loaded data

Screenshot 2025-04-12 at 09 58 41

#### Why it matters

This leads to a speedup in training, particularly with higher accumulation values, by making better use of vLLM’s batched generation capabilities.

> Note: This is a preparatory change. The full performance gains will be realized once the TRL vLLM server supports distributed parallelism (DP).

#### Review Notes

This change involves more complex indexing logic, which may make the diff harder to review. However, the code has been heavily commented for clarity and maintainability.

## Benchmark

Training time for 9 configs:
Note that for GAS=1, no speedup is expected

time_comparison

### Code used:

```python
import argparse
import subprocess

def get_git_commit_hash():
    try:
        commit_hash = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('utf-8').strip()[:7]
        return commit_hash
    except subprocess.CalledProcessError:
        return None

if __name__ == "__main__":
    parser = argparse.ArgumentParser(description="GRPO Trainer")
    parser.add_argument("--gradient_accumulation_steps", type=int, default=1)
    parser.add_argument("--per_device_train_batch_size", type=int, default=1)
    parser.add_argument("--num_iterations", type=int, default=1)
    args = parser.parse_args()

    dataset = load_dataset("trl-lib/tldr", split="train[:2048]")

        return [len(set(c)) for c in completions]

    git_commit = get_git_commit_hash()
    run_name = f"GRPO_7B_{git_commit}_GAS_{args.gradient_accumulation_steps}_BS_{args.per_device_train_batch_size}_MU_{args.num_iterations}"

    trainer = GRPOTrainer(
        model="Qwen/Qwen2.5-7B",
        train_dataset=dataset,
        args=GRPOConfig(
            run_name=run_name,
            logging_steps=10,
            bf16=True,
            max_completion_length=128,
            gradient_checkpointing=True,
            use_vllm=True,
            num_iterations=args.num_iterations,
            gradient_accumulation_steps=args.gradient_accumulation_steps,
            per_device_train_batch_size=args.per_device_train_batch_size,
            vllm_server_timeout=360,
        ),
    )
    trainer.train()
```

```sh
#!/bin/bash
#SBATCH --output=/fsx/qgallouedec/logs/%x-%j.out
#SBATCH --error=/fsx/qgallouedec/logs/%x-%j.err
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=1
#SBATCH --gres=gpu:8
#SBATCH --time=10:00:00
#SBATCH --qos=normal


# Run vLLM server
trl vllm-serve --model Qwen/Qwen2.5-7B --tensor_parallel_size 4 &

# Run trainin
CUDA_VISIBLE_DEVICES=4,5,6,7 accelerate launch \
     --config_file examples/accelerate_configs/deepspeed_zero3.yaml \
     --num_processes 4 \
     sandbox/3283.py \
     --gradient_accumulation_steps 16 \
     --per_device_train_batch_size 8 \
     --num_iterations 4
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -529,29 +529,35 @@ class GRPOTrainer(Trainer):
             "rewards": defaultdict(lambda: deque(maxlen=maxlen)),
         }
 
-        # Check if the per_device_train/eval_batch_size * num processes can be divided by the number of generations
-        num_processes = self.accelerator.num_processes
-        global_batch_size = args.per_device_train_batch_size * num_processes
-        possible_values = [n_gen for n_gen in range(2, global_batch_size + 1) if (global_batch_size) % n_gen == 0]
+        # Check if the effective batch size can be divided by the number of generations
         if self.num_generations < 2:
             raise ValueError(
-                f"GRPO requires at least 2 generations per prompt to calculate the advantages. "
-                f"You provided {self.num_generations}, which is less than the minimum required."
+                "GRPO requires at least 2 generations per prompt to calculate the advantages. You provided "
+                f"{self.num_generations}, which is less than the minimum required."
             )
+        num_processes = self.accelerator.num_processes
+        effective_batch_size = args.per_device_train_batch_size * num_processes * args.gradient_accumulation_steps
+        possible_values = [
+            n_gen for n_gen in range(2, effective_batch_size + 1) if (effective_batch_size) % n_gen == 0
+        ]
         if self.num_generations not in possible_values:
             raise ValueError(
-                f"The global train batch size ({num_processes} x {args.per_device_train_batch_size}) must be evenly "
-                f"divisible by the number of generations per prompt ({self.num_generations}). Given the current train "
-                f"batch size, the valid values for the number of generations are: {possible_values}."
+                f"The effective train batch size ({num_processes} x {args.per_device_train_batch_size} x "
+                f"{args.gradient_accumulation_steps}) must be evenly divisible by the number of generations per "
+                f"prompt ({self.num_generations}). Given the current effective train batch size, the valid values for "
+                f"the number of generations are: {possible_values}."
             )
         if self.args.eval_strategy != "no":
-            global_batch_size = args.per_device_eval_batch_size * num_processes
-            possible_values = [n_gen for n_gen in range(2, global_batch_size + 1) if (global_batch_size) % n_gen == 0]
+            effective_batch_size = args.per_device_eval_batch_size * num_processes
+            possible_values = [
+                n_gen for n_gen in range(2, effective_batch_size + 1) if (effective_batch_size) % n_gen == 0
+            ]
             if self.num_generations not in possible_values:
                 raise ValueError(
-                    f"The global eval batch size ({num_processes} x {args.per_device_eval_batch_size}) must be evenly "
-                    f"divisible by the number of generations per prompt ({self.num_generations}). Given the current "
-                    f"eval batch size, the valid values for the number of generations are: {possible_values}."
+                    f"The effective eval batch size ({num_processes} x {args.per_device_eval_batch_size}) must be "
+                    f"evenly divisible by the number of generations per prompt ({self.num_generations}). Given the "
+                    "current effective eval batch size, the valid values for the number of generations are: "
+                    f"{possible_values}."
                 )
 
         # Ensure each process receives a unique seed to prevent duplicate completions when generating with
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -671,19 +671,20 @@ class GRPOTrainer(Trainer):
 
         # In the following figure, the values are the prompt indices. The first row shows the first sampled batch, the
         # second row shows the second sampled batch, and so on.
-        #                                    |      Iteration 0      |      Iteration 1      |
-        #                                    |   GPU 0   |   GPU 1   |   GPU 0   |   GPU 1   |
         #
-        #               global_step   step    <-───>  num_generations=2
-        #                                     <-───────> per_device_train_batch_size=3
-        #  grad_accum=2  ▲   0          0     [0   0   1   1   2   2]  3   3   4   4   5   5   ▲
-        #                ▼   0          1      0   0   1   1   2   2 [ 3   3   4   4   5   5]  |
-        #                                                                                      |  num_iterations=2
-        #                    1          2     [0   0   1   1   2   2]  3   3   4   4   5   5   |
-        #                    1          3      0   0   1   1   2   2 [ 3   3   4   4   5   5]  ▼
+        #                                      |     Accum step 0      |     Accum step 1      |
+        #                                      |   GPU 0   |   GPU 1   |   GPU 0   |   GPU 1   |
         #
-        #                    2          4     [6   6   7   7   8   8]  9   9  10  10  11  11
-        #                    2          5      6   6   7   7   8   8 [ 9   9  10  10  11  11]
+        #                 global_step   step    <-───>  num_generations=2
+        #                                       <-───────> per_device_train_batch_size=3
+        #  grad_accum    ▲  ▲  0          0     [0   0   1   1   2   2]  3   3   4   4   5   5    <- Generate for the whole accumulated batch; store the completions; use the first slice to compute the loss
+        #     =2         ▼  |  0          1      0   0   1   1   2   2 [ 3   3   4   4   5   5]   <- Take the stored generations and use the second slice to compute the loss
+        #                   |
+        #                   |  1          2     [0   0   1   1   2   2]  3   3   4   4   5   5    <- Take the stored generations and use the first slice to compute the loss
+        #  num_iterations=2 ▼  1          3      0   0   1   1   2   2 [ 3   3   4   4   5   5]   <- Take the stored generations and use the second slice to compute the loss
+        #
+        #                      2          4     [6   6   7   7   8   8]  9   9  10  10  11  11    <- Generate for the whole accumulated batch; store the completions; use the first slice to compute the loss
+        #                      2          5      6   6   7   7   8   8 [ 9   9  10  10  11  11]   <- ...
         #                                          ...
         effective_batch_size = (
             self.args.per_device_train_batch_size
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -185,7 +185,6 @@ def nanstd(tensor: torch.Tensor) -> torch.Tensor:
     return torch.sqrt(variance)
 
 
-
 def split_tensor_dict(
     tensor_dict: dict[str, Optional[torch.Tensor]], num_chunks: int
 ) -> list[dict[str, Optional[torch.Tensor]]]:
@@ -212,6 +211,7 @@ def split_tensor_dict(
         }
         for i in range(num_chunks)
     ]
+
 
 def nanmin(tensor: torch.Tensor) -> torch.Tensor:
     """
@@ -241,7 +241,6 @@ def nanmax(tensor: torch.Tensor) -> torch.Tensor:
     if torch.isnan(tensor).all():
         return torch.tensor(float("nan"), dtype=tensor.dtype, device=tensor.device)
     return torch.max(tensor[~torch.isnan(tensor)])
-
 
 
 class GRPOTrainer(Trainer):
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -20,6 +20,7 @@ from collections.abc import Sized
 from contextlib import nullcontext
 from typing import Any, Callable, Optional, Union
 
+import datasets
 import torch
 import torch.utils.data
 import transformers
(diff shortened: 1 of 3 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -732,21 +796,35 @@ def _move_model_to_vllm(self):
...
-            if self.state.global_step % self.num_iterations == 0 or buffered_inputs is None:
-                # buffered_inputs=None can occur when resuming from a checkpoint
-                inputs = self._generate_and_score_completions(inputs)
-                self._buffered_inputs[buffer_index] = inputs
-            else:
-                inputs = buffered_inputs
+            generate_every = self.args.gradient_accumulation_steps * self.num_iterations
+            if self._step % generate_every == 0 or self._buffered_inputs is None:
+                # self._buffered_inputs=None can occur when resuming from a checkpoint
+                accumulated_local_batch = self._generate_and_score_completions(accumulated_local_batch)
```
I-l-l-I, line 819: I think using `accumulated_local_batch` directly might lead to OOM when calling `_get_per_token_logps`, maybe splitting `_generate_and_score_completions` into two parts would be better.

qgallouedec:
## Regression

The following experiments aim to ensure that that the results match when the same configuration is used. For all graphs, 
- red is for the current main (d625c55)
- green, this branch (a5f0f03)

*Outline:*

- `gradient_accumulation_steps=1` and `num_iterations=1`
- `gradient_accumulation_steps=1` and `num_iterations=2`
- `gradient_accumulation_steps=1` and `num_iterations=4`
- `gradient_accumulation_steps=4` and `num_iterations=1`
- `gradient_accumulation_steps=4` and `num_iterations=2`
- `gradient_accumulation_steps=4` and `num_iterations=4`
- `gradient_accumulation_steps=16` and `num_iterations=1`
- `gradient_accumulation_steps=16` and `num_iterations=2`
- `gradient_accumulation_steps=16` and `num_iterations=4`

### `gradient_accumulation_steps=1` and `num_iterations=1`



### `gradient_accumulation_steps=1` and `num_iterations=2`



### `gradient_accumulation_steps=1` and `num_iterations=4`



### `gradient_accumulation_steps=4` and `num_iterations=1`



### `gradient_accumulation_steps=4` and `num_iterations=2`



### `gradient_accumulation_steps=4` and `num_iterations=4`



### `gradient_accumulation_steps=16` and `num_iterations=1`



### `gradient_accumulation_steps=16` and `num_iterations=2`



### `gradient_accumulation_steps=16` and `num_iterations=4`

trl/trainer/grpo_trainer.py
qgallouedec: Good point. I haven't encountered such OOM yet. In fact, here, we do the forward pass with no grad, so maybe it prevents this problem? I'd have to test it with a large gradient accumulation step to confirm.

trl/trainer/grpo_trainer.py
```diff
@@ -183,6 +185,34 @@ def nanstd(tensor: torch.Tensor) -> torch.Tensor:
...
+    chunk_size = first_tensor.shape[0] // num_chunks
+    return [
+        {
+            key: tensor[i * chunk_size : (i + 1) * chunk_size] if tensor is not None else None
+            for key, tensor in tensor_dict.items()
+        }
+        for i in range(num_chunks)
+    ]
+
+
```
shirinyamani, line 215: so by this you are creating the mini-batches based on grad_accum_step size 
```python
self._buffered_inputs = split_tensor_dict(accumulated_local_batch, self.args.gradient_accumulation_steps)
```

inspecting;

```python

x = torch.arange(12).reshape(6, 2)
y = torch.arange(6).reshape(6, 1)
tensor_dict = {"x": x, "y": y}
    

result = split_tensor_dict(tensor_dict, 3)

print("Original tensors:")
print(f"x: {x}")
print(f"y: {y}")
print("\nSplit results:")
for i, chunk in enumerate(result):
    print(f"Chunk {i}:")
    print(f"x: {chunk['x']}")
    print(f"y: {chunk['y']}")

```
out tiny batches are:

```text
Original tensors:
x: tensor([[ 0,  1],
        [ 2,  3],
        [ 4,  5],
        [ 6,  7],
        [ 8,  9],
        [10, 11]])
y: tensor([[0],
        [1],
        [2],
        [3],
        [4],
        [5]])

Split results:
Chunk 0:
x: tensor([[0, 1],
        [2, 3]])
y: tensor([[0],
        [1]])
Chunk 1:
x: tensor([[4, 5],
        [6, 7]])
y: tensor([[2],
        [3]])
Chunk 2:
x: tensor([[ 8,  9],
        [10, 11]])
y: tensor([[4],
        [5]])
```

qgallouedec: Exact

trl/trainer/grpo_trainer.py
I-l-l-I: I tested it and it will be OOM when `max_completion_length` is large.

qgallouedec: Using the same setting, you don't get OOM with main?

I-l-l-I: Yes, I train 1.5B model on two gpus with GAS=8, BS=16 and max_completion_length=2048.

trl/trainer/grpo_trainer.py
```diff
@@ -631,7 +695,7 @@ def _get_train_sampler(self) -> Sampler:
             data_source=self.train_dataset,
             mini_repeat_count=self.num_generations,
             batch_size=effective_batch_size // self.num_generations,
-            repeat_count=self.num_iterations,
+            repeat_count=self.num_iterations * self.args.gradient_accumulation_steps,
```
edbeeching, line 698: Probably not for this PR, but would it be worth having a `per_device_mega_batch_size` and completely decoupling the gradient accumulation from the size of the big generation batch? Some batch shuffling would be required to ensure that the minibatches vary from one iteration to the next, but I don't think it would be a huge change.

trl/trainer/grpo_trainer.py
I-l-l-I: > I think using `accumulated_local_batch` directly might lead to OOM when calling `_get_per_token_logps`, maybe splitting `_generate_and_score_completions` into two parts would be better.

In the first part, use vLLM to generate all the `completion_ids` and calculate the `advantages` of all these completions. In the second part, compute `old_per_token_logps` and `ref_per_token_logps` with original `per_device_batchsize` (perhaps `old_per_token_logps` can be obtained directly in the vLLM generation phase).

One benefit of this is that instead of `(per_device_batchsize * num_processes) % num_generations == 0`, we need `(per_device_batchsize * num_processes * gradient_accumulation_steps) % num_generations == 0`. This solves (https://github.com/huggingface/trl/issues/3017) and (https://github.com/huggingface/trl/pull/3288) without introducing additional parameters in config, just adjust the `per_device_batchsize` and `gradient_accumulation_steps` to adapt to various scenarios.

trl/trainer/grpo_trainer.py
shirinyamani: @edbeeching Do you mean sth like this in simple terms?
```python
maga_batch_size = 524288  # ~0.5M, in number of tokens if T = 1024
B = 64 # micro batch size
T = 1024 
assert maga_batch_size % (B * T * ddp_world_size) == 0
grad_accum_steps = maga_batch_size // (B * T * ddp_world_size)
```
where we later do;
```python
for micro_step in range(grad_accum_steps):
        x, y = train_loader.next_batch()
        x, y = x.to(device), y.to(device)
        logits, loss = model(x, y)
        loss = loss / grad_accum_steps
        loss_accum += loss.detach()
        loss.backward()
```

lewtun: I think what Ed is suggesting is to align the `GRPOTrainer` partially with our other RL trainers where one defines a large effective batch size of `N_prompts x N_generations` and then performing K steps of mini-batch optimisation (i.e. the `num_minibatches` arg in the `PPOTrainer`). This is also what other RL frameworks like `verl` do.

In that context, gradient accumulation is not involved in defining the large effective batch size, but I realise it is not easy to force the `transformers.Trainer` logic to do mini-batch optimisation and should be done in a separate PR (if at all)

trl/trainer/grpo_trainer.py
lewtun: A good model to test would be any of the DeepSeek distilled ones: they easily produce 32k tokens per prompt, so you'll know quickly if improvements to the generation/scoring still OOM or not

qgallouedec: That's a very good point @I-l-l-I, I didn't even realised that with this PR, we're releasing this requirement `(per_device_batchsize * num_processes) % num_generations == 0`

qgallouedec: This should avoid OOM: eedaab5

qgallouedec: Now the memory usage is the same as with main:

W B Chart 15_04_2025, 17_24_59


```python
import random

dataset = load_dataset("trl-lib/tldr", split="train[:200]")


def reward_random(completions, **kwargs):
    return [random.random() for _ in range(len(completions))]

trainer = GRPOTrainer(
    model="Qwen/Qwen2.5-1.5B",
    reward_funcs=reward_random,
    train_dataset=dataset,
    args=GRPOConfig(
        per_device_train_batch_size=16,
        gradient_accumulation_steps=8,
        max_completion_length=2048,
        use_vllm=True,
        gradient_checkpointing=True,
        bf16=True,
    ),
)
trainer.train()
```


2 GPUs

```
CUDA_VISIBLE_DEVICES=0,1 accelerate launch --config_file examples/accelerate_configs/deepspeed_zero3.yaml --num_processes 2 3283.py 
```

Side note: with this script, for example, training time is reduced from 25 min to 10 min. 🤯

trl/trainer/grpo_trainer.py
```diff
@@ -836,13 +932,14 @@ def _generate_and_score_completions(
         attention_mask = torch.cat([prompt_mask, completion_mask], dim=1)  # (B, P+C)
 
         logits_to_keep = completion_ids.size(1)  # we only need to compute the logits for the completion tokens
+        batch_size = self.args.per_device_train_batch_size if mode == "train" else self.args.per_device_eval_batch_size
```
edbeeching, line 935: Generally for training (independent of this PR) `per_device_eval_batch_size` is `2*per_device_train_batch_size`

As we are in `no_grad` mode you can probably set the `per_device_train_batch_size` here to be `2*self.args.per_device_train_batch_size` as there will be no memory used for the storage of activations etc.

It would be worth validating in a scenario with significant memory pressure, so do not worry about it if you want to get the PR merged.

trl/trainer/grpo_trainer.py
```diff
@@ -677,18 +747,28 @@ def _get_last_hidden_state(self, model, input_ids, attention_mask, logits_to_kee
...
-        logits = logits[:, :-1, :]  # (B, L-1, V), exclude the last logit: it corresponds to the next token pred
-        input_ids = input_ids[:, -logits_to_keep:]
-        # For transformers<=4.48, logits_to_keep argument isn't supported, so here we drop logits ourselves.
-        # See https://github.com/huggingface/trl/issues/2770
-        logits = logits[:, -logits_to_keep:]
-        # Divide logits by sampling temperature.
-        # See https://huggingface.co/blog/the_n_implementation_details_of_rlhf_with_ppo#policy-training-implementation-details
-        logits = logits / self.temperature
-        return selective_log_softmax(logits, input_ids)  # compute logprobs for the input tokens
+    def _get_per_token_logps(self, model, input_ids, attention_mask, logits_to_keep, batch_size=None):
```
lewtun, line 750: nit:
```suggestion
    def _get_per_token_logps(self, model, input_ids, attention_mask, logits_to_keep, batch_size=None) -> torch.Tensor:
```

lewtun (approved):
Very nice PR with clearly documented logic; this was a pleasure to read! Looking forward to these speed-ups!

zhiqihuang:
From trl logging, I noticed sometime the generations is not correctly paired with prompts. It could a logging issue in multi-gpu training or it could be the order changed during one call per batch to the vllm server. I will try to find more trace and post here.

Issue #3284, opened 2025-04-12
OpenAI-Compatible vLLM Server with Weight Synchronization


BjarniHaukur:
### Feature request

I propose adding a new OpenAI-compatible vLLM API server for use with the `GRPOTrainer`.

The implementation mirrors the weight syncing logic from `trl/scripts/vllm_serve.py`, but offloads most complexity to the existing `vllm.entrypoints.openai.api_server` infrastructure.

This enables training on significantly more complex rollouts than the standard synchronous .generate() endpoint can support. By supporting the OpenAI API interface, it also allows seamless integration with a wide range of existing agent frameworks and products.

This direction is a step toward reproducing pipelines like [OpenHands LM 32B](https://www.all-hands.dev/blog/introducing-openhands-lm-32b----a-strong-open-coding-agent-model). I strongly suspect that Claude 3.7 Sonnet was trained in a similar fashion, iteratively reinforced using rollouts generated through its own [Claude Code](https://docs.anthropic.com/en/docs/welcome) scaffolding.

### Motivation

Currently, TRL only supports synchronous, batched `.generate()` calls for inference. This restricts the types of rollouts that can be created,  especially in domains that benefit from having multi-step approaches, tool use, or environment interaction.

I’ve been using TRL for my Master’s thesis on reinforcement learning for language models in the program repair domain. In several GRPO experiments, I repeatedly encountered the same limitation: with `.generate()`, all context construction, planning, and feedback extraction must happen within a single call. For example, in tasks from [SWE-Gym](https://huggingface.co/SWE-Gym), the model needs to generate code edits for real repositories. To do this in one `.generate()` call, the user must manually construct the relevant repo context and later parse outputs like diffs to extract useful reward signals. This makes experimentation slow and always feels like “reinventing the wheel.”

Rather than building ad-hoc scaffolding from scratch, I began exploring how to integrate existing coding agents like [Aider](https://github.com/Aider-AI/aider) directly into the training loop. These agents already support rich workflows such as repo mapping, diff parsing, and iterative interaction—and they use the OpenAI API interface. Enabling TRL to train models through this interface would allow us to run them *in situ*, inside the same environment they’re meant to be deployed in.

This proposal aims to bridge that gap and enable more realistic, multi-step training workflows as a first-class feature in TRL.

### Your contribution

I have already developed an initial working implementation in this PR: <INSERT PR>, which introduces `vllm_serve_openai_compatible.py`.

[...]

PR #3288, opened 2025-04-13
Modified GRPOTrainer to accumulate gradient within a single training batch

jarrelscy:
#  What does this PR do?

GRPOTrainer calculates advantages and then calculates loss per completion. Currently this is all done within a single batch which can take a lot of memory. Just like with gradient accumulation, we can call .backwards on the loss for each completion separately. This PR does so by introducing a new parameter into GRPOConfig called num_generations_chunks, of which num_generations needs to be a multiple of. Doing so will cause loss.backward to be called per num_generations_chunks number of completions. 

Example usage:
```
# train_grpo.py

training_args = GRPOConfig(output_dir="Qwen2-0.5B-GRPO", 
                          logging_steps=10,
                          per_device_train_batch_size=16, # needs to be a multiple of num_generations
                          num_generations=8, # needs to be a multiple of num_generations_chunks 
                          num_generations_chunks=8)
trainer = GRPOTrainer(
    reward_funcs=reward_len,
    args=training_args,
    train_dataset=dataset,
)
trainer.train()
```

Fixes # [3017](https://github.com/huggingface/trl/issues/3017)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


qgallouedec:
Thanks @jarrelscy 

I understand the motivation. Just for clarification, if 

> Just like with gradient accumulation, we can call .backwards on the loss for each completion separately. 

then why not using gradient accumulation? Is it because the generation will also be done on smaller batches, then makes things slower?

jarrelscy:
Hi @qgallouedec as @JamesBowerXanda pointed out in [here](https://github.com/huggingface/trl/issues/3017), the quality of the loss depends on the group size. In this [paper ](https://arxiv.org/abs/2502.18548) they point that you need a large group size to approximate the expected reward normalised by the standard deviation of the reward of an output sampled from the previous policy. 

In GRPO each generation is assigned a relative advantage against other generations, so if the group size is small, this can lead to erratic losses. 

In gradient accumulation (per batch), we are still comparing the advantage of each generation against other generations within that batch.

qgallouedec:
FYI, now you can pass a group as large as `gradient_accumulation * per_device_batch_size * num_devices` thanks to #3284

qgallouedec:
Closing this as I believe the motivation behind this PR has been addressed by #3283

jarrelscy:
@qgallouedec this PR is not exactly the same as 3283 - its akin to the comment [here](https://github.com/huggingface/trl/pull/3283#discussion_r2044304049) on 3283, which states that this functionality is not implemented in 3283.

qgallouedec:
Ok, sorry for the misjudgement, so I'm reopening the PR.

jiangix-paper:
Hello, any update for this? Have you merge it to the master branch?

jarrelscy:
@jiangix-paper there are some new changes in the trl main branch which I think are not compatible - the entropy masking implementation. i've just done a merge, feel free to try cloning and testing it

qgallouedec:
Now that generation and loss computation are decoupled, I think you can easily scale to any groupe size by playing with the gradient accumulation parameter. E.g.,if you want a group size of 32 you can for example:

```python
GRPOConfig(
    gradient_accumulation_steps=32,
    num_generations=32,
)
```


Hope I'm not misjudging again, please let me know if it's the case

Issue #3292, opened 2025-04-15
Value Model in PPOTrainer: Unsupported Optional type


AMindToThink:
### Reproduction

The `value_model` parameter of PPOTrainer is `value_model: Optional[nn.Module] = None`, but if it is set as None or left as default, errors occur. 

I'm unsure which of the following possibilities was intended:

1. Someone planned to later add support for skipping the use of a value model and using the rewards directly instead of the advantage. This might make sense in memory-constrained applications, at the expense of much worse performance (though really, just use RLOO).
2. This is a simple typo, and the type should be changed from `Optional[nn.Module]` to `nn.Module` and the parameter should be required.

I'm eager to make more and more substantial contributions to `trl`, so if these changes would be helpful, I'm happy to implement it!

The following is what happens if you run examples/scripts/ppo/tmp_ppo.py but set the `value_model` to `None`.

`(interpretable-fine-tuning) cs29824@sting-vm-1:~/matthew/trl-SAC$ python -i examples/scripts/ppo/tmp_ppo.py     --dataset_name trl-internal-testing/descriptiveness-sentiment-trl-style     --dataset_train_split descriptiveness     --learning_rate 3e-6     --output_dir models/minimal/ppo     --per_device_train_batch_size 64     --gradient_accumulation_steps 1     --total_episodes 10000     --model_name_or_path EleutherAI/pythia-1b-deduped     --missing_eos_penalty 1.0
[2025-04-15 04:10:55,588] [INFO] [real_accelerator.py:222:get_accelerator] Setting ds_accelerator to cuda (auto detect)
Some weights of GPTNeoXForSequenceClassification were not initialized from the model checkpoint at EleutherAI/pythia-160m and are newly initialized: ['score.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
Traceback (most recent call last):
  File "/home/cs29824/matthew/trl-SAC/examples/scripts/ppo/tmp_ppo.py", line 153, in <module>
    trainer = PPOTrainer(
              ^^^^^^^^^^^
  File "/home/cs29824/matthew/interpretable-fine-tuning/.venv/lib/python3.11/site-packages/trl/trainer/ppo_trainer.py", line 222, in __init__
    self.model = PolicyAndValueWrapper(self.policy_model, self.value_model)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #3293, opened 2025-04-15
Keep PPOTrainer Value Model (Critic Model)


AMindToThink:
### Feature request

Proposed change: Add an option in PPOConfig called `save_value_model:bool=False`. If true, ppo_trainer's save_model will set `self.model` to the value model and save it after saving the `self.model.policy` as normal.

### Motivation

Right now, the PPOTrainer trains a classification model which takes a state and outputs an estimate for how much reward it expects the policy to earn in that state.
After training it, the value model is discarded.

That seems really weird to me! It seems extremely useful to have a classifier model which can predict how well your text generation model will do given a prompt. If the value model predicts a bad response before the policy generates a response, you can take precautions. 

For instance: 

- If the value model predicts that `model1` would do poorly, you can switch to `model2`, which may do better.
- You can have a content warning for responses expected to bother the user. For instance, if the user asks "What's the best religion and why?", then you could have a pop-up window saying "You have asked a sensitive question. Please note that the views of the model do not reflect the views of `our_corp`, and that we do our best to satisfy all our customers and stakeholders." (The question has no good answer because if the AI refuses to answer, that bothers people, and if it gives an answer, many people will be upset, so the value model would give that prompt a low value.)
- The Value model is an interesting object to research, and allowing users to save it will facilitate research.

### Your contribution

I'm eager to make contributions to trl, so if these changes would be helpful, I'm happy to implement it!

qgallouedec:
Thanks for reporting, and sorry for the late reply, we're catching up. I'm going to close this one, see https://github.com/huggingface/trl/pull/6120#issuecomment-4823906635

PR #3300, opened 2025-04-15
🙅 PPO value_model can't be None, so it shouldn't be Optional

AMindToThink:
Simple change to the typing and default value of value_model in PPOTrainer. Since it cannot be None, it should not default to None or be marked Optional. See [this issue](https://github.com/huggingface/trl/issues/3292).

### DIFF
--- a/trl/trainer/ppo_trainer.py
+++ b/trl/trainer/ppo_trainer.py
@@ -107,7 +107,7 @@ class PPOTrainer(Trainer):
         ref_model: Optional[nn.Module],
         reward_model: nn.Module,
         train_dataset: Dataset,
-        value_model: Optional[nn.Module] = None,
+        value_model: nn.Module,
         data_collator: Optional[DataCollatorWithPadding] = None,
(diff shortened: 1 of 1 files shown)


qgallouedec (approved):
This looks reasonable, thanks! 🙏

qgallouedec:
@AMindToThink are your latest commits for this PR?

AMindToThink:
Oops, sorry, no. Reverted.

I'm working to allow users to save the value model along with the policy model.

I'm running into some odd errors. I'll make a bug report.

AMindToThink:
Hi @qgallouedec, 
It turns out that [this issue was already raised a year and a half ago](https://github.com/huggingface/trl/issues/3293), and you decided to keep the value model Optional with a None default to keep the documentation easier to read. Sorry for rehashing old questions!

I have made [an issue with the bugs](https://github.com/huggingface/trl/issues/3301) I ran into while trying to implement saving the value model. I would appreciate your help!

Thank you,
Matthew (@AMindToThink)

Issue #3301, opened 2025-04-15
no attribute 'policy' when pushing PPOTrainer to hub using example ppo.py script


AMindToThink:
### Reproduction

I'm working to add support for [keeping the value model (critic model)](https://github.com/huggingface/trl/issues/3293) that is created by PPOTrainer. I'm running into an odd issue where running the example script with the `push_to_hub` argument results in an error, but then the model is pushed anyways. 

This might not seem like a problem, but when I try changing save_model so that both models can be saved, I get the error and then neither model is pushed. 

I supsect this has something to do with code running in parallel interfering with each other.

I would much appreciate some help!

(Additionally, the progress bars don't indicate when the push finishes! This confused me greatly)

# Standard Command Run (straight from examples/scripts/ppo/ppo.py)
```
(trl) cs29824@sting-vm-1:~/matthew/trl$ python -i examples/scripts/ppo/ppo.py     --dataset_name trl-internal-testing/descriptiveness-sentiment-trl-style     --dataset_train_split descriptiveness     --learning_rate 3e-6     --output_dir models/minimal/ppo_push_main_12     --per_device_train_batch_size 64     --gradient_accumulation_steps 1     --total_episodes 1     --model_name_or_path EleutherAI/pythia-1b-deduped     --missing_eos_penalty 1.0 --push_to_hub
```
# Resulting error of standard command:
```
Traceback (most recent call last):
  File "/home/cs29824/matthew/trl/examples/scripts/ppo/ppo.py", line 166, in <module>
    trainer.save_model(training_args.output_dir)
  File "/home/cs29824/matthew/trl/trl/trainer/ppo_trainer.py", line 338, in save_model
    super().save_model(output_dir, _internal_call)
  File "/home/cs29824/matthew/trl/.venv/lib/python3.10/site-packages/transformers/trainer.py", line 3906, in save_model
    self.push_to_hub(commit_message="Model save")
  File "/home/cs29824/matthew/trl/.venv/lib/python3.10/site-packages/transformers/trainer.py", line 4816, in push_to_hub
    self.save_model(_internal_call=True)
  File "/home/cs29824/matthew/trl/trl/trainer/ppo_trainer.py", line 332, in save_model
    self.model = self.model.policy  # save only the policy
  File "/home/cs29824/matthew/trl/.venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1928, in __getattr__
    raise AttributeError(
[...]

PR #3308, opened 2025-04-16
Allow for saving the PPOTrainer value model (critic model)

AMindToThink:
This PR adds functionality for `save_value_model: bool=False` which determines whether to save the value model that is trained by PPOTrainer. 

During training, PPO puts quite a lot of effort training the value model to be good at predicting the reward the policy will receive. That sounds like a valuable tool, and it is a shame that previously we had simply been discarding it. This PR adds functionality for keeping the value model.

I requested this feature [here](http://github.com/issues/created?issue=huggingface%7Ctrl%7C3293), and this PR resolves that.
It also resolves the error described in this issue: [no attribute 'policy' when pushing PPOTrainer to hub using example ppo.py script #3301](https://github.com/huggingface/trl/issues/3301)
**I'm unsure whether the way I fixed it, the early return on line 338 in PPOTrainer, is legal. Please check!**

I would love your feedback, @qgallouedec!

### DIFF
--- a/examples/scripts/sft_video_llm.py
+++ b/examples/scripts/sft_video_llm.py
@@ -50,12 +50,12 @@ from typing import Any
 
 import requests
 import torch
-import wandb
 from datasets import load_dataset
 from peft import LoraConfig
 from qwen_vl_utils import process_vision_info
 from transformers import AutoModelForVision2Seq, AutoProcessor, BitsAndBytesConfig, Qwen2VLProcessor
 
+import wandb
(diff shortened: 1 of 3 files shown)


AMindToThink:
Hi, I'd like to bring attention back to this, especially since Sutskever mentioned Value Functions as a core area of future research in his recent Dwarkesh interview.

I can fix the merge conflicts, if doing so is what is needed for this to be approved. I'll also squash the commits while I'm at it.

qgallouedec:
Considering the very low usage of PPOTrainer, we decided to place the system in maintenance mode, consequently, I'll close this PR. Thanks for the contribution though.

PR #3324, opened 2025-04-18
🏰 [vllm] Support `base_url` parameter for vLLM client initialization

### LINKED ISSUE #3322: Feature Request: Support Custom URL Configuration for vLLM Client Initialization
### Feature request

Hello TRL Team!

First, I want to express my appreciation for your excellent work on this library - TRL has become an indispensable tool in our LLM development pipeline.

**Proposal**
I'd like to suggest an enhancement to improve flexibility when using vLLM with online methods. Currently, the vLLM client constructs its endpoint URL by combining host (IP address) and port parameters. To better support dynamic environments and proxy configurations, could we add direct support for URL specification during client initialization?

```python
# Current
client = vLLMClient(host="192.168.1.10", port=8000)

# Suggested
client = vLLMClient(base_url="https://llm-proxy.mycompany.com/v1")
```

**Benefits**
1.Infrastructure Agnosticism : Supports proxy setups, load balancers, and service meshes
2.K8s Friendliness : Aligns with cloud-native practices using stable service endpoints
3.Backwards Compatibility : Maintain existing IP+port support while adding URL option
4.Security : Enables easier integration with TLS-terminating reverse proxies

**Use Case**
Our RL data pipeline requires:
1.Developers to spin up ephemeral vLLM instances
2.Automatic registration with our proxy service
3.Client configuration using fixed proxy URL rather than instance-specific IPs
Would this be a valuable addition to the library? I'm happy to provide more details or contribute to implementation discussions.

Thank you for reading this request!

### Motivation

**Motivation**
In our Kubernetes-based infrastructure:
vLLM servers receive dynamic IP assignments on each deployment
We use a stable proxy service that routes requests to the current vLLM endpoint
The current IP+port binding requires manual updates whenever servers restart
A URL-based configuration would provide permanent endpoint addressing

### Your contribution

TBD

re-imagined:
related issue https://github.com/huggingface/trl/issues/3322
- Introduce base_url parameter to simplify server connection
- Update __init__ method to support both base_url and host+port configurations
- Modify URL construction in various methods to use base_url
- Update documentation to include new initialization examples

Close #3322.


tests/test_vllm_client_server.py
```diff
@@ -162,3 +238,64 @@ def tearDownClass(cls):
...
+        cls.client.close_communicator()
+
+        # vLLM x pytest (or Popen) seems not to handle process termination well. To avoid zombie processes, we need to
+        # kill the server process and its children explicitly.
+        parent = psutil.Process(cls.server_process.pid)
+        children = parent.children(recursive=True)
+        for child in children:
+            child.send_signal(signal.SIGTERM)
+        cls.server_process.terminate()
+        cls.server_process.wait()
```
qgallouedec, line 301:
```suggestion
```

I think one test is enough

trl/extras/vllm_client.py
```diff
@@ -47,10 +45,12 @@ class VLLMClient:
     weights in a distributed setting. Before using it, start the vLLM server with `trl vllm-serve`.
 
     Args:
+        base_url (`str`, *optional*, defaults to `None`):
+            Base URL for the vLLM server (e.g., "http://localhost:8000"). If provided, host and server_port are ignored.
```
qgallouedec, line 49:
```suggestion
            Base URL for the vLLM server (e.g., `"http://localhost:8000"`). If provided, host and server_port are ignored.
```

trl/extras/vllm_client.py
```diff
@@ -47,10 +45,12 @@ class VLLMClient:
...
         host (`str`, *optional*, defaults to `"0.0.0.0"`):
-            IP address of the vLLM server.
+            IP address of the vLLM server. Ignored if base_url is provided.
```
qgallouedec, line 51:
```suggestion
            IP address of the vLLM server. Ignored if `base_url` is provided.
```

trl/extras/vllm_client.py
```diff
@@ -47,10 +45,12 @@ class VLLMClient:
...
         server_port (`int`, *optional*, defaults to `8000`):
-            Port number of the vLLM server.
+            Port number of the vLLM server. Ignored if base_url is provided.
```
qgallouedec, line 53:
```suggestion
            Port number of the vLLM server. Ignored if `base_url` is provided.
```

trl/extras/vllm_client.py
```diff
@@ -83,16 +101,30 @@ class VLLMClient:
...
             raise ImportError("vLLM is not installed. Please install it with `pip install vllm`.")
 
         self.session = requests.Session()
-        self.host = host
-        self.server_port = server_port
+
+        if base_url is not None:
+            # Parse the base_url to extract host and port
+            parsed_url = urlparse(base_url)
+            scheme = parsed_url.scheme or "http"
```
qgallouedec, line 121: this allows to add something like `"localhost:8000"` instead of `"http://localhost:8000"` right?

trl/trainer/grpo_trainer.py
```diff
@@ -627,9 +627,16 @@ def data_collator(features):  # No data collation is needed in GRPO
...
+                if args.vllm_server_base_url is not None:
+                    self.vllm_client = VLLMClient(
+                        base_url=args.vllm_server_base_url, connection_timeout=args.vllm_server_timeout
+                    )
+                else:
+                    self.vllm_client = VLLMClient(
+                        host=args.vllm_server_host, 
+                        server_port=args.vllm_server_port, 
+                        connection_timeout=args.vllm_server_timeout
+                    )
```
qgallouedec, line 639: maybe this is easier?

```suggestion
                base_url = args.vllm_server_base_url if args.vllm_server_base_url is not None else f"http://{args.vllm_server_port}:{args.vllm_server_port}"
                self.vllm_client = VLLMClient(base_url=ase_url, connection_timeout=args.vllm_server_timeout)
```

qgallouedec (commented):
I think this change makes sense, thanks for suggesting it.
Sorry for the delayed review.
I've made a few comments.
Have you tested on your infrastructure?

trl/extras/vllm_client.py
re-imagined: yes

trl/extras/vllm_client.py
```diff
@@ -207,14 +239,14 @@ def init_communicator(self):
...
-        url = f"http://{self.host}:{self.server_port}/init_communicator/"
+        url = f"{self.base_url}/init_communicator/"
         # In the server side, the host is set to 0.0.0.0
         response = self.session.post(url, json={"host": "0.0.0.0", "port": self.group_port, "world_size": world_size})
         if response.status_code != 200:
             raise Exception(f"Request failed: {response.status_code}, {response.text}")
 
         # Set up the communication group for weight broadcasting
-        pg = StatelessProcessGroup.create(host=self.host, port=self.group_port, rank=self.rank, world_size=world_size)
+        pg = StatelessProcessGroup.create(host="0.0.0.0", port=self.group_port, rank=self.rank, world_size=world_size)
```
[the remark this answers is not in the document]
re-imagined: yeah, when host is not provided at initialization of the client, using `self.host` to create StatelessProcessGroup could cause error

re-imagined:
> I think this change makes sense, thanks for suggesting it. Sorry for the delayed review. I've made a few comments. Have you tested on your infrastructure?

Thanks for your review. 😄
I am trying to test it on different cases, and checking if the model output is correct, but I am struggling on some nccl error when executing the tests. (it worked well on last commit 🤣, until I merge the main branch)
I will let you know as soon as i have an update

qgallouedec:
Hey! any update?

re-imagined:
> Hey! any update?

hi @qgallouedec,
Sorry for late update, I've been out for few days.

After several tests, I realized that  using 0.0.0.0 as the host address in StatelessProcessGroup.create isn't viable—it requires the server's actual IP address to function properly. But if we use base url to init the vllm client, we don't know the server's ip. 

Proposed Approach:  Use the base url to resolve the server's IP programmatically in the client  (when calling `/health`).

I’m currently testing this approach and diving deeper into how StatelessProcessGroup interacts with PyTorch’s TCPStore

I’ll update this thread once I’ve narrowed down the root cause and validated the fix. Let me know if you have insights or suggestions!

re-imagined:
@qgallouedec 
hi, it's ready for review now

trl/extras/vllm_client.py
```diff
@@ -119,11 +152,17 @@ def check_server(self, total_timeout: float = 0.0, retry_interval: float = 2.0):
                 elapsed_time = time.time() - start_time
                 if elapsed_time >= total_timeout:
                     raise ConnectionError(
-                        f"The vLLM server can't be reached at {self.host}:{self.server_port} after {total_timeout} "
+                        f"The vLLM server can't be reached at {self.base_url} after {total_timeout} "
                         "seconds. Make sure the server is running by running `trl vllm-serve`."
                     ) from exc
             else:
                 if response.status_code == 200:
+                    if "X-Forwarded-For" in response.headers:
```
re-imagined, line 160: we still need the ip address when using base url

trl/extras/vllm_client.py
```diff
@@ -119,11 +152,19 @@ def check_server(self, total_timeout: float = 0.0, retry_interval: float = 2.0):
...
```
re-imagined, line 160: we still need the ip addrees when using base url
1. An HTTP proxy can add a header containing the backend IP so that the client can read it from the response. The most common header for passing client IPs is X-Forwarded-For
2. get the ip via a request

qgallouedec: I've updated the host to be `self.host = socket.gethostbyname(parsed_url.hostname)` during the init. It's simpler. It works locally, but I'm curious to know if it works for you.

qgallouedec (commented):
Sorry for the late review. I've updated your branch with a few minor things, plus an attempt to simplify the host recovering. Can you check that it works on your side?

re-imagined:
> Sorry for the late review. I've updated your branch with a few minor things, plus an attempt to simplify the host recovering. Can you check that it works on your side?

thank you, I will test it today

re-imagined:
> Sorry for the late review. I've updated your branch with a few minor things, plus an attempt to simplify the host recovering. Can you check that it works on your side?

@qgallouedec 
I’ve tested the latest changes on my end, and everything works as expected. The simplification to the host recovery logic is a nice improvement—thanks for taking the time to refine it.

qgallouedec (approved):
LGTM, thanks!

PR #3337, opened 2025-04-22
🎭 Fix train and eval mode checking in `GRPOTrainer` and `SFTTrainer`

I-l-l-I:
# What does this PR do?

Use `model.training` instead of `control.should_evaluate` to determine if the model is in evaluation mode.

### Motivation

It is not accurate to use `should_evaluate` to determine the `mode`. For example, if the trainer needs to evaluate and log at a certain training step, then `log` will be executed twice at this step. The first log should be for the training metrics, but at this time `should_evaluate=True`, since `should_evaluate` means to evaluate after one step. 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -882,7 +882,7 @@ class GRPOTrainer(Trainer):
         #   - Completions are generated for each batch without buffering or reuse
         # Returns a single local batch in both cases.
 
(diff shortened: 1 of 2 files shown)


qgallouedec (approved):
That's a very good point @I-l-l-I!

I've tested myself I can confirm that the current implementation has indeed the issue you describe, and that your solution works like a charm.
I've also fixed it in SFT.

PR #3340, opened 2025-04-22
fix bug: When chat_template is not defined exception  will occur

dignfei:
… an exception will occur because there is no default value.

# What does this PR do?
fix bug: When chat_template is not defined in the configuration file, an exception will occur because there is no default value.

## Before submitting
- [√ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
      to it if that's the case.
- [√ ] Did you make sure to update the documentation with your changes?
- [√ ] Did you write any new necessary tests?

## Who can review?

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -295,7 +295,10 @@ class GRPOConfig(TrainingArguments):
         default=None,
         metadata={"help": "Regex for vLLM guided decoding. If `None` (default), guided decoding is disabled."},
     )
-
+    chat_template: Optional[str] = field(
+        default=None,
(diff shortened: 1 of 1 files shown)


LeonEricsson:
I think you are mistaken, [GRPOConfig](https://github.com/huggingface/trl/blob/1faa7f9b368296e459e51346e25bf88356e822c0/trl/trainer/grpo_config.py#L24) does not expect a chat_template parameter. If the tokenizer lacks a chat_template e.g `tokenizer.chat_template = None` and you pass a conversational dataset then things should fail, but this is expected.

Could you provide the code which produces the error?

qgallouedec:
Closing, explanation from @LeonEricsson sounds valid to me!

PR #3358, opened 2025-04-24
📝 vLLM integration doc

shirinyamani:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/docs/source/vllm_integration.md
+++ b/docs/source/vllm_integration.md
@@ -1,14 +1,72 @@
 # vLLM Integration
 
-<Tip warning={true}>
+# 🎬 Flashback: why do we need to use vLLM in online methods?
 
-Section under construction. Feel free to contribute!
+Online methods such as GRPO or Online DPO require the model to generate completions. Because these completions are used to compute the reward signal, they need to be generated at regular intervals during training. This is typically done every `gradient_accumulation_steps * num_iterations` steps, where `num_iterations` is the number of iterations between two gradient updates. Now, the problem is that generating completions is a time-consuming process, especially when using large models. The reason of the time-consuming nature of the generation process is by default, this is done using the [(unwrapped)model's `generate` method](https://github.com/huggingface/trl/blob/f3e8c2304428ef16e9ae5de9e5741ed84d533b7b/trl/trainer/grpo_trainer.py#L965C39-L965C66), (ofcourse if `use_vllm` is set to False). This `unwrapped_model.generate` method is thechnically a synchronous function, meaning that it will block the execution of the program until the generation is complete. This can lead to inefficiencies, especially when generating large batches of completions or when using large models. Therefore, the generation process can become a bottleneck in the training process, leading to longer training times and reduced efficiency. So this is why we need to think of a better way to do this which is using vLLM for faster generation. 
 
-</Tip>
+# 🤔 How does vLLM solve the slow generation issue?
+if you've ever done autoregressive decoder training, you know  all the input tokens to the LLM produce their attention key and value tensors, and these tensors are kept in GPU memory to later generate next tokens (Q) based on them. These cached key and value tensors are often referred to as KV cache.  However, this storing is really a pain as it occupies a lot of memory. 
+So here is the secret sauce of vLLM, it uses a technique called PagedAttention to solve this problem. PagedAttention , which is inspired by the OS’s virtual memory concept stores continuous keys and values in **non-contiguous memory space** which is way more efficient. The detail of this is beyond the scope of this document, but in short, it allows the model to store the keys and values in a more efficient way, reducing the memory footprint and speeding up the generation process. If you are interested, make sure to check out the [vLLM PagedAttention](https://blog.vllm.ai/2023/06/20/vllm.html) for more details.
 
-## TRL vLLM server
 
-TRL provides a way to speedup generation using a dedicated vLLM server. 
+# ⚙️ How to use vLLM in practice for generation in online methods in TRL?
+
+1. To use [vLLM](https://github.com/vllm-project/vllm), first install it using:
+
+```bash
+pip install vllm
+```
+
+or 
+
+```bash
+pip install "trl[vllm]"
+```
+
+<hfoptions id="vllm examples">
+<hfoption id="Online DPO">
+
+</hfoption>
+<hfoption id="GRPO">
+
+2. Then, **start a vLLM server** by running:
+
+```bash
+trl vllm-serve --model <model_name>
+```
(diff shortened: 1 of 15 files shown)


docs/source/vllm_integration.md
```diff
@@ -1,14 +1,72 @@
-# vLLM Integration
+# vLLM Integration
```
qgallouedec, line 1:
```suggestion
# vLLM Integration
```

docs/source/vllm_integration.md
```diff
@@ -1,14 +1,72 @@
...
... the reward signal, they need to be generated at regular intervals during training. This is typically done every `gradient_accumulation_steps * num_iterations` steps, where `num_iterations` is the number of iterations between two gradient updates. Now, the problem is that generating completions is a time-consuming process, especially when using large models. The reason of the time-consuming nature of the generation process is by default, this is done using the [(unwrapped)model's `generate` method](https://github.com/huggingface/trl/blob/f3e8c2304428ef16e9ae5de9e5741ed84d533b7b/trl/trainer/grpo_trainer.py#L965C39-L965C66), (ofcourse if `use_vllm` is set to False). This `unwrapped_model.generate` method is thechnically a synchronous function, meaning that it will block the execution of the program until the generation is complete. This can lead to inefficiencies, especially when generating large batches of completions or when using large models. Therefore, the generation process can become a bottleneck in the training process, leading to longer training times and reduced efficiency. So this is why we need to think of a better way to do this which is using vLLM for faster generation.
```
qgallouedec, line 5: What we name `num_iterations` in GRPO isn't the number of iterations between two gradient updates. You should use another term to avoid confusion

qgallouedec: > meaning that it will block the execution of the program until the generation is complet

I'm not sure why you mention this? Put like this, it seems to imply that this is the major limitation that will eliminate with vLLM.

docs/source/vllm_integration.md
qgallouedec, line 1: As far as the general structure is concerned, I think that when users come across this page, the first question they have  in mind is _How can I use vLLM with TRL to make things go faster?_ (a fortiori with GRPO). So my advice is to get straight to this point. And further down, explain in more detail why generation is the major bottleneck, how vLLM optimizes generation etc.

In short, I see this page more as a practical guide.

docs/source/vllm_integration.md
```diff
@@ -40,8 +98,74 @@ options:
                         feature. (default: None)
 ```
 
-### Find the best distributed setup
+# 🥸 Okay, now that we have the server running, how can we use it to generate completions?
```
qgallouedec, line 101: There seems to be heading level issues

docs/source/vllm_integration.md
```diff
@@ -1,14 +1,72 @@
...
+<hfoption id="GRPO">
+
+2. Then, **start a vLLM server** by running:
+
+```bash
+trl vllm-serve --model <model_name>
+```
+# 🔎 What exactly happens when you run `trl vllm-serve --model <model_name>`?
+when you run for example `trl vllm-serve --model Qwen/Qwen2.5-7B --tensor-parallel-size 2 --data-parallel-size 2`, the following happens:
+![vllm](https://huggingface.co/datasets/trl-lib/documentation-images/resolve/main/vllm_integration_dp2_tp2.png)
```
qgallouedec, line 39: I have the impression that on the figure you're using DP=4, and not DP=2.

docs/source/vllm_integration.md
```diff
@@ -1,14 +1,72 @@
...
+1. First it will spawn multiple workers to handle loads of requests in parallel. To figure our exactly how many workers to spawn, it will use the `--tensor-parallel-size` and `--data-parallel-size` arguments in the command. For example, if you run `trl vllm-serve --model Qwen/Qwen2.5-7B --tensor-parallel-size 2 --data-parallel-size 2`, it will spawn 4 workers (2 for data parallelism and 2 for tensor parallelism). Tricky point here is that you need to think of it as 4 workers independent and in parallel at the same time are processing a chunk of the incoming requests. Here the requests are basically the prompts that are sent to the model on the server to generate completions. Therefore, each of these workers(gpus) will be responsible for processing a chunk of the incoming requests.
```
qgallouedec, line 40: > For example, if you run `trl vllm-serve --model Qwen/Qwen2.5-7B --tensor-parallel-size 2 --data-parallel-size 2`, it will spawn 4 workers (2 for data parallelism and 2 for tensor parallelism)

When you put it like that, I think it's a bit misleading. It seems to imply this can also be true, which is not

> For example, if you run `trl vllm-serve --model Qwen/Qwen2.5-7B --tensor-parallel-size 3 --data-parallel-size 2`, it will spawn 5 workers (2 for data parallelism and 3 for tensor parallelism)

docs/source/vllm_integration.md
```diff
@@ -1,14 +1,72 @@
...
+
+2. Now that we have the requests(prompts) ready on each of the workers, the model will start generating the completions. Note that the model (models' weights actually) itself is split across multiple gpus on the vllm side (`--tensor parallelism size`) and each gpu will be responsible for processing a chunk of the incoming requests(`--data parallelism size`).
+
+3. Although the gpus process the requests in parallel and independent of one another, they need to communicate or talk with each other. Because recall that each of them process a chunk of the incoming prompt (e.g. if you have 4 gpus and 8 requests, with dp=2, each gpu will process 2 requests). This  gpu-gpu communication is handled by Nvidia NCCL lib. This communication is just to make sure each gpu has its slice of prompt/request.  Note that you can define the num_generations, as the number of completions to generate for each request. So if you have 4 gpus and 8 requests/prompts, with dp=2 and num_generations=2, each gpu will process 2 prompts and generate 2 completions for each of them. So in total, you will have 16 completions.
```
qgallouedec, line 44: > So if you have 4 gpus and 8 requests/prompts, with dp=2 and num_generations=2, each gpu will process 2 prompts and generate 2 completions for each of them. So in total, you will have 16 completions. 

when it's formulated like this "So in total, you will have 16 completions", it seems to imply that this number is derived from the ones you gave before (number of GPUs, DP). However, this number only comes from the fact that you have 8 prompts and 2 completions per prompt.

docs/source/vllm_integration.md
shirinyamani: correct, added the *

docs/source/vllm_integration.md
shirinyamani: cuz i couldn't figure the best way to represent TP, might get complicated, so not changed it to tp=1, dp=4

PR #3367, opened 2025-04-26
Environments

August-murr:
# Environments for Customized Rollouts with VLLM and GRPO

Environments allows for **customized rollouts and usage** of the model during training with **VLLM** and **GRPO**.

## Example: Creating a Custom Environment

```python
from trl import Environment

class MyCustomEnv(Environment):
    def __init__(self, ...):
        # Initialize anything your agent needs, like code execution,
        # tool calls, or external APIs.
        pass

    def tool_call(self, ...):
        # Define any environment-specific methods you want to use.
        pass

    def generate(self, vllm_client, generation_config: VLLMClientGenerationConfig, prompts: list[str]) -> list:
        # This is the method GRPO training uses to generate responses.
        # It must return a list of completion IDs (tokenized responses).
        return completion_ids
```

---

## `CodeAgentEnvironment`

A built-in `CodeAgentEnvironment` is included, which can be used with either:
- [E2B code interpreter](https://e2b.dev/)
- A local code interpreter

Example setup:

```python
from trl import CodeAgentEnvironment, E2BExecutor

code_executor = E2BExecutor(api_key="YOUR_E2B_TOKEN")
my_env = CodeAgentEnvironment(
    code_executor=code_executor,
    tokenizer=tokenizer,
    parsing_string="<code>",
    stop_string="</code>",
)
```

---

## Running the Environment

You can **use the environment manually** with a VLLM server to test and observe:

Start a VLLM server:
```bash
trl vllm-serve --model "your_model"
```

Then run the agent:

```python
from trl import VLLMClientGenerationConfig, VLLMClient

client = VLLMClient()
gen_config = VLLMClientGenerationConfig(
    n=8,
    repetition_penalty=1.0,
    temperature=0.8,
    top_p=0.9,
    top_k=10,
    min_p=0.0,
    max_tokens=256,
)

responses = my_env.run_agent(  # Main method in CodeAgentEnvironment to generate responses
    vllm_client=client,
    generation_config=gen_config,
    prompts=prompts
)
```

Or, **use the environment for training**:

```python

training_args = GRPOConfig(
    output_dir="code_agent",
    use_vllm=True,
    # ... other config options ...
)

trainer = GRPOTrainer(
    model=...,                # Your model or model name
    reward_funcs=...,         # Your reward function(s)
    args=training_args,
    environment=my_env,       # Your custom or built-in environment
    # ... other trainer args ...
)
```

---

## Related Work

While finishing this PR, ByteDance published [*ReTool: Reinforcement Learning for Strategic Tool Use in LLMs*](https://arxiv.org/pdf/2504.11536).  
This PR with `CodeAgentEnvironment` is a similar idea, with two key differences:
1. **ReTool uses PPO**, while we use **GRPO**.
2. **ReTool masks interpreter feedback** — an idea I'll be investigating and possibly adding later.

---

## Other Notes

- I added a `stop_string` parameter to the VLLM CLI — it’s critical for agentic workflows where you want to cleanly delimit model outputs.

---

## Demo Run

To test, I set up a simple math task with a one-shot prompt for `Qwen-0.5B-instruct`.
The reward function checked if the agent **used the code interpreter** during its response.

- Here's the [wandb run](https://wandb.ai/moh-murr/Code_Agent/runs/buoluckd?nw=nwusermohmurr).
- You can specifically check:
  - `train/rewards/count_function_calls_per_response/mean`
  - [Code Interpreter Call Frequency](https://wandb.ai/moh-murr/Code_Agent/runs/buoluckd/panel/ckfy8s8wy?nw=nwusermohmurr) (a custom evaluation trace with a larger `n` for lower noise)

W B Code Interpreter Call Frequency

**Results**:  
The base model had a 14% code interpreter call frequency.  
After **24 training steps**, the frequency increased to **almost 100%**.

### DIFF
--- a/docs/source/training_agents.md
+++ b/docs/source/training_agents.md
@@ -1,0 +1,138 @@
+# Training Agents with Environments in TRL
+
+Environments in TRL provide a powerful way to customize the rollout process for reinforcement learning (RL) with language models, specifically when using vLLM as the backend. This enables advanced agent training workflows, such as allowing your model to interact with external tools or APIs during RL training.
+
+> **Note:**  
+> - Environments currently require vLLM as the backend.
+> - The only supported RL training method at this time is GRPO (Group Relative Policy Optimization).
+
+## Installation
+
+To use Environments and agent features, install TRL with the `[agents]` extra:
+
+```bash
+pip install trl[agents]
+```
+
+## What Are Environments?
+
+An **Environment** defines how rollouts (model generations) are performed during RL training. By customizing the rollout, you can enable your model to interact with external systems, execute code, or follow any custom logic during generation.
+
+TRL provides built-in environments, such as `CodeAgentEnvironment`, and you can also implement your own by subclassing `Environment`.
+
+---
+
+## Using the Built-in CodeAgentEnvironment
+
+The `CodeAgentEnvironment` allows your agent to write and execute code during training, using a code interpreter like the E2B Code Interpreter. This is useful for tasks where the model needs to solve problems by running code.
+
+You can also use the `run_agent` method to test and observe your agent's behavior outside of training.
+
+### Example: Running the CodeAgent Environment
+
+First, start a vLLM server with your model:
+
+```bash
+trl vllm-serve --model "Qwen/Qwen2.5-0.5B-Instruct"
+```
+
+Then, use the environment and client in Python:
+
+```python
+from trl import CodeAgentEnvironment, E2BExecutor, VLLMClientGenerationConfig, VLLMClient
+
+client = VLLMClient()
+gen_config = VLLMClientGenerationConfig(
+    n=8,
+    repetition_penalty=1.0,
+    temperature=0.8,
+    top_p=0.9,
+    top_k=10,
+    min_p=0.0,
+    max_tokens=256,
+)
+tokenizer = ...  # Load your tokenizer here
+code_executor = E2BExecutor(api_key="YOUR_E2B_TOKEN")
+my_env = CodeAgentEnvironment(
+    code_executor=code_executor,
+    tokenizer=tokenizer,
+    parsing_string="<code>",
+    stop_string="</code>",
+)
+prompts = [...]  # List of prompt strings
+responses = my_env.run_agent(
+    vllm_client=client,
+    generation_config=gen_config,
+    prompts=prompts
+)
+```
+
+---
+
+## Writing a Custom Environment
+
+To create your own environment, subclass `Environment` and implement a `generate` method. This method receives the vLLM client, a generation config, and a list of prompts, and must return a list of tokenized completions (as lists of token IDs).
+
+The `generate` method is called by the GRPO training loop to perform rollouts.
+
+**Example:**
+
+```python
+from trl import Environment, VLLMClientGenerationConfig
+
+class MyCustomEnv(Environment):
+    def __init__(self, ...):
+        # Your custom initialization
+        pass
+
+    def generate(self, vllm_client, generation_config: VLLMClientGenerationConfig, prompts: list[str]) -> list:
+        # Your custom rollout logic
+        completion_ids = vllm_client.generate(
+            prompts=prompts,
+            **vars(generation_config)
+        )
+        return completion_ids
+```
+
+You can add any additional methods or logic needed for your task.
+
+---
+
+## Training with a Custom Environment
+
+To use your environment for RL training, simply pass it to the `GRPOTrainer`:
+
+```python
+from trl import GRPOConfig, GRPOTrainer
+
+training_args = GRPOConfig(
+    output_dir="qwen_0.5B-agent",
+    use_vllm=True,
+    # ... other config options ...
+)
+
+trainer = GRPOTrainer(
+    model=...,                # Your model or model name
+    reward_funcs=...,         # Your reward function(s)
+    args=training_args,
+    environment=my_env,       # Your custom or built-in environment
+    # ... other trainer args ...
+)
+```
+
+Then, start training as usual:
+
+```python
+trainer.train()
+```
+
+---
+
+## Summary
+
+- **Environments** let you fully customize how rollouts are performed during RL training with TRL and vLLM.
+- Use built-in environments like `CodeAgentEnvironment` for code-executing agents, or implement your own by subclassing `Environment`.
+- The `generate` method is the key entry point for custom rollout logic.
+- Pass your environment to `GRPOTrainer` to use it during RL training.
+
+For more advanced examples, see the [notebooks](../notebooks/) or the TRL documentation.
--- a/trl/agents/environments.py
+++ b/trl/agents/environments.py
@@ -108,12 +108,13 @@ class CodeAgentEnvironment(Environment):
         """
         # Store original prompts for later use
         original_prompts = prompts.copy()
-        
+        # a buncha debug print statements
+        print(f"Original prompts: {original_prompts}")
         # Duplicate prompts to match the requested number of generations (n)
         expanded_prompts = []
         for prompt in prompts:
             expanded_prompts.extend([prompt] * generation_config.n)
-        
+        print(f"Expanded prompts: {expanded_prompts}")
         # Create a modified generation config with n=1 for individual generations
         single_gen_config = VLLMClientGenerationConfig(
             n=1,
@@ -124,20 +125,24 @@ class CodeAgentEnvironment(Environment):
             min_p=generation_config.min_p,
             max_tokens=generation_config.max_tokens,
             guided_decoding_regex=generation_config.guided_decoding_regex,
+            # Ensure stop string is always included for code parsing
             stop=[self.stop_string] if generation_config.stop is None else 
-                 generation_config.stop + [self.stop_string]
+                    list(set(generation_config.stop + [self.stop_string])) # Use set to avoid duplicates
         )
-        
+        print(f"Single generation config: {single_gen_config}")
         # Track current conversations for each prompt
         current_conversations = expanded_prompts.copy()
         completed_conversations = []
-        
+        print(f"Initial Current conversations: {current_conversations}")
+        print(f"Initial Completed conversations: {completed_conversations}")
         # Continue code execution loop until all conversations are completed
         while current_conversations:
+            print(f"\n--- Starting Generation Loop ---")
+            print(f"Current conversations ({len(current_conversations)}): {current_conversations}")
             # Generate next response segment
             outputs = vllm_client.generate(
                 prompts=current_conversations,
-                n=1,
+                n=1, # Already handled expansion, generate 1 per conversation
                 repetition_penalty=single_gen_config.repetition_penalty,
                 temperature=single_gen_config.temperature,
                 top_p=single_gen_config.top_p,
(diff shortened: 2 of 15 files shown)


August-murr:
@qgallouedec @shirinyamani Can I get a review?

August-murr:
@qgallouedec, since you disagreed with the implementation in #2810, I assumed you preferred a wrapper or a class instead. Therefore, I began writing this PR with the Environment class.

> We're not sure at this point if everything belongs in TRL or if it should be implemented in a third-party library
Agentic RL is moving fast. why shouldn't TRL support it?

shirinyamani:
> Agentic RL is moving fast. why shouldn't TRL support it?

Hi Mohammad, Thanks for your contribution. Very promising results and great work! We can perhaps work together to build this support? 
Just a couple of questions/notes;

1. I see in your environments.py file where you define the CodeAgentEnvironment as the source where the agent can pick the code interpreter tool for the given set of prompts, building on this, I am wondering, if we want to introduce other tool kits like internet search or less costly approach of getting completion/generation from a bigger/reasoning model, how the workflow would be? 
2. Following the first question, I believe ideally we would like to have a load_tools (similar to [smolagent](https://huggingface.co/docs/smolagents/en/reference/tools#smolagents.load_tool) one) or a `[tools]` so that the agent can pick from!
3. May I learn abt the code n the trainer code you used to generate the result please? for reproduction purposes

August-murr:
It's August.

> We can perhaps work together to build this support?

It looks like the TRL team is a bit unsure about what features the `environments` should include, like whether they can share on the hub or how much customization they can have.

> if we want to introduce other tool kits like internet search

I really don't think we should be adding tools to TRL, since we also have to maintain them, especially given how fast it's moving with stuff like MCPs.

> we would like to have a load_tools (similar to smolagent one) or a [tools] so that the agent can pick from!

The `LocalCodeExecuter` already has access to all the tools defined or imported into the training script, and the `prepare_for_local_agent` function takes a list of tools like `[tool_1, tool_2, ...]` to add their docstrings or source code into the system prompt, helping the agent figure out how to use them.

As for the `E2BCodeExecuter`, in the last iteration (#2936), I included a parameter that lets it use a separate tools script path and add that to the code being executed, along with the `prepare_for_e2b_agent` to include it in the system prompt. I didn’t put it in this PR, but I can if you think it’s worth it.

I'm not really sure about the issues or limitations of these features. The main thing is they only work for code agents, not tool calling agents. Honestly, tool calling agents that use JSON never really made sense to me, while code agents feel way more flexible and efficient.

If you’ve got any other ideas, feel free to share!

> May I learn abt the dataset n the trainer code you used to generate the result

I was working on this [Kaggle Notebook](https://www.kaggle.com/code/augustmurr/agentic-rl-demo), but Kaggle was super inconsistent, and the `trl vllm-serve` kept crashing randomly. So, I didn’t get to finish it and just opened a PR. In the cell `Dataset and Prompt Design`, you can check out the prompt and data I used. I know it’s not the most practical or useful way to do it, especially with budget constraints, but it’s really just a proof of concept. If anyone feels like it’s not enough and wants me to run it with something more practical, just let me know—I’d be happy to do it!

August-murr:
I really need you all to be more open and communicative about the features or functionality you’re looking for in the environments.
@lewtun 
You mentioned sharing environments like the Smolagents tools, but I wasn’t aware that sharing tools on the hub was even an option because there's no way to find them! Maybe @aymeric-roucher could explain why. It feels pointless to share environments if no one can find or use them, right?

Regarding shareable environments, if the goal is just for training with RL, a simple setup that can load a class from the hub and be used in the `GRPOTrainer` should do the trick. It’s pretty straightforward, but I’m not sure if environments should also mask the tokens that weren’t generated by the LLM for loss calculation.

However, if the idea is for users to load an environment and serve it, that complicates things a bit. I’m not too familiar with how TRL users would serve their fine-tuned models in production. Right now, they can use the environment outside of training with `trl vllm-serve`, but to make it production-ready, they’d need to work with inference frameworks like VLLM, SGLang, etc.

docs/source/training_agents.md
```diff
@@ -0,0 +1,138 @@
+# Training Agents with Environments in TRL
+
+Environments in TRL provide a powerful way to customize the rollout process for reinforcement learning (RL) with language models, specifically when using vLLM as the backend. This enables advanced agent training workflows, such as allowing your model to interact with external tools or APIs during RL training.
```
lewtun, line 3: nit since we emphasise the backend in the note below
```suggestion
Environments in TRL provide a powerful way to customize the rollout process for reinforcement learning (RL) with language models. This enables advanced agent training workflows, such as allowing your model to interact with external tools or APIs during RL training.
```

docs/source/training_agents.md
```diff
@@ -0,0 +1,138 @@
...
+
+An **Environment** defines how rollouts (model generations) are performed during RL training. By customizing the rollout, you can enable your model to interact with external systems, execute code, or follow any custom logic during generation.
+
+TRL provides built-in environments, such as `CodeAgentEnvironment`, and you can also implement your own by subclassing `Environment`.
+
+---
+
+## Using the Built-in CodeAgentEnvironment
+
+The `CodeAgentEnvironment` allows your agent to write and execute code during training, using a code interpreter like the E2B Code Interpreter. This is useful for tasks where the model needs to solve problems by running code.
```
lewtun, line 27:
```suggestion
The `CodeAgentEnvironment` allows your agent to write and execute code during training, using a code interpreter like [E2B sandboxes](https://e2b.dev). This is useful for tasks where the model needs to solve problems by running code.
```

docs/source/training_agents.md
```diff
@@ -0,0 +1,138 @@
...
+
+You can also use the `run_agent` method to test and observe your agent's behavior outside of training.
```
lewtun, line 29: turbo nit:
```suggestion
You can also use the `run_agent()` method to test and observe your agent's behavior outside of training.
```

docs/source/training_agents.md
```diff
@@ -0,0 +1,138 @@
...
+    generation_config=gen_config,
+    prompts=prompts
+)
+```
+
+---
+
+## Writing a Custom Environment
+
+To create your own environment, subclass `Environment` and implement a `generate` method. This method receives the vLLM client, a generation config, and a list of prompts, and must return a list of tokenized completions (as lists of token IDs).
```
lewtun, line 74:
```suggestion
To create your own environment, subclass `Environment` and implement a `generate()` method. This method receives the vLLM client, a generation config, and a list of prompts, and must return a list of tokenized completions (as lists of token IDs).
```

docs/source/training_agents.md
```diff
@@ -0,0 +1,138 @@
...
+
+The `generate` method is called by the GRPO training loop to perform rollouts.
```
lewtun, line 76:
```suggestion
The `generate()` method is called by the GRPO training loop to perform rollouts.
```

trl/agents/environments.py
```diff
@@ -0,0 +1,318 @@
...
+
+import abc
+from dataclasses import dataclass
+from typing import Any, Optional
+
+from transformers import PreTrainedTokenizerBase
+
+
+@dataclass
+class VLLMClientGenerationConfig:
```
lewtun, line 23: I think this could live in `vllm_client.py`

trl/agents/utils.py
```diff
@@ -0,0 +1,270 @@
...
+            Path to the user-provided script.
+
+    Returns:
+        `str`:
+            The content of the script.
+    """
+    return Path(user_script_path).read_text()
+
+
+class Localexecuter:
```
lewtun, line 64: turbo nit:
```suggestion
class LocalExecutor:
```

lewtun: Also, would it make sense to reuse the executors from `smolagents` ([example](https://github.com/huggingface/smolagents/blob/d8db112be8f208db851c0836d56de0c9b157cd6d/src/smolagents/local_python_executor.py#L1493)) or does that introduce a lot of unnecessary dependencies?

The alternative would be to copy-paste with attribution so we can potentially remove the langchain dep

trl/agents/utils.py
```diff
@@ -0,0 +1,270 @@
...
+        for code in codes:
+            try:
+                result = repl.run(code)
+                results.append(str(result))
+            except Exception as e:
+                results.append(f"Error executing code: {str(e)}")
+        return results
+
+
+class E2BExecuter:
```
lewtun, line 86: turbo nit:
```suggestion
class E2BExecutor:
```

Similar question here on whether we can reuse the `smolagents` one: https://github.com/huggingface/smolagents/blob/d8db112be8f208db851c0836d56de0c9b157cd6d/src/smolagents/remote_executors.py#L96

trl/extras/vllm_client.py
```diff
@@ -142,6 +142,7 @@ def generate(
         min_p: float = 0.0,
         max_tokens: int = 16,
         guided_decoding_regex: Optional[str] = None,
+        stop: Optional[list[str]] = None,
```
lewtun, line 145: Do I understand correctly that this is needed to enable e.g. stopping generation at the end of a code block instead of EOS?

trl/trainer/grpo_trainer.py
```diff
@@ -934,22 +948,30 @@ def _generate_and_score_completions(
 
             # Generate completions using vLLM: gather all prompts and use them in a single call in the main process
             all_prompts_text = gather_object(prompts_text)
+            # In the _generate_and_score_completions method
             if self.accelerator.is_main_process:
-                # Since 'prompts' contains 'num_generations' duplicates, we first take unique prompts, and generate
-                # num_generations outputs for each one. This is faster than generating outputs for each duplicate
-                # prompt individually.
+                # Since 'prompts' contains 'num_generations' duplicates, we first take unique prompts...
```
lewtun, line 953: Please restore the full comment :)

trl/trainer/grpo_trainer.py
```diff
@@ -934,22 +948,30 @@ def _generate_and_score_completions(
 
             # Generate completions using vLLM: gather all prompts and use them in a single call in the main process
             all_prompts_text = gather_object(prompts_text)
+            # In the _generate_and_score_completions method
```
lewtun, line 951:
```suggestion
```

trl/agents/environments.py
```diff
@@ -0,0 +1,318 @@
...
+        if generation_config is None:
+            raise ValueError("Generation config must be provided to the generate method")
+
+        return vllm_client.generate(
+            prompts=prompts,
+            **vars(generation_config),
+        )
+
+
+class CodeAgentEnvironment(Environment):
```
lewtun, line 78: Very nice implementation of multi-step code execution!

lewtun (commented):
This is a super nice feature @August-murr and great job with the clean implementation. Overall the design looks really good to me, with a question about whether we can reuse any parts from `smolagents` which has similar executors for code. I realise this adds another dependency to the project, so am also happy if we include a handful of maintained environments and let the community build their own.

To validate the feature, I was thinking we could try teaching a model to solve GSM8k problems with Python. Maybe @shirinyamani can give this a try?

lewtun:
> You mentioned sharing environments like the Smolagents tools, but I wasn’t aware that sharing tools on the hub was even an option because there's no way to find them!

Ah yes, we haven't really settled on the best way to make these discoverable, but the basic idea I had was to expose environments as Gradio apps, so people could find them on Spaces. We don't need to solve that in this PR IMO.

trl/agents/environments.py
August-murr: > Very nice implementation of multi-step code execution!

It was All Gemini😂 and some Claude
The best thing Humans can do is to stay out of the loop🫨

trl/agents/utils.py
August-murr: Not really, the biggest hurdle in integrating smolagents into TRL, as I explained before, is running things in batches and in parallel.
This E2BCodeExecuter uses asyncio to run batch of code but smolagents doesn't.

trl/agents/utils.py
August-murr: It wouldn’t make sense if it’s just for the local code executor, but if we think there’s other stuff in smolagents that could be helpful for TRL, then it could be worth having that dependency.

> The alternative would be to copy-paste with attribution so we can potentially remove the langchain dep

you mean copy-paste this:
https://github.com/langchain-ai/langchain-experimental/blob/main/libs/experimental/langchain_experimental/utilities/python.py
with attribution?

August-murr:
> To validate the feature, I was thinking we could try teaching a model to solve GSM8k problems with Python.

I got it! I’ll need to check out some of the new changes and ideas, and maybe train a bit on using tools like web search to make sure everything is running properly.

trl/agents/utils.py
lewtun: OK then let's stick with your implementation for now and later if `smolagents` supports batching / multiprocessing, we can revisit

August-murr:
@qgallouedec @lewtun @shirinyamani 

[Here](https://github.com/August-murr/AutonomousAIResearcher/tree/main/reports/ReAct%20on%20Roids-GSM8K%20Agent)'s the report and the script for the GSM8K code agent trained with the branch. It is very similar to Microsoft's paper titled ["Artist"](https://arxiv.org/pdf/2505.01441). 
I recommend reviewing the table from the [W&B](https://wandb.ai/moh-murr/ReAct_on_Roids/runs/26o633zo?nw=nwusermohmurr) run to see the results for yourself.

A few notes:
- I implemented tool output masking, which may or may not have been necessary. However, it was included in nearly every research paper addressing similar issues, such as [ReTool](https://arxiv.org/pdf/2504.11536). Keep in mind that this will complicate customized environments since each one needs to output a completion mask that masks the tool outputs.
- To ensure proper functionality of tool output masking, I had to add tokens to the tokenizer and resize the model embedding before training. If we weren't using VLLM, this would have been as simple as adding a few lines to the GRPOTrainer script. However, since VLLM pulls the model directly from a local source or the Hugging Face hub, adding the tokens must be done separately. I will write the documentation for this as we begin the review process.

Whats next?

August-murr:
@lewtun @qgallouedec @shirinyamani 

Whatever

Issue #3250, opened 2025-04-07
PPOTrainer' object has no attribute 'generate'


zoeChen119:
### Reproduction

```python
import torch
from datasets import load_dataset
from transformers import AutoTokenizer, GenerationConfig, AutoModelForCausalLM
from trl import PPOTrainer, PPOConfig, AutoModelForCausalLMWithValueHead, create_reference_model
# from trl.core import respond_to_batch

ds = load_dataset('imdb', split='train')
ds = ds.rename_columns({'text': 'review', 'label': 'sentiment'})
ds = ds.filter(lambda x: len(x["review"])>200, batched=False)

model = AutoModelForCausalLMWithValueHead.from_pretrained('gpt2')
value_model = AutoModelForCausalLM.from_pretrained('gpt2')
model_ref = create_reference_model(model)
reward_model = create_reference_model(model)

tokenizer = AutoTokenizer.from_pretrained('gpt2')
tokenizer.pad_token = tokenizer.eos_token

ppo_config = PPOConfig(batch_size=1, mini_batch_size=1,output_dir="./output/")

query_txt = "This morning I went to the "
query_tensor = tokenizer.encode(query_txt, return_tensors="pt")

# response_tensor  = respond_to_batch(model, query_tensor)
model.generation_config = GenerationConfig(eos_token_id=tokenizer.eos_token_id)

ppo_trainer = PPOTrainer(
    args=ppo_config, 
    processing_class=tokenizer,
    model=model, 
    ref_model=model_ref, 
    reward_model=reward_model, 
    train_dataset=ds,
    value_model=value_model,
)

response_tensor, values_and_logits = ppo_trainer.generate(
    query_tensor,
    return_prompt=False,
    return_values_and_logits=True,
)
# define a reward for response
# (this could be any reward such as human feedback or output from another model)
reward = [torch.tensor(1.0)]

# train model for one step with ppo
train_stats = ppo_trainer.step([query_tensor[0]], [response_tensor[0]], reward)
```

outputs:

```
Traceback (most recent call last):
  File "/workspace/multimodal/RL/ppo_test.py", line 4, in <module>
    from trl import PPOTrainer, PPOConfig, AutoModelForCausalLMWithValueHead, create_reference_model, AutoModelForCausalLM
ImportError: cannot import name 'AutoModelForCausalLM' from 'trl' (/usr/local/lib/python3.10/dist-packages/trl/__init__.py)
```

### System Info

transformers
Version: 4.48.0.dev0

[...]

Issue #3270, opened 2025-04-09
Question on the removal of `.step()` and `.generate()` methods in `Trainer`


jskaf34:
Hello everyone,

First of all, thank you for the amazing work you're doing—this library is truly a fantastic resource for the community! 😊

I had a more practical question regarding the decision to remove the `.step()` and `.generate()` methods from the `Trainer` classes. Our current implementation relies heavily on these methods, and updating to the new version would require a fair amount of engineering effort.

Before proceeding, I wanted to understand the rationale behind this change. Was there a specific motivation for deprecating these methods, and is there a recommended alternative approach?

Thank you in advance for your help and insights! 🚀

qgallouedec:
Hi, thanks for the feedback.
That sounds more like a question about transformers, doesn't it? 
`Trainer` still exposes a [`training_step`](https://github.com/huggingface/transformers/blob/f834ca2c19215f1e4fb0959cc3faafeaf56cd4f7/src/transformers/trainer.py#L3698) method, and I looked at the [codebase a year ago](https://github.com/huggingface/transformers/blob/9322576e2f49d1014fb0c00a7a7c8c34b6a5fd35/src/transformers/trainer.py), `Trainer` didn't exposed either a `step` method, or a `generate` method. What are you referring to exactly? Or maybe in an even older version?

eryawww:
Hello, thank you for the answer.

I believe I have a similar question. In the earlier version (trl==0.11), there was a [PPOTrainer.step(query, response, score)](https://github.com/huggingface/trl/blob/v0.11-release/trl/trainer/ppo_trainer.py#L655) method that was really handy for online/iterative RL scenario. From what I see in the current implementation, it looks like everything is now wrapped into [PPOTrainer.train](https://github.com/huggingface/trl/blob/main/trl/trainer/ppo_trainer.py).

I’m wondering, what is the recommended way to implement an iterative scenario with the new version?

jskaf34:
Hey thank you for your answer.
In `trl==0.11 version`, `PPOTrainer` had a `generate` and a `step`  method that were really convenient to customise our RL loops. In fact, many examples on the internet relies on this old version of TRL. Those methods were removed afterwards, I was wondering if there were any reason for that removal please, in case I had to implement them back to upgrade to TRL's new version.
Have a great day !

albertvillanova:
Hi @jskaf34 — picking this up as we tidy the backlog. 

[...]

PR #3374, opened 2025-04-27
Reintroduce `generate` method for PPOTrainer

CloseChoice:
# What does this PR do?

In the release v0.12.0 the `generate` and the `_generate_batched` methods were removed from the PPOTrainer. As there is feedback from the community (see the issues #3250 and #3270) (including [a member's comment](https://github.com/huggingface/trl/issues/3270#issuecomment-2794225816)) this PR reintroduces the method and tests it properly. Note that the reintroduced methods are basically copy-pasted from version v0.11.4 and then slightly adapted to the current code.

Supports #3250 (not strictly closing since in the reproducible example the `.step` method is used as well, which this PR does not reintroduce).

## Before submitting
- ~[ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).~
      Pull Request section?
      to it if that's the case.

## Who can review?


CloseChoice:
Just realized that there is the `generate_completions` method and the `batch_generation` function already. I would assume that reusing this functionality should be the target of this PR, not adding duplicating functionality as it is done right now. Though right now this is not possible since the `generate_completions` method does not return anything. Would it be suitable to add a return to the `generate_completions`?

qgallouedec:
What do you think is the best?

CloseChoice:
> What do you think is the best?

I'll prepare an update to the `generate_completions` function and update this PR, will ping you once that is ready for a first review.

CloseChoice:
@qgallouedec I updated the functionality and reused the `generate` method that we have already. I think this is way cleaner and simpler. My current implementation of `generate` does not match the one in `0.11.4` exactly, since I wanted to keep it simple. Let me know what you think.

CloseChoice:
@qgallouedec quick follow up, do you see this of use to the repo or shall we close the PR?

albertvillanova:
Sorry for the long silence here @CloseChoice — we're catching up on the PR backlog and you deserved an answer back in June when you asked. 

Since then PPO moved to `trl.experimental.ppo` (#4482) and the maintained `trl/trainer/ppo_trainer.py` was removed outright (#5051), so both files this PR touches no longer exist on main — and the `generate`/`batch_generation` helpers it imports from `trl.trainer.utils` now live inside the experimental module itself. 

We also settled the underlying question: the v0.11-style streaming API (`generate(...)` plus a manual `step(...)` loop) isn't coming back, which is why we closed #3250 as not-planned and #3270. 

So I'm closing this one — not a reflection on the code, the ground just moved out from under it. Thanks for the effort here, and sorry we let it sit this long.

PR #3388, opened 2025-04-29
💔 [GRPO] Decouple gradient accumulation from the number of minibatches generated

edbeeching:
This PR decouples the size of the effective batch from the gradient accumulation steps with a new argument `num_mini_batches`. 

By default `num_mini_batches = gradient_accumulation_steps`

A few caveats:

- The mini-batches are grouped in prompt order, a subsequent PR will shuffle the `accumulated_local_batch` before adding it to the buffer. PR https://github.com/huggingface/trl/pull/3391
- The minibatches are still padded to the length of the longest generation, a subsequent PR will fix this
- If `num_iterations > 1` the buffer indices should be permuted / reshuffled in order to ensure that samples are taken randomly.


lewtun (approved):
LGTM with some nits!

lewtun:
As discussed offline, it would be interesting to check that the `clip_ratio` is indeed non-zero when `num_mini_batches != gas` (could possibly be part of the integration tests?)

trl/trainer/grpo_config.py
```diff
@@ -412,3 +419,20 @@ class GRPOConfig(TrainingArguments):
...
+    def __post_init__(self):
+        super().__post_init__()
+
+        # The current default effective batch size
+        if self.num_mini_batches is None:
+            self.num_mini_batches = self.gradient_accumulation_steps
+
+        if self.num_mini_batches < self.gradient_accumulation_steps:
+            raise ValueError(
+                f"steps_per_generation ({self.num_mini_batches}) must be greater than or equal to gradient_accumulation_steps ({self.gradient_accumulation_steps})."
```
qgallouedec, line 432:
```suggestion
                f"num_mini_batches ({self.num_mini_batches}) must be greater than or equal to gradient_accumulation_steps ({self.gradient_accumulation_steps})."
```

trl/trainer/grpo_config.py
```diff
@@ -412,3 +419,20 @@ class GRPOConfig(TrainingArguments):
...
+            )
+
+        if self.num_mini_batches % self.gradient_accumulation_steps != 0:
+            raise ValueError(
+                f"steps_per_generation ({self.num_mini_batches}) must be divisible by gradient_accumulation_steps ({self.gradient_accumulation_steps})."
```
qgallouedec, line 437: Is this really necessary? One could imagine that `self.gradient_accumulation_steps=2` and `self.num_mini_batches=3`, in which case generation would occur in the middle of a gradient accumulation cycle. This would be a bit odd because the off-policyness could vary within an effective batch, but I don't see why this would be a problem.

trl/trainer/grpo_config.py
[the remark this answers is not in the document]
qgallouedec: Same remark here. It's probably suboptimal, but is there any reason that prevents `self.num_mini_batches < gradient_accumulation_steps`?

trl/trainer/grpo_config.py
```diff
@@ -61,6 +61,8 @@ class GRPOConfig(TrainingArguments):
             with vLLM generation.
         shuffle_dataset (`bool`, *optional*, defaults to `True`):
             Whether to shuffle the training dataset.
+        num_mini_batches: (`int`, *optional*, defaults to `None`):
```
qgallouedec, line 64: I would find it clearer to call this parameter `generate_every` or something similar. I can't figure out what a mini-batch is in this case. Named like that, and with the doc ("split"), I expect there to be a division by that number later, like `mini_batch_size = batch_size // num_mini_batches`.

trl/trainer/grpo_trainer.py
```diff
@@ -583,14 +583,14 @@ def data_collator(features):  # No data collation is needed in GRPO
                 f"{self.num_generations}, which is less than the minimum required."
             )
         num_processes = self.accelerator.num_processes
-        effective_batch_size = args.per_device_train_batch_size * num_processes * args.gradient_accumulation_steps
+        effective_batch_size = args.per_device_train_batch_size * num_processes * args.num_mini_batches
```
qgallouedec, line 586: To be precise, I would rename this parameter, which actually corresponds to the generation batch size (`generation_batch_size`?), not the effective batch size.

trl/trainer/grpo_config.py
lewtun: For context, the mini-batch terminology used here is the same as done in e.g. DAPO:

Screenshot 2025-04-29 at 20 29 58

In other words, @edbeeching is partitioning the effective / generation batch into `num_mini_batches` subsets to compute the loss. It takes the model slightly off-policy but allows one to generate a large batch and optimise smaller chunks without going OOM

qgallouedec: I see. Although this is technically correct and corresponds to the paper, I have the impression that it's a bit misleading, insofar as, as a user, I imagine that by increasing `num_mini_batches`, I'm consequently decreasing the size of the mini-batch and increase the number of gradient updates per rollout step.
And, the sentence in the paper is also constructed like this: "the mini-batch size is set to 512, i.e., 16 gradient updates for each rollout step."

But that's not really what happens here. But if you don't find it misleading, we can leave it like that, just document it better.

lewtun: To clarify the DAPO example above, they have 512 unique prompts per batch and then sample 16 completions per prompt to obtain a generation batch size of 512 x 16 = 8'192.

They then partition the generation batch size into mini-batches of size 512, hence the 16 gradient updates per rollout step. 

Now suppose I have 8 GPUs and `per_device_train_batch_size=16`, then I believe the corresponding setting in `trl` would be:

* `gradient_accumulation_steps = 64` (to get 8,192 = 8 x 16 x 64)
* `num_mini_batches = 16` (i.e. 16 gradient updates with slices of 512 samples)

If @edbeeching agrees with my logic then indeed it would make sense to document how (a) the number of unique prompts is computed and (b) how mini-batching and gas can be tuned to obtain large generation batch sizes

trl/trainer/grpo_config.py
edbeeching: Yes you are right, we can relax this.

trl/trainer/grpo_config.py
edbeeching: @lewtun in general I agree, but I am not sure if  `gradient_accumulation_steps = 64` in the case of the DAPO example details above.

To summarize how I see it, for each generation step:
- There will be 16 optimization steps (gradient updates) in total.
- Each step is with 512 prompt-completions pairs
- The total generation batch size is `512*16=8192`
- `per_device_train_batch_size` is unknown, but lets assume it is 16 and `num_gpus=8`. 
- Which would mean `per_device_train_batch_size=16` and `gradient_accumulation_steps=4` as `16*8*4=512`

In this case the `num_mini_batches` would be:

- `generation_batch_size / (num_gpus*per_device_train_batch_size) = 64`
- This is also equal to the `num_optmization_steps* grad_acc_steps`

trl/trainer/grpo_trainer.py
edbeeching: Yes, I agree this is more clear.

trl/trainer/grpo_config.py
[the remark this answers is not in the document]
edbeeching: You are right on this point as well, I am not sure why a user would want to do this, but I agree we can relax this. 
Should I put a warning or remove it entirely?

trl/trainer/grpo_config.py
lewtun: Ah yes indeed I made a mistake in my above reasoning - I agree with @edbeeching. 

We discussed offline and also agree it is confusing to have `num_mini_batches` affect the generation batch size.

trl/trainer/grpo_config.py
```diff
@@ -464,3 +474,52 @@ class GRPOConfig(TrainingArguments):
...
+                for n_gen in range(2, self.generation_batch_size + 1)
+                if (self.generation_batch_size) % n_gen == 0
+            ]
+            if self.num_generations not in possible_values:
+                raise ValueError(
+                    f"The effective eval batch size ({num_processes} x {self.per_device_eval_batch_size}) must be "
+                    f"evenly divisible by the number of generations per prompt ({self.num_generations}). Given the "
+                    "current effective eval batch size, the valid values for the number of generations are: "
+                    f"{possible_values}."
+                )
```
qgallouedec, line 525: I've just spotted a small error! The sampling for evaluation is in fact the default one (no batch grouping), therefore

Global batch
```suggestion
            global_eval_batch_size = args.per_device_eval_batch_size * num_processes
            possible_values = [
                n_gen
                for n_gen in range(2, global_eval_batch_size + 1)
                if (global_eval_batch_size) % n_gen == 0
            ]
            if self.num_generations not in possible_values:
                raise ValueError(
                    f"The global eval batch size ({num_processes} x {self.per_device_eval_batch_size}) must be "
                    f"evenly divisible by the number of generations per prompt ({self.num_generations}). Given the "
                    "current effective eval batch size, the valid values for the number of generations are: "
                    f"{possible_values}."
                )
```

shirinyamani (approved):

edbeeching:
> I personally prefer steps_per_generation as this is more explicit to the user than generation_batch_size

I think both are useful, so I will make it so the user can set either option, but not both.

PR #3396, opened 2025-05-01
Suggestions for Decouple gradient accumulation from the number of minibatches generated

qgallouedec:
Some suggestions for PR #3388:

- Move `generation_batch_size` under the section `Parameters that control generation`
- use `int | None` instead of `int | str` because the the CLI parser doesn't support multi-type (that's why the CI is failing)
- rephrase some comments to be more precise (are we one the same page on these terms?):

| Term | Size |
|----------|----------|
| Local batch | `per_device_train_batch_size` |
| Local accumulated batch | `per_device_train_batch_size * gradient_accumulation_steps` |
| Global batch | `per_device_train_batch_size * world_size` |
| Effective batch | `per_device_train_batch_size * world_size * gradient_accumulation_steps` |
| Local generation batch | `per_device_train_batch_size * steps_per_generation ` |
| Generation batch | `per_device_train_batch_size * world_size  * steps_per_generation ` |

- since `num_generations` will make less sense to use with this PR, I removed it from the figure, and hopefully gained clarity
- support the case `self.args._steps_per_generation < self.args.gradient_accumulation_steps` (not sure why you'd want that, but it just requires to change `!=` into `<`.

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -65,7 +65,7 @@ class GRPOConfig(TrainingArguments):
         > Parameters that control generation
 
         generation_batch_size: (`int` or `None`, *optional*, defaults to `None`):
-            Batch size to use during generation. If `None`, it defaults to the effective training batch size:
+            Batch size to use for generation. If `None`, it defaults to the effective training batch size:
             `per_device_train_batch_size * num_processes * gradient_accumulation_steps`.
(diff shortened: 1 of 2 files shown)


edbeeching (approved):
Thanks for the suggestions, I will merge them

edbeeching (approved):

PR #3402, opened 2025-05-02
🪪 Remove license classifier

qgallouedec:
License classifier are deprecated, see https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license


lewtun (approved):
LGTM

PR #3404, opened 2025-05-02
💬 Fix `setup_chat_format` and add `clone_chat_template`

qgallouedec:
This PR does two things:

1. It fixes the resizing of the embedding dimension in `setup_chat_format`.
2. It introduces a new function, `clone_chat_template`.

Both functions essentially serve the same purpose. The key difference is that `setup_chat_template` takes as input any tokenizer that already includes a chat template. I think this new function is more convenient and flexible in practice.

Also, if you look at the implementations, you'll notice that unlike `setup_chat_format`, `clone_chat_template` doesn't handle BOS or PAD tokens. Correct me if I’m wrong, but I don’t believe that’s necessary when setting up a chat template.

Also, it adds all the added tokens from the source tokenizer, ensuring tokens like `<think>` `<|eot|>` ... are properly set.
Note that the practice vary a lot regarding these added token, whether they are "special" or not. The current implementation fully relies on the source tokenizer: if a token is special in the source tokenizer, it will be special in the target, and vice et versa.

In the long term, I think we should deprecate `setup_chat_format` in favor of `clone_chat_template `, but this can be done in a future PR.


trl/models/utils.py
```diff
@@ -116,7 +120,8 @@ def setup_chat_format(
 
     # resize embedding layer to a multiple of 64, https://x.com/karpathy/status/1621578354024677377
     model.resize_token_embeddings(
-        len(tokenizer), pad_to_multiple_of=resize_to_multiple_of if resize_to_multiple_of is not None else None
+        new_num_tokens=tokenizer.vocab_size + len(tokenizer.added_tokens_encoder.keys()),
+        pad_to_multiple_of=resize_to_multiple_of if resize_to_multiple_of is not None else None,
```
qgallouedec, line 124: fix the new embedding size

trl/models/utils.py
```diff
@@ -132,6 +137,69 @@ def setup_chat_format(
     return model, tokenizer
 
 
+def setup_chat_template(
+    model: PreTrainedModel,
+    tokenizer: PreTrainedTokenizer,
+    source: str,
```
edbeeching:
```suggestion
    source_tokenizer_path: str,
```

trl/models/utils.py
```diff
@@ -132,6 +137,69 @@ def setup_chat_format(
...
+    - Update the tokenizer's EOS token and add it to the special tokens.
+    - Resize the model's token embeddings to accommodate the new special tokens.
+    - Update the model's configuration and generation configuration to use the new EOS token.
+
+    Args:
+        model (`PreTrainedModel`):
+            Model to update.
+        tokenizer (`PreTrainedTokenizer`):
+            Tokenizer to update with a new chat template.
+        source (`str`):
```
edbeeching:
```suggestion
        source_tokenizer_path (`str`):
```

trl/models/utils.py
```diff
@@ -132,6 +137,69 @@ def setup_chat_format(
     return model, tokenizer
 
 
+def setup_chat_template(
```
edbeeching: Would `clone_chat_template` be a more suitable name?

trl/models/utils.py
```diff
@@ -132,6 +137,69 @@ def setup_chat_format(
...
+    ```
+    """
+    # Load the source tokenizer
+    tokenizer_source = AutoTokenizer.from_pretrained(source)
+
+    # Set the chat template for the tokenizer
+    tokenizer.chat_template = tokenizer_source.get_chat_template()
+
+    # Set the special tokens for the tokenizer
+    tokenizer.add_special_tokens({"additional_special_tokens": [tokenizer_source.eos_token]})
```
edbeeching: What if the source tokenizer's chat template contains other special tokens such as `<think>`, <|im_start|> etc , would we not need to add these to our tokenizer? Or should this be done by the user?

trl/models/utils.py
qgallouedec: sounds good

trl/models/utils.py
qgallouedec: Done in f1fbe73d647d04161ba93c12c66ea6057947a0bc

docs/source/sft_trainer.md
```diff
@@ -63,26 +63,26 @@ If you’d like to compute loss on both the prompt **and** the completion while
 ### Add Special Tokens for Chat Format
 
 Adding special tokens to a language model is crucial for training chat models. These tokens are added between the different roles in a conversation, such as the user, assistant, and system, and help the model recognize the structure and flow of a conversation. This setup is essential for enabling the model to generate coherent and contextually appropriate responses in a chat environment. 
-The [`setup_chat_format`] function in `trl` easily sets up a model and tokenizer for conversational AI tasks. This function:
+The [`clone_chat_template`] function in `trl` easily sets up a model and tokenizer for conversational AI tasks. This function:
```
lewtun: nit: let's use this as an occasion to tidy up the prose

```suggestion
The [`clone_chat_template`] function is a useful utility to prepare a model and tokenizer for conversational AI tasks. This function:
```

lewtun:
> Also, if you look at the implementations, you'll notice that unlike setup_chat_format, clone_chat_template doesn't handle BOS or PAD tokens. Correct me if I’m wrong, but I don’t believe that’s necessary when setting up a chat template.

The reason we did this in `setup_chat_format()` was because some base models like Llama / Qwen don't always define a BOS or PAD token and this needs to be set explicitly if you want to run SFT.

qgallouedec:
For popular chat templates, there’s usually at least one model on the Hugging Face Hub that already implements that template, right?  
The goal is to simplify SFT setup by letting users specify:  
*“Use the same chat template as [reference_model]”*  
instead of manually configuring the template in trl. In any case we only support one, very simple chat template, so it's not very satisfying.

trl/models/utils.py
```diff
@@ -132,6 +140,75 @@ def setup_chat_format(
...
+        tokenizer (`PreTrainedTokenizer`):
+            Tokenizer to update.
+        source_tokenizer_path (`str`):
+            Path or identifier of the pretrained tokenizer to clone from.
+        resize_to_multiple_of (`int` or `None`, *optional*, defaults to `64`):
+            The embedding layer will be resized to the new vocabulary size. If this is not `None`, it will round up the
+            new vocabulary size to the nearest multiple of this value.
+
+    Returns:
+    model (`PreTrainedModel`):
```
qgallouedec:
```suggestion
        model (`PreTrainedModel`):
```

trl/scripts/sft.py
```diff
@@ -104,7 +104,8 @@ def main(script_args, training_args, model_args):
 
     # Set default chat template if needed
     if tokenizer.chat_template is None:
-        model, tokenizer = setup_chat_format(model, tokenizer, format="chatml")
+        # TODO: source should be passed as an argument
+        model, tokenizer = clone_chat_template(model, tokenizer, "Qwen/Qwen3-0.6B")
```
lewtun, line 108: We have found it useful internally to expose a `chat_template` arg in `SFTConfig` which allows one to define a custom template or copy-paste one from an existing model. Perhaps we could expose this along with a `chat_template_clone` arg (or something similar), now that I better understand what your intent was in this PR?

qgallouedec: sounds good!

qgallouedec: #3599

PR #3272, opened 2025-04-09
🎁 Reward takes completion ids

qgallouedec:
This PR provides access to completion tokens in reward functions. This enables, for example, the implementation of Soft Overlong Punishment (from DAPO).

PR #3248, opened 2025-04-07
☕ Overlong-filtering for GRPO

shirinyamani:
# What does this PR do?

changes:
- simplified completion_mask to not change anything in the loss compute to be compatible with the lig-loss, 
- intended to further extend it to remove the truncated_mask_completions from the Advtg 

simple script to take it for spin;
```python
from datasets import load_dataset
from trl.trainer.grpo_trainer import GRPOTrainer
from trl.trainer.grpo_config import GRPOConfig


dataset = load_dataset("trl-lib/tldr", split="train[:200]")


def reward_func(completions, **kwargs):
    return [len(set(c)) for c in completions]

args = GRPOConfig(
    output_dir="mask_truncated_ture_with_max_completion_length",
    use_vllm=False,
    bf16=True,
    gradient_checkpointing=True,
    logging_steps=50,
    mask_truncated_completions=True, 
    num_train_epochs=1,
    per_device_train_batch_size=4,
    gradient_accumulation_steps=1,
    num_generations=4,  
    max_completion_length=50,
)


trainer = GRPOTrainer(
    model="Qwen/Qwen2.5-0.5B",
    args=args,
    reward_funcs=reward_func,
    train_dataset=dataset,
)


trainer.train() 
```
then do;
```python
accelerate launch 3248.py
```
## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [x] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #3413, opened 2025-05-05
Feature: Implemented DAPO

lavaman131:
We just finished implementing DAPO and applied it to training small language models, specifically with `unsloth/Qwen2.5-0.5B-Instruct` [here](https://github.com/lavaman131/reasoning-slms) (includes DAPO config usage) and were able to get some positive results as described in my [technical blog](https://alexlavaee.me/projects/reasoning-slms/).

I thought it might be useful to submit a PR in case the TRL community would benefit from a DAPO implementation. I tried to follow the rough structure of the GRPO trainer, but am of course happy to get any feedback on the code.

Here is a rough overview of the proposed API extending off of what the current GRPO API was missing:

* Soft Overlong Punishment

```python
use_soft_overlong_punishment: Optional[bool] = field(
    default=True,
    metadata={
        "help": "A length-aware penalty mechanism designed to shape the reward for truncated samples. According to the "
        "[DAPO](https://huggingface.co/papers/2503.14476) paper, this signals to the model to avoid excessively long responses."
    },
)
soft_punish_cache: Optional[int] = field(
    default=0,
    metadata={
        "help": "The L_cache term in the soft overlong punishnment calculation. According to the "
        "[DAPO](https://huggingface.co/papers/2503.14476) paper, this term determines how close"
        "to the max completion length a response can get before being penalized."
    },
)
```

* Dynamic Sampling

```python
use_dynamic_sampling: Optional[bool] = field(
    default=True,
    metadata={
        "help": "Whether to use dynamic sampling to filter out prompts that have zero variance from the "
        "[DAPO](https://huggingface.co/papers/2503.14476) paper."
    },
)
```

### DIFF
--- a/README_dev.md
+++ b/README_dev.md
@@ -4,5 +4,4 @@
 
 ```bash
 python -m pip install -r requirements_dev.txt
-python -m pip install flash-attn --no-build-isolation
 ```
--- a/README_dev.md
+++ b/README_dev.md
@@ -1,7 +1,0 @@
-# Development Setup
-
-## Package Installation
-
-```bash
-python -m pip install -r requirements_dev.txt
-```
--- a/README_dev.md
+++ b/README_dev.md
@@ -1,0 +1,7 @@
+# Development Setup
+
+## Package Installation
+
+```bash
+python -m pip install -r requirements_dev.txt
+```
--- a/README_dev.md
+++ b/README_dev.md
@@ -4,4 +4,5 @@
 
 ```bash
 python -m pip install -r requirements_dev.txt
+python -m pip install flash-attn --no-build-isolation
 ```
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,19 +1,3 @@
 -e .[dev]
-unsloth_zoo
-packaging
-tyro
-transformers>=4.46.1
-datasets>=2.16.0
-sentencepiece>=0.2.0
-tqdm
-psutil
-wheel>=0.42.0
-numpy
-accelerate>=0.34.1
-peft>=0.7.1
-protobuf<4.0.0
-huggingface_hub
-hf_transfer
-bitsandbytes>=0.45.1
-unsloth[cu124onlytorch260]
-unsloth[flashattention]
+jupyter
+ipykernel
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -2,3 +2,4 @@
 wandb
 jupyter
 ipykernel
+ninja
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,6 +1,0 @@
--e .[dev]
-wandb
-jupyter
-ipykernel
-ninja
-hf_transfer
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,0 +1,19 @@
+-e .[dev]
+unsloth_zoo
+packaging
+tyro
+transformers>=4.46.1
+datasets>=2.16.0
+sentencepiece>=0.2.0
+tqdm
+psutil
+wheel>=0.42.0
+numpy
+accelerate>=0.34.1
+peft>=0.7.1
+protobuf<4.0.0
+huggingface_hub
+hf_transfer
+bitsandbytes>=0.45.1
+unsloth[cu124onlytorch260]
+unsloth[flashattention]
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -3,3 +3,4 @@ wandb
 jupyter
 ipykernel
 ninja
+hf_transfer
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,3 +1,4 @@
 -e .[dev]
+wandb
 jupyter
 ipykernel
--- a/setup.py
+++ b/setup.py
@@ -129,7 +129,7 @@ setup(
     zip_safe=False,
     version=__version__,
     description="Train transformer language models with reinforcement learning.",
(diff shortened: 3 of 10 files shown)


LeonEricsson:
The current implementation already supports overlong punishment through a custom reward function, so I’m not sure a separate DAPO trainer is necessary just for dynamic sampling. It's more practical+maintainable to integrate that functionality into the existing GRPO trainer. Also, have to performed any wall clock benchmarks for dynamic sampling? It certainly reduces the number of optimization steps but I'm not convinced it significantly improves training time.

lavaman131:
Thanks for the feedback @LeonEricsson.

Regarding your questions:

1. I wasn't able to find an implementation of soft overlong punishment in the documentation and also not in [open_r1](https://github.com/huggingface/open-r1/blob/main/src/open_r1/rewards.py) where @shirinyamani and @qgallouedec are making some changes. I implemented the token-level soft-overlong punishment as described in the original paper. To my knowledge, the reward function API was not yet modified to take the `completion_ids` and was proposed [here](https://github.com/huggingface/trl/pull/3272) by @qgallouedec to support this. So, I took the proposed changes there and added it to my current implementation.

I agree, this doesn't need to be its own API, but once the new reward function callable is supported I'm happy to add a short tutorial in the docs page that uses the GRPOConfig with how to extend it with the custom reward function callable that takes `completion_ids` and dynamic sampling to make it DAPO.

2. Regarding performance benchmarks I've attached a screenshot of my wandb dashboard that compares vanilla GRPO to adding dynamic sampling during training. As of now, I currently implement dynamic sampling by making the loss zero, effectively skipping an optimization step. Is this the recommended way to implement this?
Screenshot 2025-05-05 at 9 18 06 AM

shirinyamani:
Hi @lavaman131 thanks for your contribution. 
I think we have overlong_filtering support in this PR #3248 where we defined an arg to filter the overlong completions.

LeonEricsson:
> 1. I wasn't able to find an implementation of soft overlong punishment in the documentation and also not in [open_r1](https://github.com/huggingface/open-r1/blob/main/src/open_r1/rewards.py) where @shirinyamani and @qgallouedec are making some changes. I implemented the token-level soft-overlong punishment as described in the original paper. To my knowledge, the reward function API was not yet modified to take the `completion_ids` and was proposed [here](https://github.com/huggingface/trl/pull/3272) by @qgallouedec to support this. So, I took the proposed changes there and added it to my current implementation.

Aah right I see, I think its best to wait for #3272 and then implement overlong punishment in [open_r1](https://github.com/huggingface/open-r1/blob/main/src/open_r1/rewards.py). There's actually already an open [PR](https://github.com/huggingface/open-r1/pull/567) for this, the last thing discussed was a comparisons of computing soft_overlong_punishment on characters vs tokens.

> 2. Regarding performance benchmarks I've attached a screenshot of my wandb dashboard that compares vanilla GRPO to adding dynamic sampling during training. As of now, I currently implement dynamic sampling by making the loss zero, effectively skipping an optimization step. Is this the recommended way to implement this?

Hmm, this doesn’t sound quite right, though I might be misunderstanding. Dynamic sampling aims to improve sample efficiency by replacing zero-advantage samples with more informative ones. As training progresses, the number of samples where all generations are correct (or incorrect) tends to increase. These samples don’t contribute any gradients during policy updates, which leads to inefficiency.

To address this, DAPO proposes **over-sampling** - taking advantage of the fact that generation is highly parallelizable - to ensure that the batch is filled with non-zero advantage samples. This could involve continuously drawing new samples until the batch meets this criterion. Alternatively, one might naively increase the batch size by some factor and then filter out the zero-advantage samples, but that would require tuning said factor, which isn’t trivial.

lavaman131:
Ok, thank you both for confirming.

**Notes:**
Yes, I was aware of the overlong filtering, but was referring to the soft overlong punishment from #3272 @shirinyamani.

Then, is it correct to assume that the soft overlong punishment should be implemented as a custom reward function since `completion_ids` is now accepted to the reward callable?

Regarding dynamic sampling @LeonEricsson, my way of preventing the zero-advantage samples was to mask them out and if there are no samples skip the optimization step. Right, I was kind-of doing the latter, where I was using more optimization steps, rather than increasing the batch size. I like your idea of continuously sampling until the batch meets this criterion, however, actually thought it would be messier calling `vllm.generate`.

**Takeaways:**
Based off of this discussion, would the community benefit from a page in the documentation, describing the custom reward function for soft overlong filtering and maybe even adding a flag for dynamic sampling if it is practical to do so in the first place (e.g. due to possible sample inefficiency).

qgallouedec:
> Then, is it correct to assume that the soft overlong punishment should be implemented as a custom reward function since `completion_ids` is now accepted to the reward callable?

I think so yes.

> would the community benefit from a page in the documentation, describing the custom reward function for soft overlong filtering

To take a broader view, I think we could even start to take the reward functions from open-r1 and integrate them into trl as a sub-module. And consequently this sub-module would have its own dedicated documentation page.

> and maybe even adding a flag for dynamic sampling if it is practical to do so in the first place (e.g. due to possible sample inefficiency).

I don't really know what to do about dynamic sampling. For the moment, I don't see any evidence that it would make a difference in compute efficiency, and it would introduce significant complexity into the code. I don't know if there's much more to be done in that direction.

lavaman131:
Thanks, @qgallouedec, that makes sense.

I especially like the idea of a reward functions submodule from open-r1. I could create a draft PR proposing this functionality and would be happy to contribute with these efforts if agreeable?

qgallouedec:
OFC!!

qgallouedec:
> I especially like the idea of a reward functions submodule from open-r1

initial structure in #3430

PR #3415, opened 2025-05-05
🧪 Testing support for Qwen3 tiny 

shirinyamani:
# What does this PR do?

This PR adds support for [trl-internal-testing](https://huggingface.co/trl-internal-testing) for quick tests for Qwen3 tiny models of (1.7, 4 and 8B + 4B for reward modelling purpose)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


scripts/generate_tiny_models.py
```diff
@@ -120,6 +123,9 @@ def push_to_hub(model, tokenizer, prefix=None, suffix=None):
     ("microsoft/Phi-3.5-mini-instruct", Phi3Config, Phi3ForCausalLM, None),
     ("Qwen/Qwen2.5-32B-Instruct", Qwen2Config, Qwen2ForCausalLM, "2.5"),
     ("Qwen/Qwen2.5-Coder-0.5B", Qwen2Config, Qwen2ForCausalLM, "2.5-Coder"),
+    ("Qwen/Qwen3-1.7B", Qwen3Config, Qwen3ForCausalLM, "3"),
+    ("Qwen/Qwen3-4B", Qwen3Config, Qwen3ForCausalLM, "3"),
+    ("Qwen/Qwen3-8B", Qwen3Config, Qwen3ForCausalLM, "3"),
```
qgallouedec, line 128: The tiny models will all have the same id, no?

qgallouedec:
It might be interesting to use these tiny models in our test suite. At least for tests where we test all the tiny models.

scripts/generate_tiny_models.py
shirinyamani: correct, my purpose was to have different sizes of the model.

qgallouedec: The only size you get with this code is about 1M param (called tiny). Look below, the id is only used for loading the tokenizer. That the whole point actually, we don't want to use big models for unit tests.

shirinyamani:
> It might be interesting to use these tiny models in our test suite. At least for tests where we test all the tiny models.

done!

qgallouedec (approved):
- Fixed the script
- Removed the models with bad names
- Updated the collections

Now I think we're ready to merge

PR #3428, opened 2025-05-08
🛠️ quantization support for vllm generation

shirinyamani:
# What does this PR do?

This PR adds support for quantization when generating completions using vllm.  Usage would be:
```python
trl vllm-serve --model Qwen/Qwen2.5-7B  --quantization="fp8"
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/scripts/vllm_serve.py
+++ b/trl/scripts/vllm_serve.py
@@ -189,6 +189,9 @@ class ScriptArguments:
         log_level (`str`, *optional*, defaults to `"info"`):
             Log level for uvicorn. Possible choices: `"critical"`, `"error"`, `"warning"`, `"info"`, `"debug"`,
             `"trace"`.
+        quantization (`str` or `None`, *optional*, defaults to `"fp8"`):
+            If set, all Linear modules (except for the final lm_head) have their weights quantized down to FP8_E4M3
+            precision with a per-tensor scale.
     """
 
     model: str = field(
@@ -266,6 +269,13 @@ class ScriptArguments:
             "'trace'."
         },
     )
+    quantization: Optional[str] = field(
+        default="fp8",
+        metadata={
+            "help": "If set,  all Linear modules (except for the final lm_head) have their weights quantized down"
+            "to FP8_E4M3 precision with a per-tensor scale."
+        },
+    )
 
 
 def llm_worker(
@@ -291,6 +301,7 @@ def llm_worker(
         kv_cache_dtype=script_args.kv_cache_dtype,
         max_model_len=script_args.max_model_len,
         worker_extension_cls="trl.scripts.vllm_serve.WeightSyncWorkerExtension",
+        quantization=script_args.quantization,
     )
 
     # Send ready signal to parent process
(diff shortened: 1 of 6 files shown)


trl/scripts/vllm_serve.py
```diff
@@ -187,6 +187,9 @@ class ScriptArguments:
         log_level (`str`, *optional*, defaults to `"info"`):
             Log level for uvicorn. Possible choices: `"critical"`, `"error"`, `"warning"`, `"info"`, `"debug"`,
             `"trace"`.
+        quantization (`str` or `None`, *optional*, defaults to `"fp8"`):
```
qgallouedec, line 190: by default does vLLM quantize to fp8?

shirinyamani: yes! [Link](https://docs.vllm.ai/en/latest/features/quantization/fp8.html)

qgallouedec: I don't think it does actually: https://docs.vllm.ai/en/latest/api/vllm/vllm.entrypoints.llm.html#vllm.entrypoints.llm.LLM

qgallouedec: it seems like it only _supports_ it, no?

shirinyamani: > I don't think it does actually: https://docs.vllm.ai/en/latest/api/vllm/vllm.entrypoints.llm.html#vllm.entrypoints.llm.LLM

When you go down in the same link there is a parameter called quantization where they also say fp8 (experimental), I believe this means they support it. Isn't ? 
Plus in the link above if you scroll down to online quantization section there is exactly an example off llm() where they pass in quantization as parameter

qgallouedec: My question is, does vLLM quantize the model to FP8 by default? And from my reading of the documentation, it doesn't seem to. So I don't think we should overwrite this default unless we can show that it's a much better choice in our case.

shirinyamani: oh i get what you meant now, so no, it does it only if the user pass `--quantization="fp8" `. that is why I introduced it in the script args so that the user has this freedom of choice! Does this answer satisfy your question? or am i mistaken?

shirinyamani: Oh I see what you mean now and you are right. I am setting it to solely fp8 here.  so you want us to put it as optional None and the user can pick each of the supported one of choice?

> quantization – The method used to quantize the model weights. Currently, we support “awq”, “gptq”, and “fp8” (experimental). If None, we first check the quantization_config attribute in the model config file. If that is None, we assume the model weights are not quantized and use dtype to determine the data type of the weights.

shirinyamani: sure

qgallouedec:
It's possible that the difference in data types between the server's weights and those received by the trainer could cause problems. Have you checked whether it's possible to run GRPO with FP8 quantization?
Also, it might be good to benchmark the speedup, even briefly.
Ideally, we should also check that this doesn't negatively impact training (though this isn't strictly necessary to merge the PR).

qgallouedec:
closing as no recent update

PR #3430, opened 2025-05-09
🎁 Reward submodule

qgallouedec:
This PR creates a sub-module dedicated to reward functions.

Please note that this PR is not intended to integrate _all_ reward functions; this may be done in the future. Its purpose is simply to create a structure for future additions.

### DIFF
--- a/tests/test_rewards.py
+++ b/tests/test_rewards.py
@@ -1,0 +1,64 @@
+# Copyright 2020-2025 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+from trl.rewards import think_format_reward
+
+
+class ThinkFormatRewardTester(unittest.TestCase):
+    def test_valid_format(self):
+        completions = [
+            "<think>This is my reasoning.</think>This is my answer.",  # Simple, one-line reasoning
+            "<think>\nThis is my reasoning.\n</think>\nThis is my answer.",  # Multiline reasoning
+            "<think>\nThis is\nmy reasoning.\n</think>\nThis is my answer.",  # Multiline reasoning
+            "<think>\nThis is <some tag> my reasoning.</think>\nThis is my answer.",  # Reasoning including other tags
+            "<think></think>\nThis is my answer.",  # Empty reasoning
+        ]
+        completions = [[{"content": completion}] for completion in completions]
+        expected_rewards = [1.0, 1.0, 1.0, 1.0, 1.0]  # All should be valid
+        rewards = think_format_reward(completions)
+        self.assertEqual(rewards, expected_rewards)
+
+    def test_invalid_format(self):
+        completions = [
+            "<think>\nThis is my reasoning.\nThis is my answer.",  # No closing </think>
+            "<think>This is my reasoning.\nThis is my answer.",  # No closing </think>
+            "This is my reasoning. This is my answer.",  # No <think> tags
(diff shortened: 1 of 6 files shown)


trl/rewards/format_rewards.py
```diff
@@ -0,0 +1,46 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+
+
+def think_answer_format_reward(completions: list[list[dict[str, str]]], **kwargs) -> list[float]:
```
lewtun, line 18: FYI this particular format is legacy from the [R1 tech report](https://arxiv.org/pdf/2501.12948):

Screenshot 2025-05-09 at 19 36 38

However, if you distill the R1 model itself, you find it only produces `<think>` and `</think>` tags, i.e. no `<answer>` tags at all, e.g. : https://huggingface.co/datasets/open-r1/OpenR1-Math-220k/viewer/default/train?row=0

This I recommend we change this reward function to simply `think_format_reward` and enforce `<think>\nThought section\n</think>\nSolution section.` This has the added advantage of being easier to learn and is consistent with policies initialised from R1 traces

qgallouedec: Sounds good! done in 335bcc7a1ae86827f69b918ac369bf8878daeea0

lavaman131:
Hi @qgallouedec, thanks for setting up the boilerplate.

Are you intending the rewards to be importing as paths to the module or strings in a reward function registry? (that is the impression I got from skimming your changes). As a user, I would think it would be most intuitive to import reward callables from the reward module instead and include them as a `List[Callable]` type like the current API. Please correct me if I'm misunderstanding. That would also give very simple extensibility for custom reward functions.

If you do intend the registry option, maybe something similar to the gym register environment API would be interesting? For example, you could use a string argument for names of functions in the registry and have a static method to query the existing functions and their descriptions. And for custom reward functions, there could be a `register_reward_fn` which updates the registry.

qgallouedec:
> If you do intend the registry option, maybe something similar to the gym register environment API would be interesting? 

I'm strongly against this idea tbh. Gym registry is a nightmare and not pythonic at all

lavaman131:
> > If you do intend the registry option, maybe something similar to the gym register environment API would be interesting?
> 
> I'm strongly against this idea tbh. Gym registry is a nightmare and not pythonic at all

Ah ok 😅, fair enough.

So, to clarify, will the structure remain the same outside of the cli, where the user includes their rewards functions as a list of callables. But, now they can import from this module as well for common reward functions? (Just want to scope out how implementing rewards functions for the module would work).

qgallouedec:
My idea is that everything remain the same for the user. The only difference would be that they would be able to do

```python
from trl.rewards import think_format_reward
```

Instead of having to implement it themselves, or copy from another library

lavaman131:
Awesome, that makes sense and aligns with what I was thinking as well. Once this change is out, happy to work on adding to the submodule!

qgallouedec:
Perfect! Thanks!


If you want to start now, you can you create a new branch from this one

tests/test_rewards.py
```diff
@@ -0,0 +1,64 @@
...
+        rewards = think_format_reward(completions)
+        self.assertEqual(rewards, expected_rewards)
+
+    def test_invalid_format(self):
+        completions = [
+            "<think>\nThis is my reasoning.\nThis is my answer.",  # No closing </think>
+            "<think>This is my reasoning.\nThis is my answer.",  # No closing </think>
+            "This is my reasoning. This is my answer.",  # No <think> tags
+            "This is my reasoning.\nThis is my answer.",  # No <think> tags
+            "This is my reasoning</think>\nThis is my answer.",  # No opening <think>
```
lewtun, line 40: Should we also test no opening `<think>`, but also a closing `</think>` e.g.

```suggestion
            "This is my reasoning</think>\nThis is my answer.",  # No opening <think>
            "This is a preamble<think>This is my reasoning</think>This is my answer.",
```

lewtun (approved):
LGTM with a question on testing whether the think process can have a preamble (and should thus invalid)

tests/test_rewards.py
qgallouedec: Yep! e776045b63151f4f54b041d1d6cc9f1381abdb0d

PR #3434, opened 2025-05-12
feat: Implement Two-Sided Clipping for GRPO Trainer

### LINKED ISSUE #3435: Implement Two-Sided Clipping for GRPO Trainer
### Feature request

This feature proposal is to implement a two-sided clipping mechanism for the GRPO (Group Relative Policy Optimization) trainer. This modification addresses a potential stability issue in the standard GRPO formulation.

The proposed objective function is:
J_GRPO(θ) = E[ min( min(ratio, δ) * Â_t, clip(ratio, 1-ε, 1+ε) * Â_t) ]
(Full formula: J_GRPO(θ) = \mathbb{E}_{q\sim P(Q), \{o_i\}_{i=1}^G \sim \pi_{\theta_{old}}(O|q)} \left[ \frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|} \min \left( \min\left(\frac{\pi_\theta(o_{i,t}|q, o_{i,<t})}{\pi_{\theta_{old}}(o_{i,t}|q, o_{i,<t})}, \delta\right) \hat{A}_{i,t}, \text{clip}\left(\frac{\pi_\theta(o_{i,t}|q, o_{i,<t})}{\pi_{\theta_{old}}(o_{i,t}|q, o_{i,<t})}, 1-\epsilon, 1+\epsilon\right) \hat{A}_{i,t} \right) \right])

This introduces a new hyperparameter, `delta` (δ), to `GRPOConfig`. This parameter caps the probability ratio for negative advantages.
The implementation involves:
1. Adding `delta` to `trl/trainer/grpo_config.py`.
2. Modifying `_compute_loss` in `trl/trainer/grpo_trainer.py` to use this new clipping logic.
3. Adding a corresponding unit test in `tests/test_grpo_trainer.py`.

### Motivation

The standard GRPO formulation can encounter stability issues, particularly when negative advantages (Â_t < 0) coincide with very large probability ratios (π_θ / π_θ_old). In such cases, the original clipping mechanism (which only applies when the ratio is too small for negative advantages) can lead to extremely large policy updates, potentially destabilizing the training process.

The proposed two-sided clipping mechanism aims to mitigate this by introducing an upper bound (`delta`) on the probability ratio when advantages are negative. This allows for significant updates but prevents the extreme changes that could harm training stability and robustness. The recommendation is to set `delta > 1 + epsilon` to ensure this balance.

### Your contribution

Will open a PR

ucalyptus:
As seen in the new [Prime Intellect's Intellect-2 Technical Report](https://storage.googleapis.com/public-technical-paper/INTELLECT_2_Technical_Report.pdf)
# What does this PR do?

This PR introduces a two-sided clipping mechanism to the GRPO (Group Relative Policy Optimization) trainer. This change aims to address a potential stability issue in the standard GRPO formulation where, for negative advantages (\(\hat{A}_{i,t} < 0\)), the original clipping only applied when the probability ratio was too small. This could lead to extremely large updates if the ratio became very large.

The core modification implements the following objective:
image

A new hyperparameter, `delta`, has been added to `GRPOConfig`. This parameter caps the probability ratio for negative advantages. It is recommended to set \(\delta > 1+\epsilon\) to allow significant updates while preventing extreme changes that could destabilize training.

The changes include:
1.  **`trl/trainer/grpo_config.py`**: Added the `delta` hyperparameter.
2.  **`trl/trainer/grpo_trainer.py`**: Modified `_compute_loss` to implement the two-sided clipping.
3.  **`tests/test_grpo_trainer.py`**: Added `test_two_sided_clipping_loss` to verify the new logic.

Fixes #3435 

## Before submitting
      Pull Request section? (Assumed based on PR process)
      to it if that's the case.
- [x] Did you make sure to update the documentation with your changes? (Docstring for `delta` in `GRPOConfig` was added.)
- [X] Did you write any new necessary tests? (The `test_two_sided_clipping_loss` test was added.)

## Who can review?

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1091,6 +1091,8 @@ class GRPOTrainerTester(unittest.TestCase):
             previous_trainable_params = {n: param.clone() for n, param in trainer.model.named_parameters()}
             trainer.train()
 
+            self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
+            
             # Check if the model parameters have changed
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1204,7 +1204,6 @@ class GRPOTrainerTester(unittest.TestCase):
             model_id = "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5"
             tokenizer = AutoTokenizer.from_pretrained(model_id)
 
-            # Load minimal dataset
             dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train[:1]")
 
             # Create trainer
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1209,7 +1209,6 @@ class GRPOTrainerTester(unittest.TestCase):
             # Create trainer
             trainer = self._make_delta_trainer(tmp_dir, tokenizer, dataset)
 
-            # Prepare inputs
             inputs = self._delta_inputs(trainer.accelerator.device)
             inputs.update(
                 {
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1096,11 +1096,6 @@ class GRPOTrainerTester(unittest.TestCase):
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
 
-    def test_training_with_peft(self):
-        # This method is not provided in the original file or the new file
-        # It's assumed to exist as it's called in the test_training_with_mask_truncated_completions method
-        pass
-
     def test_training_with_mask_truncated_completions_all_masked(self):
         """
         Test that when all generated completions are truncated (i.e., none contain an EOS token), and
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1031,39 +1031,19 @@ class GRPOTrainerTester(unittest.TestCase):
     def test_training_with_mask_truncated_completions(self, mock_generate):
         """Test that training works with mask_truncated_completions=True parameter."""
 
-        # Initialize tokenizer locally for this test
-        model_id_for_tokenizer = "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5"
-        tokenizer = AutoTokenizer.from_pretrained(model_id_for_tokenizer)
-        if tokenizer.pad_token is None:
-            tokenizer.pad_token = tokenizer.eos_token
-
         # We mock the generate method because the model's random weights make it extremely unlikely to produce a
         # sequence containing the EOS token within the allowed max_completion_length. As a result, all tokens are
         # masked in the loss, the model doesn't update, and the final check (which verifies the update) fails.
         def fake_generate(prompt_ids, **kwargs):
-            batch_size = prompt_ids.shape[0]
-            max_completion_length = kwargs.get(
-                "generation_config"
-            ).max_new_tokens  # Get max_new_tokens from GenerationConfig
-
             # pad_token_id = 151643; eos_token_id = 151645
-            base_completions = torch.tensor(
+            completions_ids = torch.tensor(
                 [
-                    [1] * max_completion_length,  # Truncated example
-                    [9] * (max_completion_length // 2)
-                    + [tokenizer.eos_token_id]
-                    + [tokenizer.pad_token_id]
-                    * (max_completion_length - max_completion_length // 2 - 1),  # EOS example
-                    [12] * (max_completion_length - 1) + [tokenizer.eos_token_id],  # EOS at the end example
+                    [1, 2, 3, 4, 5, 6, 7, 8],  # this one is truncated
+                    [9, 10, 11, 151645, 151643, 151643, 151643, 151643],  # this one contains eos
+                    [12, 13, 14, 15, 16, 17, 18, 151645],  # particular case, eos is generated just within the limit
                 ],
                 device=prompt_ids.device,
-                dtype=torch.long,
             )
-            # Repeat the base completions to match the required batch size
-            completions_ids = base_completions.repeat(
-                (batch_size + base_completions.shape[0] - 1) // base_completions.shape[0], 1
-            )[:batch_size]  # Ensure correct batch size
-
             return torch.cat([prompt_ids, completions_ids], dim=1)
 
         mock_generate.side_effect = fake_generate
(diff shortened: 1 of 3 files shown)


ucalyptus:
@qgallouedec

kashif:
@ucalyptus what are your thoughts on making this optional with regards to the default config values?

kashif:
what i mean is to then have the default be `inf` and a recommendation in the help message for good values perhaps else code using the default values might change the behaviour unexpectedly

qgallouedec:
Agree with @kashif. Maybe even `None` for readability.

ucalyptus:
Ok so here are the changes I made:

`GRPOConfig` now defaults delta to None and has an updated docstring.
`GRPOTrainer._compute_loss` now only applies the upper delta clipping when `delta` is explicitly set to a float value.

Feel free to push back if there are further concerns.

kashif:
@ucalyptus can you kindly run the `make precommit` in the root dir of TRL to fix the formatting?

kashif:
@ucalyptus  the failing CI test is due to a known issue in the dev so can be ignored

kashif:
@ucalyptus  do we want to add a warning that this config will be ignored when the Liger kernel losss is enabled?

edbeeching:
> @ucalyptus do we want to add a warning that this config will be ignored when the Liger kernel losss is enabled?

Gvien that logs are are flooded with info / warnings from transformers and other libs. It might be better to explicitly raise a ValueError when liger + delta clip is enabled together. So the user can disable it and won't draw any wrong conclusions.

kashif (approved):

tests/test_grpo_trainer.py
```diff
@@ -1065,15 +1085,13 @@ def fake_generate(prompt_ids, **kwargs):
...
             )
 
             previous_trainable_params = {n: param.clone() for n, param in trainer.model.named_parameters()}
-
             trainer.train()
 
-            self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
-
-            # Check that the params have changed
+            # Check if the model parameters have changed
```
kashif, line 1094:
```suggestion
            self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
            
            # Check if the model parameters have changed
```

tests/test_grpo_trainer.py
```diff
@@ -1065,15 +1085,15 @@ def fake_generate(prompt_ids, **kwargs):
...
             )
 
             previous_trainable_params = {n: param.clone() for n, param in trainer.model.named_parameters()}
-
             trainer.train()
 
             self.assertIsNotNone(trainer.state.log_history[-1]["train_loss"])
-
-            # Check that the params have changed
+
```
kashif, line 1095:
```suggestion
```

tests/test_grpo_trainer.py
```diff
@@ -1031,19 +1031,39 @@ def test_training_no_scale_rewards(self):
...
+        model_id_for_tokenizer = "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5"
+        tokenizer = AutoTokenizer.from_pretrained(model_id_for_tokenizer)
+        if tokenizer.pad_token is None:
+            tokenizer.pad_token = tokenizer.eos_token
+
         # We mock the generate method because the model's random weights make it extremely unlikely to produce a
         # sequence containing the EOS token within the allowed max_completion_length. As a result, all tokens are
         # masked in the loss, the model doesn't update, and the final check (which verifies the update) fails.
         def fake_generate(prompt_ids, **kwargs):
+            batch_size = prompt_ids.shape[0]
```
lewtun, line 1044: I don't quite understand why we need to change this method to accommodate the change in the trainer. In particular, why are we changing an unrelated test? Please revert if these changes are not needed for `test_two_sided_clipping_loss()`

tests/test_grpo_trainer.py
```diff
@@ -1147,3 +1166,182 @@ def test_training_num_generations_larger_than_batch_size(self):
...
+
+    def test_two_sided_clipping_loss(self):
+        """
+        Tests the two-sided GRPO clipping logic with specific scenarios.
+        Uses a completion length of 2 to ensure logp and loss masking work.
+        """
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            model_id = "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5"
+            tokenizer = AutoTokenizer.from_pretrained(model_id)
+            if tokenizer.pad_token is None:
```
lewtun, line 1178: This model has a [pad token](https://huggingface.co/trl-internal-testing/tiny-Qwen2ForCausalLM-2.5/blob/main/tokenizer_config.json#L204), so this logic is not necessary

tests/test_grpo_trainer.py
```diff
@@ -1147,3 +1166,182 @@ def test_training_num_generations_larger_than_batch_size(self):
...
+                delta=2.0,  # delta > 1 + epsilon (1.2)
+                epsilon_high=0.2,  # Corresponds to self.epsilon_high in trainer
+                beta=0.0,  # No KL divergence for this specific loss test
+                loss_type="bnpo",  # Default, simplifies loss for single token
+                max_completion_length=2,  # Use completion length 2
+                report_to="none",
+            )
+
+            trainer = GRPOTrainer(
+                model=model_id,  # Use local model_id
```
lewtun, line 1196:
```suggestion
                model=model_id,
```

tests/test_grpo_trainer.py
```diff
@@ -1147,3 +1166,182 @@ def test_training_num_generations_larger_than_batch_size(self):
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
+
+    def test_two_sided_clipping_loss(self):
```
lewtun, line 1170: Instead of bundling all scenarios into one test, let's parametrise them with something like this:

```py
@parameterized.expand([
    # name ,  adv , old_p , new_p , expected_loss
    ("pos_ratio_in_clip",      2.0, 0.50, 0.55, -2.2),
    ("pos_ratio_above_clip",   2.0, 0.40, 0.60, -2.4),
    ("neg_ratio_in_clip",     -2.0, 0.50, 0.45,  1.8),
    ("neg_ratio_below_clip",  -2.0, 0.50, 0.35,  1.6),
    ("neg_ratio_above_delta", -2.0, 0.20, 0.50,  4.0),
    ("neg_between_clip_delta",-2.0, 0.40, 0.60,  3.0),
])
```

You could then add some helper methods like this:

```python
@staticmethod
def _make_delta_trainer(tmp_dir, tokenizer, dataset):
    cfg = GRPOConfig(
        output_dir=tmp_dir,
        epsilon=0.20, delta=2.0,
        epsilon_high=0.20, beta=0.0,
        loss_type="bnpo", max_completion_length=2,
        report_to="none",
    )
    return GRPOTrainer(
        args=cfg,
        reward_funcs=lambda x, **y: [1.0] * len(x),
        train_dataset=dataset,
        processing_class=tokenizer,
    )

@staticmethod
def _delta_inputs(device):
    return dict(
        prompt_ids      = torch.tensor([[101]],  device=device),
        prompt_mask     = torch.tensor([[1]],    device=device),
        completion_ids  = torch.tensor([[2000, 2001]], device=device),
        completion_mask = torch.tensor([[1, 1]], device=device),
    )
```

which are then called within the test. This has the advantage of making the code much simpler to read, but also ensuring we can see exactly which test is failing if there's a regression.

lewtun (changes requested):
Thank you very much for the quick implementation of this feature @ucalyptus !

The trainer logic looks sound to me, but the tests need to be refactored with `parameterised` so that we can maintain them / better catch regressions. 

I also see quite a few unnecessary code comments, so please remove most of them, except in cases where some clarification is required.

ucalyptus:
@kashif thanks for helping with the PR, and comments by @lewtun. Let me know if there are any outstanding concerns left.

lewtun (approved):
Thanks for iterating @ucalyptus and @kashif ! LGTM with some final nits

PR #3444, opened 2025-05-13
enable activation offloading on XPU

yao-matrix:


trl/models/activation_offloading.py
```diff
@@ -85,11 +86,17 @@ def __init__(
         self.use_pin_memory = use_pin_memory
         self.virtual_memory_safe_pct = 60  # we should not exceed this percentage of memory
 
-        self.s0 = torch.cuda.default_stream()  # comp stream
+        self.accelerator_type = (
+            torch.accelerator.current_accelerator().type if hasattr(torch, "accelerator") else "cuda"
+        )
+        # NOTE: xpu doesn't have `default_stream` API, use `current_stream` instead
+        self.s0 = (
+            torch.xpu.current_stream() if self.accelerator_type == "xpu" else torch.cuda.default_stream()
```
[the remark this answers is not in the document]
gujinghui: LGTM. As long as the workload has no specific assumption on cuda default stream, the current stream is good enough to replace it for functionality.

yao-matrix:
@kashif , pls help review, thx.

trl/models/activation_offloading.py
```diff
@@ -85,11 +86,17 @@ def __init__(
         self.use_pin_memory = use_pin_memory
         self.virtual_memory_safe_pct = 60  # we should not exceed this percentage of memory
 
-        self.s0 = torch.cuda.default_stream()  # comp stream
+        self.accelerator_type = (
```
kashif, line 89: just thinking... would this still be `"cuda"` if we have no cuda gpus (i.e. on a CPU only setup?)

yao-matrix: just my 2 points.
1) Yes, in a CPU-only machine, it would still be `"cuda"` with current code logic and there will be `AssertError` as below:
  File "/usr/local/lib/python3.11/dist-packages/torch/cuda/__init__.py", line 1095, in default_stream
    _lazy_init()
  File "/usr/local/lib/python3.11/dist-packages/torch/cuda/__init__.py", line 370, in _lazy_init
    raise AssertionError("Torch not compiled with CUDA enabled")
AssertionError: Torch not compiled with CUDA enabled
2) This is a acceptable behavior to users, i mean if i'm the user,:). Since the activation offloading feature needs an accelerator by its nature, so if there is no accelerator, we need fail with errors to make user know, there are a few options to let user know, `AssertError` is one of these and in this case the feature will no run in any case(i mean not like in other cases we can fallback to some other path), so assert out is acceptable.

trl/models/activation_offloading.py
```diff
@@ -24,6 +24,7 @@
 import torch
 from torch import nn
 from torch.autograd.graph import saved_tensors_hooks
+from transformers.utils import is_torch_accelerator_available
```
kashif, line 27:
```suggestion
```

kashif (approved):

PR #3449, opened 2025-05-15
[SFT] add warning if dataset's input_ids exceed max_length

HERIUN:
# What does this PR do?

setting default max_length=1024 may problem while long text dataset.

user have to know dataset's max_length and truncated with max_length.
 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@kashif @qgallouedec

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -635,7 +635,7 @@ class SFTTrainer(Trainer):
 
                         # Create a completion mask
                         completion_mask = [0] * len(prompt_ids) + [1] * (len(prompt_completion_ids) - len(prompt_ids))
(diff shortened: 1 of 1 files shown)


qgallouedec:
Thanks for the PR! I appreciate the intention behind adding visibility into truncation behavior. However, I'm not sure that raising a warning in this case is the best approach. Truncation can be a normal and intentional part of preprocessing, and emitting a warning during expected behavior can be noisy and misleading, especially in workflows where truncation is desired.

Instead, what might be more useful is a utility function (or method, I'm not exactly sure at this point) that prints or returns length statistics (e.g., quantiles) across the dataset. This would give users a clearer view of the lengths distribution and help them to choose the max length.

PR #3459, opened 2025-05-16
🛠️ Initialize reward_kwargs to prevent UnboundLocalError in GRPOTrainer

teilomillet:
This change initializes the reward_kwargs variable as an empty dictionary to avoid potential UnboundLocalError during its usage in the GRPOTrainer class. This ensures that the variable is defined before it is accessed.

# What does this PR do?

This PR addresses a potential `UnboundLocalError` that can occur in `trl.trainer.GRPOTrainer._generate_and_score_completions`.

The `reward_kwargs` dictionary is intended to store additional arguments (beyond prompt and completion) that are passed to custom, non-module-based reward functions. This dictionary is populated within an `else` block when iterating through `self.reward_funcs`.

If `self.reward_funcs` is empty (e.g., when a downstream library or user provides an empty list, intending to bypass TRL's internal reward calculation and supply pre-computed rewards/advantages) or if all provided reward functions are instances of `torch.nn.Module`, the `else` block that defines `reward_kwargs` is never entered.

Later in the same method, there's a warning mechanism that checks if all reward functions returned `None` for a sample. This warning message attempts to access `reward_kwargs` to provide context:
`row_reward_kwargs = {key: value[nan_row_idx] for key, value in reward_kwargs.items()}`.
If `reward_kwargs` was not initialized due to the conditions mentioned above, this access results in an `UnboundLocalError`, crashing the process.

This PR fixes the error by ensuring `reward_kwargs` is always defined by initializing it to an empty dictionary (`{}`) before the loop that processes reward functions. This allows the warning logic to execute safely, even if no custom reward functions populate `reward_kwargs`.

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1139,6 +1139,8 @@ class GRPOTrainer(Trainer):
             completions = completions_text
 
         rewards_per_func = torch.zeros(len(prompts), len(self.reward_funcs), device=device)
+
+        reward_kwargs = {} # Initialize reward_kwargs to an empty dict to prevent UnboundLocalError
         for i, (reward_func, reward_processing_class, reward_func_name) in enumerate(
             zip(self.reward_funcs, self.reward_processing_classes, self.reward_func_names)
         ):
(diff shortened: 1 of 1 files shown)


qgallouedec (approved):
Thanks!

PR #3470, opened 2025-05-19
Allow an user to train from a local dataset

gogo2464:
# What does this PR do?

Allow an user to train from a local dataset. Tested locally on sft. Does not have any unit testing yet. I also need to provide more documentation.

Fixes # (issue)

I started the PR before raising an issue. I did not wanted to waste developer time.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,12 @@ trl sft --model_name_or_path Qwen/Qwen2.5-0.5B \
     --output_dir Qwen2.5-0.5B-SFT
 ```
 
+```bash
+trl sft --model_name_or_path Qwen/Qwen2.5-0.5B \
+    --dataset_name ./dataset.json \
+    --output_dir Qwen2.5-0.5B-SFT
+```
+
 **DPO:**
 
 ```bash
--- a/trl/scripts/sft.py
+++ b/trl/scripts/sft.py
@@ -110,14 +110,21 @@ def main(script_args, training_args, model_args):
     ################
     # Dataset
     ################
-    print("loading: {0}".format(script_args.dataset_name))
-    
-    if os.path.exists(script_args.dataset_name) == True:
-        print("loading dataset from local file")
-        dataset = load_dataset("json", data_files=script_args.dataset_name)
+    if script_args.dataset_path is not None:
+        if os.path.exists(script_args.dataset_path) == True:
+            print("loading: {0}".format(script_args.dataset_path))
+            if os.path.isdir(script_args.dataset_path):
+                dataset = load_dataset("imagefolder", data_files=script_args.dataset_path)
+            else:
+                print("loading dataset from local file")
+                dataset = load_dataset("json", data_files=script_args.dataset_path)
     else:
-        print("loading dataset from online")
-        dataset = load_dataset(script_args.dataset_name, name=script_args.dataset_config)
+        if script_args.dataset_path is not None:
+            print("loading dataset online")
+            dataset = load_dataset(script_args.dataset_name, name=script_args.dataset_config)
+        else:
+            print("you must choose at least one argument between --dataset_path and dataset_name. Exiting.")
+        
 
     ################
     # Training
--- a/trl/scripts/sft.py
+++ b/trl/scripts/sft.py
@@ -110,16 +110,18 @@ def main(script_args, training_args, model_args):
     ################
     # Dataset
     ################
+    
+    print("{0}, {1}".format(script_args.dataset_name, script_args.dataset_path))
+    
     if script_args.dataset_path is not None:
+        print("loading dataset from local file")
         if os.path.exists(script_args.dataset_path) == True:
-            print("loading: {0}".format(script_args.dataset_path))
-            if os.path.isdir(script_args.dataset_path):
-                dataset = load_dataset("imagefolder", data_files=script_args.dataset_path)
+            if not os.path.isdir(script_args.dataset_path):
+                dataset = load_dataset(path=os.getcwd(), data_files=script_args.dataset_path)
             else:
-                print("loading dataset from local file")
-                dataset = load_dataset("json", data_files=script_args.dataset_path)
+                dataset = load_dataset("imagefolder", data_files=script_args.dataset_path)
     else:
-        if script_args.dataset_path is not None:
+        if script_args.dataset_name is not None:
             print("loading dataset online")
             dataset = load_dataset(script_args.dataset_name, name=script_args.dataset_config)
         else:
--- a/trl/scripts/sft.py
+++ b/trl/scripts/sft.py
@@ -52,6 +52,7 @@ python trl/scripts/sft.py \
 """
 
 import argparse
+import os
 
 from datasets import load_dataset
 from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
@@ -109,7 +110,14 @@ def main(script_args, training_args, model_args):
     ################
     # Dataset
     ################
-    dataset = load_dataset(script_args.dataset_name, name=script_args.dataset_config)
+    print("loading: {0}".format(script_args.dataset_name))
+    
+    if os.path.exists(script_args.dataset_name) == True:
+        print("loading dataset from local file")
+        dataset = load_dataset("json", data_files=script_args.dataset_name)
+    else:
+        print("loading dataset from online")
+        dataset = load_dataset(script_args.dataset_name, name=script_args.dataset_config)
 
     ################
     # Training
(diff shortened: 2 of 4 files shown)


trl/scripts/sft.py
```diff
@@ -109,7 +110,14 @@ def main(script_args, training_args, model_args):
...
     ################
-    dataset = load_dataset(script_args.dataset_name, name=script_args.dataset_config)
+    print("loading: {0}".format(script_args.dataset_name))
+    
+    if os.path.exists(script_args.dataset_name) == True:
+        print("loading dataset from local file")
+        dataset = load_dataset("json", data_files=script_args.dataset_name)
+    else:
+        print("loading dataset from online")
+        dataset = load_dataset(script_args.dataset_name, name=script_args.dataset_config)
```
qgallouedec, line 120:
```suggestion
    if os.path.exists(script_args.dataset_name):
        dataset = load_dataset("json", data_files=script_args.dataset_name)
    else:
        dataset = load_dataset(script_args.dataset_name, name=script_args.dataset_config)
```

qgallouedec:
Cool! thanks! Can you just document it in `ScriptArguments`?

trl/scripts/sft.py
gogo2464: sure. good idea. sounds to be a much more better coding practice!

gogo2464:
obviously! I am going to document.

gogo2464:
Also I am very proud of your enthousiasm. I was going to split into another command line option. The point is to avoid a command line option with 2 use case in one. I could implement this point in another PR latter. I currently need to dev the backend.

gogo2464:
> Cool! thanks! Can you just document it in `ScriptArguments`?

in the class documentation or in the cli arguments?

gogo2464:
@qgallouedec I am confused. What should I put at this address https://github.com/huggingface/trl/blob/main/trl/scripts/utils.py#L41 ?

qgallouedec:
Using `dataset_name` as you did before seemed fine to me.

Maybe even better: an arg `data_files`, that way you keep the consistency with the argument name in `load_dataset`.

As for the doc, there's no need to add a line in the README. Just mention in `ScriptArguments` that either `dataset_name` or `data_files` must be filled in.

gogo2464:
I could use dataset_name only to minimize code size. But then I will choose to split to command lines arguments which will be recuperated into a single variable.

gogo2464:
I would like to test with a local dataset containing a set of images. I did: https://github.com/huggingface/trl/pull/3470/files#diff-068631df1c290e19041a844db840ce5d053269604aba96a730064e0d63a95325R117 and I tested with:

`trl sft --model_name_or_path "CompVis/stable-diffusion-v1-4" --dataset_name ./dataset/ --output_dir OUTPUT_SD_MODEL/`

It actually does not supports the current model. I have the error:

ValueError: Unrecognized model in CompVis/stable-diffusion-v1-4. Should have a model_type key in its config.json, or contain one of the following strings in its name: albert, align, altclip, aria, aria_text, audio-spectrogram-transformer, autoformer, aya_vision, bamba

Could I have a cli example with stable diffusion please? I am sure my python is correct.

trl/scripts/sft.py
gogo2464: sounds to be fixed. I put more content alas.

gogo2464:
does sft inly does text to text? https://github.com/huggingface/trl/blob/main/trl/scripts/sft.py#L90

README.md
```diff
@@ -163,6 +163,12 @@ trl sft --model_name_or_path Qwen/Qwen2.5-0.5B \
     --output_dir Qwen2.5-0.5B-SFT
 ```
 
+```bash
+trl sft --model_name_or_path Qwen/Qwen2.5-0.5B \
+    --dataset_name ./dataset.json \
+    --output_dir Qwen2.5-0.5B-SFT
+```
```
qgallouedec, line 170: Let's keep the readme simple

```suggestion
```

gogo2464: ok

gogo2464: Do you suggest another place to put the official doc in please? For your information, I did it in my blog at:

https://codeberg.org/libertia/modular-AI-doc#3-train-your-own-model-for-bypassing-privacy-copyright-issue-according-to-your-view

qgallouedec:
I've looked into it a bit more, having `"json"` hard-coded is not ideal. I think the best thing would be to simply add 

```diff
- dataset = load_dataset(script_args.dataset_name, name=script_args.dataset_config)
+ dataset = load_dataset(script_args.dataset_name, name=script_args.dataset_config, data_files=script_args.data_files)
```

and the user could use the script like this:

```
trl sft --model_name_or_path Qwen/Qwen2-0.5B --dataset_name arrow --data_files test_ds/test/data-00000-of-00001.arrow
```

gogo2464:
I entierely agree with `script_args.dataset_name` usage.

I am confused with `data_files`.

I was stating that now `--dataset_path` and `--dataset_name` are two different options usable in two different use cases. For example you can point to  a folder with `trl sft --model_name_or_path Qwen/Qwen2.5-0.5B --dataset_path ./dataset.json --output_dir Qwen2.5-0.5B-SFT`

instead of `trl sft --model_name_or_path Qwen/Qwen2.5-0.5B --dataset_name ./dataset.json --output_dir Qwen2.5-0.5B-SFT`

Then, data_files might be removed. I am confused with it usage.

gogo2464:
Is the cli argparse really in examples/research_projects/stack_llama/scripts/supervised_finetuning.py please? It sounds weird.

gogo2464:
--data_files sounds optional. I am tried a better third way.

gogo2464:
Finally, just do `trl sft --model_name_or_path Qwen/Qwen2-0.5B --dataset_path data.json --output_dir ./MYDIR`. Works on my computer. Ask for more testing if you would like. :)

qgallouedec:
I think there's been some confusion. Here, it's simply a matter of adding a new parameter, `data_files`. And once we've this argument, we can simply load, let's say `arrow` datasets using

```
--dataset_name arrow --data_files data-00000-of-00001.arrow
```

gogo2464:
Then data_files`--data_files` will replace `--data_path` ? I am not a great fan of this approach with a --data_path combined with a --data_files .

I think that --data_path sounds like a bash `cd` command wrapper and does not sounds a very unix philosophy. I would prefer to simply use --data_files only instead.


I am also confused. What is in the example the name data-00000-of-00001.arrow ? is it a file format of image? Of text? A folder, etc...

gogo2464:
@qgallouedec : sorry for wasting your time. Do you want I try more and then I show you? I do not necessarely need to ask.

gogo2464:
@qgallouedec Hello. Sorry for the confusion.

There has been a little delay. I am confused as well. Could you clarify please?

gogo2464:
is it ok to merge or should I fork?

gogo2464:
I wonder if you still want this pr or if you prefer that I fork to my own repo.

Best regards.

albertvillanova (commented):
To load a local dataset, you just need to pass its directory path name as `dataset_name`:
```python
dataset = load_dataset("directory_path_name")
```
The `datasets` library automatically detects the type of files you have in that directory (JSON, CSV, Parquet) to select the corresponding dataset builder.

So I would say this PR is not necessary to "Allow a user to train from a local dataset".

gogo2464:
> To load a local dataset, you just need to pass its directory path name as `dataset_name`:
> 
> ```python
> dataset = load_dataset("directory_path_name")
> ```
> 
> The `datasets` library automatically detects the type of files you have in that directory (JSON, CSV, Parquet) to select the corresponding dataset builder.
> 
> So I would say this PR is not necessary to "Allow a user to train from a local dataset".

Thank you for the information! I did not have read the doc YET.

albertvillanova:
Thanks for the PR and sorry for the very slow turnaround here @gogo2464 — this one slipped through our backlog. 

Since you opened it, `trl sft` reworked its dataset loading: `--dataset_name` now accepts a local path directly (its help text is literally "Path or name of the dataset to load"), and there's a `--datasets`/`data_files` mixture mechanism for local files, so this use case is already covered. As a result the branch also no longer applies — it patches the old single `load_dataset(...)` call that has since been replaced by a multi-branch `get_dataset(...)` block. 

Given that, plus the leftover debug prints and the `imagefolder` path in an SFT script, I think the cleanest move is to close this — but thank you for pushing on local-dataset support, and please open an issue if `--dataset_name <local_path>` doesn't cover your case!

PR #3495, opened 2025-05-26
[GRPO] Pad per minibatch instead of per generation batch

edbeeching:
# What does this PR do?

The current implementation of GRPO pads all samples by the longest sample in the generation batch, which can lead to significant redundant computation and greater memory pressure in a ds-z3 / FSDP setting.

Example of (unsorted vs sorted padded samples):
image

Plot of fraction of padded tokens in different configurations:
image

This PR stores samples unpadded, and pads per minibatch.

stiil TODO:

- [ ] add support for `mask_truncated_completions=True`
- [ ] benchmark
- [ ] validate performance / regressions


lewtun (commented):
Logic looks sound with a question about the way we slice the completion IDs tensor

trl/trainer/grpo_trainer.py
```diff
@@ -1302,14 +1353,42 @@ def compute_liger_loss(self, unwrapped_model, inputs):
...
+        if old_per_token_logps is not None:
+            old_per_token_logps = pad(inputs["unpadded_per_token_logps"], padding_value=0.0, padding_side="right")
+
+        padded_inputs = {
+            "prompt_ids": prompt_ids,
+            "prompt_mask": prompt_mask,
+            "completion_ids": completion_ids,
+            "completion_mask": completion_mask,
+            "advantages": torch.stack(inputs["advantages"]),
+            "old_per_token_logps": old_per_token_logps,
```
qgallouedec, line 1380: I think this should be in a collator

trl/trainer/grpo_trainer.py
```diff
@@ -1244,13 +1265,43 @@ def _generate_and_score_completions(
...
+                # prompt ids were left padded
+                unpadded_prompt_ids.append(p_ids[start:prompt_ids_length])
+                # completion ids were right padded
+                unpadded_completion_ids.append(c_ids[: end - prompt_ids_length])
+                if mask[start:end].sum() != end - start:
+                    raise ValueError(f"Attention mask from {start} to {end} does not match the expected length. ")
+                if old_per_token_logps is not None:
+                    unpadded_per_token_logps.append(old_logps[: end - prompt_ids_length])
+            else:
+                # case where the attention mask is all zeros, e.g. when mask_truncated_completions is enabled
```
qgallouedec, line 1295: Can you double-check this? In case of masked truncated completion, I think only the completion mask is truncated. You'd have something like:

```python
prompt_ids = [0, 0, 1, 2]
prompt_mask = [0, 0, 1, 1]
completion_ids = [3, 4, 5, 0]
completion_mask = [0, 0, 0, 0]
```

hence:

```python
attention_mask = [0, 0, 1, 1, 0, 0, 0, 0]
```

trl/trainer/grpo_trainer.py
```diff
@@ -1244,13 +1265,43 @@ def _generate_and_score_completions(
...
+                unpadded_prompt_ids.append(None)
+                unpadded_completion_ids.append(None)
+                unpadded_per_token_logps.append(None)
+
         return {
-            "prompt_ids": prompt_ids,
-            "prompt_mask": prompt_mask,
-            "completion_ids": completion_ids,
-            "completion_mask": completion_mask,
+            "unpadded_prompt_ids": unpadded_prompt_ids,
```
qgallouedec, line 1301: nit: I think it's ok to keep the original keys
```suggestion
            "prompt_ids": unpadded_prompt_ids,
```

trl/trainer/grpo_trainer.py
```diff
@@ -1302,14 +1353,42 @@ def compute_liger_loss(self, unwrapped_model, inputs):
...
+
+        if old_per_token_logps is not None:
+            old_per_token_logps = pad(inputs["unpadded_per_token_logps"], padding_value=0.0, padding_side="right")
+
+        padded_inputs = {
+            "prompt_ids": prompt_ids,
+            "prompt_mask": prompt_mask,
+            "completion_ids": completion_ids,
+            "completion_mask": completion_mask,
+            "advantages": torch.stack(inputs["advantages"]),
```
qgallouedec, line 1379: you don't modify the advantages, you don't need to stack, do you?

qgallouedec:
Looks good Ed! I took the liberty of adding a test for the new pad option.

qgallouedec:
Closing this one, but the idea is still good and I've captured it separately.

The problem you described is unchanged: `grpo_trainer.py` still pads `completion_ids` to the longest sample in the whole generation batch, then slices rows into per-device minibatches, so every minibatch carries padding sized by the longest sequence anywhere in the generation batch. `pad()` never gained `return_mask` either.

What's gone is the branch. `grpo_trainer.py` has taken 282 commits since your last push and grown from 1508 to 3497 lines, so it's conflicting and none of the 89 lines would apply. Faster to redo than to rebase, and the three TODOs (mask_truncated_completions, benchmark, regression check) were never done, so there's no measurement to carry over either.

PR #3499, opened 2025-05-27
Clean two sided clipping

qgallouedec:
Some late review on #3434


tests/test_grpo_trainer.py
qgallouedec, line 1: It might seem risky to remove these tests, but having tests that are so tightly coupled to internal implementation details (like calls to _get_log_prob) makes future development extremy difficult. Even if it's less secured, I recommend manually inspecting intermediate results instead, and keeping unit tests focused on what they're meant for: validating external behavior.

kashif:
i can try to integrate into liger too

PR #3510, opened 2025-05-29
🎀 New defaults: `gradient_checkpointing=True`

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/bco_config.py
```diff
@@ -73,6 +73,14 @@ class BCOConfig(TrainingArguments):
...
     """
 
+    # Arguments whose default values are overridden from TrainingArguments
+    gradient_checkpointing: bool = field(
+        default=True,
+        metadata={
+            "help": "If True, use gradient checkpointing to save memory at the expense of slower backward pass."
+        },
+    )
+
```
shirinyamani, line 83: So technically you wanna overwrite by setting the default to be `True` always? 
always doing the `grad_checkpointing`

I guess in most of the trainer scripts most of us set this to `True`, so I believe its a good time to change the default!

qgallouedec: Yes! that's the underlying motivation. We should do that with all the arguments that match:

> most of the trainer scripts most of us set this to <VALUE>

trl/trainer/cpo_config.py
```diff
@@ -23,14 +23,15 @@ class CPOConfig(TrainingArguments):
...
+    differ from those in [`~transformers.TrainingArguments`].
+
     Using [`~transformers.HfArgumentParser`] we can turn this class into
     [argparse](https://docs.python.org/3/library/argparse#module-argparse) arguments that can be specified on the
     command line.
 
     Parameters:
-        learning_rate (`float`, *optional*, defaults to `1e-6`):
-            Initial learning rate for [`AdamW`] optimizer. The default value replaces that of
-            [`~transformers.TrainingArguments`].
```
qgallouedec, line 33: Giveving the added

> Note that default values in this class may differ from those in [`~transformers.TrainingArguments`]."

there is no need to double-document

shirinyamani (commented):
I know the failing is not relevant to this PR directly, but have you checked the root?

qgallouedec:
No, not yet, I have to fix the tests before merging.

lewtun (approved):
Good default! LGTM

edbeeching (approved):
I feel like we are often overriding `transformers.TrainingArguments`, which requires updating all the child classes that inherit from this class. Would it make sense to have in intermediate class called `TrainingArgumentsOverride`?


For example:

```
class TrainingArgumentsOverride(transformers.TrainingArguments):
    gradient_checkpointing: bool = field(
        default=True,
        metadata={
            "help": "If True, use gradient checkpointing to save memory at the expense of slower backward pass."
        },
    )
```
Then:
```
class BCOConfig(TrainingArgumentsOverride):
...
```

This will centralize all the overrides to one file and will mean fewer changes.

Just an idea, no need to include in this PR.


Other comments:

- Should you update the doc string on the new classes to include the new default value of True in this case.
- What about `gradient_checkpointing_kwargs`, are you happy with the defaults?

qgallouedec:
> Would it make sense to have in intermediate class called TrainingArgumentsOverride?

I think it makes sense yes.

> Should you update the doc string on the new classes to include the new default value of True in this case.

I don't think it's necessary, for two reasons:

* The default value is correctly shown in the generated documentation (even if no one is likely to notice it there).
* It's also correctly picked up by the linter (here’s an example from mine — I think that’s where most people check default values anyway).
* I also find it more readable when the docstring only includes parameters specific to the method in `trl`. Including overridden parameters might make it a bit messy.

The only downside is that someone might check the Transformers docs and get confused by the outdated default value — but I don’t think that happens often.

> What about gradient_checkpointing_kwargs, are you happy with the defaults?

In some cases, I know that we need to disable or enable reentrant. But I don't quite remember when. Happy to override this one as well.

PR #3526, opened 2025-06-02
📦 Packing with flash attn kwargs to avoid cross-contamination

thepowerfuldeez:
Requires https://github.com/huggingface/transformers/pull/38536
Requires https://github.com/huggingface/trl/pull/3521

Modifies packing so that flash attention kernel is aware of sequence boundaries, which leads to improved sparsity and quality. Works only for SFT.

Loss plots:
image

(Blue is the new version, brown is the old one)

Achieved by this training config:
```python
training_args = SFTConfig(
    run_name=args.run_name,
    per_device_eval_batch_size=2,
    learning_rate=1e-5,
    weight_decay=1e-7,
    max_grad_norm=1.0,
    num_train_epochs=3,
    lr_scheduler_type="cosine",
    warmup_ratio=0.05,
    max_length=args.max_length,
    output_dir=args.output_dir,
    padding_free=True,
    eval_steps=0.05,
    logging_steps=0.01,
    save_steps=0.15,
    save_total_limit=3,
    save_strategy="steps",
    eval_strategy="steps",
    completion_only_loss=True,
    eos_token="<|im_end|>",
    include_tokens_per_second=True,
    use_liger_kernel=True,
    report_to=["wandb"],
    dataloader_num_workers=8,
    packing=True,
    eval_packing=False,
    packing_strategy="ffd",
)
```


qgallouedec:
Super cool! Can you change the base branch?

I think it doesn't requires transformers change if we play with signature columns, let check

thepowerfuldeez:
I missed the signature columns, so now it should work :) Updated plots, now the difference is visible.
I haven't found any data collators in trl, so I believe we would need to modify transformers in that case!

thepowerfuldeez:
@qgallouedec needs to re-open with main branch now

qgallouedec:
Wait, it wasn't automatically rebased? That's what usually happens.🤨

trl/data_utils.py
```diff
@@ -478,6 +479,67 @@ def pack_examples(examples: dict[str, list[list]], seq_length: int) -> dict[str,
     return examples
 
 
+@numba.njit(["(int32[:], int32)", "(int64[:], int64)"], cache=True)
```
qgallouedec: do you mind reverting this change? so that we keep PRs separate

qgallouedec: actually that might not be that easy, your 2 PRs are quite intertwined right?

thepowerfuldeez: yep, but I will think about reverting, sorry that I haven't done that before
this should be possible

thepowerfuldeez: done!

trl/data_utils.py
```diff
@@ -539,21 +539,29 @@ def _pack_ffd(examples: pa.Table, seq_length: int) -> pa.Table:
...
             new_offsets = np.array(new_offsets, dtype=dtype)
             packed_values = pa.array(packed_values, type=values.type)
+            sequence_lengths = pa.array(sequence_lengths, type=pa.list_(pa.int32()))
             column = type(column).from_arrays(new_offsets, packed_values)
-        packed_columns.append(column)
-    return pa.Table.from_arrays(packed_columns, names=examples.column_names)
+            packed_columns.append(column)
+            packed_columns.append(sequence_lengths)
+        else:
+            packed_columns.append(column)
```
qgallouedec: This should work as well?

```suggestion
            packed_columns.append(sequence_lengths)
        packed_columns.append(column)
```

qgallouedec: actually it seems like `sequence_lengths` can be appended more than once, if you've more than one col that matches `if pyarrow.types.is_list(column.type) or pyarrow.types.is_large_list(column.type):` (which is almost always the case)

trl/trainer/sft_trainer.py
```diff
@@ -678,6 +692,9 @@ def _set_signature_columns_if_needed(self):
         # dataset. So we need to override the default signature columns to include "completion_mask" as well.
         if self._signature_columns is None:
             self._signature_columns = ["input_ids", "attention_mask", "completion_mask"]
+        # For the packing case with FFD, we need to store sequence_length returned by the data collator with flattening
+        if self.args.packing and self.args.packing_strategy == "ffd" and self.args.padding_free:
+            self._signature_columns.append("sequence_length")
```
qgallouedec: nice!

trl/data_utils.py
thepowerfuldeez: ah yea makes sense, and I also think ordering matters right? we could just ensure that we added this once and in the end, I guess?

thepowerfuldeez: updated

qgallouedec:
It seems like the input of compute loss doesn't include `sequence_length`. Checking why

qgallouedec:
After digging, you'll need other columns like `cu_seq_lens_q`, `cu_seq_lens_k`, `max_length_q`, `max_length_k`. I think the easiest is to built the position ids the the packing func, then retrieve these flash-attn related values from it in the collator. I'm trying this approach in another branch. Keep me posted if you make any progress

thepowerfuldeez:
Using position_ids should be enough, as cu_seq_lens_q, cu_seq_lens_k, max_length_q, max_length_k gets created inside https://github.com/huggingface/transformers/blob/78d771c3c21922642fc9546ccb973cc7a182ab34/src/transformers/modeling_flash_attention_utils.py#L390-L393

I wil re-check

qgallouedec:
super cool! I didn't know about that

qgallouedec:
FYI, I'm currently working on resolving the conflict :)

qgallouedec:
I seems like it doesn't work, I'm not sure why exactly. It seems easier to directly build `position_ids` from packing I think, I'm trying this approach in another branch

trl/trainer/sft_trainer.py
```diff
@@ -293,7 +293,7 @@ def __init__(
         if args.padding_free:
             if data_collator is not None:
                 raise ValueError("Passing a custom data collator is not supported when using padding-free.")
-            if args.packing:
+            if args.packing and args.packing_strategy != "ffd":
```
```suggestion
            if args.packing:
```

actually padding_free is a different method (which is less relevant once once we have ffd with flash-attn).

trl/trainer/sft_trainer.py
```diff
@@ -313,7 +313,18 @@ def __init__(
...
+                    "You are using packing with padding-free training, but the attention implementation is not set to "
+                    "'flash_attention_2'. Packing flattens batches into a single sequence, and 'flash_attention_2' is "
+                    "the only known attention mechanism that reliably supports this. Using other implementations may "
+                    "lead to unexpected behavior. To ensure compatibility, set `attn_implementation='flash_attention_2'` "
+                    "in the model configuration."
+                )
+            data_collator = DataCollatorWithFlattening(
+                return_flash_attn_kwargs=False,
+                return_position_ids=True,
+            )
```
qgallouedec: I suggest dedenting this

trl/trainer/sft_trainer.py
thepowerfuldeez: You mean if we use packing with flash attn, we inherently already using padding free? If we pass position_ids
Probably we could change warning to say something like "when FFD packing is enabled, model accepts position_ids which works the same way as padding-free, so specifying padding_free makes no effect"

trl/trainer/sft_trainer.py
thepowerfuldeez: not sure what would happen if user do not specifies flash attention 2 implementation - right now we just won't use padding free with position ids.
If we de-dent - we enable DataCollatorWithFlattening for any cases?

thepowerfuldeez:
Ah, I see the problem. After the recent Attention refactoring, all models still accept position_ids as an argument.
And even if we pass position_ids to model.forward, it doesn't go to FA2 directly.

After the investigation, here's the attention flow

1. Inside the model class we set attention interface as
`attention_interface = ALL_ATTENTION_FUNCTIONS[self.config._attn_implementation]`
2. for FA2 it calls `flash_attention_forward` inside [transformers.integrations.flash_attention](https://github.com/huggingface/transformers/blob/fa921ad85458c193b5c304b3cab2164fd2b08a9d/src/transformers/integrations/flash_attention.py#L14)
3. `flash_attention_forward` do not accept position_ids directly, only through kwargs
4. flash_attention_forward calls `_flash_attention_forward` inside [transformers.modelling_flash_attention_utils](https://github.com/huggingface/transformers/blob/fa921ad85458c193b5c304b3cab2164fd2b08a9d/src/transformers/modeling_flash_attention_utils.py#L284)
5. If `position_ids` is provided here and attention_mask is not provided, we will generate padding-free masks: [link](https://github.com/huggingface/transformers/blob/fa921ad85458c193b5c304b3cab2164fd2b08a9d/src/transformers/modeling_flash_attention_utils.py#L385)

The problem, however, is that for some models attention_mask is always not None.
For example, for [Qwen3](https://github.com/huggingface/transformers/blob/fa921ad85458c193b5c304b3cab2164fd2b08a9d/src/transformers/models/qwen3/modeling_qwen3.py#L458), we pass attention_mask depending on attention type (full / sliding) and it's always casual attention mask.

I have a hack locally for going into flash_attn_varlen branch inside `flash_attention_forward` if we have position ids (so that position ids has priority over attention_mask).

thepowerfuldeez:
In addition, adding sequence length is incorrect after resolving conflict @qgallouedec 
I got 
```
ArrowInvalid: Column 1 named sequence_length expected length 436 but got length 874
```

Reverting helped

qgallouedec:
For reference:  https://github.com/thepowerfuldeez/trl/pull/1

thepowerfuldeez:
False alarm. I just tested training with aforementioned branch + transformers main, and we are calling flash_attention_varlen using position_ids. Nice and elegant solution by preparing position_ids during packing + flatten using existing data collator.

However, this approach uses more GPU memory, which is surprising

qgallouedec:
## Experiments

### Packing vs no packing

3.7 x speedup

Screenshot 2025-06-05 at 20 11 49
Screenshot 2025-06-05 at 20 11 35

```python

dataset = load_dataset("trl-lib/tldr", split="train[:2000]")

def concat(example):
    return{"text": example["prompt"]+ example["completion"]}
dataset = dataset.map(concat, batched=True, remove_columns=dataset.column_names)

packing = True
trainer = SFTTrainer(
    model="Qwen/Qwen2.5-0.5B",
    args=SFTConfig(
        model_init_kwargs={"attn_implementation": "flash_attention_2", "torch_dtype": "bfloat16"},
        packing=packing,
        run_name="packing" if packing else "no_packing",
        gradient_accumulation_steps=1 if packing else 6,  # to get close num_tokens/opt steps 
    ),
    train_dataset=dataset,
)
trainer.train()
```

### Flash attn kwargs vs no flash attn kwargs

The curves are very close, which seems normal to me.

The advantage here is that you avoid contamination between samples, which is not trivial to observe.

Screenshot 2025-06-05 at 20 17 42


For this one, I just tweaked the code to remove `position_ids`.

qgallouedec:
I'm realising that using position_ids when the attention implementation is not flash-attn hurts the result quite a lot. I'm adding a way use position_ids or not

qgallouedec:
Now it looks good! (purple is the annoying setting where position_ids are passed, even when flash-attn is disable: not the case anymore after 9f4d9ee3811a3e3e1a4251101732a927cfc573e8)

Screenshot 2025-06-05 at 20 48 06

qgallouedec:
Thanks a lot @thepowerfuldeez, I'll merge now, but feel free to make additional comments

qgallouedec (approved):

jiosephlee:
Hi @qgallouedec @thepowerfuldeez! 

I just had a clarification on how packing is implemented with max_length. If there is a datum that can fit in only partially, will that be forced in with truncation, or will that belong to the next sequence? I know the non-packing max_length behavior is truncation. If it doesn't truncate, then how does packing handle whole datums that are longer than the set max_length?

PR #3538, opened 2025-06-04
[SFT][Bugfix] sets average_tokens_across_devices to true in SFTConfig

edbeeching:
# What does this PR do?
Sets average_tokens_across_devices to true in SFTConfig

But defautl In a multi-device setting, the training loss is not compariable even you use the same effective batch size. (world_size * per_device_train_batch_size * gradient_accumulation_steps)

Example loss curves for 8,4,2,1 nodes (green, pink, grey, peach respectively).
image

This is because in `transformers.TrainingArguments` `average_tokens_across_devices=False` by default. Setting this to true achieves parity and lower training loss in different multi-device / node. Shown in the additional blue curve below. I tested 8 vs 64 GPUs (1 vs 8 nodes) and the training curves were near identical.
image

### DIFF
--- a/trl/trainer/sft_config.py
+++ b/trl/trainer/sft_config.py
@@ -80,7 +80,11 @@ class SFTConfig(TrainingArguments):
             the full sequence for [language modeling](#language-modeling) datasets.
         activation_offloading (`bool`, *optional*, defaults to `False`):
             Whether to offload the activations to the CPU.
-
+        average_tokens_across_devices (`bool`, *optional*, defaults to `True`):
+            Whether or not to average tokens across devices. If enabled, will use all_reduce to synchronize
+            num_tokens_in_batch for precise loss calculation. Reference:
+            https://github.com/huggingface/transformers/issues/34242
+            Overrides the `average_tokens_across_devices` parameter in [`~transformers.TrainingArguments`].
     """
 
     # Parameters that control the model
@@ -184,6 +188,14 @@ class SFTConfig(TrainingArguments):
         default=None,
         metadata={
             "help": "This parameter is deprecated and will be removed in version 0.20.0. Use `max_length` instead."
+        },
+    )
+    average_tokens_across_devices: bool = field(
+        default=True,
+        metadata={
+            "help": "Whether or not to average tokens across devices. If enabled, will use all_reduce to synchronize "
+            "num_tokens_in_batch for precise loss calculation. Reference: https://github.com/huggingface/transformers/issues/34242 "
+            "Overrides the `average_tokens_across_devices` parameter in [`~transformers.TrainingArguments`]."
         },
     )
 


trl/trainer/prm_config.py
```diff
@@ -46,6 +46,11 @@ class PRMConfig(TrainingArguments):
             Whether to train only on the last step.
         dataset_num_proc (`int`, *optional*, defaults to `None`):
             Number of processes to use for processing the dataset.
+        average_tokens_across_devices (`bool`, *optional*, defaults to `True`):
+            Whether or not to average tokens across devices. If enabled, will use all_reduce to synchronize
+            num_tokens_in_batch for precise loss calculation. Reference:
+            https://github.com/huggingface/transformers/issues/34242
+            Overrides the `average_tokens_across_devices` parameter in [`~transformers.TrainingArguments`].
```
qgallouedec: Starting from https://github.com/huggingface/trl/pull/3530, we stop overriding the docstring from `TrainingArgument`. It make things more concise, and easier to read. We just mention

> Note that default values in this class may differ from those in [`~transformers.TrainingArguments`].

The new default are displayed anyway in the TRL doc.

```suggestion
```

trl/trainer/prm_config.py
```diff
@@ -94,3 +99,11 @@ class PRMConfig(TrainingArguments):
         default=None,
         metadata={"help": "Number of processes to use for processing the dataset."},
     )
+    average_tokens_across_devices: bool = field(
+        default=True,
+        metadata={
+            "help": "Whether or not to average tokens across devices. If enabled, will use all_reduce to synchronize "
+            "num_tokens_in_batch for precise loss calculation. Reference: https://github.com/huggingface/transformers/issues/34242 "
+            "Overrides the `average_tokens_across_devices` parameter in [`~transformers.TrainingArguments`]."
```
qgallouedec:
```suggestion
```

trl/trainer/reward_config.py
```diff
@@ -45,6 +45,11 @@ class may differ from those in [`~transformers.TrainingArguments`].
         remove_unused_columns (`bool`, *optional*, defaults to `False`):
             Whether to remove the columns that are not used by the model's forward pass. Can be `True` only if
             the dataset is pretokenized.
+        average_tokens_across_devices (`bool`, *optional*, defaults to `True`):
+            Whether or not to average tokens across devices. If enabled, will use all_reduce to synchronize
+            num_tokens_in_batch for precise loss calculation. Reference:
+            https://github.com/huggingface/transformers/issues/34242
+            Overrides the `average_tokens_across_devices` parameter in [`~transformers.TrainingArguments`].
```
qgallouedec:
```suggestion
```

trl/trainer/reward_config.py
```diff
@@ -87,3 +92,11 @@ class may differ from those in [`~transformers.TrainingArguments`].
             "if the dataset is pretokenized."
         },
     )
+    average_tokens_across_devices: bool = field(
+        default=True,
+        metadata={
+            "help": "Whether or not to average tokens across devices. If enabled, will use all_reduce to synchronize "
+            "num_tokens_in_batch for precise loss calculation. Reference: https://github.com/huggingface/transformers/issues/34242 "
+            "Overrides the `average_tokens_across_devices` parameter in [`~transformers.TrainingArguments`]."
```
qgallouedec:
```suggestion
```

trl/trainer/sft_config.py
```diff
@@ -81,7 +81,11 @@ class SFTConfig(TrainingArguments):
             the full sequence for [language modeling](#language-modeling) datasets.
         activation_offloading (`bool`, *optional*, defaults to `False`):
             Whether to offload the activations to the CPU.
-
+        average_tokens_across_devices (`bool`, *optional*, defaults to `True`):
+            Whether or not to average tokens across devices. If enabled, will use all_reduce to synchronize
+            num_tokens_in_batch for precise loss calculation. Reference:
+            https://github.com/huggingface/transformers/issues/34242
+            Overrides the `average_tokens_across_devices` parameter in [`~transformers.TrainingArguments`].
```
qgallouedec:
```suggestion
```

trl/trainer/sft_config.py
```diff
@@ -202,6 +206,14 @@ class SFTConfig(TrainingArguments):
             "help": "This parameter is deprecated and will be removed in version 0.20.0. Use `max_length` instead."
         },
     )
+    average_tokens_across_devices: bool = field(
+        default=True,
+        metadata={
+            "help": "Whether or not to average tokens across devices. If enabled, will use all_reduce to synchronize "
+            "num_tokens_in_batch for precise loss calculation. Reference: https://github.com/huggingface/transformers/issues/34242 "
+            "Overrides the `average_tokens_across_devices` parameter in [`~transformers.TrainingArguments`]."
```
qgallouedec:
```suggestion
```

trl/trainer/prm_config.py
qgallouedec: Screenshot 2025-06-04 at 11 57 42

trl/trainer/prm_config.py
```diff
@@ -46,6 +46,11 @@ class PRMConfig(TrainingArguments):
             Whether to train only on the last step.
         dataset_num_proc (`int`, *optional*, defaults to `None`):
             Number of processes to use for processing the dataset.
+        average_tokens_across_devices (`bool`, *optional*, defaults to `True`):
```
shirinyamani: @edbeeching  @qgallouedec  I assume this average is for all the tokens before EOS on all devices? 

This would be particularly important when used in conjunction with`mask_truncated_completions=True`.

qgallouedec (approved):
LGTM!

trl/trainer/prm_config.py
qgallouedec: Actually it doesn't apply to GRPO: The issue only occurs when `num_items_in_batch` is used, which is not the case for GRPO (as we have our own compute loss function

shirinyamani: Correct, my bad for the `mask_truncated_completions=True` which is added to grpo. But the first part of the question i guess still applies, isn't? 
So in other words, I'm wondering when we are calculating this average, do we take this average wrt "all" tokens in the batch aware of the facts that the tokens after EOS should not be considered?

qgallouedec: > I assume this average is for all the tokens before EOS on all devices?

Yes, that's correct

PR #3543, opened 2025-06-05
💽 [TRLParser] Fail when unknown args are provided in the config file.

edbeeching:
When a config.yaml file is parsed by the `TRLParser` there is an option to return remaining strings. But otherwise we do not check that all of the parsed args have been consumed by the config classes. 

It would be easy for a user to mistype an argument in a config file and not be aware that the parameter was not set correctly.

This PR adds an option to the `TRLParser` `fail_with_unknown_args`, which is `True` by default. This ensures that errors in the config file are quickly flagged to the user.

### DIFF
--- a/trl/scripts/utils.py
+++ b/trl/scripts/utils.py
@@ -214,13 +214,12 @@ class TrlParser(HfArgumentParser):
         if return_remaining_strings:
             args_remaining_strings = output[-1]
             return output[:-1] + (config_remaining_strings + args_remaining_strings,)
+        elif fail_with_unknown_args and config_remaining_strings:
+            raise ValueError(
+                f"Unknown arguments from config file: {config_remaining_strings}. Please remove them, add them to the "
+                "dataclass, or set `fail_with_unknown_args=False`."
+            )
         else:
-            if fail_with_unknown_args:
-                if config_remaining_strings:
-                    raise ValueError(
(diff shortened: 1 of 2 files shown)


trl/scripts/utils.py
```diff
@@ -212,6 +215,12 @@ def parse_args_and_config(
             args_remaining_strings = output[-1]
             return output[:-1] + (config_remaining_strings + args_remaining_strings,)
         else:
+            if fail_with_unknown_args:
+                if config_remaining_strings:
+                    raise ValueError(
+                        f"Unknown arguments from config file: {config_remaining_strings}. "
+                        "Please remove them or add them to the dataclass."
```
qgallouedec: ... or set`fail_with_unknown_args` to False

trl/scripts/utils.py
```diff
@@ -212,6 +215,12 @@ def parse_args_and_config(
             args_remaining_strings = output[-1]
             return output[:-1] + (config_remaining_strings + args_remaining_strings,)
         else:
+            if fail_with_unknown_args:
+                if config_remaining_strings:
```
qgallouedec:
```suggestion
        elif config_remaining_string and fail_with_unknown_args:
```

qgallouedec (commented):
Nice, some minor suggestions

qgallouedec (approved):
lgtm!

PR #3390, opened 2025-04-29
Add completion entropy metrics

lewtun:
# What does this PR do?

Metric from the DAPO paper: https://dapo-sia.github.io

Useful for measuring diversity / entropy collapse

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #3563, opened 2025-06-10
Add entropy based filtering inside the GRPOTrainer.

### LINKED ISSUE #3555: Add entropy based token filtering to the GRPO Trainer/Loss Function.
### Feature request

In the [paper](https://arxiv.org/pdf/2506.01939) **Beyond the 80/20 Rule: High-Entropy Minority Tokens Drive Effective Reinforcement Learning for LLM Reasoning** the authors from the Qwen team show that filtering out i.e. zeroing out the policy-loss from the bottom 80% of tokens in a model's response based on their entropy can yield better results than including the loss from all the tokens.

> We ultimately improve RLVR by restricting policy gradient updates to
forking tokens and uncover a finding even beyond the 80/20 rule: utilizing only 20% of
the tokens while maintaining performance comparable to full-gradient updates on the
Qwen3-8B base model and significantly surpassing full-gradient updates on the Qwen3-
32B

It'd be great to add a flag in the GRPO Trainer so that a user can exclude the bottom-k tokens based on entropy from the policy loss.

### Motivation

Can help users create better reasoning models!

### Your contribution

I'd like to implement this parameter/feature!

pramodith:
# What does this PR do?

This PR is in relation to #3555 which proposes to mask out the policy loss coming from tokens in the completions corresponding to positions with an entropy scores below the bottom-k percentile.

This idea is proposed by the Qwen team in their accompanying [paper](https://arxiv.org/pdf/2506.01939) **Beyond the 80/20 Rule**

## Key Proposals of the paper that guided the implementation

```math
\mathcal{J}^{\mathcal{B}}_{\text{HighEnt}}(\theta) = 
\mathbb{E}_{\mathcal{B} \sim \mathcal{D},\,(q,a) \sim \mathcal{B},\, \{o^i\}_{i=1}^{G} \sim \pi_{\text{old}}(\cdot|q)}
\left[
\frac{1}{\sum_{i=1}^{G} |o^i|} \sum_{i=1}^{G} \sum_{t=1}^{|o^i|} 
\mathbb{I}\left[H^i_t \geq \tau^{\mathcal{B}}_{\rho} \right] 
\cdot \min\left(r^i_t(\theta) \hat{A}^i_t,\ 
\text{clip}(r^i_t(\theta), 1 - \epsilon_{\text{low}}, 1 + \epsilon_{\text{high}}) \hat{A}^i_t
\right)
\right],
\quad \text{s.t. } 0 < \left|\left\{ o^i \mid \text{is\_equivalent}(a, o^i) \right\}\right| < G
```

The key difference is the term
```math
\mathbb{I}\left[H^i_t \geq \tau^{\mathcal{B}}_{\rho} \right] 
```
From the paper

> Here, $H^i_t$ denotes the entropy of token t in response i, $\mathbb{I}[·]$ is the indicator function that evaluates to 1 if the
condition inside holds and 0 otherwise, ρ ∈ (0, 1] is a predefined ratio specifying the top proportion of high-entropy tokens to be selected within a batch, and $\tau^{\mathcal{B}}$ is the corresponding entropy threshold within
the batch B such that only tokens with $H^i_t ≥ \tau^{\mathcal{B}}_{\rho}$ , **comprising the top-ρ fraction of all tokens in the batch**, are used to compute the gradient.

Entropy is calculated as normal via the formula

```math
H_t := - \sum_{j=1}^{V} p_{t,j} \log p_{t,j}
```

The paper applies the entropy mask to the DAPO loss function in their experiments, but I think we can leave it to the user to decide which loss i.e. GRPO, Dr. GRPO or DAPO to apply it to.

The paper finds that the best threshold is to keep the top-20% of tokens based on their entropy.

> We ultimately improve RLVR by restricting policy gradient updates to
forking tokens and uncover a finding even beyond the 80/20 rule: utilizing only 20% of
the tokens while maintaining performance comparable to full-gradient updates on the
Qwen3-8B base model and significantly surpassing full-gradient updates on the Qwen3-
32B (+11.04 on AIME’25 and +7.71 on AIME’24) and Qwen3-14B (+4.79 on AIME’25 and
+5.21 on AIME’24)

**I didn't run the vllm tests inside `test_grpo_trainer.py` since my machine/vm didn't have access to a gpu.**

Fixes #3555 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/scripts/grpo.py
+++ b/trl/scripts/grpo.py
@@ -100,7 +100,7 @@ def main(script_args, training_args, model_args):
     # Initialize the GRPO trainer
     trainer = GRPOTrainer(
         model=model,
-        reward_funcs=reward_model,
+        reward_funcs=reward_funcs,
         args=training_args,
         train_dataset=dataset[script_args.dataset_train_split],
         eval_dataset=dataset[script_args.dataset_test_split] if training_args.eval_strategy != "no" else None,


LeonEricsson:
You should be able to calculate the entropy directly from log probs as

```
entropy = -(per_token_logps.exp() * per_token_logps).sum(-1)
```

which means we don't have to modify `_get_per_token_logps`. I noted however that you didn't apply temperature scaling when calculating your entropy, which my proposed solution does. As I see it temperature scaling remains an unresolved variable: the paper evaluates entropy masking only at T = 1, so the optimal percentile thresholds for other temperatures are unknown. The same caveat applies to model choice—results are reported solely for the Qwen-3 series, and the threshold may shift substantially for other architectures. In practice, the 20 % rule should be treated as a starting heuristic; users will need to experiment with the temperature-model-percentile combination that best fits their own setup.

pramodith:
~Ah,  yeah that's a good point let me update the entropy calculation to use the logprobs directly. Yep the temperature and 20% threshold should be hyper-parameters that are tuned by the user.~

pramodith:
Also just realized that the temperature parameter won't affect the ranking order of entropy since all positions are affected by the same temperature, so the temp param shouldn't matter here.

trl/trainer/grpo_trainer.py
LeonEricsson, line 1: Given that liger loss doesn't support entropy filtering, raise an error if the user tries to use liger loss + entropy filter

pramodith:
> You should be able to calculate the entropy directly from log probs as
> 
> ```
> entropy = -(per_token_logps.exp() * per_token_logps).sum(-1)
> ```
Actually this isn't true the `per_token_logps` only contain the log probs of the token in the sampled response. Entropy needs to be calculated over the probability distribution over the entire vocabulary per seq position in the response, so we do need to update the `_get_per_token_logps` since it uses the `selective_log_softmax` function to gather the probs of just the tokens in sampled response.

LeonEricsson:
> > You should be able to calculate the entropy directly from log probs as
> > ```
> > entropy = -(per_token_logps.exp() * per_token_logps).sum(-1)
> > ```
> 
> Actually this isn't true the `per_token_logps` only contain the log probs of the token in the sampled response. Entropy needs to be calculated over the probability distribution over the entire vocabulary per seq position in the response, so we do need to update the `_get_per_token_logps` since it uses the `selective_log_softmax` function to gather the probs of just the tokens in sampled response.

yes of course, my bad. not really a fan of the proposed refactor of  `_get_per_token_logps` so trying to think of an alternative solution. Things of concern: 1) `logits` is a big tensor, clogs VRAM. 2) performing `torch.softmax(logits, dim=-1)` will spike memory usage. See #3390 for discussion. But maybe we just have to bite the bullet here, and in this case we should definitely only perform the entropy calculation when necessary

qgallouedec:
Nice! Thanks!
Yes, as @LeonEricsson says, the tricky part is the memory peak of the softmax. There as to be a way to compute the entropy efficiently, similarly to what we do with `selective_log_softmax`.

Another recommendation: 1 argument is probably enough.

```diff
- if self.filter_on_entropy:
+ if self.token_entropy_percentile_threshold < 1.0:
```

and add that the recommended value is 0.2 in the documentation.

pramodith:
Cool, let me look into making the entropy calculation less memory intensive.

pramodith:
Updated the code to make sure that only a mini-batch of logits are materialized at any given point of time and entropies for those mini-batches of logits are optionally calculated. The `rowise_entropy` function is inspired by the `selective_log_softmax` function to calculate entropy per row of the mini-batch. Introduces a bit of redundant compute because we compute the log_softmax twice in entropy calc and `selective_log_softmax`, but I guess its worth the tradeoff with the gains in clean code and memory.

trl/trainer/grpo_trainer.py
```diff
@@ -1360,6 +1392,9 @@ def _compute_loss(self, model, inputs):
         per_token_loss1 = coef_1 * advantages.unsqueeze(1)
         per_token_loss2 = coef_2 * advantages.unsqueeze(1)
         per_token_loss = -torch.min(per_token_loss1, per_token_loss2)
+        if entropy_mask is not None:
+            # mask out tokens based on entropy
```
```suggestion
```

already clear and self-explanatory

trl/trainer/grpo_trainer.py
```diff
@@ -1324,20 +1338,38 @@ def _compute_loss(self, model, inputs):
...
-        per_token_logps = self._get_per_token_logps(model, input_ids, attention_mask, logits_to_keep)
+        # Compute the entropy at each position in the completion
+        if self.token_entropy_percentile_threshold > 0.0:
+            logs_and_entropies = self._get_per_token_logps_and_entropies(
+                model, input_ids, attention_mask, logits_to_keep, do_compute_entropy=True
+            )
+            per_token_logps = logs_and_entropies["logps"]
+            entropies = logs_and_entropies["entropies"]
+            # we need to find the entropy threshold based on all tokens in the
+            # batch and not just along a given batch index.
```
```suggestion
            # compute the entropy threshold across all tokens in the batch
```

trl/trainer/grpo_trainer.py
```diff
@@ -1324,20 +1338,38 @@ def _compute_loss(self, model, inputs):
...
+            entropy_threshold = torch.quantile(entropies.flatten(), self.token_entropy_percentile_threshold)
+            # Create a mask for tokens corresponding to positions in the completion
+            # where the entropy is not in the top `token_entropy_percentile_threshold` percentile.
```
```suggestion
```
think this is redundant as well

LeonEricsson:
> Updated the code to make sure that only a mini-batch of logits are materialized at any given point of time and entropies for those mini-batches of logits are optionally calculated. The `rowise_entropy` function is inspired by the `selective_log_softmax` function to calculate entropy per row of the mini-batch. Introduces a bit of redundant compute because we compute the log_softmax twice in entropy calc and `selective_log_softmax`, but I guess its worth the tradeoff with the gains in clean code and memory.

nice work. left a few minor comments

pramodith:
Thanks for the review Leon! Made the suggested changes.

pramodith:
@qgallouedec please take another look at the PR when you have the time.

LeonEricsson:
I spent some time benchmarking different entropy calculations, [scripts here](https://github.com/LeonEricsson/trl/blob/memory_efficient_entropy_calculation/entropy_benches.py). Long story short I recommend:

```
def compute_entropy(logits: torch.Tensor, chunk_size: int = 1) -> torch.Tensor:
    """
    Compute the Shannon entropy (nats) for each row of `logits`
    with a memory-friendly, `log_softmax → exp → reduce` sequence.

    Args:
        logits (`torch.Tensor`):
            Tensor of shape `(..., num_classes)` holding unnormalised
            log-probabilities.
        chunk_size (`int`, *optional*, defaults to `1`):
            Number of rows processed together. Keeping it
            at `1` minimises peak memory; larger values trade a small
            rise in memory for fewer launches.  Default is `1`.

    Returns:
        `torch.Tensor`:
            Entropy values with shape `logits.shape[:-1]`.
    """
    outs = []
    for chunk in logits.split(chunk_size, dim=0): # loop to reduce peak mem consumption
        logp = F.log_softmax(chunk, dim=-1)
        entropy = -(logp.exp() * logp).sum(-1)
        outs.append(entropy)
    return torch.cat(outs, dim=0)
```

this is basically what you already had, but importantly we sum inside the loop which avoids materializing the [S, V] tensor. I observed no latency improvements when increasing `chunk_size` on a 4090, because it's already memory-bound at `1`

1485840691:
@qgallouedec @LeonEricsson  this PR has a bit of duplicate with my implementation of entropy regularization loss in https://github.com/huggingface/trl/pull/3628. We need to sync on both given the 2 address entropy in two different directions. The entropy regularization loss is proposed in [issue ](https://capricious-hydrogen-41c.notion.site/Skywork-Open-Reasoner-Series-1d0bc9ae823a80459b46c149e4f51680?pvs=25#1d1bc9ae823a801592a0c3891ea5328f) and also officially implemented in verl.

pramodith:
@1485840691 I think that despite the overlap both of the purposes are complementary and once this PR is pushed including the entropy loss in the final loss should be fairly simple.

pramodith:
@LeonEricsson thanks for those cool benchmarks! I've updated the code to compute the sum inside the for loop as you've suggested.

1485840691:
I have a comment on entropy from logits. Given verl has already provided an implementation https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/utils/torch_functional.py#L150, why do we not re-use that?

Sorry, I tried benchmark it. Indeed verl's implementation does not have better running performance
"
Duration of rowise_entropy: 0.7136 seconds
Duration of entropy_from_logits: 1.4799 seconds
"

But given the entropy from logits function is embed inside the chunk loop of get_per_token_logps, why do we need to support chunking here?

1485840691:
And there is another question regarding the interaction between entropy loss and entropy mask: Do we need to consider the entropy mask in computing entropy loss? Now entropy loss is computed using completion mask @qgallouedec @LeonEricsson

pramodith:
> I have a comment on entropy from logits. Given verl has already provided an implementation https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/utils/torch_functional.py#L150, why do we not re-use that? Given the entropy from logits function is embed inside the chunk loop of get_per_token_logps, I think we might not need to support chunking here. Even if we do need chunking, verl has in-place implementation: https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/utils/torch_functional.py#L150

I'm assuming you mean why we don't simply import the function from `veRL`. I think that we'd like to avoid introducing new dependencies for fairly simple utility functions. Good point on supporting chunking inside the entropy function, while you correctly point out that we don't need to support chunking for the current implementation of GRPO since the `get_per_token_logps` does it, I think it makes sense to support chunking inside the `rowise_entropy` function to make it more re-usable for other usecases.

pramodith:
> And there is another question regarding the interaction between entropy loss and entropy mask: Do we need to consider the entropy mask in computing entropy loss? Now entropy loss is computed using completion mask @qgallouedec @LeonEricsson

I don't think the entropy loss should take the entropy mask into consideration. In the entropy masking paper linked above, their intention was to just affect the policy loss. So similar to how KL-div doesn't consider the entropy mask, I believe that the entropy loss shouldn't either, especially if we want to reproduce the behavior of veRL.

1485840691:
> > I have a comment on entropy from logits. Given verl has already provided an implementation https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/utils/torch_functional.py#L150, why do we not re-use that? Given the entropy from logits function is embed inside the chunk loop of get_per_token_logps, I think we might not need to support chunking here. Even if we do need chunking, verl has in-place implementation: https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/utils/torch_functional.py#L150
> 
> I'm assuming you mean why we don't simply import the function from `veRL`. I think that we'd like to avoid introducing new dependencies for fairly simple utility functions. Good point on supporting chunking inside the entropy function, while you correctly point out that we don't need to support chunking for the current implementation of GRPO since the `get_per_token_logps` does it, I think it makes sense to support chunking inside the `rowise_entropy` function to make it more re-usable for other usecases.

I think we might follow verl to support entropy_from_logits(https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/utils/torch_functional.py#L143) and entropy_from_logits_chunking(https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/utils/torch_functional.py#L150). And I do not mean importing verl lib to support such small util function ,just copy the code and add a comment. But given I have benchmarked it. verl's implementation does not have better performance.

pramodith:
> I have a comment on entropy from logits. Given verl has already provided an implementation https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/utils/torch_functional.py#L150, why do we not re-use that?
> 
> Sorry, I tried benchmark it. Indeed verl's implementation does not have better running performance " Duration of rowise_entropy: 0.7136 seconds Duration of entropy_from_logits: 1.4799 seconds "
> 
> But given the entropy from logits function is embed inside the chunk loop of get_per_token_logps, why do we need to support chunking here?

Sorry, I missed this comment. I think the objective here is to reduce peak memory usage rather than optimize on run-time, based on the comments from Leon and Quentin

> Nice! Thanks!
Yes, as @LeonEricsson says, the tricky part is the memory peak of the softmax. There as to be a way to compute the entropy efficiently, similarly to what we do with selective_log_softmax.

LeonEricsson:
> But given the entropy from logits function is embed inside the chunk loop of get_per_token_logps, why do we need to support chunking here?

I agree that chunking isn't necessary here, but since it's defined as a general utility function and could be useful in future scenarios, its nice to have; provides a convenient way to trade memory for latency.

> I think we might follow verl to support entropy_from_logits(https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/utils/torch_functional.py#L143) and entropy_from_logits_chunking(https://github.com/volcengine/verl/blob/9b7bb69ea3165b691cc908d7f3f2f14c4a65a59e/verl/utils/torch_functional.py#L150).

Yeah, `entropy_from_logits_chunking` looks very similar to what I benchmarked [here](https://github.com/LeonEricsson/trl/blob/7231f7babba274df5424b1f9ae2507cc82eaf679/entropy_benches.py#L62-L70). I’d be fine with it as well, but based on my isolated benchmarks, I found it has the same memory usage while being slower.

> I don't think the entropy loss should take the entropy mask into consideration. In the entropy masking paper linked above, their intention was to just affect the policy loss. So similar to how KL-div doesn't consider the entropy mask, I believe that the entropy loss shouldn't either, especially if we want to reproduce the behavior of veRL.

~looking into this~
I agree with @pramodith, entropy filtering is applied to the policy loss, and the entropy loss should be entirely separate from the policy loss.

trl/trainer/utils.py
```diff
@@ -1781,6 +1781,32 @@ def selective_log_softmax(logits, index) -> torch.Tensor:
...
+        `torch.Tensor`:
+            Entropy of each position in a sequence
+    """
+    per_token_entropies = []
+    for i in range(0, logits.shape[0], chunk_size):
+        logits_chunk = logits[i : i + chunk_size]
+        for row_logits in logits_chunk:  # loop to reduce peak mem consumption
+            row_logps = F.log_softmax(row_logits, dim=-1)
+            row_entropy = -(torch.exp(row_logps) * row_logps).sum(-1)
+            per_token_entropies.append(row_entropy)
```
LeonEricsson: doesn't the inner loop negate the purpose of `chunk_size`? we can achieve the same thing by:
```suggestion
    for logits_chunk in logits.split(chunk_size, dim=0):
        logps = F.log_softmax(logits_chunk, dim=-1)    
        chunk_entropy = -(torch.exp(logps) * logps).sum(-1)
        per_token_entropies.append(chunk_entropy)
```

with a default `chunk_size=1`.

trl/trainer/utils.py
```diff
@@ -1781,6 +1781,30 @@ def selective_log_softmax(logits, index) -> torch.Tensor:
     return per_token_logps
 
 
+def rowise_entropy(logits, chunk_size: int = 1) -> torch.Tensor:
```
LeonEricsson:
```suggestion
def entropy_from_logits(logits, chunk_size: int = 1) -> torch.Tensor:
```

make sure to change test names as well

trl/trainer/utils.py
```diff
@@ -1781,6 +1781,30 @@ def selective_log_softmax(logits, index) -> torch.Tensor:
...
+
+    Args:
+        logits (`torch.Tensor`):
+            Logits tensor of shape `(..., num_classes)`.
+        chunk_size (`int`, *optional*, defaults to `2048`):
+            The number of chunks the batch will be split into to reduce memory consumption.
+    Returns:
+        `torch.Tensor`:
+            Entropy of each position in a sequence
+    """
```
LeonEricsson:
```suggestion
    """
    Compute the Shannon entropy (in nats) for each row of *logits* without
    materialising the full soft-max in memory.

    The batch dimension is processed in chunks of size `chunk_size` so that
    only a subset of rows is expanded to probabilities at any one time.

    Args:
        logits (`torch.Tensor`):
            Logits tensor of shape `(..., num_classes)`. Entropy is taken along the last axis; all
            leading dimensions are preserved.
        chunk_size (`int`, *optional*, defaults to `1`):
            Number of rows to process per iteration.
    Returns:
        `torch.Tensor`:
            Entropy values with shape `logits.shape[:-1]`.
    """
```

tests/test_utils.py
```diff
@@ -616,3 +617,23 @@ def test_num_samples(self, mock_stdout):
...
+    def test_rowise_entropy(self, dtype, chunk_size):
+        batch_size, seq_len, vocab_size = 64, 384, 768
+        logits = torch.randn(batch_size, seq_len, vocab_size, dtype=dtype)
+        log_probs = torch.log_softmax(logits, dim=-1)
+        expected_entropy = -(torch.exp(log_probs) * log_probs).sum(-1)
+        predicted_entropy = rowise_entropy(logits, chunk_size=chunk_size)
+        if dtype in [torch.float16, torch.bfloat16]:
+            self.assertTrue(torch.equal(predicted_entropy, expected_entropy))
+        else:
+            torch.testing.assert_close(predicted_entropy, expected_entropy, rtol=1e-5, atol=1e-5)
```
LeonEricsson:
```suggestion
        if dtype in (torch.float64, torch.float32):
            p = logits.softmax(-1)
            entropy = -torch.sum(p * p.log(), dim=-1)
        else:
            logps = logits.log_softmax(dim=-1)
            entropy = -(torch.exp(logps) * logps).sum(-1)

        predicted_entropy = entropy_from_logits(logits, chunk_size=chunk_size)
        torch.testing.assert_close(predicted_entropy, entropy, rtol=1e-5, atol=1e-5)
```

we can use the exact calculation for float32/64

trl/trainer/grpo_trainer.py
```diff
@@ -1364,7 +1383,22 @@ def _compute_loss(self, model, inputs):
...
         logits_to_keep = completion_ids.size(1)  # we only need to compute the logits for the completion tokens
 
-        per_token_logps = self._get_per_token_logps(model, input_ids, attention_mask, logits_to_keep)
+        # Compute the entropy at each position in the completion
+        if self.token_entropy_percentile_threshold > 0.0:
+            logs_and_entropies = self._get_per_token_logps_and_entropies(
+                model, input_ids, attention_mask, logits_to_keep, do_compute_entropy=True
+            )
+            per_token_logps = logs_and_entropies["logps"]
+            entropies = logs_and_entropies["entropies"]
```
```suggestion
            logps_and_entropies = self._get_per_token_logps_and_entropies(
                model, input_ids, attention_mask, logits_to_keep, do_compute_entropy=True
            )
            per_token_logps = logps_and_entropies["logps"]
            entropies = logps_and_entropies["entropies"]
```

LeonEricsson:
some final comments then I'm happy to merge.

LeonEricsson (approved):

PR #3598, opened 2025-06-16
⚔️ Fix bf16 fp16 config conflict issue

yao-matrix:
w/ this PR https://github.com/huggingface/trl/pull/3515, it by default set `bf16` set to True regardless what `fp16` setting is. In case where user set `fp16` explicitly to `True`(as in ut case here https://github.com/huggingface/trl/blob/main/tests/slow/test_dpo_slow.py#L132), there will be error message "At most one of fp16 and bf16 can be True, but not both".

Fixing this issue by setting `bf16` to True only when `fp16` is False and `bf16` is not set.

@qgallouedec , pls help review, thx.

### DIFF
--- a/trl/trainer/bco_config.py
+++ b/trl/trainer/bco_config.py
@@ -199,3 +199,8 @@ class BCOConfig(TrainingArguments):
         default=10.0,
         metadata={"help": "Maximum value of the density ratio. The estimated density ratio is clamped to this value."},
     )
+
+    def __post_init__(self):
+        self.bf16 = not (self.fp16) if self.bf16 is None else self.bf16
+
+        super().__post_init__()
--- a/trl/trainer/cpo_config.py
+++ b/trl/trainer/cpo_config.py
@@ -187,7 +187,6 @@ class CPOConfig(TrainingArguments):
     )
 
     def __post_init__(self):
-        if self.bf16 is None:
-            self.bf16 = True if not self.fp16 else False
+        self.bf16 = not (self.fp16) if self.bf16 is None else self.bf16
 
         super().__post_init__()
--- a/trl/trainer/dpo_config.py
+++ b/trl/trainer/dpo_config.py
@@ -438,7 +438,6 @@ class DPOConfig(TrainingArguments):
     )
 
     def __post_init__(self):
-        if self.bf16 is None:
-            self.bf16 = True if not self.fp16 else False
+        self.bf16 = not (self.fp16) if self.bf16 is None else self.bf16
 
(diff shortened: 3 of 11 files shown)


trl/trainer/bco_config.py
```diff
@@ -202,3 +202,9 @@ class BCOConfig(TrainingArguments):
         default=10.0,
         metadata={"help": "Maximum value of the density ratio. The estimated density ratio is clamped to this value."},
     )
+
+    def __post_init__(self):
+        if self.bf16 is None:
+            self.bf16 = True if not self.fp16 else False
+
```
yao-matrix, line 205: when `bf16` is set, directly pass to `TrainingArguments` `post_init_` to do argument validation and setting.
when `bf16` is not set, if `fp16` is `True`, set `bf16` to `False`, else to `True`

qgallouedec:
Yes thank you, this PR makes sense. By the way, that's why slow tests fail on main.

The only thing that bothers me is that None by default isn't very informative, maybe even misleading. For the moment, I don't really see an alternative.

yao-matrix:
> Yes thank you, this PR makes sense. By the way, that's why slow tests fail on main.
> 
> The only thing that bothers me is that None by default isn't very informative, maybe even misleading. For the moment, I don't really see an alternative.

@qgallouedec Yes, since it's a `post_init_`, and we need to know whether the `bf16` is set explicitly by users or just a automatic default, so the best way i can figure to set it as optional w/ default None. Yes, it's not informative, maybe we can enhance docstring to improve that?

Like    "Whether to use bf16 (mixed) precision instead of 32-bit. Requires Ampere or higher NVIDIA "
          "architecture or Intel XPU or using CPU (use_cpu) or Ascend NPU. **If user doesn't set it explicitly and `fp16` not set to True, we default set `bf16` to True.** This is an experimental API and it may change."

WDYT?

qgallouedec:
> Yes, it's not informative, maybe we can enhance docstring to improve that?

Yes, I think that's the best way to go at this stage. I'm fine with what you suggested, just delete this

> This is an experimental API and it may change.

trl/trainer/bco_config.py
qgallouedec: I think it's equivalent to this?

```python
self.bf16 = not(self.fp16) if self.bf16 is None else self.bf16
```

qgallouedec:
I just realized that I asked you to remove something that was already in the doc. But still, can you remove it? I think bf16 is stable enough now

PR #3599, opened 2025-06-16
📜 Add `chat_template_path` parameter to `SFTConfig`

qgallouedec:
Take the chat template from `"Qwen/Qwen3-4B"` to train a tiny GPTNeoX model:

```python

dataset = load_dataset("trl-internal-testing/zen", "conversational_language_modeling", split="train")
training_args = SFTConfig(chat_template_path="Qwen/Qwen3-4B")
trainer = SFTTrainer(
    model="trl-internal-testing/tiny-GPTNeoXForCausalLM",  # trl-internal-testing/tiny-GPTNeoXForCausalLM doesn't have a chat template set by default
    args=training_args,
    train_dataset=dataset
)
trainer.train()
```

or from a path:

```python
training_args = SFTConfig(chat_template_path="path_to_template.jinja")
```

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1350,3 +1350,29 @@ class SFTTrainerTester2(unittest.TestCase):
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.allclose(param, new_param), f"Parameter {n} has not changed")
+
+    def test_train_with_set_chat_template(self):
+        # Get the dataset
+        dataset = load_dataset("trl-internal-testing/zen", "conversational_language_modeling", split="train")
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            # Initialize the trainer
+            training_args = SFTConfig(output_dir=tmp_dir, chat_template_source="Qwen/Qwen3-4B", report_to="none")
+            # trl-internal-testing/tiny-GPTNeoXForCausalLM doesn't have a chat template set by default
+            trainer = SFTTrainer(
+                model="trl-internal-testing/tiny-GPTNeoXForCausalLM", args=training_args, train_dataset=dataset
+            )
+
+            # Save the initial parameters to compare them later
+            previous_trainable_params = {n: param.clone() for n, param in trainer.model.named_parameters()}
+
+            # Train the model
+            trainer.train()
+
+            # Check that the training loss is not None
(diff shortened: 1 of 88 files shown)


tests/test_sft_trainer.py
```diff
@@ -1350,3 +1350,29 @@ def test_train_packing(self, packing_strategy):
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.allclose(param, new_param), f"Parameter {n} has not changed")
+
+    def test_train_with_set_chat_template(self):
```
lewtun: Should we also test that setting this chat template matches the model one or is that covered by the `clone_chat_template` unit tests?

trl/trainer/sft_config.py
```diff
@@ -38,6 +38,9 @@ class SFTConfig(TrainingArguments):
         model_init_kwargs (`dict[str, Any]` or `None`, *optional*, defaults to `None`):
             Keyword arguments for [`~transformers.AutoModelForCausalLM.from_pretrained`], used when the `model`
             argument of the [`SFTTrainer`] is provided as a string.
+        chat_template (`str` or `None`, *optional*, defaults to `None`):
```
lewtun: Don't we need to update the tokenizer when this is set? That is, `tokenizer.chat_template = args.chat_template`?

tests/test_sft_trainer.py
qgallouedec: Right, I think this part belongs to the `clone_chat_template` test:

https://github.com/huggingface/trl/blob/02a48b732328b808087d0586806c134f27d57d1f/tests/test_dataset_formatting.py#L178-L190

Here the idea is simply to test "can we train a model that initially has no chat template on conversational data"

trl/trainer/sft_config.py
qgallouedec: Following discussion:

For now, `chat_template` does not support full Jinja templates. Setting custom Jinja template is considered an advanced use case, and we assume users who rely on it typically modify the tokenizer manually. Most users will likely just want to clone an existing chat template.

We may support full template definitions in the future, but managing added tokens won’t be automatic: users will probably need to ensure all required tokens from the tokenizer are properly added.

lewtun: As discussed offline, let's change this to `chat_template_path` and let it point to either a local Jinja file or a Hub repo ID

edbeeching (approved):

PR #3617, opened 2025-06-18
🧬 Add `generation_kwargs` as a property of `GRPOConfig` to support additional generation arguments.

### LINKED ISSUE #3562: add generation_kwargs to `GRPOTrainer`, so people have more control when training
### Feature request

I would like to have more control over grpo generation hyperparameters.
currently `GRPOConfig` allows only some of the features for the generations: e.g. temp, top_k, top_p, repetition_penalty and maybe some others.
The code overrides generation_config that the model uses, which means that it is not possible to change the models config to get the desired affect.

https://github.com/huggingface/trl/blob/1314aac502979d6ca7064e32e33b844e22f78823/trl/trainer/grpo_trainer.py#L668-L680

https://github.com/huggingface/trl/blob/1314aac502979d6ca7064e32e33b844e22f78823/trl/trainer/grpo_trainer.py#L1079-L1089

### Motivation

This could give more flexibility when training using the trainer. e.g. we may want to supress some tokens when generating but it is currently not possible to do this

### Your contribution

I might be able to help at a later date. so if someone would like to help, it will be great

pramodith:
# What does this PR do?

This PR is an attempt to address #3562 to allow users to specify additional arguments that can be passed to either vllms `SamplingParams` or HFs `GenerationConfig` to give a user more control over how the old policy generates sample responses.

Fixes #3562

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/scripts/vllm_serve.py
+++ b/trl/scripts/vllm_serve.py
@@ -487,8 +487,7 @@ def main(script_args: ScriptArguments):
             "max_tokens": request.max_tokens,
             "guided_decoding": guided_decoding,
         }
-        if request.generation_kwargs is not None:
-            generation_kwargs.update(request.generation_kwargs)
+        generation_kwargs.update(request.generation_kwargs)
         sampling_params = SamplingParams(**generation_kwargs)
 
         # Evenly distribute prompts across DP ranks
--- a/trl/scripts/vllm_serve.py
+++ b/trl/scripts/vllm_serve.py
@@ -433,6 +433,7 @@ def main(script_args: ScriptArguments):
         min_p: float = 0.0
         max_tokens: int = 16
         guided_decoding_regex: Optional[str] = None
+        generation_kwargs: dict = field(default_factory=dict)
 
     class GenerateResponse(BaseModel):
         completion_ids: list[list[int]]
@@ -445,6 +446,15 @@ def main(script_args: ScriptArguments):
         Args:
             request (`GenerateRequest`):
                 - `prompts` (list of `str`): A list of prompts (text strings) for the model to generate completions.
+                - `n` (`int`, *optional*, defaults to `1`): Number of completions to generate for each prompt.
+                - `repetition_penalty` (`float`, *optional*, defaults to `1.0`): Repetition penalty to apply during generation.
+                - `temperature` (`float`, *optional*, defaults to `1.0`): Temperature for sampling. Higher values lead to more random outputs.
+                - `top_p` (`float`, *optional*, defaults to `1.0`): Top-p (nucleus) sampling parameter. It controls the diversity of the generated text.
+                - `top_k` (`int`, *optional*, defaults to `-1`): Top-k sampling parameter. If set to `-1`, it disables top-k sampling.
+                - `min_p` (`float`, *optional*, defaults to `0.0`): Minimum probability threshold for sampling.
+                - `max_tokens` (`int`, *optional*, defaults to `16`): Maximum number of tokens to generate for each completion.
+                - `guided_decoding_regex` (`str`, *optional*): A regex pattern for guided decoding. If provided, the model will only generate tokens that match this regex pattern.
+                - `generation_kwargs` (`dict`, *optional*): Additional generation parameters to pass to the vLLM `SamplingParams`. This can include parameters like `seed`, `frequency_penalty`, etc. If it contains keys that conflict with the other parameters, they will override them.
 
         Returns:
             `GenerateResponse`:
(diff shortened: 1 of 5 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -670,17 +685,11 @@ def __init__(
             self.accelerator.wait_for_everyone()
         else:
             self.generation_config = GenerationConfig(
-                max_new_tokens=self.max_completion_length,
-                do_sample=True,
                 pad_token_id=processing_class.pad_token_id,
                 bos_token_id=processing_class.bos_token_id,
                 eos_token_id=processing_class.eos_token_id,
```
qgallouedec: why not these ones as well?

trl/trainer/grpo_trainer.py
```diff
@@ -506,6 +506,21 @@ def __init__(
...
+                ("top_k", self.top_k),
+                ("temperature", self.temperature),
+                ("min_p", self.min_p),
+                ("repetition_penalty", self.repetition_penalty),
+                ("max_new_tokens", self.max_completion_length),
+                ("do_sample", True),
+            ]
+            for key, value in default_args:
+                if key not in self.generation_kwargs:
+                    self.generation_kwargs[key] = value
```
qgallouedec: You should be able to simply this with

```python
dict_a.update(dict_b)
```

trl/trainer/grpo_trainer.py
pramodith: I guess from my point of view these are properties stemming from the tokenizer and shouldn't really be tinkered with.

trl/trainer/grpo_trainer.py
pramodith: Hmmm, I think I went for something else, so my position was that any (key, value) pair that the user includes in `generation_kwargs` override any default args like `self.top_p`. An update would end up overriding the argument set in `generation_kwargs`. So in this code block here, I only set a currently supported generation related argument when its not already included in `generation_kwargs`

qgallouedec: Ok, your approach makes sense I think. But still, you can simply with something like:

```python
generation_kwargs = {"top_p": args.top_p, "top_k": args.top_k, ...}
generation_kwargs.update(args.generation_kwargs)
```

qgallouedec: Also, this:

> so my position was that any (key, value) pair that the user includes in generation_kwargs override any default args

should be clearly stated in the doc. That `args.generation_kwargs` has the precedence.

pramodith: Cool done!

pramodith: > Also, this:
> 
> > so my position was that any (key, value) pair that the user includes in generation_kwargs override any default args
> 
> should be clearly stated in the doc. That `args.generation_kwargs` has the precedence.

I thought I did do that here in the documentation: https://github.com/pramodith/trl/blob/2fea0c49ad365b74d84b5e9de5c5d84b25d34513/trl/trainer/grpo_config.py#L91

or is there another place where I need to include that info?

trl/trainer/grpo_config.py
```diff
@@ -85,6 +85,10 @@ class GRPOConfig(TrainingArguments):
             Float that penalizes new tokens based on whether they appear in the prompt and the generated text so far.
             Values > `1.0` encourage the model to use new tokens, while values < `1.0` encourage the model to repeat
             tokens.
+        generation_kwargs (`dict[str, Any]` or `None`, *optional*, defaults to `None`):
+            Additional keyword arguments to pass to the model's `generate` method. If `None`, it defaults to an empty
```
qgallouedec: Actually, it's passed to GenerationConfig, not `generate`. I would clarify that, depending on how you generate (using vLLM or not), this dictionary will populate either `SamplingParams` (for vLLM) or `GenerationConfig` (for other frameworks).

> override the default recommended generation arguments

To be precise, it will also override any generation config passed like `min_p`

trl/trainer/grpo_trainer.py
```diff
@@ -515,6 +515,16 @@ def __init__(
...
+        self.generation_kwargs = {
+            "top_p": self.top_p,
+            "top_k": self.top_k,
+            "temperature": self.temperature,
+            "min_p": self.min_p,
+            "repetition_penalty": self.repetition_penalty,
+            "max_new_tokens": self.max_completion_length,
+            "do_sample": True,
+        }
+        self.generation_kwargs.update(args.generation_kwargs)
```
qgallouedec: very good

trl/trainer/grpo_trainer.py
qgallouedec: ok makes sense

qgallouedec:
I think there is an issue with top_p. The values are not handled the same in vllm and in transformers. I don't think we can have a shared `generation_kwargs`. Trying to make it local, keep you posted

qgallouedec:
And it seems not to work with vLLM server

qgallouedec:
Ok it should be better now. And I also added the training to the test, because the sampling happens during training. Just initializing was a bit lite

pramodith:
Thanks for those changes! Definitely dropped the ball on testing out with vllm.

trl/scripts/vllm_serve.py
```diff
@@ -487,7 +487,8 @@ async def generate(request: GenerateRequest):
             "max_tokens": request.max_tokens,
             "guided_decoding": guided_decoding,
         }
-        generation_kwargs.update(request.generation_kwargs)
+        if request.generation_kwargs is not None:
```
qgallouedec: I don't think this can be None. If not provided, it should be an empty dict

Issue #3320, opened 2025-04-17
Add Adaptive Entropy Control to GRPOTrainer


cuts2k:
### Feature request

Reading Skywork's blogpost on their OR1 model I found this interesting modification where they use entropy as an extra regularization but with a dynamic weight such that it would try to aim for a specific value. 
The relevant part of the blog is here:
https://capricious-hydrogen-41c.notion.site/Skywork-Open-Reasoner-Series-1d0bc9ae823a80459b46c149e4f51680?pvs=25#1d1bc9ae823a801592a0c3891ea5328f
While I haven't tested this myself this looks like a promising enhancement to the GRPOTrainer and hopefully shouldn't be very hard to add.

### Motivation

Just thought I'd point out an interesting enhancement in a likely overlooked post. I'm not in any way associated with Skywork and haven't tested this, but theoretically makes sense.

### Your contribution

Unlikely I'll be able to help, sorry.

1485840691-eng:
@qgallouedec what do you think of the entropy regularization in helping improve model generalization?  Similar entropy regularization methods have also been applied to other RL algos, its pitfall is its high sensitivity to training data and difficulty to control the  coefficient, adaptive entropy control provides one alternative to these problems, but not sure how much it could help. If we think the feature is good for GPRO, I could help provide a PR for it.

vinceamaz:
@1485840691-eng  Would be nice to see official support of Entropy Loss in GRPO. It seems to be standard in frameworks like verl. I also came across a paper called "Reinforcement Learning for Reasoning in Large Language Models with One Training Example". They did an ablation study with entropy loss and found that it improves LLM on MATH500 from 70.8 to 74.8.

qgallouedec:
Sounds promising! We are very opened to contributions

avishaiElmakies:
I could try doing a PR for this

vatsal-kr:
Is someone working on this PR as of now?

1485840691-eng:
Sorry missing reply, I am working on that and would send a PR soon.

1485840691-eng:
@vinceamaz  thanks for your knowledge on that. I have read that and related code in verl, would try to leverage as much as possible. The entropy part would be like acting upon the log probabilities of whole vocab.

![Image](https://github.com/user-attachments/assets/533b875c-bd57-413c-9ab2-8333199fd156)

albertvillanova:
[...]

PR #3628, opened 2025-06-22
Add Entropy Control to GRPOTrainer

1485840691:
# What does this PR do?

Fixes # (3320) 
https://github.com/huggingface/trl/issues/3320

The initial step is to support static entropy control
Next step is to support adaptive entropy control

## Before submitting
- [ N] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ Y] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ Y] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [Y ] Did you make sure to update the documentation with your changes?
- [ N] Did you write any new necessary tests?

## Who can review?


LeonEricsson:
Note that there is a parallel PR (#3563) working on entropy based filtering, we're going to need to sync these

trl/trainer/grpo_config.py
```diff
@@ -194,6 +194,8 @@ class GRPOConfig(TrainingArguments):
             masked. Recommended value is `0.8`.
         use_liger_loss (`bool`, *optional*, defaults to `False`):
             Whether to use the Liger GRPO loss.
+        entropy_coef (`float` or `None`, *optional*, defaults to `0.0`):
+            Coef of entropy regularization loss. If not specified, default to 0.0
```
```suggestion
            Weight of the entropy regularization term in the loss. A positive coefficient adds an entropy bonus to encourage exploration.
```
also change below

trl/trainer/grpo_trainer.py
```diff
@@ -555,6 +555,9 @@ def __init__(
         # This acts as a flag to indicate that the warning has already been issued.
         model.warnings_issued["estimate_tokens"] = True
 
+        # Entropy loss coef
+        self.entropy_coef = args.entropy_coef
```
```suggestion
        self.entropy_coef = args.entropy_coef
```

move up under `# Training arguments`

trl/trainer/grpo_trainer.py
```diff
@@ -1442,6 +1442,21 @@ def _compute_loss(self, model, inputs):
...
+            if self.loss_type == "grpo":
+                entropy_loss = (
+                    (per_token_entropy * completion_mask).sum(-1) / completion_mask.sum(-1).clamp(min=1.0)
+                ).mean()
+            elif self.loss_type == "bnpo":
+                entropy_loss = (per_token_entropy * completion_mask).sum() / completion_mask.sum().clamp(min=1.0)
+            elif self.loss_type == "dr_grpo":
+                entropy_loss = (per_token_entropy * completion_mask).sum() / (
+                    per_token_entropy.size(0) * self.max_completion_length
+                )
```
LeonEricsson: how about we refactor this into a loss aggregation function to avoid duplicating the policy loss calculation just above (L1433-1440).

1485840691: Like https://github.com/volcengine/verl/blob/433544f0be95ed0d35af9d408c451c61b75ade58/verl/trainer/ppo/core_algos.py#L563 ? It has specified the loss type and aggregates loss in different ways.

trl/trainer/grpo_trainer.py
1485840691: Done

trl/trainer/grpo_config.py
1485840691: Done

trl/trainer/grpo_trainer.py
LeonEricsson: Sure, something like that seems reasonable, but stick with our naming convention (bnpo, dr_gpro, ..)

1485840691:
@LeonEricsson @qgallouedec   could you please help review the latest changes? Thanks

trl/trainer/grpo_trainer.py
```diff
@@ -1797,18 +1853,23 @@ def _compute_loss(self, model, inputs):
...
-            raise ValueError(f"Unknown loss type: {self.loss_type}")
+        loss = agg_loss(
+            per_token_loss, completion_mask, self.loss_type, max_completion_length=self.max_completion_length
+        )
 
         # Log the metrics
         mode = "train" if self.model.training else "eval"
 
+        if self.entropy_coef != 0:
+            per_token_entropy = logps_and_entropies["entropies"]
```
LeonEricsson: need to update, entropies is now in `entropies` variable

trl/trainer/grpo_config.py
```diff
@@ -534,6 +536,43 @@ class GRPOConfig(TrainingArguments):
...
+            "Used to adjust entorpy term based on current entropy in adaptive entropy control."
+        },
+    )
+    target_entropy: float = field(
+        default=0.0,
+        metadata={
+            "help": "Target entropy value as the upper bound of entropy loss. "
+            "Target entropy value in adaptive entropy control."
+        },
+    )
```
LeonEricsson: these should added to the class docstring as well

trl/trainer/grpo_config.py
LeonEricsson: Seems like your fix was undone? Should be same as the field help in the definition further down in the config

trl/trainer/grpo_config.py
LeonEricsson: Due to the difficulty of setting these parameters we should provide a recommended value in the comments (see other config params for inspo)

trl/trainer/grpo_config.py
```diff
@@ -534,6 +536,43 @@ class GRPOConfig(TrainingArguments):
         default=False,
         metadata={"help": "Whether to use the Liger GRPO loss."},
     )
+    entropy_coef: float = field(
+        default=0.0,
+        metadata={
+            "help": "Weight of the entropy regularization term in the loss. "
+            "A positive coefficient adds an entropy bonus to encourage exploration."
+            "A negative coefficient dynamically adjusts the weight based on entropy loss."
```
LeonEricsson: Not a fan of this kind of configuration. I understand that you want to minimize the number of config params but I would prefer to separate entropy regularization and adaptive entropy control. Have adaptive entropy control as an explicit boolean config param that overrides entropy_coef.

LeonEricsson (commented):
Thanks for the work. A few comments on my end

1485840691:
> Thanks for the work. A few comments on my end

Thanks for your comments. Resolved

trl/trainer/grpo_config.py
```diff
@@ -201,6 +201,24 @@ class GRPOConfig(TrainingArguments):
             If used with `mask_truncated_completions=True`, only tokens from non-truncated completions are considered.
         use_liger_loss (`bool`, *optional*, defaults to `False`):
             Whether to use the Liger GRPO loss.
+        ent_coef (`float`, *optional*, defaults to `0.0`):
```
```suggestion
        entropy_coef (`float`, *optional*, defaults to `0.0`):
```

trl/trainer/grpo_config.py
```diff
@@ -201,6 +201,24 @@ class GRPOConfig(TrainingArguments):
...
+            Static weight of the entropy regularization term in the loss.
+            A positive coefficient adds an entropy bonus to encourage exploration.
+            A zero or negative coefficient disables the entropy regularization term.
```
```suggestion
            Weight of the entropy regularization term in the loss.
            A positive coefficient adds an entropy bonus to encourage exploration.
```

trl/trainer/grpo_config.py
```diff
@@ -201,6 +201,24 @@ class GRPOConfig(TrainingArguments):
...
+        use_adapt_ent (`bool`, *optional*, defaults to `False`):
+            Whether to apply adaptive entropy control. Override ent_coef if set to True.
+        min_ent_coef (`float`, *optional*, defaults to `0.0`):
+            Minimum weight of the entropy regularization term.
+            Used to cap the minimum entropy term weight in adaptive entropy control.
```
```suggestion
        entropy_coef_min (`float`, *optional*, defaults to `0.0`):
            Lower bound for entropy coefficient when using adaptive entropy control.
```

trl/trainer/grpo_config.py
```diff
@@ -201,6 +201,24 @@ class GRPOConfig(TrainingArguments):
...
+        max_ent_coef (`float`, *optional*, defaults to `1.0`):
+            Maximum weight of the entropy regularization term.
+            Used to cap the minimum entropy term weight in adaptive entropy control.
```
```suggestion
        entropy_coef_max (`float`, *optional*, defaults to `1.0`):
            Upper bound for entropy coefficient when using adaptive entropy control.
```

trl/trainer/grpo_config.py
```diff
@@ -201,6 +201,24 @@ class GRPOConfig(TrainingArguments):
...
+        delta_ent_coef (`float`, *optional*, defaults to `0.005`):
+            Per step delta weight of the entropy regularization term.
+            Used to adjust the entropy term based on current entropy in adaptive entropy control.
```
```suggestion
        entropy_coef_delta (`float`, *optional*, defaults to `0.005`):
            Step size for adjusting entropy coefficient during adaptive entropy control.
```

trl/trainer/grpo_config.py
```diff
@@ -201,6 +201,24 @@ class GRPOConfig(TrainingArguments):
...
+        target_ent (`float`, *optional*, defaults to `0.2`):
+            Target entropy value as the upper bound of entropy loss.
+            The entropy we aim to maintain in adaptive entropy control.
```
```suggestion
        entropy_target (`float`, *optional*, defaults to `0.2`):
            Desired target entropy value maintained through adaptive entropy control.
```

trl/trainer/grpo_trainer.py
```diff
@@ -300,6 +301,49 @@ def identity(x):
...
+
+        Returns:
+            `float`: Adjusted entropy coefficient.
+        """
+        if ent < self.target_ent:
+            self.value += self.delta_ent_coef
+        else:
+            self.value -= self.delta_ent_coef
+
+        self.value = float(np.clip(self.value, self.min_ent_coef, self.max_ent_coef))
```
LeonEricsson: let's avoid importing numpy just for this

```suggestion
        self.value = float(max(self.min_ent_coef, min(self.value, self.max_ent_coef)))
```

trl/trainer/grpo_config.py
```diff
@@ -201,6 +201,24 @@ class GRPOConfig(TrainingArguments):
             If used with `mask_truncated_completions=True`, only tokens from non-truncated completions are considered.
         use_liger_loss (`bool`, *optional*, defaults to `False`):
             Whether to use the Liger GRPO loss.
+        ent_coef (`float`, *optional*, defaults to `0.0`):
+            Static weight of the entropy regularization term in the loss.
+            A positive coefficient adds an entropy bonus to encourage exploration.
+            A zero or negative coefficient disables the entropy regularization term.
+        use_adapt_ent (`bool`, *optional*, defaults to `False`):
+            Whether to apply adaptive entropy control. Override ent_coef if set to True.
```
```suggestion
        use_adapt_entropy (`bool`, *optional*, defaults to `False`):
            Whether to apply adaptive entropy control. Overrides `entropy_coef`.
```

trl/trainer/grpo_trainer.py
```diff
@@ -676,6 +720,18 @@ def __init__(
             raise NotImplementedError(
                 "Liger Kernels don't currently support masking token positions based on entropy."
             )
+        # Entropy loss weight
+        self.ent_coef = max(args.ent_coef, 0.0)
```
LeonEricsson: I think we can allow the user to set a negative weight if they choose to. I don’t see a specific use case for it, but I don't see the harm in allowing it

trl/trainer/grpo_trainer.py
LeonEricsson: What do you think about changing this to a EntropyController, that handles adaptive and static entropy control. That means we always call EntropyController to retrieve our entropy coefficient, and it handles whether that is an adaptive entropy or a static. Or do you see a clear benefit to having the branching be outside of the controller?

LeonEricsson (commented):
few more comments

trl/trainer/grpo_trainer.py
1485840691: What if the user set a negative weight for it. Should we directly multiply the negative weight to the entropy loss?

trl/trainer/grpo_trainer.py
1485840691: > What do you think about changing this to a EntropyController, that handles adaptive and static entropy control. That means we always call EntropyController to retrieve our entropy coefficient, and it handles whether that is an adaptive entropy or a static. Or do you see a clear benefit to having the branching be outside of the controller?

Your suggestion make sense. Will do it

trl/trainer/grpo_trainer.py
LeonEricsson: yes

trl/trainer/grpo_config.py
```diff
@@ -207,6 +207,20 @@ class GRPOConfig(TrainingArguments):
             If used with `mask_truncated_completions=True`, only tokens from non-truncated completions are considered.
         use_liger_loss (`bool`, *optional*, defaults to `False`):
             Whether to use the Liger GRPO loss.
+        entropy_coef (`float`, *optional*, defaults to `0.0`):
+            Static coefficient of the entropy regularization term in the loss.
+            A positive coefficient adds an entropy bonus to encourage exploration.
+            It is also used as the initial entropy coefficient when using adaptive entropy control.
+        use_adapt_entropy (`bool`, *optional*, defaults to `False`):
```
LeonEricsson: sorry for pettiness but can can we do
```suggestion
        use_adaptive_entropy (`bool`, *optional*, defaults to `False`):
```

trl/trainer/grpo_trainer.py
```diff
@@ -300,6 +300,62 @@ def identity(x):
...
+class EntropyController:
+    def __init__(
+        self, use_adapt_ent=False, ent_coef=0.0, min_ent_coef=0.0, max_ent_coef=0.0, delta_ent_coef=0.0, target_ent=0.0
+    ):
+        self.use_adapt_ent = use_adapt_ent
+        self.ent_coef = ent_coef
+        self.min_ent_coef = min_ent_coef
+        self.max_ent_coef = max_ent_coef
+        self.delta_ent_coef = delta_ent_coef
+        self.target_ent = target_ent
```
LeonEricsson: change these everywhere the same way we did in config, e.g `entropy_coef_min`. Also use `entropy` instead of `ent` throughout.

LeonEricsson:
While reviewing the updated entropy controller I noted the following issues, which I should of realized sooner, apologies for that.

1. **Hidden mutable state**
   The class keeps an internal entropy coefficient that mutates every call. Because that state lives outside the trainer/optimizer stack, it’s easy to miss in tests, logs, or checkpoints and makes debugging non-deterministic behaviour harder.

2. **Distributed training**
   Right now every rank updates the coefficient from its *local* entropy, so the values drift apart. That means different GPUs are optimising slightly different objectives. The paper intends for a single global coefficient.
   

I suggest moving ownership of the entropy coefficient parameter to GRPOTrainer, making the entropy controller a pure strategy object which holds logic to step the entropy coefficient (rename `__call__` to `step()`), change name to `EntropyScheduler`, and use global entropy to step the entropy coefficient. Also broadcast coefficient to all ranks. Something like this (reduce() and broadcast() are placeholders)

```
entropy = agg_loss(...)

world_entropy = reduce(entropy.detach(), reduction="mean")

    self.entropy_coef = self.entropy_scheduler.step(
        self.entropy_coef, world_entropy
    )                   

broadcast(self.ent_coef, src=0)

loss = loss - self.ent_coef * entropy_loss
```

1485840691:
> While reviewing the updated entropy controller I noted the following issues, which I should of realized sooner, apologies for that.
> 
> 1. **Hidden mutable state**
>    The class keeps an internal entropy coefficient that mutates every call. Because that state lives outside the trainer/optimizer stack, it’s easy to miss in tests, logs, or checkpoints and makes debugging non-deterministic behaviour harder.
> 2. **Distributed training**
>    Right now every rank updates the coefficient from its _local_ entropy, so the values drift apart. That means different GPUs are optimising slightly different objectives. The paper intends for a single global coefficient.
> 
> I suggest moving ownership of the entropy coefficient parameter to GRPOTrainer, making the entropy controller a pure strategy object which holds logic to step the entropy coefficient (rename `__call__` to `step()`), change name to `EntropyScheduler`, and use global entropy to step the entropy coefficient. Also broadcast coefficient to all ranks. Something like this (reduce() and broadcast() are placeholders)
> 
> ```
> entropy_loss = agg_loss(...)
> 
> world_entropy = reduce(entropy_loss.detach(), reduction="mean")
> 
> if self.accelerator.is_main_process:
>     self.entropy_coef = self.entropy_scheduler.step(
>         self.entropy_coef, world_entropy
>     )                   
> 
> broadcast(self.ent_coef, src=0)
> 
> loss = loss - self.ent_coef * entropy_loss
> ```

Yes, I also think it might be better to use a global scheduler to schedule the update of entropy coef based on global entropy loss gathered from all ranks. I took a look at the original code of skywork and think that it might be using a per rank scheduler to control entropy coef. If you have time, could you please help confirm it?

The entropy loss apply entropy coef: https://github.com/SkyworkAI/Skywork-OR1/blob/64e96afa213ae89d0ad21932106d3b8aafe9ace2/verl/workers/actor/dp_actor.py#L234

The entropy controller defined inside trainer

https://github.com/SkyworkAI/Skywork-OR1/blob/64e96afa213ae89d0ad21932106d3b8aafe9ace2/verl/trainer/ppo/ray_trainer.py#L391

https://github.com/SkyworkAI/Skywork-OR1/blob/64e96afa213ae89d0ad21932106d3b8aafe9ace2/verl/trainer/ppo/ray_trainer.py#L1097C25-L1098C1

LeonEricsson:
@qgallouedec would appreciate your thoughts on dealing with the stateful entropy coefficient. To recap, Adaptive Entropy Control maintains the entropy coefficient $\alpha_k$ as an adaptive (or running) coefficient, which is incrementally updated on each optimizer step based on the batches entropy. Is something like this sufficient for maintaining a global entropy coefficient?

```
entropy = agg_loss(...)

world_entropy = all_reduce(entropy.detach(), reduction="mean")

self.entropy_coef = self.entropy_scheduler.step(
        self.entropy_coef, world_entropy
)                   

loss = loss - self.entropy_coef * entropy_loss
```

1485840691:
> @qgallouedec would appreciate your thoughts on dealing with the stateful entropy coefficient. To recap, Adaptive Entropy Control maintains the entropy coefficient α k as an adaptive (or running) coefficient, which is incrementally updated on each optimizer step based on the batches entropy. Is something like this sufficient for maintaining a global entropy coefficient?
> 
> ```
> entropy = agg_loss(...)
> 
> world_entropy = reduce(entropy.detach(), reduction="mean")
> 
> if self.accelerator.is_main_process:
>     self.entropy_coef = self.entropy_scheduler.step(
>         self.entropy_coef, world_entropy
>     )                   
> 
> broadcast(self.entropy_coef)
> 
> loss = loss - self.entropy_coef * entropy_loss
> ```

https://github.com/huggingface/trl/pull/3628/commits/8c0868290f83091e330782d34d92a5adcd199d7e

I have submitted an initial update

My comment:

    Could we only use a global entropy loss gathered from all rank?
    In each rank, the process retains its own state of entropy coef and update based on global entropy loss
    Any potential risk of applying this simpler update? 

@LeonEricsson  what do you think of my suggested approach of only using global entropy loss but keep per rank entropy scheduler? The entropy coef computation could be scattered in each rank but not blocked by main process. Besides, given each rank computes entropy coef based on global entropy loss, the entropy coef in each rank might be updated almost in the same path

tests/test_grpo_trainer.py
```diff
@@ -251,6 +253,62 @@ def test_sampler_with_mini_repeat_count_and_batch_size_3(self):
         assert sampled[24:28] == sampled[28:32] == sampled[32:36]
 
 
+class EntropyControlTester(unittest.TestCase):
```
LeonEricsson: These tests need updating to match the new class signature

tests/test_grpo_trainer.py
```diff
@@ -251,6 +253,62 @@ def test_sampler_with_mini_repeat_count_and_batch_size_3(self):
...
+            entropy_coef_delta=0.01,
+            entropy_target=0.2,
+        )
+        entropies = torch.tensor([[0.1, 0.2, 0.3, 1.4, 0.5, 0.14], [0.5, 0.6, 0.7, 0.8, 0.9, 1.0]])
+        mask = torch.tensor([[1, 1, 1, 1, 0, 0], [1, 1, 1, 1, 0, 0]])
+        entropy_loss = aggregate_loss(entropies, mask, "grpo")
+        entropy_loss_coef = ent_ctrl(entropy_loss.item())
+        self.assertEqual(
+            entropy_loss_coef, 0.0, f"Expected entropy loss coefficient to be 0.0, got {entropy_loss_coef}"
+        )
```
LeonEricsson: Why is this expected to be 0? Should it not be 0.1 + (-1 * 0.01)?

trl/trainer/grpo_trainer.py
```diff
@@ -186,6 +186,34 @@ def __len__(self) -> int:
...
+        Args:
+            entropy (`float`): Current entropy value.
+            entropy_coef (`float`): Current entropy coef.
+
+        Returns:
+            `float`: Adjusted entropy coefficient.
+        """
+        new_entropy_coef = entropy_coef + (1 if entropy < self.entropy_target else -1) * self.entropy_coef_delta
+        new_entropy_coef = float(max(self.entropy_coef_min, min(new_entropy_coef, self.entropy_coef_max)))
+        return new_entropy_coef * (entropy < self.entropy_target)
```
LeonEricsson: What is the purpose of this? This will set the entropy coefficient to 0 whenever the entropy is less than the target entropy.

1485840691: According to the original post of adaptive entropy control release by skywork-OR1

image

https://github.com/SkyworkAI/Skywork-OR1/blob/64e96afa213ae89d0ad21932106d3b8aafe9ace2/verl/trainer/ppo/core_algos.py#L74
https://github.com/SkyworkAI/Skywork-OR1/blob/64e96afa213ae89d0ad21932106d3b8aafe9ace2/verl/workers/actor/dp_actor.py#L284

tests/test_grpo_trainer.py
1485840691: Because of the below  code:

      new_entropy_coef * (entropy < self.entropy_target)

The test case gives entropy loss higher than entropy target, so the coef will be zero.

trl/trainer/grpo_trainer.py
LeonEricsson: Aaah, I missed this part, thanks for that. That means we need to track $c_k$ separately from $a_k$. I'm pushing a change now, look good to you?

trl/trainer/grpo_trainer.py
```diff
@@ -185,6 +185,48 @@ def __len__(self) -> int:
         return (self.num_samples // self.batch_size) * self.batch_size * self.mini_repeat_count * self.repeat_count
 
 
+class EntropyController:
```
qgallouedec, line 189: I don't think this controller is needed. Maybe it's simpler with only attributes in GRPO?

trl/trainer/grpo_trainer.py
```diff
@@ -299,6 +341,19 @@ def identity(x):
...
+    if loss_type == "grpo":
+        loss = ((per_token_loss * loss_mask).sum(-1) / loss_mask.sum(-1).clamp(min=1.0)).mean()
+    elif loss_type == "bnpo":
+        loss = (per_token_loss * loss_mask).sum() / loss_mask.sum().clamp(min=1.0)
+    elif loss_type == "dr_grpo":
+        max_completion_length = kwargs.get("max_completion_length", 1)
+        loss = (per_token_loss * loss_mask).sum() / (per_token_loss.size(0) * max_completion_length)
+    else:
+        raise ValueError(f"Unknown loss type: {loss_type}")
+    return loss
```
qgallouedec, line 362: is this change needed?

qgallouedec: oh ok I get it, it's because you re-use this function for the entropy as well

1485840691:
@lewtun @qgallouedec @kashif @sergiopaniego  any more feedback on this PR?

trl/trainer/grpo_trainer.py
albertvillanova: This is a fair point worth discussing. The controller class works well when the adaptive logic grows (e.g. more scheduling strategies), but for now it is a thin wrapper.

albertvillanova (commented):
Hey @1485840691 — apologies for going quiet on this. We have been working through a backlog of open PRs and should have responded sooner; that's on us. 

The work you have put in here is appreciated, and the feature itself (entropy control in GRPO, based on Skywork OR1) is genuinely useful. 

Are you still interested in contributing this feature?

PR #3639, opened 2025-06-24
☕️ GRPO script reward_funcs error

tcapelle:
We should pass the list of functions to the Trainer, and not just the reward model (that is actually None most of the time)

### DIFF
--- a/trl/scripts/grpo.py
+++ b/trl/scripts/grpo.py
@@ -100,7 +100,7 @@ def main(script_args, training_args, model_args):
     # Initialize the GRPO trainer
     trainer = GRPOTrainer(
         model=model,
-        reward_funcs=reward_model,
+        reward_funcs=reward_funcs,
         args=training_args,
         train_dataset=dataset[script_args.dataset_train_split],
         eval_dataset=dataset[script_args.dataset_test_split] if training_args.eval_strategy != "no" else None,


shirinyamani:
Hi @tcapelle, thanks for pointing it. 
But i think the reason is technically grpo supports both model and func in terms of acceptable reward, and in this specific example, since `reward_funcs=[]` while we have a reward_model, that is why we put the reward_model instead of the func.


 
```
    # Get the reward models and functions
    reward_funcs = []
    if script_args.reward_model_name_or_path:
        reward_model = AutoModelForSequenceClassification.from_pretrained(
            script_args.reward_model_name_or_path, trust_remote_code=model_args.trust_remote_code, num_labels=1
        )
        reward_funcs.append(reward_model)

    if script_args.reward_funcs:
        for func_name in script_args.reward_funcs:
            if func_name in reward_funcs_registry:
                reward_funcs.append(reward_funcs_registry[func_name])
            elif "." in func_name:
                module_path, func_name = func_name.rsplit(".", 1)
                sys.path.insert(0, os.getcwd())
                module = importlib.import_module(module_path)
                reward_func = getattr(module, func_name)
                reward_funcs.append(reward_func)
            else:
                raise ValueError(
                    f"Could not load reward function '{func_name}'. Expected one of "
                    f"{list(reward_funcs_registry.keys())} or a valid import path."
                ) 
```

tcapelle:
As of right now, if you pass reward funcs they don't get passed to the Trainer.

shirinyamani:
@tcapelle correct!

shirinyamani (approved):

Issue #3644, opened 2025-06-25
0.19.0 breaks the chat format in GRPO trainer leading to severely mangled chat generations


mchugha2:
### Reproduction

At line 1068 in grpo_trainer.py:

```python
prompts_text = self.processing_class.batch_decode(
    prompt_ids, skip_special_tokens=True, clean_up_tokenization_spaces=False
)
```

```python
skip_special_tokens=True
```

This removes the necessary chat format tokens for Qwen instruct models and many others. This leads to severely mangled generations. 

I assume it should be:

```python
skip_special_tokens=False
```

### System Info

- Platform: Linux-5.15.0-139-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- TRL version: 0.19.0
- PyTorch version: 2.7.0
- accelerator(s): NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3, NVIDIA H100 80GB HBM3
- Transformers version: 4.52.4
- Accelerate version: 1.8.1
- Accelerate config: not found
- Datasets version: 3.6.0
- HF Hub version: 0.33.0
- bitsandbytes version: not installed
- DeepSpeed version: 0.17.1
- Diffusers version: not installed
- Liger-Kernel version: 0.5.10
- LLM-Blender version: not installed
- OpenAI version: 1.91.0
- PEFT version: not installed
- vLLM version: 0.9.1

### Checklist

- [x] I have checked that my issue isn't already filed (see [open issues](https://github.com/huggingface/trl/issues?q=is%3Aissue))
- [x] I have included my system information
- [x] Any code provided is minimal, complete, and reproducible ([more on MREs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks))
- [x] Any code provided is properly formatted in code blocks, (no screenshot, [more on code blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks))
- [x] Any traceback provided is complete

pramodith:
I'm looking into this one.

pramodith:
[...]

PR #3646, opened 2025-06-25
Prompt Decoding after truncation should not skip special tokens

pramodith:
# What does this PR do?

Quick fix for #3644. Skipping special tokens gets rid of the `<|im_start|>` and `<|im_end|>` chat templating and this text subsequently passed to vllm leads to malformed generations.

However, the way the prompt is currently truncated has a few other issues. For one it gets rid of the very first <|im_start|> if we exceed the `max_prompt_length` configuration. I'll work on a subsequent PR to truncate and reformat the prompt post truncation cleanly and ensure that the chat template remains in-tact post truncation.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1065,7 +1065,7 @@ class GRPOTrainer(Trainer):
             prompt_ids = prompt_ids[:, -self.max_prompt_length :]
             prompt_mask = prompt_mask[:, -self.max_prompt_length :]
             prompts_text = self.processing_class.batch_decode(
-                prompt_ids, skip_special_tokens=True, clean_up_tokenization_spaces=False
+                prompt_ids, skip_special_tokens=False, clean_up_tokenization_spaces=False
             )
 
         # Generate completions using either vLLM or regular generation


LeonEricsson:
Do you plan to split the `skip_special_tokens=False` fix into a separate PR? I think it makes sense to separate them – this way, we can get the fix merged as soon as possible.

pramodith:
Cool, I'll have a new PR with just `skip_special_tokens=False` in the next hour.

pramodith:
@LeonEricsson this one is ready for review now. I've mentioned some of the tradeoffs this solution entails in the description of the PR.

trl/trainer/grpo_config.py
```diff
@@ -260,7 +262,10 @@ class GRPOConfig(TrainingArguments):
     max_prompt_length: Optional[int] = field(
         default=512,
         metadata={
-            "help": "Maximum length of the prompt. If the prompt is longer than this value, it will be truncated left."
+            "help": "Maximum length of the prompt. If the prompt is longer than this value, it will be truncated left. If the "
+            "prompt is conversational, we only truncate the message tokens starting from the top of the conversation
+            "and do not account for any tokens introduced
```
```suggestion
            "prompt is conversational, we only truncate the message tokens starting from the top of the conversation "
            "and do not account for any tokens introduced "
```

trl/trainer/grpo_trainer.py
```diff
@@ -1066,26 +1066,75 @@ def _calculate_rewards(self, inputs, prompts, completions, completion_ids_list):
...
+        # If it is conversational the truncation preserves the chat template.
+        if not is_conversational(prompts[0]):
+            prompt_text = [x["prompt"] for x in prompts]
+            prompt_inputs = self.processing_class(
+                text=prompt_text, return_tensors="pt", padding=True, padding_side="left", add_special_tokens=False
+            )
+            prompt_inputs = super()._prepare_inputs(prompt_inputs)
+            prompt_ids, prompt_mask = prompt_inputs["input_ids"], prompt_inputs["attention_mask"]
+
+            if self.max_prompt_length is not None and prompt_mask.sum(-1).max() > self.max_prompt_length:
```
LeonEricsson, line 1089: could you motivate the need for `prompt_mask.sum(-1).max() > self.max_prompt_length`? is it to avoid unnecessary decode if we don't need to truncate?

trl/trainer/grpo_trainer.py
```diff
@@ -1066,26 +1066,75 @@ def _calculate_rewards(self, inputs, prompts, completions, completion_ids_list):
...
-        prompt_inputs = super()._prepare_inputs(prompt_inputs)
-        prompt_ids, prompt_mask = prompt_inputs["input_ids"], prompt_inputs["attention_mask"]
-
-        if self.max_prompt_length is not None:
-            prompt_ids = prompt_ids[:, -self.max_prompt_length :]
-            prompt_mask = prompt_mask[:, -self.max_prompt_length :]
-            prompts_text = self.processing_class.batch_decode(
-                prompt_ids, skip_special_tokens=False, clean_up_tokenization_spaces=False
-            )
+        prompt_ids, prompt_mask, prompts_text = self._get_prompt_inputs(inputs)
```
LeonEricsson: let's make naming consistent `prompts_ids`, `prompts_mask` ...

LeonEricsson (commented):
I tested running with the following conversational dataset:

```
SYSTEM_PROMPT = """
Respond in the following format:
<reasoning>
...
</reasoning>
<answer>
...
</answer>

"""
def get_gsm8k_questions(split="train") -> Dataset:
    data = load_dataset("openai/gsm8k", "main")[split]  # type: ignore
    data = data.map(
        lambda x: {  # type: ignore
            "prompt": [{"role": "system", "content": SYSTEM_PROMPT}, {"role": "user", "content": x["question"]}],
            "answer": extract_hash_answer(x["answer"]),
        }
    )  # type: ignore
    return data  # type: ignore
```

and the `prompts_text` returned by `_get_prompt_inputs` had the wrong system prompt. They contained the tokenizer's default system prompt instead.

trl/trainer/grpo_trainer.py
```diff
@@ -1066,26 +1066,75 @@ def _calculate_rewards(self, inputs, prompts, completions, completion_ids_list):
...
+                                num_tokens_to_truncate = 0
+                                truncated_messages[-1].append(msg)
+
+                prompt_inputs = self.processing_class.apply_chat_template(
+                    truncated_messages, return_dict=True, add_generation_prompt=True
+                )
+                prompt_inputs = super()._prepare_inputs(prompt_inputs)
+                prompt_ids = prompt_inputs["input_ids"]
+                prompt_mask = prompt_inputs["attention_mask"]
+                prompt_text = self.processing_class.batch_decode(prompt_ids, skip_special_tokens=False)
```
LeonEricsson: this should be outside/after the for-loop, right?

trl/trainer/grpo_trainer.py
pramodith: Yes exactly, gets rid of any redundant ops.

LeonEricsson: Got it. Given we've already padded, won't `prompt_mask.sum(-1).max()` always equal `prompt_ids.shape[-1]` or `prompt_mask.shape[-1]` (we're doing 'longest' padding - _pad to the longest sequence in the batch_)?

pramodith: Yeah that's fair. I'll change that to just use the shape.

trl/trainer/grpo_trainer.py
pramodith: Yep, good catch.

trl/trainer/grpo_trainer.py
pramodith: changing the naming to `prompts_ids` and `prompts_mask` means that we might also want to update the dictionary keys in the returned dict of this function. Meaning that we'll also need to update the keys used to access these tensors throughout the file, just wanted to double check if you're okay with that.

https://github.com/huggingface/trl/blob/02cce41d062a6b2754dfc2f5a31437d475003e45/trl/trainer/grpo_trainer.py#L1363-L1371

LeonEricsson: okay, leave it as is. This inconsistency was around before this PR.

pramodith:
> I tested running with the following conversational dataset:
> 
> ```
> SYSTEM_PROMPT = """
> Respond in the following format:
> <reasoning>
> ...
> </reasoning>
> <answer>
> ...
> </answer>
> 
> """
> def get_gsm8k_questions(split="train") -> Dataset:
>     data = load_dataset("openai/gsm8k", "main")[split]  # type: ignore
>     data = data.map(
>         lambda x: {  # type: ignore
>             "prompt": [{"role": "system", "content": SYSTEM_PROMPT}, {"role": "user", "content": x["question"]}],
>             "answer": extract_hash_answer(x["answer"]),
>         }
>     )  # type: ignore
>     return data  # type: ignore
> ```
> 
> and the `prompts_text` returned by `_get_prompt_inputs` had the wrong system prompt. They contained the tokenizer's default system prompt instead.

What was the `max_prompt_length` configured to? Some chat templates automatically introduce the default system prompt if the system message is missing in the messages for a given batch. This happens when running the `apply_chat_template` function.

So I'm assuming that the `max_prompt_length` that was set for this example eliminated the custom system prompt. Now that I think about it this means that any custom system prompt will always be pruned. Might make more sense for us to start pruning from the first user message instead of starting from the system prompt and only prune from the system prompt in case all turns barring the last user message was pruned.

But I suppose that would make this approach inconsistent with how the non-chat prompt truncation is handled. Let me know what you think the best approach is.

LeonEricsson:
> What was the `max_prompt_length` configured to? Some chat templates automatically introduce the default system prompt if the system message is missing in the messages for a given batch. This happens when running the `apply_chat_template` function.

yes good catch, i believe this was indeed the case.

> So I'm assuming that the `max_prompt_length` that was set for this example eliminated the custom system prompt. Now that I think about it this means that any custom system prompt will always be pruned. Might make more sense for us to start pruning from the first user message instead of starting from the system prompt and only prune from the system prompt in case all turns barring the last user message was pruned.

> But I suppose that would make this approach inconsistent with how the non-chat prompt truncation is handled. Let me know what you think the best approach is.

tbh, I'm not entirely sure what the correct approach is here. I feel like user content should indeed be truncated first. However, I agree that we're moving into territory where handling conversational versus non-conversational formats significantly diverges.

pramodith:
> tbh, I'm not entirely sure what the correct approach is here. I feel like user content should indeed be truncated first. However, I agree that we're moving into territory where handling conversational versus non-conversational formats significantly diverges.

Do we want to rope in someone else for a third opinion? I don't think there's a right approach/answer here prompt truncation is going to always be less than ideal in terms of how it affects training and one can make a case for either approach.

pramodith:
@qgallouedec can you please give a third opinion on this PR?

qgallouedec:
Sorry for the late reply. I'm slowly catching up after my vacation
I agree with @LeonEricsson, I understand the intuition of not breaking the chat template, but do we have solid results to validate this?:

> a naive truncation will break the template leading to poor training results.

And are there any results to show that this approach gives significantly better results? I've never seen a paper or lib using this approach, do you know of any?

pramodith:
> And are there any results to show that this approach gives significantly better results? I've never seen a paper or lib using this approach, do you know of any?

That's a fair question, its perhaps necessary to quantify the impact of this. I can create a colab notebook to test out the results of a model trained on a dataset where 20% of the prompts are truncated. We can then compare the results on a test set when preserving and not preserving the chat template.

Are there any datasets that you'd recommend for this and what model should I be training, is Qwen3-0.6B good enough or do we want to train a larger model?

LeonEricsson:
We're left truncating, so cutting the system prompt first and foremost. We should see this impact training, given the importance of the system prompt in common GRPO applications.   

> Are there any datasets that you'd recommend for this and what model should I be training, is Qwen3-0.6B good enough or do we want to train a larger model?

How about [open-r1/DAPO-Math-17k-Processed ](https://huggingface.co/datasets/open-r1/DAPO-Math-17k-Processed ) with Qwen3 1.7B

PR #3669, opened 2025-06-30
feat: Pass trainer state to reward functions

### LINKED ISSUE #3668: GRPOTrainer reward functions cannot access training progress
### Proposed Solution

Add the trainer state to the reward function kwargs before calling the reward function:

```python
# In GRPOTrainer, before calling the reward function:
reward_kwargs["trainer_state"] = self.state
```

This would allow reward functions to access training progress:

```python
def custom_reward_func(
    completions: List[List[Dict[str, str]]],
    sources: List[Optional[str]],
    trainer_state: Any,
    **kwargs
) -> List[float]:
    """Custom reward function with curriculum learning."""
    
    # Implement reward curriculum based on training progress
    training_progress = trainer_state.global_step / trainer_state.max_steps
    
    # Example: No reward for first 30% of training
    if training_progress < 0.3:
        return [0.0] * len(completions)
    
    # Apply actual reward function after warmup
    responses = [comp[0]['content'] for comp in completions]
    return [compute_reward(response, source) for response, source in zip(responses, sources)]
```

### Benefits

1. **Flexibility**: Enables dynamic reward strategies based on training progress
2. **Backward Compatible**: Existing reward functions would continue to work (they would just ignore the extra parameter)
3. **Minimal Change**: Requires only a single line addition to pass the trainer state
4. **Enables Advanced Training Strategies**: Curriculum learning, warm-up periods, and progressive reward shaping

### Alternative Solutions Considered

1. **Global variable**: Not clean and could cause issues with multiple trainer instances
2. **Callback system**: More complex and overkill for this simple use case
3. **Subclassing GRPOTrainer**: Requires users to maintain custom trainer code

### Additional Context

This feature would be particularly useful for:
- Research on reward curriculum and progressive training
- Production systems that need stable training with gradual reward introduction
- Complex reward functions that benefit from warm-up periods

### Description

Currently, when using custom reward functions with `GRPOTrainer` in TRL, there's no way to access the training progress information (such as `global_step` and `max_steps`) within the reward function. This limitation prevents implementing dynamic reward strategies like reward curriculum, where rewards can be adjusted based on training progress.

### Use Case

A common use case is implementing reward curriculum, where different reward strategies are applied at different stages of training. For example:
- Early training: Use simpler rewards or no rewards to allow the model to explore
- Mid training: Gradually introduce more complex reward signals
- Late training: Apply full reward function

### Problem

The reward function signature only receives:
- `prompts`
- `completions`
- `completion_ids` 
- Additional custom kwargs

However, it lacks access to:
- Current global step
- Total training steps
- Overall training progress

This makes it impossible to implement training-aware reward strategies.

### Your contribution

I will write a simple PR for it.

seungduk-yanolja:
This allows for implementing dynamic reward strategies like curriculum learning, where rewards can be adjusted based on training progress. Fixes #3668

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1276,8 +1276,9 @@ class GRPOTrainerTester(unittest.TestCase):
         # Test if trainer can handle reward function that accesses the trainer state
         dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
 
-        def reward_func(completions, trainer_state, **kwargs):
+        def reward_func(completions, **kwargs):
             # Assert that trainer_state is passed and has the expected attributes
+            trainer_state = kwargs.get("trainer_state")
             assert trainer_state is not None
             assert hasattr(trainer_state, "global_step")
             return [float(len(set(completion))) for completion in completions]
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1273,13 +1273,12 @@ class GRPOTrainerTester(unittest.TestCase):
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
 
     def test_training_with_reward_func_accessing_trainer_state(self):
-        # Test if trainer can handle reward function that accesses the trainer state
         dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
 
         def reward_func(completions, **kwargs):
-            # Assert that trainer_state is passed and has the expected attributes
             trainer_state = kwargs.get("trainer_state")
             assert trainer_state is not None
+            # transformers.TrainerState instance should have a `global_step` property.
             assert hasattr(trainer_state, "global_step")
             return [float(len(set(completion))) for completion in completions]
 
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1271,3 +1271,29 @@ class GRPOTrainerTester(unittest.TestCase):
             for n, param in previous_trainable_params.items():
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
+
+    def test_training_with_reward_func_accessing_trainer_state(self):
+        # Test if trainer can handle reward function that accesses the trainer state
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        def reward_func(completions, trainer_state, **kwargs):
+            # Assert that trainer_state is passed and has the expected attributes
+            assert trainer_state is not None
+            assert hasattr(trainer_state, "global_step")
+            return [float(len(set(completion))) for completion in completions]
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            training_args = GRPOConfig(
+                output_dir=tmp_dir,
+                per_device_train_batch_size=2,
+                num_generations=2,
+                max_completion_length=8,
+                report_to="none",
+            )
+            trainer = GRPOTrainer(
+                model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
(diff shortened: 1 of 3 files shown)


tests/test_grpo_trainer.py
```diff
@@ -1271,3 +1271,29 @@ def test_training_with_generation_kwargs(self):
...
+
+    def test_training_with_reward_func_accessing_trainer_state(self):
+        # Test if trainer can handle reward function that accesses the trainer state
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        def reward_func(completions, trainer_state, **kwargs):
+            # Assert that trainer_state is passed and has the expected attributes
+            assert trainer_state is not None
+            assert hasattr(trainer_state, "global_step")
+            return [float(len(set(completion))) for completion in completions]
```
kashif:
```suggestion
        def reward_func(completions, **kwargs):
            # Assert that trainer_state is passed and has the expected attributes
            trainer_state = kwargs.get("trainer_state")
            assert trainer_state is not None
            assert hasattr(trainer_state, "global_step")
            return [float(len(set(completion))) for completion in completions]
```

kashif:
small nit with the test.. but seemslike a useful thing to have!

tests/test_grpo_trainer.py
```diff
@@ -1271,3 +1271,30 @@ def test_training_with_generation_kwargs(self):
...
+    def test_training_with_reward_func_accessing_trainer_state(self):
+        # Test if trainer can handle reward function that accesses the trainer state
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        def reward_func(completions, **kwargs):
+            # Assert that trainer_state is passed and has the expected attributes
+            trainer_state = kwargs.get("trainer_state")
+            assert trainer_state is not None
+            assert hasattr(trainer_state, "global_step")
+            return [float(len(set(completion))) for completion in completions]
```
seungduk-yanolja:
```suggestion
    def test_training_with_reward_func_accessing_trainer_state(self):
        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")

        def reward_func(completions, **kwargs):
            trainer_state = kwargs.get("trainer_state")
            assert trainer_state is not None
            # transformers.TrainerState instance should have a `global_step` property.
            assert hasattr(trainer_state, "global_step")
            return [float(len(set(completion))) for completion in completions]
```

trl/trainer/grpo_trainer.py
```diff
@@ -346,6 +346,10 @@ def reward_func(completions, **kwargs):
                   reward function returns None for a sample, that reward function is excluded from the reward
                   calculation for that sample. For more details, see [Using a custom reward
                   function](#using-a-custom-reward-function).
+
+                  The trainer's state is also passed to the reward function. The trainer's state is an
+                  instance of [`transformers.TrainerState`](https://huggingface.co/docs/transformers/main/main_classes/callback#transformers.TrainerState) and can be accessed by adding a `trainer_state` argument
```
```suggestion
                  instance of [`transformers.TrainerState`](https://huggingface.co/docs/transformers/main/main_classes/callback#transformers.TrainerState)
                  and can be accessed by adding a `trainer_state` argument
```

trl/trainer/grpo_trainer.py
```diff
@@ -346,6 +346,11 @@ def reward_func(completions, **kwargs):
                   reward function returns None for a sample, that reward function is excluded from the reward
                   calculation for that sample. For more details, see [Using a custom reward
                   function](#using-a-custom-reward-function).
+
+                  The trainer's state is also passed to the reward function. The trainer's state is an
+                  instance of [`transformers.TrainerState`](https://huggingface.co/docs/transformers/main/main_classes/callback#transformers.TrainerState)
+                  and can be accessed by adding a `trainer_state` argument
```
```suggestion
                  and can be accessed by accessing the `trainer_state` argument
```

seungduk-yanolja:
Does it require approval whenever I create a merge commit?

PR #3317, opened 2025-04-17
✌️ Add support for FSDP2

lewtun:
# What does this PR do?

This PR adds support for FSDP2 by (a) adding an `accelerate` config and (b) updating `prepare_fsdp()`.

Useful migration guide: https://huggingface.co/docs/accelerate/main/en/concept_guides/fsdp1_vs_fsdp2

The FSDP2 API is still in beta, so we can hold off on merging this if we feel the `accelerate` version is too high to bump right now. The alternative would be to check the accelerate version being used and only allow FSDP2 to be run for versions > 1.6.0 

~In particular, `FULL_STATE_DICT` is not supported in FSDP2 yet, so it is not possible to save `transformers` friendly checkpoints. This is currently being worked on in `accelerate`, so we could hold off until then~

Update: now supported on `accelerate@main` and tested with:

```shell
uv pip install accelerate @ git+https://github.com/huggingface/accelerate.git@c5caa11e8557633ba2187a84931b78cc25098c05
```

A quick test on SFT shows the two FSDP versions are almost identical, modulo grad norms:

Screenshot 2025-04-17 at 16 03 02

## Commands to test with on 2 GPUs

```shell
accelerate launch --config_file examples/accelerate_configs/fsdp2.yaml --num_processes 2  trl/scripts/sft.py \
    --model_name_or_path Qwen/Qwen2-0.5B \
    --dataset_name trl-lib/Capybara \
    --learning_rate 2.0e-5 \
    --num_train_epochs 1 \
    --packing \
    --per_device_train_batch_size 2 \
    --gradient_accumulation_steps 8 \
    --gradient_checkpointing \
    --eos_token '<|im_end|>' \
    --logging_steps 25 \
    --eval_strategy steps \
    --eval_steps 100 \
    --output_dir Qwen2-0.5B-SFT

# LoRA
accelerate launch --config_file examples/accelerate_configs/fsdp2.yaml --num_processes 2 trl/scripts/sft.py \
    --model_name_or_path Qwen/Qwen2-0.5B \
    --dataset_name trl-lib/Capybara \
    --learning_rate 2.0e-4 \
    --num_train_epochs 1 \
    --packing \
    --per_device_train_batch_size 2 \
    --gradient_accumulation_steps 8 \
    --gradient_checkpointing \
    --eos_token '<|im_end|>' \
    --logging_steps 25 \
    --eval_strategy steps \
    --eval_steps 100 \
    --use_peft \
    --lora_r 32 \
    --lora_alpha 16 \
[...]

PR #3670, opened 2025-06-30
Support FSDP2 in GRPOTrainer

thepowerfuldeez:
Following the idea from https://github.com/huggingface/trl/pull/3317#issuecomment-2848259987 I updated _sync_fsdp_params_to_vllm with FSDP2 summon code


thepowerfuldeez:
I'm not strong at FSDP2 so this impl might be incomplete

But without this addition previously params were never synced from vllm to trl code, making the model never train with FSDP2

kashif:
thanks @thepowerfuldeez   If its ok with you, we are going to use the #3687 instead of this PR

thepowerfuldeez:
@kashif no worries!

kashif:
thank you!

PR #3682, opened 2025-07-02
fix wrong logit slicing in grpo _get_per_token_logps_and_entropies

### LINKED ISSUE #3681: _get_per_token_logps_and_entropies in GRPO implements the wrong logit slicing
### Reproduction

Simply as the title suggests, [_get_per_token_logps_and_entropies](https://github.com/huggingface/trl/blob/e04f7eb3b993b5278733a793853ed3a85d1c69c8/trl/trainer/grpo_trainer.py#L867) incorrectly assumes that the sequence length of the `logits` tensor should just be reduced by one. However, what's needed is to select only the `logits` that correspond to the `completion_ids`. Since the `logits` for the entire `prompt + completion` sequence are returned, we need to slice out only the part that aligns with the completions.

### System Info

```
vllm=0.9.1
trl[vllm]=0.19.0
transformers=4.52.3
torch=2.6.0
```

### Checklist

ahatamiz:
# What does this PR do?

[_get_per_token_logps_and_entropies](https://github.com/huggingface/trl/blob/e04f7eb3b993b5278733a793853ed3a85d1c69c8/trl/trainer/grpo_trainer.py#L867) incorrectly assumes that the sequence length of the `logits` tensor should just be reduced by one. However, what's needed is to select only the `logits` that correspond to the `completion_ids`. Since the `logits` for the entire `prompt + completion` sequence are returned, we need to slice out only the part that aligns with the completions.

Fixes #3681.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review? 

@shirinyamani @qgallouedec

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -881,7 +881,9 @@ class GRPOTrainer(Trainer):
                 attention_mask=attention_mask_batch,
                 logits_to_keep=logits_to_keep + 1,
             ).logits
-            logits = logits[:, :-1, :]  # (B, L-1, V), exclude the last logit: it corresponds to the next token pred
+            # Correctly slice the logits to only include the parts corresponding to the completions.
+            # The logits for the completion tokens are at the end of the sequence.
+            logits = logits[:, -logits_to_keep-1:-1, :]
             # Divide logits by sampling temperature.
             # See https://huggingface.co/blog/the_n_implementation_details_of_rlhf_with_ppo#policy-training-implementation-details
             logits = logits / self.temperature
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -883,7 +883,7 @@ class GRPOTrainer(Trainer):
             ).logits
             # Correctly slice the logits to only include the parts corresponding to the completions.
             # The logits for the completion tokens are at the end of the sequence.
-            logits = logits[:, -logits_to_keep-1:-1, :]
+            logits = logits[:, -logits_to_keep - 1 : -1, :]
             # Divide logits by sampling temperature.
             # See https://huggingface.co/blog/the_n_implementation_details_of_rlhf_with_ppo#policy-training-implementation-details
             logits = logits / self.temperature


shirinyamani (commented):
Hi @ahatamiz jan, thanks for your catch! 
I'm testing this with the debugger, this certainly makes sense to me. But one question; 
prior to defining the logits as `logits = logits[:, :-1, :] (B, T-1, vocab_size)` in below, we are getting the logits while setting the  `logits_to_keep=logits_to_keep + 1,`! and here the logits_to_keep is only on completions, isnt? 
```python
            # We add 1 to `logits_to_keep` because the last logits of the sequence is later excluded
            logits = model(
                input_ids=input_ids_batch,
                attention_mask=attention_mask_batch,
                logits_to_keep=logits_to_keep + 1,
            ).logits
            logits = logits[:, :-1, :]  # (B, L-1, V), exclude the last logit: it corresponds to the next token pred
            # Divide logits by sampling temperature.
```
So what im trying to say is don't you think that the logits we have is already on completions only based on the pasted code or am i wrong?

ahatamiz:
Hi @shirinyamani ! thanks for raising this excellent point ! 

The reason behind this PR is that I ran into some mismatch when using [Nemotron-H-8B-Reasoning-128K](https://huggingface.co/nvidia/Nemotron-H-8B-Reasoning-128K). Probably you can just test it out with this model or any non-Qwen model, and see the mismatch and hence the bug. 

The issue is that  standard Hugging Face `AutoModelForCausalLM` does not have a `logits_to_keep` parameter in its `forward` method.

When you pass an unrecognized keyword argument like this, Python's `**kwargs` functionality simply allows the model's `forward` method to accept and then ignore it. The model proceeds as it normally would: it processes the entire input sequence (`input_ids_batch`) and returns the logits for that full sequence.

The solution I provided (`logits = logits[:, -logits_to_keep-1:-1, :]`) manually performs the slicing that the `logits_to_keep` argument was intended to trigger automatically. It correctly aligns the `logits` tensor with the `input_ids_batch` by selecting only the logits corresponding to the completion tokens.

I have already trained a model with this new change and validated the results !

shirinyamani (approved):
@ahatamiz Hi Ali Jan, 
I've just tested! you are right! and thank you so much for the explanation! Since majority of our tests use Qwen we did not catch it. But my question is why this does not happen with Qwen family of model?

shirinyamani (commented):
Hi Ali Jan @ahatamiz, 

a couple of notes, 
I've just investigated abit abt the root cause of this; 

My main source of investigation is the standard models on **transformers** lib including the non-qwen family of models. Our investigation flow was to look for `logits_to_keep` in the forward pass, and if the models happen to have this arg in their forward pass then next step is to look into the slicing strategy!

It turned out that for famous non-Qwen models like [llama](https://github.com/huggingface/transformers/blob/7d9e52f376ad4b351ae696b0a62280cb9c63f70b/src/transformers/models/llama/modeling_llama.py#L521C9-L521C23) or for [deepseek-v3](https://github.com/huggingface/transformers/blob/7d9e52f376ad4b351ae696b0a62280cb9c63f70b/src/transformers/models/deepseek_v3/modeling_deepseek_v3.py#L709) we have the `logits_to_keep` arg followed by the same slicing approach, e.g. [here](https://github.com/huggingface/transformers/blob/7d9e52f376ad4b351ae696b0a62280cb9c63f70b/src/transformers/models/deepseek_v3/modeling_deepseek_v3.py#L755)!

In contract, when I investigated your customized mentioned [Nemotron](https://huggingface.co/nvidia/Nemotron-H-8B-Reasoning-128K/blob/main/modeling_nemotron_h.py) model I realized that this particular model do not have `logits_to_keep` arg and hence the bug!

Therefore, I did discuss this with my team n after advising with team, we wanted to kindly advice you to update the model to match the standard practice of models on transformer lib which technically include the `logits_to_keep` in the forward pass, also we would greatly appreciate if you mention the other models (except the Nemotron) you came across who are missing this arg?

Always happy to discuss/learn more from you!

ahatamiz:
Hi @shirinyamani 

Thank you for diving into this! I’ve only observed this behavior in the Nemotron family so far. I'd let the team know about this, but there's probably no guarantee they would change it for this release as it may be a non-issue for their workflow.  That said, this PR addresses the root cause regardless of whether a model preserves logits_to_keep in its forward pass. I believe it is difficult to enforce this uniformly across many model variants. 

The change itself is minimal yet robust, ensuring that once merged, this issue won’t resurface. It’s a low-risk update with high confidence in preventing future regressions.

PR #3687, opened 2025-07-03
👐 FSDP2+GRPO

SalmanMohammadi:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/grpo_trainer.py
```diff
@@ -901,6 +900,17 @@ def _get_per_token_logps_and_entropies(
...
+            if param.is_cpu:
+                param = param.to(torch.device("cuda"))
+            param = param.full_tensor()
+            
+            if self.vllm_mode == "server" and self.accelerator.is_main_process:
+                self.vllm_client.update_named_param(name, param)
+            elif self.vllm_mode == "colocate":
+                llm_model = self.llm.llm_engine.model_executor.driver_worker.model_runner.model
+                llm_model.load_weights([(name, param)])
+
```
winglian, line 1140: Does this get duplicated and sent twice with the now L934-938?

salmanmohammadi: I think you're right - this should be the only logic for syncing params, and I believe we'll be dropping support for FSDP1, right? @kashif

kashif: yes we are dropping support for FSDP1

qgallouedec (approved):
This won't work for peft, but we can do that for the next patch release

fabianlim:
@qgallouedec one possible issue I can see with calling `state_dict` is that it wont scale for large models? Have we tried this for large models like 70b? cc: @toslali-ibm

salmanmohammadi:
> @qgallouedec one possible issue I can see with calling `state_dict` is that it wont scale for large models? Have we tried this for large models like 70b? cc: @toslali-ibm

In FSDP2, `state_dict` doesn't gather the full state dict, this only happens per-parameter when you convert the parameter's `DTensor` to a full local tensor.

fabianlim:
@SalmanMohammadi I see thanks for clarifying. So you are saying that the gathering only happens when you do `param = param.full_tensor()`

Kirill-Kravtsov:
Doesn't the `prepare_fsdp` function require more changes, since FSDP2 has a completely different API compared to FSDP1?

PR #3649, opened 2025-06-25
Faster `position_ids` computation for FFD packing

mariosasko:
# What does this PR do?

This PR removes the slow conversion from PyArrow to Python needed to compute `position_ids` for FFD packing, which was introduced in https://github.com/huggingface/trl/pull/3526. To do this, it uses `seq_lengths` that save disk space and can be efficiently converted to `position_ids` in a vectorized manner in the SFT collator.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #3699, opened 2025-07-06
[SFT] Add `seq_lengths` to signature columns

LeonEricsson:
## What does this PR do?

To avoid cross-contamination when training with packed sequences, we pass `position_ids` to FA2. This was introduced in PR #3526, which calculated `position_ids` during packing and added them to the dataset. To ensure these are retained, `position_ids` must be set as a signature column; otherwise, the `Trainer` will discard them as unused.

https://github.com/huggingface/transformers/blob/ca7e1a3756c022bf31429c452b2f313f043f32de/src/transformers/trainer.py#L1007

While this was implemented correctly, PR #3649 introduced an optimized approach where `position_ids` are computed within the data collator using a new dataset column, `seq_lengths`. However, the latter PR failed to update the signature column from `position_ids` to `seq_lengths`. As a result, `seq_lengths` is not passed to the collator, causing it to behave as if the sequences are not packed.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -829,7 +829,7 @@ class SFTTrainer(Trainer):
             self._signature_columns = [
                 "input_ids",
                 "labels",
-                "position_ids",
+                "seq_lengths",
                 "completion_mask",
                 "assistant_masks",
             ]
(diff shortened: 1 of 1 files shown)


LeonEricsson:
waiting to merge until we've confirmed this fixes the issues raised in #3705

qgallouedec (approved):
Good!

jiosephlee:
```
if args.use_liger_kernel:
    dataset = dataset.select_columns(
        {"input_ids", "position_ids", "completion_mask"}.intersection(dataset.column_names)
    )
```
This line of code may need to be adjusted as well (edit: nvm just saw the new commit)

kashif:
@jiosephlee  fixed and tested and works thanks!

PR #3700, opened 2025-07-07
[GRPO] Log generation entropy

### LINKED ISSUE #3571: [GRPO] Entropy metric
### Feature request

Entropy metrics for the generation during training, averaged across the sequence.

### Motivation

This has been up for discussion previously, particularly after the release of the [DAPO paper](https://arxiv.org/pdf/2503.14476) but was never implemented. Policy entropy is a real-time barometer of exploration capacity. Logging entropy gives practitioners an early-warning gauge of whether downstream performance is plateauing. 

Recent RLVR work makes a strong case for tracking entropy.

1. [The Entropy Mechanism of Reinforcement Learning for Reasoning Language Models](https://arxiv.org/pdf/2505.22617)
2. [ProRL: Prolonged Reinforcement Learning Expands Reasoning Boundaries in Large Language Models](https://arxiv.org/pdf/2505.24864)
3. [Beyond the 80/20 Rule: High-Entropy Minority Tokens Drive Effective Reinforcement Learning for LLM Reasoning](https://arxiv.org/pdf/2506.01939)
4. [The Surprising Effectiveness of Negative Reinforcement](https://arxiv.org/pdf/2506.01347)

### Your contribution

Will PR

LeonEricsson:
# What does this PR do?

Log the average per-token entropy of the generations/completions (normalized over the local batch, BNPO-style). Since entropy is not computed by default, you need to set `log_entropy` in the config. However, given the importance of this metric, I'm open to the idea of always computing entropy–provided that it doesn't introduce noticeable overhead.

Fixes #3571

This PR also includes a minor refactoring of the _compute_loss function.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


LeonEricsson:
Test run comparing the compute cost of calculating entropy. Currently, I'm only using a batch size of 12 and a maximum completion length of 400.

W B Chart 07_07_2025, 13_25_13

trl/trainer/grpo_trainer.py
```diff
@@ -1430,35 +1431,67 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
         logits_to_keep = completion_ids.size(1)  # we only need to compute the logits for the completion tokens
 
-        # Compute the entropy at each position in the completion
-        if self.token_entropy_percentile_threshold > 0.0:
-            logps_and_entropies = self._get_per_token_logps_and_entropies(
-                model, input_ids, attention_mask, logits_to_keep, compute_entropy=True
+        entropies = None
+        entropy_mask = None
+
+        compute_entropy = self.log_entropy or self.token_entropy_percentile_threshold > 0.0
```
pramodith: Shouldn't we be using `torch.no_grad()`, if we're computing entropy just for logging purposes?

LeonEricsson: yes!

trl/trainer/grpo_trainer.py
```diff
@@ -1430,35 +1431,64 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+        if compute_entropy_mask:
             entropy_mask = self._compute_entropy_mask(entropies, completion_mask)
-        else:
-            per_token_logps = self._get_per_token_logps_and_entropies(
-                model, input_ids, attention_mask, logits_to_keep
-            )["logps"]
-            entropy_mask = None
+
+        if self.log_entropy and entropies is not None:
+            with torch.no_grad():
```
pramodith: I think it'd be more efficient if the `torch.no_grad()` context is here https://github.com/huggingface/trl/blob/640a9f39164ce3f9bbac7d325c1149ba023314e6/trl/trainer/grpo_trainer.py#L905-L907. It should theoretically be faster and have fewer memory ops.

LeonEricsson: Yeah, I was considering this but didn’t want to complicate things with `_get_per_token_logps_and_entropies`. I’ll take another look and see if I can find a clean way to incorporate it. Either way, I’m leaning toward always logging entropy instead of making it configurable, which will probably change things.

trl/trainer/grpo_trainer.py
```diff
@@ -927,7 +927,8 @@ def _get_per_token_logps_and_entropies(
             all_logps.append(logps)
 
             if compute_entropy:
-                entropies = entropy_from_logits(logits)
+                with torch.no_grad():
+                    entropies = entropy_from_logits(logits)
```
LeonEricsson, line 1099: the entropies, as they are used by _compute_loss (masking and/or logging), don't require gradients. However, this may not be true for future use cases. Is this too naive?

@pramodith

LeonEricsson:
Updated PR with new changes on main

trl/trainer/grpo_trainer.py
pramodith: I agree that the entropies don't need gradients, but when we compute the entropies for the entropy masking loss, the logits are involved in ops and I think these ops should be considered in the computational graph for gradient calculation. So I do think we need to differentiate between when we compute entropy for logging vs masking.

trl/trainer/grpo_trainer.py
```diff
@@ -1537,6 +1538,9 @@ def _compute_loss(self, model, inputs):
             mean_kl = (per_token_kl * completion_mask).sum() / completion_mask.sum()
             self._metrics[mode]["kl"].append(self.accelerator.gather(mean_kl).nanmean().item())
 
+        mean_entropy = (entropies * completion_mask).sum() / completion_mask.sum()
```
pramodith: I think we can run into an edge case here where completion_mask.sum() is 0, when someone uses `mask_truncated_completions`all of the logging steps should probably include a small offset in the denominator, like (completion_mask.sum + 1e-8) to prevent a _nan_ from popping up.

trl/trainer/grpo_trainer.py
```diff
@@ -903,7 +903,7 @@ def _get_last_hidden_state(self, unwrapped_model, input_ids, attention_mask, log
     def _get_per_token_logps_and_entropies(
         self, model, input_ids, attention_mask, logits_to_keep, batch_size=None, compute_entropy=False
     ) -> dict[str, Optional[torch.Tensor]]:
-        """Compute log‐probs and (optionally) entropies for each token."""
+        """Compute log-probs and (optionally) entropies for each token."""
```
pramodith, line 1060: Wondering if there's any benefit to having the `compute_entropy` bool, seems like it'd always be True and we don't seem to be providing an option to the user to set it.

pramodith (commented):
Thanks for working on this! Had a few questions.

qgallouedec:
Thanks! Reviewing asap

trl/trainer/grpo_trainer.py
LeonEricsson: these calls for example have `compute_entropy=False`

https://github.com/huggingface/trl/blob/56f4201db6973b82cfe55efab9ac7e1d64b2677a/trl/trainer/grpo_trainer.py#L1560-L1597

trl/trainer/grpo_trainer.py
LeonEricsson: take a look now and let me know what you think

trl/trainer/grpo_trainer.py
pramodith: My bad, I realised this a few hours later.

trl/trainer/grpo_trainer.py
pramodith: Looks great!

trl/trainer/grpo_trainer.py
LeonEricsson: Since we only use entropy to mask tokens (a hard 0/1 gates) there’s no gradient path through that thresholding anyway, masking is non differentiable. So wrapping in torch.no_grad() should be safe for both masking and logging.

pramodith (approved):
LGTM for what its worth 😅

PR #3710, opened 2025-07-09
Remove the negative value of KL divergence

ENg-122:
# What does this PR do?
Regarding the dangers of negative KL divergence mentioned in https://huggingface.co/docs/trl/main/en/how_to_train, why not remove the negative cases and add a warning prompt?

Fixes # (issue)

## Before submitting
- [×] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [√] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [×] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [√] Did you make sure to update the documentation with your changes?
- [×] Did you write any new necessary tests?

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1461,7 +1461,8 @@ class GRPOTrainer(Trainer):
                 torch.exp(ref_per_token_logps - per_token_logps) - (ref_per_token_logps - per_token_logps) - 1
             )
             per_token_kl = torch.clamp(per_token_kl, min=0.0)
-            warnings.warn(f"KL divergence value is 0")
+            if (per_token_kl == 0.0).all():
+                warnings.warn(f"KL divergence value is 0")
 
         # Compute the loss
         advantages = inputs["advantages"]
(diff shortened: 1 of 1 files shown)


shirinyamani:
Hello @ENg-122 I understand the motivation and why you would want to not have negative kl, but im thinking this is also a sing that if we get negative kl then it would mean sth is not making sense or there is bug, versus if we set the min to zero in `torch.clamp` then we would never know abt a potential bug, does that make sense?

hjh0119:
The KL K3 approximation is theoretically always greater than or equal to 0, so applying a clamp operation is redundant 
```
per_token_kl = (
                torch.exp(ref_per_token_logps - per_token_logps) - (ref_per_token_logps - per_token_logps) - 1)
```

In summary, exp(x) >= x + 1 for all x, so per_token_kl is always non-negative and an additional clamp is unnecessary.

qgallouedec:
I think @hjh0119 is right. The only case where you might see negative value is because of numerical approximation, but never seen in practice. I'm closing this PR for now. If you have an example of setting  that hit such case.

PR #3723, opened 2025-07-11
📥 Set environment variables for vLLM distributed training in GRPOTrainer

qgallouedec:
Late review for #3638

If you've also time to review @toslali-ibm 

```python

train_dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")

def main():
    training_args = GRPOConfig(
        max_completion_length=8,
        per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
        use_vllm=True,
        vllm_mode="colocate",
        report_to="none",
    )

    trainer = GRPOTrainer(
        model="Qwen/Qwen3-0.6B",
        args=training_args,
        train_dataset=train_dataset,
    )

    trainer.train()

if __name__ == "__main__":
    main()
```

this now work, tested with 2 GPUs

### DIFF
--- a/docs/source/grpo_trainer.md
+++ b/docs/source/grpo_trainer.md
@@ -223,17 +223,6 @@ training_args = GRPOConfig(
     vllm_mode="colocate",
 )
 ```
-Note that in this mode, if you want to run with an accelerate config file of your choice (e.g. [accelerate-configs](https://github.com/huggingface/trl/tree/main/examples/accelerate_configs)), you need to set some environment variable before launching the training script:
-
-```bash
-#required environment variables for vLLM colocate mode on a single node single GPU
-os.environ["RANK"] = "0"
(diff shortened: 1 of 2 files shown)


kashif:
thanks! I was adding them to the slow_grpo_test.py and now i can remove that when merged

kashif (approved):

kashif:
@qgallouedec  perhaps in the docs we can say that we use these defaults and one can change it by setting the env variables if needed?

qgallouedec:
> @qgallouedec perhaps in the docs we can say that we use these defaults and one can change it by setting the env variables if needed?

👍
Done in [d48ad8f](https://github.com/huggingface/trl/pull/3723/commits/d48ad8f47fac6eba5a65919bea9600481ae90a91)

PR #3725, opened 2025-07-12
🔧 Fix GRPO sampling logic

qgallouedec:
~No idea how we ended up with this erroneous sampling logic, but after debugging in depth, I realize that GRPO is incorrect in several places. This PR fixes those problems.~

EDIT: I was initially wrong, but there are still a couple of errors to fix

If it can help the review:
Untitled-2025-07-08-1423

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -828,7 +828,7 @@ class GRPOTrainer(Trainer):
             data_source=dataset,
             mini_repeat_count=self.num_generations,
             batch_size=self.args.generation_batch_size // self.num_generations,
-            repeat_count=self.num_iterations,
+            repeat_count=self.num_iterations * self.args.steps_per_generation,
             shuffle=self.shuffle_dataset,
             seed=self.args.seed,
         )
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1332,3 +1332,55 @@ class GRPOTrainerTester(unittest.TestCase):
         entropy_mask = trainer._compute_entropy_mask(entropies, completion_mask)
 
         self.assertTrue(torch.equal(entropy_mask, expected_mask))
+
+    def test_prepare_input_called_with_correct_data(self):
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            training_args = GRPOConfig(
+                output_dir=tmp_dir,
+                learning_rate=0.1,  # increase the learning rate to speed up the test
+                max_completion_length=8,  # reduce the completion length to reduce memory usage
+                gradient_accumulation_steps=3,  # can be anything in this test
+                # steps_per_generation*per_device_train_batch_size=24 is divisible by num_generations=4
+                steps_per_generation=4,
+                num_generations=4,
+                per_device_train_batch_size=6,  # reduce the batch size to reduce memory usage
+                num_iterations=2,
+                shuffle_dataset=False,
+                report_to="none",
+            )
+            trainer = GRPOTrainer(
+                model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
+                reward_funcs="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
+                args=training_args,
+                train_dataset=dataset,
+            )
+            # steps_per_generation=4, per_device_train_batch_size=6 and num_generations=4, so we expect a
+            # generation batch of 24 samples (steps_per_generation * per_device_train_batch_size), containing 6
+            # different prompts (steps_per_generation * per_device_train_batch_size // num_generations), each repeated
+            # 4 times (num_generations).
+            expected_first_generation_batch = (
+                [{"prompt": "Beautiful is better than"}] * 4
+                + [{"prompt": "Explicit is"}] * 4
+                + [{"prompt": "Simple is better"}] * 4
+                + [{"prompt": "Complex"}] * 4
+                + [{"prompt": "Flat is better than"}] * 4
+                + [{"prompt": "Sparse is better"}] * 4
+            )
+            expected_second_generation_batch = (
+                [{"prompt": "Readability"}] * 4
+                + [{"prompt": "Special cases aren't special"}] * 4
+                + [{"prompt": "Although practicality beats"}] * 4
+                + [{"prompt": "Errors should never"}] * 4
+                + [{"prompt": "Unless explicitly"}] * 4
+                + [{"prompt": "In the face of ambiguity, refuse"}] * 4
+            )
+
+            with patch.object(GRPOTrainer, "training_step", wraps=trainer.training_step) as mock_prepare:
+                trainer.train()
+                # 3 epochs * 2 iterations * 2 generation batches to cover the dataset * 4 steps_per_generation
+                self.assertEqual(mock_prepare.call_count, 48)
+                for i in range(0, 8):  # Generation batch repeated 8 times (steps_per_generation*num_iterations)
+                    assert mock_prepare.call_args_list[i].args[1] == expected_first_generation_batch
+                for i in range(8, 16):
+                    assert mock_prepare.call_args_list[i].args[1] == expected_second_generation_batch


trl/trainer/grpo_config.py
```diff
@@ -67,9 +67,11 @@ class GRPOConfig(TrainingArguments):
 
         generation_batch_size: (`int` or `None`, *optional*, defaults to `None`):
             Batch size to use for generation. If `None`, it defaults to the effective training batch size:
-            `per_device_train_batch_size * num_processes * steps_per_generation`.
-        steps_per_generations: (`int` or `None`, *optional*, defaults to `None`):
-            Number of optimization steps per generation. If `None`, it defaults to gradient_accumulation_steps.
+            `per_device_train_batch_size * num_processes * steps_per_generation`. In other words, there is one
+            generation batch processed per optimization step. Mutually exclusive with `steps_per_generation`.
+        steps_per_generation: (`int` or `None`, *optional*, defaults to `None`):
```
qgallouedec, line 72:
```diff
-         steps_per_generations: (`int` or `None`, *optional*, defaults to `None`):
+         steps_per_generation: (`int` or `None`, *optional*, defaults to `None`):
```

trl/trainer/grpo_config.py
```diff
@@ -67,9 +67,11 @@ class GRPOConfig(TrainingArguments):
...
+            Number of steps per generation. If `None`, it defaults to `gradient_accumulation_steps`. Mutually exclusive
```
qgallouedec, line 73: It's not the number of opt step, but the number of steps.

trl/trainer/grpo_config.py
```diff
@@ -555,53 +555,30 @@ def __post_init__(self):
...
 
-        self.steps_per_generation = self.generation_batch_size // (self.per_device_train_batch_size * num_processes)
-
         # Check if the effective batch size can be divided by the number of generations
         if self.num_generations < 2:
             raise ValueError(
                 "GRPO requires at least 2 generations per prompt to calculate the advantages. You provided "
                 f"{self.num_generations}, which is less than the minimum required."
             )
-        possible_values = [
```
qgallouedec, line 588: The checks below aren't relevant anymore. The only test we need is 

```python
if self.generation_batch_size % (self.per_device_train_batch_size * num_processes) != 0:
    raise ValueError(
        f"generation_batch_size ({self.generation_batch_size}) must be divisible by the global batch size "
        f"({self.per_device_train_batch_size * num_processes})."
    )
```

trl/trainer/grpo_trainer.py
```diff
@@ -179,7 +179,7 @@ def __iter__(self):
                         yield index
 
     def __len__(self) -> int:
-        return self.num_samples * self.mini_repeat_count * self.repeat_count
+        return (self.num_samples // self.batch_size) * self.batch_size * self.mini_repeat_count * self.repeat_count
```
qgallouedec, line 182: Because of this (just above):


```python
        #    [[2, 4, 3], [1, 0, 6], [5]]
        # -> [[2, 4, 3], [1, 0, 6]]
        indexes = [chunk for chunk in indexes if len(chunk) == self.batch_size]
```

trl/trainer/grpo_trainer.py
```diff
@@ -617,14 +617,12 @@ def __init__(
...
+        # Keep logs sized to the generation batch to record only outputs from the latest model update.
         self._textual_logs = {
-            "prompt": deque(maxlen=maxlen),
-            "completion": deque(maxlen=maxlen),
-            "rewards": defaultdict(lambda: deque(maxlen=maxlen)),
-            "advantages": deque(maxlen=maxlen),
+            "prompt": deque(maxlen=args.generation_batch_size),
+            "completion": deque(maxlen=args.generation_batch_size),
+            "rewards": defaultdict(lambda: deque(maxlen=args.generation_batch_size)),
+            "advantages": deque(maxlen=args.generation_batch_size),
```
qgallouedec, line 625: Equivalent but simplify by using `generation_batch_size` directly

trl/trainer/grpo_trainer.py
```diff
@@ -819,7 +817,7 @@ def _get_train_sampler(self, dataset: Optional[Dataset] = None) -> Sampler:
             data_source=dataset,
             mini_repeat_count=self.num_generations,
             batch_size=self.args.generation_batch_size // self.num_generations,
-            repeat_count=self.num_iterations * self.args.steps_per_generation,
+            repeat_count=self.num_iterations,
```
qgallouedec: I think this is a big one. No idea why we did it in the first place. There is no need to repeat the sampling `self.args.steps_per_generation` times

trl/trainer/grpo_trainer.py
```diff
@@ -1274,10 +1272,13 @@ def _generate_and_score_completions(
...
-            # old_per_token_logps == per_token_logps, so we can skip it's computation here, and use
-            # per_token_logps.detach() instead.
-            if self.num_iterations > 1 or self.args.steps_per_generation > self.args.gradient_accumulation_steps:
+            # If the generation and optimization steps are misaligned—i.e., if generation does not occur at the end of
+            # a full optimizer step (when gradient_accumulation_steps is not a multiple of generate_every)—then the
+            # samples may come from an earlier version of the model. In that case, we need to track old_per_token_logps
+            # for importance sampling. If the steps are aligned, importance sampling isn't necessary and we set
+            # old_per_token_logps to None.
+            generate_every = self.args.steps_per_generation * self.num_iterations  # generation frequency
+            if self.args.gradient_accumulation_steps % generate_every != 0:
```
qgallouedec, line 1291: Hoping the comment is enough for review

edbeeching:
Would it be possible to write an integration test with dummy data to ensure this is now working as expected in a few different configurations?

trl/trainer/grpo_trainer.py
hjh0119: I think the original logic might be correct, because in `_prepare_inputs`, data is only sampled once every `generate_every` steps. For the remaining `generate_every - 1` steps, we still need to load data. Therefore, multiplying by `self.args.steps_per_generation` here actually helps to avoid skipping any data.

kashif:
i have been using:


```python
    def test_repeat_sampler_length_calculation(self):
        dataset = Dataset.from_list([{"text": f"sample_{i}"} for i in range(10)])

        sampler = RepeatSampler(
            data_source=dataset,
            mini_repeat_count=2,
            batch_size=3,  # 10 samples / 3 batch_size = 3 complete batches, 1 sample dropped
            repeat_count=1,
            shuffle=False,
        )

        actual_samples = list(sampler)

        # With batch_size=3 and 10 samples, only 9 samples form complete batches (1 dropped)
        # So we expect 9 * 2 (mini_repeat_count) * 1 (repeat_count) = 18 samples
        expected_actual_samples = 18

        self.assertEqual(len(actual_samples), expected_actual_samples)
        self.assertEqual(
            len(sampler),
            len(actual_samples),
            f"RepeatSampler.__len__() should match actual iterations. "
            f"Expected {len(actual_samples)}, but got {len(sampler)}",
        )

    def test_grpo_sampling_no_steps_per_generation_multiplier(self):

            config = GRPOConfig(
                output_dir=tmp_dir,
                num_generations=8,
                num_iterations=2,
                steps_per_generation=4,  # Use steps_per_generation without generation_batch_size
                max_steps=1,  # Just test sampler creation
                report_to="none",
            )

            trainer = GRPOTrainer(
                args=config,
                train_dataset=dataset,
            )

            # Get the train sampler
            sampler = trainer._get_train_sampler()

            # repeat_count = num_iterations, NOT num_iterations * steps_per_generation
            self.assertEqual(
                sampler.repeat_count,
                config.num_iterations,
                f"RepeatSampler.repeat_count should not include steps_per_generation multiplier. "
                f"Expected {config.num_iterations}, but got {sampler.repeat_count}",
            )
```

trl/trainer/grpo_trainer.py
edbeeching: This was what I was thinking when I made the change. Although perhaps there were some details of the sampling that I have misunderstood, which introduced the bug.

kashif (approved):

konstantinjdobler:
Hey, after reading through this, I am not entirely clear on which cases were bugged before this PR. Could you give a quick overview which settings caused wrong sampling logic before this PR landed?

trl/trainer/grpo_trainer.py
tangyd: The total number of completion = mini_repeat_count *  batch_size *  repeat_count.
If repeat_count=self.num_iterations * self.args.steps_per_generation, then the total number of completion in each generation will be:
num_generations * generation_batch_size // num_generations * num_iterations * steps_per_generation = generation_batch_size  * num_iterations * args.steps_per_generation = per_device_train_batch_size * num_processes  * num_iterations * steps_per_generation**2. I'm confused if we really need this square.

trl/trainer/grpo_trainer.py
tangyd: I'm still confused. The generate_every is for fixing distribution shift and I totally understand why it times steps_per_generation.
While in generating new samples, we have `batch_size=self.args.generation_batch_size // self.num_generations` and `repeat_count=self.num_iterations * self.args.steps_per_generation`. According to the document, `generation_batch_size = per_device_train_batch_size * num_processes * steps_per_generation`. It seems both parameters in RepeatSampler, batch_size and repeat_count, include `steps_per_generation`. 
Is this what we really want? If yes, could you give more detailed explanation in comment? @qgallouedec

PR #3727, opened 2025-07-12
🏗️ Refactor top-entropy in GRPO

qgallouedec:
# What does this PR do?

Some minor change, late review from https://github.com/huggingface/trl/pull/3563

1. Don't use/unittest private method: `_compute_entropy_mask` -> `get_high_entropy_mask`
2. `token_entropy_percentile_threshold` -> `top_entropy_quantile` top better align with the _top-20%_ message (its more a quantile than a percentile here.
3. Fix the doc: link and var name (rho, not tau)
4. I better like `_get_per_token_logps_and_entropies` returning a tuple, but personal preference

cc @pramodith happy to have you thoughts on this

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -896,7 +896,7 @@ class GRPOTrainerTester(unittest.TestCase):
                 num_generations=3,  # reduce the number of generations to reduce memory usage
                 max_completion_length=8,  # reduce the completion length to reduce memory usage
                 report_to="none",
-                top_entropy_quantile=0.8,
+                top_entropy_quantile=0.2,
             )
             trainer = GRPOTrainer(
                 model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -216,7 +216,7 @@ class RepeatRandomSamplerTester(unittest.TestCase):
 class GetHighEntropyMaskTester(unittest.TestCase):
     def test_compute_entropy_mask_0(self):
         # We have a total of 12 tokens out of which 10 are non-pad.
-        # for a token_entropy_percentile_threshold of 0.8, we expect the top 20% i.e 2 non-pad tokens corresponding to
+        # for a top_entropy_quantile of 0.8, we expect the top 20% i.e 2 non-pad tokens corresponding to
         # the highest entropy to be unmasked.
         # In our example these will be the tokens corresponding to the entropies 0.9 and 1.0 since 1.1 and 1.2 are pad
         # tokens they are excluded from the entropy threshold calculation.
@@ -243,19 +243,19 @@ class GetHighEntropyMaskTester(unittest.TestCase):
         torch.testing.assert_close(entropy_mask, expected_mask)
 
     def test_compute_entropy_mask_all_tokens(self):
-        # For a threshold of 0.0 we expect all non-pad tokens BUT ONE to be unmasked.
+        # For a threshold of 0.0 we expect all non-pad tokens to be unmasked.
         entropies = torch.tensor([[0.1, 0.2, 0.3, 0.4, 0.5, 0.6], [0.7, 0.8, 0.9, 1.0, 1.1, 1.2]])
         mask = torch.tensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 1, 0, 0]])
         entropy_mask = get_high_entropy_mask(entropies, mask, threshold=0.0)
-        expected_mask = torch.tensor([[0, 1, 1, 1, 1, 1], [1, 1, 1, 1, 0, 0]], dtype=torch.bool)
+        expected_mask = torch.tensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 1, 0, 0]], dtype=torch.bool)
         torch.testing.assert_close(entropy_mask, expected_mask)
 
     def test_compute_entropy_mask_no_tokens(self):
-        # If there are no non-pad tokens we expect an empty mask.
+        # If there are no non-pad tokens we expect the mask to be all zeros BUT ONE VALUE.
         entropies = torch.tensor([[0.1, 0.2, 0.3, 0.4, 0.5, 0.6], [0.7, 0.8, 0.9, 1.0, 1.1, 1.2]])
         mask = torch.tensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 1, 0, 0]])
         entropy_mask = get_high_entropy_mask(entropies, mask, threshold=1.0)
-        expected_mask = torch.tensor([[0, 0, 0, 0, 0, 0], [0, 0, 0, 0, 0, 0]], dtype=torch.bool)
+        expected_mask = torch.tensor([[0, 0, 0, 0, 0, 0], [0, 0, 0, 1, 0, 0]], dtype=torch.bool)
         torch.testing.assert_close(entropy_mask, expected_mask)
 
 
@@ -896,7 +896,7 @@ class GRPOTrainerTester(unittest.TestCase):
                 num_generations=3,  # reduce the number of generations to reduce memory usage
                 max_completion_length=8,  # reduce the completion length to reduce memory usage
                 report_to="none",
-                token_entropy_percentile_threshold=0.8,
+                top_entropy_quantile=0.8,
             )
             trainer = GRPOTrainer(
                 model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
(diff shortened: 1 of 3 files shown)


shirinyamani (commented):
Im wondering the default being 1 has better result? cuz it seems like this would mask all but the highest entropy token?
i think cuz actually `top_entropy_quantile = max(entropies of non-padding tokens)`
So this way only the tokens with exactly the highest entropy will be marked `True` ?

trl/trainer/grpo_config.py
```diff
@@ -191,12 +191,12 @@ class GRPOConfig(TrainingArguments):
...
-            τ parameter from the [Beyond the 80/20 Rule](https://huggingface/papers/2506.01939) paper, which finds that
-            masking out the bottom τ percentile of tokens based on the entropy of the probability distribution at a
-            given sequence position, in the policy loss term yields better results. The range of the parameter is
-            [0.0-1.0] a value of 0.0 means that none the tokens are filtered out and 1.0 means that all the tokens are
-            masked. Recommended value is `0.8`.
+        top_entropy_quantile (`float`, *optional*, defaults to `1.0`):
+            ρ parameter from [Beyond the 80/20 Rule](https://huggingface.co/papers/2506.01939). Keeps in the policy
+            loss term only the top-ρ quantile of tokens by entropy of the probability distribution at each sequence
+            position, improving results. Range: `[0.0-1.0]`. A value of `1.0` masks all but the highest entropy token;
+            `0.0` keeps all tokens. The paper recommends a value of `0.2`.
```
qgallouedec:
```suggestion
            position, improving results. Range: `[0.0-1.0]`. A value of `0.0` masks all but the highest entropy token;
            `1.0` keeps all tokens. The paper recommends a value of `0.2`.
```

qgallouedec:
> Im wondering the default being 1 has better result? cuz it seems like this would mask all but the highest entropy token?

Actually it's the opposite, I fixed the doc here:
[5621e1a](https://github.com/huggingface/trl/pull/3727/commits/5621e1aa4fbabb68f834a196b651feefd0dc864a)

pramodith:
Thanks for the changes! I'll take a look later today.

trl/trainer/grpo_trainer.py
```diff
@@ -290,6 +290,30 @@ def identity(x):
...
+        threshold (`float`):
+            Quantile threshold between `0.0` and `1.0` to select high-entropy tokens.
+
+    Returns:
+        `torch.Tensor`:
+            Boolean mask of shape (batch_size, seq_len), where `True` indicates tokens with entropy >= threshold and
+            `False` otherwise.
+    """
+    non_pad_entropies = entropies[mask.bool()].float()
+    entropy_threshold = torch.quantile(non_pad_entropies, threshold)
```
qgallouedec, line 313: In practice, there are cases where the mask can be all-zeros (for example with `mask_truncated_completions=True`). I still need to support this case

trl/trainer/grpo_trainer.py
```diff
@@ -1456,18 +1471,14 @@ def _compute_loss(self, model, inputs):
...
 
-        # Compute the entropy at each position in the completion
-        if self.token_entropy_percentile_threshold > 0.0:
-            logps_and_entropies = self._get_per_token_logps_and_entropies(
-                model, input_ids, attention_mask, logits_to_keep, compute_entropy=True
-            )
-            per_token_logps = logps_and_entropies["logps"]
-            entropies = logps_and_entropies["entropies"]
-            entropy_mask = self._compute_entropy_mask(entropies, completion_mask)
+        # Compute the per_token_logps the entropy if necessary at each position in the completion
```
```suggestion
        # Compute the per_token_logps and the entropy (if necessary) at each position in the completion
```

tests/test_grpo_trainer.py
```diff
@@ -213,6 +213,52 @@ def test_sampler_with_mini_repeat_count_and_batch_size_3(self):
...
+        entropies = torch.tensor([[0.1, 0.2, 0.3, 0.4, 0.5, 0.6], [0.7, 0.8, 0.9, 1.0, 1.1, 1.2]])
+        mask = torch.tensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 1, 0, 0]])
+        entropy_mask = get_high_entropy_mask(entropies, mask, threshold=0.0)
+        expected_mask = torch.tensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 1, 0, 0]], dtype=torch.bool)
+        torch.testing.assert_close(entropy_mask, expected_mask)
+
+    def test_compute_entropy_mask_no_tokens(self):
+        # If there are no non-pad tokens we expect the mask to be all zeros BUT ONE VALUE.
+        entropies = torch.tensor([[0.1, 0.2, 0.3, 0.4, 0.5, 0.6], [0.7, 0.8, 0.9, 1.0, 1.1, 1.2]])
+        mask = torch.tensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 1, 0, 0]])
```
pramodith, line 259: If there are _no non-pad tokens i.e. all the tokens are pad tokens_, shouldn't the mask here be all 1s?

trl/trainer/grpo_trainer.py
pramodith: Ahhh, that was the case I hadn't thought about.

tests/test_grpo_trainer.py
pramodith: I think the intention here was to check the case when the threshold is set to 1.0, so maybe renaming it to `test_compute_entropy_mask_all` might be more appropriate.

trl/trainer/grpo_trainer.py
LeonEricsson: how about an early return
```python
if non_pad_entropies.numel() == 0:
        return torch.zeros_like(entropies, dtype=torch.bool)
```

tests/test_grpo_trainer.py
LeonEricsson: and `expected_mask = torch.tensor([[0, 0, 0, 0, 0, 0], [0, 0, 0, 0, 0, 1]], dtype=torch.bool)`

trl/trainer/grpo_trainer.py
qgallouedec: thanks, done in https://github.com/huggingface/trl/pull/3727/commits/24fc4ab65cfc531b4eb71c44243f9fc970231356

tests/test_grpo_trainer.py
qgallouedec: right, actually the name of the test is not correct, fixed now

LeonEricsson (approved):

tests/test_grpo_trainer.py
```diff
@@ -216,6 +216,60 @@ def test_sampler_with_mini_repeat_count_and_batch_size_3(self):
...
+    def test_compute_entropy_threshold_1(self):
+        # If the threshold is 1.0 then we expect the mask to be all zeros BUT ONE VALUE.
+        entropies = torch.tensor([[0.1, 0.2, 0.3, 0.4, 0.5, 0.6], [0.7, 0.8, 0.9, 1.0, 1.1, 1.2]])
+        mask = torch.tensor([[1, 1, 1, 1, 1, 1], [1, 1, 1, 1, 0, 0]])
+        entropy_mask = get_high_entropy_mask(entropies, mask, threshold=1.0)
+        expected_mask = torch.tensor([[0, 0, 0, 0, 0, 0], [0, 0, 0, 1, 0, 0]], dtype=torch.bool)
+        torch.testing.assert_close(entropy_mask, expected_mask)
+
+    def test_compute_entropy_all_masked(self):
+        # If there are no non-pad tokens we expect the mask to be all zeros BUT ONE VALUE.
```
pramodith: I think the comment needs to be updated to just `be all zeros`

trl/trainer/grpo_trainer.py
```diff
@@ -290,6 +290,32 @@ def identity(x):
     return x
 
 
+def get_high_entropy_mask(entropies: torch.Tensor, mask: torch.Tensor, threshold: float) -> torch.Tensor:
```
pramodith, line 293: Are you going to address the cases of all completion tokens being masked because of truncation and entropy thresholds being inaccurate if training is in a multi-gpu setting in another PR? I can tackle those cases if you're short on time. 😄

pramodith:
The test case corresponding to training a model with the entropy mask needs to be updated to use the new argument `top_entropy_quantile` for the test to pass.

trl/trainer/grpo_trainer.py
qgallouedec: Yes, let's keep PRs as minimal as possible. I'd be very happy to review a subsequent PR from you addressing this issue

tests/test_grpo_trainer.py
```diff
@@ -1301,38 +1355,6 @@ def reward_func(completions, **kwargs):
             )
             trainer.train()
 
-    def test_compute_entropy_mask(self):
-        """Test the _compute_entropy_mask method."""
```
qgallouedec, line 1305: let's avoid testing private methods

Issue #2807, opened 2025-02-08
Please make GRPO support data_collator and hence multimodal LLM


thusinh1969:
### Feature request

I found that **GRPO does NOT support data_collator** which in fact TRL with GRPO can finetune  text-only, which is pitty. 

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[52], line 3
      1 from trl import GRPOTrainer
----> 3 trainer = GRPOTrainer(
      4     model=model,
      5     args=training_args,
      6     train_dataset=train_dataset_formated_no_load,
      7     data_collator=GRPO_qwen2vl_collate_fn,
      8     tokenizer=processor.tokenizer,
      9     optimizers=(optimizer_adamw, lr_scheduler_adamw),
     10     reward_funcs=[format_reward_func, accuracy_reward_func, len_reward_func],
     11 )

TypeError: GRPOTrainer.__init__() got an unexpected keyword argument 'data_collator'

```
Should we expect it any time soon ?

Thanks in advanced,
Steve

### Motivation

We shoiuld be able to finetune GRPO with data_collator (same way PPO does) which TRL can then support multimodal LLM at will.

### Your contribution

Should we expect it any time soon ?

We are at EraX are happy to assist with some coding and testing.

Thanks in advanced,
Steve

Benjoyo:
Multimodal support would be great!

Syazvinski:
+999

davidhughhenrymack:
I am essentially patching this in a local fork so that I can train a vision model

If I made a simple pr to support this just for non-VLLM generation, would that be acceptable? I'm envisioning a sort of minimal viable solution.


I'm envisioning 
- Trainer, accepts collator function argument
- During the prepare and generate function, this is used preferentially to its own applying of chat template and applying processing class
- All of the columns outputted by the collator function are passed into the generate call so that you can have pixel data as well as token ids and attention masks

I need to think a little bit how this interacts with the prompt text generation currently, maybe as a simple interim step we require the collator to provide both tokenized untokenized prompt

Benjoyo:
@davidhughhenrymack I'd say if you have something working, make a PR for sure. Even if it doesn’t get merged, people can use it and tinker with it!

qgallouedec:
The way to go is not to support collations. Instead one should modify the generation part and the compute loss

davidhughhenrymack:
[...]

PR #3731, opened 2025-07-13
add custom data_collator to GRPOTrainer

morrisalp:
# What does this PR do?

Some use cases (including mine) require custom data collation + GRPO. Before GRPOTrainer could only use a default collation method.

Fixes # (issue)

Makes progress towards #2807
* GPRO for multimodal LLMs requires custom data collation
* Does not fully implement it because multimodal generation also requires passing additional modality features during generation.
* PR #3460 is complementary but is specific to standard vision-language data.


morrisalp:
Closing it since I see there are further issues with the generation functionality, and maybe #3460 obviates it.

PR #3740, opened 2025-07-16
[GRPO] Fix: Processing ref logprobs in batches

idanshen:
Until commit d6a969f, ref_per_token_logps was calculated inside _compute_loss or compute_liger_loss, which gets input of size per_device_batch_size. 
Since it moved to _generate_and_score_completions it now need to be passed a batch_size. Without it, memory consumption is not independent of the number of gradient accumulations and can spike when beta != 0


LeonEricsson (approved):
LGTM! Nice catch

PR #3743, opened 2025-07-18
Support dLLM in GRPO reference model creation

### LINKED ISSUE #3742: Bug when creating a reference model in GRPO with deepspeed zero3 for diffusion language models
In line 420 of `trl/trainer/grpo_trainer.py`, 

```python
 # Reference model
        self.beta = args.beta
        if self.beta == 0.0:
            self.ref_model = None
        elif is_deepspeed_zero3_enabled():
```

when deepspeed zero3 is enabled, the reference model is loaded using `AutoModelForCausalLM`. However, recently, there have been diffusion language models like [DiffuCoder](https://github.com/apple/ml-diffucoder). These models need to be loaded using `AutoModel` instead of `AutoModelForCausalLM`. Hence, the code will report an error, stating that there is no `DreamConfig` from `AutoModelForCausalLM`.

xijia-tao:
The current `GRPOTrainer` creates a reference model with `AutoModelForCausalLM` when deepspeed zero3 is enabled. This is not compatible with recent diffusion language models like [DiffuCoder](https://github.com/apple/ml-diffucoder). These models need to be loaded using `AutoModel` instead of `AutoModelForCausalLM`. 

The proposed change is a simple one-liner.

Fixes #3742

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -34,6 +34,7 @@ from torch import nn
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 from torch.utils.data import DataLoader, Sampler
 from transformers import (
+    AutoModel,
     AutoModelForCausalLM,
     AutoModelForSequenceClassification,
     AutoTokenizer,
@@ -584,7 +585,7 @@ class GRPOTrainer(Trainer):
             self.ref_model = None
         else:
             # For deepspeed, fsdp or non-distributed models, create a reference model from scratch
-            self.ref_model = AutoModelForCausalLM.from_pretrained(model_id, **model_init_kwargs)
+            self.ref_model = AutoModel.from_pretrained(model_id, **model_init_kwargs)
 
         # Disable dropout in the models
         if args.disable_dropout:


Tavish9 (changes requested):
Not working here for language model.

AutoModel would not load `lm_head`.

qgallouedec:
would this work instead?

```python
config = AutoConfig.from_pretrained(model_id)
architecture = getattr(transformers, config.architectures[0])
model = architecture.from_pretrained(model_id)
```

Tavish9:
> would this work instead?
> 
> ```python
> config = AutoConfig.from_pretrained(model_id)
> architecture = getattr(transformers, config.architectures[0])
> model = architecture.from_pretrained(model_id)
> ```

Not working for transformers un-registered model.

`AutoModelForCausalLM` can load local model if `MyLocalModelForCausalLM.register_for_auto_class("AutoModelForCausalLM")`

qgallouedec:
True, not sure what to do in this case

qgallouedec:
closing, see https://github.com/huggingface/trl/issues/3742#issuecomment-4941308985

Issue #3709, opened 2025-07-09
question about entropy_threshold calculating


AGI-player:
### Reproduction

entropy_threshold = torch.quantile(entropies.flatten().float(), self.token_entropy_percentile_threshold) maybe affected by the padding part when padding length is long as follows:
for a batch 3 x 266
completion_mask = tensor([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
         1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
         1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 0],
[...]

PR #3744, opened 2025-07-18
fix loss normalization with entropy mask

hjh0119:
# What does this PR do?

Fix a bug where, when using entropy_mask, tokens that were filtered out would still be included in loss normalization even though their loss was set to zero.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1504,7 +1504,7 @@ class GRPOTrainer(Trainer):
         per_token_loss2 = coef_2 * advantages.unsqueeze(1)
         per_token_loss = -torch.min(per_token_loss1, per_token_loss2)
         if entropy_mask is not None:
-            per_token_loss = per_token_loss * entropy_mask
+            completion_mask = completion_mask & entropy_mask
         if self.beta != 0.0:
             per_token_loss = per_token_loss + self.beta * per_token_kl
 


kashif:
@hjh0119 if we only mask the `completion_mask` then I believe the KL will only be computed on the high entropy tokens and the clip ratio as well... and other issues... i think it's best to do:

```python
        if entropy_mask is not None:
            per_token_loss = per_token_loss & entropy_mask
            # Combine entropy_mask with completion_mask for proper normalization
            effective_mask = completion_mask & entropy_mask
        else:
            effective_mask = completion_mask
        if self.beta != 0.0:

        if self.loss_type == "grpo":
            loss = ((per_token_loss * effective_mask).sum(-1) / effective_mask.sum(-1).clamp(min=1.0)).mean()
        elif self.loss_type == "bnpo":
            loss = (per_token_loss * effective_mask).sum() / effective_mask.sum().clamp(min=1.0)
        elif self.loss_type == "dr_grpo":
            loss = (per_token_loss * completion_mask).sum() / (per_token_loss.size(0) * self.max_completion_length)
        else:
            raise ValueError(f"Unknown loss type: {self.loss_type}")
```

as this isolates entropy masking to only the loss computation while preserving all other functionality

can you try this out?

qgallouedec:
I don't think it's an issue actually. We have discussed this question here: https://github.com/huggingface/trl/issues/3709#issuecomment-3064518856

hjh0119:
That makes sense, I’ll close this PR.

PR #3749, opened 2025-07-20
👨‍💼 [SFT] Packing with completion_only and assistant_only training

### LINKED ISSUE #3728: The assistant_only_loss field becomes ineffective when packing=True is enabled.
During training, when packing=False is enabled, each piece of data will have a completion_mask constructed to mask the relevant fields of template, user, and system. However, when packing=True is set, the completion_mask no longer appears, causing issues with the packing training method. As a result, the trained model will continuously output <|im_start|>.

LeonEricsson:
# What does this PR do?

When training on completions only or assistant messages only the SFT collator expects specific masks in the dataset. However, when packing is enabled these masks are currently lost

https://github.com/huggingface/trl/blob/116ec493fadebb53e6520ddfa66623576b6cf3af/trl/trainer/sft_trainer.py#L808

This is fixed simply by selecting the correct columns.

Fixes #3728

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/sft_trainer.py
```diff
@@ -805,7 +805,15 @@ def tokenize(example, processing_class, dataset_text_field, assistant_only_loss)
                     raise ValueError("When packing is enabled, `max_length` can't be `None`.")
                 if isinstance(dataset, Dataset):  # `IterableDataset.map` does not support `desc`
                     map_kwargs["desc"] = f"Packing {dataset_name} dataset"
-                dataset = dataset.select_columns("input_ids")
+
+                columns = ["input_ids"]
+                if "completion_mask" in dataset.column_names:
+                    columns.append("completion_mask")
+                elif "assistant_mask" in dataset.column_names:
+                    columns.append("assistant_mask")
```
LeonEricsson: @qgallouedec From what I understand, "completion only" and "assistant only" modes are mutually exclusive -  assistant mask training is enabled only for language modeling datasets, whereas "completion only" requires a prompt–completion dataset.

Looking forward though, Is there anything that prevents "assistant only" training from being used with a prompt–completion dataset?

qgallouedec: Technically nothing prevents these two settings to be used together. But I need to check if they actually work together as expected

qgallouedec (approved):

trl/trainer/sft_trainer.py
qgallouedec: Ok now it should be compatible. I added a test for this case. For the record though, this test wouldn't fail before this PR, it would just silently not  the desired training.

qgallouedec:
Feel free to merge if it look good to you 👍

LeonEricsson:
Looks good thanks @qgallouedec

PR #3758, opened 2025-07-23
Dynamic sampling option in GRPO trainer based on DAPO paper

### LINKED ISSUE #3708: [Feature Request] support dynamic sampling for GRPO trainer
### Feature request

Filter most easy and hard samples can increased the diversity of examples, which is beneficial for fine-tuning performance. 

trl can provide a custom function between generation and computing per_token_logps for user to evaluate the quality of prompt, to determinate whether current sample be used in next step or not. 

### Motivation

- dynamic sampling from [DAPO](https://arxiv.org/abs/2503.14476), page 5, 3.2 section. 
- Rollout Rescue Mechanism and Intra-Batch Informative Substitution in [Polaris](https://hkunlp.github.io/blog/2025/Polaris/)

### Your contribution

I'm pleased to contribute this feature

almeidava93:
# What does this PR do?

This PR adds the option to enable dynamic sampling in the GRPO trainer. Dynamic sampling was described in the [DAPO paper](https://arxiv.org/abs/2503.14476) and consists in oversampling prompts which have zero variance in their rewards. This makes the algorithm more efficient by avoiding training in samples with no learning signal. As stated in the DAPO paper, this procedure alone was able to increase their model accuracy on AIME24 (avg@32) from 42% to 50%. 

Specifically, this contribution adds two GRPOConfig options:

- `use_dynamic_sampling`: a boolean argument that enables dynamic sampling
- `max_num_samplings`: an integer argument that provides a maximum number of resamplings. 

This algorithm resamples completions from the model until a non-zero standard deviation is found in the computed rewards or until the maximum number of samplings is reached. `GRPOConfig` docstring was also updated.

Fixes #3708 ([Feature Request] support dynamic sampling for GRPO trainer)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


qgallouedec:
Thanks for the PR!
Were you able to reproduce this result?

almeidava93:
Unfortunately not. They've used a 32B base model. I could try something similar with a much smaller model, but wouldn't be the same thing.

almeidava93:
I have a hardware limitation. I'm trying to reproduce the experiment with a llama 1B model, but this will take time. I would need to first do GRPO training and then retrain using dynamic sampling. But looking at their implementation, it is pretty similar to what I've added.

LeonEricsson:
I would also like to see a wall-time comparison; dynamic sampling certainly increases sample efficiency, but since it requires re-sampling, does it actually result in a net gain in training time?

qgallouedec:
Yes! it was the lack of information on this that prompted us to wait a bit before implementing it.

almeidava93:
Do you believe that an experiment with smaller models is enough? I'm talking about a 1B model trained with LoRA. I thought about using the same data they used in the paper for training and evaluation and compare the performance between GRPO only and GRPO + dynamic sampling

LeonEricsson:
> Do you believe that an experiment with smaller models is enough? I'm talking about a 1B model trained with LoRA. I thought about using the same data they used in the paper for training and evaluation and compare the performance between GRPO only and GRPO + dynamic sampling

That sounds like a solid starting point. Depending on the results, we might consider running tests on larger models later. I can potentially handle that if necessary, but please proceed with the initial experiment first, would be great!

almeidava93:
I have some results to share. I did two experiments using Llama 3.2 1B model with LoRA, one with standard GRPO, another with GRPO + dynamic sampling. 

In the paper, they mentioned that they resampled the model until every sample had rewards with a standard deviation different from zero. I had to limit resampling to a maximum due to hardware limitations. It was set to 10. The maximum number of generated tokens was set to 1000 tokens. Other techniques presented in the paper (clip-higher, Token-level Loss, Soft Overlong Punishment) were not applied. Also, to try to answer your question regarding training time, the GRPO training was done for 1000 training steps and GRPO + dynamic sampling was done for the same time, regardless of the number of training steps. 

Performance on AIME-2024:
- Llama 1B GRPO - 1000 training steps: 0.0052 %
- Llama 1B GRPO + dynamic sampling - 250 training steps: 0.0031%

Total training time was about 7h30 in total for each experiment. Here's the hardware I'm using: 13th Gen Intel(R) Core(TM) i9-13900K (3.00 GHz, 24 cores), 64 GB DDR5 RAM, and a NVIDIA GeForce RTX 4090 GPU with 24 GB VRAM.

Different from my finding, they specifically state that training time was not longer with dynamic sampling. I wonder what kind of optimizations should I do to replicate their setup. What are your thoughts?

One possible optimization would be oversampling instead of only resampling completions.

LeonEricsson:
Am I understanding you correctly that with just `max_num_samplings=10`, training was 4× slower in terms of steps?

almeidava93:
Yep, that was my finding. But I didn't use a couple of strategies to reduce the need for resampling (oversampling) or speeding up inference (with vllm). Do you have any suggestions on how to tackle this considering the hardware limitation I have?

I believe this result may also change depending on what base model you have. If the base model has a higher probability of generating a correct answer, the need for resampling is lower and the training may become more efficient as well. This Llama 3.2 1B model is very limited if compared with the Qwen 32B they've used.

noevernier:
Hey, thanks for the work on this !

I wanted to highlight a potential issue I've run into with the current implementation. Right now, it looks like the entire batch is regenerated if any group in the batch has a zero reward std. In my case (batch size = 16, num_generation=8), the probability that all groups simultaneously have non-zero std is quite low, which means we often end up regenerating full batches multiple times.

This can be quite inefficient, especially when some groups are perfectly usable. Would it make sense to instead preserve the groups with non-zero std and only regenerate the problematic ones? That might significantly reduce the number of redundant forward passes.

Curious to hear your thoughts! Maybe there's a tradeoff I’m not seeing here.

almeidava93:
You're absolutely right. When I implemented this, I focused on handling generations for a single prompt just because I wouldn't be able to have a meaningful number of generations (16 or greater) while batching more than one prompt a time in my local hardware. That needs a fix. 

Either way, I believe the base model you start with has a great impact. If the probability of the base model generating valid answers is very low, the number of generations needed to have at least one sample with valid answers can be very high. I wonder how the training time changes by running this with more powerful models.

shivanis1406:
In the current implementation, If the `for _ in range(self.max_num_samplings):` loop completes by exhausting all its iterations, both rewards and completion_ids will hold the values from the final iteration of the loop. Shouldn't training be skipped until expected batch is created similar to verl implementation https://github.com/volcengine/verl/blob/main/recipe/dapo/dapo_ray_trainer.py

qgallouedec:
Thanks a lot for the work here, and for the experiment.

Doing dynamic sampling right means filtering at the group level and generating until the batch is filled, which is what the comments above point at: regenerating the whole batch throws away usable groups, and exhausting max_num_samplings leaves you training on a batch you meant to discard. That requires decoupling generation from the optimization step and buffering pending groups, and the sync trainer keeps them in lockstep, so it would add a lot of complexity and still not match the paper.

The async GRPO trainer already has that structure: generation runs in its own loop and completions are buffered. That is where we plan to add dynamic sampling.

So we won't merge this one, but it was really useful in mapping out the problem, in particular the wall-time data point. Thanks again.

almeidava93:
Thanks, @qgallouedec 
I would be happy to contribute with the async GRPO trainer and workout the implementation there considering what was discussed. Is there a PR already opened for it? If not, I can start working on it.

PR #3760, opened 2025-07-23
🌋 [GRPO] add support for `pixel_attention_mask` (SmolVLM2) and `image_sizes` (LLaVa-Next)

kashif:
# What does this PR do?

Added support for SmolVLM2's `pixel_attention_mask` and llavanext's `image_sizes` in the GRPO trainer.


trl/trainer/grpo_trainer.py
```diff
@@ -1020,7 +1020,15 @@ def _enable_gradient_checkpointing(self, model: PreTrainedModel, args: GRPOConfi
...
-        self, unwrapped_model, input_ids, attention_mask, logits_to_keep, pixel_values, image_grid_thw
+        self,
+        unwrapped_model,
+        input_ids,
+        attention_mask,
+        logits_to_keep,
+        pixel_values=None,
+        image_grid_thw=None,
+        pixel_attention_mask=None,
+        image_sizes=None,
```
sergiopaniego, line 1037: Should we add a small comment with the model family that these lines are targeting?

qgallouedec:
I just need to generate a tiny model, add a test case and we're good to merge.

examples/scripts/grpo_vlm.py
```diff
@@ -24,21 +24,43 @@
...
-  --gradient_checkpointing \
-  --torch_dtype bfloat16 \
-  --max_prompt_length 2048 \
-  --max_completion_length 1024 \
-  --use_vllm \
-  --vllm_mode colocate \
-  --use_peft \
-  --lora_target_modules "q_proj", "v_proj" \
-  --log_completions
+    --config_file=examples/accelerate_configs/deepspeed_zero3.yaml \
```
lewtun: turbo nit:
```suggestion
    --config_file examples/accelerate_configs/deepspeed_zero3.yaml \
```

examples/scripts/grpo_vlm.py
```diff
@@ -24,21 +24,43 @@
...
+    --vllm_mode colocate \
+    --use_peft \
+    --lora_target_modules "q_proj", "v_proj" \
+    --log_completions
+
+# For HuggingFaceTB/SmolVLM2-2.2B-Instruct
+pip install num2words
+
+accelerate launch \
+    --config_file=examples/accelerate_configs/deepspeed_zero3.yaml \
```
lewtun:
```suggestion
    --config_file examples/accelerate_configs/deepspeed_zero3.yaml \
```

trl/trainer/grpo_trainer.py
```diff
@@ -585,7 +585,9 @@ def __init__(
 
         # Processing class
         if processing_class is None:
-            processing_class = AutoProcessor.from_pretrained(model.config._name_or_path, padding_side="left")
+            processing_class = AutoProcessor.from_pretrained(
+                model.config._name_or_path, padding_side="left", use_fast=True
```
lewtun: Why do we need to specify `use_fast=True`?

trl/trainer/grpo_trainer.py
```diff
@@ -1026,18 +1028,34 @@ def _enable_gradient_checkpointing(self, model: PreTrainedModel, args: GRPOConfi
...
+        pixel_attention_mask=None,
+        image_sizes=None,
     ):
         if is_peft_model(unwrapped_model):
             unwrapped_model = unwrapped_model.base_model.model
 
         # Build model inputs - check if the model supports logits_to_keep (some models and VLMs don't)
         model_inputs = {"input_ids": input_ids, "attention_mask": attention_mask}
 
+        # For Qwen models:
```
lewtun, line 1045: Would it make sense to be more explicit by adjusting the `model_inputs` based on the architecture defined in the model config or is that asking for trouble? The reason I ask is because the code below isn't explicitly tested and I wonder if we will be able to easier catch issues with future models by having a dedicated method we unit test

trl/trainer/grpo_trainer.py
sergiopaniego: No needed actually. Removed

trl/trainer/grpo_trainer.py
qgallouedec: For now, I would vote against it. My hope is that VLMs will gradually standardize their inputs over time, which means we’ll be able to handle all (or at least most) of them with a small, generic set of checks like the ones we have here. If we start explicitly branching based on the model architecture in the config, we’ll end up maintaining a growing list of special cases, and that’s exactly the kind of thing that’s going to break as new models appear. But lest see in the future if we want to be explicit here.

> the code below isn't explicitly tested

It's tested though, all cases should be covered: https://github.com/huggingface/trl/blob/979c1885904cd667b5d75211b1eb5a2154effefb/tests/test_grpo_trainer.py#L1601-L1607

trl/trainer/grpo_trainer.py
```diff
@@ -1026,18 +1026,34 @@ def _enable_gradient_checkpointing(self, model: PreTrainedModel, args: GRPOConfi
...
         if image_grid_thw is not None and pixel_values is not None:
             model_inputs["image_grid_thw"] = image_grid_thw
+        # For Gemma, SmolVLM2, LlavaNext etc.:
         if pixel_values is not None:
             model_inputs["pixel_values"] = pixel_values
+        # For SmolVLM2
+        if pixel_attention_mask is not None:
+            model_inputs["pixel_attention_mask"] = pixel_attention_mask
+        # For LlavaNext
```
```suggestion
        # For LLaVa-Next
```

trl/trainer/grpo_trainer.py
```diff
@@ -1026,18 +1026,34 @@ def _enable_gradient_checkpointing(self, model: PreTrainedModel, args: GRPOConfi
...
         if is_peft_model(unwrapped_model):
             unwrapped_model = unwrapped_model.base_model.model
 
         # Build model inputs - check if the model supports logits_to_keep (some models and VLMs don't)
         model_inputs = {"input_ids": input_ids, "attention_mask": attention_mask}
 
+        # For Qwen models:
         if image_grid_thw is not None and pixel_values is not None:
             model_inputs["image_grid_thw"] = image_grid_thw
+        # For Gemma, SmolVLM2, LlavaNext etc.:
```
```suggestion
        # For Gemma, SmolVLM2, LLaVa-Next etc.:
```

PR #3761, opened 2025-07-23
change doc for `num_iterations` and `steps_per_generation` to hopefully make them more clear and differentiate between them more clearly

### LINKED ISSUE #3197: GRPO - Multiple Policy updates for one completion is not desired
Hi Team,
[Here](https://github.com/huggingface/trl/blob/c5ba70d4fc7cfc2e4515654eb03332d0ff4704d9/trl/trainer/grpo_trainer.py#L571), num_iterations($\mu$) optimisation steps are made on the same completion(trajectory). Please see the attached image from PPO paper - 
Screenshot 2025-03-31 at 6 15 44 PM

Attaching algorithm from DeepseekMath paper - 
Screenshot 2025-03-31 at 6 17 50 PM
I think, for line 11 in the above Algorithm - a different batch size $B$ is used such that $\mu * B \approx |D_b|$. A total of $\mu$ updates are being done using the batch size $B$. 

Please see the PPO algorithm - 
Screenshot 2025-03-31 at 6 44 42 PM
I think $k$ in the PPO-Algo corresponds to $\mu$ in the GRPO. PPO Authors have mentioned hyperparams in appendix A, wherein it can be seen that K*M is less than or approximately equal to NT.

avishaiElmakies:
change the phrasing of docs for `steps_per_generation` and `num_iterations` to make them clearer and to differentiate what they do better.

fixes #3197 

## Before submitting
      Pull Request section?
      to it if that's the case.

@qgallouedec as we have discussed in #3197. this is what I think could make the docs clearer. would love to hear your opinion

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -70,8 +70,9 @@ class GRPOConfig(TrainingArguments):
             `per_device_train_batch_size * num_processes * steps_per_generation`. In other words, there is one
             generation batch processed per optimization step. Mutually exclusive with `steps_per_generation`.
         steps_per_generation: (`int` or `None`, *optional*, defaults to `None`):
-            Number of steps per generation. If `None`, it defaults to `gradient_accumulation_steps`. Mutually exclusive
-            with `generation_batch_size`.
+            Number of mini-steps per generation. generations will be split into `steps_per_generation` steps, for each step a forward pass will be performed on.
+            if `steps_per_generation` > gradient_accumulation_steps, we will have off-policy training.
+            If `None`, it defaults to `gradient_accumulation_steps`. Mutually exclusive with `generation_batch_size`.
         temperature (`float`, defaults to `1.0`):
             Temperature for sampling. The higher the temperature, the more random the completions.
         top_p (`float`, *optional*, defaults to `1.0`):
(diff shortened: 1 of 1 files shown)


ankur6ue:
@qgallouedec Thanks for clarifying the docs. I would recommend adding the following:

1- Clearly explain the use-case for steps-per-generation. Is it performance improvement related to generating completions in advance and reusing them? If so, has it been quantified? Isn't that already handled by simply setting num_iterations to a higher number? If there is no clear use-case, I urge you guys to get rid of steps-per-generation and simplify the API 
2- If a user doesn't want to get into the complication introduced by steps-per-generation and wants to exactly reproduce the math in the deep seek paper, and still use gradient accumulation, what should they set steps-per-generation to?
3- The comment in lines 966-979 in grpo_trainer.py is very helpful (really should be in the docs), but should be expanded to include: which of the steps are on-policy and which off-policy; and how does num_iterations factor in (it is not mentioned at all in the comment)

avishaiElmakies:
Imo Docs for the parameters/arguments should be clear and concise. So i want the explaining to be simple and no more than about 3-4 lines of text. So i don't think i should add ehat you said to the argument documentation
If @qgallouedec agrees i am willing to add more things to `grpo_trainer.md` and go into some of what you asked for there. 

Just to make it clear, `steps_per_generation` and `num_iterations` do different things and are normal hyperparameters for PPO style algorithms. Niether is redundant.

ankur6ue:
Thanks, adding a clear explanation of the use-cases (s) behind steps_per_generation and interaction with gradient_accumulation and num_iterations to grpo_trainer.md would be much appreciated!

qgallouedec:
closing as no recent activity. AFAIK, we haven't received other clarification requests so for now I think we're good

Issue #3706, opened 2025-07-08
trl with vllm integration not working


PodoprikhinMaxim:
### Reproduction

I am trying to run trl with vllm integration and get with `.init_communicator()` problem:

run vllm server with
```
CUDA_VISIBLE_DEVICES=0,1,2 trl vllm-serve --model Qwen/Qwen2-0.5B --tensor-parallel-size 2 --data-parallel-size 1 --port=8000
```

and main script is just

```
from trl.extras.vllm_client import VLLMClient

vllm_client = VLLMClient(
    "http://0.0.0.0:8000",
    connection_timeout=10
)
print("started init_communicator")
vllm_client.init_communicator()
print("done!")
```

Error log:

```
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8000/ (Press CTRL+C to quit)
INFO:     [127.0.0.1:39726](http://127.0.0.1:39726/) - "GET /health/ HTTP/1.1" 200 OK
INFO:     [127.0.0.1:39728](http://127.0.0.1:39728/) - "GET /get_world_size/ HTTP/1.1" 200 OK
INFO:     [127.0.0.1:39744](http://127.0.0.1:39744/) - "POST /init_communicator/ HTTP/1.1" 200 OK
(VllmWorker rank=1 pid=4888) INFO 07-08 10:09:09 [utils.py:1126] Found nccl from library libnccl.so.2
(VllmWorker rank=1 pid=4888) INFO 07-08 10:09:09 [pynccl.py:70] vLLM is using nccl==2.26.2
(VllmWorker rank=0 pid=4887) INFO 07-08 10:09:09 [utils.py:1126] Found nccl from library libnccl.so.2
(VllmWorker rank=0 pid=4887) INFO 07-08 10:09:09 [pynccl.py:70] vLLM is using nccl==2.26.2

[2025-07-08 10:09:09] research-cluster-podoprihin-2em:4887:4887 [0] init.cc:722 NCCL WARN Duplicate GPU detected : rank 0 and rank 2 both on CUDA device 85000
(VllmWorker rank=0 pid=4887) ERROR 07-08 10:09:09 [multiproc_executor.py:527] WorkerProc hit an exception.
(VllmWorker rank=0 pid=4887) ERROR 07-08 10:09:09 [multiproc_executor.py:527] Traceback (most recent call last):
[...]

PR #3762, opened 2025-07-23
Prevent NCCL Device Conflicts Between vLLM Server and Trainers

CarlosArguilar:
# What does this PR do?

This PR addresses the issue #3706.

## Problem
In the issue the reported error in a simple setup including a vllm server and vllm client is the following:

```
ERROR 07-08 10:09:09 [core.py:583] RuntimeError: Worker failed with error 'NCCL error: invalid usage (run with NCCL_DEBUG=WARN for details)', please check the stack trace above for the root cause
```

The error occurs when calling the `trl.extras.vllm_client.VLLMClient.init_communicator` method, that through `vllm.distributed.device_communicators.pynccl.PyNcclCommunicator` calls the `ncclCommInitRank
` function from NCCL API.

The problem is being caused when the device specified in `trl.extras.vllm_client.VLLMClient.init_communicator` is the same as the device of one of the vllm workers, since NCCL Communicator doesn't support two ranks in the same device:

[Docs NCCL Communicator](https://docs.nvidia.com/deeplearning/nccl/archives/nccl_2262/user-guide/docs/usage/communicators.html)
> Using the same CUDA device multiple times as different ranks of the same NCCL communicator is not supported and may lead to hangs.

## Solution
With that in mind, the error reported in the issue is not a bug per say, it is caused by limitation due to unsupported configuration of the NCCL group.

Considering that, this PR addresses the problem by verifying if the vllm workers devices and the device specified in VLLMClient are different. In case this condition is not met, an exception will be raised in the vllm server to make clear the cause of the error.

For the example reported in the issue the following exception would be raised in the vllm server:

```
RuntimeError: Attempting to use the same CUDA device (UUID: ...) 
for multiple distinct roles/ranks within the same communicator. 
Ensure that trainer is using different devices than vLLM server.
```

Making very clear that the problem is using a common device for both vLLM server and trainer (VLLMClient) process.

---

## Summary of changes
The changes applied in this PR are:

- **Check for usage of the same device in VLLMClient and vLLM server (`trl/extras/vllm_client.py` and `trl/scripts/vllm_serve.py`)**: Check through device UUID if vllm worker is on the same device as the trainer main process. If it is, it raises a RuntimeError with a message to make it clear the cause of the problem.
  

- **VLLM integration docs update (`docs/source/vllm_integration.py`)**: Small updates in the VLLM integration documentation page to make it clear that the devices used for the VLLM server and trainers cannot be the same.

Some additional changes were also made to improve the VLLM Client functionalities and tests:

- **Add device parameter in `VLLMClient.init_communicator` method (`trl/extras/vllm_client.py`)**: Currently the device is hard coded to `0` in the `init_communicator` method. A `device` parameter was added to handle rare cases where the main process is not on logical cuda:0. The default value is still `0` so it doesn't break any existing usage.
- **Pass accelerator main process device to `VLLMClient.init_communicator` device parameter in GRPOTrainer (`trl/trainer/grpo_trainer.py`)**: Ensure that the communicator is receiving the correct device of the main process instead of using the default `0` device. This is the only trainer that calls the `VLLMClient.init_communicator` directly.
- **Add tests `TestVLLMClientServerDeviceParameter` (`tests/test_vllm_client_server.py`)**: Tests to ensure proper functioning of the new parameter of `VLLMClient.init_communicator` 

Also two additional corrections were made in existing test code:

- **Fix checking of `require_3_accelerators` decorator (`tests/testing_utils.py`)**: This decorator is used to ensure that the system has at least 3 accelerators to run some tests but it was checking for **more** than 3 (at least 4). It was updated to check `>=3` instead of `>3`.
- **Fix test `TestVLLMClientServerDP` (`tests/test_vllm_client_server.py`)**: The `VLLMClient.init_communicator` method was not being called in the `setUpClass` method, and the parameter update tests of the class were failing (even in the main branch) due to the communicator not being initialized.

All the `tests/test_vllm_client_server` were executed and all passed.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


docs/source/vllm_integration.md
```diff
@@ -6,21 +6,23 @@ This document will guide you through the process of using vLLM with TRL for fast
 
 💡 **Note**: Resources required for this specific example: a single node with 8 GPUs.
 
+🚨 **Important**: vLLM server and TRL trainer must use different CUDA devices to avoid conflicts.
```
qgallouedec:
```suggestion
<Tip warning=true>
vLLM server and TRL trainer must use different CUDA devices to avoid conflicts.
</Tip>
```

qgallouedec (approved):
Thanks @CarlosArguilar! very clean PR. Waiting for the CI to pass and we're good to merge

PR #3763, opened 2025-07-24
Fix clone chat template

qgallouedec:
# What does this PR do?

This PR addresses 2 issues:

- Fix `clone_chat_template` to handle dummy token addition for vocabulary size alignment. Before this change, the returned model and tokenizer may not have the same vocab size, which causes the loading of the trained model to fail.

- Enhance `SFTTrainer` to handle added tokens from chat template cloning: it ensures that both the `lm_head` and the new embed tokens are trainable

### DIFF
--- a/trl/models/utils.py
+++ b/trl/models/utils.py
@@ -149,7 +149,7 @@ def clone_chat_template(
     tokenizer: PreTrainedTokenizer,
     source_tokenizer_path: str,
     resize_to_multiple_of: Optional[int] = 64,
-) -> tuple[PreTrainedModel, PreTrainedTokenizer, list[AddedToken]]:
+) -> tuple[PreTrainedModel, PreTrainedTokenizer, list[int]]:
     """
     Clones a chat template from a source tokenizer to the target tokenizer and updates the model accordingly.
 
@@ -177,7 +177,7 @@ def clone_chat_template(
             Updated model with resized token embeddings and EOS token configured.
         tokenizer (`~transformers.PreTrainedTokenizer`):
             Updated tokenizer with the chat template and special tokens applied.
-        added_tokens (`list[AddedToken]`):
+        added_tokens (`list[int]`):
             List of tokens that were added to the tokenizer from the source tokenizer.
 
     Example:
--- a/trl/models/utils.py
+++ b/trl/models/utils.py
@@ -20,7 +20,7 @@ from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 
 import torch.nn as nn
 from packaging import version
-from transformers import AutoTokenizer, PreTrainedModel, PreTrainedTokenizer
+from transformers import AddedToken, AutoTokenizer, PreTrainedModel, PreTrainedTokenizer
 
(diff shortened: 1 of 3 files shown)


trl/models/utils.py
```diff
@@ -210,7 +216,27 @@ def clone_chat_template(
         pad_to_multiple_of=resize_to_multiple_of if resize_to_multiple_of is not None else None,
     )
 
-    return model, tokenizer
+    # After padding, the embedding matrix size may exceed the vocabulary size. Add dummy tokens to the tokenizer to
+    # ensure vocabulary size matches the embedding matrix dimensions.
+    new_embedding_size = model.get_input_embeddings().weight.size(0)
```
kashif:
```suggestion
    new_embedding_size = model.vocab_size
```

qgallouedec: Thanks! didn't know about this one

edbeeching (approved):

trl/models/utils.py
```diff
@@ -210,7 +216,27 @@ def clone_chat_template(
         pad_to_multiple_of=resize_to_multiple_of if resize_to_multiple_of is not None else None,
     )
 
-    return model, tokenizer
+    # After padding, the embedding matrix size may exceed the vocabulary size. Add dummy tokens to the tokenizer to
```
lewtun: turbo nit as I was initially confused why we were talking about padding:
```suggestion
    # After resizing, the embedding matrix size may exceed the vocabulary size. Add dummy tokens to the tokenizer to
```

trl/models/utils.py
```diff
@@ -210,7 +216,27 @@ def clone_chat_template(
...
+    # ensure vocabulary size matches the embedding matrix dimensions.
+    new_embedding_size = model.vocab_size
+    vocab_size_before_padding = len(tokenizer.vocab)
+
+    if new_embedding_size > vocab_size_before_padding:
+        # Add dummy tokens to fill the gap created by padding
+        num_dummy_tokens = new_embedding_size - vocab_size_before_padding
+        dummy_tokens = [AddedToken(f"<extra_id_{idx}>") for idx in range(num_dummy_tokens)]
```
lewtun: In the rare case where the tokenizer already has pre-existing `<extra_id_{idx}>` tokens, do this cause an issue with the vocab size not getting extended correctly?

lewtun (approved):
LGTM with a question about a rare edge case when extending the vocab size with tokens that may already exist in the base tokenizer

trl/models/utils.py
qgallouedec: agree, did the same here: https://github.com/huggingface/trl/pull/3763/commits/098a33b53370f49e17deb320935282b4a17457d5

trl/models/utils.py
qgallouedec: good point, let me fix this

qgallouedec: fixed + added test case here: [57060d0](https://github.com/huggingface/trl/pull/3763/commits/57060d0e172521ca28410cf73828d1067d235ec8)

PR #3767, opened 2025-07-24
Add uv scripts headers

lhoestq:
I did my best to guess the dependencies, lmk if it sounds good to you !

Btw for `trl` I require `trl @ git+https://github.com/huggingface/trl.git` but for releases it would be nice to require `trl==VERSION`, and then set the git URL back when setting the dev version `VERSION.dev0`

`transformers` has a `utils/release.py` file to do this automatically, that I updated to account for uv scripts in https://github.com/huggingface/transformers/pull/39635

cc @lewtun @qgallouedec

### DIFF
--- a/examples/scripts/alignprop.py
+++ b/examples/scripts/alignprop.py
@@ -12,6 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# /// script
+# dependencies = [
+#     "trl @ git+https://github.com/huggingface/trl.git",
+#     "Pillow>=9.4.0",
+# ]
+# ///
+
 """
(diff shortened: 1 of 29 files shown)


examples/scripts/alignprop.py
```diff
@@ -12,6 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# /// script
+# dependencies = [
+#     "trl @ git+https://github.com/huggingface/trl.git",
+#     "Pillow>=9.4.0",
```
lhoestq, line 18: fyi this is the minimum version to support EXIF metadata

qgallouedec (approved):

PR #3771, opened 2025-07-25
🍿 [SFT] Fix dataset indexing which crashed with a IterableDataset 

LeonEricsson:
# What does this PR do?

Fixes #3768 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [x] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #3789, opened 2025-07-28
Fix bug when `train_dataset = None` in `SFTTrainer`

h-tonywu:
# What does this PR do?

Fixes a bug when an `SFTTrainer` instance is created with `train_dataset=None`.

## Cause

The bug was introduced in https://github.com/huggingface/trl/pull/3771 as we are using `next(iter(train_dataset))`, where `train_dataset` can be `None` (weird that it wasn't caught by mypy btw).

## Note

After looking at the `Trainer` code from `transformers`, I believe the user should be able to have `train_dataset=None`, otherwise we wouldn't have `train_dataset: Optional[Union[Dataset, IterableDataset, "datasets.Dataset"]] = None` nor these lines:

https://github.com/huggingface/transformers/blob/45caf5c32bd5085a58edc6ec90c8269a54576d0e/src/transformers/trainer.py#L709-L714

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -432,11 +432,11 @@ class SFTTrainer(Trainer):
                     "to at least 2."
                 )
 
-        dataset_sample = next(iter(train_dataset))
-        if args.completion_only_loss is None:
-            self.completion_only_loss = "prompt" in dataset_sample
-        else:
+        if args.completion_only_loss is not None:
             self.completion_only_loss = args.completion_only_loss
+        else:
+            dataset_sample = next(iter(train_dataset))
+            self.completion_only_loss = "prompt" in dataset_sample
 
         if data_collator is None:
             # Get the pad token: if not provided, use the one from the processing class or the eos token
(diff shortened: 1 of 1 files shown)


qgallouedec:
how do you train if you don't have a training dataset?

qgallouedec:
Technically, `train_dataset=None` is valid according to the type hint — but the reason it’s set to `None` isn’t necessarily because we expect people to train without a dataset. It’s more about function signature design:

If `train_dataset` didn’t allow `None`, we’d either have to (1) provide a different default value, or (2) remove the default entirely, which would break the ordering of arguments.

Since `data_collator` (which comes before `train_dataset`) already allows `None`, keeping the argument order consistent with `Trainer` means we also have to assign a default value to `train_dataset`, and `None` is the logical choice.

h-tonywu:
> how do you train if you don't have a training dataset?

In our code, we override `trainer.get_train_dataloader` with our custom dataloader to have full control on the underlying processes.

> Technically, `train_dataset=None` is valid according to the type hint — but the reason it’s set to `None` isn’t necessarily because we expect people to train without a dataset. It’s more about function signature design:
> 
> If `train_dataset` didn’t allow `None`, we’d either have to (1) provide a different default value, or (2) remove the default entirely, which would break the ordering of arguments.
> 
> Since `data_collator` (which comes before `train_dataset`) already allows `None`, keeping the argument order consistent with `Trainer` means we also have to assign a default value to `train_dataset`, and `None` is the logical choice.

Thanks for the thorough answer! Regarding my need, I guess the right move is to leverage `accelerate` instead to have more control on the dataloader. Closing the PR :)

PR #3791, opened 2025-07-28
🎚️ Add dataset mixer

lewtun:
# What does this PR do?

This PR adds support to mix datasets / config / splits for more effective training recipes. The usage is best demonstrated via a YAML config with something like:

```yaml
dataset_mixture:
    datasets:
    - id: dataset_id_1
        config: config_name
        columns:
        - col1
        - col2
        weight: 0.5
    - id: dataset_id_2
        config: config_name
        columns:
        - col1
        - col2
        weight: 0.5
    seed: 42
    test_split_size: 0.1
```

The PR also introduces a helper `get_dataset()` method to make it simpler to return the dataset from either the mixture or the single dataset setup we had previously.

A few questions:

1. I am not sure if `weight` is the right terminology since it's not relative to other configs. An alternative would be `frac` or `pct`
2. Where should this be documented?
3. Should I add the `get_dataset()` utitlity function to our training scripts or is this feature best left for power users?
4. One limitation with the current implementation is that you cannot independently specify which subsets should be mixed as a train or test split (we create a separate test split if requested by the user). Do we need to support this?

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


qgallouedec:
Nice! Can you replace `load_dataset` by `get_dataset` in the scripts (trl/scripts)?

qgallouedec:
I chose a slightly different approach than the one initially proposed:

1. `dataset_mixture` is no longer a parameter of `ScriptArguments` but a dataclass in its own right. The parser does not handle nested structures well, so putting them all on the same level actually makes things much easier.

  ```diff
  - dataset_mixture:
  -   datasets:
  -     ...
  + datasets:
  +   ...
  ```

2. Remove seed and dataset shuffling. IMO it's up to the trainer to shuffle (or not) the dataset. As for the splitting test/train, the seed used is actually the one that will be set by the training argument. (no need to have several)

3. The arg name are now aligned with `datasets.load_dataset`. They may not be the best names, but consistency makes things much simpler.

  ```diff
  - - id: dataset_id_1
  + - path: dataset_id_1
  -   config: config_name
  +   name: config_name
  ```

4. `weight` is removed in favour of the already existing (and more advanced) support for slicing in the split name.

  ```diff
  - split: train
  - weight: 0.5
  + split: train[:50%]
  ```

5. A streaming mode is added (`streaming`)


*Consequently*

Before:

```yaml 
dataset_mixture:
    datasets:
    - id: dataset_id_1
      config: config_name
      split: train
      columns:
      - col1
      - col2
      weight: 0.5
    - id: dataset_id_2
      config: config_name
      split: train
      columns:
      - col1
      - col2
      weight: 0.5
    seed: 42
    test_split_size: 0.1
```

Now: 

```yaml 
datasets:
- path: dataset_id_1
  name: config_name
  split: train[:50%]
  columns:
  - col1
  - col2
- path: dataset_id_2
  name: config_name
  split: train[:50%]
  columns:
  - col1
  - col2
test_split_size: 0.1
```


Not a definitive design choice, if you want me to revert some of these modifications, I'd be happy to do so.

qgallouedec:
> * I am not sure if `weight` is the right terminology since it's not relative to other configs. An alternative would be `frac` or `pct`

addressed by relying on split

> * Where should this be documented?

now it's documented

> * Should I add the `get_dataset()` utitlity function to our training scripts or is this feature best left for power users?

done

> * One limitation with the current implementation is that you cannot independently specify which subsets should be mixed as a train or test split (we create a separate test split if requested by the user). Do we need to support this?

Not sure either. Maybe having `train_datasets` (instead of `datasets`) and a new `test_datasets` in `DatasetMixtureConfig`? I'm ok to leave it like this for now, let's improve this if needed in a following PR

qgallouedec (approved):

PR #3804, opened 2025-07-30
💇 Add soft overlong punishment reward function and update documentation

qgallouedec:
Just get soft overlong punishment from Open-R1, because I saw it in several papers


trl/rewards/other_rewards.py
```diff
@@ -0,0 +1,63 @@
...
+    0, & |y| \le L_{\max} - L_{\text{cache}} \\
+    \dfrac{(L_{\max} - L_{\text{cache}}) - |y|}{L_{\text{cache}}}, & L_{\max} - L_{\text{cache}} < |y| \le L_{\max} \\
+    -1, & L_{\max} < |y|
+    \end{cases}
+    $$
+
+    Args:
+        max_completion_len (`int`):
+            Maximum length of the completion,  \( L_{\max} \).
+        soft_punish_cache (`int`):
```
1485840691, line 36: Why not name it as "min_completion_len" to align with comment?

pramodith (approved):
LGTM!

trl/rewards/other_rewards.py
qgallouedec: Thanks for the suggestion. Here we tried to align with the paper:

image

PR #3814, opened 2025-07-30
fix CI docs and grpo slow test

kashif:
# What does this PR do?

Fix the CI docs failure now that we have the liquid tags escaped:

Screenshot 2025-07-30 at 20 11 42


kashif:
Screenshot 2025-07-30 at 20 38 05

kashif:
this seems to work for liquid and doc-builder

Issue #3811, opened 2025-07-30
GSPO-token implementation


BounharAbdelaziz:
### Feature request

I realised that the GSPO implementation in #3775 is the vanilla one and was wondering if you wanted to include the GSPO-token variant (which reduces to the vanilla one when the advantage is the same for all tokens).

### Motivation

It seems that the main author of GSPO has some preference towards the token variant as it gives more [flexibility](https://github.com/volcengine/verl/pull/2775#discussion_r2239322506).

### Your contribution

I can take care of the implementation and submit a PR in this regard.

hjh0119:
According to equations (11) and (18) in the paper, the difference between the two is that gspo-token supports token-level advantage.

However, with the current reward/advantage calculation, all tokens have the same value, so I think there is actually no difference between them.

BounharAbdelaziz:
> According to equations (11) and (18) in the paper, the difference between the two is that gspo-token supports token-level advantage.
> 
> However, with the current reward/advantage calculation, all tokens have the same value, so I think there is actually no difference between them.

Indeed @hjh0119 , that's what I've said; "GSPO-token variant (which reduces to the vanilla one when the advantage is the same for all tokens)".

I believe having the general one gives more freedom to the user to choose.

qgallouedec:
Hi @BounharAbdelaziz, sorry for the very late close here, we're catching up on the backlog.

Good news: GSPO-token was implemented in #3820 (merged) and is on current main under `trl.experimental.gspo_token`, activated with `importance_sampling_level="sequence_token"`; it uses exactly the paper's stop-gradient formulation (see https://github.com/huggingface/trl/blob/d71a6ca90806bd6dc62cdea5cc26b69f4c376403/trl/experimental/gspo_token/grpo_trainer.py#L77-L81). As you and @hjh0119 noted, it's mathematically equivalent to GSPO (`importance_sampling_level="sequence"`) under the standard per-sequence advantages, and only diverges once you supply per-token advantages. The docs cover this at https://huggingface.co/docs/trl/main/en/gspo_token. Closing as completed; thanks for the request, and thanks to @hjh0119 for the PR!

PR #3820, opened 2025-07-31
🪙 [Experimental] Support GSPO-token

hjh0119:
Support for GSPO-token as described in [GSPO paper](https://arxiv.org/abs/2507.18071), Section 4.3.

related issue: https://github.com/huggingface/trl/issues/3811

**GSPO** 
$w_{i}^{\mathrm{GSPO}} = \left[ \frac{\pi_{\theta}(y_i \mid x)}{\pi_{\theta_{\mathrm{old}}}(y_i \mid x)} \right]^{\frac{1}{|y_i|}} = \exp(\frac{1}{|y_i|} \sum_{t=1}^{|y_i|} \log \frac{\pi_{\theta}(y_{i, t} \mid x, y_{i, <t})}{\pi_{\theta_{\mathrm{old}}}(y_{i, t} \mid x, y_{i, <t})})$

**GSPO-token**
$w_{i, t}^{\mathrm{GSPO_token}} = \mathrm{sg}\left[w_i^{\mathrm{GSPO}}\right] \cdot \frac{\pi_{\theta}(y_{i, t} \mid x, y_{i, < t})}{\mathrm{sg}\left[\pi_{\theta}(y_{i, t} \mid x, y_{i, < t})\right]}$

> where $\mathrm{sg}[\cdot]$ denotes the stop-gradient (detach) operation.

> 💡 **NOTE:** GSPO-token enables support for fine-grained (token-level) advantages. 
> However, given the current formulation for advantage computation, all tokens within a sentence share the same value. In this case, GSPO and GSPO-token are theoretically equivalent, as shown in equations (11) and (18) of the paper.


LeonEricsson:
Thanks for this. Since GSPO-token is a generalized version of vanilla GSPO, I suggest we fully transition to GSPO-token instead of supporting both versions. Consequently, we would rename/remove `importance_sampling_level`, as both methods operate at the token level.

trl/trainer/grpo_trainer.py
```diff
@@ -1837,6 +1837,11 @@ def _compute_loss(self, model, inputs):
         elif self.importance_sampling_level == "sequence":
             log_importance_weights = (log_ratio * completion_mask).sum(-1) / completion_mask.sum(-1).clamp(min=1.0)
             log_importance_weights = log_importance_weights.unsqueeze(-1)
+        elif self.importance_sampling_level == 'sequence_token':
+            # GSPO-token: sg[si(θ)] * πθ(yi,t)/sg[πθ(yi,t)]
+            seq_level_log_weight = (log_ratio * completion_mask).sum(-1) / completion_mask.sum(-1).clamp(min=1.0)
+            seq_level_log_weight = seq_level_log_weight.unsqueeze(-1).detach()  # Stop gradient
```
```suggestion
            seq_level_log_weight = seq_level_log_weight.detach().unsqueeze(-1) # Stop gradient
```

trl/trainer/grpo_trainer.py
```diff
@@ -1837,6 +1837,11 @@ def _compute_loss(self, model, inputs):
...
+            log_importance_weights = seq_level_log_weight + per_token_logps - per_token_logps.detach()
```
```suggestion
            log_importance_weights = per_token_logps - per_token_logps.detach() + seq_level_log_weight
```
reorder to avoid precision errors

trl/trainer/grpo_trainer.py
pramodith: `(log_ratio * completion_mask).sum(-1) / completion_mask.sum(-1).clamp(min=1.0)`

This op is common across GSPO and GSPO-token, would be good to have a single variable pointing to this value under an if condition like

`if self.importance_sampling_level != 'token'`

trl/trainer/grpo_trainer.py
hjh0119: It should still be float32 after reordering

So maybe we need to manually convert the dtype like

```python
log_importance_weights = (seq_level_log_weight + per_token_logps - \
per_token_logps.detach()).to(per_token_logps.dtype)
```

trl/trainer/grpo_trainer.py
hjh0119: make sense, so shall we move the invalid value check for importance_sampling_level into the model parameter initialization?

hjh0119:
> Thanks for this. Since GSPO-token is a generalized version of vanilla GSPO, I suggest we fully transition to GSPO-token instead of supporting both versions. Consequently, we would rename/remove `importance_sampling_level`, as both methods operate at the token level.

Agreed to keep GSPO-token. Should we retain this parameter for compatibility with previous usage, or introduce an additional parameter instead? Which is better?

trl/trainer/grpo_trainer.py
LeonEricsson: wasn't referring to the dtype, just better to perform the subtraction `per_token_logps - per_token_logps.detach()` first, to avoid catastrophic cancellation

hjh0119: that make sense, fixed

hjh0119:
@qgallouedec @lewtun @edbeeching @kashif If there are any concerns or suggestions, please feel free to let me know. Thank you very much in advance

LeonEricsson:
> Agreed to keep GSPO-token. Should we retain this parameter for compatibility with previous usage, or introduce an additional parameter instead? Which is better?

imo it should be removed, however, since it's already been published as part of TRL v0.20, we may need to keep it for backward comp. I can't speak to it myself, so I'll leave it to someone else to decide.

qgallouedec:
Thanks for the contribution, and apologies for the delay in reviewing it.  

After reading the paper, I don’t think this PR fully achieves GSPO-token. This variation is most relevant when you have a fine-grained advantage— i.e. when $\hat{A_{i,t}}$ varies with $t$—which isn't the case here, since $\hat{A_{i,t}}=\hat{A_{i}}$.

hjh0119:
> After reading the paper, I don’t think this PR fully achieves GSPO-token. This variation is most relevant when you have a fine-grained advantage— i.e. when A i , t ^ varies with t —which isn't the case here, since A i , t ^ = A i ^ .

Sure. As I mentioned, when there is no fine-grained advantage, the gspo-token gradient is equivalent to the original implementation. 

However, do we need to implement this algorithm in advance to accommodate possible future fine-grained advantage, or to make it easier for downstream users to implement their own customized fine-grained advantage? Anyway, thank you for your review.

hjh0119:
Any feedback before I close this PR, or should we go ahead and merge it?

qgallouedec:
Please leave it open, we are working hard to provide fast review for all the PRs 🙏

qgallouedec (approved):

PR #3832, opened 2025-08-01
[GRPO]  👁️ Fix vLLM server mode for VLM GRPO training incompatibility for certain AutoProcessors

ghubnerr:
# What does this PR do?

This PR addresses a recent (unreported) issue regarding using vLLM server mode for VLM training with GRPO. Certain models, such as `Gemma3ForConditionalGeneration` don't require `<image_soft_token>` to be included in the prompt text used to call its `AutoProcessor`. 

The code block below tries to de-tokenize a chat-templated tensor and send over a fresh text prompt to vLLM. It does so by collapsing the expanded `trainer.image_token`'s placeholders into the format expected by some chat templates (e.g. `SmolVLM-Instruct`'s chat template is: 

`
"chat_template": "<|im_start|>{% for message in messages %}{{message['role'] | capitalize}}{% if message['content'][0]['type'] == 'image' %}{{':'}}{% else %}{{': '}}{% endif %}{% for line in message['content'] %}{% if line['type'] == 'text' %}{{line['text']}}{% elif line['type'] == 'image' %}{{ '<image>' }}{% endif %}{% endfor %}<end_of_utterance>\n{% endfor %}{% if add_generation_prompt %}{{ 'Assistant:' }}{% endif %}"`

Which inserts the `<image>` token (`trainer.image_token`) into the prompt. On the other hand, Gemma3 only uses `<start_of_image>` to indicate an image placement in the chat template. This means there will be an additional  (256 + 1 = 257) `<image_soft_token>` instance received by vLLM after the prompt is re-expanded by the `AutoProcessor` inside the vLLM server since it is never required in the first place.

Resulting in the error:
```
Attempted to assign 256 = 256 multimodal tokens to 257 placeholders
```

The following lines are the culprit:
```py
if self.max_prompt_length is not None:
    # If max_prompt_length is set, we trim the prompt to keep only the last `max_prompt_length` tokens.
    # Then we decode those tokens back into text. We manually remove leading pad tokens from the decoded text,
    # because we can't use `skip_special_tokens=True` (some special tokens are still needed for generation).
    protected = [self.image_token_id, self.vision_start_token_id, self.vision_end_token_id]
    protected = [token for token in protected if token is not None]
    prompt_ids, prompt_mask = truncate_with_protected_tokens(
        prompt_ids, prompt_mask, self.max_prompt_length, protected
    )

    prompts_text = self.processing_class.batch_decode(
        prompt_ids, skip_special_tokens=False, clean_up_tokenization_spaces=False
    )
    prompts_text = [re.sub(rf"^({re.escape(self.pad_token)})+", "", text) for text in prompts_text]

    # The chat template inserts a single image token into the prompt text. However, when this text is later
    # tokenized, the single image token string is expanded into multiple image token IDs, depending on the
    # image size. Since we're detokenizing here, we may see repeated image tokens in the decoded text. We
    # collapse them back into a single token string to match the original template.
    if self.image_token is not None:
        prompts_text = [
            re.sub(rf"({re.escape(self.image_token)})+", self.image_token, text) for text in prompts_text
        ]
```

### Solution
To solve for this, I edited the code block to only leave a single `trainer.image_token` behind in case that token is present in the chat template at all. Otherwise, it collapses everything, including the `trainer.vision_end_token_id` token that is added afterwards, leaving behind only `trainer.vision_start_token_id`. 

I've also added both a test that includes Qwen2.5VL + Gemma3 VLM + vLLM testing, and I have confirmed on my end that both work with GSPO + PEFT :)

## Before submitting
      Pull Request section?


kashif:
thanks @ghubnerr having a look

kashif:
Just wanted to double check, the regex you have `({escaped_img_token})+{escaped_eoi_token}` might not work as expected because it's looking for image tokens immediately followed by end tokens, but they might be separated by other content? 

Do you think we would need to collapse multiple `vision_start_token_id` instances?

Could it be that the `convert_ids_to_tokens` might not give the same results, due to merging,  as `decode([token_id])`? not sure about that...

what are your thoughts?

kashif:
@ghubnerr would something like:
```py
            for i, text in enumerate(prompts_text):
                # Collapse repeated image tokens
                if self.image_token is not None:
                    text = re.sub(rf"({re.escape(self.image_token)})+", self.image_token, text)
                
                # Collapse repeated vision start tokens
                if self.vision_start_token_id is not None:
                    vision_start_text = self.processing_class.decode([self.vision_start_token_id])
                    text = re.sub(rf"({re.escape(vision_start_text)})+", vision_start_text, text)
                
                # Remove vision end tokens (not needed in chat templates)
                if self.vision_end_token_id is not None:
                    vision_end_text = self.processing_class.decode([self.vision_end_token_id])
                    text = re.sub(rf"({re.escape(vision_end_text)})+", "", text)
                
                prompts_text[i] = text
```
work?

ghubnerr:
> @ghubnerr would something like:
> 
> ```python
>             for i, text in enumerate(prompts_text):
>                 # Collapse repeated image tokens
>                 if self.image_token is not None:
>                     text = re.sub(rf"({re.escape(self.image_token)})+", self.image_token, text)
>                 
>                 # Collapse repeated vision start tokens
>                 if self.vision_start_token_id is not None:
>                     vision_start_text = self.processing_class.decode([self.vision_start_token_id])
>                     text = re.sub(rf"({re.escape(vision_start_text)})+", vision_start_text, text)
>                 
>                 # Remove vision end tokens (not needed in chat templates)
>                 if self.vision_end_token_id is not None:
>                     vision_end_text = self.processing_class.decode([self.vision_end_token_id])
>                     text = re.sub(rf"({re.escape(vision_end_text)})+", "", text)
>                 
>                 prompts_text[i] = text
> ```
> 
> work?

Hi! Thanks for the suggestion! I'm not sure if this would work:
> Do you think we would need to collapse multiple vision_start_token_id instances?

There hasn't been an instance in my experiments where there were more `vision_start_token_id` instances than images to be placed. The core of the issue lies in whether the model's AutoProcessor needs the `vision_start_token_id` + `image_token_id`+`vision_end_token_id` or just `vision_start_token_id` (like Gemma 3). 

The issue with this line (and the sequential ones) is that it is not dynamic to their chat templates, and it still carries the underlying issue of unconditionally re-collapsing all `image_token_id`s into a single one, which is not the desired behavior all of the time (compared to not including them at all).

```python
text = re.sub(rf"({re.escape(self.image_token)})+", self.image_token, text)
                                                    ---------------- <- the issue, should be "" sometimes
```

As to this comment:
> Just wanted to double check, the regex you have ({escaped_img_token})+{escaped_eoi_token} might not work as expected because it's looking for image tokens immediately followed by end tokens, but they might be separated by other content?

I see the point you're making -- and you're right to say that it's looking for image tokens immediately followed by end tokens, but it follows the convention and the assumption that the chat template which contains the `image_token_id` would be surrounded by `vision_start_token_id` as well as `vision_end_token_id`. Let me know your thoughts!

Also, yes, I think you're right on this part. I didn't see the nuance in the method. `decode([token_id])` may be better indeed.
> Could it be that the convert_ids_to_tokens might not give the same results, due to merging, as decode([token_id])? not sure about that...

kashif:
@ghubnerr  yes I see what you mean too.. ok take the parts you feel and the rest then looks fine to me

ghubnerr:
> @ghubnerr yes I see what you mean too.. ok take the parts you feel and the rest then looks fine to me

Okay awesome, I just pushed the change using the `decode` method instead of the `convert_ids_to_token`.

An example string is now being collapsed from:

```py
['<bos><start_of_turn>user\nHello, this is my image\n\n\n\n<start_of_image><image_soft_token>.....(image_seq_len times)...<image_soft_token><end_of_image>\n\nHello world, this is my prompt<end_of_turn>\n<start_of_turn>model\n']
```
To:
```py
['<bos><start_of_turn>user\nHello, this is my image\n\n\n\n<start_of_image>\n\nHello world, this is my prompt<end_of_turn>\n<start_of_turn>model\n']
```
Instead of
```py
['<bos><start_of_turn>user\nHello, this is my image\n\n\n\n<start_of_image><image_soft_token><end_of_image>\n\nHello world, this is my prompt<end_of_turn>\n<start_of_turn>model\n']
```
If the model does not include the `image_token` on its chat template.

Where `<start_of_image>` is `vision_start_token_id` and `<end_of_image>` is `vision_end_token_id`.

ghubnerr:
Hey @kashif ! Sorry for the ping -- did you want me to review anything else? It's my first time contributing, so I'm not sure what the steps are for someone to give a Review approval 😄 Thanks for merging main again.

kashif:
no worries! I will test tomorow and get it merged

trl/trainer/grpo_trainer.py
```diff
@@ -1408,14 +1408,22 @@ def _generate_and_score_completions(
...
-                    re.sub(rf"({re.escape(self.image_token)})+", self.image_token, text) for text in prompts_text
-                ]
+                escaped_img_token = re.escape(trainer.image_token)
+                # Search for the image token in the chat template
+                if re.search(escaped_img_token, trainer.processing_class.chat_template):
+                    prompts_text = [re.sub(rf"({escaped_img_token})+", trainer.image_token, text) for text in prompts_text]
+                else:
+                    # If the chat template doesn't use the image token, we remove all instances of it + vision_end_token_id
+                    escaped_eoi_token = re.escape(trainer.processing_class.tokenizer.decode([trainer.vision_end_token_id]))
+                    prompts_text = [re.sub(rf"({escaped_img_token})+{escaped_eoi_token}", "", text) for text in prompts_text]
```
```suggestion
                escaped_img_token = re.escape(self.image_token)
                # Search for the image token in the chat template
                if re.search(escaped_img_token, self.processing_class.chat_template):
                    prompts_text = [re.sub(rf"({escaped_img_token})+", self.image_token, text) for text in prompts_text]
                else:
                    # If the chat template doesn't use the image token, we remove all instances of it + vision_end_token_id
                    escaped_eoi_token = re.escape(self.processing_class.tokenizer.decode([self.vision_end_token_id]))
                    prompts_text = [re.sub(rf"({escaped_img_token})+{escaped_eoi_token}", "", text) for text in prompts_text]
```

tests/test_grpo_trainer.py
```diff
@@ -1834,6 +1834,51 @@ def test_training_vlm_and_prompt_truncation(self):
...
+                vllm_mode="server",
+            )
+            trainer = GRPOTrainer(
+                model=model_id,
+                reward_funcs="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
+                args=training_args,
+                train_dataset=dataset,
+            )
+
+            if "Gemma3" in model_id:
```
sergiopaniego: Could this "if" be removed as this case is already managed inside the trainer?

tests/test_grpo_trainer.py
```diff
@@ -1834,6 +1834,51 @@ def test_training_vlm_and_prompt_truncation(self):
...
+                trainer.vision_start_token_id = trainer.model.config.boi_token_id
+                trainer.vision_end_token_id = trainer.model.config.eoi_token_id
```
sergiopaniego:
```suggestion
```

sergiopaniego (approved):
Thanks @ghubnerr!!

PR #3835, opened 2025-08-01
[GSPO]: Refactor `_compute_loss`

### LINKED ISSUE #3823: Loss not calculated correctly when using GSPO (`importance_sampling_level == "sequence"`)
### Reproduction

```python
from trl import GRPOConfig

training_args = GRPOConfig(
    importance_sampling_level="sequence",
    loss_type="grpo",
    beta=0.00,  
    epsilon=3e-4,
)
```

I believe that currently the GSPO loss is calculated in a different manner than in the paper. If i am misunderstanding something, please let me know.

The loss in the paper is:

Image

which means that each example has only one Advantage.

The current code is:
https://github.com/huggingface/trl/blob/dbbc770f45139b71ecb8a2871b56db1f3ca559d0/trl/trainer/grpo_trainer.py#L1859-L1896

if we use `importance_sampling_level == "sequence"` we get that `per_token_loss` shape is ` (B, 1) `. which makes sense. but when we go to calculate the loss, we multiply `per_token_loss` with `completion_mask` which gives us a shape of ` (B, T) ` (beacuse of broadcasting) which means that the advantage is multiplied by the number of non-padding tokens for each sample. and then we sum/mean to get the loss.

e.g:

Image

as far as i understand this is not how the calculation should be. and when we use GSPO we should just use `per_token_loss.mean() ` as the loss.

I think a similar issue arises if we use kl token loss. which will also have this broadcasting.

Maybe I am misunderstanding something, if that is the case. would love an explanation!

@qgallouedec  

### System Info

- Platform: Linux-5.14.0-503.21.1.el9_5.x86_64-x86_64-with-glibc2.34
- Python version: 3.11.12
- TRL version: 0.19.0.dev0+8f490b3
- PyTorch version: 2.7.0
- accelerator(s): NVIDIA A100-SXM4-80GB
- Transformers version: 4.53.0.dev0
- Accelerate version: 1.7.0
- Datasets version: 3.6.0
- HF Hub version: 0.33.4
- bitsandbytes version: 0.46.0
- Liger-Kernel version: 0.5.10
- OpenAI version: 1.90.0
- PEFT version: 0.15.2
- vLLM version: 0.9.2

### Checklist

pramodith:
# What does this PR do?
With the addition of GSPO the `GPRO` trainer deals with loss functions that operate at the token and entire sequence/response level.

Additionally recent variants of GRPO have eliminated the use of KL divergence in the loss. In this PR we refactor the `_compute_loss` function inside the trainer to distinguish between token level and sequence level loss and also start using the term `policy_loss` and distinguish it from the `kl_loss` instead of clubbing everything under `per_token_loss`.

Earlier the term `per_token_loss` was used even if GSPO loss was being computed, despite GSPO loss being a tensor of shape (B, 1) rather than (B, T). This can be confusing for readers. 

The PR also separates the kl_loss from the per_token_loss before aggregating them to make the logic of how the total loss is calculated a bit clearer.

An additional benefit of the refactor is that users won't need to explicitly set the `loss_type="grpo"`, when using GSPO.

Fixes #3823 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/grpo_trainer.py
```diff
@@ -1878,22 +1879,29 @@ def _compute_loss(self, model, inputs):
...
-        elif self.loss_type == "bnpo":
-            loss = (per_token_loss * completion_mask).sum() / completion_mask.sum().clamp(min=1.0)
-        elif self.loss_type == "dr_grpo":
-            loss = (per_token_loss * completion_mask).sum() / (per_token_loss.size(0) * self.max_completion_length)
+            policy_loss = policy_loss * entropy_mask
+
+        if self.importance_sampling_level == "sequence":
+            loss = policy_loss
+            if self.beta != 0.0:
+                loss += (per_token_kl * completion_mask).sum(-1) / completion_mask.sum(-1).clamp(min=1.0)
```
shirinyamani, line 1910: Hey @pramodith thanks for your contribution! If i understand correctly you wanna include the kl in loss calculation. 
here i see you wanna get the kl on sequence level then keep adding it to the loss, therefore the overall loss score will have the kl incase of `beta !=0` this to me seems like RLOO, isn't?  

i see the addition supports both sequence and token level!

pramodith: Hey Shirin, not sure if I understand you correctly. But the intention here is to be able to add the KL-div loss to the GSPO policy loss.

`policy_loss` -> Shape (B, 1)
`(per_token_kl * completion_mask).sum(-1) / completion_mask.sum(-1).clamp(min=1.0)` -> will reduce the KL loss to shape (B, 1) as well.

This has the same behavior as the current code

https://github.com/huggingface/trl/blob/072d7dd5a6e43424aee65d7c1ab8b74d672399f2/trl/trainer/grpo_trainer.py#L1883-L1891

Under the

```
if self.beta!=0.0:
```

Would convert per_token_loss to a tensor of shape (B,1 ) to (B, S). The naming convention of  `per_token_loss` before the self.beta!=0.0 block is confusing because if we use GSPO its per sequence loss rather than per token loss.

And the 

```
if loss_type=='grpo':
```     

block reduces the per_token_loss to a tensor of shape (B,1)

In the refactored code a user won't need to worry about setting loss type.

Let me know if this doesn't answer your question.

shirinyamani: @pramodith thanks for the explanation, yes exactly!  So overall you are trying to convert everything to sequence-level bc gspo is on seq-level rather than token-level (including the kl (seq-level) and loss (seq-level)), this reminded me of RLOO bc RLOO is also sequence-level!

we the current code, we need to document somewhere that when they pick grpo loss type, its on sequence level, but i get that it can be confusing!

qgallouedec:
Thanks @pramodith, and sorry for the long silence here, this one sat in our backlog.

Since it was opened, `_compute_loss` in `grpo_trainer.py` has been substantially rewritten: sequence-level importance sampling (GSPO) is now a first-class path, `policy_loss` is separated from the KL term and logged as its own metric, and users no longer need to set `loss_type="grpo"` for GSPO. So the goals of this refactor already landed in `main`.

The linked issue #3823 was fixed and closed back in June, and the branch now conflicts heavily with the current file. Given that, I'm going to close this as superseded, but thank you for pushing the naming/clarity that this was after.

PR #3838, opened 2025-08-02
Consolidate gather calls into single collective operation

LeonEricsson:
## Motivation

Many `*Trainer` classes synchronise metrics across ranks by issuing **multiple blocking `all_gather` collectives in sequence**. For example, `GRPOTrainer::_compute_loss` calls `self.accelerator.gather(...)` three times consecutively 

https://github.com/huggingface/trl/blob/5d914a41255d2a63e1cf0ee91a80b4199befd6d4/trl/trainer/grpo_trainer.py#L1928-L1935

Each call incurs fixed latency, so chaining them amplifies communication overhead.

## What does this PR do?

Introduces **`gather_metrics`**, a helper that:
* Buckets metric tensors that share `shape`, `dtype`, and `device`.
* Per bucket, launches one `all_gather` instead of one per metric, then redistributes the results back to their original names.

For now, this is only implemented in `GRPOTrainer`, but if this approach is-well received we can promote `gather_metrics` to a shared utility and roll it out to all trainers.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


avishaiElmakies:
Hi, did you benchmark this to test this PR effectiveness?

LeonEricsson:
> Hi, did you benchmark this to test this PR effectiveness?

I have not. In most cases the improvement is negligible, the metric tensors are tiny. Any measurable benefit would probably only be visible on large-scale clusters with many nodes. However, reducing the number of collective operations when possible is good practice; consolidating related operations improves hardware utilization and reduces unnecessary synchronization overhead.

avishaiElmakies:
OK, thanks for the information!

trl/trainer/grpo_trainer.py
```diff
@@ -1646,14 +1688,19 @@ def _generate_and_score_completions(
             self.state.num_input_tokens_seen += self.accelerator.gather(attention_mask.sum()).sum().item()
         self._metrics[mode]["num_tokens"] = [self.state.num_input_tokens_seen]
 
+        metrics_to_gather = {"completion_lengths": completion_lengths, "is_eos": is_eos.any(dim=1)}
+
+        gathered_metrics = self.accelerator.gather(metrics_to_gather)
+        self._metrics[mode]["num_tokens"] = [self.state.num_input_tokens_seen]
```
pramodith: Probably need to get rid of this line, since its a repeat of line 1689 and not using `gathered_metrics`?

trl/trainer/grpo_trainer.py
```diff
@@ -1855,14 +1901,22 @@ def masked_batch_mean(x):
...
+            "entropy": mean_entropy,
+            "low_clip": low_clip,
+            "high_clip": high_clip,
+            "clip_ratio": clip_ratio,
+        }
+
+        gathered_metrics = self.gather_metrics(metrics_to_gather)
+
+        # Extract gathered metrics and update the metrics dictionary
+        self._metrics[mode]["entropy"].append(gathered_metrics["entropy"].nanmean().item())
```
pramodith, line 1977: Might be cleaner if you rename the keys in `self._metrics` wherever needed and just iterate over the keys and append rather than listing each of the keys out manually.

trl/trainer/grpo_trainer.py
```diff
@@ -1024,6 +1024,48 @@ def _enable_gradient_checkpointing(self, model: PreTrainedModel, args: GRPOConfi
...
+        """
+        Efficiently gather metrics by stacking compatible tensors when possible.
+
+        Args:
+            metrics_to_gather (dict): Dictionary of {metric_name: tensor} to gather
+
+        Returns:
+            dict: Dictionary of {metric_name: gathered_tensor}
+        """
+        if not metrics_to_gather:
```
pramodith: Do we really need this check?

trl/trainer/grpo_trainer.py
```diff
@@ -1024,6 +1024,48 @@ def _enable_gradient_checkpointing(self, model: PreTrainedModel, args: GRPOConfi
...
+            return {}
+
+        shape_groups = {}
+        for name, tensor in metrics_to_gather.items():
+            key = (tuple(tensor.shape), tensor.dtype, tensor.device)
```
pramodith, line 1011: I'm wondering if the device needs to be included in the key here.

trl/trainer/grpo_trainer.py
LeonEricsson: no, debug code that slipped through

trl/trainer/grpo_trainer.py
LeonEricsson: i knew i had forgotten something, thanks.

trl/trainer/grpo_trainer.py
LeonEricsson: i'm wasn't certain either, but it feels nice as a safety precaution?

trl/trainer/grpo_trainer.py
LeonEricsson: Hmm, I might be misunderstanding you, but the only way I see this is possible is through something like

```python
metrics_to_gather = {
    "entropy": mean_entropy,
    "low_clip": low_clip,
    "high_clip": high_clip,
    "clip_ratio": clip_ratio,
}
gathered_metrics = self.gather_metrics(metrics_to_gather)

metric_mappings = [
    ("entropy", "entropy", lambda x: x.nanmean().item()),
    ("low_clip", "clip_ratio/low_mean", lambda x: x.nanmean().item()),
    ("low_clip", "clip_ratio/low_min", lambda x: nanmin(x).item()),
    ("high_clip", "clip_ratio/high_mean", lambda x: x.nanmean().item()),
    ("high_clip", "clip_ratio/high_max", lambda x: nanmax(x).item()),
    ("clip_ratio", "clip_ratio/region_mean", lambda x: x.nanmean().item()),
]

for source_key, target_key, agg_func in metric_mappings:
    self._metrics[mode][target_key].append(agg_func(gathered_metrics[source_key]))
```

which i don't really like tbh.

trl/trainer/grpo_trainer.py
pramodith: Cool, I suppose there's no harm in having it since it should be the same value for all the tensors running on a given device.

pramodith (approved):
LGTM!

qgallouedec:
I would be very curious to know the overhead involved. Can you measure it? If you don't have access to a distributed setup, I can take care of it, just let me know.

If it's less than, let's say 0.1%, in some standard training, I think we can leave it as it is.

PR #3841, opened 2025-08-04
Add py.typed

cyyever:
# What does this PR do?

Add `py.typed`. Otherwise, mypy says ```[mypy] Skipping analyzing "trl": module is installed, but missing library stubs or py.typed marker [import-untyped]```                               "


qgallouedec:
Thanks for your contribution. We don't use mypy for type checking, so I think we don't need this new file

cyyever:
Indeed that file is not specific to mypy and is required by PEP 561 as `Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing.`

qgallouedec (approved):
Makes sense, thanks

PR #3783, opened 2025-07-27
⚖️ Add vLLM server mode and VLM support to OnlineDPOTrainer

vaelev:
## Summary

Fixes #3782

This PR adds support for vLLM server mode to OnlineDPOTrainer, matching the functionality available in GRPO. Previously, OnlineDPO only supported "colocate" mode where vLLM runs in the same process. Now users can use an external vLLM server (launched with `trl vllm-serve`) for generation while keeping the training process separate.

### Key Changes

- **Configuration**: Added vLLM server mode parameters to `OnlineDPOConfig` (`vllm_mode`, `vllm_server_base_url`, `vllm_server_host`, `vllm_server_port`, etc.)
- **VLLMClient Integration**: Added server communication support using the existing `VLLMClient` class
- **Weight Synchronization**: Implemented model weight sync between training process and vLLM server via `_move_model_to_vllm`
- **Dual Mode Support**: Refactored generation logic to support both "server" and "colocate" modes cleanly
- **Tests**: Added comprehensive test coverage for server mode functionality

### Implementation Details

The implementation follows the same patterns as GRPO's vLLM integration to ensure consistency across the TRL library:

1. **Server Mode**: Uses `VLLMClient` to communicate with external vLLM server, synchronizes weights before generation
2. **Colocate Mode**: Maintains existing behavior with in-process vLLM `LLM` instance
3. **Error Handling**: Graceful fallback and proper error messages for configuration issues

## Test Plan

- [x] Basic import and configuration tests pass
- [x] Server mode initialization works correctly  
- [x] Weight synchronization functionality implemented
- [x] Generation logic handles both modes properly
- [x] Existing colocate mode behavior preserved
- [x] Code formatting and linting passes

PR #3866, opened 2025-08-07
Validate `vllm_mode` param in GRPO

### LINKED ISSUE #3865: Validate the `vllm_mode` is either `server` or `colocate`
### Feature request

Add a check for vllm_mode parameter in online methods so it can only be either `server` or `colocate`

### Motivation

Prevent user miswriting the flag that could cause weird behaviors

### Your contribution

Will add this support! :)

sergiopaniego:
# What does this PR do?

Validates `vllm_mode` value to prevent possible weird behaviors. It's also been added to Online DPO in #3783.
Fixes #3865

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -820,6 +820,8 @@ class GRPOTrainer(Trainer):
                     max_num_batched_tokens=4096,
                     model_impl=self.args.vllm_model_impl,
                 )
+            else:
+                raise ValueError(f"vllm_mode must be either 'server' or 'colocate', got '{self.vllm_mode}'.")
 
             # vLLM specific sampling arguments
             self.guided_decoding_regex = args.vllm_guided_decoding_regex
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -623,3 +623,7 @@ class GRPOConfig(TrainingArguments):
 
         if self.delta is not None and self.use_liger_loss:
             raise ValueError("Liger loss does not support two-sided GRPO loss yet.")
+
+        # Validate vLLM configuration
+        if self.vllm_mode not in ["server", "colocate"]:
+            raise ValueError(f"vllm_mode must be either 'server' or 'colocate', got '{self.vllm_mode}'.")


kashif:
how about switching back to the transformer mode when its not colocate or server mode? I am in 2 minds about that option as it could be a user error if use vllm is true and mode is wrong... so perhaps we should throw an exception...

sergiopaniego:
I see your point!
From a user perspective, I'd prefer that it throws an error since otherwise it could be training with a model version that I don't want (imagine a user that wants the vLLM model implementation for some reason) just because I had a typo 😅. 
If this approach is followed, a new `else` needs to be added in https://github.com/huggingface/trl/pull/3783 inside the trainer (in addition to the one in the config)

trl/trainer/grpo_trainer.py
```diff
@@ -820,6 +820,8 @@ def __init__(
                     max_num_batched_tokens=4096,
                     model_impl=self.args.vllm_model_impl,
                 )
+            else:
```
pramodith, line 823: I don't think we need a check in `grpo_trainer`. If no config object is passed, this check will never occur because `self.use_vllm`  will be false since the Config object sets this to false by default.

https://github.com/huggingface/trl/blob/2fb70902313c907a06f55aeeab9f81ce72caaff5/trl/trainer/grpo_trainer.py#L509-L512

If the user passes a config object, the check in the config file will raise the exception.

sergiopaniego: Fair point! 
The only scenario where this would be raised is something like:

```
training_args = GRPOConfig()
training_args.use_vllm = True
training_args.vllm_mode = 'colocated'
...
trainer = GRPOTrainer(
    ...
    args=training_args,
    ...
)
trainer.train()
```

Maybe I'm being overly cautious with the check? Open to discuss it 🤔

pramodith: Ahhh I see, it'd be quite strange for someone to do that but I don't have a strong opinion, so feel free to leave that check in.

qgallouedec: I would only keep this one. It's more consistent with the rest of the codebase

sergiopaniego: Updated!

Kirill-Kravtsov:
```
if self.vllm_mode == "server" and self.accelerator.is_main_process:
    ...
elif self.vllm_mode == "colocate":
    ...
else:
    raise ValueError(f"vllm_mode must be either 'server' or 'colocate', got '{self.vllm_mode}'.")
```
So now when `self.vllm_mode == "server"` it raises an error on every process except the main one when there are > 1 processes.
@qgallouedec @sergiopaniego

Kirill-Kravtsov:
Created a pr with to fix it - https://github.com/huggingface/trl/pull/3886

PR #3867, opened 2025-08-07
[GRPO] Truncated Importance Sampling to address rollout-training mismatch

LeonEricsson:
## Motivation
TRL provides the option of using vLLM for rollouts, enabling fast and scalable generation. However, the token probabilities for the generated completions—used in the GRPO objective—do not come directly from vLLM. Instead, these probabilities are recomputed by the training backend. It has been known for a [while](https://discuss.vllm.ai/t/numerical-difference-between-vllm-logprobs-and-huggingface-logprobs/151) that vLLM probabilities differ from Hugging Face, which ultimately means we inadvertently train off policy from our generation policy, despite using the same weights. A recent [blog post](https://fengyao.notion.site/off-policy-rl) highlights the effect of this discrepancy and proposes a solution in the form of an importance sampling factor

## What does this PR do?

Initially, we document the numerical differences in token probabilities when using vLLM.

Depending on the results, we may address the issue through the recommended Truncated Importance Sampling method.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


LeonEricsson:
First experiment results, these are with vLLM rollout (in server mode), and a vanilla training backend (no fsdp/deepspeed). 


Screenshot 2025-08-07 at 17 57 35

We observe a considerable difference, in line with the blog.

Run on [2ef3af6](https://github.com/huggingface/trl/pull/3867/commits/2ef3af617404e8f919cd931bc33932a4bcc1ac2c).

<details>
<summary> Code to run </summary>
```python

dataset = load_dataset("trl-lib/tldr", split="train[:50]")

training_args = GRPOConfig(
    num_train_epochs=1,
    steps_per_generation=4,
    num_generations=4,
    logging_steps=1,
    report_to="wandb"
)

trainer = GRPOTrainer(
    reward_funcs=reward_len,
    args=training_args,
    train_dataset=dataset,
)
trainer.train()
```
</details>

qgallouedec:
This is a very important one, thanks! Is it ready for review?

LeonEricsson:
> This is a very important one, thanks! Is it ready for review?

I'm wavering on how we want to address this. Either we keep recomputing and introduce the Truncated Importance Sampling approach from the blog, or we move away from recomputing and use vLLM logprobs directly, everywhere. Both are valid; I see this as more a question of which approach scales better

LeonEricsson:
Some training runs with TIS compared to baseline (TIS=-1). I wouldn't expect things to look much different, both are stable. Generally KL($\pi_{inference}, \pi_{training}$) is low.

For context this is single gpu training, on gsm8k.

Screenshot 2025-08-20 at 11 20 03

Screenshot 2025-08-20 at 11 19 15

Screenshot 2025-08-20 at 11 19 41

Screenshot 2025-08-20 at 11 23 29

LeonEricsson:
same experiments with PPO-IS

Screenshot 2025-08-24 at 16 29 20

trl/scripts/vllm_serve.py
```diff
@@ -464,9 +464,12 @@ class GenerateRequest(BaseModel):
         max_tokens: int = 16
         guided_decoding_regex: Optional[str] = None
         generation_kwargs: dict = field(default_factory=dict)
+        logprobs: Optional[int] = None
+        prompt_logprobs: Optional[int] = None
```
qgallouedec: What is this exactly?

trl/trainer/grpo_trainer.py
```diff
@@ -1742,7 +1757,13 @@ def _generate_and_score_completions(
         if has_images:
             self._logs["image"].extend(gather_object(images))
 
+        if old_per_token_logps is not None:
+            probs_diff = torch.exp(old_per_token_logps - rollout_old_per_token_logps)
+            self._metrics[mode]["Token Probability Difference/max"].append(probs_diff.max().exp().item())
+            self._metrics[mode]["Token Probability Difference/mean"].append(probs_diff.mean().exp().item())
+
         output = {
+            "prompt_texts": prompts_text,
```
qgallouedec: Is it needed?

trl/trainer/grpo_trainer.py
```diff
@@ -1461,17 +1464,26 @@ def _generate_and_score_completions(
                             max_tokens=self.max_completion_length,
                             guided_decoding_regex=self.guided_decoding_regex,
                             generation_kwargs=self.args.generation_kwargs,
+                            logprobs=0 if self.vllm_importance_sampling_correction else None,
```
LeonEricsson: Doing this is slightly more effective, but it means we need to deal with ouput not containing logprobs. Is it worth the trade-off?

LeonEricsson:
I've cleaned up the implementation now. 

There's one existing issue which is that vLLM sometimes spits out a NaN logprobs for the chosen token. This needs to be handled.

LeonEricsson:
we should update to use the final processed logprobs from vLLM, from https://github.com/vllm-project/vllm/pull/22387. prior versions of vLLM didn't support retrieving the sampled logprobs.

EDIT: the vllm patch has not been released yet, so we can hold of on that change for a future pr

trl/trainer/grpo_trainer.py
qgallouedec: I think so. Why not returning logprobs in all cases?

LeonEricsson: we could return logprobs in all cases, and change the line to:

`logprobs=0`

It would save us some work. But they would be discarded/unused if `self.vllm_importance_sampling_correction = False`

trl/scripts/vllm_serve.py
```diff
@@ -561,7 +573,19 @@ async def generate(request: GenerateRequest):
         # Flatten and combine all results
         all_outputs = list(chain.from_iterable(all_outputs))  # from list of list to single list
         completion_ids = [list(output.token_ids) for outputs in all_outputs for output in outputs.outputs]
-        return {"completion_ids": completion_ids}
+
+        response = {"completion_ids": completion_ids}
+
+        if request.logprobs is not None:
+            logprobs: list[list[float]] = [
+                [sanitize_logprob(next(iter(lp.values()))) for lp in output.logprobs]
```
qgallouedec: I think this is not correct. Because this is the kind of struture that logp has:

```

[
    {12095: Logprob(logprob=-0.4203352630138397, rank=1, decoded_token=" Paris")},
    {
        382: Logprob(logprob=-4.204380512237549, rank=4, decoded_token=".\n\n"),
        13: Logprob(logprob=-0.7043806910514832, rank=1, decoded_token="."),
    },
    {785: Logprob(logprob=-1.6503105163574219, rank=1, decoded_token="The")},
    {6722: Logprob(logprob=-0.22225184738636017, rank=1, decoded_token=" capital")},
    {315: Logprob(logprob=-0.01588992401957512, rank=1, decoded_token=" of")},
    {
        5616: Logprob(logprob=-3.9320058822631836, rank=9, decoded_token=" China"),
        9625: Logprob(logprob=-1.1820058822631836, rank=1, decoded_token=" France"),
    },
    {374: Logprob(logprob=-0.008184465579688549, rank=1, decoded_token=" is")},
    {26549: Logprob(logprob=-0.14375460147857666, rank=1, decoded_token=" Beijing")},
    {382: Logprob(logprob=-0.19107574224472046, rank=1, decoded_token=".\n\n")},
    {
        4854: Logprob(logprob=-4.891144275665283, rank=20, decoded_token="Each"),
        785: Logprob(logprob=-1.2036441564559937, rank=1, decoded_token="The"),
    },
    {
        4399: Logprob(logprob=-8.162134170532227, rank=172, decoded_token=" fall"),
        315: Logprob(logprob=-0.7871338129043579, rank=1, decoded_token=" of"),
    },
    {11: Logprob(logprob=-0.2011128067970276, rank=1, decoded_token=",")},
    {279: Logprob(logprob=-0.443490207195282, rank=1, decoded_token=" the")},
    {
        7718: Logprob(logprob=-9.457632064819336, rank=1087, decoded_token=" Common"),
        6722: Logprob(logprob=-2.207632064819336, rank=1, decoded_token=" capital"),
    },
    {
        7684: Logprob(logprob=-2.805245876312256, rank=4, decoded_token=" Good"),
        47588: Logprob(logprob=-1.6177459955215454, rank=1, decoded_token=" Era"),
    },
    {
        30414: Logprob(logprob=-10.371889114379883, rank=1986, decoded_token="fulness"),
        315: Logprob(logprob=-1.6843888759613037, rank=1, decoded_token=" of"),
    },
]
```

qgallouedec: Oh actually you're right I think. Basically, the first element in the dict is the sampled token. And sometimes it returns more tokens, when the sampled is not the most likely token.

LeonEricsson: yes, you're correct, its returning the sampled/chosen token, and the most likely token (rank 1).

I don't know why it's returning the most likely token though, the [documentation clearly states](https://docs.vllm.ai/en/stable/api/vllm/sampling_params.html#vllm.sampling_params.SamplingParams.logprobs) that "_If set to a non-None value, the result includes the log probabilities of the specified number of most likely tokens, as well as the chosen tokens._". And I've specified `logprobs=0`.

trl/trainer/grpo_trainer.py
```diff
@@ -1626,8 +1643,10 @@ def _generate_and_score_completions(
             # samples may come from an earlier version of the model. In that case, we need to track old_per_token_logps
             # for importance sampling. If the steps are aligned, importance sampling isn't necessary and we set
             # old_per_token_logps to None.
+            # When using vLLM, we always compute old_per_token_logps for importance sampling, it was shown that the
+            # distribution mismatch between vLLM and the training model can be large and harm the training.
             generate_every = self.args.steps_per_generation * self.num_iterations  # generation frequency
-            if self.args.gradient_accumulation_steps % generate_every != 0:
+            if self.args.gradient_accumulation_steps % generate_every != 0 or self.use_vllm:
```
LeonEricsson: correct me if im wrong but when self.args.gradient_accumulation_steps % generate_every are aligned, our objective reduces to _roughly_ vanilla PG and we don't have to concern ourselves with the mismatch problem. So we only have to apply TIS when we are doing multi step?

qgallouedec: even in single-step case, you have a mismatch between old (in this case, it's the current policy; but let's use "old" for consistency) and the vllm distribution. So in my understanding there is still an issue right?

qgallouedec:
I think we're good @LeonEricsson, right? Or is this PR still draft?

trl/trainer/grpo_trainer.py
LeonEricsson: right, agreed. 

so that means with vllm we'll always be using `old_per_token_logps` in the loss.

https://github.com/huggingface/trl/blob/edbe8234bc7e528f72ac76607de9d3e4753e2709/trl/trainer/grpo_trainer.py#L1587-L1591

do we want to update the comment?

trl/trainer/grpo_config.py
```diff
@@ -589,6 +598,23 @@ class GRPOConfig(TrainingArguments):
         default=False,
         metadata={"help": "Whether to use the Liger GRPO loss."},
     )
+    vllm_importance_sampling_correction: bool = field(
+        default=True,
```
LeonEricsson, line 602: @qgallouedec you want this enabled by default? I changed some if statements that were failing 
```
if self.vllm_importance_sampling_correction:
            output["importance_sampling_ratio"] = importance_sampling_ratio
```
to
```
if self.use_vllm and self.vllm_importance_sampling_correction:
            output["importance_sampling_ratio"] = importance_sampling_ratio
```

but we could also change the vllm_importance_sampling_correction default to false, and have it default enabled only when self.use_vllm is enabled?

qgallouedec: I think this

```python
if self.use_vllm and self.vllm_importance_sampling_correction:
    output["importance_sampling_ratio"] = importance_sampling_ratio
```

is better

qgallouedec (approved):
LGTM!

trl/trainer/grpo_trainer.py
qgallouedec: yes, good catch!

PR #3876, opened 2025-08-09
Fix reward_processing_classes validation in GRPOTrainer (#2839)

### LINKED ISSUE #2839: A bug in grpo_trainer.py
### Reproduction

**Issue Description:**

I’m reading the original code in the file grpo_trainer.py and found an issue related to the matching of reward_processing_classes and reward_funcs. The problem occurs when reward_processing_classes is a single object, which gets transformed into a list, but no length check is performed to ensure it matches the number of reward_funcs. This leads to unexpected behavior when there are multiple reward_funcs.

**Code in question:**
The issue appears in lines 225 to 234 in grpo_trainer.py:
```python

        # Reward processing class
        if reward_processing_classes is None:
            reward_processing_classes = [None] * len(reward_funcs)
        elif not isinstance(reward_processing_classes, list):
            reward_processing_classes = [reward_processing_classes]
        else:
            if len(reward_processing_classes) != len(reward_funcs):
                raise ValueError("The number of reward processing classes must match the number of reward functions.")

        for i, (reward_processing_class, reward_func) in enumerate(zip(reward_processing_classes, reward_funcs)):
              ...

```
**Bug Explanation:**

If reward_processing_classes is a pretrained object (not a list), it gets converted into a list with one element. However, no check is made to ensure that the number of elements in reward_processing_classes matches the number of reward_funcs.
The following code(line 234):
```
        for i, (reward_processing_class, reward_func) in enumerate(zip(reward_processing_classes, reward_funcs)):
```
will zip reward_processing_classes (which now has only one element) with reward_funcs. As a result, if len(reward_funcs) > 1, only the first reward_func will be processed, and the rest will be discarded.

**Expected Behavior:**
If reward_processing_classes is provided as a single object (not a list), it should automatically be expanded to match the number of reward_funcs.
If the length of reward_processing_classes is 1, the code should duplicate the object to match the length of reward_funcs.

I hope this issue can be addressed soon.

### System Info

trl = 0.14.0

### Checklist

- [ ] Any code provided is minimal, complete, and reproducible ([more on MREs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks))

chi2liu:
## What does this PR do?

Fixes #2839 - When a single `reward_processing_class` is provided for multiple reward functions in GRPOTrainer, it now correctly replicates the processing class for each reward function instead of silently ignoring all but the first reward function.

## Bug description

Previously, when users provided:
- Multiple reward functions (e.g., `[func1, func2, func3]`)
- A single reward processing class (not a list)

The code would convert the single processing class to a list with one element `[processing_class]`, then zip it with the reward functions. This resulted in only the first reward function being properly processed, while the rest were silently ignored.

## Solution

The fix now replicates the single processing class to match the number of reward functions:
```python
reward_processing_classes = [reward_processing_classes] * len(reward_funcs)
```

This ensures all reward functions get the same processing class, which is the expected behavior when providing a single processing class.

## Testing

Added comprehensive unit test `test_reward_processing_classes_replication` that verifies:
1. Single processing class is correctly replicated for multiple reward functions
2. List validation still works (mismatched lengths raise ValueError)
3. Correct list length continues to work as expected

All existing tests pass without regression.

## Before submitting
- [x] This PR fixes an issue
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request)
- [x] Did you make sure to update the documentation with your changes? (No doc changes needed)

## Who can review?

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1907,6 +1907,71 @@ class GRPOTrainerTester(unittest.TestCase):
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
 
+    def test_reward_processing_classes_replication(self):
+        """Test that a single reward_processing_class is replicated for multiple reward functions."""
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        # Define multiple reward functions
+        def reward_func1(prompts, completions, **kwargs):
+            return [1.0] * len(prompts)
+
+        def reward_func2(prompts, completions, **kwargs):
+            return [2.0] * len(prompts)
+
+        reward_funcs = [reward_func1, reward_func2]
+
+        # Create a single processing class (tokenizer)
+        single_processing_class = AutoTokenizer.from_pretrained("gpt2")
+        single_processing_class.pad_token = single_processing_class.eos_token
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            training_args = GRPOConfig(
+                output_dir=tmp_dir,
+                max_steps=1,
+                report_to="none",
+            )
+
+            # Test 1: Single processing class for multiple reward functions
+            trainer = GRPOTrainer(
+                model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
+                reward_funcs=reward_funcs,
+                reward_processing_classes=single_processing_class,  # Single object
+                args=training_args,
+                train_dataset=dataset,
+            )
+
+            # Verify that the processing classes were properly replicated
+            self.assertEqual(len(trainer.reward_processing_classes), len(reward_funcs))
+            for pc in trainer.reward_processing_classes:
+                self.assertEqual(pc, single_processing_class)
+
+            # Test 2: List with wrong length should raise ValueError
+            mismatched_processing_classes = [single_processing_class]  # Only one, but need two
+
+            with self.assertRaises(ValueError) as context:
+                GRPOTrainer(
+                    model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
+                    reward_funcs=reward_funcs,
+                    reward_processing_classes=mismatched_processing_classes,
+                    args=training_args,
+                    train_dataset=dataset,
+                )
+
+            self.assertIn("must match", str(context.exception))
+
+            # Test 3: Correct list length should work
+            correct_processing_classes = [single_processing_class, single_processing_class]
+
+            trainer2 = GRPOTrainer(
+                model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
+                reward_funcs=reward_funcs,
+                reward_processing_classes=correct_processing_classes,
+                args=training_args,
+                train_dataset=dataset,
+            )
+
+            self.assertEqual(len(trainer2.reward_processing_classes), len(reward_funcs))
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -609,7 +609,9 @@ class GRPOTrainer(Trainer):
         if reward_processing_classes is None:
             reward_processing_classes = [None] * len(reward_funcs)
         elif not isinstance(reward_processing_classes, list):
-            reward_processing_classes = [reward_processing_classes]
+            # If a single processing class is provided for multiple reward functions,
+            # replicate it for each reward function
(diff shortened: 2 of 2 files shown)


qgallouedec:
We don't expect a different number of elements in reward_funcs and reward_processing_class. I think we should raise and exception if this case occurs

chi2liu:
> We don't expect a different number of elements in reward_funcs and reward_processing_class. I think we should raise and exception if this case occurs

sure, updated it

tests/test_grpo_trainer.py
```diff
@@ -1907,6 +1907,84 @@ def test_training_sequence_importance_sampling(self):
...
+        reward_funcs = [reward_func1, reward_func2]
+
+        # Create a single processing class (tokenizer)
+        single_processing_class = AutoTokenizer.from_pretrained("gpt2")
+        single_processing_class.pad_token = single_processing_class.eos_token
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            training_args = GRPOConfig(
+                output_dir=tmp_dir,
+                max_steps=1,
```
```suggestion
```

tests/test_grpo_trainer.py
```diff
@@ -1907,6 +1907,84 @@ def test_training_sequence_importance_sampling(self):
                 new_param = trainer.model.get_parameter(n)
                 self.assertFalse(torch.equal(param, new_param), f"Parameter {n} has not changed.")
 
+    def test_reward_processing_classes_validation(self):
```
qgallouedec: Il not quite sure about this test. I think it would make more sense to pass two reward **models** (not function) and one reward processing class

tests/test_grpo_trainer.py
```diff
@@ -1907,6 +1907,84 @@ def test_training_sequence_importance_sampling(self):
...
+                    args=training_args,
+                    train_dataset=dataset,
+                )
+
+            self.assertIn("must match", str(context.exception))
+
+            # Test 3: Correct list length should work
+            correct_processing_classes = [single_processing_class, single_processing_class]
+
+            trainer2 = GRPOTrainer(
```
qgallouedec: Let's try to avoid two instance of trainer in one test case. If you want to test two cases, best is to have two tests

tests/test_grpo_trainer.py
chi2liu: updated

tests/test_grpo_trainer.py
chi2liu: updated

tests/test_grpo_trainer.py
chi2liu: updated

trl/trainer/grpo_trainer.py
```diff
@@ -609,10 +609,20 @@ def __init__(
...
+                    f"Received a single reward processing class instead."
+                )
             reward_processing_classes = [reward_processing_classes]
         else:
             if len(reward_processing_classes) != len(reward_funcs):
-                raise ValueError("The number of reward processing classes must match the number of reward functions.")
+                raise ValueError(
+                    f"The number of reward processing classes ({len(reward_processing_classes)}) must match "
+                    f"the number of reward functions ({len(reward_funcs)})."
+                )
```
```suggestion
        elif not isinstance(reward_processing_classes, list):
            reward_processing_classes = [reward_processing_classes]
        if len(reward_processing_classes) != len(reward_funcs):
            raise ValueError(
                f"The number of reward processing classes ({len(reward_processing_classes)}) must match the number of "
                f"reward functions ({len(reward_funcs)})."
            )
```

maybe this is simpler?

tests/test_grpo_trainer.py
```diff
@@ -1907,6 +1907,136 @@ def test_training_sequence_importance_sampling(self):
...
+
+        # Use two reward models
+        reward_models = [
+            "trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
+            "trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
+        ]
+
+        # Create a single processing class (tokenizer)
+        single_processing_class = AutoTokenizer.from_pretrained("gpt2")
+        single_processing_class.pad_token = single_processing_class.eos_token
```
qgallouedec: why using gpt2 here? just use the same model (`"trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5"`)

trl/trainer/grpo_trainer.py
chi2liu: updated

tests/test_grpo_trainer.py
chi2liu: updated

qgallouedec (approved):
thanks!

PR #3879, opened 2025-08-11
Fix prompt truncation for multimodal inputs with multiple image tokens

artem-spector:
# What does this PR do?

This PR fixes an issue in prompt truncation when max_prompt_length is set for multimodal inputs.
Previously, the code generated prompt_inputs by calling the entire processing_class (processor) on prompts_text and images.
When the processor included image inputs, it could insert multiple image tokens into the tokenized sequence.
This sometimes caused truncate_with_protected_tokens to fail when reducing the sequence to max_prompt_length because the multiple protected tokens consumed too much of the allowed space.
Now we call only the tokenizer on the prompts_texts, thus ensuring that only the textual prompt is tokenized for truncation, while still respecting protected tokens.

## Before submitting
- [ x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1380,21 +1380,21 @@ class GRPOTrainer(Trainer):
 
         prompts_text = [maybe_apply_chat_template(example, self.processing_class)["prompt"] for example in inputs]
 
-        prompt_inputs = self.processing_class(
-            text=prompts_text,
-            return_tensors="pt",
-            padding=True,
-            padding_side="left",
-            add_special_tokens=False,
-            **kwargs,
-        )
-        prompt_inputs = super()._prepare_inputs(prompt_inputs)
-        prompt_ids, prompt_mask = prompt_inputs["input_ids"], prompt_inputs["attention_mask"]
-
         if self.max_prompt_length is not None:
             # If max_prompt_length is set, we trim the prompt to keep only the last `max_prompt_length` tokens.
             # Then we decode those tokens back into text. We manually remove leading pad tokens from the decoded text,
             # because we can't use `skip_special_tokens=True` (some special tokens are still needed for generation).
+            tokenizer = self.processing_class.tokenizer if has_images else self.processing_class
+            prompt_inputs = tokenizer(
+                prompts_text,
+                return_tensors="pt",
+                padding=True,
+                padding_side="left",
+                add_special_tokens=False
+            )
+            prompt_inputs = Trainer._prepare_inputs(self, prompt_inputs)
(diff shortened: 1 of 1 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -1380,21 +1380,21 @@ def _generate_and_score_completions(
...
             # because we can't use `skip_special_tokens=True` (some special tokens are still needed for generation).
+            tokenizer = self.processing_class.tokenizer if has_images else self.processing_class
+            prompt_inputs = tokenizer(
+                prompts_text,
+                return_tensors="pt",
+                padding=True,
+                padding_side="left",
+                add_special_tokens=False
+            )
+            prompt_inputs = Trainer._prepare_inputs(self, prompt_inputs)
```
pramodith:
```suggestion
            prompt_inputs = super()._prepare_inputs(self, prompt_inputs)
```

pramodith: This block of code also ensures that prompts_text never contains more than one `image_token` so I'm wondering where you were running into failures, an example of where the current code breaks would be helpful.

https://github.com/huggingface/trl/blob/de27d612b026526ba39b88eee348994d7636e033/trl/trainer/grpo_trainer.py#L1416-L1434

artem-spector: The problem is that this code:
        prompt_inputs = self.processing_class(
            text=prompts_text,
            return_tensors="pt",
            padding=True,
            padding_side="left",
            add_special_tokens=False,
            **kwargs,
        )
may generate input_ids that include both prompt and image tokens.
For example, LlavaNextProcessor for a short prompt and an image 459x320 gives prompt_ids of length 1822. Most of those are image_tokens, and truncate_with_protected_tokens would fail trying to truncate it to 512.

In the fixed version 
            prompt_inputs = tokenizer(
                prompts_text,
                return_tensors="pt",
                padding=True,
                padding_side="left",
                add_special_tokens=False
            )
the tokenizer processes only the prompt text, and the prompt_ids length is 41, nothing to truncate

qgallouedec:
So if I understand correctly, instead of ensuring that the max number of tokens is on text tokens + image tokens, you apply it only on text token?

artem-spector:
> So if I understand correctly, instead of ensuring that the max number of tokens is on text tokens + image tokens, you apply it only on text token?

correct

qgallouedec:
The issue though is that the final sequence (the one taken by the model as input) may be longer than `max_length`

PR #3883, opened 2025-08-12
Add tests for get_position_ids_from_packed_seq_lengths

pramodith:
# What does this PR do?

This PR does some refactoring to move the `_convert_seq_lengths_to_position_ids` in sft_trainer.py into the utils.py file and rename it to `get_position_ids_from_packed_seq_lengths`.

Given this function isn't straight forward, it seemed that it would benefit from adding test cases to ease understanding and ensure correct behavior.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/utils.py
```diff
@@ -1541,7 +1541,6 @@ def print_prompt_completions_sample(
     │ │ The sun is │  in the sky. │        0.46 │   0.10 │      0.65 │ │
     │ └────────────┴──────────────┴─────────────┴────────┴───────────┘ │
     ╰──────────────────────────────────────────────────────────────────╯
-    ```
```
qgallouedec: I think this should be reverted

qgallouedec:
When a function is only used once in the codebase, I think it's better not to have it in utils (should be reserved for shared function).
Testing this function is useful though. What about turning the method into a public static method (instead of protected static method) and test it in `TestDataCollatorForLanguageModeling`?

pramodith:
> When a function is only used once in the codebase, I think it's better not to have it in utils (should be reserved for shared function). Testing this function is useful though. What about turning the method into a public static method (instead of protected static method) and test it in `TestDataCollatorForLanguageModeling`?

I'm open to this but from my pov, this is a function that can easily be re-used in the future since packing sequences is a common operation that isn't super specific just to SFT as in, there can be scenarios in the future where a certain algorithm needs packing.

trl/trainer/utils.py
pramodith: Good catch, no idea how that even changed.

qgallouedec:
> I'm open to this but from my pov, this is a function that can easily be re-used in the future since packing sequences is a common operation that isn't super specific just to SFT as in, there can be scenarios in the future where a certain algorithm needs packing.

Agreed 100%, but to avoid bloat, I’d suggest moving this function outside the class only when it’s being reused, not before.

pramodith:
Done! Moved it back inside the DataCollator and added test cases in its corresponding class.

trl/trainer/sft_trainer.py
```diff
@@ -53,7 +53,11 @@
 )
 from ..models import clone_chat_template, get_act_offloading_ctx_manager, prepare_peft_model
 from .sft_config import SFTConfig
-from .utils import generate_model_card, get_comet_experiment_url, pad
+from .utils import (
+    generate_model_card,
+    get_comet_experiment_url,
+    pad,
+)
```
```suggestion
from .utils import generate_model_card, get_comet_experiment_url, pad,
```

trl/trainer/sft_trainer.py
```diff
@@ -249,15 +253,27 @@ def torch_call(self, examples: list[Union[list[int], Any, dict[str, Any]]]) -> d
...
     @staticmethod
-    def _convert_seq_lengths_to_position_ids(batch_seq_lengths: list[list[int]]) -> list[torch.Tensor]:
+    def get_position_ids_from_packed_seq_lengths(batch_seq_lengths: list[list[int]]) -> list[torch.Tensor]:
+        """
+        Get position IDs for packed sequences.
+        Args:
+            batch_seq_lengths (list[list[int]]): A list of lists containing the lengths of each
+                individual document in the packed batch.
+        Return:
+            list[torch.Tensor]: A list of tensors containing the position IDs for each packed sequence.
```
```suggestion

        Args:
            batch_seq_lengths (`list[list[int]]`):
                A list of lists containing the lengths of each individual document in the packed batch.

        Return:
            `list[torch.Tensor]`:
                A list of tensors containing the position IDs for each packed sequence.
```

trl/trainer/sft_trainer.py
```diff
@@ -249,15 +253,27 @@ def torch_call(self, examples: list[Union[list[int], Any, dict[str, Any]]]) -> d
...
+        # Flat list of lengths
         batch_seq_lengths = torch.tensor(
             [seq_length for seq_lengths in batch_seq_lengths for seq_length in seq_lengths]
         )
         position_ids = torch.ones(sum(example_lengths), dtype=batch_seq_lengths.dtype)
         position_ids[0] = 0
+        # Reset position ids to 0 at the start of each sequence
         position_ids[batch_seq_lengths[:-1].cumsum(0)] = -(batch_seq_lengths[:-1] - 1)
         position_ids = position_ids.cumsum(0)
+        # Reshape to match batch size
```
```suggestion
        # Split back into one tensor per example
```

tests/test_sft_trainer.py
```diff
@@ -222,6 +222,26 @@ def test_assistant_masks(self):
...
+        batch_seq_lengths = [[5]]
+        result = DataCollatorForLanguageModeling.get_position_ids_from_packed_seq_lengths(batch_seq_lengths)
+        self.assertEqual(len(result), 1)
+        self.assertTrue(torch.equal(result[0], torch.arange(5)))
+
+    def test_single_example_multiple_docs(self):
+        batch_seq_lengths = [[3, 2]]
+        result = DataCollatorForLanguageModeling.get_position_ids_from_packed_seq_lengths(batch_seq_lengths)
+        self.assertEqual(len(result), 1)
+        # First doc: 0,1,2; second doc: 0,1
```
```suggestion
        # First sequence: 0, 1, 2; second sequence: 0, 1
```

trl/trainer/sft_trainer.py
```diff
@@ -53,7 +53,7 @@
 )
 from ..models import clone_chat_template, get_act_offloading_ctx_manager, prepare_peft_model
 from .sft_config import SFTConfig
-from .utils import generate_model_card, get_comet_experiment_url, pad
+from .utils import generate_model_card, get_comet_experiment_url, pad,
```
```suggestion
from .utils import generate_model_card, get_comet_experiment_url, pad
```

qgallouedec (approved):
nice!

PR #3884, opened 2025-08-12
Minor optimizations in SFT.

pramodith:
# What does this PR do?
A few minor code optimizations to improve run time and memory consumed.

1. Check if the dataset is conversational just once rather than in every invocation of dataset.map.
2. Add a `no.grad()` block when computing accuracy for logging.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -700,10 +700,12 @@ class SFTTrainer(Trainer):
 
                 dataset = dataset.map(_func, batched=False, **map_kwargs)
 
+            is_dataset_conversational = False
             if not is_processed:
                 # Convert the dataset to ChatML if needed
                 first_example = next(iter(dataset))
                 if is_conversational_from_value(first_example):
+                    is_dataset_conversational = True
                     if isinstance(dataset, Dataset):  # `IterableDataset.map` does not support `desc`
                         map_kwargs["desc"] = f"Converting {dataset_name} dataset to ChatML"
                     column_names = next(iter(dataset)).keys()
@@ -732,6 +734,8 @@ class SFTTrainer(Trainer):
                         remove_columns="messages" if "messages" in column_names else None,  # renamed to "text"
                         **map_kwargs,
                     )
+                else:
+                    is_dataset_conversational = True
 
                 # Tokenize the dataset
                 if isinstance(dataset, Dataset):  # `IterableDataset.map` does not support `desc`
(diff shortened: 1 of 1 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -740,7 +744,7 @@ def add_eos(example, eos_token):
                 def tokenize(example, processing_class, dataset_text_field, assistant_only_loss):
                     if "prompt" in example:  # prompt-completion case
                         output = {}
-                        if is_conversational(example):
+                        if is_dataset_conversational:
```
qgallouedec: By doing this, I'm afraid that since a variable outside the scope of the function is used, it breaks the possibility for `map` to use the fingerprint to cache/retrieve the result. I'll check quickly.

pramodith: I think inner functions should have access to variables declared in the outer function, I can also just pass it in `fn_kwargs` if it breaks.

pramodith: I did a quick check by adding
```
output["is_dataset_conversational"] = is_dataset_conversational
```
at the end of `tokenize` for a conversational dataset and printed out that value after the `map` function is applied and it does correctly map to `True`.

qgallouedec: Yes, it looks like caching is working as expected.

After reconsidering, I actually prefer the original approach. It keeps each *section* self-contained—even if it sacrifices a bit of performance. That said, the impact should be very minimal since it’s just a dictionary access. Would you mind reverting to the initial style?

pramodith: Sure will do in a couple of hours

pramodith: Decided to add `is_dataset_conversational` as an argument to `tokenize`, that way the tokenize function remains self-contained and we avoid the redundant ops.

qgallouedec: What I meant by self-contained is that it doesn't depend on variables stored previously. It's not critical, but with this function I found that it was more readable and practical to work that way.
I ran a small benchmark on a dataset of 158k lines, and you gain 2 seconds (0.6%). I don't think it's worth breaking up the section partitioning for that.

pramodith:
@qgallouedec  I can make the changes to set `batched=True` in another PR tomorrow, when I get the time if you'd like to merge this one today.

qgallouedec:
Yep, best is to keep things separate 👌

PR #3888, opened 2025-08-13
🌳 Enhance segment tree implementation for non-power-of-2 values

### LINKED ISSUE #3887: Bug in BFD packing
### Reproduction

BFD packing does not perform packing for certain values of max_length. I have tracked down the problem to the following minimum example

```python
import trl
import datasets
import numpy as np
from datasets.utils.logging import disable_progress_bar
disable_progress_bar()

# Create input_ids with zero values, of shape
N = 3000
lens = np.random.normal(4096, 512, N).astype(int).clip(100, 8192).tolist()
input_ids = [[0 for _ in range(n)] for n in lens]
dataset = datasets.Dataset.from_dict({'input_ids': input_ids}).with_format("arrow")

pack = lambda seq_len: dataset.map(trl.data_utils._pack_bfd, batched=True, fn_kwargs={"seq_length": seq_len})

print('Input length:', len(dataset))
print('Trying packing into powers of 2 ------')
print('Output length:', len(pack(8192)))
print('Output length:', len(pack(8192*2)))
print('Output length:', len(pack(8192*4)))
print('Trying powers of 2 + 1------')
print('Output length:', len(pack(8192+1)))
print('Output length:', len(pack(8192*2+1)))
print('Output length:', len(pack(8192*4+1)))
```

outputs:

```
Input length: 3000
Trying packing into powers of 2 ------
Output length: 1524
Output length: 795
Output length: 385
Trying powers of 2 + 1------
Output length: 3000
Output length: 3000
Output length: 3000
```

As you can see, for 8192*n + 1, the dataset is not being packed.

### System Info

- Platform: Linux-5.15.0-89-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- TRL version: 0.22.0.dev0+7f53ac0
- PyTorch version: 2.5.1
- accelerator(s): NVIDIA H100
- Transformers version: 4.55.0
- Accelerate version: 1.10.0
- Datasets version: 3.0.1
- HF Hub version: 0.34.4
- DeepSpeed version: 0.15.1
- Liger-Kernel version: 0.6.1
- OpenAI version: 1.60.2
- PEFT version: 0.12.0
- vLLM version: 0.7.0

### Checklist

MengAiDev:
# What does this PR do?

- Properly handle non-power-of-2 seq_length values by rounding up to the next power of 2 for the tree size
- This ensures correct functionality of the search method for all seq_length values
- Improves bin packing for all seq_length values, not just powers of 2

Fixes #3887 

## Before submitting

      Pull Request section?
      to it if that's the case.


trl/data_utils.py
```diff
@@ -517,16 +528,24 @@ def _pack_bfd(examples: pa.Table, seq_length: int) -> pa.Table:
...
             bin = space_to_bin[space].popleft()
         else:
+            # Create a new bin
             bin = {"ids": [], "length": 0}
             bins.append(bin)
 
         bin["ids"].append(idx)
         bin["length"] += length
-        if space < seq_length and not space_to_bin[space]:
+        if space_to_bin[space] and not space_to_bin[space]:
```
pramodith: This condition will always be False 😅 `x and not x`

trl/data_utils.py
```diff
@@ -491,7 +498,11 @@ def search(self, val):
 
 
 def _pack_bfd(examples: pa.Table, seq_length: int) -> pa.Table:
-    """Pack sequences in a pyarrow Table using Best Fit Decreasing strategy."""
+    """Pack sequences in a pyarrow Table using Best Fit Decreasing strategy.
+    
+    Fix: Ensure segment tree properly handles non-power-of-2 seq_length values by using a correctly
```
pramodith: We can get rid of the new comments here, since its added in the init of SegmentTree.

trl/data_utils.py
```diff
@@ -517,16 +528,24 @@ def _pack_bfd(examples: pa.Table, seq_length: int) -> pa.Table:
     bins: list[dict] = []
     for length, idx in zip(lengths.field(0).to_numpy(), lengths.field(1).to_numpy()):
         space = segment_tree.search(length)
-
-        if space < seq_length:
+        
+        # Handle case where search returns 0 (no suitable bin found)
+        # This can happen with certain seq_length values
+        if space == 0 or length > space:
```
pramodith: I don't think this block of code changes is required.

`space < seq_length == False` (the else block) covers the scenario of `space == 0`

if `length>seq_length` `length` will always be greater than `space`.
if `length<seq_length` `length` can never be greater than `space`

tests/test_bfd_packing_fix.py
```diff
@@ -0,0 +1,58 @@
+import unittest
```
pramodith: These test cases should be moved to `TestPackDatasetBfd` in `test_data_utils.py`

tests/test_bfd_packing_fix.py
```diff
@@ -0,0 +1,58 @@
...
+    def test_bfd_packing_identical_results(self):
+        """Test that BFD packing gives similar results for power of 2 and power of 2 + 1."""
+        # Create input_ids with varying lengths
+        N = 50
+        lens = np.random.normal(2048, 256, N).astype(int).clip(100, 4096).tolist()
+        input_ids = [[0 for _ in range(n)] for n in lens]
+        dataset = datasets.Dataset.from_dict({'input_ids': input_ids}).with_format("arrow")
+        
+        # Pack with power of 2 and power of 2 + 1
+        packed_8192 = trl.data_utils.pack_dataset(dataset, seq_length=8192, strategy="bfd")
```
pramodith: To make this more robust, I'd create a new test where N different seq_lengths are randomly chosen.

pramodith (changes requested):
Thanks for the fix! Had a few suggestions.

tests/test_data_utils.py
```diff
@@ -765,6 +770,66 @@ def test_with_truncation(self):
...
+        input_ids = [[0 for _ in range(n)] for n in lens]
+        dataset = Dataset.from_dict({'input_ids': input_ids}).with_format("arrow")
+        
+        # Pack with power of 2 + 1
+        packed_dataset = pack_dataset(dataset, seq_length=8193, strategy="bfd")
+        
+        # Verify that packing occurred (output length should be less than input)
+        self.assertLess(len(packed_dataset), len(dataset))
+        
+    def test_bfd_packing_identical_results(self):
```
pramodith: This test case seems to  be a combination of the two tests above so I don't think it adds any values.

tests/test_data_utils.py
```diff
@@ -765,6 +770,66 @@ def test_with_truncation(self):
...
         dataset = pack_dataset(dataset, seq_length, strategy="bfd")
         self.assertEqual(dataset.to_dict(), expected_output)
+        
+    def test_bfd_packing_power_of_2(self):
+        """Test BFD packing with power of 2 seq_length."""
+        # Create input_ids with varying lengths
+        N = 100
+        lens = np.random.normal(4096, 512, N).astype(int).clip(100, 8192).tolist()
+        input_ids = [[0 for _ in range(n)] for n in lens]
+        dataset = Dataset.from_dict({'input_ids': input_ids}).with_format("arrow")
```
pramodith: Let's make this `dataset` a pytest fixture so that we can re-use the same dataset across all the test cases, to make things cleaner.

tests/test_data_utils.py
```diff
@@ -765,6 +770,66 @@ def test_with_truncation(self):
         }
         dataset = pack_dataset(dataset, seq_length, strategy="bfd")
         self.assertEqual(dataset.to_dict(), expected_output)
+        
+    def test_bfd_packing_power_of_2(self):
```
pramodith: The test cases above this one do test `bfd` for powers of 2, so I think we can remove this specific test case.

trl/data_utils.py
```diff
@@ -517,16 +521,23 @@ def _pack_bfd(examples: pa.Table, seq_length: int) -> pa.Table:
...
             bin = space_to_bin[space].popleft()
         else:
+            # Create a new bin
             bin = {"ids": [], "length": 0}
             bins.append(bin)
 
         bin["ids"].append(idx)
         bin["length"] += length
-        if space < seq_length and not space_to_bin[space]:
+        # Fix the condition that was always False
```
pramodith: please get rid of the comment. Also an empty deque is evaluated to False when used as a condition. So `space_to_bin[space]` will always be the same as `len(space_to_bin[space]) == 0` so this condition will always be False too. I think we can just retain the old condition 

`if space < seq_length and not space_to_bin[space]`

pramodith: The goal of this if condition originally is to remove the `space` from the `SegmentTree` if there aren't any bins with a document `space` tokens long.

trl/data_utils.py
pramodith: Still need some clarity on this.

MengAiDev:
I have fixed it, and the test all passed (except one of cannot access huggingface.co)

pramodith:
@MengAiDev I made a few changes to your code by adding a new test case for when the examples length is greater than the seq length, in this case we should have no packing.

Restored some of the original conditions, in the place where you had an 
`if ... elif ... else` block since I didn't think they were required and were captured correctly by the original code.

If you think that I'm missing something feel free to revert/update the code.

pramodith:
@qgallouedec can you please take a look at this one when you get the chance.

trl/data_utils.py
```diff
@@ -457,11 +457,15 @@ class _SegmentTree:
 
     def __init__(self, maxval: int):
         self.maxval = maxval
-        self.tree = [0] * (2 * maxval)
+        # For non-power-of-2 values, we need to round up to the next power of 2 for the tree size
+        self.tree_size = 1
+        while self.tree_size < maxval:
+            self.tree_size <<= 1
```
qgallouedec:
```suggestion
        self.tree_size = 1 << (maxval - 1).bit_length()
```

qgallouedec (approved):

PR #3891, opened 2025-08-13
♻️ `use_cache` should be set in the forward pass

qgallouedec:
To suppress this warning:

```
use_cache=True` is incompatible with gradient checkpointing. Setting `use_cache=False`...
```


trl/trainer/sft_trainer.py
```diff
@@ -975,6 +975,8 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
         Compute training loss and additionally compute token accuracies
         """
         mode = "train" if self.model.training else "eval"
+        # If not set, defaults from model config and may warn since cache isn't compatible with gradient checkpointing
+        inputs["use_cache"] = False
```
lewtun, line 979: Note that we may want to set `use_cache=True` at the end of the SFT/DPO scripts, because otherwise I believe the final model with have `use_cache=False` and much slower inference.

PR #3895, opened 2025-08-14
[SFTTrainer]: Use batched=True for datasets.map

pramodith:
# What does this PR do?
Uses batched map operations to improve the time it takes to prepare a SFT dataset.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/tests/test_data_utils.py
+++ b/tests/test_data_utils.py
@@ -849,21 +849,13 @@ class TestMaybeConvertToChatML(TrlTestCase):
             {"from": "user", "value": "What is the capital of France?"},
             {"from": "assistant", "value": "The capital of France is Paris."},
         ]
-        example = {
-            "conversations": conversation_1
-        }
-        expected_output = {
-            "messages": conversation_1
-        }
+        example = {"conversations": conversation_1}
+        expected_output = {"messages": conversation_1}
         self.assertEqual(maybe_convert_to_chatml(example), expected_output)
 
         # Test for batched examples
-        examples = {
-            "conversations": [conversation_1, conversation_2]
-        }
-        expected_output = {
-            "messages": [conversation_1, conversation_2]
-        }
+        examples = {"conversations": [conversation_1, conversation_2]}
+        expected_output = {"messages": [conversation_1, conversation_2]}
         self.assertEqual(maybe_convert_to_chatml(examples), expected_output)
 
     def test_without_conversations_key(self):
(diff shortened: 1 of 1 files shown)


pramodith:
@qgallouedec this one is ready for review.

trl/trainer/sft_trainer.py
```diff
@@ -942,15 +943,19 @@ def _func(example):
...
+
+                        if "text" in examples:  # language modeling case
+                            examples["text"] = [append_eos(text) for text in examples["text"]]
+                        elif "completion" in examples:
+                            examples["completion"] = [append_eos(completion) for completion in examples["completion"]]
+                        return examples
 
                     dataset = dataset.map(
                         add_eos,
+                        batched=True,
```
qgallouedec, line 958: in my understanding, making this batched won't speed things up, because at the end of the day, we still have a for loop. If I'm right on this, I'd revert this change

qgallouedec:
I think you're still not leveraging the batched implementation of the tokenizer. To simplify:

```python

import time

processor = AutoProcessor.from_pretrained("Qwen/Qwen3-0.6B")
dataset = load_dataset("trl-lib/Capybara", split="train")


def currently(example, processor):
    output = processor.apply_chat_template(
        example["messages"],
        return_dict=True,
    )
    return output

before = time.time()
dataset0 = dataset.map(currently, fn_kwargs={"processor": processor})
after = time.time()
print(f"Processing time: {after - before} seconds")

def what_you_want(examples, processor):
    texts = []
    for messages in examples["messages"]:
        text = processor.apply_chat_template(
            messages,
            tokenize=False,
        )
        texts.append(text)
    output = processor(texts, add_special_tokens=False)
    return output

before = time.time()
dataset1 = dataset.map(what_you_want, batched=True, fn_kwargs={"processor": processor})
after = time.time()
print(f"Processing time: {after - before} seconds")
```

This is the speedup that I get:

```python
Processing time: 37.06554651260376 seconds
Processing time: 9.677833080291748 seconds
```

trl/trainer/sft_trainer.py
```diff
@@ -932,6 +932,7 @@ def _func(example):
                     column_names = next(iter(dataset)).keys()
                     dataset = dataset.map(
                         maybe_convert_to_chatml,
+                        batched=True,
```
qgallouedec, line 935: same here, I don't think using batched here speeds things. Unless it does (to confirm), I'd revert this

PR #3900, opened 2025-08-14
[SFTTrainer]: Compute accuracy in log to avoid gather every step.

pramodith:
# What does this PR do?

Computes the accuracy in the `log` function to avoid `gather` calls after every forward step.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -945,50 +945,38 @@ class SFTTrainer(Trainer):
 
     def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=None):
         """
-        Compute training loss and additionally compute token accuracies
+        Compute training loss and collect token statistics for later aggregation.
         """
         mode = "train" if self.model.training else "eval"
         (loss, outputs) = super().compute_loss(
             model, inputs, return_outputs=True, num_items_in_batch=num_items_in_batch
         )
+
         if mode == "train":
-            # When using padding-free, the attention_mask is not present in the inputs, instead we have cu_seq_lens_q,
-            # cu_seq_lens_k, and max_length_k, max_length_q and position_ids.
             if "attention_mask" in inputs:
-                num_tokens_in_batch = self.accelerator.gather_for_metrics(inputs["attention_mask"].sum()).sum().item()
+                num_tokens_in_batch = inputs["attention_mask"].sum().item()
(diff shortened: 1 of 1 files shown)


qgallouedec:
So ideally the log should just _log_ (and not compute metrics), but if it really helps improving the speed, maybe we can make an exception here. Do you have a benchmark, to see it speeds things up?

pramodith:
> So ideally the log should just _log_ (and not compute metrics), but if it really helps improving the speed, maybe we can make an exception here. Do you have a benchmark, to see it speeds things up?

I ran this code block with and without the changes and only got a 0.3 second improvement (might be noise) so I guess its not worth it?

GPUs -> 2 A10s
<details>
<summary>Code</summary>

```
from peft import LoraConfig
import time
from accelerate import PartialState


device_map="DDP" # for DDP and running with 

if device_map == "DDP":
    device_string = PartialState().process_index
    device_map={'':device_string}

    "Qwen/Qwen3-0.6B",
    trust_remote_code=True,
    use_cache = False,
    device_map = device_map,  # Use DDP device map
)

config = SFTConfig(
    max_steps=100,
    logging_steps=10,
    gradient_checkpointing=True,
    gradient_checkpointing_kwargs = {"use_reentrant": False}, #must be false for DDP
)


trainer = SFTTrainer(
    model=model,
    args=config,
    train_dataset=load_dataset("trl-lib/Capybara", split="train"),
    peft_config=LoraConfig(),
    
)

start_time = time.time()
trainer.train()
print(f"Training took {time.time() - start_time} seconds")
```
</details>

qgallouedec:
I'll try with 2 nodes of 8 H100

qgallouedec:
After extensive benchmarking, I found that gathering introduces negligible overhead. With your code and 2 nodes of 8×H100, the gathering time is only about 0.9 second / 673 seconds (0.13% of training time). Plus, here, we’re training a small model, this is a worst-case scenario.

pramodith:
> After extensive benchmarking, I found that gathering introduces negligible overhead. With your code and 2 nodes of 8×H100, the gathering time is only about 0.9 second / 673 seconds (0.13% of training time). Plus, here, we’re training a small model, this is a worst-case scenario.

Thanks for running the benchmark! This is an interesting and good learning for me, didn't expect the gather operations to be that efficient.

PR #3908, opened 2025-08-16
➡️ SFTTrainer for VLM: support completion-only loss

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/trainer/sft_trainer.py
```diff
@@ -292,6 +292,9 @@ class DataCollatorForVisionLanguageModeling(DataCollatorMixin):
             include a `tokenizer` with a defined `pad_token_id`.
         max_length (`int` or `None`, optional, defaults to `None`):
             Maximum sequence length for input tokens. If `None`, no truncation is applied.
+        completion_only_loss (`bool`, *optional*, defaults to `False`):
```
merveenoyan, line 310: I think we can go with by default True, the model shouldn't learn anything from the input itself, wdyt?

merveenoyan (approved):
let's go! 💗

trl/trainer/sft_trainer.py
qgallouedec: It's a bit confusing, but in fact, this default value is never used, because SFT always instantiates the collator with a value. And this value actually depends on the type of dataset and the args passed by the user.
To put it simply, if you pass a prompt-completion dataset, unless you explicitly pass `args.args.completion_only_loss=False`, the loss will be computed on completion only. See  https://github.com/huggingface/trl/blob/e22ac2475d4e7042c848e779403e3ca1ebfb419a/trl/trainer/sft_trainer.py#L756-L760

qgallouedec: added a comment to reduce confusion

PR #3916, opened 2025-08-18
🤹‍♂️ Multi-image testing dataset

qgallouedec:
See https://huggingface.co/datasets/trl-internal-testing/zen-multi-image


scripts/generate_zen_image_dataset.py
```diff
@@ -52,364 +52,6 @@ class ScriptArguments:
 
 def main(test_size, push_to_hub, repo_id):
     # fmt: off
-    sizes = np.random.randint(32, 64, size=(19, 2))
```
qgallouedec, line 55: doesn't make sense to include standard format since they don't contain any image token

qgallouedec:
merging without review, it's just about adding a dataset

Issue #3927, opened 2025-08-20
Using assistant_only_loss=True with sequence length > max_length fails silently


jonnyli1125:
### Reproduction

Script:
```py
from datasets import Dataset
from trl import SFTTrainer, SFTConfig

ds = Dataset.from_list(
    [
        {"messages": [{"role": "user", "content": "hello"}, {"role": "assistant", "content": "world"}]}
        for _ in range(100)
    ]
)
args = SFTConfig(
    output_dir="./test",
    max_length=64,
    assistant_only_loss=True,
)
trainer = SFTTrainer(
    model="HuggingFaceTB/SmolLM3-3B",
    args=args,
    train_dataset=ds,
)
trainer.train()
```

Exception:
```
Traceback (most recent call last):
 File "/tmp/ray/session_2025-08-04_10-52-00_527934_1/runtime_resources/working_dir_files/_ray_pkg_c5c76ec2e920915a/train_debug.py", line 38, in <module>
 ray.get(main.remote())
 File "/home/ray/anaconda3/lib/python3.9/site-packages/ray/_private/auto_init_hook.py", line 22, in auto_init_wrapper
 return fn(*args, **kwargs)
 File "/home/ray/anaconda3/lib/python3.9/site-packages/ray/_private/client_mode_hook.py", line 104, in wrapper
 return func(*args, **kwargs)
 File "/home/ray/anaconda3/lib/python3.9/site-packages/ray/_private/worker.py", line 2858, in get
 values, debugger_breakpoint = worker.get_objects(object_refs, timeout=timeout)
 File "/home/ray/anaconda3/lib/python3.9/site-packages/ray/_private/worker.py", line 958, in get_objects
 raise value.as_instanceof_cause()
ray.exceptions.RayTaskError(ValueError): [36mray::main()[39m (pid=461502, ip=10.244.57.31)
 File "/tmp/ray/session_2025-08-04_10-52-00_527934_1/runtime_resources/working_dir_files/_ray_pkg_c5c76ec2e920915a/train_debug.py", line 33, in main
 trainer.train()
 File "/tmp/ray/session_2025-08-04_10-52-00_527934_1/runtime_resources/pip/b3037a2716f5d158eb3870df23418846e39eeb31/virtualenv/lib/python3.9/site-packages/transformers/trainer.py", line 2238, in train
 return inner_training_loop(
[...]

PR #3930, opened 2025-08-20
Update assistant mask exception.

pramodith:
# What does this PR do?

Addresses #3927 where it's possible for all the `assistant_mask` tokens are 0 when the inputs are truncated if `max_length` is set. 

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -987,7 +987,9 @@ class SFTTrainer(Trainer):
                         "processing_class": processing_class,
                         "dataset_text_field": args.dataset_text_field,
                         "assistant_only_loss": args.assistant_only_loss,
-                        "max_length": args.max_length if args.max_length is not None else -1,
+                        "max_length": args.max_length
+                        if args.max_length is not None
+                        else processing_class.model_max_length,
                     },
                     **map_kwargs,
                 )
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -921,7 +921,7 @@ class SFTTrainer(Trainer):
                 if isinstance(dataset, Dataset):  # `IterableDataset.map` does not support `desc`
                     map_kwargs["desc"] = f"Tokenizing {dataset_name} dataset"
 
-                def tokenize(example, processing_class, dataset_text_field, assistant_only_loss):
+                def tokenize(example, processing_class, dataset_text_field, assistant_only_loss, max_length):
                     if "prompt" in example:  # prompt-completion case
                         output = {}
                         if is_conversational(example):
@@ -968,7 +968,7 @@ class SFTTrainer(Trainer):
                                 tools=example.get("tools"),
                                 **example.get("chat_template_kwargs", {}),
                             )
-                            if "assistant_masks" in processed and 1 not in processed["assistant_masks"]:
+                            if "assistant_masks" in processed and 1 not in processed["assistant_masks"][:max_length]:
                                 raise RuntimeError(
                                     "You're using `assistant_only_loss=True`, but at least one example has no "
                                     "assistant tokens. This usually means the tokenizer's chat template doesn't "
@@ -987,6 +987,7 @@ class SFTTrainer(Trainer):
                         "processing_class": processing_class,
                         "dataset_text_field": args.dataset_text_field,
                         "assistant_only_loss": args.assistant_only_loss,
+                        "max_length": args.max_length if args.max_length is not None else -1,
                     },
                     **map_kwargs,
                 )


qgallouedec:
In my understanding, if there is one sample in the dataset for which the prompt is too long, resulting in the prompt+completion part being truncated, this would make the training fail?

EDIT for clarification:
Example if there is one sample with prompt_len=70, completion_len=32 and max_length=64, this would make the training fail instead of ignoring this example?

pramodith:
> In my understanding, if there is one sample in the dataset for which the prompt is too long, resulting in the prompt+completion part being truncated, this would make the training fail?

True! Should we check for some threshold ratio of rows and then just log a warning if that threshold is reached instead of raising an exception to handle the case of truncation?

pramodith:
> EDIT for clarification:
Example if there is one sample with prompt_len=70, completion_len=32 and max_length=64, this would make the training fail instead of ignoring this example?

I actually like the idea of ignoring a row! I think we should drop/ignore the rows in the dataset that have no assistant tokens post truncation and log to the user the % of rows ignored. I think I can accomplish this with some dataset ops.

qgallouedec:
yep agree. Maybe log the number of active tokens (not masked)

qgallouedec:
Maybe we can even check directly the labels, it would allow to account for all unmasked tokens (just check label!=-100)

pramodith:
> Maybe we can even check directly the labels, it would allow to account for all unmasked tokens (just check label!=-100)

Hmmm my understanding is that the `labels` column is populated in the data collator and not at dataset preparation stage. This'd mean that we need to keep track of the running sum of the number of tokens since the data collator is called per iteration of the Dataloader. Is this something we want to do?

I also think that just accounting for the assistant tokens count would be the same as the total number of trainable tokens, everything else would either be a pad token or non-assistant token both of which would need to be masked in assistant only training mode.

qgallouedec:
> Hmmm my understanding is that the labels column is populated in the data collator

Ah yes you're right!

> I also think that just accounting for the assistant tokens count would be the same as the total number of trainable tokens

not exactly, because when you train on completion-only, you've a `completion_mask` that can also be full of 0.0 after truncation

pramodith:
> not exactly, because when you train on completion-only, you've a `completion_mask` that can also be full of 0.0 after truncation

Ahh yes forgot about that, made some changes to reflect all the three types of `datasets conversational, completions/instruction-tuning and language modeling.`

trl/trainer/sft_trainer.py
```diff
@@ -1010,9 +1010,50 @@ def tokenize(example, processing_class, dataset_text_field, assistant_only_loss)
...
+                        return sum(len(row[col]) for row in dataset)
+                    return sum(sum(row[col]) for row in dataset)
+
+                first_row = dataset[0] if isinstance(dataset, Dataset) else next(iter(dataset))
+                # Conversational Dataset
+                if args.assistant_only_loss:
+                    total_trainable_tokens_before_truncation = get_trainable_tokens(dataset, "assistant_masks")
+                # Prompt Completions/Instruction Tuning Dataset
+                elif "completion_mask" in first_row:
+                    total_trainable_tokens_before_truncation = get_trainable_tokens(dataset, "completion_mask")
```
qgallouedec: so it can be both conversational and prompt-completion. In such case, the loss is only computed for the tokens 1. in the completion and 2 from role "user".

qgallouedec:
@albertvillanova can you review this one?
Basically the idea is that we have a dataset containing the column `input_ids` (list of ints), maybe a column `assistant_mask` (list of bools), maybe a column `completion_mask` (list of bools).
A token contributes to the loss only if:
- `assistant_masks` is not a column or `assistant_masks=1` for this token AND
- `completion_mask` is not a column or `completion_mask=1` for this token

We want to log the number of tokens contributing to the loss and filter the examples where none of the tokens contribute to the loss. 
I'm not sure if it's the best way to leverage `datasets` in this case

pramodith:
> I'd recommend re-implementing with `dataset.map(..., batched=True, num_proc=...)` + `dataset.filter(...)`, which keeps the whole pipeline efficient, parallel, and robust.

@albertvillanova let me know if I'm wrong but what we're trying is more of a `reduce` operation than a `map` operation since we're trying to compute the total number of non masked tokens.

I can run a `map` with `batched=True` to get the row wise sum but will still need to iterate over the entire dataset to reduce that to a final dataset level sum. Am I missing something here?

I came across an example of converting a dataset to a polars dataframe and running reduce operation on the polars dataframe [here](https://github.com/huggingface/datasets/pull/5533#issuecomment-2498180088). Would that help in this scenario? My main concern with this approach is if we'll be effectively duplicating the dataset in memory with this approach.

albertvillanova (commented):
Yes, I think that is a good approach:
- Do as many intermediate steps with `map(batched=True` and/or `filter`.
  - These are efficient and fast: :hugs: Datasets caches all the transformations in Arrow
  - At this stage, data is not in-memory, but memory-mapped on disk
- For the final aggregation, convert to Polars and run the reduce.
  - This operation works with zero-copy on the underlying Arrow buffers

pramodith:
@albertvillanova re-opened this one per your request 😄 .

Adding my comment in the new PR that I had opened (and subsequently closed after re-opening this) here so that it doesn't get lost.

> I don't think I can avoid the for-loops even with batched I have to iterate through all the rows in the batch to get the sum of tokens in each row, so I'm not sure how much this'll help.

qgallouedec:
can we try to do it without polars? I'd like to avoid adding this dependency just for this feature

pramodith:
Feels like there isn't a nice  and efficient way of counting the number of trainable tokens ahead of time without multiple passes through the entire dataset and I don't feel like this is a valuable enough addition to the library to pursue it any further. Unless anyone has a simple solution in mind, I'm considering closing out this PR.

PR #3931, opened 2025-08-20
🐳 Docker update + Simplify Jobs doc

qgallouedec:
Our new simple docker image mostly intended to be used with [trl jobs](https://github.com/huggingface/trl-jobs)

Note that the image is already pushed to https://hub.docker.com/r/huggingface/trl-source-gpu (soon rename to just "huggingface/trl")


.github/workflows/docker-build.yml
```diff
@@ -1,95 +1,87 @@
-name: Build Docker images (scheduled)
+name: Build TRL Dev Docker image
```
qgallouedec:
```suggestion
name: Build TRL Docker image
```

.github/workflows/docker-build.yml
```diff
@@ -1,95 +1,84 @@
-name: Build Docker images (scheduled)
+name: Build TRL Docker image
 
 on:
+  push:
+    branches:
+      - main
```
qgallouedec, line 6: in theory, we need would need to build

- dev:
    - every commit on main
    - every new release of a dependency
- stable:
    - every patch release
    - every new release of a dependency

however, since we typically make a few commits per day on main, let's simplify things and just build both images when a new commit is made on main.

docker/trl/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM pytorch/pytorch:2.8.0-cuda12.8-cudnn9-runtime
+RUN pip install --upgrade pip uv
```
sergiopaniego, line 2:
```suggestion
FROM pytorch/pytorch:2.8.0-cuda12.8-cudnn9-runtime
RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/*
RUN pip install --upgrade pip uv
```

The image works for TRL jobs but when trying to use it directly via HF jobs cli, it fails since it doesn't have git installed. For example the following command:

```
hf jobs uv run --flavor a100-large --secrets HF_TOKEN --image huggingface/trl  trl/scripts/sft.py --model_name_or_path Qwen/Qwen2-0.5B --dataset_name trl-lib/Capybara   
```

Fails with the following error:
```
   Updating https://github.com/huggingface/trl.git (HEAD)
  × Failed to download and build `trl @
  │ git+https://github.com/huggingface/trl.git`
  ├─▶ Git operation failed
  ╰─▶ Git executable not found. Ensure that Git is installed and available.
```

This comes sin in that  [trl/scripts/sft.py](https://github.com/huggingface/trl/blob/e8b8499f1f8d76838155b515e414ee98f757d6d5/trl/scripts/sft.py#L17) script, the uv dependency will try to install the latest trl version from git:

```
#     "trl @ git+https://github.com/huggingface/trl.git",
```

This error is not present when using the dev image since git is installed in the Dockerfile.

qgallouedec: so maybe we should not try to install it from source and rely on the installed version? Lemme try

qgallouedec: removing

```
# /// script
# dependencies = [
#     "trl @ git+https://github.com/huggingface/trl.git",
#     "peft",
#     "trackio",
#     "kernels",
# ]
# ///
```

actually works just fine!

qgallouedec: I just need to add kernels in the ~doc~ EDIT: docker image

qgallouedec: Actually I'm thinking that maybe these scripts are going to be used with other images, so it's probably better to keep the uv dependencies. Just not forcing installing trl from source is better

qgallouedec:
I took the opportunity to simplify the jobs documentation:

- if it was not specific to training/trl, I redirected to HF Jobs doc (env var, hardware, timeout etc)
- add some warnings about important things like pushing the model and set a proper timeout
- remove the section on trackio as we now have a dedicated section in the doc for it
- made the training script part a bit more general (script don't necessarily come from the repo
- moved up trl-jobs, as it's the simpler solution so I expect user to go further in the doc for further customization (you don't want the user to discover at the end that a simpler solution existed)
- move the docker part down as it's probably more for power users to choose their own docker image

I'll merge this PR, as it's been opened for a while now, and I want the docker image to be operating. we can still refine in the future if needed

trl/scripts/env.py
```diff
@@ -27,15 +27,15 @@
 from transformers import is_bitsandbytes_available
 from transformers.utils import is_openai_available, is_peft_available
 
-from .. import __version__
-from ..import_utils import (
+from trl import __version__
```
qgallouedec, line 30: for scripts, we don't want relative imports

PR #3936, opened 2025-08-21
🗞 bugfix 'TrainerState' object is not subscriptable

ErezYosef:
# Fix: Prevent TrainerState object from being subscripted in GRPO Trainer

### Description
This PR fixes a bug in `GRPOTrainer` where `TrainerState` objects inside `reward_kwargs` were incorrectly treated as subscriptable, leading to the following error:

```
File ".../trl/trainer/grpo_trainer.py", line 1332, in _calculate_rewards
    row_reward_kwargs = {key: value[nan_row_idx] for key, value in reward_kwargs.items()}
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'TrainerState' object is not subscriptable
```
The issue was introduced in [#3669: *feat: Pass trainer state to reward functions*](https://github.com/huggingface/trl/pull/3669), which added `trainer_state` to `reward_kwargs`:

https://github.com/huggingface/trl/blob/e04f7eb3b993b5278733a793853ed3a85d1c69c8/trl/trainer/grpo_trainer.py#L1035

### Fix
This PR ensures that the `trainer_state` entry is excluded from subscripting:
```python
row_reward_kwargs = {key: value[nan_row_idx] for key, value in reward_kwargs.items() if key!='trainer_state'}
```
This prevents the TypeError while preserving the intended reward calculation logic.

Erez Yosef.

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1385,6 +1385,35 @@ class GRPOTrainerTester(TrlTestCase):
             new_param = trainer.model.get_parameter(n)
             self.assertTrue(torch.equal(param, new_param), f"Parameter {n} has changed.")
 
+    def test_warning_raised_all_rewards_none(self):
+        """Test that a proper warning is raised when all rewards are None."""
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        def always_none_reward_func(completions, **kwargs):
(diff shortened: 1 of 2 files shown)


qgallouedec (approved):
Thanks! added a test as well

PR #3938, opened 2025-08-22
🏌️ DAPO loss type

qgallouedec:
This PR introduces the true DAPO aggregation method for the loss. Previously, we were only able to achieve partial equivalence, because the normalizer was local and not accumulated. In other words, you could only achieve DAPO with gas=1 and num GPUs=1.

I checked that this new loss type is working and correct with both gradient accumulation and distributed training!

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -226,13 +226,18 @@ def split_tensor_dict(
     """
     first_tensor = next(tensor for tensor in tensor_dict.values() if tensor is not None)
     chunk_size = first_tensor.shape[0] // num_chunks
-    return [
-        {
-            key: tensor[i * chunk_size : (i + 1) * chunk_size] if tensor is not None else None
-            for key, tensor in tensor_dict.items()
-        }
-        for i in range(num_chunks)
-    ]
+    chunks = []
+    for i in range(num_chunks):
+        chunk_dict = {}
+        for key, tensor in tensor_dict.items():
+            if tensor is not None and tensor.ndim > 0:
+                chunk_dict[key] = tensor[i * chunk_size : (i + 1) * chunk_size]
+            elif tensor is not None and tensor.ndim == 0:
+                chunk_dict[key] = tensor
+            else:
+                chunk_dict[key] = None
+        chunks.append(chunk_dict)
+    return chunks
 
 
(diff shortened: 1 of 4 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -258,7 +263,9 @@ def shuffle_sequence_dict(seq_dict: dict[str, Optional[Sequence]]) -> dict[str,
     def permute(v: Optional[Sequence]) -> Optional[Sequence]:
         if v is None:
             return None
-        if isinstance(v, torch.Tensor):
+        if isinstance(v, torch.Tensor) and v.ndim == 0:
```
qgallouedec, line 266: this is for `"num_items_in_batch"` that is a single float that needs to be copied in every chunks

trl/trainer/grpo_trainer.py
```diff
@@ -1894,10 +1909,16 @@ def _compute_loss(self, model, inputs):
 
         if self.loss_type == "grpo":
             loss = ((per_token_loss * completion_mask).sum(-1) / completion_mask.sum(-1).clamp(min=1.0)).mean()
+            loss = loss / self.current_gradient_accumulation_steps
```
qgallouedec, line 1912: now that we don't rely on trainer to scale the reward, we need to do it here.

shirinyamani (approved):
I watched you on X doing it, perfectly LGTM!

trl/trainer/grpo_trainer.py
```diff
@@ -700,6 +707,12 @@ def __init__(
             processing_class=processing_class,
             callbacks=callbacks,
             optimizers=optimizers,
+            # In Trainer, `training_step` scales the loss by `gradient_accumulation_steps` only if `compute_loss_func`
+            # is None. For DAPO, loss scaling instead depends on the total number of completions tokens across the
+            # global accumulated batch. To control scaling ourselves, we must disable Trainer’s built-in scaling. The
+            # simplest (though a bit hacky) way is to set `compute_loss_func` to any non-None value, which bypasses
+            # that behavior without rewriting `training_step`.
+            compute_loss_func="non-None value to disable scaling",
```
qgallouedec, line 715: cc @SunMarc, idk if you'd have done it this way

SunMarc: yeah it's a bit hacky but i guess that one of the simplest solution. Maybe you can also do it with modifying `model_accepts_loss_kwargs` and `num_items_in_batches`

PR #3949, opened 2025-08-25
add default dataset and related precessing to make this example work

kaixuanliu:
Hi, I cannot get this example work by just specify a dataset using the default reward_funcs `think_format_reward` or `get_soft_overlong_punishment` in this example. Hence I added a default dataset and related preprocessing.

### DIFF
--- a/trl/scripts/grpo.py
+++ b/trl/scripts/grpo.py
@@ -123,23 +123,24 @@ def main(script_args, training_args, model_args, dataset_args):
     else:
         raise ValueError("Either `datasets` or `dataset_name` must be provided.")
 
-    SYSTEM_PROMPT = (
-        "A conversation between user and assistant. The user asks a question, and the assistant solves it. The "
-        "assistant first thinks about the reasoning process in the mind and then provides the user with the answer. "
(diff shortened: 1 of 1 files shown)


yao-matrix:
@kashif , pls help review, thx very much.

qgallouedec (changes requested):
This is a script, not an example, it's mainly here for being used with the TRL CLI. Consequently, I don't think we want to have any dataset-specific dataset preprocessing in it.
Also, `GRPOScriptArguments` inherits from `ScriptArguments` so it should already have a `dataset_name`  parameter

kaixuanliu:
Hi @qgallouedec , can you give an example cmd line that can run using TRL CLI? I tried a lot but they all failed.

qgallouedec:
After #3972:

```
trl grpo --model_name_or_path Qwen/Qwen3-0.6B --reward_funcs get_soft_overlong_punishment --dataset_name trl-lib/ultrafeedback-prompt --max_completion_length 2048
```

kaixuanliu:
Close this PR.

PR #3968, opened 2025-08-28
[GRPO] Adds an option to sleep vllm when running in colocated mode

edbeeching:
# What does this PR do?
This PR adds an option which using vllm colocate model to offload vllm's weights and cache to CPU memory during the optimization step, freeing up more ram for model activations, etc.

The current implementation is as follows:
image

With `vllm_sleep_enabled=True`:
image

This option was mentioned in https://huggingface.co/blog/vllm-colocate#sleep-mode-in-vllm but has not been upstreamed to TRL yet.

Comparison of training curves with / without this option:

image

Comparable loss / reward curves, but slower training due to sleep / wake.

In a future PR, we can potentially add CPU offloading to the model / optimizer. So vllm has access to more memory for the generation step.

Memory benchmarks for max seq length in the two configurations:

(TODO)


trl/trainer/grpo_trainer.py
```diff
@@ -835,7 +835,10 @@ def __init__(
                     # Latest vLLM v1 memory profiler is misled by the high default value (i.e., 32768) - thinking there's not enough memory
                     max_num_batched_tokens=4096,
                     model_impl=self.args.vllm_model_impl,
+                    enable_sleep_mode=self.args.vllm_sleep_enabled,
                 )
+                if self.args.vllm_sleep_enabled:
+                    self.llm.sleep(level=1)
```
toslali-ibm, line 542: Every time vllm wakes up, it wakes up to an updated model, right? So why not use level = 2 to further improve efficiency?

docs/source/grpo_trainer.md
```diff
@@ -251,6 +251,10 @@ We provide a [HF Space](https://huggingface.co/spaces/trl-lib/recommend-vllm-mem
 
 If the recommended value does not work in your environment, we suggest adding a small buffer (e.g., +0.05 or +0.1) to the recommended value to ensure stability.
 
+<Tip>
+If you still find you are getting out-of-memory errors set `vllm_sleep_enabled` to True and the vllm parameters and cache will be offloaded during the optimization step.
```
lewtun:
```suggestion
If you still find you are getting out-of-memory errors, set `vllm_sleep_enabled=True` and the vLLM model parameters and cache will be offloaded to CPU during the optimization step.
```

trl/trainer/grpo_config.py
```diff
@@ -140,6 +140,8 @@ class GRPOConfig(TrainingArguments):
             Model implementation to use for vLLM. Must be one of `"transformers"` or `"vllm"`. `"transformers"`: Use
             the `transformers` backend for model implementation. `"vllm"`: Use the `vllm` library for model
             implementation.
+        vllm_sleep_enabled (`bool`, *optional*, defaults to `False`):
+            Whether to enable sleep mode for vLLM. If `True`, vLLM will sleep during the optimization step and woken for weight sync and generation.
```
lewtun:
```suggestion
            Whether to enable sleep mode for vLLM. If `True`, vLLM will sleep during the optimization step and be awoken for weight sync and generation.
```

trl/trainer/grpo_config.py
```diff
@@ -416,6 +418,12 @@ class GRPOConfig(TrainingArguments):
             "model implementation."
         },
     )
+    vllm_sleep_enabled: bool = field(
+        default=False,
+        metadata={
+            "help": "Whether to enable sleep mode for vLLM. If `True`, vLLM will sleep during the optimization step and woken for weight sync and generation."
```
lewtun:
```suggestion
            "help": "Whether to enable sleep mode for vLLM. If `True`, vLLM will sleep during the optimization step and be awoken for weight sync and generation."
```

trl/trainer/grpo_trainer.py
edbeeching: Ah good point, let me rerun the benchmark with level=2.

trl/trainer/grpo_trainer.py
```diff
@@ -835,7 +835,10 @@ def __init__(
                     # Latest vLLM v1 memory profiler is misled by the high default value (i.e., 32768) - thinking there's not enough memory
                     max_num_batched_tokens=4096,
                     model_impl=self.args.vllm_model_impl,
+                    enable_sleep_mode=self.args.vllm_sleep_enabled,
```
lewtun: A bit of a turbo nit, but for other vllm params we take their name exactly as-is with a `vllm_` prefix. Following that, we should technically do:

```suggestion
                    enable_sleep_mode=self.args.vllm_enable_sleep_mode
```

If you agree, then we need to propagate the change elsewhere

lewtun (commented):
LGTM once the comment from @toslali-ibm is tested

docs/source/grpo_trainer.md
```diff
@@ -251,6 +251,10 @@ We provide a [HF Space](https://huggingface.co/spaces/trl-lib/recommend-vllm-mem
 
 If the recommended value does not work in your environment, we suggest adding a small buffer (e.g., +0.05 or +0.1) to the recommended value to ensure stability.
 
+<Tip>
```
sergiopaniego:
```suggestion
<Tip>

```

docs/source/grpo_trainer.md
```diff
@@ -251,6 +251,10 @@ We provide a [HF Space](https://huggingface.co/spaces/trl-lib/recommend-vllm-mem
...
+If you still find you are getting out-of-memory errors set `vllm_sleep_enabled` to True and the vllm parameters and cache will be offloaded during the optimization step.
+</Tip>
```
sergiopaniego:
```suggestion

</Tip>
```

To properly render content enclosed in \` \`, an extra blank line is required at both the beginning and the end of Tips

Currently: 

Screenshot 2025-08-28 at 16 20 59

trl/trainer/grpo_trainer.py
toslali-ibm: For level = 2, you will also need to wake-up then sleep in `_move_model_to_vllm`, `_sync_fsdp2_params_to_vllm` and `_sync_fsdp1_params_to_vllm`. 

BAsically anytime you touch vllm (in generation or loading/syncying the model), you wake up, do the work, and then go back to sleep.

Additional note: the reason we did not move sleep to upstream was because of a [vllm bug](https://github.com/vllm-project/vllm/issues/16993), which is just recently fixed by this [PR](https://github.com/vllm-project/vllm/pull/23477). So that means, you need to use vllm version that incorporates the fix to be able to use sleep level 2 without segmentation fault.

qgallouedec (approved):
LGTM, I let you change the name as suggested by @lewtun if you want to

trl/trainer/grpo_trainer.py
edbeeching: Yes, I have already included a wake up call before the weight sync. I don't re-sleep as we have the gen step straight after.

It looks like the level 2 fix has not made it into their most recent release. I will leave level=1 for now for better backward compatability. Unless there are other reasons to go with level 2?

PR #3906, opened 2025-08-15
[DRAFT] Refactor DPO

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #3986, opened 2025-09-01
Enable saving and loading precomputed reference log probabilities in …

### LINKED ISSUE #3985: Feature Request: Save/Load Precomputed Ref Log-Probabilities in DPOTrainer
### Feature request

Save/Load Precomputed Ref Log-Probabilities in DPOTrainer

### Motivation

Currently, when `precompute_ref_log_probs=True`, `DPOTrainer` always recomputes the reference model log-probs (ref_chosen_logps, ref_rejected_logps) for the training and evaluation datasets.
This process can be very time-consuming, and for repeated experiments on the same dataset/model setup, it leads to redundant computation.

It would be highly beneficial to have an option to cache precomputed reference log-probs to disk and reload them later, avoiding unnecessary recomputation.

### Your contribution

Introduce two new arguments:
* `save_ref_logps_dir`: (optional) directory path where precomputed log-probs will be stored
* `load_ref_logps_dir`: (optional) directory path to load precomputed log-probs from
  * When provided, the trainer checks dataset fingerprint, number of rows, and model/tokenizer info to ensure compatibility
  * If they match, cached values are loaded instead of recomputing
  * If they differ, print log and the cache is ignored and recomputation proceeds as usual

ginkyenglee:
- Added two new DPOConfig arguments: `load_ref_logps_dir` and `save_ref_logps_dir`
- Allows DPOTrainer to reuse precomputed `ref_chosen_logps` and `ref_rejected_logps` from disk
- Saves new log probabilities during training/evaluation when configured
- Added safeguards for distributed training (save only on main process, synchronize with wait_for_everyone)
- Improves reproducibility and avoids expensive recomputation across runs

# What does this PR do?

Fixes #3985

## Before submitting
- [] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
      to it if that's the case.
- [] Did you make sure to update the documentation with your changes?

## Who can review?

### DIFF
--- a/trl/trainer/dpo_config.py
+++ b/trl/trainer/dpo_config.py
@@ -319,6 +319,18 @@ class DPOConfig(TrainingArguments):
             "`per_device_train_batch_size` for training and `per_device_eval_batch_size` for evaluation."
         },
     )
+
+    save_ref_logps_dir: Optional[str] = field(
+        default=None,
+        metadata={
+            "help": "Directory to save the precomputed reference model log probabilities."
+            }) 
+            
+    load_ref_logps_dir: Optional[str] = field(
+        default=None,
+        metadata={"help": "Directory to load the precomputed reference model log probabilities from."
+        })
+
     tools: Optional[list[Union[dict, Callable]]] = field(
         default=None,
         metadata={
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -20,7 +20,7 @@ from collections import defaultdict
 from contextlib import contextmanager, nullcontext
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable, Literal, Optional, Union
+from typing import Any, Callable, Literal, Optional, Union, Dict
 
 import pandas as pd
 import torch
@@ -28,7 +28,7 @@ import torch.nn as nn
 import torch.nn.functional as F
 from accelerate import PartialState, logging
 from accelerate.utils import tqdm
-from datasets import Dataset, IterableDataset
+from datasets import Dataset, IterableDataset, load_from_disk
 from torch import autocast
 from torch.utils.data import DataLoader
 from transformers import (
@@ -388,6 +388,8 @@ class DPOTrainer(Trainer):
         self.max_length = args.max_length
         self.truncation_mode = args.truncation_mode
         self.precompute_ref_log_probs = args.precompute_ref_log_probs
+        self.save_ref_logps_dir = args.save_ref_logps_dir
+        self.load_ref_logps_dir = args.load_ref_logps_dir
         self.use_logits_to_keep = args.use_logits_to_keep
 
         if args.padding_free:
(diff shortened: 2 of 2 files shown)


qgallouedec:
Thanks for the PR — I completely understand the need.

`datasets` already provides a caching mechanism intended to handle this exact use case. However, it doesn’t seem to be working as expected here, and I don’t see an obvious fix at the moment. I’m a bit hesitant to support the feature in its current form because, while useful, it depends on a workaround rather than properly leveraging the caching system in `datasets`.

That said, it could still bring value to the community. While we wait for a proper fix, I see two options:

* Merge it, but include a warning that this argument will be removed at any time without notice.
* Leave the PR open in the meantime, and reference it in the documentation.

what do you think @albertvillanova @lewtun @kashif @edbeeching?

edbeeching:
> Thanks for the PR — I completely understand the need.
> 
> `datasets` already provides a caching mechanism intended to handle this exact use case. However, it doesn’t seem to be working as expected here, and I don’t see an obvious fix at the moment. I’m a bit hesitant to support the feature in its current form because, while useful, it depends on a workaround rather than properly leveraging the caching system in `datasets`.
> 
> That said, it could still bring value to the community. While we wait for a proper fix, I see two options:
> 
> * Merge it, but include a warning that this argument will be removed at any time without notice.
> * Leave the PR open in the meantime, and reference it in the documentation.
> 
> what do you think @albertvillanova @lewtun @kashif @edbeeching?

I think it would be ok as long as there is a warning. Perhaps you could just have a `precomputed_logprob_dir` rather than both the save/load args. Then if that file exists, load the `ref_logprobs`. If it does not exist, compute the `ref_logprobs` and save them.

ginkyenglee:
> > Thanks for the PR — I completely understand the need.
> > `datasets` already provides a caching mechanism intended to handle this exact use case. However, it doesn’t seem to be working as expected here, and I don’t see an obvious fix at the moment. I’m a bit hesitant to support the feature in its current form because, while useful, it depends on a workaround rather than properly leveraging the caching system in `datasets`.
> > That said, it could still bring value to the community. While we wait for a proper fix, I see two options:
> > 
> > * Merge it, but include a warning that this argument will be removed at any time without notice.
> > * Leave the PR open in the meantime, and reference it in the documentation.
> > 
> > what do you think @albertvillanova @lewtun @kashif @edbeeching?
> 
> I think it would be ok as long as there is a warning. Perhaps you could just have a `precomputed_logprob_dir` rather than both the save/load args. Then if that file exists, load the `ref_logprobs`. If it does not exist, compute the `ref_logprobs` and save them.

Thanks for the comments!

**Proposal: make `precomputed_logprob_dir` the single switch.**
* If `None` → no precompute.
* If set → load cache if present; otherwise compute+save in the same run.
I’ll deprecate (and ignore when dir is set) `precompute_ref_log_probs`.

**Rationale**: simpler, single source of truth and aligns with the earlier “one dir + warning” suggestion.

Does this look good?

albertvillanova (commented):
> `datasets` already provides a caching mechanism intended to handle this exact use case. However, it doesn’t seem to be working as expected here, and I don’t see an obvious fix at the moment.

Hi, if you think there is an underlying issue with `datasets` for this use case, I could have a look. I can try to identify whether the fix should happen upstream in the `datasets` lib, or on our side (if it is a misuse).

qgallouedec:
closing in favour of #3906

PR #803, opened 2023-09-21
Unify sentiment documentation

vwxyzjn:

PR #3993, opened 2025-09-02
Add pre-commit as dev dependency

albertvillanova:
Add `pre-commit` as `dev` dependency within a new `quality` extra:
- Add pre-commit as a new `quality` extra dependency in `setup.cfg`
- Include pre-commit in the `dev` extra for comprehensive development setup
- Enables consistent code quality checks across the project

Note that `pre-commit` was previously removed as dev dependency in PR:
- #803

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,6 +51,8 @@ liger =
     liger-kernel>=0.6.2
 peft =
     peft>=0.8.0
+quality =
+    pre-commit
 quantization =
     bitsandbytes
 scikit =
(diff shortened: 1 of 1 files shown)


qgallouedec (approved):
Thanks!

albertvillanova:
@qgallouedec, I just discovered that `make precommit` uses `doc-builder` (`hf-doc-builder` in PyPI) in this command: 
```
doc-builder style trl tests docs/source --max_len 119
```
- do we really need this command? Then, I would need to add the `hf-doc-builder` dependency as well.
- or can I remove it because of ruff setting `line-length = 119`?

qgallouedec:
Doc builder style will also format the doc. Ruff doesn't. So I think you should add it to the deps

PR #3995, opened 2025-09-02
Fix: Make sft script work when chat template is None

rabinadk1:
# What does this PR do?

This pull request makes improvements to the `trl/scripts/sft.py` script, primarily focusing on import optimization, correcting lazy importing, and minor API adjustments for clarity and correctness. The most significant changes are grouped below:

**Import optimization and lazy loading:**

* Removed the eager import of `AutoModelForCausalLM` at the top of the file and moved it inside the relevant code path in the `main` function to load it lazily.

**API and function signature improvements:**

* Added `_added_tokens` as the third argument because `clone_chat_template` returns three values, not two.
* Refactored the `make_parser` function to make the type hint for the `subparsers` accurate and simplified the logic for returning the correct parser instance.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/scripts/sft.py
+++ b/trl/scripts/sft.py
@@ -66,7 +66,7 @@ import os
 
 from accelerate import logging
 from datasets import load_dataset
-from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
+from transformers import AutoConfig, AutoTokenizer
 from transformers.models.auto.modeling_auto import MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES
 
 from trl import (
@@ -113,6 +113,8 @@ def main(script_args, training_args, model_args, dataset_args):
 
         model = AutoModelForImageTextToText.from_pretrained(model_args.model_name_or_path, **model_kwargs)
     else:
+        from transformers import AutoModelForCausalLM
+        
         model = AutoModelForCausalLM.from_pretrained(model_args.model_name_or_path, **model_kwargs)
 
     # Create tokenizer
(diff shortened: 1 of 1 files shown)


trl/scripts/sft.py
```diff
@@ -123,7 +126,7 @@ def main(script_args, training_args, model_args, dataset_args):
     # Set default chat template if needed
     if tokenizer.chat_template is None:
         # TODO: source should be passed as an argument
-        model, tokenizer = clone_chat_template(model, tokenizer, "Qwen/Qwen3-0.6B")
+        model, tokenizer, _added_tokens = clone_chat_template(model, tokenizer, "Qwen/Qwen3-0.6B")
```
kashif, line 129:
```suggestion
        model, tokenizer, _ = clone_chat_template(model, tokenizer, "Qwen/Qwen3-0.6B")
```

rabinadk1: I wanted to do this as well, but then I made it more explicit to the users.

trl/scripts/sft.py
```diff
@@ -63,10 +63,11 @@
 
 import argparse
 import os
+from typing import Optional
 
 from accelerate import logging
 from datasets import load_dataset
-from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
```
qgallouedec, line 69: I think you can revert this change

trl/scripts/sft.py
```diff
@@ -159,13 +162,11 @@ def main(script_args, training_args, model_args, dataset_args):
...
 
 
-def make_parser(subparsers: argparse._SubParsersAction = None):
+def make_parser(subparsers: Optional[argparse._SubParsersAction] = None):
     dataclass_types = (ScriptArguments, SFTConfig, ModelConfig, DatasetMixtureConfig)
-    if subparsers is not None:
-        parser = subparsers.add_parser("sft", help="Run the SFT training script", dataclass_types=dataclass_types)
-    else:
-        parser = TrlParser(dataclass_types)
-    return parser
```
qgallouedec, line 168: IMO, using a consistent indent level for the return is a better practice

trl/scripts/sft.py
kashif: as you like... ruff etc. might complain... i think this and the `Optional` should be the only change in this PR

qgallouedec: agree with @kashif

PR #3960, opened 2025-08-27
Dft

1485840691:
# What does this PR do?
Support dynamic fine tuning 

Fixes 3877 (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #4003, opened 2025-09-03
Add missing doc strings in SFTrainer

pramodith:
# What does this PR do?

The `SFTTrainer` currently misses doc strings for the `compute_loss_func` and `compute_metrics` constructor args. Both of these are passed on to the parent `Trainer` Class's constructor. This PR adds them, copied over from the Trainer class from the Transformers lib.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


qgallouedec:
I think `compute_loss_func` is actually never used in `SFTTrainer`. I recommend removing it (deprecate and remove in next version)

trl/trainer/sft_trainer.py
```diff
@@ -521,6 +521,15 @@ class SFTTrainer(Trainer):
...
             If the processing class has not set a padding token, `tokenizer.eos_token` will be used as the default.
+        compute_loss_func (`Callable`, *optional*):
+            A function that accepts the raw model outputs, labels, and the number of items in the entire accumulated
+            batch (batch_size * gradient_accumulation_steps) and returns the loss. For example, see the default [loss function](https://github.com/huggingface/transformers/blob/052e652d6d53c2b26ffde87e039b723949a53493/src/transformers/trainer.py#L3618) used by [`Trainer`].
+        compute_metrics (`Callable[[EvalPrediction], Dict]`, *optional*):
+            The function that will be used to compute metrics at evaluation. Must take a [`EvalPrediction`] and return
+            a dictionary string to metric values. *Note* When passing TrainingArgs with `batch_eval_metrics` set to
+            `True`, your compute_metrics function must take a boolean `compute_result` argument. This will be triggered
+            after the last eval batch to signal that the function needs to calculate and return the global summary
+            statistics rather than accumulating the batch-level statistics
```
```suggestion
        compute_metrics (`Callable[[EvalPrediction], dict]` or `None`, *optional*, defaults to `None`):
            The function that will be used to compute metrics at evaluation. Must take a
            [`~transformers.EvalPrediction`] and return a dictionary string to metric values. When passing
            [`SFTConfig`] with `batch_eval_metrics` set to `True`, your `compute_metrics` function must take a boolean
            `compute_result` argument. This will be triggered after the last eval batch to signal that the function
            needs to calculate and return the global summary statistics rather than accumulating the batch-level
            statistics.
```

qgallouedec: just improving slightly the doc from `Trainer`

pramodith:
> I think `compute_loss_func` is actually never used in `SFTTrainer`. I recommend removing it (deprecate and remove in next version)

I think it is used because it is passed to the `transformer.Trainer`'s constructor which allows the user to pass in a custom loss function.

https://github.com/huggingface/trl/blob/67991605c0e6aaf1ef3c2bf64e11da914948c4a4/trl/trainer/sft_trainer.py#L822-L829

The Trainer calls the passed `compute_loss_func` in its own compute_loss function.
https://github.com/huggingface/transformers/blob/25b4a0d8aef515ed309c935607473da319d9291c/src/transformers/trainer.py#L4123-L4125

I actually think having this is a nice way to allow users to experiment with their own loss functions, so would refrain from getting rid of it. This PR in progress also serves as an example of how one can use the `compute_loss_func` for POCs.
https://github.com/huggingface/trl/pull/3960

qgallouedec:
`compute_loss` is overwritten in SFT, so this function is never user

pramodith:
> `compute_loss` is overwritten in SFT, so this function is never user

Yes but even the overwritten version calls the trainer's implementation of `compute_loss` which in turn uses `compute_loss_func`

https://github.com/huggingface/trl/blob/0c69fd2867bf93d56f2dd5c33d772e2720f7c4ab/trl/trainer/sft_trainer.py#L1041-L1043

Relevant code in `Trainer`'s `compute_loss`

```
if self.compute_loss_func is not None:
                loss = self.compute_loss_func(outputs, labels, num_items_in_batch=num_items_in_batch)
            elif model_name in MODEL_FOR_CAUSAL_LM_MAPPING_NAMES.values():
                loss = self.label_smoother(outputs, labels, shift_labels=True)
```

qgallouedec:
My bad you're right!

PR #4005, opened 2025-09-04
Improve SFT doc

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -476,7 +476,7 @@ class SFTTrainer(Trainer):
     """
     Trainer for Supervised Fine-Tuning (SFT) method.
 
-    This class is a wrapper around the [`transformers.Trainer`] class and inherits all of its attributes and methods.
+    This class is a wrapper around the [`~transformers.Trainer`] class and inherits all of its attributes and methods.
 
     Example:
 
@@ -502,9 +502,10 @@ class SFTTrainer(Trainer):
             - A [`~transformers.PreTrainedModel`] object. Only causal language models are supported.
         args ([`SFTConfig`], *optional*, defaults to `None`):
             Configuration for this trainer. If `None`, a default configuration is used.
-        data_collator (`DataCollator`, *optional*):
+        data_collator ([`~transformers.DataCollator`] or `None`, *optional*, defaults to `None`):
             Function to use to form a batch from a list of elements of the processed `train_dataset` or `eval_dataset`.
-            Will default to a custom [`DataCollatorForLanguageModeling`].
+            Will default to a custom [`DataCollatorForLanguageModeling`] if the model is a language model and
+            [`DataCollatorForVisionLanguageModeling`] if the model is a vision-language model.
         train_dataset ([`~datasets.Dataset`] or [`~datasets.IterableDataset`]):
             Dataset to use for training. SFT supports both [language modeling](#language-modeling) type and
             [prompt-completion](#prompt-completion) type. The format of the samples can be either:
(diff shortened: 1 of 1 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -504,9 +504,10 @@ class SFTTrainer(Trainer):
             as a part of the final loss, remember to set the `output_router_logits` config of the model to `True`.
         args ([`SFTConfig`], *optional*, defaults to `None`):
             Configuration for this trainer. If `None`, a default configuration is used.
-        data_collator (`DataCollator`, *optional*):
+        data_collator ([`~transformers.DataCollator`] or `None`, *optional*):
             Function to use to form a batch from a list of elements of the processed `train_dataset` or `eval_dataset`.
-            Will default to a custom [`DataCollatorForLanguageModeling`].
+            Will default to [`trainer.sft_trainer.DataCollatorForLanguageModeling`] if the model is a language model
+            and [`trainer.sft_trainer.DataCollatorForVisionLanguageModeling`] if the model is a vision-language model.
```
albertvillanova: Nit:
```suggestion
            Will default to [`~trainer.sft_trainer.DataCollatorForLanguageModeling`] if the model is a language model
            and [`~trainer.sft_trainer.DataCollatorForVisionLanguageModeling`] if the model is a vision-language model.
```

trl/trainer/sft_trainer.py
```diff
@@ -535,33 +535,30 @@ class SFTTrainer(Trainer):
...
-        preprocess_logits_for_metrics (`Callable[[torch.Tensor, torch.Tensor], torch.Tensor]`, *optional*, defaults to
-        `None`):
+        preprocess_logits_for_metrics (`Callable[[torch.Tensor, torch.Tensor], torch.Tensor]`, *optional*, defaults to `None`):
             A function that preprocess the logits right before caching them at each evaluation step. Must take two
             tensors, the logits and the labels, and return the logits once processed as desired. The modifications made
             by this function will be reflected in the predictions received by `compute_metrics`.
 
             Note that the labels (second parameter) will be `None` if the dataset does not have them.
-        peft_config ([`~peft.PeftConfig`], *optional*, defaults to `None`):
+        peft_config ([`~peft.PeftConfig`] or `None`, *optional*, defaults to `None`):
```
albertvillanova, line 581: I think the ``or `None` `` part may be redundant here, since the type hint and default value already make it clear that None is accepted.

I would say that no adding it would be simpler and more consistent with the rest of the codebase. What do you think?

qgallouedec: > since the type hint

I find the type hint `Optional[X]` misleading—I hope it gets deprecated eventually. It’s not intuitive that it actually means "`X` or `None`". That’s why I prefer writing "`X` or `None`" explicitly in docstrings instead of using `Optional[X]`. I see many user confuser by this.

> and default value

True in this case, and in most cases, `None` is indeed the default, but not always. For example, with `SFTConfig.max_length`, `None` is accepted but isn’t the default. So without the "or `None`", it may not be clear that `None` is valid.

Because of these, I wouldn’t rely solely on type hints.
Both approaches work for me (remove or keep "or `None`"): if the default makes it obvious, we can omit "or `None`", but for consistency, we should remove it everywhere in the codebase.

qgallouedec: leaving like this for now, the whole codebase needs harmonization, see https://github.com/huggingface/trl/issues/4036

PR #4019, opened 2025-09-05
Fix passing model kwargs

qgallouedec:
passing

```python
model = AutoModelForCausalLM.from_pretrained("my_model", quantization_config=None)
```

isn't the same as

```python
model = AutoModelForCausalLM.from_pretrained("my_model")
```

causing gpt oss model to fail loading when used with trl cli

### DIFF
--- a/trl/trainer/gkd_trainer.py
+++ b/trl/trainer/gkd_trainer.py
@@ -155,7 +155,6 @@ class GKDTrainer(SFTTrainer):
             temperature=args.temperature,
             do_sample=True,
             top_k=0,
-            use_cache=False if args.gradient_checkpointing else True,
             pad_token_id=self.processing_class.pad_token_id,
         )
         # Set custom EOS tokens if they are specified by the model's generation
--- a/trl/trainer/online_dpo_trainer.py
+++ b/trl/trainer/online_dpo_trainer.py
@@ -303,7 +303,6 @@ class OnlineDPOTrainer(Trainer):
                 top_k=50,
                 top_p=1.0,
                 do_sample=True,
-                use_cache=False if args.gradient_checkpointing else True,
             )
 
         # The trainer estimates the number of FLOPs (floating-point operations) using the number of elements in the
--- a/examples/scripts/dpo_online.py
+++ b/examples/scripts/dpo_online.py
@@ -83,16 +83,19 @@ if __name__ == "__main__":
     script_args, training_args, model_args = parser.parse_args_and_config()
     training_args.gradient_checkpointing_kwargs = {"use_reentrant": True}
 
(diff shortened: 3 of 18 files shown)


trl/trainer/gkd_trainer.py
```diff
@@ -194,7 +194,6 @@ def __init__(
             temperature=args.temperature,
             do_sample=True,
             top_k=0,
-            use_cache=False if args.gradient_checkpointing else True,
```
albertvillanova: I guess this change in unrelated?

trl/trainer/online_dpo_trainer.py
```diff
@@ -563,7 +563,6 @@ def __init__(
                 "top_k": self.top_k,
                 "top_p": self.top_p,
                 "repetition_penalty": self.repetition_penalty,
-                "use_cache": True if not self.args.gradient_checkpointing else False,
```
albertvillanova: The same as above?

albertvillanova (commented):
Thanks for the PR: definitely, GPT OSS models should be supported in TRL!!!

That said, I'm not fully convinced we need to apply the "filter out all None kwargs" policy across the board. Most parameters in `from_pretrained` are robust to `None`: have you seen any issue with any parameter other than `quantization_config`?

In my view, there are two complementary points here:
1. **Upstream issue in Transformers**: do you know why transformers treat differently missing and `None` `quantization_config`? I think this might be a bug. In my opinion, passing `None` should be treated the same as omitting the key altogether, not as an invalid quantization config. It would be worth raising or linking an issue upstream so we can align behavior there.
2. **Targeted fix in TRL**: While investigating and waiting for an upstream fix, it makes sense for TRL to guard specifically against this problem.

What about stripping only the `quantization_config` (and associated `device_map`) if it is `None`. This would keep the fix minimal, and avoid unnecessarily rewriting the kwargs for unrelated arguments. What do you think?

trl/trainer/gkd_trainer.py
qgallouedec: Yes, it shouldn't be in this PR

qgallouedec:
yeah good point. 

> have you seen any issue with any parameter other than quantization_config?

no


> do you know why transformers treat differently missing and None quantization_config?

I think it comes from these lines, probably a bug, but I haven't had time to look into it further.

https://github.com/huggingface/transformers/blob/37c14430c99edca79dfcdcb76f1209f291b12fab/src/transformers/configuration_utils.py#L962-L967

> Targeted fix in TRL: While investigating and waiting for an upstream fix, it makes sense for TRL to guard specifically against this problem.

yep, agree, done

---

I will merge this PR now, as I'd like to include it in the release, but we should definitely do this 

> It would be worth raising or linking an issue upstream so we can align behavior there.

PR #4024, opened 2025-09-07
Made ref_model as None in PPO trainer for refined args

complete-dope:
# What does this PR do?

Fixes 
1. Explicit Passing of ref_model=None in PPOTrainer , by adding a default None 

## Tests
- No tests required , as its a minor fix 

Thanks


qgallouedec:
I believe that most code bases use `ref_model` as positional arguments, and this change would be breaking without providing any significant benefit. For now, I oppose this change.

PR #4032, opened 2025-09-08
[GRPO VLM] Update split sizes to generalize

zucchini-nlp:
# What does this PR do?

As discussed internally, we can rely on transformer utilities to split pixels to a list of batches. That way we dont have to handle each edge case manually

The split pixels will always be of length `batch_size`. In most cases the split sizes are a list of ones, if pixels have no patches/crops/uncommon shapes. In other cases, each batch will contain variable sized image crops per single image

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1079,12 +1079,18 @@ class GRPOTrainer(Trainer):
         # [{"role": "user", "content": [{"type": "image"}, {"type": "text", "text": "What color is the sky?"}]}]
         kwargs = {}
         has_images = "image" in inputs[0]
+        image_split_sizes = None
         if has_images:
             images = [example.get("image") for example in inputs]
             kwargs = {"images": [[img] for img in images]}
             for prompt in prompts:
                 if isinstance(prompt, list):  # i.e., when using conversational data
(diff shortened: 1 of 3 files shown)


qgallouedec (approved):
LGTM, can you just take a final look before I merge @zucchini-nlp?

trl/trainer/grpo_trainer.py
```diff
@@ -1097,6 +1106,10 @@ def _generate_and_score_completions(
         prompt_inputs = super()._prepare_inputs(prompt_inputs)
         prompt_ids, prompt_mask = prompt_inputs["input_ids"], prompt_inputs["attention_mask"]
 
+        if "image_grid_thw" in prompt_inputs and image_split_sizes is None:
+            # Fallback for VLMs that require image_grid_thw but don't provide _get_num_multimodal_tokens
+            image_split_sizes = prompt_inputs["image_grid_thw"].prod(dim=1).tolist()
```
zucchini-nlp, line 1109: not sure if we have such models (except for qwen-omni). I will update all VLMs for next release to have helpers, even if not used in vLLM

zucchini-nlp:
Thanks for updating it, didn't have time to come back to tests yesterday

PR #4034, opened 2025-09-08
Add autodoc for BestOfNSampler and improve docstrings

albertvillanova:
Add autodoc for `BestOfNSampler` classes and improve docstrings.

### DIFF
--- a/trl/extras/best_of_n_sampler.py
+++ b/trl/extras/best_of_n_sampler.py
@@ -84,22 +84,22 @@ class BestOfNSampler:
         **generation_kwargs,
     ) -> list[list[str]]:
         r"""
-        Generate the best of n samples for input queries
+        Generate the best of n samples for input queries.
 
         Args:
             tokenized_query (`list[int]` or `torch.Tensor` or `list[torch.Tensor]` or `list[int]`):
                 represents either a single tokenized query (a single tensor or a list of integers) or a batch of
-                tokenized queries (a list of tensors or a list of lists of integers)
+                tokenized queries (a list of tensors or a list of lists of integers).
             skip_special_tokens (`bool`):
-                Whether to remove the special tokens from the output
+                Whether to remove the special tokens from the output.
             device (`str` or `torch.device`, *optional*):
-                The device on which the model will be loaded
+                The device on which the model will be loaded.
             **generation_kwargs (`dict`, *optional*):
                 Additional keyword arguments passed along to the underlying model's `generate` method. This is used to
-                override generation config
+                override generation config.
 
         Returns:
-            list[list[str]]: A list of lists of generated texts
+            list[list[str]]: A list of lists of generated texts.
         """
         queries = None
 
--- a/trl/extras/best_of_n_sampler.py
+++ b/trl/extras/best_of_n_sampler.py
@@ -87,7 +87,7 @@ class BestOfNSampler:
         Generate the best of n samples for input queries.
 
         Args:
-            tokenized_query (`list[int]` or `torch.Tensor` or `list[torch.Tensor]` or `list[int]`):
+            tokenized_query (`list[int]` or `torch.Tensor` or `list[torch.Tensor]` or `list[list[int]]`):
                 Either a single tokenized query (a single tensor or a list of integers) or a batch of tokenized queries
                 (a list of tensors or a list of lists of integers).
             skip_special_tokens (`bool`, *optional*, defaults to `True`):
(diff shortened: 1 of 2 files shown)


trl/extras/best_of_n_sampler.py
```diff
@@ -21,6 +21,29 @@
 
 
 class BestOfNSampler:
+    """Initialize the sampler for best-of-n generation
```
qgallouedec:
```suggestion
    """
    Initialize the sampler for best-of-n generation
```

trl/extras/best_of_n_sampler.py
```diff
@@ -21,6 +21,29 @@
...
+    Args:
+        model (`PreTrainedModelWrapper`):
+            The pretrained model to use for generation
+        tokenizer (`PreTrainedTokenizer` or `PreTrainedTokenizerFast`):
+            Tokenizer associated with the pretrained model
+        queries_to_scores (`Callable[[list[str]], list[float]]`):
+            Callable that takes a list of generated texts and returns the associated reward scores
+        length_sampler (`Any`):
+            Sampler used to sample the length of the generated text
+        sample_size (`int`):
```
qgallouedec:
```suggestion
        sample_size (`int`, *optional*, default to `4`):
```

trl/extras/best_of_n_sampler.py
```diff
@@ -21,6 +21,29 @@
...
+            Number of samples to generate for each query
+        seed (`int`, *optional*):
```
qgallouedec, line 38:
```suggestion
        seed (`int` or `None`, *optional*, default to `None`):
```

trl/extras/best_of_n_sampler.py
```diff
@@ -21,6 +21,29 @@
...
+            Random seed used to control generation
+        n_candidates (`int`):
```
qgallouedec:
```suggestion
        n_candidates (`int`, *optional*, defaults to `1`):
```

trl/extras/best_of_n_sampler.py
```diff
@@ -21,6 +21,29 @@
...
+            Number of candidates to return for each query
+        generation_config (`GenerationConfig`, *optional*):
```
qgallouedec:
```suggestion
        generation_config (`GenerationConfig` or `None`, *optional*, defaults to `None`):
```

qgallouedec (approved):
A few remarks, otherwise Lgtm!

trl/extras/best_of_n_sampler.py
albertvillanova: @qgallouedec, I wanted to ask you, as I commented in your PR: https://github.com/huggingface/trl/pull/4005#discussion_r2329387182
- I think the `or None` addition is redundant
- Additionally, I think even the `defaults to None` part is also redundant

To support my opinion:
- In the doc-builder docs says: [Parmeter typing and default value](https://github.com/huggingface/doc-builder?tab=readme-ov-file#parmeter-typing-and-default-value)
> Imagine we have a function with the following signature:
> `def my_function(x: str = None, a: float = 1):`
> then its documentation should look like this:
> ```x (`str`, *optional*):```
> Note that we always omit the "defaults to `None`" when None is the default for any argument
- Also note that this is the case for the rest of huggingface projects: transformers, accelerate, datasets, smolagents,...

What do you think? IMHO, that makes the docstring simpler and easier to read, removing redundant information. Note that our docs also show the type hints of the signature, where more precise type info can be found.

albertvillanova:
Thanks for the review, @qgallouedec.

Note that I just added the autodoc and moved the existing docstring from `__init__` to the class declaration part, so it appears in the online docs. I did not worked further in correction of types, inter-links, description, etc, because I prioritized the addition of the docstring. But I can work on it as you suggested and change the title of the PR: :sweat_smile: 
> Add autodoc for BestOfNSampler and improve docstring

trl/extras/best_of_n_sampler.py
albertvillanova: I can make a first pass with the evident corrections and a second pass with this one...

qgallouedec:
Yes you're right, let's stay organized ;)

trl/extras/best_of_n_sampler.py
qgallouedec: I'll answer here  #4036

albertvillanova:
I guess I can merge this PR as it is.

qgallouedec:
Sure

PR #4037, opened 2025-09-08
💬 Remove setting chat template in sft script

qgallouedec:
SFT script is not necessarily used for instruction tuning. Plus, the right way to set a chat template is now to use the arg `SFTConfig.chat_template_path`

Closes #3995


qgallouedec:
Merging without review as I need it for release, feel free to give late feedback!

PR #4045, opened 2025-09-09
 [Draft] Add configurable dataset column logging to GRPOTrainer W&B tables

davanstrien:
This PR adds a wandb_log_extra_columns parameter to GRPOConfig that allows logging additional dataset columns to W&B tables during GRPO training. I found this helpful for including ground truth labels or other metadata alongside generated completions for easier evaluation.

Implementation
-  Added wandb_log_extra_columns to GRPOConfig:
- `None` (default): No extra columns logged
- []: Log all available dataset columns
- ["col1", "col2"]: Log specific columns only
- Modified GRPOTrainer to collect and log specified columns
- Added support for both standard generation and vLLM backends

Example Usage

```python
  config = GRPOConfig(
      wandb_log_extra_columns=["ground_truth", "difficulty"],
      # ... other config
  )
```

 This is quite a rough POC developed for a specific use case. Opening as a draft to discuss whether this functionality would be useful for TRL more broadly. 

Example of logging an extra column for `grount_truth`

Screenshot 2025-09-09 at 12 09 51

I have done some minimal testing, but mostly made this for myself to start, so would probably benefit from some better tests if it seemed interesting to add to TRL.


davanstrien:
@qgallouedec I've been using this for my own projects successfully now. Let me know if this seems useful for TRL more generally, and I can make a cleaner PR. No worries if not, can also keep this on my own fork otherwise :)

qgallouedec:
thanks @davanstrien!

Ideally we would have something that is compatible with other loggers as well, like trackio. I'll look into it when I have the bandwidth

davanstrien:
No rush from my side! Also happy to take a look at a more generic approach, but wanted to check the idea in general was of interest for TRL first :)

qgallouedec:
Closing this one, we ended up building the generic version you and I talked about.

`log_completions` now writes the completions table to wandb and trackio both, plus a parquet dump and an optional Hub dataset via `log_completions_hub_repo`. Extra columns ride along through a `log_extra(column, values)` callable that reward functions receive, so anything in your dataset can land in the table: dataset columns are passed to reward functions as kwargs, and `log_extra` puts them in the same row as the prompt, completion, rewards and advantage.

The documented example is your use case almost word for word:

```python
def reward_func(completions, ground_truth, log_extra=None, **kwargs):
    ...
    if log_extra:
        log_extra("golden_answer", list(ground_truth))
```

`trl.rewards.accuracy_reward` does the same for `solution`, `gold_parsed` and `answer_parsed`.

The one difference from your proposal is that the columns are named by the reward function rather than declared in the config, so logging a column no reward function touches needs a small function of its own. If that turns out to be a real annoyance, say so and we can look at a config-level shortcut on top of the same mechanism.

Thanks for raising it, the multi-logger constraint I mentioned back then is what the current design came from.

PR #4047, opened 2025-09-09
Update transformers minimum version to 4.56.1

albertvillanova:
Update transformers minimum version to 4.56.1.

That patch version includes the fix:
- https://github.com/huggingface/transformers/pull/40668

This should fix the CI tests with minimum versions: https://github.com/huggingface/trl/actions/runs/17571124955/job/49907398546
```python
AttributeError: 'Accelerator' object has no attribute 'parallelism_config'
```


albertvillanova:
Most CI tests with minimum versions are green now; only some red left: https://github.com/huggingface/trl/actions/runs/17583974907/job/49946949816

PR #4060, opened 2025-09-10
[GRPO]: Sample from a Replay Buffer To Substitute Groups with 0 std.

pramodith:
# What does this PR do?
Introduces the idea of a `ReplayBuffer` to GRPO. Implementation details

- A `ReplayBuffer` is implemented as a heap/priority queue. The buffer stores a `score` and a dict with the same keys as `_generate_and_score_completions`.
- The `ReplayBuffer` stores entire groups and all the keys associated with a group that'd be needed for computing the loss. Storing the old/ref_log_probs ensures that we don't run any extra forward passes through models.
- Currently the scoring is based on the summed product of absolute advantages and std of a group. 
- Everytime `_generate_and_score_completions` is called we check if 1. There are any groups with non-zero variance, these are candidates to be added to the ReplayBuffer. 2. There are any groups with 0 variance, these need to be substituted out with values from the replay buffer.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


pramodith:
I'm still mulling over the multi-gpu scenario, I'm wondering if we should have the same buffer used on all gpus/processes or if its okay for each gpu/process to have its own buffer. Happy to hear your views on this @qgallouedec 

Also still need to add an e2e test for training with the ReplayBuffer.

pramodith:
I should probably break the `update_with_replay_buffer` function into two smaller functions, its too big rn:
1. add_to_buffer
2. replace_from_buffer.

Also need to add new test cases to confirm that the code works when the seq lengths in the buffer are different from the current batch.

qgallouedec:
can you migrate this into `trl.experimental`? 🙏

pramodith:
> can you migrate this into `trl.experimental`? 🙏

Yes will do.

trl/experimental/grpo_with_replay_buffer/grpo_with_replay_buffer_config.py
```diff
@@ -0,0 +1,34 @@
...
+from dataclasses import dataclass, field
+
+from trl.trainer.grpo_config import GRPOConfig
+
+
+@dataclass
+class GRPOWithReplayBufferConfig(GRPOConfig):
+    """
+    New Parameters:
+        replay_buffer_size (`int`, *optional*, defaults to `0`):
```
qgallouedec, line 24:
```suggestion
        replay_buffer_size (`int`, *optional*, defaults to `64`):
```

qgallouedec (approved):
lgtm!

PR #4065, opened 2025-09-11
🧹 Remove `max_batch_tokens`, `num_blocks` and `block_size` from generation kwargs

qgallouedec:
These parameters are now automatically computed, see https://github.com/huggingface/transformers/pull/40426


remi-or:
As a reminder, while `max_batch_tokens` and `num_blocks` are infered, `block_size` isn't and defaults to 32. You might want to test out how the change from 128 to 32 will impact you :slightly_smiling_face: 
Not really familiar with `trl` so I can't really review this, sorry!

qgallouedec:
thanks for the feedback @remi-or! 

>  change from 128 to 32 will impact you 🙂

will do!

qgallouedec:
Tried with 0.5B and 1.7B, no difference between block_size=32 and 128👍

PR #4076, opened 2025-09-13
Fix CI

qgallouedec:
This PR https://github.com/huggingface/transformers/pull/40416 introduced a new key in the log history, which makes the CI fail. We just need to check that all the keys/values are *included* in the log history, no need to check this new key

### DIFF
--- a/tests/test_callbacks.py
+++ b/tests/test_callbacks.py
@@ -118,7 +118,8 @@ class WinRateCallbackTester(TrlTestCase):
         trainer.add_callback(win_rate_callback)
         trainer.train()
         winrate_history = [h for h in trainer.state.log_history if "eval_win_rate" in h]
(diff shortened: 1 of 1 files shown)


scripts/generate_tiny_models.py
```diff
@@ -256,23 +254,13 @@ def init_weights_tiny_model(model):
...
-        d_model=16,
-        encoder_layers=2,
-        decoder_layers=2,
-        d_kv=2,
-        d_ff=64,
-        num_layers=6,
-        num_heads=8,
-        decoder_start_token_id=0,
-        is_encoder_decoder=True,
-    )
```
qgallouedec, line 275: the issue came from this config that was not correct. It's way easier to just modify the config from an existing model

kashif (approved):

Issue #3957, opened 2025-08-26
sft_gemma3 example doesn't work


yao-matrix:
## Command
`accelerate launch --config_file examples/accelerate_configs/deepspeed_zero3.yaml examples/scripts/sft_gemma3.py`

## Error Message

> [rank0]: Traceback (most recent call last):
> [rank0]:   File "/workspace/trl/examples/scripts/sft_gemma3.py", line 67, in <module>
> [rank0]:     main()
> [rank0]:   File "/workspace/trl/examples/scripts/sft_gemma3.py", line 60, in main
> [rank0]:     trainer.train()
> [rank0]:   File "/workspace/transformers/src/transformers/trainer.py", line 2318, in train
> [rank0]:     return inner_training_loop(
> [rank0]:            ^^^^^^^^^^^^^^^^^^^^
> [rank0]:   File "/workspace/transformers/src/transformers/trainer.py", line 2613, in _inner_training_loop
> [rank0]:     batch_samples, num_items_in_batch = self.get_batch_samples(epoch_iterator, num_batches, args.device)
> [rank0]:                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> [rank0]:   File "/workspace/transformers/src/transformers/trainer.py", line 5436, in get_batch_samples
> [rank0]:     batch_samples.append(next(epoch_iterator))
> [rank0]:                          ^^^^^^^^^^^^^^^^^^^^
> [rank0]:   File "/workspace/accelerate/src/accelerate/data_loader.py", line 567, in __iter__
> [rank0]:     current_batch = next(dataloader_iter)
> [rank0]:                     ^^^^^^^^^^^^^^^^^^^^^
> [rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/utils/data/dataloader.py", line 734, in __next__
> [rank0]:     data = self._next_data()
> [rank0]:            ^^^^^^^^^^^^^^^^^
> [rank0]:   File "/usr/local/lib/python3.12/dist-packages/torch/utils/data/dataloader.py", line 790, in _next_data
> [rank0]:     data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
> [rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4080, opened 2025-09-14
[WIP] Fix usage of VLM using text only

SamuelBarryCS:
## What
- Fixes https://github.com/huggingface/trl/issues/3957

## How to review
- WIP

## Test performed
- WIP

## Benchmark
- WIP

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -377,11 +377,17 @@ class DataCollatorForVisionLanguageModeling(DataCollatorMixin):
             raise KeyError(f"Unexpected input keys in examples: {list(examples[0].keys())}.")
 
     def _collate_language_modeling(self, examples: list[Union[list[int], Any, dict[str, Any]]]) -> dict[str, Any]:
-        images = [example["images"] for example in examples]
+        # Check if examples contain images - some VLMs can be used for text-only tasks
+        has_images = "images" in examples[0]
+        if has_images:
+            images = [example["images"] for example in examples]
+        else:
+            images = None
 
         if "messages" in examples[0]:  # conversational case
             for example in examples:
-                prepare_multimodal_messages(example["messages"], len(example["images"]))
+                num_images = len(example["images"]) if has_images else 0
+                prepare_multimodal_messages(example["messages"], num_images)
             messages = [example["messages"] for example in examples]
             texts = self.processor.apply_chat_template(messages)
         elif self.dataset_text_field in examples[0]:  # standard case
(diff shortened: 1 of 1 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -392,17 +398,21 @@ def _collate_language_modeling(self, examples: list[Union[list[int], Any, dict[s
...
-            padding=True,
-            padding_side="right",
-            pad_to_multiple_of=self.pad_to_multiple_of,
-            truncation=self.max_length is not None,
-            max_length=self.max_length,
-            return_tensors=self.return_tensors,
-            add_special_tokens=False,  # to avoid adding the BOS, twice see https://huggingface.co/blog/qgallouedec/gotchas-in-tokenizer-behavior#7-chat-template-and-tokenization-dont-compose-due-to-special-tokens
-        )
+        # For text-only data with VLM models, don't pass images to the processor
+        processor_kwargs = {
```
SamuelBarryCS: I used a kwargs structure to avoid redundancy.
Otherwise we have to do:

```
if images: ...
output = self.processor(images, texts, ...)
else:
output = self.processor(texts, ...)
```

which I personally find quite ugly, but I can change if needed

test_vlm_text_only_issue.py
```diff
@@ -0,0 +1,31 @@
+"""Test for issue #3957 - VLM KeyError fix"""
```
SamuelBarryCS: Will be deleted before merging

SamuelBarryCS:
cc @qgallouedec for review when you get the time.
Merci!

sergiopaniego (approved):
Thanks for the fix!
We could add a modified version of the script as a test in `test_sft_trainer.py` and remove the script

SamuelBarryCS:
> Thanks for the fix! We could add a modified version of the script as a test in `test_sft_trainer.py` and remove the script

Thanks for your quick reply @sergiopaniego !
Done in https://github.com/huggingface/trl/pull/4080/commits/646559485752a951d07ad05dc779d2ce95da0b3c :), and the test is passing fine:
image

Good to merge on my side (but it looks like we still need approval for the workflows to run)

SamuelBarryCS:
@sergiopaniego do you understand why the tests are failing ? It doesn't look related to my changes ... :x
Otherwise ready to merge on my side (once this test issue is solved).

qgallouedec:
Hi, thanks for your contribution. I opted for a different approach: when the dataset does not contain images, it can be pre-processed, and there is no point in doing it on the fly.

SamuelBarryCS:
> Hi, thanks for your contribution. I opted for a different approach: when the dataset does not contain images, it can be pre-processed, and there is no point in doing it on the fly.

This works as well! Thanks for the edit @qgallouedec.

PR #4082, opened 2025-09-15
Add support for testing experimental features

albertvillanova:
Add support for testing experimental features.

This PR introduces a minimal testing workflow for the new `experimental` submodule:
- Default behavior unchanged: pytest and make test now exclude experimental tests automatically via `addopts = ["-k", "not experimental"]` in pyproject.toml
- New target: `make test_experimental`: runs only the experimental tests
- Directory structure: All tests for features under the `trl.experimental` namespace should be placed in the directory:
  - `tests/experimental/`

This setup keeps CI fast and stable (running only the main test suite) while making it easy for contributors to run experimental tests locally (or in a future separate CI job).

Follow-up to:
- #4073

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: test precommit common_tests slow_tests test_examples tests_gpu
+.PHONY: test precommit common_tests slow_tests test_examples tests_gpu test_experimental
 
 check_dirs := examples tests trl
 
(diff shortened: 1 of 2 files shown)


qgallouedec:
I initially thought that a marker would be more suitable, but the subdirectory solution in test is also good IMO, as it allows the codebases to remain truly separate.

qgallouedec (approved):
lgtm!

PR #4083, opened 2025-09-15
Add deprecation warnings to docstrings

albertvillanova:
Add deprecation warnings to docstrings.

This PR adds deprecation notices to docstrings for documentation consistency: the docstring should match the actual behavior of the code.

This approach is better than removing it entirely because:
- The parameter still exists in the code and will accept values (even if it shows a warning)
- Users need the documentation of the parameter to better understand the deprecation message

### DIFF
--- a/trl/trainer/online_dpo_trainer.py
+++ b/trl/trainer/online_dpo_trainer.py
@@ -170,6 +170,16 @@ class OnlineDPOTrainer(Trainer):
 
         > Deprecated parameters
 
+        reward_model:
+
+            <Deprecated version="0.22.0">
+
+            This parameter is deprecated and will be removed in version 0.25.0. Use `reward_funcs` instead.
+
+            </Deprecated>
+
+        > Deprecated parameters
+
         <Deprecated version="0.22.0">
 
         The following parameters are deprecated and will be removed in version 0.25.0:
--- a/trl/trainer/online_dpo_trainer.py
+++ b/trl/trainer/online_dpo_trainer.py
@@ -175,15 +175,6 @@ class OnlineDPOTrainer(Trainer):
             This parameter is deprecated and will be removed in version 0.25.0. Use `reward_funcs` instead.
 
             </Deprecated>
-
-        reward_processing_class:
-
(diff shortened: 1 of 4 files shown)


albertvillanova:
Screenshot from 2025-09-15 11-35-48

trl/trainer/online_dpo_trainer.py
```diff
@@ -168,12 +168,22 @@ class OnlineDPOTrainer(Trainer):
...
+            </Deprecated>
+
+        reward_processing_class:
+
+            <Deprecated version="0.22.0">
+
+            This parameter is deprecated and will be removed in version 0.25.0. Use `reward_processing_classes`
+            instead.
+
+            </Deprecated>
```
qgallouedec:
```suggestion
```

this one never existed, the previous doc was actually wrong

qgallouedec:
Screenshot 2025-09-15 at 8 50 52 AM

for rloo it's not properly rendered

albertvillanova:
> for rloo it's not properly rendered

Yes, I discovered this bug in doc-builder: the warning is not properly rendered when the parameter description is under a parameter subsection `> Deprecated parameters`.
- For the rest of classes, I just put the deprecated params under no subsection, just at the root parameters section `Args:`
- But for `RLOOConfig`:
  - There already exist other param subsections: 
    ```python
    Parameters:
        > Parameters that control the model and reference model

        model_init_kwargs (`str`, `dict[str, Any]`, *optional*):
    ```
  - Therefore the possible solutions are:
    - Either fix the rendering by putting the deprecated parameters in the root `Args` section, before all the rest of parameters: but I don't think it makes sense to document the deprecated parameters before the functional ones
    - Or put the deprecated parameters in a `> Deprecated parameters` subsection after all the other params, with their malformed rendering
      - I could open an issue (or PR) in doc-builder and eventually have these parameters properly rendered

What do you think?

albertvillanova:
I opened an issue in doc-builder to report this bug:
- huggingface/doc-builder#627

qgallouedec:
I mean, let's keep the malformed warning, not a big deal

PR #4085, opened 2025-09-15
Align slow tests with regular tests

albertvillanova:
Align slow tests with regular tests.

This PR:
- Aligns slow tests with regular tests architecture for consistency and maintainability
- Replaces custom Docker images (`huggingface/trl-latest-gpu`, `huggingface/trl-source-gpu`) with standard `pytorch/pytorch:2.6.0-cuda12.6-cudnn9-devel`
- Removes problematic `--no-deps` flag that prevented VLM dependencies from being installed

## Problem

See previous PR addressing the issue of misaligned dependencies:
- #4064

The slow tests used custom Docker images with pre-installed dependencies and `--no-deps` to avoid conflicts. When VLM extras were added to the install command, the `--no-deps` flag prevented Pillow and other VLM dependencies from being installed, causing import errors.

## Solution

- Use the same proven approach as regular tests: standard PyTorch image + `uv pip install ".[dev]"`
- Eliminates custom Docker image usage burden
- Ensures all extras (including VLM) are properly installed
- Provides consistency across all test workflows

## Additional problem
Also note that `conda` `trl` venv is no longer present in the Docker images after the merge of:
- #3931

and the slow tests (which use the Docker containers) now raise: https://github.com/huggingface/trl/actions/runs/17713423927/job/50335140091
```python
EnvironmentNameNotFound: Could not find conda environment: trl
You can list all discoverable environments with `conda info --envs`.
```

### DIFF
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -20,7 +20,7 @@ jobs:
       CUDA_VISIBLE_DEVICES: "0"
       TEST_TYPE: "single_gpu"
     container:
-      image: pytorch/pytorch:2.6.0-cuda12.6-cudnn9-devel
+      image: pytorch/pytorch:2.8.0-cuda12.8-cudnn9-devel
       options: --gpus all --shm-size "16gb"
     defaults:
       run:
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -14,90 +14,106 @@ env:
 
 jobs:
   run_all_tests_single_gpu:
-    strategy:
-      fail-fast: false
-      matrix:
-        docker-image-name:
-          [
-            "huggingface/trl",
-            "huggingface/trl:dev",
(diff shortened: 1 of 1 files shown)


kashif:
should we test with pytorch 2.7? or even 2.8?

albertvillanova:
Thanks, @kashif, you are right! I just copied the torch version used in regular tests, but it is true that the Dockerfiles (used by the slow-test until now) are using pytorch 2.8.0!
https://github.com/huggingface/trl/blob/5a1c2f9b3b4e314de6df8304f6a32caf0f6737b5/docker/trl/Dockerfile#L1

One question: should I align regular tests and make all use pytorch 2.8? Or is there a specific reason why regular tests are using an older version?

qgallouedec (approved):
Nice!

for te record, I'd like to extend the multi-gpu testing to more variation, see #3902, but for now I'm not sure what's the best way to do it

PR #4092, opened 2025-09-15
📜 Convert `set` to `list` of tags

qgallouedec:
#3550 changed `tags` from `list` to `set`. It added conversion of `tags` to `list` when calling `generate_model_card` in `SFTTrainer` but not in other trainers. This PR _fixes_, though nothing is really broken since `tags` is later implicitly converted to list with 

https://github.com/huggingface/trl/blob/e2b18ec4e7eb63bf9cd7829e0c39b8473b09f9a8/trl/trainer/utils.py#L1252


qgallouedec:
Yes it was to avoid duplicates, I didn't know ModelCardData removes duplicates.
No strong opinion on this but I feel like it's cleaner not to pass duplicates anyway

PR #4102, opened 2025-09-18
docs: correct option name to enable vllm sleep mode

muupan:
# What does this PR do?

The option `vllm_sleep_enabled` mentioned in the docs does not exist. It seems like `vllm_enable_sleep_mode` is the correct one. This PR fixes this. https://github.com/huggingface/trl/blob/20cc58d7772ae660792c7b5249d8b817986a547d/trl/trainer/grpo_config.py#L143C9-L143C31

## Before submitting

## Who can review?

### DIFF
--- a/docs/source/grpo_trainer.md
+++ b/docs/source/grpo_trainer.md
@@ -257,7 +257,7 @@ We provide a [HF Space](https://huggingface.co/spaces/trl-lib/recommend-vllm-mem
 
 If the recommended value does not work in your environment, we suggest adding a small buffer (e.g., +0.05 or +0.1) to the recommended value to ensure stability.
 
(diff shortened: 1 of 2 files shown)


sergiopaniego (approved):
Thanks for the fix!
I can see that for `RLOOConfig`, `vllm_enable_sleep_mode` is not present, so the 

```
training_args = RLOOConfig(..., vllm_enable_sleep_mode=True)
```

wouldn’t work as expected. We could go ahead and merge this PR as is, and then add the functionality in a follow-up. Alternatively, if you’d like to mirror the behavior from GRPOConfig and GRPOTrainer, that could work too.

muupan:
I see, I didn't notice that RLOOConfig doesn't have it.

How about stopping mentioning RLOO for now? I pushed a new commit for it.

sergiopaniego (changes requested):
I've created a PR #4107 with the addition, so we can keep the `RLOO` option here. Would you mind restoring the last changes? 😄

muupan:
Cool, I removed the last commit.

PR #4114, opened 2025-09-20
😷 Refactor GRPO/RLOO to isolate `_generate`

qgallouedec:
This PR is the third of a sequence of PR (after https://github.com/huggingface/trl/pull/4111 and #4113) that aims to refactor the generation part of GRPO/RLOO to allow for easier customization.

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1890,7 +1890,7 @@ class GRPOTrainer(Trainer):
             model_name=model_name,
             hub_model_id=self.hub_model_id,
             dataset_name=dataset_name,
-            tags=tags,
+            tags=list(tags),
             wandb_url=wandb.run.url if is_wandb_available() and wandb.run is not None else None,
             comet_url=get_comet_experiment_url(),
             trainer_name="GRPO",
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -47,7 +47,7 @@ from transformers import (
 from transformers.trainer_utils import seed_worker
 from transformers.utils import is_datasets_available, is_peft_available, is_rich_available
 
-from ..data_utils import apply_chat_template, is_conversational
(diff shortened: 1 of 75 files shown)


qgallouedec:
cc @Peter-Chou, customization is made easier with this one

trl/trainer/grpo_trainer.py
```diff
@@ -1361,7 +1354,72 @@ def _generate_and_score_completions(
...
+            prompt_ids,
+            completion_ids,
+            prompt_mask,
+            completion_mask,
+            num_items_in_batch,
+            sampling_per_token_logps,
+            forward_kwargs,
+        )
+
+    def _generate_and_score_completions(
```
lewtun, line 1392: Would it be cleaner to have separate `_generate()` and `_score()` methods or is it annoying to split them entirely because you need to construct the `prompts` for each?

lewtun (approved):
LGTM once the tests pass and a question about whether we should split `_generate` from scoring entirely

Peter-Chou:
@qgallouedec Yes. The original `_generate_and_score_completions` method was way too lengthy.   
Breaking it down into finer-grained sub-methods and chaining them together like this is an excellent approach!

trl/trainer/grpo_trainer.py
qgallouedec: Yes, it's probably cleaner, but you would have to build the `prompts` in `_score` as well, or pass it as an argument, which would be a bit strange because you would also pass `prompt_ids`. I'm not sure what's the best approach for now. For now, I'll leave it as it is and see if we can do better later.

LeonMalteW:
I'm quite interested in following the refactoring process, as the biggest problem for me is scaling to context sizes of up to 32K, and a lot of OOM bombs get hit right now.
Will the final refactor turn _generate_and_score_completions and all the subfunctions into a pipeline so that we don't need to keep all the data on all GPUs the same, and only the necessary data gets synced? Or is this too specific and would it not hit the target of the refactor?

qgallouedec:
This refactoring is actually for enabling multi-turn RL (with tool calling)

PR #4115, opened 2025-09-21
a bit messy!

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import inspect
+import json
 import os
 import re
 import textwrap
@@ -60,6 +61,8 @@ from .utils import (
     RepeatSampler,
     disable_dropout_in_model,
     entropy_from_logits,
+    flush_left,
+    flush_right,
     generate_model_card,
     get_comet_experiment_url,
     identity,
@@ -96,6 +99,20 @@ logger = logging.get_logger(__name__)
 # What we call a reward function is a callable that takes a list of prompts and completions and returns a list of
 # rewards. When it's a string, it's a model ID, so it's loaded as a pretrained model.
 RewardFunc = Union[str, PreTrainedModel, Callable[[list, list], list[float]]]
+
+
+def extract_tool_calls(text: str) -> dict[str, Any]:
+    """
+    Given a list of strings, extract all <tool_call> JSON blocks and return them as a list of dictionaries.
+    """
+    pattern = re.compile(r"<tool_call>\s*(\{.*?\})\s*</tool_call>", re.DOTALL)
+
+    for match in pattern.findall(text):
+        try:
+            return json.loads(match)
+        except json.JSONDecodeError:
+            pass
+    return None
 
 
 class GRPOTrainer(Trainer):
@@ -214,7 +231,10 @@ class GRPOTrainer(Trainer):
         callbacks: Optional[list[TrainerCallback]] = None,
         optimizers: tuple[Optional[torch.optim.Optimizer], Optional[torch.optim.lr_scheduler.LambdaLR]] = (None, None),
         peft_config: Optional["PeftConfig"] = None,
+        tools=None,
     ):
+        self.tools = tools or []
+        self._tool_dict = {name: tool for name, tool in zip([tool.__name__ for tool in self.tools], self.tools)}
         # Args
         if args is None:
             model_name = model if isinstance(model, str) else model.config._name_or_path
@@ -1084,7 +1104,8 @@ class GRPOTrainer(Trainer):
                     prepare_multimodal_messages(prompt, num_images=len(image_list))
 
         prompts_text = [
-            maybe_apply_chat_template({"prompt": prompt}, self.processing_class)["prompt"] for prompt in prompts
+            maybe_apply_chat_template({"prompt": prompt}, self.processing_class, tools=self.tools)["prompt"]
+            for prompt in prompts
         ]
 
         prompt_inputs = self.processing_class(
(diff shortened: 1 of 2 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -98,6 +101,20 @@
 RewardFunc = Union[str, PreTrainedModel, Callable[[list, list], list[float]]]
 
 
+def extract_tool_calls(text: str) -> dict[str, Any]:
+    """
+    Given a list of strings, extract all <tool_call> JSON blocks and return them as a list of dictionaries.
+    """
+    pattern = re.compile(r"<tool_call>\s*(\{.*?\})\s*</tool_call>", re.DOTALL)
```
lewtun, line 109: There is unfortunately no standardisation for the tool call tags across model families, but Matt is working on extending the chat templates so they can auto-parse tools calls (internal Slack thread): https://huggingface.slack.com/archives/C06JKEMK6BZ/p1757691450090859

Note that `vllm` works around this by providing a dedicated set of parsers that can be set when spinning up the server: https://docs.vllm.ai/en/stable/features/tool_calling.html

I'm not sure we want to go down this route, since it's quite messy in my experience to match the parser to the desired model (e.g. some Qwen models use the `hermes` parser, others not)

So in the meantime, we might want to give uses the ability to provide their own parsing function and default to yours (which is the most common I've seen)

trl/trainer/grpo_trainer.py
```diff
@@ -1413,6 +1434,53 @@ def _generate_and_score_completions(
...
+                new_completion_mask,
+                new_num_items_in_batch,
+                new_sampling_per_token_logps,
+                new_forward_kwargs,
+            ) = self._generate(new_prompts, images)
+            num_tool_ids = new_prompt_mask.sum(-1) - torch.cat(
+                [prompt_mask[needs_tool], completion_mask[needs_tool]], dim=1
+            ).sum(-1)
+            tool_ids = [ids[-num:] for ids, num in zip(new_prompt_ids, num_tool_ids)]
+            tool_mask = [torch.ones_like(ids) for ids in tool_ids]
```
lewtun, line 1464: Would be cool to have a unit test for this masking so we're confident it is behaving as expected

trl/trainer/grpo_trainer.py
lewtun: Perfect timing :) 

https://moon-ci-docs.huggingface.co/docs/transformers/pr_40894/en/chat_response_parsing

albertvillanova: That is the approach we usually followed on smolagents: to provide a sensible default, but allow users to fully customize the function/object instance.

trl/trainer/grpo_trainer.py
```diff
@@ -98,6 +101,20 @@
 RewardFunc = Union[str, PreTrainedModel, Callable[[list, list], list[float]]]
 
 
+def extract_tool_calls(text: str) -> dict[str, Any]:
```
albertvillanova, line 105: What about moving this function to a non-specific trainer module, so it can be used by any trainer in the future?

trl/trainer/grpo_trainer.py
albertvillanova, line 109: Note that the compile function will be called at each function call. If optimizing performance is necessary in this case, we should set the compilation as a constant at the module level, so it is called only once at import time.

trl/trainer/grpo_trainer.py
```diff
@@ -214,7 +231,10 @@ def __init__(
         callbacks: Optional[list[TrainerCallback]] = None,
         optimizers: tuple[Optional[torch.optim.Optimizer], Optional[torch.optim.lr_scheduler.LambdaLR]] = (None, None),
         peft_config: Optional["PeftConfig"] = None,
+        tools=None,
     ):
+        self.tools = tools or []
+        self._tool_dict = {name: tool for name, tool in zip([tool.__name__ for tool in self.tools], self.tools)}
```
albertvillanova: Simpler:
```suggestion
        self._tool_dict = {tool.__name__: tool for tool in self.tools}
```

trl/trainer/grpo_trainer.py
```diff
@@ -214,7 +231,10 @@ def __init__(
         callbacks: Optional[list[TrainerCallback]] = None,
         optimizers: tuple[Optional[torch.optim.Optimizer], Optional[torch.optim.lr_scheduler.LambdaLR]] = (None, None),
         peft_config: Optional["PeftConfig"] = None,
+        tools=None,
```
albertvillanova, line 249: You will need to add the `tools` param to the trainer docstring. And give a type hint.

trl/trainer/grpo_trainer.py
```diff
@@ -98,6 +101,20 @@
...
 
+def extract_tool_calls(text: str) -> dict[str, Any]:
+    """
+    Given a list of strings, extract all <tool_call> JSON blocks and return them as a list of dictionaries.
+    """
+    pattern = re.compile(r"<tool_call>\s*(\{.*?\})\s*</tool_call>", re.DOTALL)
+
+    for match in pattern.findall(text):
+        try:
+            return json.loads(match)
```
albertvillanova, line 113: You only return the first match?

trl/trainer/grpo_trainer.py
```diff
@@ -1405,6 +1426,53 @@ def _generate_and_score_completions(
             sampling_per_token_logps,
             forward_kwargs,
         ) = self._generate(prompts, images)
+        completions = self.processing_class.batch_decode(completion_ids, skip_special_tokens=True)
+        tool_calls = [extract_tool_calls(completion) for completion in completions]
+        tool_results = [self._tool_dict[tc["name"]](**tc["arguments"]) if tc else None for tc in tool_calls]
+        tool_messages = [
+            [{"role": "tool", "name": tc["name"], "content": str(tr)}] if tc else None
+            for tc, tr in zip(tool_calls, tool_results)
+        ]
```
albertvillanova, line 1445: Not sure of this handles potential multiple tool calls in a single completion...

albertvillanova, line 1445: Before the messages with the tool results (`"role": "tool"`), shouldn't we prepend the messages with the tool calls themselves (`"role": "assistant", "tool_calls":...`)? Not sure of this though... A real question! 😅

August-murr:
I suggest that we remove any references to including tools as a parameter in the `GRPOTrainer`:

```
trainer = GRPOTrainer(…,
tools = [tool_1, tool_2]
)
```

While I am not necessarily against chat templates or tool parsing implementations, I believe the attribute `tools` should be eliminated.

To maximize scalability, a better approach would be to create an all-in-one tools sandbox. This means that the only tool the training script interacts with would be the sandboxed code executors, with all the necessary tools defined within it, like using a Docker image that contains the dependencies of the tools to initialize the sandbox 

Therefore, I propose that we only add the parameter `Environment`, which would encompass the code executor along with all initialized tools.

**The environment needs to be responsible for handling tool usage and execution.**

Then you could create your own built-in environments using your preferred parsing and chat templates.

qgallouedec:
Yes, I have seen some works that suggest that when scaling up, multiplying tools works less well than a smaller, more generic set of tools, such as the one you describe. However, it seems to me that the approach proposed here is in fact compatible:

```python
trainer = GRPOTrainer(…,
    tools = [my_big_containerized_all_in_one_tool],
)
```

in the end it's up to the user decide how to design it, most important here is to allow for this flexibility

August-murr:
but the name `tools` as a param in `GRPOTrainer` is misleading.
What I suggest is more like
```py
MyEnv = DefaultEnv(code_executer=my_big_containerized_all_in_one_tool)

trainer = GRPOTrainer(…,
Environment=MyEnv
)

```
The user can then customize tool use in their own environment.

qgallouedec:
the name tool is actually quite standard now:

https://huggingface.co/blog/unified-tool-use

https://huggingface.co/docs/transformers/main/chat_extras

https://huggingface.co/docs/transformers/main/en/internal/tokenization_utils#transformers.PreTrainedTokenizerBase.apply_chat_template.tools

trl/trainer/grpo_trainer.py
qgallouedec: it must have been very late 😂 😂

qgallouedec:
closed in favor of #4300

PR #4101, opened 2025-09-17
Fix VLM configs in generate_tiny_models

albertvillanova:
This PR fixes the configs for VLMs within `scripts/generate_tiny_models.py`. The main change is switching from direct attribute assignment on the config object to passing configuration dictionaries as arguments to `AutoConfig.from_pretrained`, which streamlines and clarifies the configuration process.

Configuration refactoring:

* Replace manual assignment of individual attributes on `config.text_config` and `config.vision_config` with `text_config` and `vision_config` dictionaries passed directly to `AutoConfig.from_pretrained`. This makes the code cleaner and less error-prone.
* Add a `"layer_types": None` entry to `text_config`, indicating it should be set automatically based on `num_hidden_layers`, so we avoid a ValueError

Fix #4099.

## Validation
We can check that the fix works.
Before the fix:
```python
from transformers import AutoModelForSequenceClassification

AutoModelForSequenceClassification.from_pretrained('trl-internal-testing/tiny-Gemma3ForConditionalGeneration')

ValueError: `num_hidden_layers` (2) must be equal to the number of layer types (34)
```
After the fix: We opened a PR in 'trl-internal-testing/tiny-Gemma3ForConditionalGeneration' with the fix: https://huggingface.co/trl-internal-testing/tiny-Gemma3ForConditionalGeneration/discussions/2
```python
from transformers import AutoModelForSequenceClassification

AutoModelForSequenceClassification.from_pretrained('trl-internal-testing/tiny-Gemma3ForConditionalGeneration', revision='refs/pr/2')
```

PR #4118, opened 2025-09-22
🔭 Align param passing to VLM configs in generate_tiny_models

albertvillanova:
Align the remaining VLM config params and pass them as arguments to `AutoConfig.from_pretrained`.

This is the recommended way to update subconfigs: https://github.com/huggingface/transformers/issues/41020#issuecomment-3314879813

Follow-up to:
- #4101


qgallouedec:
merging so that I can use it in the sequence of PRs #4113, #4114, #4115

Issue #4061, opened 2025-09-11
GRPO on InternVL3.5 does not support dynamic vision tokens


sanchit97:
### Description
The issue likely occurs because InternVL uses a dynamic amount of Vision tokens, which are replaced by <IMG_CONTEXT>.
Setting the size of each image to a fixed resolution fixes the error, but it is less than ideal.

### Reproduction

```python
training_args = GRPOConfig(
        output_dir=test
        bf16=True,
        remove_unused_columns = False,
        per_device_train_batch_size=4,
        num_train_epochs=4,
        logging_steps=50,
        max_prompt_length = 4096,
        eval_strategy="steps",
        eval_steps=500,
        max_completion_length = 512,
        num_generations = 4,
        learning_rate = 2e-7, 
        )

        trainer = GRPOTrainer(
            model=model, # internvl3.5 type
            args=training_args,
            reward_funcs=[..],
            train_dataset=grpo_train_dataset,
            eval_dataset=grpo_eval_dataset,  
            processing_class=processor,
        )
        
        trainer.train()

```

outputs:

```
Traceback (most recent call last):
  ...
[rank0]:   File "/mnt/home/../miniconda3/envs/../lib/python3.10/site-packages/transformers/models/internvl/modeling_internvl.py", line 654, in get_placeholder_mask
[rank0]:     raise ValueError(
[rank0]: ValueError: Image features and image tokens do not match: tokens: 1536, features 512
```

### System Info

trl==0.22

### Checklist

- [x] I have checked that my issue isn't already filed (see [open issues](https://github.com/huggingface/trl/issues?q=is%3Aissue))
- [x] I have included my system information
- [x] Any code provided is minimal, complete, and reproducible ([more on MREs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks))
- [x] Any code provided is properly formatted in code blocks, (no screenshot, [more on code blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks))
- [x] Any traceback provided is complete

ycma8:
Maybe try increasing max_prompt_length and see if that helps.

sanchit97:
> Maybe try increasing max_prompt_length and see if that helps.

I did try that, to no success. Have you had success in doing that?

ckddls1321:
[...]

PR #4121, opened 2025-09-22
[GRPO] Ensure max_prompt_length can accommodate protected tokens

kashif:
fixes issue #4061

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1114,8 +1114,19 @@ class GRPOTrainer(Trainer):
             # because we can't use `skip_special_tokens=True` (some special tokens are still needed for generation).
             protected = [self.image_token_id, self.vision_start_token_id, self.vision_end_token_id]
             protected = [token for token in protected if token is not None]
+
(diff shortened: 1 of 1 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -1114,8 +1114,19 @@ def _generate_and_score_completions(
...
+            if protected:
+                protected_tensor = torch.tensor(protected, device=prompt_ids.device)
+                max_protected = max(
+                    torch.isin(prompt_ids[i], protected_tensor).sum().item() for i in range(prompt_ids.shape[0])
+                )
+                effective_max_length = max(self.max_prompt_length, max_protected)
+            else:
+                effective_max_length = self.max_prompt_length
+
             prompt_ids, prompt_mask = truncate_with_protected_tokens(
```
qgallouedec: so here, if you've a prompt that contains more than `self.max_prompt_length` image tokens, you'll effectively get a `prompt_ids.shape[1] > self.max_prompt_length`?

kashif: yes that is correct

kashif:
Closing as the issue is not fixed

PR #4133, opened 2025-09-24
Disable cache in compute_loss only when checkpointing is enabled

cyyever:
# What does this PR do?

Currently `use_cache` may be unconditionally disabled on certain models. It should be respected if `checkpointing` is explicitly set to `False` by the user.


qgallouedec:
Thanks, can you elaborate why you want the cache to be enabled?

cyyever:
The simple reason is let the user use what they specify. It is indeed strange to use cache in training... But it may help in eval

qgallouedec:
But I don't understand how cache is useful in the forward, since we don't pass nor re-use this cache

cyyever:
You're right, let me check the trainer code of transformers

PR #4152, opened 2025-09-26
Use `prompt_ids` from generation

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/tests/test_vllm_client_server.py
+++ b/tests/test_vllm_client_server.py
@@ -13,10 +13,8 @@
 # limitations under the License.
 
 import os
-import signal
 import subprocess
 
-import psutil
 import pytest
 from transformers import AutoModelForCausalLM
 from transformers.testing_utils import require_torch_multi_accelerator, torch_device
@@ -24,7 +22,7 @@ from transformers.testing_utils import require_torch_multi_accelerator, torch_de
 from trl.extras.vllm_client import VLLMClient
 from trl.scripts.vllm_serve import chunk_list
 
-from .testing_utils import TrlTestCase, require_3_accelerators
+from .testing_utils import TrlTestCase, kill_process, require_3_accelerators
 
 
 class TestChunkList(TrlTestCase):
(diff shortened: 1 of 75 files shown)


tests/test_vllm_client_server.py
```diff
@@ -74,36 +74,42 @@ def setup_class(cls):
...
         prompts = ["Hello, AI!", "Tell me a joke"]
-        outputs = self.client.generate(prompts)["completion_ids"]
+        outputs = self.client.generate(prompts)
+        prompt_ids = outputs["prompt_ids"]
+        completion_ids = outputs["completion_ids"]
 
-        # Check that the output is a list
-        assert isinstance(outputs, list)
+        # Check that the outputs are lists
+        assert isinstance(prompt_ids, list)
```
lewtun, line 82: nit: I'm not sure if we changed our conventions, but I think the previous way was to rely on the `unittest` methods like this:

```suggestion
         self.assertIsInstance(prompt_ids, list)
```

lewtun (approved):
Thanks for the clean refactor! 

Overall LGTM with a question about whether the difference across benchmark runs is within the variance of a repeated run (i.e. if you run one of these again on `main` does it also fluctuate from the original run on `main`)?

Screenshot 2025-10-07 at 07 46 10

tests/test_vllm_client_server.py
qgallouedec: yes, now we use pytest instead of unittest, see https://github.com/huggingface/trl/pull/4188

qgallouedec:
> with a question about whether the difference across benchmark runs is within the variance

For some reason, even with

```python

# Strctly ensure reproducibility
import random
import numpy as np
import torch
random.seed(42)
np.random.seed(42)
torch.manual_seed(42)
torch.cuda.manual_seed_all(42)
torch.backends.cudnn.deterministic = True
torch.backends.cudnn.benchmark = False

args=GRPOConfig(
    ...,
    seed=42,
)

```

I cannot have exact same results between runs after the second step.

green/yellow: main
blue/red: this PR

Screenshot 2025-10-07 at 11 38 41 AM

PR #4146, opened 2025-09-26
🧺 [1/N] Refactor `_generate` in GRPO/RLOO: list of ints instead of tensors

qgallouedec:
This PR is the third of a sequence of PR (after https://github.com/huggingface/trl/pull/4111, https://github.com/huggingface/trl/pull/4113 and https://github.com/huggingface/trl/pull/4114) that aims to refactor the generation part of GRPO/RLOO to allow for easier customization and ultimately tool calling

The idea with this PR is to make `_generate` return list of ints instead of tensors. This will help a lots when implementing tool calling.

Several modifications:

1. `truncate_with_protected_tokens`: instead of operating on 2D tensors (ids and mask), it will operate on sequence ids directly:

    before
    
    ```python
    >>> ids = torch.tensor([[1, 2, 3], [4, 5, 6]])
    >>> mask = torch.tensor([[1, 1, 1], [0, 0, 1]])
    >>> truncate_with_protected_tokens(ids, mask, target_length=2, protected_tokens=[])
    tensor([[2, 3], [5, 6]]), tensor([[1, 1], [0, 1]])
    ```
    
    after
    
    ```python
    >>> ids = [[1, 2, 3], [4, 5, 6]]
    >>> [truncate_with_protected_tokens(seq, target_length=2, protected_tokens=[]) for seq in ids]
    [[2, 3], [5, 6]]
    ```

2. `_generate` now returns list of ids, instead of tensor + mask
    conversion to tensor is handle in `_generate_and_score_completions`

3. The generation part is moved to a function `_generate_single_turn`, which is called by `_generate`

PR #4153, opened 2025-09-26
🧺 [3/N] Refactor `_generate` in GRPO/RLOO: Rely on generator for prompt truncation

qgallouedec:
This PR belongs to a sequence of PR that aims to refactor the generation part of GRPO/RLOO to allow for easier customization and ultimately tool calling

Previous:
- #4111
- #4113
- #4114
- #4146
- #4152

Next:
- #4154 
- #4155

The idea in this PR is to rely on the inference engine (vLLM, transformers) for the prompt truncation.
- For transformers, it's just about passing `truncation=True` and `max_length=self.max_prompt_length`
- For vLLM we need to pass `truncate_prompt_tokens=self.max_prompt_length`

The other major change in this PR is the removal of the custom, complex prompt truncation logic for visual data. I believe this responsibility shouldn’t fall on TRL—especially since the current approach is difficult to maintain, only works for a handful of models, and likely doesn’t provide much benefit. If prompt truncation with image tokens isn’t supported by design in Transformers, then TRL shouldn’t reimplement or partially support it.

### DIFF
--- a/tests/test_dpo_trainer.py
+++ b/tests/test_dpo_trainer.py
@@ -33,12 +33,18 @@ from transformers import (
 from transformers.testing_utils import (
     get_device_properties,
     require_liger_kernel,
-    require_torch_gpu_if_bnb_not_multi_backend_enabled,
 )
 
 from trl import DPOConfig, DPOTrainer, FDivergenceType
 
-from .testing_utils import TrlTestCase, require_bitsandbytes, require_no_wandb, require_peft, require_vision
+from .testing_utils import (
+    TrlTestCase,
+    require_bitsandbytes,
+    require_no_wandb,
+    require_peft,
+    require_torch_gpu_if_bnb_not_multi_backend_enabled,
+    require_vision,
+)
 
 
 if is_vision_available():
--- a/tests/test_peft_models.py
+++ b/tests/test_peft_models.py
@@ -16,12 +16,11 @@ import os
 
 import torch
 from transformers import AutoModelForCausalLM
-from transformers.testing_utils import require_torch_gpu_if_bnb_not_multi_backend_enabled
 from transformers.utils import is_peft_available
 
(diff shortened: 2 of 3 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -1087,58 +1085,12 @@ def _generate_single_turn(self, prompts: list[str], images: Optional[list]):
...
-                        ]
-                    else:
-                        # If vision_end_token_id is None, just remove the image tokens
-                        prompts_text = [re.sub(rf"({escaped_img_token})+", "", text) for text in prompts_text]
+        if images is not None:
+            prompt_inputs = self.processing_class(text=prompts_text, return_tensors="pt", **kwargs)
+            prompt_inputs = super()._prepare_inputs(prompt_inputs)
+            forward_kwargs = {k: v for k, v in prompt_inputs.items() if k not in ["input_ids", "attention_mask"]}
+        else:
+            forward_kwargs = {}
```
qgallouedec, line 1094: this part is just to get the forward kwargs, but it will be moved outside this function in a subsequent pr

trl/trainer/grpo_trainer.py
```diff
@@ -1311,10 +1265,18 @@ def _generate_single_turn(self, prompts: list[str], images: Optional[list]):
 
         else:
             # Regular generation path
-            prompt_ids = [torch.tensor(ids, device=device) for ids in prompt_ids]
-            prompt_mask = [torch.ones_like(ids, dtype=torch.long) for ids in prompt_ids]
-            prompt_ids = pad(prompt_ids, padding_value=self.pad_token_id, padding_side="left")
-            prompt_mask = pad(prompt_mask, padding_value=0, padding_side="left")
+            self.processing_class.truncation_side = "left"  # ensure left truncation for generation
```
```suggestion
            self.processing_class.truncation_side = "left"
```

PR #4155, opened 2025-09-26
🧺 [5/N] Refactor `_generate` in GRPO/RLOO: Insert images in the prompt

qgallouedec:
…y and enhance handling of structured messages

# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/data_utils.py
+++ b/trl/data_utils.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import copy
 from collections import defaultdict, deque
 from collections.abc import Sequence
 from itertools import takewhile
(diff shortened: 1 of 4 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -1069,23 +1074,9 @@ def _calculate_rewards(self, inputs, prompts, completions, completion_ids_list):
         rewards_per_func = gather(rewards_per_func)
         return rewards_per_func
 
-    def _generate_single_turn(self, prompts: list[str], images: Optional[list]):
+    def _generate_single_turn(self, prompts: list):
```
lewtun, line 1077: nit: should this be the correct type?
```suggestion
    def _generate_single_turn(self, prompts: list[Any]):
```

trl/trainer/grpo_trainer.py
```diff
@@ -1098,38 +1089,35 @@ def _generate_single_turn(self, prompts: list[str], images: Optional[list]):
...
-                            top_p=self.top_p,
-                            top_k=-1 if self.top_k is None else self.top_k,
-                            min_p=0.0 if self.min_p is None else self.min_p,
-                            max_tokens=self.max_completion_length,
-                            truncate_prompt_tokens=self.max_prompt_length,
-                            guided_decoding_regex=self.guided_decoding_regex,
-                            generation_kwargs=self.args.generation_kwargs,
-                        )
+                        if is_conversational({"prompt": ordered_set_of_prompts[0]}):
+                            output = self.vllm_client.chat(prompts=ordered_set_of_prompts, **sampling_params)
```
lewtun, line 1118: FYI there are subtle differences between the chat and completions endpoints regarding BOS tokens: https://github.com/vllm-project/vllm/issues/9519

I think they are largely fixed, but good to be aware of

trl/trainer/grpo_trainer.py
```diff
@@ -1176,31 +1164,18 @@ def _generate_single_turn(self, prompts: list[str], images: Optional[list]):
                 if self.vllm_tensor_parallel_size > 1:
                     # Gather prompts from all ranks in the TP group and flatten.
                     # Each rank starts with its own prompts; after gathering, all ranks see the full group set.
-                    orig_size = len(prompts_text)
+                    orig_size = len(prompts)
                     gathered_prompts = [None for _ in range(self.vllm_tensor_parallel_size)]
-                    torch.distributed.all_gather_object(gathered_prompts, prompts_text, group=self.tp_group)
```
lewtun, line 1181: So much cleaner now!

trl/data_utils.py
```diff
@@ -50,31 +62,85 @@ def prepare_multimodal_messages(messages: list[dict[str, Any]], num_images: int)
...
+                img_idx += 1
+
+    return messages
+
+
+def prepare_multimodal_messages_vllm(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    # docstyle-ignore  # because <Image> is not parsable in the code block
+    """
+    Convert structured multimodal messages into a format compatible with vLLM. Replaces `"type": "image"` blocks with
+    `"type": "image_pil"` blocks, and `"image": Image` with `"image_pil": Image`.
```
lewtun, line 116: Why do we need a separate convention for vLLM vs the default? In other words, would it be simpler to just have `image_pil` as the type or does that cause issues with other uses?

trl/trainer/grpo_trainer.py
qgallouedec: Functionally, it’s the same thing, but I feel like `list[Any]` is less correct — it suggests that the list could contain absolutely anything, which isn’t true. The proper type hint would actually be something more like:

```python
list[str] | list[list[dict[str, str]]] | list[list[dict[str, list[dict[str, str]]]]]
```

…but that’s really hard to read.

So by just writing `list`, it feels like a reasonable compromise: it’s a simple type hint — not entirely accurate, but not wrong either.

trl/trainer/grpo_trainer.py
qgallouedec: Good point — for reference, this is likely due to the point described here: https://huggingface.co/blog/qgallouedec/gotchas-in-tokenizer-behavior#7-chat-template-and-tokenization-dont-compose-due-to-special-tokens.

- In my experience, this only affects a very small number of models.
- I think it’s reasonable to rely on vLLM to have addressed it by now, and we don't want to "patch" vLLM here

trl/data_utils.py
qgallouedec: because vLLM only supports `{"type": "image_pil", "image_pil": <PIL.Image>}` and transformers only supports `{"type": "image", "image": <PIL.Image>}` :)

crying-happy

PR #4178, opened 2025-09-30
Fix PEFT interlinks in docstrings

albertvillanova:
Fix PEFT interlinks in docstrings

The interlink will be properly rendered (as it is already the case for other HF libs, like `transformers` or `datasets`) once this PR in `doc-builder` is merged:
- https://github.com/huggingface/doc-builder/pull/628

### DIFF
--- a/trl/trainer/gkd_trainer.py
+++ b/trl/trainer/gkd_trainer.py
@@ -85,7 +85,7 @@ class GKDTrainer(SFTTrainer):
         preprocess_logits_for_metrics (`Callable`, *optional*):
             Function to preprocess the logits before computing the metrics. Must take in the `logits` and `labels` and
             return the logits to be used for metrics computation.
-        peft_config ([`~peft.config.PeftConfig`], *optional*):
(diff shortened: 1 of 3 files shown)


trl/trainer/gkd_trainer.py
```diff
@@ -85,7 +85,7 @@ class GKDTrainer(SFTTrainer):
         preprocess_logits_for_metrics (`Callable`, *optional*):
             Function to preprocess the logits before computing the metrics. Must take in the `logits` and `labels` and
             return the logits to be used for metrics computation.
-        peft_config ([`~peft.config.PeftConfig`], *optional*):
+        peft_config ([`~peft.PeftConfig`], *optional*):
```
qgallouedec, line 88: there is a similar mistake in PPO as well

albertvillanova: Oups, I had missed that! :sweat_smile:

albertvillanova: Done!

PR #4174, opened 2025-09-30
Fix docstrings with 'deprecated' Sphinx directive

albertvillanova:
This PR replaces the `.. deprecated::` Sphinx directives with the `<Deprecated>` tag in the docstrings of:
- `NashMDTrainer`
- `XPOTrainer`

The Sphinx directives were introduced in:
- #3783

PR #4179, opened 2025-09-30
Fix docstring interlink to parent class for NashMDTrainer and XPOTrainer

albertvillanova:
Fix docstring interlink to parent class for NashMDTrainer and XPOTrainer.

Note that they were linking to the `OnlineDPOConfig` config class, instead of the `OnlineDPOTrainer` trainer class as their parent class.

Follow-up to:
- #4174


kashif (approved):

albertvillanova:
The failing CI with dev dependencies are unrelated. See comment in fixing PR: https://github.com/huggingface/trl/pull/4139#pullrequestreview-3284871581
- #4139

PR #4183, opened 2025-09-30
🐍 Drop Python 3.9

qgallouedec:
# What does this PR do?

Python 3.14 release: 2025-10-07, see https://devguide.python.org/versions/

TRL usage:

Screenshot 2025-09-30 at 9 15 31 AM

Main changes:
- Drop support for Python 3.9; Ruff target python version 3.9 -> 3.10
- `Union[X, Y]` becomes `X | Y` (syntax Python 3.10+)
- `Optional[X] becomes `X | None`
- Pass `strict: bool` when using `zip`. In most cases, it's `True`, expect in `apply_chat_template`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,6 @@ dependencies = [
     "accelerate>=1.4.0",
     "datasets>=3.0.0",
     "transformers>=4.56.1",
-    "transformers!=4.57.0; python_version == '3.9'"
 ]
 dynamic = ["version"]
 
--- a/trl/__init__.py
+++ b/trl/__init__.py
@@ -20,18 +20,6 @@ from typing import TYPE_CHECKING
 from .import_utils import _LazyModule
 
 
-if sys.version_info[:2] == (3, 9):
-    warnings.warn(
-        (
-            "Support for Python 3.9 will be dropped in the next release "
-            "(after its end-of-life on October 31, 2025). "
-            "Please upgrade to Python 3.10 or newer."
-        ),
-        category=FutureWarning,
-        stacklevel=2,
-    )
-
-
 try:
     __version__ = version("trl")
 except PackageNotFoundError:


.github/workflows/tests.yml
albertvillanova, line 62: I think this setting is necessary to allow 3.14 release candidate:
```suggestion
          allow-prereleases: true
```

See: https://github.com/huggingface/trl/actions/runs/18134859419/job/51610824765?pr=4183
```python
Error: The version '3.14' with architecture 'x64' was not found for this operating system.
```

albertvillanova (changes requested):

trl/data_utils.py
```diff
@@ -174,22 +174,26 @@ def apply_chat_template(
...
 
             chosen = prompt_chosen[len(prompt) :]
         if "rejected" in example and "prompt" in example:  # explicit prompt
             prompt_rejected = tokenizer.apply_chat_template(
                 example["prompt"] + example["rejected"], tools=tools, tokenize=False, **template_kwargs
             )
             # Handle DeepSeek-R1 <tool_call> token, see the above comment for details
-            prompt = "".join(x for x, _ in takewhile(lambda x: x[0] == x[1], zip(prompt, prompt_rejected)))
+            prompt = "".join(
+                x for x, _ in takewhile(lambda x: x[0] == x[1], zip(prompt, prompt_rejected, strict=False))
```
qgallouedec, line 267: only exceptions for `strict=False` I think

.github/workflows/tests.yml
qgallouedec: It seems like torch (or torchvision) doesn't yet support python 3.14. I suggest adding python 3.14 in a future PR

albertvillanova (commented):
I see that you removed the CI tests for Python 3.9, which makes sense given its upcoming end of life.

However, I noticed that you're also replacing `typing.Optional` and `typing.Union`. Since these are not deprecated, these changes don't seem strictly necessary. Are we intentionally introducing incompatibility with Python 3.9, even if we're no longer officially supporting it?

pyproject.toml
```diff
@@ -21,13 +21,12 @@ classifiers = [
...
     "Operating System :: OS Independent",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13"
 ]
-requires-python = ">=3.9"
+requires-python = ">=3.10"
```
albertvillanova, line 29: OK, I just saw this now! 😅

qgallouedec:
You're right, it's not strictly needed. And this is a result of upgrading the ruff target version to py310

This would indeed break the support for python3.9. Which I think is ok, only 1% of users are still using it, and I suspect that these users actually use an old version of TRL

qgallouedec:
Wdyt?

edbeeching (approved):

PR #4193, opened 2025-10-02
[Online-DPO] fix the completion_len == max_new_tokens crash

kashif:
# What does this PR do?

Fix a crash when we have Empty prompt after truncation due to conservative configs.  Else we get a off-by-1 error
also prepeare the ref_model properly

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,6 +16,7 @@ on:
 env:
   TQDM_DISABLE: 1
   CI_SLACK_CHANNEL: ${{ secrets.CI_PUSH_MAIN_CHANNEL }}
+  PYTORCH_CUDA_ALLOC_CONF: "expandable_segments:True"
 
 jobs:
   check_code_quality:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,35 @@
+# Copyright 2020-2025 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import gc
+
+import pytest
+import torch
+
+
+@pytest.fixture(autouse=True)
+def cleanup_gpu():
+    """
+    Automatically cleanup GPU memory after each test.
+
+    This fixture helps prevent CUDA out of memory errors when running tests in parallel
+    with pytest-xdist by ensuring models and tensors are properly garbage collected
+    and GPU memory caches are cleared between tests.
+    """
+    yield
+    # Cleanup after test
+    gc.collect()
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
+        torch.cuda.synchronize()


trl/trainer/online_dpo_trainer.py
```diff
@@ -58,7 +58,7 @@
 from ..extras.vllm_client import VLLMClient
 from ..import_utils import is_vllm_available
 from ..models import create_reference_model, prepare_peft_model
-from ..models.utils import unwrap_model_for_generation
+from ..models.utils import prepare_deepspeed, unwrap_model_for_generation
```
albertvillanova: I guess you meant:
```suggestion
from ..models.utils import prepare_fsdp, unwrap_model_for_generation
```

trl/trainer/online_dpo_trainer.py
```diff
@@ -591,9 +591,23 @@ def __init__(
...
+            # Prepare ref model for FSDP training
+            if self.ref_model is not None:
+                self.ref_model = prepare_fsdp(self.ref_model, self.accelerator)
+            if self.reward_funcs is not None:
+                for i, reward_func in enumerate(self.reward_funcs):
+                    if isinstance(reward_func, PreTrainedModel):
+                        # Set device placement to True to make `prepare_model` move `reward_func` to device when using fsdp
+                        self.reward_funcs[i] = self.accelerator.prepare_model(
+                            reward_func, evaluation_mode=True, device_placement=True
+                        )
```
albertvillanova: Refactoring, but feel free to ignore it:
```suggestion
                # Set device placement to True to make `prepare_model` move `reward_func` to device when using fsdp
                self.reward_funcs = [
                    self.accelerator.prepare_model(reward_func, evaluation_mode=True, device_placement=True)
                    if isinstance(reward_func, PreTrainedModel)
                    else reward_func
                    for reward_func in self.reward_funcs
                ]
```

kashif:
thanks @albertvillanova fixed.. note this fixes an edge case when we have conservative configs and all the prompts get clipped

kashif:
@albertvillanova  added a 3rd fix for updating the vllm weights only when needed

trl/trainer/online_dpo_trainer.py
qgallouedec: it seems duplicated with the above?

qgallouedec:
What setting results in completely removing the prompt?

kashif:
@qgallouedec  when the `max_tokens` is less than `max_new_tokens` i thnk

kashif:
i believe here it was returning  a large number so that the  `prompt_ids` are empty
```
        num_tokens_to_truncate = max(prompt_ids.size(1) + completion_ids.size(1) - self.max_length, 0)

```

trl/trainer/online_dpo_trainer.py
kashif: Yes the ref model prep. is not duplicated, but the reward model in both cases just needs to be on the appropriate device, so we have that logic duplicated

trl/trainer/online_dpo_config.py
```diff
@@ -412,3 +412,11 @@ def __post_init__(self):
 
         if hasattr(self.beta, "__len__") and len(self.beta) == 1:
             self.beta = self.beta[0]
+
+        if self.max_new_tokens >= self.max_length:
+            warnings.warn(
+                f"The configuration has `max_new_tokens` ({self.max_new_tokens}) >= `max_length` ({self.max_length}). "
+                "This will cause prompts to be truncated or completely removed in the forward pass. "
+                "To preserve prompts, ensure  e.g. `max_length > max_new_tokens + 512`. ",
+                UserWarning,
```
qgallouedec:
```suggestion
            logger.warning(
                f"The configuration has `max_new_tokens` ({self.max_new_tokens}) >= `max_length` ({self.max_length}). "
                "This will cause prompts to be truncated or completely removed in the forward pass. "
                "To preserve prompts, ensure  e.g. `max_length > max_new_tokens + 512`. ",
```

trl/trainer/online_dpo_trainer.py
qgallouedec: Right! So currently we've this:

```python
if is_deepspeed_enabled:
    if ref_model is not None:
        ref_model = prepare_deepspeed(ref_model)
    if reward_funcs is not None:
        reward_funcs = prepare_deepspeed(reward_funcs)

elif is_fsdp_enabled:
    if ref_model is not None:
        ref_model = prepare_fsdp(ref_model)
    if reward_funcs is not None:
        reward_funcs = accelerator.prepare_model(reward_funcs)

else:
    if ref_model is not None:
        ref_model = accelerator.prepare_model(ref_model)
    if reward_funcs is not None:
        reward_funcs = accelerator.prepare_model(reward_funcs)
```

I'd recommend this instead, this would align with GRPO/RLOO and would be easier to maintain, wdyt?


```python
if ref_model is not None:
    if is_deepspeed_enabled:
        ref_model = prepare_deepspeed(ref_model)
    elif is_fsdp_enabled:
        ref_model = prepare_fsdp(ref_model)
    else:
        ref_model = accelerator.prepare_model(ref_model)

if reward_funcs is not None:
    if is_deepspeed_enabled:
        reward_funcs = prepare_deepspeed(reward_funcs)
    else:
        reward_funcs = accelerator.prepare_model(reward_funcs)
```

I think you can just copy-paste from grpo

PR #4194, opened 2025-10-02
Replace setup with pyproject and fix packaging unintended modules

albertvillanova:
Replace setup with pyproject, as discussed internally in HF open source team.


pyproject.toml
```diff
@@ -1,3 +1,132 @@
...
+    "requests",
+    "uvicorn",
+    "Pillow",
+    "torchvision",
+    "num2words==0.5.14"
+]
+
+[tool.setuptools]
+include-package-data = true
+
```
qgallouedec, line 120:
```suggestion
[tool.setuptools.package-data]
"trl" = [
    "templates/*.md",
    "accelerate_configs/*.yaml",
    "LICENSE",
    "CONTRIBUTING.md",
    "README.md",
]

[tool.setuptools.exclude-package-data]
"*" = ["__pycache__"]
```

maybe we can also drop `MANIFEST.in` and have everything here

pyproject.toml
```diff
@@ -1,3 +1,132 @@
...
+[tool.setuptools.dynamic]
+version = { file = "VERSION" }
+
+[tool.setuptools.packages.find]
+where = ["."]
```
qgallouedec:
```suggestion
where = ["trl*"]
```

otherwise building will include the examples, which we don't want

qgallouedec: and maybe you don't even need `exclude = ["tests*"]` in this case

qgallouedec: I'm realizing that `setup.cfg` was actually mistakenly  including examples and scripts.

albertvillanova:
> I'm realizing that setup.cfg was actually mistakenly including examples and scripts.

Yes, @qgallouedec!! 

Initially, my goal was just to replicate the previous behavior while migrating from `setup.cfg` to `pyproject.toml`.

But after looking deeper, I noticed that we were previously packaging and distributing unintended modules!

I did some investigation and found that we're not the only ones. For example, see the issue I opened in `spanner-graph-notebook` (which is preinstalled in all Collab environments):
- https://github.com/cloudspannerecosystem/spanner-graph-notebook/issues/61

So this PR will actually fix a long-standing packaging issue on our side as well.

albertvillanova:
At the end I had to keep `MANIFEST.in` because it is the only way we can exclude `tests` from being packaged due to default behavior of `setuptools` backend. See: https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html#controlling-files-in-the-distribution
> More precisely, the following files are included in a source distribution by default:
> ...
> - Files that match the following glob patterns: `tests/test*.py`, `test/test*.py`;

PR #4197, opened 2025-10-02
Remove tokenizer creation from `sft` example script

sergiopaniego:
# What does this PR do?

In the `sft` example script, the tokenizer is currently created before creating the trainer. This causes errors for VLM since they use processor. It can safely be removed since it's created inside the trainer

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -14,6 +14,5 @@ jobs:
       commit_sha: ${{ github.sha }}
       package: trl
       version_tag_suffix: ""
-      custom_container: huggingface/transformers-doc-builder
     secrets:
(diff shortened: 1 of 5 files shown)


sergiopaniego:
We also have `sft_vlm.py` and more example scripts. I think we can keep those for some time and then drop them, creating a generic one for all the cases. wdyt?

qgallouedec:
> We also have sft_vlm.py and more example scripts. I think we can keep those for some time and then drop them, creating a generic one for all the cases. wdyt?

I think that unless we have a good reason to create the tokenizer outside the trainer, we should let the trainer create it. I'm fine with removing in other scripts now if you want

PR #4201, opened 2025-10-03
Fix prompt-completion labeling with add_generation_prompt and warning

### LINKED ISSUE #4147: Prompt-completion does not use correct labels when passing conversational format to SFT
### Reproduction

Hi!

With the switch to the new dataset format types for SFT I was wondering how to properly provide the data for training, in my case prompt-completion
I have used `DataCollatorForCompletionOnlyLM` in the past and now want to update to the newest trl version to get all the nice new features.

In the past, I had my raw data in the message format:
```
# example 1
example = {
    "messages": [
        {"role": "system", "content": "SYSTEM"},
        {"role": "user", "content": "QUESTION"},
        {"role": "assistant", "content": "ANSWER"},
    ]
}
```
And I applied the chat template on it, which gave me something like this (generic chat tempalte):
```
<system>SYSTEM<end><user>QUESTION<end><assistant>ANSWER<end>
```
With the `DataCollatorForCompletionOnlyLM` I had to specify the instruction and response templates as `<user>` and `<assistant>` respectively and it then correctly gave me the labels and masks.

Now with the new format (as described [here](https://huggingface.co/docs/trl/v0.23.0/en/dataset_formats#conversational)) I need to bring my data in the `prompt-completion` format.
```
# example 2
example = {
    "prompt": [
        {"role": "system", "content": "SYSTEM"},
        {"role": "user", "content": "QUESTION"}
    ],
    "completion": [
        {"role": "assistant", "content": "ANSWER"}
    ]
}
```

Now to my question: How do I pass these examples to SFT?

Naively, I would have assumed that I can just pass the examples in the role-content format to SFT and the chat template is applied accordingly.
However, when inspecting the tokenized inputs and labels I see the following
```
Input: <system>SYSTEM<end><user>QUESTION<end><assistant>ANSWER<end>
Label: --------------------------------------<assistant>ANSWER<end>
```
I.e. the generation prompt is added in the labels and not in the prompt.

When I process example 2 with `apply_chat_template` before (as recommended [here](https://huggingface.co/docs/trl/v0.23.0/en/dataset_formats#working-with-conversational-datasets-in-trl)), the example looks as follows:
```
# processed example 2
example = {
    "prompt": "<system>SYSTEM<end><user>QUESTION<end><assistant>",
    "completion": "ANSWER<end>"
}
```

And the corresponding input and label is as I would have expected it 
```
Input: <system>SYSTEM<end><user>QUESTION<end><assistant>ANSWER<end>
Label: -------------------------------------------------ANSWER<end>
```

Comparing how the chat template is applied in `apply_chat_template` and `_prepare_dataset` ([here](https://github.com/huggingface/trl/blob/main/trl/trainer/sft_trainer.py#L968)) the generation prompt is added in the former since the last role in `prompt` is naturally `user`.

I am leaning towards saying that this is a bug in `_prepare_dataset` but could also be that I have misunderstood how to use the new formats.

Any help is very much appreciated!

Here are the minimal examples:
```
# Passing Conversational format directly
from trl import SFTTrainer
from datasets import Dataset

example = Dataset.from_dict({
    "prompt": [[
        {"role": "system", "content": "SYSTEM"},
        {"role": "user", "content": "QUESTION"}
    ]],
    "completion": [[
        {"role": "assistant", "content": "ANSWER"}
    ]]
})

trainer = SFTTrainer(
    model="Qwen/Qwen3-0.6B",
    train_dataset=example,
)

tokenized_data = trainer.data_collator.torch_call(trainer.train_dataset)
print(repr(trainer.processing_class.decode(tokenized_data["input_ids"][0].tolist())))
print(repr(trainer.processing_class.decode([i for i in tokenized_data["labels"][0].tolist() if i != -100])))
```

```
# Applying chat template before with apply_chat_template and passing standard format
from datasets import Dataset
from trl import apply_chat_template
from trl import SFTTrainer

example = Dataset.from_dict({
    "prompt": [[
        {"role": "system", "content": "SYSTEM"},
        {"role": "user", "content": "QUESTION"}
    ]],
    "completion": [[
        {"role": "assistant", "content": "ANSWER"}
    ]]
})


tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3-0.6B")
example = example.map(apply_chat_template, fn_kwargs={"tokenizer": tokenizer})

trainer = SFTTrainer(
    model="Qwen/Qwen3-0.6B",
    train_dataset=example,
)

tokenized_data = trainer.data_collator.torch_call(trainer.train_dataset)
print(repr(trainer.processing_class.decode(tokenized_data["input_ids"][0].tolist())))
print(repr(trainer.processing_class.decode([i for i in tokenized_data["labels"][0].tolist() if i != -100])))
```

### System Info

- Platform: Linux-6.8.0-1031-aws-x86_64-with-glibc2.35
- Python version: 3.12.10
- TRL version: 0.23.0
- PyTorch version: 2.8.0
- accelerator(s): NVIDIA A10G
- Transformers version: 4.56.2
- Accelerate version: 1.10.1
- Datasets version: 3.6.0
- HF Hub version: 0.34.4
- bitsandbytes version: 0.46.1
- OpenAI version: 1.95.1
- PEFT version: 0.16.0
- vLLM version: not installed

### Checklist

behroozazarkhalili:
## Summary

Fixes #4147 

This PR implements the solution discussed with @qgallouedec:

- Add `add_generation_prompt=True` for more precise token masking in prompt-completion format
- Add warning when `assistant_masks` are all zeros (chat template lacks `{% generation %}` support)

## Changes

1. **Line 940**: Added `add_generation_prompt=True` to ensure prompt tokens properly include assistant role markers
2. **Lines 980-988**: Added warning when `assistant_masks` contain only zeros, alerting users their chat template is incompatible with `assistant_only_loss=True`

## Behavior

When chat templates lack `{% generation %}` keyword support:
- ⚠️ Warning is logged explaining the incompatibility
- Users are directed to either use a compatible chat template or switch to `completion_only_loss=True`


trl/trainer/sft_trainer.py
```diff
@@ -974,10 +975,17 @@ def tokenize_fn(example, processing_class, dataset_text_field, assistant_only_lo
                                 "token handling. Verify that the tokenizer is processing text consistently."
                             )
 
-                        # Create a completion mask
-                        completion_mask = [0] * len(prompt_ids) + [1] * (len(prompt_completion_ids) - len(prompt_ids))
                         output["input_ids"] = prompt_completion_ids
-                        output["completion_mask"] = completion_mask
```
qgallouedec, line 980: I don't think you want to remove the completion mask

trl/trainer/sft_trainer.py
```diff
@@ -974,10 +975,17 @@ def tokenize_fn(example, processing_class, dataset_text_field, assistant_only_lo
...
+
+                        # Warn if assistant_masks are all zeros (chat template doesn't support {% generation %})
+                        # This means assistant_only_loss won't work correctly with prompt-completion format
```
qgallouedec: I think it means it won't work *at all*, ie with prompt-completion but also with language modeling

trl/trainer/sft_trainer.py
```diff
@@ -974,10 +975,17 @@ def tokenize_fn(example, processing_class, dataset_text_field, assistant_only_lo
...
+                        if "assistant_masks" in output and 1 not in output["assistant_masks"]:
```
qgallouedec: I'm not even sure that this condition is ever met since `return_assistant_tokens_mask=assistant_only_loss` is also missing from the tokenizer args

trl/trainer/sft_trainer.py
behroozazarkhalili: I suddenly removed it. Returned.

trl/trainer/sft_trainer.py
behroozazarkhalili: What should it be, then? I became a bit confused, tbh :)

trl/trainer/sft_trainer.py
behroozazarkhalili: How about the following condition?

```python
if "assistant_masks" in output and 1 not in output["assistant_masks"] and assistant_only_loss:
                            logger.warning(
                                "The chat template does not support generation tokens (missing '{% generation %}' keyword). "
                                "When using prompt-completion format with assistant_only_loss=True, this will result in "
                                "all tokens being masked. Consider using a chat template that supports generation tokens, "
                                "or use completion_only_loss=True instead."
                            )
```
Can we raise a proper error instead of a warning?

trl/trainer/sft_trainer.py
```diff
@@ -974,11 +975,22 @@ def tokenize_fn(example, processing_class, dataset_text_field, assistant_only_lo
...
-                        # Create a completion mask
-                        completion_mask = [0] * len(prompt_ids) + [1] * (len(prompt_completion_ids) - len(prompt_ids))
                         output["input_ids"] = prompt_completion_ids
+
+                        # Create completion_mask - more precise thanks to add_generation_prompt=True
+                        completion_mask = [0] * len(prompt_ids) + [1] * (len(prompt_completion_ids) - len(prompt_ids))
                         output["completion_mask"] = completion_mask
 
+                        # Warn if assistant_masks are all zeros (chat template doesn't support {% generation %})
+                        # This means assistant_only_loss won't work correctly with prompt-completion format
```
```suggestion
                        # Warn if assistant_masks are all zeros (usually means that chat template doesn't support {% generation %})
```

trl/trainer/sft_trainer.py
```diff
@@ -974,11 +975,22 @@ def tokenize_fn(example, processing_class, dataset_text_field, assistant_only_lo
...
+                        if "assistant_masks" in output and 1 not in output["assistant_masks"] and assistant_only_loss:
+                            logger.warning(
+                                "You're using `assistant_only_loss=True` with prompt-completion format, but the chat template "
```
```suggestion
                                "You're using `assistant_only_loss=True` but the chat template "
```

yes but not relevant

trl/trainer/sft_trainer.py
```diff
@@ -974,11 +975,22 @@ def tokenize_fn(example, processing_class, dataset_text_field, assistant_only_lo
...
+                                "doesn't generate valid assistant masks — it may be missing the `{% generation %}` keyword. "
+                                "This will cause all tokens to be masked during training. Please use `completion_only_loss=True` "
+                                "instead to train on completion tokens."
```
```suggestion
                                "This will cause all tokens to be masked during training."
```

no, I think you're mixing completion-only and assistant-only. See https://huggingface.co/docs/trl/en/sft_trainer#train-on-assistant-messages-only and https://huggingface.co/docs/trl/en/sft_trainer#train-on-completion-only

qgallouedec:
Actually I think that assistant-only is currently not supported for prompt-completion dataset. Checking

qgallouedec:
It's wrong actually. I realize that in the LM case we already have an error for this case. The easiest is just to make this error common to both LM and prompt-completion, see my latest commit

behroozazarkhalili:
> It's wrong actually. I realize that in the LM case we already have an error for this case. The easiest is just to make this error common to both LM and prompt-completion, see my latest commit

I saw your point.

PR #4203, opened 2025-10-04
🎨 Support mixing image+text and text-only examples

qgallouedec:
`SFTTrainer`, `GRPOTrainer` and `RLOOTrainer` support multi-image dataset, but don't support mixed dataset (text-only and image dataset). Thanks to @zucchini-nlp in https://github.com/huggingface/transformers/pull/36682, this is now supported in transformers. In this PR we extend the tests and the support for this case. Example

```python

dataset = load_dataset("trl-internal-testing/zen-multi-image", "conversational_prompt_completion", split="train")

trainer = SFTTrainer(
    model="Qwen/Qwen2.5-VL-3B-Instruct",
    args=SFTConfig(
        max_length=None,  # For VLMs, truncating can remove image tokens, leading to errors
    ),
    train_dataset=dataset,
)
trainer.train()
```


qgallouedec:
Thanks for the review; we already have test_training_vlm_multi_image in GRPO and RLOO, but there was a filtering to exclude text-only samples. This PR remove this filter so that we have a truly mixed dataset

sergiopaniego (approved):
lgtm! thanks 😄

PR #4217, opened 2025-10-07
[DOCS] fix prose in lora guide

burtenshaw:
This make the text about lr match the figures in the above table.


sergiopaniego (approved):
lgtm!

PR #4196, opened 2025-10-02
Fix entropy and accuracy calculation for prompt_tuning techniques.

pramodith:
# What does this PR do?
Three of the popularly used peft techniques are:

1. `P-tuning`
2. `Prompt-Tuning`
3. `Prefix-Tuning`

While the first two introduce logit scores for the virtual tokens `prefix-tuning` doesn't. This behavior affects how accuracy and entropy are calculated in `SFTTrainer`. 

This PR makes the required changes to ensure that the trainer doesn't crash and correctly calculates metrics.

Fixes #4168 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [X] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [X] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [X] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #4218, opened 2025-10-07
Fix/attention mask mismatch

KatharinaJacoby:
The issue seems to come from the concatenation of the virtual attention mask, which changes the dimensions of the attention_mask and causes a mismatch with per_token_entropy.
Checking that the dimensions of per_token_entropy and attention_mask match before multiplication. I changed the code block so the attention_mask is truncated to match the sequence length of per_token_entropy. 
And added a unit test:
-set up common test data in the setUp method.
-test the scenario where the attention mask and per-token entropy have matching dimensions.
-test the scenario where the attention mask and per-token entropy have mismatched dimensions.
-test the scenario where position_ids is used instead of attention_mask.
-test the scenario where neither attention_mask nor position_ids is present, expecting a ValueError.

### DIFF
--- a/docs/source/lora_without_regret.md
+++ b/docs/source/lora_without_regret.md
@@ -42,7 +42,7 @@ from trl import SFTTrainer, SFTConfig
 
 dataset = load_dataset("open-thoughts/OpenThoughts-114k", split="train")
 
-peft_config = LoraConfig(lora_r=256, lora_alpha=16, lora_target_modules="all-linear")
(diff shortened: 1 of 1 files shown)


pramodith:
Hey @KatharinaJacoby I think another PR that I opened last week resolves the issue that your PR covers too. #4196

qgallouedec:
closing in favor of #4196, thanks a lot @KatharinaJacoby and @pramodith

PR #4231, opened 2025-10-08
[Utils] fix DataCollatorForChatML

kashif:
# What does this PR do?

Fixes the DataCollatorForChatML when the tokenizer / inputs has the no-think flag


trl/trainer/utils.py
```diff
@@ -173,6 +118,188 @@ def __call__(self, examples: list[dict[str, Any]]) -> dict[str, torch.Tensor]:
...
+        ids = tokenized["input_ids"]
+        mask = tokenized.get("attention_mask")
+        if mask is None:
+            mask = [1] * len(ids)
+        if max_length is not None and len(ids) > max_length:
+            ids = ids[:max_length]
+            mask = mask[:max_length]
+        return {"input_ids": ids, "attention_mask": mask}
+
+    def _extract_enable_thinking_metadata(self, example: Mapping[str, Any]) -> Optional[bool]:
```
albertvillanova, line 272: I think this method could also be made static:
```suggestion
    @staticmethod
    def _extract_enable_thinking_metadata(self, example: Mapping[str, Any]) -> Optional[bool]:
```

albertvillanova (approved):
Thanks!

kashif:
@albertvillanova  its still a bit ugly... so checking how to make it better

kashif:
@albertvillanova  i'll still need to debug this so moving to draft for now

PR #4234, opened 2025-10-08
Fix CI unittest asserts

albertvillanova:
Fix CI unittest asserts: https://github.com/huggingface/trl/actions/runs/18351090937/job/52270992538
```python
AttributeError: 'TestSFTTrainer' object has no attribute 'assertIsNotNone'
```
```python
FAILED tests/test_sft_trainer.py::TestSFTTrainer::test_train_vlm_prompt_completion_gemma - AttributeError: 'TestSFTTrainer' object has no attribute 'assertIsNotNone'
```

This was introduced in:
- #4190

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1466,12 +1466,12 @@ class TestSFTTrainer(TrlTestCase):
         trainer.train()
 
         # Check that the training loss is not None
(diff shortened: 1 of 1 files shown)


albertvillanova:
It's normal: so long used to unitest assertions... :sweat_smile:

PR #4244, opened 2025-10-09
Added SFT LoRA notebook

sergiopaniego:
# What does this PR do?

Add SFT LoRA notebook

Colab: https://colab.research.google.com/github/huggingface/trl/blob/sft-lora-notebook/examples/notebooks/sft_trl_lora_qlora.ipynb

Notebook can be early-reviewed.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/examples/notebooks/sft_trl_lora_qlora.ipynb
+++ b/examples/notebooks/sft_trl_lora_qlora.ipynb
@@ -1,1914 +1,1107 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "id": "ke2YKcr_iw-7"
-      },
-      "source": [
-        "# Supervised Fine-Tuning (SFT) with LoRA/QLoRA using TRL — on a Free Colab Notebook\n",
-        "\n",
-        "[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/huggingface/cookbook/blob/main/notebooks/en/fine_tuning_llm_grpo_trl.ipynb)"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "id": "AXdmDb9kiw-9"
-      },
-      "source": [
-        "![trl banner](https://huggingface.co/datasets/trl-lib/documentation-images/resolve/main/trl_banner_dark.png)"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "id": "6jbXmhLGiw-9"
-      },
-      "source": [
-        "Easily fine-tune Large Language Models (LLMs) or Vision-Language Models (VLMs) with **LoRA** or **QLoRA** using the [**Transformers Reinforcement Learning (TRL)**](https://github.com/huggingface/trl) library built by Hugging Face — all within a **free Google Colab notebook** (powered by a **T4 GPU**.).  \n",
-        "\n",
-        "- [TRL GitHub Repository](https://github.com/huggingface/trl) — star us to support the project!  \n",
-        "- [Official TRL Examples](https://huggingface.co/docs/trl/example_overview)  \n",
-        "- [Community Tutorials](https://huggingface.co/docs/trl/community_tutorials)"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "id": "bYFKZvHOiw--"
-      },
-      "source": [
-        "## Key concepts\n",
-        "\n",
-        "- **SFT**: Trains models from example input-output pairs to align behavior with human preferences.\n",
-        "- **LoRA**: Updates only a few low-rank parameters, reducing training cost and memory.\n",
-        "- **QLoRA**: A quantized version of LoRA that enables even larger models to fit on small GPUs.\n",
-        "- **TRL**: The Hugging Face library that makes fine-tuning and reinforcement learning simple and efficient.\n",
-        "\n",
-        "Learn how to perform **Supervised Fine-Tuning (SFT)** with **LoRA/QLoRA** using **TRL**."
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "id": "Pk7PwfYRiw--"
-      },
-      "source": [
-        "## Install dependencies\n",
-        "\n",
-        "We'll install **TRL** with the **PEFT** extra, which ensures all main dependencies such as **Transformers** and **PEFT** (a package for parameter-efficient fine-tuning, e.g., LoRA/QLoRA) are included. Additionally, we'll install **trackio** to log and monitor our experiments, and **bitsandbytes** to enable quantization of LLMs, reducing memory consumption for both inference and training."
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": null,
-      "metadata": {
-        "id": "bRL1aItXiw--"
-      },
-      "outputs": [],
-      "source": [
-        "!pip install -Uq \"trl[peft]\" trackio bitsandbytes"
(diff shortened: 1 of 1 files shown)


burtenshaw:
Nice work. Reviewing is really hard with nbdime so I'm going to use screenshots. Sorry!

❤️ Capybara. But I think we could go for a more up to date dataset that contains new samples for the model. i.e. a math, code, or reasoning task.
image

I would simplify this cell down to 5 or so models. But I like the explanation.
image

I would also only do QLoRA and really go for the lowest memory. You can add a note saying, why not try without?
image

We should explain target modules/ useful because it's intuitive to understand.
image

qgallouedec:
lgtm! I just simplified the notebook (removed the metadata) 18k lines to 500.

```
nb-clean clean -M examples/notebooks/sft_trl_lora_qlora.ipynb 
```

examples/notebooks/sft_trl_lora_qlora.ipynb
```diff
@@ -0,0 +1,562 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip install -Uq trl trackio"
+   ]
```
albertvillanova: This is slightly misleading because `trl` alone does not install `peft` as a dependency. It worked in Colab only because `peft` happens to be preinstalled there. To make the tutorial reliable in any environment and in case `peft` stops being preinstalled in Colab, I would recommend to explicitly install `trl[peft]`, which guarantees that `peft` is available.
```suggestion
    "We'll install **TRL** with the **PEFT** extra, which ensures all main dependencies such as **Transformers** and **PEFT** are included.\n",
    "Additionally, we'll install **trackio** to log and monitor our experiments."
   ]
  },
  {
   "cell_type": "code",
   "execution_count": null,
   "metadata": {},
   "outputs": [],
   "source": [
    "!pip install -Uq \"trl[peft]\" trackio"
   ]
```

examples/notebooks/sft_trl_lora_qlora.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "\n",
+    "## Granite QLoRA\n",
+    "#model_id, output_dir = \"ibm-granite/granite-4.0-micro\", \"granite-4.0-micro\"  # ✅ ~3.3 GB VRAM"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let's lead the selected model using `transformers`, configuring QLoRA via `bitsandbytes` if needed. We don't need to configure the tokenizer since the trainer takes care of that automatically."
```
albertvillanova: Typo?

```suggestion
    "Let's load the selected model using `transformers`, configuring QLoRA via `bitsandbytes` if needed. We don't need to configure the tokenizer since the trainer takes care of that automatically."
```

examples/notebooks/sft_trl_lora_qlora.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "- [Community Tutorials](https://huggingface.co/docs/trl/community_tutorials)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Learn how to perform **Supervised Fine-Tuning (SFT)** with **LoRA/QLoRA** using **TRL**."
+   ]
+  },
```
albertvillanova: What about adding an introductory section explaining the main concepts? This would help beginner users understand the key concepts before diving into installation or code. On the other hand, I would keep this section short (concise, bullet-style; maybe with links to more extensive explanations), so we avoid users to skip it (if too long) and make it clear the essentials quickly before jumping into the code.

Something like:
```markdown
## Key concepts
- SFT: Trains models from example input-output pairs to align behavior with human preferences.
- LoRA: Updates only a few low-rank parameters, reducing training cost and memory.
- QLoRA: A quantized version of LoRA that enables even larger models to fit on small GPUs.
- TRL: The Hugging Face library that makes fine-tuning and reinforcement learning simple and efficient.
```

Additionally, we could also define `base model` and `lora/qlora adapter`, which are mentioned in the notebook without explaining them.

examples/notebooks/sft_trl_lora_qlora.ipynb
```diff
@@ -0,0 +1,562 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# If using QLoRA\n",
+    "!pip install -Uq bitsandbytes\n",
```
albertvillanova: Maybe worth adding a sentence in the markdown above explaining what `bitsandbytes` is?

examples/notebooks/sft_trl_lora_qlora.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "from huggingface_hub import notebook_login\n",
+    "\n",
+    "notebook_login()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We’ll retrieve your **Hugging Face username** to use it as part of the path for saving and loading the fine-tuned model."
```
albertvillanova: I have a concern here: do we need to handle this with the use of `username`? I think this creates noise within the neat flow of the notebook.
- I think the use of `username` is already the default of the trainer if passed `push_to_hub=True` arg
- If we want to give the users the option to customize this and pass and organization name instead, what about using `hub_model_id` arg instead? 

Nit:
```suggestion
    "We'll retrieve your **Hugging Face username** to use it as part of the path for saving and loading the fine-tuned model."
```

sergiopaniego:
Thanks for the feedback I've updated the notebook based on it!

I still have a few todos to cover (I've also added them at the top of the notebook). Once they're covered, we're ready for a final review 😄 

@burtenshaw  I've updated the dataset to [`HuggingFaceH4/Multilingual-Thinking`](https://huggingface.co/datasets/HuggingFaceH4/Multilingual-Thinking). Contains reasoning traces in different languages and was used in the [gpt-oss recipe](https://cookbook.openai.com/articles/gpt-oss/fine-tune-transfomers). Since it's a reasoning dataset, we may need to add another option for non-reasoning models (I'd just add a comment about it in the notebook). 
I'm now only using QLoRA. I've added more details about it and how to use LoRA if needed.

@qgallouedec should we clean completely the output in the notebook? From a reader view, they can miss some details that are in the outputs (training traces, prints...)

@albertvillanova thanks for the review! I've included your suggestions!

TODOS:
- Check if dataset needs to be adapted (maybe convert it and upload it to org).
- In case of using this reasoning dataset, add details or fallback for non-reasoning models (another dataset).
- Run training for one epoch.
- Update trackio screenshot.

qgallouedec:
> @qgallouedec should we clean completely the output in the notebook? From a reader view, they can miss some details that are in the outputs (training traces, prints...)

I think you can keep the outputs that facilitate understanding and delete the others (pip installation logs, etc.).

sergiopaniego:
Ready for final review!

For the record, I cleaned the notebook with this answer to prevent GitHub from not rendering it: https://github.com/orgs/community/discussions/155944#discussioncomment-14611780

albertvillanova (approved):
Amazing work! Thanks, @sergiopaniego!

I think there is an issue with the image: https://huggingface.co/datasets/trl-lib/documentation-images/resolve/main/sft-lora-notebook-trackio.png
> request_id: 01K80DK5CG94YGDQW5KGG5KZ2H; (10) DB Error: dispatch; code: 2

Apart from that, everything OK from my side.

PR #4247, opened 2025-10-10
[Activation-checkpointing] add tensor dedup and param offloading

kashif:
# What does this PR do?
*  Prevents redundant offloading when multiple tensor views share the same storage
* Tracks and filters out model parameters during offloading


kashif:
@sywangyi would you be kind enough to test this on your hardware and give me some feedback? thank you!

trl/models/activation_offloading.py
```diff
@@ -30,10 +30,31 @@
 if is_torch_npu_available():
     import torch_npu  # noqa: F401
 
+# Try to import DTensor for FSDP v2 support
+try:
+    from torch.distributed._tensor import DTensor
```
qgallouedec: do you know in which version DTensor was introduced? I'm wondering is this try/expect is needed

trl/models/activation_offloading.py
```diff
@@ -152,60 +179,82 @@ def pack_tensor(activation: torch.Tensor) -> int:
...
+                return tensor_id
+
+            # Check if tensor is too small
+            if num_bytes < self.min_tensor_size_bytes:
+                self.tracker[tensor_id] = (activation, False)
+                return tensor_id
+
+            # Check if tensor is a parameter or buffer
+            if isinstance(activation, torch.nn.Parameter) or (
+                hasattr(torch.nn, "Buffer") and isinstance(activation, torch.nn.Buffer)
```
qgallouedec, line 231: same question here, is Buffer a recent addition?

trl/models/activation_offloading.py
SunMarc: For reference we have the following in accelerate: 
```python
if is_torch_version(">=", "2.5.0"):
        from torch.distributed.tensor import DTensor
    else:
        # from torch 2.0.0 (oldest supported accelerate torch version), DTensor is in torch.distributed._tensor
        from torch.distributed._tensor import DTensor
```

but we also need to check for `torch.distributed.is_available()`, otherwise you might get import issue.

trl/models/activation_offloading.py
kashif: no buffer has always been there from the start, I can clean this up

trl/models/activation_offloading.py
```diff
@@ -30,10 +30,31 @@
...
+    share the same underlying storage. From:
+    https://github.com/volcengine/verl/blob/main/verl/utils/activation_offload.py
+
+    Args:
+        tensor: The tensor to get the key for
+
+    Returns:
+        A tuple of (storage_pointer, dtype) that uniquely identifies the tensor's storage
+    """
+    storage_ptr = tensor.untyped_storage().data_ptr() + tensor.storage_offset()
```
S1ro1: Using `data_ptr()` can be a bit tricky with for example TorchAO quantized tensors etc, as those can return 0 for `data_ptr()`. I don't have a concrete example, just something to be aware of.

trl/models/activation_offloading.py
```diff
@@ -368,6 +417,36 @@ def hook(outputs, inputs):
...
+        Args:
+            model: The model whose parameters should be tracked
+        """
+        param_storages = set()
+
+        for p in model.parameters():
+            # For FSDP v2: extract local tensor from DTensor
+            actual_tensor = p
+            if DTensor is not None and isinstance(p, DTensor) and hasattr(p, "_local_tensor"):
+                actual_tensor = p._local_tensor
```
S1ro1, line 547: Again, something to care for. If fp8 is used, it can return 0, viz here: https://github.com/huggingface/accelerate/blob/f0313a64a2f3de359924c85a98ee010c47b846ec/src/accelerate/accelerator.py#L3842

trl/models/activation_offloading.py
S1ro1: Edit: here is an example (i.e. float8linear, which can sometimes happen). https://github.com/huggingface/accelerate/blob/f0313a64a2f3de359924c85a98ee010c47b846ec/src/accelerate/accelerator.py#L3842

sywangyi:
> @sywangyi would you be kind enough to test this on your hardware and give me some feedback? thank you!

pytest tests/test_activation_offloading.py::TestActivationOffloading::test_parameter_filtering
pytest tests/test_activation_offloading.py::TestActivationOffloading::test_tensor_deduplication

these two cases pass in intel xpu

kashif:
@S1ro1 ok i'll just skip FP8 activations

kashif:
@SunMarc I have added support for broadcast and non-contiguous tensors

Issue #2660, opened 2025-01-25
Issues with DPOTrainer and Qwen2-VL processor


baichuanzhou:
Hey guys, I am trying to dig into the DPO implementation for VLM and I encountered this issue:
[Here](https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L642) in the `process_row` function,
```python
processor, tokenizer = processing_class, processing_class.tokenizer  # the processing class is a processor
processed_features = processor(images=features["images"], text=features["prompt"], add_special_tokens=False)

prompt_input_ids = processed_features["input_ids"][0]
pixel_values = processed_features["pixel_values"][0]
```

 `images` will be turned into `pixel_values` by indexing on the first returned `pixel_values`.(I assume this is because dataset format request that inputs should only have one image). However, after playing with Qwen2-VL's processor, I found that Qwen2-VL's processor always returns a 2D tensor, which would make indexing on the first element essensially indexing on the first row of the pixel values.

If this is the case, I don't think `pixel_values` should be handled that way.

Here's the code I used to test Qwen2-VL's processor:
```python
from PIL import Image
from transformers import AutoProcessor

processor = AutoProcessor.from_pretrained('Qwen/Qwen2-VL-7B-Instruct')
image = Image.open('some image')
outputs = processor(images=[image, image], text="<image><image>", return_tensors="pt")
print(outputs.pixel_values.size())
# should be a 2D tensor, in my case it was torch.Size([3128, 1176])
```
And [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen2_vl/image_processing_qwen2_vl.py#L312) I checked the Qwen2-VL's preprocessing logic, it seems that the processor will always return a flattened out 2D tensor.

Wouldn't indexing a 2D image tensor be a problem?

baichuanzhou:
Hi guys, to confirm my assumption, I further did the following experiment:
1. set trace in the `process_row` function like this:
```python
@staticmethod
    def process_row(features, processing_class, max_prompt_length, max_completion_length, add_special_tokens):
        """
        Same as `tokenize_row` but for vision models. Please refer to `tokenize_row` for more information.
        """
        processor, tokenizer = processing_class, processing_class.tokenizer  # the processing class is a processor
        processed_features = processor(images=features["images"], text=features["prompt"], add_special_tokens=False)

        prompt_input_ids = processed_features["input_ids"][0]
        pixel_values = processed_features["pixel_values"][0]
[...]

PR #4257, opened 2025-10-11
Fix DPO Trainer Bug For Qwen2-VL (Issue 2660)

FabianSchuetze:
Fixes # (issue)

Fixes issue https://github.com/huggingface/trl/issues/2660 .  The DPO trainer doesn't work for Qwen2.5-VL models b/c:

- The pre-processing function only ingest tokens worth of one "row" of an image, but the image-encoder is 2-D: https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L762 
- Qwen2.5-VL requires a `image_grid_thw` argument in the batch inputs but that isn't passed: https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L1246 

To solve the issue, I have:
- added a kwarg that allows skipping pre-processing of data. That allows users to write their own collate-fn, as is common practice in the SFT trainer.
- Pass the image_grid_thw` value to the model-args, as is done in the grpo trainer.

## Before submitting
- [x ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [x ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.

## Who can review?
@baichuanzhou raised the issue initially, and I hit on the same problem when using DPO for Qwen vl.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -1257,7 +1257,7 @@ class DPOTrainer(BaseTrainer):
             model_kwargs["image_sizes"] = concatenated_batch["image_sizes"]
         # For Qwen-VL models
         if "image_grid_thw" in concatenated_batch:
-            model_kwargs["image_grid_thw"] = concatenated_batch["image_grid_tw"]
+            model_kwargs["image_grid_thw"] = concatenated_batch["image_grid_thw"]
 
         prompt_attention_mask = concatenated_batch["prompt_attention_mask"]
         completion_attention_mask = concatenated_batch["completion_attention_mask"]
@@ -1505,6 +1505,9 @@ class DPOTrainer(BaseTrainer):
             model_kwargs["pixel_attention_mask"] = concatenated_batch["pixel_attention_mask"]
         if "image_sizes" in concatenated_batch:
             model_kwargs["image_sizes"] = concatenated_batch["image_sizes"]
+        # For Qwen-VL models
+        if "image_grid_thw" in concatenated_batch:
+            model_kwargs["image_grid_thw"] = concatenated_batch["image_grid_thw"]
 
         prompt_input_ids = concatenated_batch["prompt_input_ids"]
         prompt_attention_mask = concatenated_batch["prompt_attention_mask"]


FabianSchuetze:
@qgallouedec I saw you worked on a few DPO patches. Could you take a look at the PR?

defdet:
Thank you for the PR, it would solve my problem as well. However, doesn't your PR still leave first bug unfixed?

> The pre-processing function only ingest tokens worth of one "row" of an image, but the image-encoder is 2-D

The line responsible for the bug is the same: https://github.com/FabianSchuetze/trl/blob/adb34f339895e2f92a33d47195cc46c3c1f3bda9/trl/trainer/dpo_trainer.py#L766

FabianSchuetze:
> Thank you for the PR, it would solve my problem as well. However, doesn't your PR still leave first bug unfixed?
> 
> > The pre-processing function only ingest tokens worth of one "row" of an image, but the image-encoder is 2-D
> 
> The line responsible for the bug is the same: https://github.com/FabianSchuetze/trl/blob/adb34f339895e2f92a33d47195cc46c3c1f3bda9/trl/trainer/dpo_trainer.py#L766

Yes and No. I deliberately introduced the `skip_prepare_dataset` keyword. When that is set, the dataset is not proccessed and you can use your own collate_fn function. That's exactly the same behavior as in the SFTTrainer.

VietHoang1512:
hi @FabianSchuetze, I am wondering if you could provide an example training script for QwenVL with your DPOTrainer?

GrayChan04:
@FabianSchuetze Thanks for your PR. But it still doesn't work and still throws the error: 

`But for t, h, w in grid_thw:
TypeError: 'NoneType' object is not iterable`

I have already modified three parts of the code in dpo_train.py: 
`if "image_grid_thw" in batch:
output["image_grid_thw"] = torch.cat([batch["image_grid_thw"], batch["image_grid_thw"]], dim=0)`

`if "image_grid_thw" in concatenated_batch:
model_kwargs["image_grid_thw"] = concatenated_batch["image_grid_thw"]`

`if "image_grid_thw" in concatenated_batch:
model_kwargs["image_grid_thw"] = concatenated_batch["image_grid_thw"]`

Do I need to modify other .py files as well? Or could you please provide the code you used to train Qwen-2.5-VL-7B-Instruct with DPOTrainer?

qgallouedec:
thanks for the fix, and sorry for the late reply, we were refactoring DPO in #3906. This issue is now solved, and you can reliably train Qwen2-VL with DPO :)

PR #4264, opened 2025-10-13
[SFT] add support for unified conversion logic for both images and videos

kashif:
# What does this PR do?

 - Unified conversion logic for both images and videos
 - Parses `<image>` and `<video>` placeholder tags

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -346,74 +346,75 @@ class DataCollatorForVisionLanguageModeling(DataCollatorMixin):
         else:
             raise KeyError(f"Unexpected input keys in examples: {list(examples[0].keys())}.")
 
-    def _collate_language_modeling(self, examples: list[dict[str, Any]]) -> dict[str, Any]:
-        # Handle images
-        images = [example.get("images", []) for example in examples]
-        # Transformers requires at least one image in the batch, otherwise it throws an error
-        if all(img_list == [] for img_list in images):
-            images = None
+    def _has_structured_content(self, messages: list[dict]) -> tuple[bool, bool]:
+        """
+        Check if messages contain structured content with images or videos.
 
-        # Handle videos
-        videos = [example.get("videos", []) for example in examples]
-        if all(vid_list == [] for vid_list in videos):
-            videos = None
-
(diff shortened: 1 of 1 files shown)


qgallouedec:
Nice!

I'm a bit hesitant about one point — adding this would introduce a new logic for how we prepare multimodal data, and it would reduce the overall coherence of the codebase.

Currently, for VLMs, we support the following formats:

1. **Unstructured messages with images**
   (the image or images are added before the first user message)

   ```python
   messages = [{"role": "user", "content": "What is it?"},
               {"role": "assistant", "content": "It's a park"}]
   image = <PIL.Image>  # or images = [<PIL.Image>, <PIL.Image>]
   ```

2. **Structured messages**

   ```python
   messages = [{"role": "user", "content": [{"type": "text", "text": "What is it?"}, {"type": "image"}]},
               {"role": "assistant", "content": "It's a park"}]
   image = <PIL.Image>
   ```

Here, the proposal would be to add support for a third format:

3. **Unstructured messages with an inline image tag**

   ```python
   messages = [{"role": "user", "content": "<image> What is it?"},
               {"role": "assistant", "content": "It's a park"}]
   image = <PIL.Image>
   ```

It doesn’t seem to me that this third format is really necessary. I’d simply recommend that users with datasets in this form preprocess them to match one of the two supported formats.

trl/trainer/sft_trainer.py
```diff
@@ -390,26 +475,43 @@ def _collate_prompt_completion(self, examples: list[dict[str, Any]]) -> dict[str
...
+        videos = [example.get("videos", []) for example in examples]
+        if all(vid_list == [] for vid_list in videos):
+            videos = None
+
         if is_conversational(examples[0]):  # conversational case
             for example in examples:
-                prepare_multimodal_messages(example["prompt"] + example["completion"], len(example["images"]))
+                num_images = len(example.get("images", []))
+                num_videos = len(example.get("videos", []))
+                # Only prepare multimodal messages for images; videos use native <video> tags
```
qgallouedec: > videos use native <video> tags

this doesn't make sense to me. For `_collate_language_modeling` we need to add `{"type": "video"}`, but not here?

PR #4278, opened 2025-10-15
Fix CI issue for vlm_gemma_3n model

kaixuanliu:
When we run test case `pytest -rA tests/test_sft_trainer.py::TestSFTTrainer::test_train_vlm_gemma_3n`, it will fail both on CUDA and Intel XPU. Further investigation shows there are 2 reasons:
1. audio tower part will not update related weight during finetune
2. when using bf16 datatype, small changes for model weight will be rounded off
This PR fixes this bug.

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1494,10 +1494,11 @@ class TestSFTTrainer(TrlTestCase):
         # Initialize the trainer
         training_args = SFTConfig(
             output_dir=self.tmp_dir,
+            learning_rate=0.1,
             max_length=None,
             per_device_train_batch_size=1,
             gradient_checkpointing=True,
-            model_init_kwargs={"dtype": "float16"},
+            model_init_kwargs={"dtype": "bfloat16"},
             report_to="none",
         )
         trainer = SFTTrainer(model="google/gemma-3n-E2B-it", args=training_args, train_dataset=dataset)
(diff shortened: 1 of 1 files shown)


yao-matrix:
@kashif , pls help review, thx very much.

tests/test_sft_trainer.py
```diff
@@ -1514,8 +1514,11 @@ def test_train_vlm_gemma_3n(self):
         # Check the params have changed
         for n, param in previous_trainable_params.items():
             new_param = trainer.model.get_parameter(n)
-            if "model.vision_tower" in n:
-                # The vision tower is not updated, not sure why at this point.
+            if "model.vision_tower" in n or "model.audio_tower" in n:
+                # The vision tower and audio tower will not update.
```
albertvillanova: Nit:
```suggestion
                # The vision/audio towers are frozen during training and will not update
```

qgallouedec: Oh is this the reason?

tests/test_sft_trainer.py
```diff
@@ -1497,7 +1497,7 @@ def test_train_vlm_gemma_3n(self):
             max_length=None,
             per_device_train_batch_size=1,
             gradient_checkpointing=True,
-            model_init_kwargs={"dtype": "bfloat16"},
+            model_init_kwargs={"dtype": "float16"},
```
```suggestion
            model_init_kwargs={"dtype": "bfloat16"},
```

tests/test_sft_trainer.py
```diff
@@ -1497,7 +1497,7 @@ def test_train_vlm_gemma_3n(self):
             max_length=None,
```
qgallouedec, line 1498:
```suggestion
            learning_rate=0.1,  # increase lr to ensure updates are not lost due to bf16 rounding
            max_length=None,
```

qgallouedec (commented):
I'd prefer keeping bf16, as it's usually the precision used. Instead we can simply increase precision

PR #4297, opened 2025-10-16
Replace unittest skipTest from transformers with pytest.skip

albertvillanova:
Replace unittest skipTest from transformers (implicitly used by requirement markers) with pytest.skip markers:
- `require_liger_kernel`
- `require_flash_attn`
- `require_torch_accelerator`
- `require_torch_multi_accelerator`
- `require_wandb`

### DIFF
--- a/tests/testing_utils.py
+++ b/tests/testing_utils.py
@@ -85,8 +85,7 @@ require_torch_gpu_if_bnb_not_multi_backend_enabled = pytest.mark.skipif(
 )
 
 
-# Function ported from transformers.testing_utils
-def require_flash_attn():
+def is_flash_attn_available():
     flash_attn_available = is_flash_attn_2_available()
     kernels_available = is_kernels_available()
     try:
@@ -96,7 +95,11 @@ def require_flash_attn():
     except Exception:
         kernels_available = False
 
-    return pytest.mark.skipif(not (kernels_available or flash_attn_available), reason="test requires Flash Attention")
+    return kernels_available or flash_attn_available
+
+
+# Function ported from transformers.testing_utils
+require_flash_attn = pytest.mark.skipif(not is_flash_attn_available(), reason="test requires Flash Attention")
 
 
 class RandomBinaryJudge(BaseBinaryJudge):


tests/testing_utils.py
```diff
@@ -69,6 +85,20 @@ def is_bitsandbytes_multi_backend_available() -> bool:
...
+    flash_attn_available = is_flash_attn_2_available()
+    kernels_available = is_kernels_available()
+    try:
+        from kernels import get_kernel
+
+        get_kernel("kernels-community/flash-attn")
+    except Exception:
+        kernels_available = False
+
+    return pytest.mark.skipif(not (kernels_available or flash_attn_available), reason="test requires Flash Attention")
```
kashif:
```suggestion
# Function ported from transformers.testing_utils
def require_flash_attn(func=None):
    flash_attn_available = is_flash_attn_2_available()
    kernels_available = is_kernels_available()
    try:
        from kernels import get_kernel

        get_kernel("kernels-community/flash-attn")
    except Exception:
        kernels_available = False

    marker = pytest.mark.skipif(not (kernels_available or flash_attn_available), reason="test requires Flash Attention")
    if func is not None:
        return marker(func)
    return marker
```

kashif: other wise we have to change our decorators to `@require_flash_attn()`

albertvillanova: Thanks a lot for your review, @kashif! :hugs: 

Finally, I aligned the implementation with the rest of the "require" pytest-markers, by calling an auxiliary function `is_x_available`.

tests/testing_utils.py
```diff
@@ -69,6 +85,23 @@ def is_bitsandbytes_multi_backend_available() -> bool:
...
+        get_kernel("kernels-community/flash-attn")
+    except Exception:
+        kernels_available = False
+
+    return kernels_available or flash_attn_available
+
+
+# Function ported from transformers.testing_utils
+require_flash_attn = pytest.mark.skipif(not is_flash_attn_available(), reason="test requires Flash Attention")
+
```
qgallouedec, line 103: I don't think this one it 100% correct, because if kernels is available, but not flash-attn, and you try to use `atto_implementation="flash_attention2"`, it will most likely fail.

albertvillanova: I just ported it from transformers. :sweat_smile: 
https://github.com/huggingface/transformers/blob/12a50f294d50e3d0e124511f2b6f43625f73ffce/src/transformers/testing_utils.py#L575-L591
```python
def require_flash_attn(test_case):
    flash_attn_available = is_flash_attn_2_available()
    kernels_available = is_kernels_available()
    try:
        from kernels import get_kernel

        get_kernel("kernels-community/flash-attn")
    except Exception as _:
        kernels_available = False

    return unittest.skipUnless(kernels_available | flash_attn_available, "test requires Flash Attention")(test_case)
```
Do you think I made an error in porting it? @qgallouedec

albertvillanova: The logic on `transformers` side was recently changed by this PR:
- https://github.com/huggingface/transformers/pull/40178

> Modified `require_flash_attn` in `testing_utils.py` to allow tests to run if either FlashAttention2 or the community kernel is available, broadening test coverage and reliability.

qgallouedec: OK I see.
A lot of our tests rely on flash-attn lib; it's probably a good time to drop `flash-attn` and rely only on `kernels`:

- replace all `attn_implementation="flash_attention_2"` -> `attn_implementation="kernels-community/flash-attn"`
- replace `require_flash_attn` by `require_kernels`

@albertvillanova we can do this in a future PR

albertvillanova: I agree! I created an issue for that:
- #4319

PR #4300, opened 2025-10-18
🕵️‍♂️ GRPO: Agent training

qgallouedec:
# What does this PR do?

This PR implements tool calling for GRPO. The API is as follows:

⚠️ requires transformers v5.0.0.dev0

```python
from datasets import Dataset
from trl import GRPOTrainer

def multiply(a: int, b: int) -> int:
    """
    Multiplies two integers.

    Args:
        a: The first integer.
        b: The second integer.

    Returns:
        The product of the two integers.
    """
    return a * b


dataset = Dataset.from_list(
    [
        {"prompt": [{"role": "user", "content": "What is 3 multiplied by 4?"}], "answer": 12},
        {"prompt": [{"role": "user", "content": "Calculate 7 times 8."}], "answer": 56},
        {"prompt": [{"role": "user", "content": "Find the product of 5 and 6."}], "answer": 30},
        {"prompt": [{"role": "user", "content": "What do you get when you multiply 9 by 9?"}], "answer": 81},
        {"prompt": [{"role": "user", "content": "Compute 12 multiplied by 11."}], "answer": 132},
        {"prompt": [{"role": "user", "content": "What is 15 times 14?"}], "answer": 210},
    ]
)

def accuracy(completions, answer, **kwargs):
    rewards = []
    for completion, ans in zip(completions, answer):
        if str(ans) in completion[-1]["content"]:
            rewards.append(1.0)
        else:
            rewards.append(0.0)
    return rewards

trainer = GRPOTrainer(
    model="Qwen/Qwen3-0.6B",
    train_dataset=dataset,
    tools=[multiply],
    reward_funcs=accuracy,
)
trainer.train()
```

This PR contains a few important changes:

### 🚨 Removal of `max_prompt_length`

This PR contains a breaking change: `max_prompt_length` has been removed from GRPO.

Here are the reasons: (tldr: because it’s extremely hard to implement reliably with multi-turn tool calling, likely harmful to training anyway, likely not used in practice, and dropping it simplifies the API while keeping it consistent across LLMs and VLMs.)

1. **Supporting `max_prompt_length` with tool calling is extremely complex.**
   For single-turn generation it works fine, but multi-turn generation introduces a major challenge: the prompt grows after every step. Since the model is called repeatedly with an increasingly long prompt, we would need to recalculate the allowed prompt length dynamically based on how many tokens have already been generated. Implementing this reliably is tricky and adds significant complexity.

2. **Truncating prompts is likely worse than dropping samples altogether.**
   Although I’m not aware of formal studies, intuition suggests that truncation can remove information necessary to solve the task. Training on such incomplete examples can lead to strong biases, whereas simply skipping overly long samples avoids this risk.

3. **It simplifies the API and removes confusing edge cases.**
   Previously, when training VLMs, we had to tell users to disable prompt truncation entirely because Transformers does not support truncating multimodal prompts. This led to inconsistent, non-user-friendly recommendations. Removing `max_prompt_length` allows us to provide one clean, unified API that works for all model types.

4. **It very likely not a widely used feature anyway**

### Online decoding

Before calling the reward function, we need to decode the completion. Previously, this was done here:

https://github.com/huggingface/trl/blob/1a9ff522308331ac32e3bd7076a09f3c1a922c1e/trl/trainer/grpo_trainer.py#L1605-L1617

The issue is that, while this works for single-turn outputs, it does not allow reliable parsing of multi-turn text. See this [internal discussion](https://huggingface.slack.com/archives/C06JKEMK6BZ/p1763175180698879). The workaround is to parse after each turn, which requires moving the decoding logic inside the generation loop (in `_generate`):

https://github.com/huggingface/trl/blob/c54bf4fd48c60c3060fd902686464854f6660c2d/trl/trainer/grpo_trainer.py#L1483-L1495

https://github.com/huggingface/trl/blob/c54bf4fd48c60c3060fd902686464854f6660c2d/trl/trainer/grpo_trainer.py#L1543

https://github.com/huggingface/trl/blob/c54bf4fd48c60c3060fd902686464854f6660c2d/trl/trainer/grpo_trainer.py#L1614-L1618

The method then returns the list of messages:

https://github.com/huggingface/trl/blob/c54bf4fd48c60c3060fd902686464854f6660c2d/trl/trainer/grpo_trainer.py#L1669

Note that this change removes support for the "bootstrap" feature. I haven’t had time to investigate adding support for it.

### Tool mask

We don't want the loss to be computed on the tokens corresponding to the tool result. Consequently, `_generate` builds and return a `tool_mask`

https://github.com/huggingface/trl/blob/c54bf4fd48c60c3060fd902686464854f6660c2d/trl/trainer/grpo_trainer.py#L1668

which is then used to mask these tokens in the loss computation.

https://github.com/huggingface/trl/blob/c54bf4fd48c60c3060fd902686464854f6660c2d/trl/trainer/grpo_trainer.py#L2100

#### Chat template

To make this feature work, we need the chat template to be prefix-preserving. Ie:

https://github.com/huggingface/trl/blob/9f0aa3db1ddf244c9cefad45a984f18a5e71d7da/trl/chat_template_utils.py#L195-L212

The issue is that some widely used tokenizers, such as GPT-OSS and Qwen3, are not prefix-preserving due to the way they handle think tokens. To address this, I suggest using a slightly modified version of the template that ensures it is prefix-preserving. Additionally, as @lewtun pointed out, it’s not even clear whether these templates might make the inference OOD

Screenshot 2025-11-19 at 11 59 14 PM

#### Response schema

To parse tool calls from the model’s response, we rely on `tokenizer.parse_response`, introduced in https://github.com/huggingface/transformers/pull/40894. This requires the tokenizer to have a `response_schema` (integrated in a similar way as chat templates). However, very few (no?) model repositories currently include such a schema.

To enable this feature despite the lack of adoption, I propose adding a mapping for some popular chat templates to their response schemas (currently only Qwen3).

https://github.com/huggingface/trl/blob/fbb625f1904c79549c2cbff07706d341250dc143/trl/chat_template_utils.py#L172-L174

 Ideally, once adoption increases and model repos start including proper response schemas, we can remove this custom mapping entirely.

### A fair amount of complexity in the generation

This PR adds 60+ lines of intricate code with many special cases in the generation method. While it’s admittedly hard to follow, after a lot of iteration this is likely the simplest reliable way to implement the feature. Normally, I would be very reluctant to introduce this level of complexity, but given the impact of this feature, I believe it’s truly worth it.

https://github.com/huggingface/trl/blob/c54bf4fd48c60c3060fd902686464854f6660c2d/trl/trainer/grpo_trainer.py#L1509-L1623

## Next steps

* Add tool support to the vLLM client/server (currently supported only for collocated vLLM and Transformers generation)
* Provide the same integration for RLOO

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1854,7 +1854,8 @@ class GRPOTrainer(BaseTrainer):
 
             # Compute the importance sampling ratio when using vLLM, to correct for potential distribution mismatch
             if self.use_vllm and self.vllm_importance_sampling_correction:
-                per_token_logps_diff = (old_per_token_logps - sampling_per_token_logps) * completion_mask
+                mask = completion_mask if not self.tools else completion_mask * tool_mask
+                per_token_logps_diff = (old_per_token_logps - sampling_per_token_logps) * mask
 
                 sequence_level_is = self.vllm_importance_sampling_mode in ["sequence_mask", "sequence_truncate"]
                 if sequence_level_is:
@@ -2132,7 +2133,8 @@ class GRPOTrainer(BaseTrainer):
         )
 
         if self.top_entropy_quantile < 1.0:
-            entropy_mask = self.get_high_entropy_mask(entropies, completion_mask, 1 - self.top_entropy_quantile)
+            mask = completion_mask if not self.tools else completion_mask * inputs["tool_mask"]
+            entropy_mask = self.get_high_entropy_mask(entropies, mask, 1 - self.top_entropy_quantile)
         else:
             entropy_mask = None
 
@@ -2161,7 +2163,8 @@ class GRPOTrainer(BaseTrainer):
         if self.importance_sampling_level == "token":
             log_importance_weights = log_ratio
         elif self.importance_sampling_level == "sequence":
-            log_importance_weights = (log_ratio * completion_mask).sum(-1) / completion_mask.sum(-1).clamp(min=1.0)
+            mask = completion_mask if not self.tools else completion_mask * inputs["tool_mask"]
+            log_importance_weights = (log_ratio * mask).sum(-1) / mask.sum(-1).clamp(min=1.0)
             log_importance_weights = log_importance_weights.unsqueeze(-1)
         else:
             raise ValueError(


docs/source/grpo_trainer.md
```diff
@@ -514,6 +514,10 @@ and the reward will be computed as the sum of the rewards from each function, or
 
 Note that [`GRPOTrainer`] supports multiple reward functions of different types. See the parameters documentation for more details.
 
+## Agent Training
+
+To write...
```
edbeeching: Would you like some help here? We could also have a novel example for the release?

burtenshaw: @sergiopaniego is working on this I believe.

pyproject.toml
sergiopaniego, line 1: The `transformers` dependency could be upgraded to `>=5.0.0` when available since it's required by the PR

qgallouedec: The problem is that it's not yet released. It has to be installed from source for the moment.

qgallouedec:
For information:
| Tokenizer | Tool Calls | Prefix-Preserving |
| --- | --- | --- |
| trl-internal-testing/tiny-CohereForCausalLM | ❌ | ✅ |
| trl-internal-testing/tiny-DbrxForCausalLM | ❌ | ✅ |
| trl-internal-testing/tiny-DeepseekV3ForCausalLM | ✅ | ❌ |
| trl-internal-testing/tiny-DeepseekV3ForCausalLM-0528 | ✅ | ✅ |
| trl-internal-testing/tiny-FalconMambaForCausalLM | ❌ | ✅ |
| trl-internal-testing/tiny-Gemma2ForCausalLM | ❌ | ✅ |
| trl-internal-testing/tiny-GemmaForCausalLM | ❌ | ✅ |
| trl-internal-testing/tiny-GptOssForCausalLM | ✅ | ❌ |
| trl-internal-testing/tiny-LlamaForCausalLM-3.1 | ✅ | ✅ |
| trl-internal-testing/tiny-LlamaForCausalLM-3.2 | ✅ | ✅ |
| trl-internal-testing/tiny-LlamaForCausalLM-3 | ❌ | ✅ |
| trl-internal-testing/tiny-MistralForCausalLM-0.1 | ❌ | ✅ |
| trl-internal-testing/tiny-MistralForCausalLM-0.2 | ❌ | ✅ |
| trl-internal-testing/tiny-Phi3ForCausalLM | ❌ | ❌ |
| trl-internal-testing/tiny-Qwen2ForCausalLM-2.5 | ✅ | ✅ |
| trl-internal-testing/tiny-Qwen3ForCausalLM | ✅ | ❌ |
```python
from trl.chat_template_utils import is_chat_template_prefix_preserving

tokenizers = [
    "trl-internal-testing/tiny-CohereForCausalLM",
    "trl-internal-testing/tiny-DbrxForCausalLM",
    "trl-internal-testing/tiny-DeepseekV3ForCausalLM",
    "trl-internal-testing/tiny-DeepseekV3ForCausalLM-0528",
    "trl-internal-testing/tiny-FalconMambaForCausalLM",
    "trl-internal-testing/tiny-Gemma2ForCausalLM",
    "trl-internal-testing/tiny-GemmaForCausalLM",
    "trl-internal-testing/tiny-GptOssForCausalLM",
    "trl-internal-testing/tiny-LlamaForCausalLM-3.1",
    "trl-internal-testing/tiny-LlamaForCausalLM-3.2",
    "trl-internal-testing/tiny-LlamaForCausalLM-3",
    "trl-internal-testing/tiny-MistralForCausalLM-0.1",
    "trl-internal-testing/tiny-MistralForCausalLM-0.2",
    "trl-internal-testing/tiny-Phi3ForCausalLM",
    "trl-internal-testing/tiny-Qwen3ForCausalLM",
]

print(f"| Tokenizer | Tool Calls | Prefix-Preserving |")
print("| --- | --- | --- |")
for tokenizer_name in tokenizers:
    tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
    tool_support = "✅" if "tool_call" in tokenizer.chat_template else "❌"
    prefix_preserving = "✅" if is_chat_template_prefix_preserving(tokenizer) else "❌"
    print(f"| {tokenizer_name} | {tool_support} | {prefix_preserving} |")
```

trl/experimental/grpo_with_replay_buffer/grpo_with_replay_buffer_trainer.py
```diff
@@ -106,6 +112,9 @@ def _generate_and_score_completions(
             sampling_per_token_logps = pad(sampling_per_token_logps, padding_value=0.0, padding_side="right")
         else:
             sampling_per_token_logps = None
+        if self.tools:
+            tool_mask = [torch.tensor(mask, device=device) for mask in tool_mask_list]
+            tool_mask = pad(tool_mask, padding_value=0, padding_side="right")
```
edbeeching: It might be easier to follow if the `(1 - tool_mask)` from L289 is done here? I was initialy confused about why the mask was padded with zeroes until I saw it was flipped later.

trl/trainer/grpo_trainer.py
```diff
@@ -1332,22 +1382,18 @@ def _generate_single_turn(self, prompts: list):
                         self.llm.sleep(level=2)
 
         elif self.use_transformers_paged:
-            processor_kwargs = {
-                "max_length": self.max_prompt_length,
-                "truncation": True,
-                "add_special_tokens": False,
```
edbeeching, line 1372: It was your intention to also drop `"add_special_tokens": False,` ?

trl/trainer/grpo_trainer.py
```diff
@@ -1431,8 +1475,155 @@ def _generate(self, prompts: list):
...
+                completions = [[parse_response(self.processing_class, ids)] for ids in completion_ids]
+            else:
+                contents = self.processing_class.batch_decode(completion_ids, skip_special_tokens=True)
+                completions = [[{"role": "assistant", "content": content}] for content in contents]
+        else:
+            completions = self.processing_class.batch_decode(completion_ids, skip_special_tokens=True)
+
+        # Extract tool calls from the completions
+        if self.tools:
+            tool_calls = [completion[0].get("tool_calls") for completion in completions]
```
edbeeching: If it is not too complicated then L1510-1625 should go in their own method called `_run_tool_calls` or similar

This block can then be:
```
        if self.tools:
            tool_mask, completions, ... = self._run_tool_calls(...)
        else:
            tool_mask = None

```

It would make it simpler to parse the codebase when you are not using tools. It may make the tool call block more testable? There is a lot of stuff going on here, overall I understand the logic but there may be edge cases that slip though. Is there are way we can break up while loop into smaller methods and unit test them?

trl/trainer/grpo_trainer.py
```diff
@@ -1510,6 +1725,9 @@ def _generate_and_score_completions(
             sampling_per_token_logps = pad(sampling_per_token_logps, padding_value=0.0, padding_side="right")
         else:
             sampling_per_token_logps = None
+        if self.tools:
+            tool_mask = [torch.tensor(mask, device=device) for mask in tool_mask_list]
+            tool_mask = pad(tool_mask, padding_value=0, padding_side="right")
```
edbeeching: Same comment as above, would it make sense to do `(1 - tool_mask)` directly here? Alternatively the mask could be not flipped when it is created, then you could pad with 1's here.

trl/trainer/grpo_trainer.py
edbeeching, line 1696: `agg_completion_lengths.sum()  # = num_items_in_batch`
Does / should this include the tool call tokens?

edbeeching (approved):
Thanks for this implementation. In general it is great, the only issue I have is that there is a lot of new code in L1510-1625 of the `grpo_trainer.py` file. It would be great to extract this into at least one method, or several methods for some of the blocks of code:

For example make a `_run_tool_calls` method which contains:
```
...
while idxs_with_tool:
  ... 
  # Call the tools, and build the new prompt for generation
  ... = call_the_tools(..)
  # Tokenize and filter samples whose length exceeds max allowed length.
  ... = tokenize_and_filter(...)
  # Generate new completions after tool execution
  prompt_completion_tool_ids, post_tool_ids, post_tool_logprobs, _ = self._generate_single_turn(prompt_completion_tools)
# Sanity check: from experience
... etc

```
Then add tests for those methods, there seem to be a lot of edge cases you have already covered, capturing those in tests would be good and will make this more robust.

The level of granularity you want with these methods is up to you of course. But at least one method for  `_run_tool_calls(...)` would make the `_generate` method easier to parse, particularly for new users who are trying to underand the codebase and are not training llms to use tools (yet!).

trl/trainer/grpo_trainer.py
```diff
@@ -589,11 +628,6 @@ def cast_outputs_to_original_dtype(module, args, output):
                 # Ensure distributed rendezvous variables are set without colliding across concurrent runs
                 ensure_master_addr_port()
 
-                if self.max_prompt_length is not None and self.max_completion_length is not None:
```
lewtun, line 611: Does this mean we now use the model's default `max_model_len`?

If so, would it make sense to expose `max_model_len` as an arg in the config to accommodate cases where the sequences one is training on are much smaller than the default value? This allows one to get better throughput / lower memory usage.

trl/trainer/grpo_trainer.py
```diff
@@ -1431,8 +1475,155 @@ def _generate(self, prompts: list):
...
+                        name = tool_call.get("name", "unknown")
+                        result = {"error": f"Unsupported tool call type: {tool_call['type']}"}
+                    tool_message = {"role": "tool", "name": name, "content": str(result)}
+                    prompt_completion_tool.append(tool_message)
+                    completions[idx_with_tool].append(tool_message)
+
+            # Tokenize and filter samples whose length exceeds max allowed length. This is important, because if vLLM
+            # is called with an input longer than its max model length, it will error out.
+            pct_ids = self.processing_class.apply_chat_template(prompt_completion_tools, **kwargs)["input_ids"]
+            overlong = [len(pct) - len(p) >= self.max_completion_length for p, pct in zip(p_ids, pct_ids, strict=True)]
```
lewtun: Should this refer instead to `max_model_len` because technically one could have a small `max_completion_length` but a long context that is built up over many steps of turn calling.

trl/trainer/grpo_trainer.py
```diff
@@ -1431,8 +1475,155 @@ def _generate(self, prompts: list):
...
+            # Generate new completions after tool execution
+            prompt_completion_tool_ids, post_tool_ids, post_tool_logprobs, _ = self._generate_single_turn(
+                prompt_completion_tools
+            )
+
+            # Sanity check: from experience, this is useful to catch bugs in the chat template
+            for idx in range(len(idxs_with_tool)):
+                idx_with_tool = idxs_with_tool[idx]
+                pct = prompt_completion_tool_ids[idx]  # = prompt-completion-tool
+                assert prompt_ids[idx_with_tool] == pct[: len(prompt_ids[idx_with_tool])]
```
lewtun: Should we raise a proper `ValueError` here with an informative error message for the user?

trl/trainer/grpo_trainer.py
```diff
@@ -1431,8 +1475,155 @@ def _generate(self, prompts: list):
...
+
+            # Truncate so that pct[len(prompt_ids[idx]) :] + post_tool does not exceed max_completion_length
```
lewtun, line 1597: Same comment here, that I don't think the issue is about exceeding the `max_completion_length`, rather that we should left-truncate to ensure we have enough tokens to avoid exceeding the model's context length (set via `max_model_len`)

trl/trainer/grpo_trainer.py
lewtun: I agree with Ed that it would be great if we can refactor this somewhat to isolate the tool-calling logic and enable it to be tested independently

trl/trainer/grpo_trainer.py
```diff
@@ -1463,7 +1654,25 @@ def _generate(self, prompts: list):
         self._metrics[mode]["completions/min_terminated_length"].append(term_completion_lengths.float().min().item())
         self._metrics[mode]["completions/max_terminated_length"].append(term_completion_lengths.float().max().item())
 
-        return prompt_ids, completion_ids, total_completion_tokens, logprobs, extra_fields
+        if self.tools:
+            agg_tool_call_count = self.accelerator.gather(torch.tensor(tool_call_count, device=device)).sum()
```
lewtun, line 1721: Very nice metrics!

trl/chat_template_utils.py
```diff
@@ -0,0 +1,404 @@
...
+    >>> tokenizer.parse_response(assistant_text)
+    {'role': 'assistant', 'content': '', 'tool_calls': [{'type': 'function', 'function': {'name': 'multiply', 'arguments': {'a': 3, 'b': 4}}}]}
+    ```
+    """
+    if tokenizer.chat_template == qwen3_chat_template:
+        tokenizer.response_schema = qwen3_schema
+        return tokenizer
+    raise ValueError(
+        "Unrecognized chat template, failed to add response schema. Please manually set the response schema on the "
+        "tokenizer or processor."
```
lewtun: To help users, shall we add a pointer to the docs?
```suggestion
        "tokenizer or processor. See the Transformers [docs](https://huggingface.co/docs/transformers/main/en/chat_response_parsing#response-parsing) for more details on response parsing."
```

trl/trainer/grpo_trainer.py
kashif: can this be made independent of GRPO trainer as other online trainers could then potentially benifit from this helper?

trl/experimental/grpo_with_replay_buffer/grpo_with_replay_buffer_trainer.py
qgallouedec: it seems reasonable to me, done in caf1ad259f11a6d0571ce53f742b49ac002f0e4e

trl/trainer/grpo_trainer.py
qgallouedec: Yes:

- in the conversational case, `add_special_tokens` is (properly) hardcoded to `False` anyway, see [here](https://github.com/huggingface/transformers/blob/69f003696b55de75b7f18888c03111909a7cd537/src/transformers/tokenization_utils_base.py#L1714)
- in the standard case, `add_special_tokens` should be set to `True` (default, see [here](https://github.com/huggingface/transformers/blob/69f003696b55de75b7f18888c03111909a7cd537/src/transformers/tokenization_utils_base.py#L2980)) so that any BOS token is added.

I don't know why we had `"add_special_tokens": False,` in the first place

trl/trainer/grpo_trainer.py
qgallouedec: I refactored to isolate the `_tool_call_loop`. 94c2ff21e183068bf81e7036402809d5c2c3c8ee. Now we have, as @edbeeching suggested:

```python
        if self.tools:
            tool_mask, ... = self._tool_call_loop(prompts, prompt_ids, completion_ids, completions, logprobs)
        else:
            tool_mask = None
```

much easier to read indeed.

---

> It may make the tool call block more testable? There is a lot of stuff going on here

>  enable it to be tested independently

I get the motivation, and I agree that "there is a lot of stuff going on here." My take is that it would be better to split this into smaller helper functions or utilities. That makes the code easier to follow, easier to test, and more flexible overall. I'll try to do it.

However, I usually avoid testing private methods or internal logic directly, because those tests can break easily for the wrong reasons: any small change inside can force a test rewrite. I prefer testing through the public API, even if it means patching a bit to make sure the right part runs. Like here

https://github.com/huggingface/trl/blob/94c2ff21e183068bf81e7036402809d5c2c3c8ee/tests/test_grpo_trainer.py#L1721-L1797

It is more difficult than it seems to find the right balance. 😅 I'd be curious to know what @albertvillanova thinks about it (when he is back).

---
> can this be made independent of GRPO trainer as other online trainers could then potentially benefit from this helper?

It’s quite complex because a lot of the internal logic of this new method depends on other methods and attributes (`self.chat_template_kwargs`, `self._tool_dict`, `self.processing_class`, `self._generate_single_turn`, ...). In this case, I would lean more towards copying. From experience, even though this increases the amount of duplicated code in the codebase, it doesn’t really make maintenance harder, in fact, it often makes maintenance much easier. For example, RLOO and GRPO have a lot of duplicated code, but they are very easy to maintain together with very little shared abstraction.

trl/trainer/grpo_trainer.py
qgallouedec: It does include the tool call tokens, but shouldn't. I'll need to fix it. Excellent catch, thanks!

sergiopaniego:
Added agent training example script ([trackio](https://huggingface.co/spaces/sergiopaniego/trl-trackio))!
400 steps to reach that point
Screenshot 2025-11-25 at 17 10 58

qgallouedec:
@codex review

qgallouedec:
@codex review

trl/trainer/grpo_trainer.py
```diff
@@ -375,8 +389,33 @@ def __init__(
...
+                raise ImportError(
+                    "Using tools with GRPOTrainer requires the jmespath library for response parsing. Please install "
+                    "it with `pip install jmespath` to use this feature."
+                )
+        self.tools = tools or []
+        self._tool_dict = {tool.__name__: tool for tool in self.tools}
+        # At the time of initial implementation, most tokenizers do not have built-in support for response schemas.
+        # While waiting for broader adoption, we provide this utility function to manually set the response schema for
+        # known chat templates.
+        if tools and not processing_class.response_schema:
```
albertvillanova: To fix #4609:
```suggestion
        if tools and not getattr(processing_class, "response_schema", None):
```

albertvillanova (commented):
Just a minor change to fix #4609.

After the merge of https://github.com/huggingface/transformers/pull/40936, the attribute does not necessarily exist. Before it was `None`.

Feel free to ignore it if there is a better solution!

examples/scripts/grpo_agent.py
```diff
@@ -63,45 +63,63 @@
...
+    """
+    Reward Yes/No correctness.
+    Robust to extra spaces, punctuation, or surrounding stars.
+    """
     rewards = []
     for completion, ans in zip(completions, answer, strict=False):
-        guess = completion[-1]["content"].strip()
+        guess = completion[-1]["content"].strip().lower()
+        guess_clean = guess.replace("*", "").replace("`", "").strip()
         reward = 0.0
```
edbeeching, line 67: Could L75-82 be simplified to:
```

if guess_clean == ans.lower():
    reward = 0.5
else:
    reward = -0.2
```

examples/scripts/grpo_agent.py
```diff
@@ -63,45 +63,63 @@
...
-        elif reward == 0.0:
-            reward += 0.25
+            if turn.get("role") == "assistant" and turn.get("tool_calls"):
+                tool_called = True
+            if turn.get("role") == "tool" and turn.get("content"):
+                tool_response_ok = True
+                if "error" in turn["content"].lower():
+                    reward -= 0.3  # penalize errors
+
+        if tool_called and tool_response_ok:
```
edbeeching: 107-112 would be easier to parse for a reader like this:
```
if tool_called:
    if tool_response_ok:
        reward += 0.25
    else:
        reward -= 0.2
else:
    reward -= 0.3
```

trl/trainer/grpo_trainer.py
qgallouedec: good point

qgallouedec: bd8905de486577426d7f006e2063e18bf942b783

trl/trainer/grpo_trainer.py
qgallouedec: the problem is that, with left-truncation, the logprobs we get after a left-truncate will reflect the model’s probabilities given only the truncated context we actually sent, not the full original history.

Before truncate, tokens are scored as $(\log p(x_t \mid \text{full history}))$.
After truncate, tokens are scored as $(\log p(x_t \mid \text{truncated history}))$.

So I think it would introduce a bias when we compute the loss and the importance sampling (for training-inference mismatch)

trl/trainer/grpo_trainer.py
qgallouedec: Ah yes very good catch, I double-checked and you're right!
Fixed here https://github.com/huggingface/trl/pull/4300/commits/c7ba06db43f872f8608367df90bfb57a3bc4e344

trl/trainer/grpo_trainer.py
```diff
@@ -1461,15 +1508,183 @@ def _generate_single_turn(self, prompts: list):
...
+            # Generate new completions after tool execution
+            prompt_completion_tool_ids, post_tool_ids, post_tool_logprobs, _ = self._generate_single_turn(
+                prompt_completion_tools
+            )
+
+            # Sanity check: from experience, this is useful to catch bugs in the chat template
+            for idx in range(len(idxs_with_tool)):
+                idx_with_tool = idxs_with_tool[idx]
+                pct = prompt_completion_tool_ids[idx]  # = prompt-completion-tool
+                assert prompt_ids[idx_with_tool] == pct[: len(prompt_ids[idx_with_tool])]
```
```suggestion
                if prompt_ids[idx_with_tool] != pct[: len(prompt_ids[idx_with_tool])]:

                    raise ValueError("The chat template has been detected as not being a prefix-preserving template. Please correct the chat template.")
```

trl/trainer/grpo_trainer.py
qgallouedec: Yes, thanks, done

austinmw:
Would this be plug-and-play with smolagents for inference?

Also curious about extensibility to langgraph or strands agent harness.

PR #4302, opened 2025-10-18
[SFT] Log mean token accuracy from Liger kernel

kashif:
# What does this PR do?

We log the mean token accuracy from Liger kernel if it is available

Requires https://github.com/linkedin/Liger-Kernel/pull/910

to test do:
```bash
pip install git+https://github.com/kashif/Liger-Kernel.git@mean_token_accuracy
pip install git+https://github.com/huggingface/trl.git@iiger-mean-acc
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1518,7 +1518,6 @@ class TestGRPOTrainer(TrlTestCase):
             num_generations=3,  # reduce the number of generations to reduce memory usage
             max_completion_length=8,  # reduce the completion length to reduce memory usage
             use_liger_kernel=True,  # enable Liger kernel
-            loss_type="bnpo",  # default dapo is not supported yet
             report_to="none",
         )
         trainer = GRPOTrainer(


trl/trainer/sft_trainer.py
```diff
@@ -1133,8 +1138,12 @@ def compute_loss(
             self._total_train_tokens += num_tokens_in_batch
         self._metrics[mode]["num_tokens"] = [self._total_train_tokens]
 
-        # Compute token accuracy if we have labels and if the model is not using Liger (no logits)
-        if not self.args.use_liger_kernel:
+        if self.args.use_liger_kernel:
+            if hasattr(outputs, "token_accuracy") and outputs.token_accuracy is not None:
```
qgallouedec: I prefer using an explicit condition, like this:

```python
if self.args.use_liger_kernel and version("liger_kernel") >= "0.6.4":
    token_accuracy = self.accelerator.gather_for_metrics(outputs.token_accuracy).mean().item()
    self._metrics[mode]["mean_token_accuracy"].append(token_accuracy)
```

Having the explicit condition makes it much easier to spot when the check can be removed after we bump the minimum `liger_kernel` version. It’s clearer and more maintainable than hiding the logic elsewhere.

qgallouedec: the question is then, which version will include this feature

kashif: the issue with this is that I cannot test it since the dev version is the current version in their setup

qgallouedec: ah true

lewtun:
Tested with both DDP and FSDP2 - accuracies and loss are ~identical both with and without Liger. 

I suggest we wait until the Liger PR is merged before merging this one with @qgallouedec's change:

Screenshot 2025-11-04 at 20 08 58
Screenshot 2025-11-04 at 20 09 02

kashif:
Liger PR merged... waiting for a release, and then can pin the version to that

trl/trainer/sft_trainer.py
kashif: checking explicitly now

trl/trainer/sft_trainer.py
```diff
@@ -1151,8 +1158,12 @@ def compute_loss(
...
```
qgallouedec: Can we have an explicit version check instead of hasattr?

qgallouedec: Also, if you have the right version, there is no reason for token_accuracy to be None, is there?

qgallouedec (approved):
Lgtm, just a small remark

trl/trainer/sft_trainer.py
kashif: there could be, as the outputs depend on how the model is patched... a model might not be patched properly, and then we get an exception for that model. I am fixing models as i find them but not all might have this.

PR #4303, opened 2025-10-18
GRPO: ScaleRL -> Support casting LM Head to FP32

pramodith:
# What does this PR do?

Per [ScaleRL](https://huggingface.co/papers/2510.13786)

> "FP32 Precision for LLM logits The generators and trainers rely on different kernels for inference and training, leading to small numerical mismatches in their token probabilities (He & Lab, 2025). RL training is highly sensitive to such discrepancies, since they directly affect the IS ratio in the surrogate objective. MiniMax et al. (2025) identified that these mismatches are especially pronounced at the language model head, and mitigate this by FP32 computations at the head for both the generator and trainer. As shown in Figure 5b, the precision fix dramatically improves the asymptotic performance A from 0.52 to 0.61. Given this clear benefit, we include the FP32 precision fix in our ScaleRL recipe."

This paper adds the option to cast the LM head to full-precision.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


pramodith:
I wasn't able to find a config in `vllm` that forces the lm head to fp32. Likewise for `liger_loss`, lemme know if I should just raise `Unsupported` exceptions` for when `cast_lm_head_to_fp32` and `use_vllm` or `use_liger_loss` are enabled.

qgallouedec:
I would approach this a bit differently: if the user passes a string as the model ID, it means they’re relying on trl to choose the best way to initialize the model. In that case, we can safely cast the `lm_head` to `float32`:

```python
GRPOTrainer(
    model="my_model",  # lm_head automatically cast to float32
    ...
)
```

However, if the user wants more fine-grained control over how the model is initialized, they should handle the initialization themselves, for example:

```python

model = AutoModelForCausalLM.from_pretrained("my_model")

GRPOTrainer(
    model=model,
    ...
)
```

consequently, we wouldn't need a new parameter in in `GRPOConfig`. 
- Note 1: that it's not very consistent with `disable_dropout`, but I think it's ok. 
- Note 2: it will need to be clearly documented in `GRPOTrainer.model`

What do you think?

qgallouedec:
And also, do you mind adding a section in paper index as well. Something like "currently, only the lm head to fp32 trick is supported and enabled by default"

pramodith:
> Note 1: that it's not very consistent with `disable_dropout`, but I think it's ok.
> Note 2: it will need to be clearly documented in GRPOTrainer.model

I did consider the approach that you're proposing too and I do see its merit. My concern with the lib automatically setting the LM Head to fp32 without the user explicitly stating it is that the memory consumed by the LM Head and subsequently logits, entropy etc. is going to be a fair amount which could lead to OOM issues.

The user would then need to spend time understanding why there's an OOM. If we document things really well like throwing in a log message stating `LM Head is initialized as FP32, if you run into OOM errors consider down-casting this by ...` they might be able to identify the reason for OOM quickly but would then have to go on to add the 2 lines of code that you pointed whereas for a config it'd just be one line. They'd also be much more aware of the cause for OOM errors because they explicitly set the config to run the LM Head in fp32.

pramodith:
> And also, do you mind adding a section in paper index as well. Something like "currently, only the lm head to fp32 trick is supported and enabled by default"

I'm going to be adding some of the other parts of ScaleRL too, and was planning on adding ScaleRL to the paper-index once we had all of those along with the CISPO pr merged.

trl/trainer/grpo_config.py
```diff
@@ -40,9 +40,11 @@ class GRPOConfig(TrainingArguments):
         disable_dropout (`bool`, *optional*, defaults to `False`):
             Whether to disable dropout in the model. This is useful for training with a reference model, as it prevents
             the model from generating different logprobs for the same input.
+        cast_lm_head_to_fp32 (`bool`, *optional*, defaults to `False`):
+            Whether to cast the Language Modeling Head of the policy and reference models to float32. As recommended by
```
lewtun: turbo nit:
```suggestion
            Whether to cast the language modeling head of the policy and reference models to float32. As recommended by
```

trl/trainer/grpo_config.py
```diff
@@ -286,6 +288,13 @@ class GRPOConfig(TrainingArguments):
             "it prevents the model from generating different logprobs for the same input."
         },
     )
+    cast_lm_head_to_fp32: bool = field(
+        default=False,
+        metadata={
+            "help": "Whether to cast the Language Modeling Head of the policy and reference, models to float32."
```
lewtun: nit:
```suggestion
            "help": "Whether to cast the language modeling head of the policy and reference models to float32."
```

lewtun (approved):
Very clean PR @pramodith ! Overall LGTM. If time / compute permits, it could be quite interesting to see what training / inference probabilities look like with/without FP32, as done in the MiniMax M1 paper: https://arxiv.org/abs/2506.13585

Image

trl/trainer/grpo_config.py
```diff
@@ -286,6 +288,13 @@ class GRPOConfig(TrainingArguments):
             "it prevents the model from generating different logprobs for the same input."
         },
     )
+    cast_lm_head_to_fp32: bool = field(
+        default=False,
```
qgallouedec, line 304: When/if we confirm the results from the paper, this could be changed to True in my opinion.

tests/test_grpo_trainer.py
```diff
@@ -661,7 +661,8 @@ def test_training_with_sync_ref_model(self):
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
 
-    def test_training_beta_non_zero(self):
+    @parameterized.expand([(False,), (True,)])
+    def test_training_beta_non_zero(self, cast_lm_head_to_fp32):
```
qgallouedec: Can we have a separate test for "cast lm head to fp32" instead? It's not really related to beta

pramodith: Done!

trl/trainer/grpo_config.py
pramodith: Lemme take a look at the paper that Lewis linked, and get back to you on if I can do something like that with the limited gpu resources I have.

pramodith: Seems fairly easy to implement! All I have to do is take a dataset with a thousand prompts. load the model in bf16 mode, cast its lm_head to fp32 and store the logits of the tokens in the completions and then do the same for a model loaded in vllm.

I can then plot something similar to the Minimax paper. Will have this done in the next few hours.

tests/test_grpo_trainer.py
```diff
@@ -690,6 +690,36 @@ def test_training_beta_non_zero(self):
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
 
+    def test_training_with_cast_lm_head_to_fp32(self):
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+        training_args = GRPOConfig(
+            fp16=True,
```
```suggestion
```

for consistency with other tests. Unless there is something specific with fp16?

tests/test_grpo_trainer.py
```diff
@@ -690,6 +690,36 @@ def test_training_beta_non_zero(self):
...
+            model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
+            reward_funcs="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
+            args=training_args,
+            train_dataset=dataset,
+        )
+
+        previous_trainable_params = {n: param.clone() for n, param in trainer.model.named_parameters()}
+
+        trainer.train()
+
```
qgallouedec, line 726: maybe here it would make sense to assert that `trainer.model.lm_head` is indeed in fp32

trl/trainer/grpo_config.py
qgallouedec: wow so cool, if you have any trouble reproducing, I think it's perfectly fine to open a subsequent PR to change the default

tests/test_grpo_trainer.py
pramodith: Good call.

trl/trainer/grpo_config.py
pramodith: Realized that there were a few edge cases, going to try to wrap it up as soon as possible. If I run into issues reproducing in the next couple of days I'll go ahead and merge this.

pramodith: Trying to reproduce the paper has resulted in finding a neat edge case. So a number of models tie the weights of the `LM` head with the `token embedding` layer, in such a scenario casting the `lm_head` also casts the token_embedding layer to `float32` and this breaks the forward pass for subsequent layers.

In this scenario we'll have to `clone` the `token_embedding` layer and then assign the clone to the LM head as a new layer and then cast that to `float()`. This'll lead to a further increase in GPU memory consumed.

@qgallouedec , @lewtun do you see any other alternative solution to the tied weights issue?

qgallouedec: Interesting! Do you have 1/2 models as example?

pramodith: Any of the Qwen-3 (0.6B, 4B, 8B) or Qwen-2.5 (7.5B) models. It seems like the smaller models have `tie_word_embeddings` set to True and the larger models (like 32B) do not.

pramodith: Here are some results from Qwen3-32B (had similar results for the smaller models 4B, 8B, 14B). We see marginal improvement in Correlation and Summed absolute differences in probs

qwen32_bf16
qwen32_fp32

pramodith: Qwen3-30B-A3-Thinking similar results.

qwen_moe_bf16
qwen_moe_fp32

pramodith: @qgallouedec I'm gpu resource constrained so if you want to run it on larger MoE models this is the [code](https://github.com/pramodith/llm_exploration/blob/main/transformers_vs_vllm_logits.py) you might need to tinker with the `tensor_parallel_size` and `enable_expert_parallel`.

trl/trainer/grpo_trainer.py
```diff
@@ -1249,6 +1265,8 @@ def _generate_single_turn(self, prompts: list[str], images: Optional[list]):
                     unwrapped_model.to(torch.bfloat16)
                 elif self.args.fp16:
                     unwrapped_model.to(torch.float16)
+                if self.args.cast_lm_head_to_fp32:
+                    unwrapped_model.lm_head.to(torch.float32)
```
pramodith, line 1321: I don't think I need to register the pre-hook here again, let me know if I'm wrong.

pramodith:
@qgallouedec please take another look at this PR and if all's good feel free to merge it.

PR #4305, opened 2025-10-20
refactor: simplify parameter freezing in modeling_base.py

Ki-Seki:
# What does this PR do?

Refactor the parameter freezing logic to simplify the handling of shared and unshared parameters. The original implementation was confusing; this version is clearer. Additionally, the current shared approach merely performs a deep copy—consider adding a TODO to switch to a shallow copy for shared layers in the future to save GPU memory.

## Before submitting
      to it if that's the case.

## Who can review?

@lvwerra @lewtun @albertvillanova  @qgallouedec

### DIFF
--- a/trl/models/modeling_base.py
+++ b/trl/models/modeling_base.py
@@ -622,12 +622,14 @@ def create_reference_model(
     parameter_names = [n for n, _ in model.named_parameters()]
     ref_model = deepcopy(model)
 
+    for param_name in parameter_names:
+        param = ref_model.get_parameter(param_name)
+        param.requires_grad = False
+    ref_model = ref_model.eval()
+
     # if no layers are shared, return copy of model
     if num_shared_layers is None:
-        for param_name in parameter_names:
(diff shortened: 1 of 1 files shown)


Ki-Seki:
@qgallouedec , thank you for running checks. I found that two CI checks failed, which may be some issues with the main branch. This PR is just a simple refactor.🤗

qgallouedec:
Thanks @Ki-Seki, and sorry for the long delay here. Since this PR was opened, `create_reference_model` was moved out of `trl/models/modeling_base.py` (that file no longer exists); the function now lives in `trl/experimental/utils.py`, so this diff no longer applies. More importantly, the memory-saving idea you flagged as a TODO (sharing the frozen layers' storage with the reference model instead of duplicating them via deepcopy) has since been implemented in #6053. What remains is a readability-only refactor that also subtly changes which reference-model params get frozen, with no test to pin the behavior. Given all that I'll close this one, but thank you! your suggestion pointed in the same direction as #6053.

PR #3469, opened 2025-05-19
Add async tool-enabled vLLM server for GRPO training via OpenAI-compatible interface

BjarniHaukur:
# What does this PR do?

This PR adds a new `vllm_serve_async.py` script to TRL. It:

* Enables asynchronous, OpenAI-compatible inference with vLLM
* Supports models that use tool calls (e.g., search APIs, python tool, general terminal usage)
* Mirrors the weight syncing logic from `vllm_serve.py`
* Delegates endpoint complexity to `vllm.entrypoints.openai.api_server`
* Exposes a `rollout_func` interface that lets users define custom input/output structures and tool definitions to forward into reward functions

Fixes #3284 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #4310, opened 2025-10-20
🕹️ Add rollout function for OpenEnv integration

lewtun:
# What does this PR do?

Simplified version of https://github.com/huggingface/trl/pull/3469 which exposes a `rollout_func` to enable users to define custom logic for tool calling, environments like those from [OpenEnv](https://github.com/meta-pytorch/OpenEnv/tree/main) etc. Very much WIP :)

Example scripts included in `examples/scripts/openenv`

## TODO

- [x] Propagate reward from envs in `rollout_func` to `_calculate_rewards`
- [x] Streamline `rollout_func` signature
- [x] Find a better way to handle text environments in `rollout_func` because `/generate` endpoint only returns `prompt_ids` and `completion_ids`
- [ ] Mask tool calls for multi-step 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/examples/scripts/openenv/catch.py
+++ b/examples/scripts/openenv/catch.py
@@ -233,8 +233,6 @@ training_args = GRPOConfig(
     num_train_epochs=1,
     max_completion_length=4,
     gradient_accumulation_steps=4,
-    log_completions=True,
-    num_completions_to_print=1,
 )
 trainer = GRPOTrainer(
     model="Qwen/Qwen2.5-0.5B-Instruct",


trl/trainer/grpo_trainer.py
```diff
@@ -1116,20 +1129,35 @@ def _generate_single_turn(self, prompts: list[str], images: Optional[list]):
...
-                            min_p=0.0 if self.min_p is None else self.min_p,
-                            max_tokens=self.max_completion_length,
-                            truncate_prompt_tokens=self.max_prompt_length,
-                            guided_decoding_regex=self.guided_decoding_regex,
-                            generation_kwargs=self.args.generation_kwargs,
-                        )
+                        if self.rollout_func is not None:
+                            output = self.rollout_func(
+                                prompts=ordered_set_of_prompts,
+                                n=self.num_generations,
```
lewtun: I found it necessary to propagate the sampling parameters because e.g. if we set `num_generations` in the `GRPOConfig`, this information must be aligned in the `rollout_func` for consistency. 

The alternative would be to remove the sampling parameters altogether, and then assume the user aligns some of these params in their implementation of `rollout_func`

trl/experimental/openenv/echo.py
```diff
@@ -0,0 +1,146 @@
...
+8000
+
+-- Run this script -- CUDA_VISIBLE_DEVICES=1 python trl/experimental/openenv/echo.py
+"""
+
+GEN_URL = "http://0.0.0.0:8000/generate/"
+ENV_URL = "http://0.0.0.0:8001"
+
+# Start the Echo server in background
+print("⚡ Starting FastAPI server for Echo Environment...")
```
lewtun: We can't run Docker containers on the HF cluster, so I'm using FastAPI locally

trl/trainer/grpo_trainer.py
```diff
@@ -1116,28 +1131,39 @@ def _generate_single_turn(self, prompts: list[str], images: Optional[list]):
...
+                                top_k=-1 if self.top_k is None else self.top_k,
+                                min_p=0.0 if self.min_p is None else self.min_p,
+                                max_tokens=self.max_completion_length,
+                                truncate_prompt_tokens=self.max_prompt_length,
+                                guided_decoding_regex=self.guided_decoding_regex,
+                                generation_kwargs=self.args.generation_kwargs,
+                            )
+                        # Extract required fields and collect any extra fields for reward functions
+                        required_keys = {"prompt_ids", "completion_ids", "logprobs"}
+                        extra_fields = {k: v for k, v in output.items() if k not in required_keys}
```
lewtun, line 1160: Here is how I propagate the reward from the environment in `rollout_func` to the reward functions. I could have done it separately for the `rollout_func` branch, but then I would have to duplicate all the broadcasting logic

trl/trainer/grpo_trainer.py
```diff
@@ -1223,6 +1258,8 @@ def _generate_single_turn(self, prompts: list[str], images: Optional[list]):
                     completion_ids = all_completion_ids
                     logprobs = all_logprobs
 
+                extra_fields = {}  # No extra fields for colocate mode
```
lewtun, line 1252: For now I just focus on server mode. I can extend to the other modes later

trl/experimental/openenv/echo.py
```diff
@@ -0,0 +1,178 @@
...
+    gradient_accumulation_steps=4,
+)
+trainer = GRPOTrainer(
+    model="Qwen/Qwen2.5-0.5B-Instruct",
+    reward_funcs=reward_from_env,
+    args=training_args,
+    train_dataset=dataset,
+    rollout_func=rollout_func,
+)
+trainer.train()
```
kashif, line 168: after this perhaps we should stop the echo server?

trl/experimental/openenv/echo.py
qgallouedec, line 1: can you move this to examples/scripts instead? either have a dedicated `openenv` directory, are a standalone script, as you want

trl/experimental/openenv/echo.py
kashif: i can send a PR for that to cleanly stop things

trl/trainer/grpo_trainer.py
```diff
@@ -1116,28 +1129,39 @@ def _generate_single_turn(self, prompts: list[str], images: Optional[list]):
...
-                            generation_kwargs=self.args.generation_kwargs,
-                        )
-                        payload = (output["prompt_ids"], output["completion_ids"], output["logprobs"])
+                        if self.rollout_func is not None:
+                            output = self.rollout_func(
+                                prompts=ordered_set_of_prompts,
+                                images=ordered_set_of_images,
+                                args=self.args,
+                                processing_class=self.processing_class,
+                            )
```
qgallouedec, line 1151: rollout_func then requires  `use_vllm=True` and `vllm_mode="server"`, right?

examples/scripts/openenv/catch.py
```diff
@@ -0,0 +1,249 @@
...
+
+def reward_from_env(completions, **kwargs):
+    """Reward function that uses the environment reward from the catch game."""
+    # Extract environment rewards from kwargs (propagated via extra_fields)
+    env_rewards = kwargs.get("env_reward", [])
+    if env_rewards:
+        return [float(reward) for reward in env_rewards]
+    else:
+        # Fallback if env_reward is not available
+        return [0.0] * len(completions)
```
qgallouedec, line 224: ideally, we shouldn't have this. We could 

- allow `reward_func=None` in `GRPOTrainer.__init__`
- check if the rollout function returns a `"reward"` key, and if so, we use this as a column in `rewards_per_func` (I might be missing something here)

trl/trainer/grpo_trainer.py
lewtun: Ah yes, I assumed this was covered by placing it within this part of the if block: https://github.com/huggingface/trl/pull/4310/files#diff-964e6fd373aa93037604064cb2b822d7f8e2735e33f791065acf2c4c3552d393R1115

trl/experimental/openenv/echo.py
lewtun: done in https://github.com/huggingface/trl/pull/4310/commits/d8baa434f03cc66db809244dcdf74c27280f1833

examples/scripts/openenv/catch.py
lewtun: I'm going to leave this for follow up refactoring.

PR #4325, opened 2025-10-23
Use explicit tiny-Qwen2_5_VL model_id parameter in CI tests

albertvillanova:
Use explicit "tiny-Qwen2_5_VL" `model_id` param in CI tests.

This change improves test flexibility and maintainability by allowing tests to easily run with different model IDs. All hardcoded references to the model "trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration" have been replaced with the parametrized variable, and the corresponding test function signatures were updated.

No changes were made to the core test logic or assertions; the focus is on improving test parametrization and maintainability.


albertvillanova:
Thanks, @qgallouedec: good point!

I proposed this change with 2 ideas in mind:
1. Extendibility: It makes it straightforward to extend the tests to additional models in the future
2. Debugging: If multiple tests fail at once, having the model name in the test summary (via `pytest.mark.parametrize`) makes it much easier to identify whether the failure is specific to a given model.

So while it may look unnecessary for a single case right now, it helps keep things scalable and easier to debug later on.

PR #4331, opened 2025-10-23
Use explicit tiny-Qwen2ForCausalLM-2.5 model_id param in CI tests

albertvillanova:
Use explicit tiny-Qwen2ForCausalLM-2.5 model_id param in CI tests.

This change improves test flexibility and maintainability by allowing tests to easily run with different model IDs. All hardcoded references to the model "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5" have been replaced with the parametrized variable, and the corresponding test function signatures were updated.

No changes were made to the core test logic or assertions; the focus is on improving test parametrization and maintainability.

Follow-up to:
- #4325


qgallouedec:
In my opinion, the numerous repetitions of `model_id="Qwen/tiny-Qwen2ForCausalLM-2.5"` are not a problem. On the contrary, they allow the tests to be self-contained and easy to copy and paste. However, I find that using a fixture reduces consistency somewhat, because, for example, some tests in `TestSFTTrainer` will use the fixture `model_id`, but others (e.g., tests specific to GPT-OSS and VLMs) will keep an explicit `model_id`. Let me know what you think

albertvillanova:
@qgallouedec, I have addressed your requests and fixed the conflicts.

qgallouedec:
Thanks, Albert. As I mentioned, I don't think there's much point in parameterizing tests that only use one model anyway (and in most cases we don't want to extend the tests to more models).

The motivation is to increase maintainability, but for me it's not clear if it really increases maintainability.

PR #4335, opened 2025-10-24
Added custom `prepare_model_for_kbit_training` to save VRAM

sergiopaniego:
# What does this PR do?

This PR introduces a custom `prepare_model_for_kbit_training`  function used when training quantized models  + QLoRA.

The function is a small modification from [PEFT's version](https://github.com/huggingface/peft/blob/a18ba67f242ab2eb74cdabab76ea2fd836b5cd83/src/peft/utils/other.py#L170).That function upcasts to float32 a lot of params from the model, increasing the VRAM consumption. In constrained setups this can be problematic, so a more detailed selection of upcasted layers could benefit VRAM consumption.

I've identified 3 possible ways for instantiating the trainer and they're displayed in this [testing notebook](https://colab.research.google.com/drive/1hSJ_v8sQg61lFQaXcKLBOvKshCAPVmet?usp=sharing)

Testing them, the memory consumption differs as follows for that case (19.105 vs 14.156GB peak reserved VRAM memory):

PEFT func vs custom TRL func (VRAM usage)

I've tested the training procedure for 100 steps and the result can be found in [this trackio Space](https://huggingface.co/spaces/sergiopaniego/qwen3-14b-unsloth-bnb-4bit-testing-func).

Screenshot 2025-10-24 at 15 47 23

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec @kashif

### DIFF
--- a/trl/models/utils.py
+++ b/trl/models/utils.py
@@ -33,7 +33,7 @@ from .modeling_value_head import AutoModelForCausalLMWithValueHead, AutoModelFor
 
 if is_peft_available():
     import peft
-    from peft import PeftConfig, PeftModel, get_peft_model, prepare_model_for_kbit_training
+    from peft import PeftConfig, PeftModel, get_peft_model
 
 
 if TYPE_CHECKING:


sergiopaniego:
There are some trainers (BCO, CPO, DPO, KTO, ORPO) that won't be affected by this change since they currently call `prepare_model_for_kbit_training` from `peft` directly from the trainer ([example](https://github.com/huggingface/trl/blob/436e1c04309a77bad70cb0f04722164468ed5c36/trl/trainer/cpo_trainer.py#L199)). 

The ones affected are GRPO, Online DPO, PRM, Reward, RLOO, SFT as they call `prepare_peft_model` which is the function that then calls the custom `prepare_model_for_kbit_training` ([example](https://github.com/huggingface/trl/blob/436e1c04309a77bad70cb0f04722164468ed5c36/trl/trainer/grpo_trainer.py#L290)).

Let me know if this behavior needs to be reviewed

qgallouedec:
> There are some trainers (BCO, CPO, DPO, KTO, ORPO) that won't be affected by this change since they currently call `prepare_model_for_kbit_training` from `peft` directly from the trainer ([example](https://github.com/huggingface/trl/blob/436e1c04309a77bad70cb0f04722164468ed5c36/trl/trainer/cpo_trainer.py#L199)).
> 
> The ones affected are GRPO, Online DPO, PRM, Reward, RLOO, SFT as they call `prepare_peft_model` which is the function that then calls the custom `prepare_model_for_kbit_training` ([example](https://github.com/huggingface/trl/blob/436e1c04309a77bad70cb0f04722164468ed5c36/trl/trainer/grpo_trainer.py#L290)).
> 
> Let me know if this behavior needs to be reviewed

Yes, you're right. I just want to mention that this approach is not deliberate design choice and that ideally, all trainers should instantiate the model in the same way. But this is obviously out of the scope of this PR

PR #4344, opened 2025-10-27
[OpenENV] Openenv rollout_func signature proposal

kashif:
# What does this PR do?

Proposal to refactor the rollout_func to take the trainer as an argument. With this change, we should be able to use the trainer's generation helpers to generate.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/extras/vllm_client.py
+++ b/trl/extras/vllm_client.py
@@ -282,7 +282,7 @@ class VLLMClient:
     async def generate_async(
         self,
         prompts: list[str],
-        images: Optional[list] = None,
+        images: list | None = None,
         n: int = 1,
         repetition_penalty: float = 1.0,
         temperature: float = 1.0,
@@ -290,9 +290,9 @@ class VLLMClient:
         top_k: int = -1,
         min_p: float = 0.0,
         max_tokens: int = 16,
-        truncate_prompt_tokens: Optional[int] = None,
-        guided_decoding_regex: Optional[str] = None,
-        generation_kwargs: Optional[dict] = None,
+        truncate_prompt_tokens: int | None = None,
+        guided_decoding_regex: str | None = None,
+        generation_kwargs: dict | None = None,
     ) -> dict:
         """
         Asynchronously generates model completions for the provided prompts.


examples/scripts/openenv/echo.py
```diff
@@ -88,19 +88,19 @@
...
         prompts: List of prompts to generate from
-        args: GRPOConfig containing all sampling parameters
-        processing_class: Tokenizer/processor for decoding completions
+        trainer: The GRPOTrainer instance (required parameter)
 
     Returns:
         Dict containing prompt_ids, completion_ids, logprobs, and env_reward
     """
     # 1. Generate completions via vLLM inference server (running on port 8000)
+    args = trainer.args
```
burtenshaw, line 103: When using the current `rollout_func`, I found the definition of the vllm payload and how you need to use `requests` kinda messy/ low level. Mainly because I have to unpack and redefine the config (now trainer). 

Here's an idea: could we define a `generate` function within the trainer that uses closures from the trainer scope to set configuration defaults. The user could then use `generate` function as a param of rollout_func without having to deal with the underlying vllm logic. They can then override any config params of the function.

examples/scripts/openenv/catch_async_colocate.py
```diff
@@ -0,0 +1,294 @@
...
+import time
+from pathlib import Path
+
+import requests
+from datasets import Dataset
+from envs.openspiel_env import OpenSpielEnv
+from envs.openspiel_env.models import OpenSpielAction
+
+from trl import GRPOConfig, GRPOTrainer, RichProgressCallback, apply_chat_template
+from trl.extras.vllm_colocate_async import AsyncVLLMColocateWrapper
```
qgallouedec, line 53: I think you forgot to push this new file

qgallouedec:
I haven't played with OpenEnv much, so I'll leave it to @lewtun or @burtenshaw to review this one. Note that for now this rollout function is considered very experimental and it's likely that's we'll have to change/remove it in the future (depending on how OpenEnv is adopted by the community), so I wouldn't put too much effort to propagate/consolidate it.

kashif:
closing in favour of https://github.com/huggingface/trl/pull/4510

PR #4349, opened 2025-10-27
👑 [experimental] GOLD Trainer

kashif:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/docs/source/gold_trainer.md
+++ b/docs/source/gold_trainer.md
@@ -111,7 +111,7 @@ boundaries.
 
 ## GOLDTrainer
 
-[[autodoc]] GOLDTrainer
+[[autodoc]] experimental.gold.GOLDTrainer
     - train
     - generate_on_policy_outputs
     - save_model
@@ -119,4 +119,4 @@ boundaries.
 
 ## GOLDConfig
 
-[[autodoc]] GOLDConfig
+[[autodoc]] experimental.gold.GOLDConfig


trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+                self.trainer._last_student_sync_step = state.global_step
+
+
+class GOLDTrainer(SFTTrainer):
+    _tag_names = ["trl", "gold"]
+
+    def __init__(
+        self,
+        model: Optional[Union[PreTrainedModel, nn.Module, str]] = None,
+        model_revision: Optional[str] = None,
```
qgallouedec: any reason for having model_revision in the trainer arg instead of trainer config?

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+        eval_dataset: Optional[Union[Dataset, dict[str, Dataset]]] = None,
+        processing_class: Optional[
+            Union[PreTrainedTokenizerBase, BaseImageProcessor, FeatureExtractionMixin, ProcessorMixin]
+        ] = None,
+        compute_metrics: Optional[Callable[[EvalPrediction], dict]] = None,
+        callbacks: Optional[list[TrainerCallback]] = None,
+        optimizers: tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
+        preprocess_logits_for_metrics: Optional[Callable[[torch.Tensor, torch.Tensor], torch.Tensor]] = None,
+        peft_config: Optional["PeftConfig"] = None,
+        formatting_func: Optional[Callable] = None,
```
qgallouedec: Unless necessary, I would remove the formatting_func, as its use is now discouraged and it exists primarily for backward compatibility reasons.

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+    ):
+        self.model_name_or_path = model if isinstance(model, str) else model.config._name_or_path
+        self.model_revision = model_revision
+        # Ensure Trainer does not drop non-signature columns used by the collator (e.g., "prompts")
+        args.remove_unused_columns = False
```
qgallouedec: I recommend setting the signature column instead of modifying the args (see the other trainers)

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+        # Respect a user-provided data_collator; otherwise, provide a ChatML collator that
+        if data_collator is None:
+            data_collator = DataCollatorForChatML(tokenizer=processing_class, max_length=args.max_length)
+
+        # Ensure SFTTrainer does not pre-process the dataset when using a ChatML collator,
+        # so that raw conversational fields (e.g., "messages") remain available to the collator.
+        if args.dataset_kwargs is None:
+            args.dataset_kwargs = {"skip_prepare_dataset": True}
+        else:
+            args.dataset_kwargs["skip_prepare_dataset"] = True
```
qgallouedec: I don't think you need this if you've set the right signature column

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+            raise ValueError(
+                "You passed teacher_model_init_kwargs to the GOLDConfig, but your teacher_model is already instantiated."
+            )
+        else:
+            teacher_model_init_kwargs = args.teacher_model_init_kwargs
+            teacher_model_init_kwargs["torch_dtype"] = (
+                teacher_model_init_kwargs["torch_dtype"]
+                if teacher_model_init_kwargs["torch_dtype"] in ["auto", None]
+                else getattr(torch, teacher_model_init_kwargs["torch_dtype"])
+            )
```
qgallouedec, line 788: It could be nice to use `create_model_from_path` for the models creation

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+        self.beta = args.beta
+        self.temperature = args.temperature
+        self.top_p = args.top_p
+        self.seq_kd = args.seq_kd
+
+        # Separate loss logs
+        self._on_sum = 0.0
+        self._off_sum = 0.0
+        self._on_step_eq = 0.0
+        self._off_step_eq = 0.0
```
qgallouedec: can you find more informative names for these variable? Or at least comment so that we know what it is

trl/experimental/gold/gold_config.py
```diff
@@ -0,0 +1,419 @@
...
+        student_vllm_tensor_parallel_size (`int`, *optional*, defaults to `1`):
+            Tensor parallel size for the colocated student vLLM engine (if `student_vllm_mode="colocate"`).
+        student_vllm_guided_decoding_regex (`str` or `None`, *optional*, defaults to `None`):
+            Regex for vLLM guided decoding for the student model.
+        student_vllm_sync_frequency (`int`, *optional*, defaults to `1`):
+            Frequency (in training steps) to synchronize student model weights to vLLM engine. Set to 1 to sync after
+            every step.
+        student_vllm_enable_sleep_mode (`bool`, *optional*, defaults to `False`):
+            Whether to enable sleep mode for the student vLLM engine. If set to `True`, the engine will enter sleep
+            mode after each training step to save resources.
```
qgallouedec: IMO, you don't need the prefix `student`.

trl/experimental/gold/gold_config.py
```diff
@@ -0,0 +1,419 @@
...
+    )
+    beta: float = field(
+        default=0.5,
+        metadata={
+            "help": "Interpolation coefficient between `0.0` and `1.0` of the Generalized Jensen-Shannon Divergence "
+            "loss. When beta is `0.0`, the loss is the KL divergence. When beta is `1.0`, the loss is the Inverse KL "
+            "Divergence."
+        },
+    )
+    max_new_tokens: int = field(
```
qgallouedec: I'd use `max_completion_length` to align with other trainers

trl/experimental/gold/gold_config.py
```diff
@@ -0,0 +1,419 @@
...
+            "help": "Number of optimization steps per generation. If `None`, it defaults to gradient_accumulation_steps."
+        },
+    )
+    data_subset: Optional[float] = field(
+        default=None,
+        metadata={
+            "help": "Subset of the training data to use. If a float, it is interpreted as a fraction of the dataset. "
+            "If an integer, it is interpreted as the number of samples to use."
+        },
+    )
```
qgallouedec: this is not used 

```suggestion
```

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+    is_rich_available,
+)
+
+from trl.extras.profiling import profiling_decorator
+from trl.extras.vllm_client import VLLMClient
+from trl.import_utils import is_vllm_available
+from trl.models import prepare_deepspeed
+from trl.models.utils import unwrap_model_for_generation
+from trl.trainer.sft_trainer import SFTTrainer
+from trl.trainer.utils import (
```
qgallouedec: can you use relative imports instead? for consistency

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+        """
+        # Check if dataset is already processed
+        column_names = list(next(iter(dataset)).keys())
+        is_processed = "input_ids" in column_names
+
+        # Use our enhanced dataset preparation for:
+        # 1. ULD loss with cross-tokenizer (need original text preservation)
+        # 2. Any unprocessed dataset (need attention_mask for DataCollatorForChatML)
+        if not is_processed or (
+            hasattr(self, "use_uld_loss")
```
qgallouedec: Unless very good reason, I'd avoid using `hasattr` as much as possible. An _explicit_ condition exists almost every time. Here it's easier: `use_uld_loss` is set at init, so we know it's always true

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+                teacher_tokenizer=self.teacher_tokenizer,
+            )
+
+        self.generation_config = GenerationConfig(
+            max_new_tokens=args.max_new_tokens,
+            temperature=args.temperature,
+            top_p=args.top_p,
+            do_sample=True,
+            top_k=args.top_k,
+            use_cache=False if args.gradient_checkpointing else True,
```
qgallouedec:
```suggestion
```

cache must be enabled for generation and disabled for forward

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+                            )
+                            for i in range(self.accelerator.num_processes // args.student_vllm_tensor_parallel_size)
+                        ]
+                    )
+
+                # vLLM requires the environment variables to be set for distributed training.
+                os.environ["RANK"] = str(self.accelerator.process_index)
+                os.environ["LOCAL_RANK"] = str(self.accelerator.local_process_index)
+                os.environ["WORLD_SIZE"] = str(self.accelerator.num_processes)
+                os.environ["MASTER_ADDR"] = os.environ.get("MASTER_ADDR", "localhost")
```
qgallouedec: you should be able to use `ensure_master_addr_port()` now

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+                answers_index.append(0)
+            else:
+                diff_indices = indices[1:] - indices[:-1]
+                break_index = (diff_indices != 1).nonzero()
+                length = (break_index[0].item() + 1) if len(break_index) > 0 else len(indices)
+                answers_index.append(length - 1)
+        return answers_index, answers_size
+
+
+class GOLDStudentVLLMSyncCallback(TrainerCallback):
```
qgallouedec: why do we use a callback here but not in other other trainers?

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,2038 @@
...
+            and state.global_step % self.trainer.student_vllm_sync_frequency == 0
+        ):
+            # Check if this is a step where gradients are synchronized
+            # This happens at the end of gradient accumulation cycles
+            if hasattr(self.trainer.accelerator, "sync_gradients") and self.trainer.accelerator.sync_gradients:
+                self.trainer._move_student_model_to_vllm()
+                self.trainer._last_student_sync_step = state.global_step
+
+
+class GOLDTrainer(SFTTrainer):
```
qgallouedec, line 723: almost all methods from SFT are overridden in this trainer. For maintainability reasons, I suggest inheriting this trainer from BaseTrainer rather than SFTTrainer. The same applies to its configuration.

qgallouedec (commented):
First review

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,1984 @@
...
+        packing: bool,
+        formatting_func: Optional[Callable[[dict], str]],
+        dataset_name: str,
+    ) -> Union[Dataset, IterableDataset]:
+        """
+        Prepare dataset while preserving original text for cross-tokenizer distillation.
+        """
+        from accelerate import PartialState
+
+        from ..data_utils import is_conversational, maybe_convert_to_chatml, pack_dataset, truncate_dataset
```
qgallouedec: always use top-level import

trl/experimental/gold/gold_config.py
```diff
@@ -0,0 +1,419 @@
...
+from dataclasses import dataclass, field
+from typing import Any, Optional
+
+from transformers import TrainingArguments
+
+from trl.trainer.sft_config import SFTConfig
+
+
+@dataclass
+class GOLDConfig(SFTConfig):
```
qgallouedec, line 24: @cmpatino mentioned that the the method is quite sensitive to hyperparameters. Consequently, I recommend having good defaults here (and not rely on defaults from SFT). Eg lr=1e-7

docs/source/gold_trainer.md
```diff
@@ -0,0 +1,99 @@
...
+
+student_name = "meta-llama/Llama-3.2-1B-Instruct"
+teacher_name = "Qwen/Qwen2.5-0.5B-Instruct"
+
+tokenizer = AutoTokenizer.from_pretrained(student_name)
+if tokenizer.pad_token is None:
+    tokenizer.pad_token = tokenizer.eos_token
+
+model = AutoModelForCausalLM.from_pretrained(student_name)
+teacher_model = AutoModelForCausalLM.from_pretrained(teacher_name)
```
qgallouedec, line 71: loading model outside the trainer should be for advance use case, and not for quick-start. Same remark for setting training arguments and loading the tokenizer.

 Ideally, the api should look like this:

```python
from trl.experimental.gold import GoldConfig, GoldTrainer

dataset = load_dataset("HuggingFaceTB/OpenR1-Math-220k-default-verified", split="train")

trainer = GOLDTrainer(
    model="meta-llama/Llama-3.2-1B-Instruct",
    teacher="Qwen/Qwen2.5-0.5B-Instruct",
    train_dataset=train_dataset,
)
trainer.train()
```

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,1984 @@
...
+        formatting_func: Optional[Callable[[dict], str]],
+        dataset_name: str,
+    ) -> Union[Dataset, IterableDataset]:
+        """
+        Override dataset preparation to preserve original text for cross-tokenizer distillation and ensure
+        attention_mask is always added for DataCollatorForChatML compatibility.
+        """
+        # Check if dataset is already processed
+        column_names = list(next(iter(dataset)).keys())
+        is_processed = "input_ids" in column_names
```
qgallouedec, line 1005: unless there is a case that I'm not aware of, I think we should assume that the data is always not processed. It would simplify things

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,1984 @@
...
+                completion_texts,
+            )
+
+            teacher_input_ids = teacher_input_ids.to(self.accelerator.device)
+            teacher_labels = teacher_labels.to(self.accelerator.device)
+            teacher_attention_mask = teacher_attention_mask.to(self.accelerator.device)
+
+            outputs_student = model(
+                input_ids=inputs["input_ids"],
+                attention_mask=inputs["attention_mask"],
```
qgallouedec, line 1373:
```suggestion
                attention_mask=inputs["attention_mask"],
                use_cache=False,
```

to avoid warning with gradient checkpointing

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,1984 @@
...
+                # labels mask and labels (shifted)
+                labels_mask = inputs["labels"] != -100
+                masked_input_ids = torch.where(
+                    labels_mask, inputs["input_ids"], torch.full_like(inputs["input_ids"], -100)
+                )
+                true_labels = masked_input_ids[:, 1:].contiguous()
+
+                # Release intermediate tensors
+                del labels_mask, masked_input_ids
+
```
qgallouedec:
```suggestion
```

these are tiny, you can safely keep them

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,1984 @@
...
+                    if self.uld_loss_fn.last_unmatched_loss is not None
+                    else 0.0
+                )
+
+                self._matched_sum += matched_val
+                self._unmatched_sum += unmatched_val
+                self._matched_step_eq += step_eq
+                self._unmatched_step_eq += step_eq
+
+        empty_cache()
```
qgallouedec, line 1531: out of curiosity, did you try to ablate this? normally we don't have to handle cache emptying so I'm always surprised to see this

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,1984 @@
...
+    def generate_on_policy_outputs(self, model, inputs, generation_config, pad_token_id=None):
+        # Generate output with respect to the prompt only
+        if self.use_transformers_paged:
+            previous_attn = self.model.config._attn_implementation
+            if is_flash_attn_2_available():
+                model.config._attn_implementation = "paged_attention"
+            else:
+                model.config._attn_implementation = "sdpa_paged"
+            generated_outputs = model.generate_batch(
+                inputs["prompts"].tolist(),
```
qgallouedec: if you just use `tolist`, most prompts will start with padding tokens, which is a waste of ressources. Instead you could use something like

```python
ids = [row[m.bool()].tolist() for row, m in zip(tensor, mask)]
```

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,1984 @@
...
+                generation_config=generation_config,
+            )
+            model.config._attn_implementation = previous_attn
+
+            completion_ids = [output.generated_tokens for output in generated_outputs.values()]
+            generated_tokens = torch.stack([torch.tensor(ids, device=model.device) for ids in completion_ids])
```
qgallouedec, line 1556: I'm quite surprised that this doesn't fail, because it's likely that most completions don't share the same length

trl/experimental/gold/gold_trainer.py
```diff
@@ -0,0 +1,1993 @@
...
+        self.model_name_or_path = model if isinstance(model, str) else model.config._name_or_path
+        revision = args.student_model_revision
+        if isinstance(model, str) and revision is not None:
+            args.model_init_kwargs = args.model_init_kwargs or {}
+            args.model_init_kwargs["revision"] = revision
+
+        # Respect a user-provided data_collator; otherwise, provide a ChatML collator that
+        if data_collator is None:
+            data_collator = DataCollatorForChatML(tokenizer=processing_class, max_length=args.max_length)
+
```
qgallouedec, line 764: If I understand correctly, the data is already preprocessed, so most of the logic from `DataCollatorForChatML` isn't used. Consequently, I recommend having a custom collator, like in SFT and Reward, that only features what's necessary for the collation.

trl/experimental/gold/gold_trainer.py
kashif: so here we are in the generate_batch part that returns a dict whose values are already right padded to the same length so it doesnt fail

PR #4361, opened 2025-10-28
Add `add_generation_prompt` to processor_kwargs in GRPO and RLOO trainer

qgallouedec:
This fixes a bug introduced in #4155


trl/trainer/rloo_trainer.py
pramodith, line 1138: Quick question the else block will also include `add_generation_prompt` wonder if that'll just be ignored by the tokenizers __call__ or if it'd raise an error?

trl/trainer/grpo_trainer.py
pramodith, line 1312: Same comment for grpo too!

PR #4366, opened 2025-10-29
Fix CI experimental tests TypeError for GRPOWithReplayBufferTrainer.update_with_replay_buffer

albertvillanova:
Fix CI experimental tests TypeError for GRPOWithReplayBufferTrainer.update_with_replay_buffer:
- Rename prompt_inputs to forward_kwargs param to update_with_replay_buffer

Fix #4365.

PR #4391, opened 2025-10-30
Fix GRPO with replay buffer by inserting images in the prompt

albertvillanova:
Fix GRPO with replay buffer by inserting images in the prompt. Additionally, fix the CI test `test_training_with_replay_buffer`.

Follow-up to:
- #4155

Currently, GRPO with Replay Buffer raises an error: https://github.com/huggingface/trl/actions/runs/18940392458/job/54077463859
> TypeError: GRPOTrainer._generate() takes 2 positional arguments but 3 were given

Stacktrace:
```python
>       trainer.train()

tests/experimental/test_grpo_with_replay_buffer_trainer.py:282: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.13/site-packages/transformers/trainer.py:2325: in train
    return inner_training_loop(
.venv/lib/python3.13/site-packages/transformers/trainer.py:2674: in _inner_training_loop
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.13/site-packages/transformers/trainer.py:4014: in training_step
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
trl/extras/profiling.py:98: in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
trl/trainer/grpo_trainer.py:1037: in _prepare_inputs
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <trl.experimental.grpo_with_replay_buffer.grpo_with_replay_buffer_trainer.GRPOWithReplayBufferTrainer object at 0x7fb2382151d0>
inputs = [{'prompt': "Although that way may not be obvious at first unless you're"}, {'prompt': "Although that way may not be o...may not be obvious at first unless you're"}, {'prompt': "Although that way may not be obvious at first unless you're"}]

    def _generate_and_score_completions(
        self, inputs: list[dict[str, Union[torch.Tensor, Any]]]
    ) -> dict[str, Union[torch.Tensor, Any]]:
    
        if "images" in inputs[0]:
            images = [example.get("images") for example in inputs]
        elif "image" in inputs[0]:
            images = [[example.get("image")] if example.get("image") is not None else None for example in inputs]
        else:
            images = None
        # Transformers requires at least one image in the batch, otherwise it throws an error
        if images is not None and all(img_list == [] for img_list in images):
            images = None
    
        (
            prompt_ids,
            completion_ids,
            prompt_mask,
            completion_mask,
            num_items_in_batch,
            sampling_per_token_logps,
            forward_kwargs,
>       ) = self._generate(prompts, images)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       TypeError: GRPOTrainer._generate() takes 2 positional arguments but 3 were given

trl/experimental/grpo_with_replay_buffer/grpo_with_replay_buffer_trainer.py:91: TypeError
```


trl/experimental/grpo_with_replay_buffer/grpo_with_replay_buffer_trainer.py
```diff
@@ -209,10 +270,6 @@ def _generate_and_score_completions(
         )
         all_process_advantages = advantages.clone()  # keep the aggregated advantages for logging
         advantages = advantages[process_slice]
-        if std_rewards is None:
-            std_rewards = rewards.view(-1, self.num_generations).std(dim=1)
-            std_rewards = std_rewards.repeat_interleave(self.num_generations, dim=0)
-        std_rewards = std_rewards[process_slice] if std_rewards is not None else None
```
albertvillanova, line 215: This last line is not in GRPOTrainer: is it necessary? If so, shouldn't we implement it in GRPOTrainer as well?
```python
std_rewards = std_rewards[process_slice]
```

qgallouedec: @pramodith if you've some time to check

pramodith: Will take a look a bit later this evening!

pramodith: We need the sliced `std_rewards` in this trainer because we decide if a specific example should be added to the replay buffer or sampled from the buffer based on `std_reward` of that specific rollout. Since each gpu sees a unique batch of data we need to only perform the buffer lookup and update based on the slice residing in the gpu.

`GRPOTrainer` doesn't **need the std after advantage scores are computed** so it can be discarded in GRPOTrainer.

pramodith: This shows how `std_rewards` is used for updating the replay buffer https://github.com/huggingface/trl/blob/1c2322eb7d78b60560cc1f1e52aca1d1d3ef2047/trl/experimental/grpo_with_replay_buffer/grpo_with_replay_buffer_trainer.py#L407-L411

pramodith: This entire block of code removed should remain in `grpo_with_replay_buffer_trainer.py` we always need the group level std to determine what goes into the replay. buffer.

albertvillanova: Thanks for the explanation about `std_rewards = std_rewards[process_slice]`,  @pramodith. :hugs: 

Respect to the block lines just before, I removed the condition `if std_rewards is None` because I think this is always False. Just some lines above, we have this code: https://github.com/albertvillanova/trl/blob/d0324230761e7860646f4d15d7ff8beb433103ac/trl/experimental/grpo_with_replay_buffer/grpo_with_replay_buffer_trainer.py#L250-L260
```python
        if self.scale_rewards in ["group", "none"]:
            # If self.scale_rewards = "none", we'll still log group level std
            std_rewards = rewards.view(-1, self.num_generations).std(dim=1)
            std_rewards = std_rewards.repeat_interleave(self.num_generations, dim=0)
        elif self.scale_rewards == "batch":
            # Compute global std
            std_rewards = rewards.std().expand_as(rewards)
        else:
            raise ValueError(
                f"Invalid value for scale_rewards: {self.scale_rewards}. Must be one of 'batch', 'group', or 'none'."
            )
```
Therefore, `std_rewards` can't be None if I understand correctly. It should always be a `torch.Tensor`.

pramodith: Hmmm, yeah you're right but there's a bug here. The replay buffer requires the `std` to be computed over the group. I'll fix that in a subsequent PR, getting rid of that block in this PR is fine.

albertvillanova: I re-added just that line: https://github.com/huggingface/trl/pull/4391/commits/4266550534ac69feb0729ea0c0f3bb0ba4666ab2
:wink:

albertvillanova:
The failing test will be fixed after the merge of:
- #4366

qgallouedec:
Mmh wait we've

```
tests/experimental/test_grpo_with_replay_buffer_trainer.py::TestUpdateWithReplayBuffer::test_update_with_inputs_different_seq_len - TypeError: GRPOWithReplayBufferTrainer.update_with_replay_buffer() got an unexpected keyword argument 'prompt_inputs'. Did you mean 'prompt_ids'?
```

albertvillanova:
That is fixed in my previous PR. See my comment above, @qgallouedec.

albertvillanova:
@qgallouedec, there were 2 different bugs:
- TypeError: GRPOWithReplayBufferTrainer.update_with_replay_buffer() got an unexpected keyword argument 'prompt_inputs'. Did you mean 'prompt_ids'?
  - Fixed by:
    - #4366
- TypeError: GRPOTrainer._generate() takes 2 positional arguments but 3 were given
  - Fixed by this PR:
    - #4391

qgallouedec:
oh my bad, thanks for fixing it

qgallouedec:
~Let's merge #4366 to keep things clean~


EDIT Let's merge #4366 first to keep things clean

PR #4410, opened 2025-10-30
Add On-Policy Distillation from thinking labs to paper index.

pramodith:
# What does this PR do?

Add On-Policy Distillation from thinking labs to paper index. https://thinkingmachines.ai/blog/on-policy-distillation/

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@kashif @qgallouedec 


trl/experimental/gold/gold.py
```diff
@@ -21,34 +21,17 @@
...
-    --model_name_or_path meta-llama/Llama-3.2-1B-Instruct \
-    --teacher_model_name_or_path Qwen/Qwen2-1.5B-Instruct \
-    --dataset_name trl-lib/chatbot_arena_completions \
-    --learning_rate 2e-5 \
-    --per_device_train_batch_size 4 \
-    --gradient_accumulation_steps 8 \
-    --output_dir gold-model \
-    --num_train_epochs 1 \
-    --push_to_hub \
+# Full training: python trl/experimental/gold/gold.py \
```
qgallouedec: I don't know why it tries to format this, it doesn't do it for the other scripts.
Anyway, you should revert this change

docs/source/paper_index.md
```diff
@@ -580,3 +580,47 @@ def add_margin(example):
 
 dataset = dataset.map(add_margin)
 ```
+
+## Distillation
+Papers relating to training a student model with the help of a teacher model.
+
+### On-Policy Distillation
```
qgallouedec, line 612: Usually in this section we try to have the paper/blog title as the title for the subsection. So here, we would like to have 2 subsections, one for each blog post.

docs/source/paper_index.md
```diff
@@ -580,3 +580,47 @@ def add_margin(example):
...
+Additionally on-policy distillation is more compute efficient and is less prone to overfitting when trained with limited data.
+
+To train a model with on-policy distillation using TRL, you can use the following configuration, with the [`GKDTrainer`] and [`GKDConfig`]:
+
+```python
+from trl import GKDConfig
+
+config = GKDConfig(
+    lmbda=1.0, # student produces rollouts for all batches
+    beta=1.0, # to ensure reverse-kl as the loss function
```
pramodith, line 634: Can someone please confirm if `beta` should be 1.0? The thinking lab's block mentions that we should use reverse-kl, but I see that in the HF blog recommends setting beta=0 for the distillation step., so I'm a bit confused.

From thinking labs:

>Reverse KL has natural synergy with RL, which generally optimizes a form of sequence-level reverse KL induced by the reward model. However, unlike most reward models in practice, the reverse KL is “unhackable” in the sense that low KL always corresponds to a high probability of desirable behavior from the teacher model’s point of view. Two other useful properties of reverse KL are that it is “mode seeking”See [Eric Jang’s post](https://blog.evjang.com/2016/08/variational-bayes.html) for more discussion of mode seeking behaviors. — it learns one specific behavior (the teacher’s) instead of spreading its distribution across several suboptimal options — and it reduces [exposure bias](https://arxiv.org/abs/1506.03099).

HF blog's linked recipe for distillation

```
accelerate launch \
  --config_file examples/accelerate_configs/multi_gpu.yaml trl/experimental/gold/gold.py \
  --model_name_or_path <sft-model> \
  --dtype auto \
  --attn_implementation kernels-community/flash-attn \
  --dataset_name allenai/tulu-3-sft-mixture \
  --dataset_train_split train \
  --bf16 \
  --learning_rate 1e-7 \
  --gradient_checkpointing \
  --gradient_accumulation_steps 64 \
  --num_train_epochs 1 \
  --eval_strategy steps \
  --eval_steps 100 \
  --temperature 1.0 \
  --top_p 0.95 \
  --top_k 0 \
  --max_new_tokens 2048 \
  --max_prompt_length 512 \
  --lmbda 0.25 \
  --beta 0.0 \
```

docs/source/paper_index.md
pramodith: Ummm, a bit confused the title of thinking machine's blog is `On-Policy Distillation`

docs/source/paper_index.md
qgallouedec: cc @kashif

docs/source/paper_index.md
qgallouedec: My bad you are right

docs/source/paper_index.md
```diff
@@ -580,3 +580,47 @@ def add_margin(example):
...
+
+)
+```
+
+Alternatively, you can use the [`GoldTrainer`] and [`GoldConfig`] to perform on-policy distillation with a similar configuration:
+
+```python
+from trl.experimental import GoldConfig
+
+config = GoldConfig(
```
qgallouedec: To confirm, this is supposed to be equivalent to the above configuration right?

pramodith: Yes! Supposed to be the exact same.

```suggestion
Alternatively, you can use the [`GOLDTrainer`] and [`GOLDConfig`] to perform on-policy distillation with a similar configuration:

```python
from trl.experimental import GOLDConfig

config = GOLDConfig(
```

docs/source/paper_index.md
kashif: lets see the GKD paper had beta from 0.1, 0.5 and 0.9, and as the beta-> 1 the gradient of the loss behaves like inverse-KL. In the code when `beta=1.0` we do: `            jsd = F.kl_div(teacher_log_probs, student_log_probs, reduction="none", log_target=True)` which is the KL(student || teacher)

pramodith: Yes, so it seems like setting `beta=1` is the right thing to do for reproducing what the Thinking Machines blog state. Was the decision to use beta=0 instead of 1.0 in the HF blog intended?

cmpatino: That's a good catch.

Using `beta=0` was intentional because it was more sample efficient compared to `beta=1` for some ablations with the Countdown task. We also used `lambda=0.25` instead of `lambda=1.0`, as lower lambda values are faster to run than `lambda=1` and with comparable results to those of an entirely online setup (see Figure 5 from the blogpost).

You're right that for exact reproduction of the Thinking Machine blog, we should use `beta=1` and `lambda=1`. However, we wanted to include the parameters we used for the blog post in case someone wanted to reproduce those results. 

Running the experiments with the Thinking Machines setup will likely result in marginally better performance, but the conclusion would be essentially the same.

pramodith: Thanks for the clarification!

pramodith: @qgallouedec this one should be fine to merge.

PR #4415, opened 2025-10-31
Replace duplicate test with model_id parametrized test

albertvillanova:
This PR updates the vision-language model (VLM) training tests to handle special cases for the Gemma model more consistently. The main change is the consolidation of Gemma-specific logic into parameterized tests, removing the need for a separate test function.

**Test suite improvements:**

* Added `trl-internal-testing/tiny-Gemma3ForConditionalGeneration` to the parameterized model list in `test_train_vlm_prompt_completion`, ensuring Gemma-specific behavior is tested alongside other models.

* Removed the separate `test_train_vlm_prompt_completion_gemma` function, consolidating Gemma testing into the main parameterized test and reducing duplication.

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1430,6 +1430,8 @@ class TestSFTTrainer(TrlTestCase):
         "model_id",
         [
             "trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration",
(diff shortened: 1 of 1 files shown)


qgallouedec (approved):
I can't remember why we didn't write it this way in the first place 🧐

PR #4442, opened 2025-11-03
[GRPOTrainer bug fix] a little bug in completions with bootstrap

SolarWindRider:
## 1. Fix a little bug in GRPOTrainer.

When inputs are conversational, bootstrap is a list not str, causing a str concat error.

I add a bootstrap type check to fix the bug while maintain the original program

## 2. Add a new GRPO family Trainer


qgallouedec:
thanks, can you slit this PR into two PRs instead please?

SolarWindRider:
> thanks, can you slit this PR into two PRs instead please?

Sure, I split the grpotrainer bug fix in there: 
https://github.com/huggingface/trl/pull/4452/#issue-3589701263

PR #4450, opened 2025-11-04
💬 Add chat to vLLM client and server, update trainer calls

qgallouedec:
Slow tests pass locally, and GRPO training works as well:

```python

dataset = load_dataset("trl-lib/ultrafeedback-prompt", split="train")

    return [len(c[0]["content"]) for c in completions]

trainer = GRPOTrainer(
    model="Qwen/Qwen3-0.6B",
    train_dataset=dataset,
    args=GRPOConfig(
        use_vllm=True,
    ),
)
trainer.train()
```


trl/scripts/vllm_serve.py
```diff
@@ -616,6 +616,143 @@ async def generate(request: GenerateRequest):
         ]
         return {"prompt_ids": prompt_ids, "completion_ids": completion_ids, "logprobs": logprobs}
 
+    class ChatRequest(BaseModel):
```
qgallouedec, line 618: Exactly the same as generate, expect:

- images are within the messages (so we drop images from args)
- chat_template_kwargs argument added

trl/trainer/grpo_trainer.py
```diff
@@ -1189,9 +1189,8 @@ def _generate_single_turn(self, prompts: list):
                             )
                         else:
                             if is_conversational({"prompt": ordered_set_of_prompts[0]}):
-                                # FIXME: this endpoint doesn't exist in vllm_client
                                 output = self.vllm_client.chat(
-                                    prompts=ordered_set_of_prompts,
+                                    messages=ordered_set_of_prompts,
```
qgallouedec, line 1194: I use "messages" instead of "prompt" to align with vLLM

trl/extras/vllm_client.py
```diff
@@ -265,6 +266,59 @@ def pil_to_base64(image):
...
+                "generation_kwargs": generation_kwargs or {},
+                "chat_template_kwargs": chat_template_kwargs or {},
+            },
+        )
+        if response.status_code == 200:
+            json_response = response.json()
+            return {
+                "prompt_ids": json_response["prompt_ids"],
+                "completion_ids": json_response["completion_ids"],
+                "logprobs": json_response["logprobs"],
```
kashif, line 361: will the logprobs be returned always or is there a flag?

tests/test_vllm_client_server.py
```diff
@@ -99,6 +99,26 @@ def test_generate(self):
         for seq in completion_ids:
             assert all(isinstance(tok, int) for tok in seq)
 
+    def test_chat(self):
```
lewtun, line 102: It would be good to check that the issues with double BOS tokens getting inserted have been fully resolved (e.g. for a Llama model): https://github.com/vllm-project/vllm/issues/9519

@edbeeching ran into this during https://huggingface.co/spaces/HuggingFaceH4/blogpost-scaling-test-time-compute and it has a subtle, but negative impact on the generations.

trl/scripts/vllm_serve.py
```diff
@@ -615,6 +615,143 @@ async def generate(request: GenerateRequest):
...
+        messages: list[list[dict]]
+        n: int = 1
```
lewtun, line 620: Not for this PR, but more generally is there a reason we don't expose the full set of `SamplingParams` from vLLM?

trl/scripts/vllm_serve.py
```diff
@@ -615,6 +615,143 @@ async def generate(request: GenerateRequest):
...
+            "n": request.n,
+            "repetition_penalty": request.repetition_penalty,
+            "temperature": request.temperature,
+            "top_p": request.top_p,
+            "top_k": request.top_k,
+            "min_p": request.min_p,
+            "max_tokens": request.max_tokens,
+            "truncate_prompt_tokens": request.truncate_prompt_tokens,
+            "guided_decoding": guided_decoding,
+            "logprobs": 0,
```
lewtun: Why are `logprobs` being provided as a generation kwarg? If we want zero logprobs returned, it should be `None`: https://docs.vllm.ai/en/latest/api/vllm/index.html#vllm.SamplingParams.logprobs

trl/extras/vllm_client.py
```diff
@@ -265,6 +266,59 @@ def pil_to_base64(image):
...
+        top_p: float = 1.0,
+        top_k: int = -1,
+        min_p: float = 0.0,
+        max_tokens: int = 16,
+        truncate_prompt_tokens: int | None = None,
+        guided_decoding_regex: str | None = None,
+        generation_kwargs: dict | None = None,
+        chat_template_kwargs: dict | None = None,
+    ) -> list[list[int]]:
+        """ """
```
albertvillanova: The docstring of the `chat` method is missing here.

trl/extras/vllm_client.py
```diff
@@ -265,6 +266,59 @@ def pil_to_base64(image):
...
+        temperature: float = 1.0,
+        top_p: float = 1.0,
+        top_k: int = -1,
+        min_p: float = 0.0,
+        max_tokens: int = 16,
+        truncate_prompt_tokens: int | None = None,
+        guided_decoding_regex: str | None = None,
+        generation_kwargs: dict | None = None,
+        chat_template_kwargs: dict | None = None,
+    ) -> list[list[int]]:
```
albertvillanova: The return type hint is wrong: it actually returns a `dict` not a `list`.

trl/extras/vllm_client.py
```diff
@@ -265,6 +266,59 @@ def pil_to_base64(image):
...
+        """ """
+        url = f"{self.base_url}/chat/"
+
+        # Convert PIL images to base64 strings
+        for message_list in messages:
+            for message in message_list:
+                if isinstance(message["content"], list):
+                    for part in message["content"]:
+                        if part["type"] == "image_pil":
+                            part["image_pil"] = pil_to_base64(part["image_pil"])
```
albertvillanova, line 337: Note that here you are modifying the input `messages` in-place: is this intended?
If the caller reuses `messages`, they will have base63 strings instead of PIL images.

trl/scripts/vllm_serve.py
```diff
@@ -615,6 +615,143 @@ async def generate(request: GenerateRequest):
...
+        completion_ids: list[list[int]]
+        logprobs: list[list[float]]
+
+    @app.post("/chat/", response_model=ChatResponse)
+    async def chat(request: ChatRequest):
+        """
+        Generates completions for the provided chat messages.
+
+        Args:
+            request (`GenerateRequest`):
```
albertvillanova: Copy-paste error?
```suggestion
            request (`ChatRequest`):
```

trl/scripts/vllm_serve.py
```diff
@@ -615,6 +615,143 @@ async def generate(request: GenerateRequest):
...
+                - `guided_decoding_regex` (`str`, *optional*): A regex pattern for guided decoding. If provided, the
+                  model will only generate tokens that match this regex pattern.
+                - `generation_kwargs` (`dict`, *optional*): Additional generation parameters to pass to the vLLM
+                  `SamplingParams`. This can include parameters like `seed`, `frequency_penalty`, etc. If it contains
+                  keys that conflict with the other parameters, they will override them.
+                - `chat_template_kwargs` (`dict`, *optional*): Additional keyword arguments to pass to the chat
+                  template.
+
+        Returns:
+            `GenerateResponse`:
```
albertvillanova: The same:

```suggestion
            `ChatResponse`:
```

lewtun (approved):
LGTM with a suggestion to double-check models like Llama are not getting a double BOS token.

trl/extras/vllm_client.py
kashif: perhaps before doing this we should check if the `isinstance(part["image_pil"], Image.Image)`

trl/extras/vllm_client.py
qgallouedec: oups, fixed!

trl/extras/vllm_client.py
qgallouedec: @albertvillanova good point, fixed!

@kashif, at this point we shouldn't have something else than a PIL Image. If it's not a PIL Image, there is an issue with the data preparation

trl/extras/vllm_client.py
qgallouedec: yep, always, thanks to 

```python
generation_kwargs = {
    ...
    "logprobs": 0,
}
```

in the server

trl/scripts/vllm_serve.py
qgallouedec: The API isn't very intuitive, but when you pass `logprobs=0`, it actually means that you want the generation to return `logprobs`, but only for sampled tokens. `None` means no logprobs returned at all.
I'll add a comment to reduce confusion.

trl/scripts/vllm_serve.py
qgallouedec: Yes, discussed https://github.com/huggingface/trl/issues/3562#issuecomment-2964355101 and https://github.com/huggingface/trl/pull/2728#issuecomment-2635166424. TLDR: `SamplingParams` doesn't have  the same default as `GenerationConfig` so we would have a mismatch; default values in `SamplingParams` aren't necessarily good; `SamplingParams` is likely not json-serializable

qgallouedec: Note that you can still pass any additional argument with `generation_kwargs`

lewtun: Thanks, that's a nice approach

tests/test_vllm_client_server.py
albertvillanova, line 102: @qgallouedec I'm sorry but I'm not able to run this test.
Could you please give me some hint about the environment requirements so I can run it?
Thanks! 🤗

kashif: We might have to mock the response?

PR #4452, opened 2025-11-05
Support completion bootstrap for VLM in GRPO/RLOO

SolarWindRider:
## 1. Fix a little bug in GRPOTrainer.

When inputs are conversational, bootstrap is a list not str, causing a str concat error.

I add a bootstrap type check to fix the bug while maintain the original program

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1575,8 +1575,8 @@ class GRPOTrainer(BaseTrainer):
             completions = []
             for prompt, completion in zip(prompts, completions_text, strict=True):
                 bootstrap = prompt.pop()["content"] if prompt[-1]["role"] == "assistant" else ""
-                if isinstance(bootstrap, list):  # for VLM, the bootstrap might have the format [{"type": "text", "text": "..."}]
-                    assert len(bootstrap) == 1 and bootstrap[0]["type"] == "text":
+                if isinstance(bootstrap, list):  # for VLM, the format might be [{"type": "text", "text": "..."}]
+                    assert len(bootstrap) == 1 and bootstrap[0]["type"] == "text"
                     bootstrap = bootstrap[0]["text"]
                 completions.append([{"role": "assistant", "content": bootstrap + completion}])
         else:
--- a/trl/experimental/gfpo/gfpo_trainer.py
+++ b/trl/experimental/gfpo/gfpo_trainer.py
@@ -206,6 +206,9 @@ class GFPOTrainer(_GRPOTrainer):
             completions = []
             for prompt, completion in zip(prompts, completions_text, strict=True):
                 bootstrap = prompt.pop()["content"] if prompt[-1]["role"] == "assistant" else ""
+                if isinstance(bootstrap, list):  # for VLM, the format might be [{"type": "text", "text": "..."}]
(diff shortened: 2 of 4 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -1573,8 +1573,12 @@ def _generate_and_score_completions(
         completions_text = self.processing_class.batch_decode(completion_ids, skip_special_tokens=True)
         if is_conversational(inputs[0]):
             completions = []
-            for prompt, completion in zip(prompts, completions_text, strict=True):
-                bootstrap = prompt.pop()["content"] if prompt[-1]["role"] == "assistant" else ""
+            for prompt, completion in zip(prompts, completions_text):
```
albertvillanova: You should keep the change introduced by:
- #4183
```suggestion
            for prompt, completion in zip(prompts, completions_text, strict=True):
```

trl/trainer/grpo_trainer.py
```diff
@@ -1573,8 +1573,12 @@ def _generate_and_score_completions(
...
+                bootstrap = ""
+                if prompt[-1]["role"] == "assistant":
+                    bootstrap = prompt.pop()["content"] # bootstrap should be str
+                    if type(bootstrap) == list:
+                        bootstrap = bootstrap.pop()["text"]
```
albertvillanova: Maybe we can keep most of the previous code; additionally, it is recommended to use `isinstance` rather that `type` in this context: 
```suggestion
                bootstrap = prompt.pop()["content"] if prompt[-1]["role"] == "assistant" else ""
                if isinstance(bootstrap, list):
                    bootstrap = bootstrap[-1].get("text", "")
```

albertvillanova (changes requested):
I think you mixed 2 contributions in this PR:
- Fix bug in GRPOTrainer
- Add experimental RTPO

I would suggest you split your contributions in 2 PRs, and keep here only the bug fix.

SolarWindRider:
Thank you for your suggestion

SolarWindRider:
> I think you mixed 2 contributions in this PR:
> 
>     * Fix bug in GRPOTrainer
> 
>     * Add experimental RTPO
> 
> 
> I would suggest you split your contributions in 2 PRs, and keep here only the bug fix.

The problem is I don't know how to split it into 2 PR2 properly.

trl/trainer/grpo_trainer.py
```diff
@@ -1575,6 +1575,8 @@ def _generate_and_score_completions(
             completions = []
             for prompt, completion in zip(prompts, completions_text, strict=True):
                 bootstrap = prompt.pop()["content"] if prompt[-1]["role"] == "assistant" else ""
+                if isinstance(bootstrap, list):
+                    bootstrap = bootstrap[-1].get("text", "")
```
qgallouedec: let's be a bit more constrained to avoid silent bug

```suggestion
                if isinstance(bootstrap, list):  # for VLM, the bootstrap might have the format [{"type": "text", "text": "..."}]
                    assert len(bootstrap) == 1 and bootstrap[0]["type"] == "text":
                    bootstrap = bootstrap[0]["text"]
```

qgallouedec (approved):
LGTM thanks!

SolarWindRider:
Thank you guys! @qgallouedec @albertvillanova 
♪(･ω･)ﾉ

Issue #4374, opened 2025-10-30
Road to v1


qgallouedec:
Cette issue a pour but de referencer les taches nécessaires qui nous séparent de la v1. Cette list est evolutive et est modifiée en fonction des discussion et des avancées récentes

## Documentation

- [x] Remove `how_to_train.md` #4267 
- [x] Remove `using_llama_models.md` #4268
- [x] Remove `logging.md` #4269
- [ ] #4376
- [ ] #4375 
- [ ] #4377 
- [ ] #4378
- [ ] #4379 
- [ ] #4381
- [ ] #4382
- [ ] #4383
- [ ] #4384
- [ ] #4385
- [ ] #4386

## Addition to the codebase

- [x] Add accuracy reward to the `trl.rewards` module https://github.com/huggingface/trl/pull/4270
- [ ] Add an option (default to True) to use `RichProgressCallback` in scripts (`trl.scripts`).

## Remove from the codebase

- [ ] Remove `log_example_reports.py` #4241
- [ ] Remove `commands` directory: I don't think it's used anywhere. Slurm examples should be in the documentation or in `examples`.
- [ ] Remove `examples/research_projects`: the examples contained in this folder are most likely all obsolete. It is not wise to keep non-functional code in the directory. Furthermore, as the library now aims to exist beyond research, this type of code should be placed in separate directories, possibly with a link in our documentation.
- [ ] Remove `trl.extra.dataset_formatting`: the functions in this subbmodule aren't used anymore. I'd let a few version between deprecation and removal (3?) as they are in a lot of exmaples.
- [ ] #4387
- [ ] Remove `BestOfNSampler`. #4291 

## Modify the codebase

- [ ] Refactor `OnlineDPOTrainer` to aling with `GRPOTrainer`: This would allow easy maintenace.
- [ ] Move everything related to Mergekit to hte experimental submodule. I'm not seeing any usage for this part of the library, we may remove it in the future.
- [ ] Move everything related to Judges to `trl.experimental`. I'm not seeing any usage for this part of the library, we may remove it in the future.
- [ ] Move `masked_mean`, `masked_var` and `masked_whiten` to `ppo.py`: these functions are now only used in PPO, so we should move them there. I'd would either have no notice, or only a 1 version notice ("this function are now in `trl.trainer.ppo_trainer`, please import it from there").
- [ ] Move everything in`test/slow` to their respective trainer test files and rely on pytest slow marker instead.
- [ ] #4380

[...]

PR #4487, opened 2025-11-06
docs: change XPOTrainer import experimental.xpo, mark testXPOTrainer as low priority

Invidia19:
Hi! I’m still pretty new here, so please let me know if I’m stepping on anyone’s toes.
Following #4312 way of moving files to experimental, i think a few more stuff can be updated.

This small PR does the following:

- Updates `XPOTrainer` import to use `experimental.xpo`
- Marks `testXPOTrainer` as low priority

Contributes to #4374 
Fixes to #4485


qgallouedec:
thanks! good catch!

PR #4494, opened 2025-11-06
Add tiny model Qwen3VLForConditionalGeneration to CI

albertvillanova:
Add tiny model Qwen3VLForConditionalGeneration to CI.

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -21,6 +21,7 @@ import torch
 import transformers
 from accelerate.utils.memory import release_memory
 from datasets import load_dataset
+from packaging.version import Version
 from packaging.version import parse as parse_version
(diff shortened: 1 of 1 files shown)


albertvillanova:
We need to skip this test case for old `transformers` used in CI tests with minimum versions:
> ValueError: The checkpoint you are trying to load has model type `qwen3_vl` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.

albertvillanova:
- The Qwen3-VL series were introduced in `transformers-4.57.0`: https://github.com/huggingface/transformers/commit/c0dbe095b0fdceff933dbca7978ed76716d97fb2
- The CI tests with minimum versions use `transformers-4.56.1`

PR #4496, opened 2025-11-06
Add vLLM quantization option for colocate

sergiopaniego:
# What does this PR do?

Adds vLLM quantization option for colocate via `vllm_quantization` param.

TODO:

- [x] test it
- [x] extend to more online trainers

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@kashif

### DIFF
--- a/trl/experimental/gold/gold_config.py
+++ b/trl/experimental/gold/gold_config.py
@@ -304,10 +304,6 @@ class GOLDConfig(SFTConfig):
         default=None,
         metadata={"help": "Regex pattern used for vLLM guided decoding (optional)."},
     )
-    vllm_quantization: str | None = field(
-        default=None,
-        metadata={"help": "Quantization method to use for vLLM. If `None` (default), no quantization is applied."},
-    )
     vllm_sync_frequency: int = field(
         default=1,
         metadata={
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -489,10 +489,6 @@ class GRPOConfig(TrainingArguments):
         default=None,
         metadata={"help": "Regex for vLLM guided decoding. If `None` (default), guided decoding is disabled."},
     )
-    vllm_quantization: str | None = field(
-        default=None,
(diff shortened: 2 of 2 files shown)


sergiopaniego:
I've tested, and here's an example [running notebook](https://colab.research.google.com/drive/1bG3vBdKaAPUvuLcLns3bVKqbgaWk2lCv?usp=sharing).

It was raising an exception with `bnb_4bit_quant_storage=model_args.dtype,` so I've introduced a new param for passing it from command line. Additionally, by default that param is torch.uint8 in BitsAndBytes so just leaving it blank (None) would retrieve the same value.

sergiopaniego:
@kashif @qgallouedec @hmellor

qgallouedec:
In my understanding, it only works with `vllm_quantization="bitsandbytes"` right? and it requires the trained model to be loaded using `bitsandbytes`?

sergiopaniego:
Yes for both. Both models need to be quantized in order to work, hence opening the door for QLoRA here. If one of the models state is different (quantized/unquantized), it raises an exception when updating the vLLM weights during training.

We could even make it a boolean but I've used the current signature just to follow the used one in [vLLM](https://docs.vllm.ai/en/latest/features/quantization/bnb/)

qgallouedec:
What about this alternative approach: if the trained model is quantized, then pass `vllm_quantization="bitsandbytes"` to vLLM? maybe more user-friendly?

trl/trainer/grpo_trainer.py
```diff
@@ -596,6 +599,13 @@ def cast_outputs_to_original_dtype(module, args, output):
                     max_model_len = self.max_prompt_length + self.max_completion_length
                 else:
                     max_model_len = None
+
+                vllm_quantization = None
+                if is_bitsandbytes_available():
+                    for _, module in model.named_modules():
+                        if isinstance(module, bnb.nn.Linear4bit):
```
kashif, line 606:
```suggestion
                        if isinstance(module, (bnb.nn.Linear4bit, bnb.nn. Linear8bitLt)):
```

PR #4359, opened 2025-10-28
Add support for Trackio completions logging in GRPOTrainer

taha-yassine:
# What does this PR do?

This PR adds support for Trackio completions logging in GRPOTrainer alongside the already existing W&B support.
Currently Trackio doesn't support adding images to tables (opened https://github.com/gradio-app/trackio/issues/327 for tracking), so we skip them. The code is structured in such a way that adding that in the future would be straightforward.

Additionaly, `wandb_log_unique_prompts` also works for Trackio and I don't see any reason why it shouldn't. I propose we deprecate it and simply rename it to `log_unique_prompts` to avoid confusion.

The feature was tested using a modified version of `examples/scripts/grpo_vlm.py`.
image
image

I can also extend support to other trainer if the idea is accepted.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [x] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #4505, opened 2025-11-10
Add support for images inside tables with Trackio completions logging

taha-yassine:
# What does this PR do?

Follow up to https://github.com/huggingface/trl/pull/4359 now that support for images in tables in Trackio has landed.

Also sorry about the bug introduced in my last PR. I reused part of that code here to avoid code duplication, but this time it should work fine.

Also, since this requires the latest Trackio version to work correctly, what's the recommended way for pinning it on TRL's side?

Screenshot From 2025-11-10 18-55-52

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1939,6 +1939,12 @@ class GRPOTrainer(BaseTrainer):
                     self.num_completions_to_print,
                 )
 
+            logging_backends = []
+            if self.args.report_to and "wandb" in self.args.report_to and wandb.run is not None:
+                logging_backends.append(wandb)
+            if self.args.report_to and "trackio" in self.args.report_to:
+                logging_backends.append(trackio)
+
             table = {
                 "step": [str(self.state.global_step)] * len(self._logs["prompt"]),
                 "prompt": self._logs["prompt"],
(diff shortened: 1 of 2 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -1950,34 +1956,23 @@ def log(self, logs: dict[str, float], start_time: float | None = None) -> None:
...
-
-                if logging_backend == "trackio":
-                    if images_raw:
-                        # TODO: Implement once supported upstream https://github.com/gradio-app/trackio/issues/334
-                        logger.info("Skipping image logging for Trackio")
-                        df = df_base
-                        # images = []
-                        # for image_list in self._logs["images"]:
-                        #     images.append([trackio.Image(image) for image in image_list])
-                        # df = pd.concat([df_base, pd.Series(images, name="image")], axis=1, copy=False)
```
qgallouedec, line 1976: I think you just need to uncomment this

taha-yassine: This would lead to duplicated code since the part under `if logging_backend == "wandb":` is exactly the same as the one under `if logging_backend == "trackio":`, but if you think it's a better solution I'll do it that way

qgallouedec:
> Also, since this requires the latest Trackio version to work correctly, what's the recommended way for pinning it on TRL's side?

Trackio is still very alpha, so IMO versioning isn’t a major concern for now, as most users are likely to stay up-to-date or even use the dev version.

qgallouedec:
The runner isn't taking the job right now, I'll merge without the CI being ran: the CI doesn't test the tracking anyway

PR #4514, opened 2025-11-12
Rename `flash-attn` to `flash-attn2`

qgallouedec:
In kernels, flash-attn was renamed flash-attn2, see https://github.com/huggingface/kernels/issues/179#issuecomment-3503099190


trl/trainer/sft_trainer.py
```diff
@@ -73,8 +73,9 @@
     "flash_attention_2",
     "flash_attention_3",
     "kernels-community/flash-attn",
```
sergiopaniego: Can we remove it?
```suggestion
```

qgallouedec: I guess we can. It's not working anyway

Issue #2897, opened 2025-02-18
`GRPO` completions skip special tokens ?


MohamedAliRashad:
So i was trying to add some special tokens to my model but i noticed that they never show in completions and this sucks because I rely on them in my reward function.

Can there be a flag that disables skiping special tokens ?

NJordan72:
I think vLLM skips special tokens by default.  `skip_special_tokens=False` in SamplingParams would probably do the trick.

nopepper:
I think this should be False by default, special tokens can be part of the reward computation (especially when you have custom special tokens in the training pipeline)

MohamedAliRashad:
@0x404 That's what i am doing right now but i think an argument for it in `GRPOConfig` will be better

MohamedAliRashad:
@nopepper I see the same

unlikezy:
I encounter the same issue here, looking forward to the PR..

Fransou:
Hey, I am having the same issue, so I used the following temporary workaround:

```
from tokenizers import Tokenizer
from copy import deepcopy


def wrap_tokenizer(tokenizer: Tokenizer) -> Tokenizer:
    """
    Wraps a normal tokenizer to make sure its batch_decode does not skip special tokens
    """
    new_tokenizer = deepcopy(tokenizer)

    old_batch_decode = new_tokenizer.batch_decode

    def batch_decode(*args, **kwargs):
        """
        Batch decode sequences
        """
        kwargs["skip_special_tokens"] = False
        return old_batch_decode(*args, **kwargs)

    new_tokenizer.batch_decode = batch_decode
    return new_tokenizer
```

The new tokenizer obtained using this function will systematically override the "skip_special_tokens" parameter if it is part of the kwargs.

Serzhanov:
still waiting for PR

qgallouedec:
closing this one, that we decided not to address. see this discussion https://github.com/huggingface/trl/pull/4521

PR #4521, opened 2025-11-13
Make `skip_special_tokens` configurable

taha-yassine:
# What does this PR do?

Fixes https://github.com/huggingface/trl/issues/2897

As mentioned in the original issue, special tokens can be especially useful when computing reward functions.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -102,6 +102,8 @@ class GRPOConfig(TrainingArguments):
             parameter is only effective when `use_vllm` is set to `False`.
         cache_implementation (`str`, *optional*):
             Implementation of the cache method for faster generation when `use_vllm` is set to `False`.
+        skip_special_tokens (`bool`, *optional*, defaults to `True`):
+            Whether to skip special tokens when decoding completions. This affects both reward computation and logging.
 
         > Parameters that control generation acceleration powered by vLLM
 
@@ -450,6 +452,13 @@ class GRPOConfig(TrainingArguments):
     cache_implementation: str | None = field(
         default=None,
         metadata={"help": "Implementation of the cache method for faster generation when use_vllm is set to False."},
+    )
+    skip_special_tokens: bool = field(
+        default=True,
+        metadata={
+            "help": "Whether to skip special tokens when decoding completions. This affects both reward computation "
+            "and logging."
+        },
     )
 
     # Parameters that control generation acceleration powered by vLLM
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -534,6 +534,7 @@ class GRPOTrainer(BaseTrainer):
         self.log_completions = args.log_completions
         self.log_unique_prompts = args.log_unique_prompts
         self.num_completions_to_print = args.num_completions_to_print
+        self.skip_special_tokens = args.skip_special_tokens
         # Keep logs sized to the generation batch to record only outputs from the latest model update.
         self._logs = {
             "images": deque(maxlen=args.generation_batch_size),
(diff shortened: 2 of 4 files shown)


qgallouedec:
Thanks for the PR. I went through the issue. I understand that the motivation is to compute a reward that depends on the special tokens.
Would computing the reward based on the completion_ids (ints) instead of on the completion (strs) would do the trick?

taha-yassine:
> Would computing the reward based on the completion_ids (ints) instead of on the completion (strs) would do the trick?

I think `completion_ids` is useful when the reward depends on whether or not some specific predetermined tokens are present. For example, it can be used to check if the policy respects a given answer format.
Now say we're training a policy to answer questions and to put the result between `<answer>` and `</answer>`, and say these are special tokens. `completion_ids` would allow us to detect whether the completion contains those tokens, but we won't be able to precisely locate what part of `completion` they surround. Also, we can't manually decode `completion_ids` because we don't have access to the tokenizer inside the reward functions.

taha-yassine:
Hi @qgallouedec , just checking in to see if you've had a chance to look at my last comment. Lmk what you think.

qgallouedec:
Actually, I'm still wondering about this feature. In your example, in my opinion, `<answer>` shouldn't be a special token in the first place. In most models, these kind of tokens are not special: 

- https://huggingface.co/Qwen/Qwen3-8B/blob/main/tokenizer_config.json#L203
- https://huggingface.co/deepseek-ai/DeepSeek-V3.2-Exp/blob/main/tokenizer_config.json
- https://huggingface.co/moonshotai/Kimi-K2-Thinking/blob/main/tokenizer_config.json#L129

Another difficulty I foresee is that we will soon be relying on "response parsing" (see #4300) instead of simple decoding. And in that case, I don't even see how we  could include special tokens; in fact, I don't think this question would even make sense.

I recommend the following:
- re-consider the choice of making a token "special" (in the vast majority of cases, only BOS, EOS and PAD should be special), or
- have a tokenizer in the reward function:
  ```python
  tokenizer = AutoTokenizer.from_pretrained(...)

  def reward_func(completion_ids, **kwargs):
      text = tokenizer.decode(completion_ids, skip_special_tokens=False)
      ...
      return rewards
  ```

taha-yassine:
I agree that it doesn't make sense to have these kinds of tokens marked as special. The issue is that some models still do that ([example](https://huggingface.co/PleIAs/Baguettotron/blob/main/tokenizer_config.json#L57)). And for some more context, the model I struggled with the most and that led me to open this PR was gpt-oss because of their harmony response format. Extracting the content of each channel requires access to the special tokens.
 
> Another difficulty I foresee is that we will soon be relying on "response parsing" (see #4300) instead of simple decoding. And in that case, I don't even see how we could include special tokens; in fact, I don't think this question would even make sense.

I wasn't aware of this new response parsing feature. It seems useful and it would indeed solve a lot of the headache around extracting the different parts of the response for reward design. My only concern is the time it would take for it to make it into TRL and for general adoption by past and future models.

On the other hand, what issues do you see with this PR in its current?

In the mean time, while having a globally available tokenizer that one could use for rewards isn't the most elegant of solutions, I guess it's an acceptable workaround.

qgallouedec:
Got it. For now, we can’t extract the content of each channel in the case of GPT-OSS, but this will become possible soon thanks to the response schema, [see how we use it with tools](https://github.com/huggingface/trl/blob/e002a25f005741753312886ac754c5adf274b1ce/trl/chat_template_utils.py#L20). Given that, I don’t think adding a new parameter, as suggested in this PR, is the right solution. I want to reiterate that using a tokenizer inside the reward function is a more appropriate approach in the meantime. I’m closing this PR, and thanks again for the contribution!

PR #4522, opened 2025-11-13
Fix Docker images for Liger

lewtun:
# What does this PR do?

Liger kernels require a `gcc` compiler, which the `runtime` image from `torch` doesn't contain. This PR switches to the `devel` image to enable Liger. I've also removed `hf_transfer` as a dep, as this is deprecated in favour of `xet`

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/docker/trl-dev/Dockerfile
+++ b/docker/trl-dev/Dockerfile
@@ -1,5 +1,5 @@
-FROM pytorch/pytorch:2.8.0-cuda12.8-cudnn9-runtime
+FROM pytorch/pytorch:2.8.0-cuda12.8-cudnn9-devel
 RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


docker/trl-dev/Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM pytorch/pytorch:2.8.0-cuda12.8-cudnn9-runtime
+FROM pytorch/pytorch:2.8.0-cuda12.8-cudnn9-devel
```
lewtun, line 1: Note that if we want to use `vllm` in the images for e.g. GRPO, a better approach is to use the official vllm image like this:

```suggestion
FROM vllm/vllm-openai:v0.10.2
```

qgallouedec: Nice, I don't know what to think of it right now, I've opened #4523 the meantime

Issue #3729, opened 2025-07-12
GRPOTrainer.__init__() got an unexpected keyword argument 'compute_metrics'


CHOUDDIDIBUYAOSONG:
### Reproduction

def compute_metrics(eval_preds):
    # eval_preds 是一个元组 (predictions, label_ids)
    predictions, label_ids = eval_preds

    # predictions 是多个生成结果，需要提取其中的最终内容
    if isinstance(predictions, list):  # list of list of dicts
        extracted_preds = [extract_answer_from_model_output(pred[0]['content']) for pred in predictions]
    else:
        extracted_preds = predictions

    # 提取 ground truth
    true_labels = [extract_answer_from_dataset(a) for a in label_ids]

    acc = accuracy_score(true_labels, extracted_preds)

    # print("\n📊 验证集分类报告:")
    # print(classification_report(true_labels, extracted_preds, digits=4))

    return {
        "eval_accuracy": acc,
    }

from trl import GRPOConfig, GRPOTrainer

train_args = GRPOConfig(
    # output_dir="outputs/qwen1.5b-grpo-lora-fp16",
    fp16=True,
    deepspeed="ds_config.json",
    per_device_train_batch_size=16,
    per_device_eval_batch_size=16,
    gradient_accumulation_steps=2,
    learning_rate=2e-4,
    num_train_epochs=3,
    lr_scheduler_type="cosine",
    warmup_ratio=0.05,
    max_grad_norm=0.3,
    logging_steps=10,
    save_steps=5,
    output_dir=OUTPUT_DIR,
    report_to=["wandb"],
    max_prompt_length=1024,
    max_completion_length=156,
    num_generations=8,
    use_vllm=False,
    eval_steps=10,  # 每100步评估一次
    eval_strategy="steps",
    save_strategy="steps",
    # metric_for_best_model="eval_accuracy",
    # save_total_limit=1,
    # compute_metrics=compute_metrics,
)

trainer = GRPOTrainer(
    model=model,
    processing_class=tokenizer,
    reward_funcs=[
        accuracy_reward,
        format_reward
    ],
    args=train_args,
    train_dataset=dataset["train"],
    eval_dataset=dataset["test"],
    compute_metrics=compute_metrics,
)

### System Info

trl==0.16.0

### Checklist

- [x] I have checked that my issue isn't already filed (see [open issues](https://github.com/huggingface/trl/issues?q=is%3Aissue))
- [x] I have included my system information
[...]

Issue #2959, opened 2025-02-26
`compute_metrics` in `GROPTrainer`


dipta007:
### Feature request

Can you please add `compute_metrics` in the `GROPTrainer`?

I can see its available in other, i.e., [DPOTrainer](https://huggingface.co/docs/trl/v0.15.2/dpo_trainer#trl.DPOTrainer.compute_metrics)

### Motivation

1. We can compute accuracy or downstream eval metrics over the evaluation dataset
2. That will help in more fine-grained logging
3. Better metric for `metric_for_best_model`

### Your contribution

I am open to submit a PR

qgallouedec:
Hey @dipta007, sorry for the very slow reply here, we're catching up on the backlog. This is a real gap, and a plain passthrough wouldn't have fixed it: `prediction_step` returns `(loss, None, None)` https://github.com/huggingface/trl/blob/c061050d3c4eb62dda345adf6f1a4e892619adec/trl/trainer/grpo_trainer.py#L3292-L3298
so the eval loop never has predictions or labels to hand your function: it would be silently ignored rather than error.

One note on your third motivation: using `metric_for_best_model` with reward metrics like `eval_reward` was a separate bug, and that one is fixed as of v1.5.0 (see #3384), so that part works today.

The main open question is what exactly `EvalPrediction` should carry (in particular how reference answers reach your metric), and input from you on that would be welcome.

JulesRoussel2001:
Hi @qgallouedec, I have a working implementation of this in #5790 (currently sitting at the first-time-contributor CI approval gate), and since you asked for input on what EvalPrediction should carry, here is the design I used and what I learned from using it in a real experiment.

[...]

PR #4534, opened 2025-11-17
Add `compute_metrics` parameter for `GRPOTrainer`

colinzhaoxp:
# Add `compute_metrics` parameter for `GRPOTrainer`

This is my first open-source PR contribution, I would greatly appreciate any feedback on areas for improvement. Please don't hesitate to suggest changes - I'm eager to learn and make this contribution as good as possible!

## What does this PR do?
This PR adds `compute_metrics` parameter for `GRPOTrainer`, which is already supported by `Trainer`. We can compute accuracy or downstream eval metrics over the evaluation dataset

## Fixes related issues
https://github.com/huggingface/trl/issues/3729
https://github.com/huggingface/trl/issues/2959

## Changes Made
Added `compute_metrics` parameter to `GRPOTrainer`
`trl/trainer/grpo_trainer.py`

Added a new optional parameter after num_generations:

```python
from transformers.trainer_utils import seed_worker, EvalLoopOutput

class GRPOTrainer(BaseTrainer):
    """
    ...
    The function that will be used to compute metrics at evaluation. Must take a [`EvalPrediction`] and return
    a dictionary string to metric values. *Note* When passing TrainingArgs with `batch_eval_metrics` set to
    `True`, your compute_metrics function must take a boolean `compute_result` argument. This will be triggered
    after the last eval batch to signal that the function needs to calculate and return the global summary
    statistics rather than accumulating the batch-level statistics
    ...
    """
    def __init__(
        self,
        ...
        compute_metrics: Callable[[EvalLoopOutput], dict] | None = None,
        ...
    ):
    ...
    super().__init__(
        ...
        compute_metrics=compute_metrics,
        ...
    )

```

## Example Usage
```python
def my_eval_function(eval_predict):
       pass

trainer = GRPOTrainer(
    reward_funcs=reward_func,
    args=args,
    train_dataset=train_dataset,
    eval_dataset=eval_dataset,
    compute_metrics=my_eval_function,
)
# trainer.train() # evaluation during training
trainer.evaluate() # or directly evaluate your model.
```
More examples are available in this [blog](https://huggingface.co/docs/evaluate/transformers_integrations#trainer)

## Benefits
- Flexible: Users can choose their own function to evaluate their model during the trianing.

## Who can review?
Any community member is welcome to provide feedback.
As my first open-source contribution, I'm excited to learn - please don't hesitate to suggest any enhancements!


colinzhaoxp:
@kashif @qgallouedec @burtenshaw

qgallouedec:
thanks! I don't think it would work just like this though. Because in `Trainer`, `compute_metrics` is called only if `compute_loss` returns logits and labels. And in GRPO, it's not clear what would be the labels? Consequently, `compute_loss` doesn't support `return_output` 

https://github.com/huggingface/trl/blob/f7ac9741d43528cf5050ba3ecd41cffec34c5a1a/trl/trainer/grpo_trainer.py#L1790-L1792

You can run this code, and see that `my_eval_function` is never called:

```python

dataset = load_dataset("trl-internal-testing/zen", "conversational_prompt_only")

def my_eval_function(eval_predict):
    print(eval_predict)


    return [len(set(c[0]["content"])) for c in completions]

trainer = GRPOTrainer(
    model="Qwen/Qwen3-0.6B",
    train_dataset=dataset["train"],
    eval_dataset=dataset["test"],
    args=GRPOConfig(
        eval_steps=2,
        eval_strategy="steps",
    ),
    compute_metrics=my_eval_function,
)
trainer.train()
```

colinzhaoxp:
Thanks for your reply! 

Yes, the demo above given by me is not the minimal runnable. I just want to explain the how to add `compute_metric` to `GPROTrain`. Just like your comments,  if want to normally use this function, we need the the `compute_loss` returns logits and labels, and some other changes to fit.

So the next thing I should do is to give a minimal runnable demo? Because I think add other code, like change the code `compute_loss`, is not reasonable. I need your suggestions to improve this PR.

By the way, I will firstly provide a minimal runnable demo.

Thanks

albertvillanova:
Hi @colinzhaoxp — sorry for the long delay here, and thank you for such a clean first PR! 

Unfortunately, adding `compute_metrics` to `__init__` doesn't actually enable it: GRPO's `prediction_step` returns `(loss, None, None)`, so the eval loop never has predictions/labels to pass to your function — it would be silently ignored rather than raising, which is a worse trap than the current error. Making it genuinely work means surfacing the generated completions as `predictions` (and reference answers as `labels`) inside `prediction_step`, and the same change would have to land in `rloo_trainer.py`, which has the identical structure. Also the annotation should be `Callable[[EvalPrediction], dict]`, not `EvalLoopOutput`. 

Given the size of the real fix I'll close this for now, but the feature is genuinely wanted (#2959) — please don't hesitate to open a follow-up that surfaces predictions in the eval loop, and I'm happy to help review it!

PR #4558, opened 2025-11-22
Move `WinRateCallback` to experimental

qgallouedec:
- Usage is very low for this callback, and 
- It doesn't support fast generation
- And the design isn't very good, as you need to pass the trainer as argument, which is a bit circular

I don't think it's worth having it in the stable part of TRL. I suggest to move it to experimental

Part of #4374

### DIFF
--- a/trl/trainer/callbacks.py
+++ b/trl/trainer/callbacks.py
@@ -266,14 +266,14 @@ def _win_rate_completions_df(
 
 
 class WinRateCallback(_WinRateCallback):
-    def __post_init__(self):
+    def __init__(self, *args, **kwargs):
         warnings.warn(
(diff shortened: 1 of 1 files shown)


trl/trainer/callbacks.py
```diff
@@ -262,196 +265,15 @@ def _win_rate_completions_df(
...
-                    prompts=prompts,
-                    completions=completions,
-                    winner_indices=winner_indices,
-                )
-                log_table_to_comet_experiment(
-                    name="win_rate_completions.csv",
-                    table=df,
-                )
+class WinRateCallback(_WinRateCallback):
+    def __post_init__(self):
```
kashif: since this is not a dataclass this will not be called so changing to `__init__`

PR #4270, opened 2025-10-13
Add accuracy reward

pramodith:
Ports the `accuracy_reward` function from [open-r1](https://github.com/huggingface/open-r1/blob/main/src/open_r1/rewards.py) to `trl.rewards`.

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [X] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [X] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [X] Did you make sure to update the documentation with your changes?
- [X] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #4563, opened 2025-11-23
Reasoning reward

lewtun:
# What does this PR do?

This PR introduces a `reasoning_accuracy_reward()` function which:

* Removes the reasoning content
* Computes math accuracy on the final answer (i.e. everything after the closing `</think>` tag)

This is useful because reasoning models often have very long reasoning content that can bias the grader (e.g. partial, correct solutions, but not those given as the model's final answer).

If the reasoning content is not complete, we assign zero reward to penalise the model to generate more concise answers. Here's some examples on how this compares with the existing `accuracy_reward()` function (generated by Codex):

```py
from trl.rewards import accuracy_reward, reasoning_accuracy_reward

examples = [
    {
        "label": "1) Complete reasoning, correct final answer → both reward 1",
        "completions": [
            [
                {
                    "role": "assistant",
                    "content": (
                        "<think> Reasoning here. </think> "
                        "The final answer is \\boxed{\\frac{1}{3}}"
                    ),
                }
            ]
        ],
        "solutions": [r"\frac{1}{3}"],
    },
    {
        "label": "2) Complete reasoning, wrong final answer → both reward 0",
        "completions": [
            [
                {
                    "role": "assistant",
                    "content": (
                        "<think> Reasoning here. </think> "
                        "The final answer is \\boxed{\\frac{1}{2}}"
                    ),
                }
            ]
        ],
        "solutions": [r"\frac{1}{3}"],
    },
    {
        "label": (
            "3) Correct answer appears in unfinished reasoning → "
            "accuracy_reward=1, reasoning_accuracy_reward=0"
        ),
        "completions": [
            [
                {
                    "role": "assistant",
                    "content": (
                        "<think> I got \\boxed{\\frac{1}{3}} but I never close </think> "
                        "(no explicit final answer section)."
                    ),
                }
            ]
        ],
        "solutions": [r"\frac{1}{3}"],
    },
    {
        "label": (
            "4) Reasoning contains correct intermediate box, "
            "final answer after </think> is wrong → accuracy_reward=1, reasoning_accuracy_reward=0"
        ),
        "completions": [
            [
                {
                    "role": "assistant",
                    "content": (
                        "<think> Intermediate step gives \\boxed{\\frac{1}{3}}. </think> "
                        "Therefore the final answer is bonkers"
                    ),
                }
            ]
        ],
        "solutions": [r"\frac{1}{3}"],
    },
]

reasoning_delimiters = ["</think>"]

for ex in examples:
    acc = accuracy_reward(ex["completions"], ex["solutions"])
    reasoning_acc = reasoning_accuracy_reward(
        ex["completions"], ex["solutions"], reasoning_delimiters=reasoning_delimiters
    )
    print(f"{ex['label']}\n  accuracy_reward -> {acc[0]}\n  reasoning_accuracy_reward -> {reasoning_acc[0]}\n")

# 1) Complete reasoning, correct final answer → both reward 1
#   accuracy_reward -> 1.0
#   reasoning_accuracy_reward -> 1.0

# 2) Complete reasoning, wrong final answer → both reward 0
#   accuracy_reward -> 0.0
#   reasoning_accuracy_reward -> 0.0

# 3) Correct answer appears in unfinished reasoning → accuracy_reward=1, reasoning_accuracy_reward=0
#   accuracy_reward -> 1.0
#   reasoning_accuracy_reward -> 0.0

# 4) Reasoning contains correct intermediate box, final answer after </think> is wrong → accuracy_reward=1, reasoning_accuracy_reward=0
#   accuracy_reward -> 1.0
#   reasoning_accuracy_reward -> 0.0
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


trl/rewards/accuracy_rewards.py
```diff
@@ -70,10 +74,119 @@ def accuracy_reward(completions: list[list[dict[str, str]]], solution: list[str]
...
+            )
+            # Compute binary rewards if verifiable, `None` otherwise to skip this example
+            try:
+                reward = float(verify(gold_parsed, answer_parsed))
+            except Exception as e:
+                logger.warning(f"Verification failed with error: {e}. Assigning reward=None for this example.")
+                reward = None
         else:
-            # If the gold solution is not parseable, we assign `None` to skip this example
-            reward = float(content.strip().lower() == sol.strip().lower())
```
lewtun, line 76: I don't think this is desired behaviour, so have opted to return `None` if the gold answer cannot be parsed. LMK if you disagree

trl/rewards/accuracy_rewards.py
```diff
@@ -70,10 +74,119 @@ def accuracy_reward(completions: list[list[dict[str, str]]], solution: list[str]
...
+                reward = None
+        else:
+            # If the gold solution cannot be parsed, we assign `None` to skip this example
+            reward = None
+        rewards.append(reward)
+
+    return rewards
+
+
+def reasoning_accuracy_reward(
```
lewtun, line 81: We could technically merge this logic with `accuracy_reward()`, but for now I kept it separate to avoid bundling too many things together. Happy to refactor if you prefer it all in one function.

trl/rewards/accuracy_rewards.py
```diff
@@ -12,25 +12,29 @@
...
 
+logger = logging.getLogger(__name__)
+
 if is_math_verify_available():
     from latex2sympy2_extended import NormalizationConfig
     from math_verify import LatexExtractionConfig, parse, verify
 
 
-def accuracy_reward(completions: list[list[dict[str, str]]], solution: list[str], **kwargs) -> list[float | None]:
+def accuracy_reward(completions: list[list[dict[str, str]]], solutions: list[str], **kwargs) -> list[float | None]:
```
lewtun: Note to selves: this should technically be called a `math_reward()` function since it's math-specific

qgallouedec:
nice, can you add it to https://github.com/huggingface/trl/blob/4cb1a2567c25dfc89a1a3e64c538f31e7068eb83/trl/scripts/grpo.py#L55-L59 and https://github.com/huggingface/trl/blob/4cb1a2567c25dfc89a1a3e64c538f31e7068eb83/trl/scripts/rloo.py#L52-L56

trl/rewards/accuracy_rewards.py
qgallouedec: lgtm!

trl/rewards/accuracy_rewards.py
```diff
@@ -70,10 +74,119 @@ def accuracy_reward(completions: list[list[dict[str, str]]], solution: list[str]
                 extraction_mode="first_match",
             )
             # Compute binary rewards if verifiable, `None` otherwise to skip this example
-            reward = float(verify(gold_parsed, answer_parsed))
+            try:
+                reward = float(verify(gold_parsed, answer_parsed))
+            except Exception as e:
+                logger.warning(f"Verification failed with error: {e}. Assigning reward=None for this example.")
+                reward = None
```
qgallouedec: do you have an example that would raise an exception here 

```python
float(verify(gold_parsed, answer_parsed))
```

?

Since both are parsed, I don't think it can fail.

qgallouedec:
and here as well https://github.com/huggingface/trl/blob/4cb1a2567c25dfc89a1a3e64c538f31e7068eb83/docs/source/rewards.md

trl/rewards/accuracy_rewards.py
qgallouedec: I'm not sure about this renaming, a lot of datasets use `"solution"`, and not `"solutions"`

qgallouedec: > Note to selves: this should technically be called a math_reward() function since it's math-specific

I'm fine with the renaming 👍, it was added recently in https://github.com/huggingface/trl/pull/4270, so I think we can just rename without deprecation notice

lewtun: > I'm not sure about this renaming, a lot of datasets use "solution", and not "solutions"

Good catch, fixed in https://github.com/huggingface/trl/pull/4563/commits/efc68b90b4783bf6ac5cc9876ff11e6edfd371d3

edbeeching (approved):
LGTM, nice test coverage. I couldn't think of any missing edge cases.

trl/rewards/accuracy_rewards.py
lewtun: > I'm fine with the renaming 👍, it was added recently in https://github.com/huggingface/trl/pull/4270, so I think we can just rename without deprecation notice

I'll do this in a follow-up PR to avoid making this one huge

trl/rewards/accuracy_rewards.py
lewtun: Good point, I was following some old logic we had in `open-r1`. I checked a few edge cases and indeed even if the parser returns `[]` for the `answer_parsed` we still get 0 reward and no exception. One can trigger the exception with `raise_on_error=True` inside `verify`, but that doesn't seem super helpful for RL

docs/source/lora_without_regret.md
```diff
@@ -135,94 +135,6 @@ The blog post performs GRPO on a range of models and datasets from the Hub, and
 
 For reinforcement learning, the blog uses a math reasoning task that we can reproduce as a Python function.
 
-<details>
-<summary>Reward function</summary>
-
-```python
-def strip_reasoning_accuracy_reward(
```
lewtun, line 142: For viz @burtenshaw a close variant of this reward function is now native in TRL as `reasoning_accuracy_reward()`

tests/test_rewards.py
```diff
@@ -117,27 +117,76 @@ def test_accuracy_reward_unparseable_gold(self):
         """Test accuracy_reward with an unparseable gold solution."""
         completion = [
             [{"content": "Answer is forty two."}],
-            [{"content": "Some other content."}],
```
lewtun, line 120: Removed these cases as they weren't testing what was advertised :)

lewtun:
@qgallouedec I've added the reward function to the registry and docs. I'll follow up with a new PR to handle the renaming of the reward functions

lewtun:
> Out of curiosity, why use a list of strings for delimiters, rather than a single string? Do you have an example where the end of the reasoning can be represented by more than one delimiter?

What I had in mind was scenarios where a user wants to RL a variety of different reasoning models within the same script. Then they can define e.g. `reasoning_delimiters=["</think>", "[/THINK]"]` to RL both Qwen/DeepSeek models and Magistral without needing to change their code each time.

PR #4564, opened 2025-11-23
Add `shuffle_dataset` option to `SFTTrainer`

qgallouedec:
# What does this PR do?

Adds an option to shuffle the dataset in SFT

```python
SFTConfig(
    shuffle_dataset=True,
)
```

Code example with packing:

```python
from datasets import Dataset

dataset = Dataset.from_dict(
    {
        "input_ids": [
            [1],
            [2, 2],
            [3, 3, 3],
            [4, 4, 4, 4],
            [5, 5, 5, 5, 5],
            [6, 6, 6, 6, 6, 6],
        ],
    }
)

trainer = SFTTrainer(
    model="Qwen/Qwen3-0.6B",
    train_dataset=dataset,
    args=SFTConfig(
        packing=True,
        max_length=4,
        packing_strategy="bfd",  # or wrapped
        shuffle_dataset=False,  # or True
    ),
)
print(trainer.train_dataset["input_ids"])
```

### Before

Longest sequences are placed first:

```txt
Column([[4, 4, 4, 4], [5, 5, 5, 5], [6, 6, 6, 6], [3, 3, 3, 1], [2, 2]])
```

### After with `shuffle_dataset=True`

The same bins are used, but their contents are interleaved:

```txt
Column([[2, 2], [5, 5, 5, 5], [3, 3, 3, 1], [6, 6, 6, 6], [4, 4, 4, 4]])
```

### Before

Sequences are packed strictly in input order:

```txt
Column([[1, 2, 2, 3], [3, 3, 4, 4], [4, 4, 5, 5], [5, 5, 5, 6], [6, 6, 6, 6]])
```

### After with `shuffle_dataset=True`

Input order is ignored and sequences are mixed:

```txt
Column([[6, 5, 5, 5], [6, 6, 6, 6], [1], [5, 5, 2, 2], [3, 3, 3, 6]])
```


MilkClouds:
I think the name shuffle_dataset is bit misleading. I thought that trl wasn't shuffling data on training when I see this on 0.26 patch note. How about shuffle_dataset_on_packing or shuffle_dataset_before_packing?

qgallouedec:
I think you're making a mistake here. Packing or not, if you enable this `shuffle_dataset`, the dataset is shuffled.

MilkClouds:
@qgallouedec I mean, the argument name is confusing with `DistributedSampler(shuffle=True)`. I did and other users may be confused with "is this argument for enable/disabling sampler's shuffle?"

qgallouedec:
Ok I see what you mean. In my opinion, this is unintended behavior from the `transformers.Trainer`, which should not use `RandomSampler` by default, but rather `SequentialSampler`.

PR #4567, opened 2025-11-24
Support async reward functions and parallelize call to reward functions.

### LINKED ISSUE #4130: Async Support for Reward Functions
### Feature request

Add support in TRL for async reward functions so users can run batched external API calls (e.g. OpenAI/Deepseek or local inference) with asyncio.gather when computing rewards in the GRPO trainer.

### Motivation

I am creating my reward functions for the GRPO trainer to have api calls to deepseek/openai (In fact this could be useful for any api calls, even for local servers/inference if your node setup supports it)

whenever I want to call my reward function over my batch, I need to do so in an async manner. 

One Example of how I want to use it: 

async def check_content(prompts, completions, **kwargs) -> list[float]:
    # I put 1 to get the user content I do not care about the system content
    prompts = [prompt[1]['content'] for prompt in prompts] 
    client = OpenAI(api_key=deepseek_api_key, base_url="https://api.deepseek.com")
    async def call_model(prompt: str, completion:str) -> int:
        """Call Deepseek API to get relevance score, for one prompt-completion pair."""
        response = await client.chat.completions.create(
            model="deepseek-chat",
            messages=[
                {"role": "system", "content": "You are a judge ...."},
                {"role": "user", "content": f"This is the prompt: {prompt}"},
                {"role": "user", "content": f"This is the answer: {completion}"},
            ],
            stream=False
        )
        ....
    tasks = []
    for prompt, completion in zip(prompts, completions):
        tasks.append(call_model(prompt, completion[0]['content']))
    scores = await asyncio.gather(*tasks)
    return scores
    

### Your contribution

I can do a PR, which I fixed it locally if people are interested

pramodith:
# What does this PR do?
1. Checks if a reward function is not a `nn.Module`
2. If not checks if the reward function is synchronous or async.
3. Execution of synchronous reward functions are parallelized by running each on the next available thread.
4. Execution on non-async reward functions are parallelized by asyncio io's current event loop.
5. We await until all the created tasks are complete.
6. To avoid conflicts with any existing event loop that the code is running under we check if we're in an active event-loop if so we create a new event loop and run all the reward functions under the new event loop.

Fixes #4130 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/docs/source/grpo_trainer.md
+++ b/docs/source/grpo_trainer.md
@@ -559,10 +559,6 @@ async def async_reward_func(prompts, completions, **kwargs):
     return [1.0 if completion else 0.0 for completion in completions]
 ```
 
-### Rapid Experimentation for GRPO
-
-RapidFire AI is an open-source experimentation engine that sits on top of TRL and lets you launch multiple GRPO configurations at once, even on a single GPU. Instead of trying configurations sequentially, RapidFire lets you **see all their learning curves earlier, stop underperforming runs, and clone promising ones with new settings in flight** without restarting. For more information, see [RapidFire AI Integration](rapidfire_integration).
-
 #### Passing the reward function to the trainer
 
 To use your custom reward function, pass it to the [`GRPOTrainer`] as follows:
@@ -590,6 +586,10 @@ trainer = GRPOTrainer(
 and the reward will be computed as the sum of the rewards from each function, or the weighted sum if `reward_weights` is provided in the config.
 
 Note that [`GRPOTrainer`] supports multiple reward functions of different types. See the parameters documentation for more details.
+
+### Rapid Experimentation for GRPO
+
+RapidFire AI is an open-source experimentation engine that sits on top of TRL and lets you launch multiple GRPO configurations at once, even on a single GPU. Instead of trying configurations sequentially, RapidFire lets you **see all their learning curves earlier, stop underperforming runs, and clone promising ones with new settings in flight** without restarting. For more information, see [RapidFire AI Integration](rapidfire_integration).
 
 ## Agent Training
 


pramodith:
On revision I think it might be more prudent to use `ThreadPoolExecutor` for synchronous reward functions since these are going to be CPU bound/compute heavy, running them in an async fashion might not lead to true parallelism. Happy to hear your thoughts on this.

qgallouedec:
> On revision I think it might be more prudent to use `ThreadPoolExecutor` for synchronous reward functions since these are going to be CPU bound/compute heavy, running them in an async fashion might not lead to true parallelism. Happy to hear your thoughts on this.

I think I agree with this. I don't have a big background with asyncio though

pramodith:
Done! Runs each synchronous reward function in its own thread now.

qgallouedec:
I think the implementation looks good.

However, I'm still wondering if we should constraint the reward function to be synchronous. If execution is parallelized, we should achieve the same performance, no?

pramodith:
> I think the implementation looks good.
> 
> However, I'm still wondering if we should constraint the reward function to be synchronous. If execution is parallelized, we should achieve the same performance, no?

There aren't any constraints the reward function can sync or async. Sync functions use ThreadPoolExecutor, async functions use async.gather.

Perhaps I'm misunderstanding you?

```
if asyncio.iscoroutinefunction(reward_func):
          async_funcs_info.append((i, reward_func, reward_func_name))
else:
    sync_funcs_info.append((i, reward_func, reward_func_name))

```

The test case added also tests for this by including one sync and one async function.

qgallouedec:
Sorry for the delay but it took me some time to find a clear way to transcript my mind:

Currently we have something like this:

```python
import time


def reward_length(texts):
    time.sleep(0.5)
    return [float(len(t)) for t in texts]


def reward_vowel_count(texts):
    time.sleep(0.5)
    vowels = set("aeiouAEIOU")
    return [float(sum(ch in vowels for ch in t)) for t in texts]

if __name__ == "__main__":
    texts = [
        "The quick brown fox",
        "ChatGPT is helpful",
        "Async rewards can reduce latency",
    ]

    rewards_per_func = []
    for reward_fn in [reward_length, reward_vowel_count]:
        rewards_per_func.append(reward_fn(texts))
    print(rewards_per_func)
```

But it's not ideal, because each reward calculation works in sequence. So we could have something like this instead:

```python
import multiprocessing
import time
from concurrent.futures import ProcessPoolExecutor
from typing import Callable


def reward_length(texts):
    time.sleep(0.5)
    return [float(len(t)) for t in texts]


def reward_vowel_count(texts):
    time.sleep(0.5)
    vowels = set("aeiouAEIOU")
    return [float(sum(ch in vowels for ch in t)) for t in texts]


def _worker_reward(fn, texts):
    return fn(texts)

if __name__ == "__main__":
    texts = [
        "The quick brown fox",
        "ChatGPT is helpful",
        "Async rewards can reduce latency",
    ]

    funcs = [reward_length, reward_vowel_count]
    with ProcessPoolExecutor(max_workers=len(funcs), mp_context=multiprocessing.get_context("spawn")) as executor:
        futures = [executor.submit(_worker_reward, fn, texts) for fn in funcs]
        rewards_per_func = [f.result() for f in futures]
    print(rewards_per_func)
```

which is nice in my opinion. The thing is, it doesn't natively supports asynchronous reward function.

to support async functions we should have something like this:

```python
import asyncio


async def reward_length_async(texts):
    await asyncio.sleep(0.5)
    return [float(len(t)) for t in texts]


async def reward_vowel_count_async(texts):
    await asyncio.sleep(0.5)
    vowels = set("aeiouAEIOU")
    return [float(sum(ch in vowels for ch in t)) for t in texts]


if __name__ == "__main__":
    texts = [
        "The quick brown fox",
        "ChatGPT is helpful",
        "Async rewards can reduce latency",
    ]

    async def asyncio_concurrent(fns, texts: list[str]) -> list[list[float]]:
        tasks = [asyncio.create_task(fn(texts)) for fn in fns]
        return await asyncio.gather(*tasks)

    funcs = [reward_length_async, reward_vowel_count_async]
    rewards_per_func = asyncio.run(asyncio_concurrent(funcs, texts))
    print(rewards_per_func)
```


The issue is that, is introduces a good amount of complexity, because now we have to
- use multiprocessing for synchronous reward functions
- use ayncio for asynchronous reward functions

So I'm wondering, to reduce code complexity inside the GRPOTrainer, why not imposing the reward function to be synchronous, and it would be up to the use to call `asyncio.run` inside some wrapper, like here:


```python
import asyncio
import multiprocessing
from concurrent.futures import ProcessPoolExecutor


async def reward_length_async(texts):
    await asyncio.sleep(0.5)
    return [float(len(t)) for t in texts]


async def reward_vowel_count_async(texts):
    await asyncio.sleep(0.5)
    vowels = set("aeiouAEIOU")
    return [float(sum(ch in vowels for ch in t)) for t in texts]


def run_async_in_process(async_fn, texts): # this
    return asyncio.run(async_fn(texts))


if __name__ == "__main__":
    texts = [
        "The quick brown fox",
        "ChatGPT is helpful",
        "Async rewards can reduce latency",
    ]

    funcs = [reward_length_async, reward_vowel_count_async]

    with ProcessPoolExecutor( max_workers=len(funcs), mp_context=multiprocessing.get_context("spawn")) as executor:
        futures = [executor.submit(run_async_in_process, fn, texts) for fn in funcs]
        rewards_per_func = [f.result() for f in futures]
    print(rewards_per_func)
```

I know it's not ideal, because async functions are meant for concurrency within a single process. It would probably make it slightly slower (because of IPC overhead, process creation, event loop creation) but it would more simple on the GRPO side

qgallouedec:
Another option for the user would be to use a single reward function, like this:

```python
import asyncio
import multiprocessing
from concurrent.futures import ProcessPoolExecutor


async def reward_length_async(texts):
    await asyncio.sleep(0.5)
    return [float(len(t)) for t in texts]


async def reward_vowel_count_async(texts):
    await asyncio.sleep(0.5)
    vowels = set("aeiouAEIOU")
    return [float(sum(ch in vowels for ch in t)) for t in texts]


async def compute_rewards_async(texts):
    funcs = [reward_length_async, reward_vowel_count_async]
    tasks = [asyncio.create_task(fn(texts)) for fn in funcs]
    return await asyncio.gather(*tasks)


def compute_rewards(texts):
    rewards = asyncio.run(compute_rewards_async(texts))
    combined_rewards = [sum(reward_tuple) for reward_tuple in zip(*rewards)]
    return combined_rewards


def _worker_reward(fn, texts):
    return fn(texts)


if __name__ == "__main__":
    texts = [
        "The quick brown fox",
        "ChatGPT is helpful",
        "Async rewards can reduce latency",
    ]

    funcs = [compute_rewards]
    with ProcessPoolExecutor(max_workers=len(funcs), mp_context=multiprocessing.get_context("spawn")) as executor:
        futures = [executor.submit(_worker_reward, fn, texts) for fn in funcs]
        rewards_per_func = [f.result() for f in futures]
    print(rewards_per_func)
```


it will achieve the same speed as asyncio directly, the only drawback is that in the logs, you will get only the mean/std only for this global reward, and not per-reward.

pramodith:
I see where you're coming from! I don't really have a strong stance on this. My main pushback is that at least going by this PR it doesn't seem too complicated to handle both sync and async rewards within the library. To me it just feels much easier to handle the async wrapping ourselves rather than ask each user of the library to write the same bit of wrapper code and I can't foresee maintaining this to be much of an hassle, given that `asyncio` and how we're using it is a stable and well known use case.

Also I can see a future where RL training inside TRL starts supporting async rollouts and training similar to PipeLineRL or PrimeRL and we might need to add tons of async support then, perhaps this can be a stepping stone for that?

trl/trainer/grpo_trainer.py
```diff
@@ -1195,14 +1203,73 @@ def _calculate_rewards(self, inputs, prompts, completions, completion_ids_list):
...
+            async def _run_async_funcs():
+                coros = [_invoke_async_reward(i, func, func_name) for (i, func, func_name) in async_funcs_info]
+                return await asyncio.gather(*coros)
+
+            # Detect if a loop is already running (e.g., pytest-asyncio). If so create a new loop to avoid
+            # creating coroutines that never get awaited due to fallback.
+            try:
+                running_loop = asyncio.get_running_loop()
+            except RuntimeError:
+                running_loop = None
```
qgallouedec: This may be a silly question, but I'm not very familiar with asyncio. How can we have a loop already running?

qgallouedec:
> I see where you're coming from! I don't really have a strong stance on this. My main pushback is that at least going by this PR it doesn't seem too complicated to handle both sync and async rewards within the library. To me it just feels much easier to handle the async wrapping ourselves rather than ask each user of the library to write the same bit of wrapper code and I can't foresee maintaining this to be much of an hassle, given that `asyncio` and how we're using it is a stable and well known use case.
> 
> Also I can see a future where RL training inside TRL starts supporting async rollouts and training similar to PipeLineRL or PrimeRL and we might need to add tons of async support then, perhaps this can be a stepping stone for that?

Yep I guess you're right

---

```
2025-12-10T17:26:27.6600554Z FAILED tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm_and_importance_sampling[trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration] - AttributeError: Can't pickle local object 'GRPOTrainer._calculate_rewards.<locals>._invoke_sync_reward'
```

I think you need to move the `_invoke_sync_reward` outside `GRPOTrainer`:
```python
def _invoke_sync_reward(...): ...

class GRPOTrainer:
    def _calculate_rewards(self, ...): ...
```

pramodith:
Ran a few quick tests and it seems like the overhead of creating Processes or new Threads is more expensive than just running stuff sequentially because the reward functions aren't too CPU intensive.

```
import time
from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor

def simple_string_ops(text_list):
    return [text.upper().strip() for text in text_list]

data = [["hello world"] * 100 for _ in range(4)]

# Sequential
start = time.time()
results_seq = [simple_string_ops(d) for d in data]
seq_time = time.time() - start
print(f"Sequential: {seq_time:.4f}s")

# ThreadPool
start = time.time()
with ThreadPoolExecutor(max_workers=4) as executor:
    results_thread = list(executor.map(simple_string_ops, data))
thread_time = time.time() - start
print(f"ThreadPool: {thread_time:.4f}s")

# ProcessPool
start = time.time()
with ProcessPoolExecutor(max_workers=4) as executor:
    results_thread = list(executor.map(simple_string_ops, data))
process_time = time.time() - start
print(f"ProcessPool: {process_time:.4f}s")
```

The result:

```
Sequential: 0.0002s
ThreadPool: 0.0032s
ProcessPool: 0.0343s
```

So we're going to run all the sync reward functions sequentially, and use asyncio to run async reward functions in parallel.

trl/trainer/grpo_trainer.py
pramodith: Yeah this is probably overkill, the only way we can have an already running loop is if `_calculate_rewards `is called in an async context which never happens, so getting rid of it.

trl/trainer/grpo_trainer.py
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import asyncio
```
qgallouedec, line 15:
```suggestion

import asyncio
```

qgallouedec (approved):
LGTM, Thanks!
I made a few style modifications. 
It could be nice to have the same in RLOO in a follow up PR

PR #4571, opened 2025-11-25
Fix vLLM sleep mode: add collective RPC call to reload weights in vLLM wake-up process

qgallouedec:
See https://github.com/vllm-project/vllm/issues/29341

### DIFF
--- a/trl/experimental/openenv/utils.py
+++ b/trl/experimental/openenv/utils.py
@@ -101,6 +101,7 @@ def generate_rollout_completions(
 
     if trainer.args.vllm_enable_sleep_mode:
         trainer.llm.wake_up(tags=["kv_cache"])
+        # Work around for https://github.com/vllm-project/vllm/issues/29341
         trainer.llm.collective_rpc("reload_weights")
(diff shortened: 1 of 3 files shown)


trl/experimental/gold/gold_trainer.py
```diff
@@ -1845,6 +1845,7 @@ def _move_model_to_vllm(self):
         if self.vllm_mode == "colocate" and self.vllm_enable_sleep_mode:
             empty_cache()
             self.vllm_engine.wake_up(tags=["weights"])
+            self.vllm_engine.collective_rpc("reload_weights")
```
lewtun, line 1849: Shall we add an internal note to our future selves?
```suggestion
            # Work around for https://github.com/vllm-project/vllm/issues/29341
            self.vllm_engine.collective_rpc("reload_weights")
```

lewtun (approved):
LGTM with a suggestion that we include a comment in the changes since this is quite cryptic

edbeeching (approved):

PR #4420, opened 2025-11-01
[ALST/Ulysses] Added ALST/Ulysses documentation

kashif:
# What does this PR do?

Added an ALST/Ulysses section to the distributed training doc with yaml config.

Needs: 
- [x] https://github.com/huggingface/accelerate/pull/3817 
- [x] https://github.com/huggingface/transformers/pull/41832
- [x] https://github.com/huggingface/transformers/pull/42444
- [ ] a new release with these PRs merged
 

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #4579, opened 2025-11-26
fix: use shift_labels for metrics when using CP or SP

jue-jue-zi:
When appling deepspeed sequence parallelism, it patches the dataloader and replace the `labels` to `shift_labels`. This patch make the label prefetch broken and raise KeyError.


kashif:
thanks @jue-jue-zi currently i am documenting this feature here: https://github.com/huggingface/trl/pull/4420

kashif:
@jue-jue-zi do you know what error you get and on which version of accerlate/transformers you tested? SP is not released yet BTW. Also deepspeed does not replace the label but adds the `"shift_labels"`

jue-jue-zi:
> @jue-jue-zi do you know what error you get and on which version of accerlate/transformers you tested? SP is not released yet BTW. Also deepspeed does not replace the label but adds the `"shift_labels"`

I patch the latest transformers by cherry picking the latest Deepspeed SP commit https://github.com/huggingface/transformers/commit/7e0ea6997411f2633712cec5c475b791efe69785 (see https://github.com/huggingface/transformers/issues/42414). I used the SFTTrainer from trl v0.23.1 version and encountered the KeyError problem. I have not checked carefully if Deepspeed remove the labels key or not, but this patch works for me.

PR #4324, opened 2025-10-22
Hotfix: Fall back to config.text_config._name_or_path if missing config._name_or_path

albertvillanova:
Fall back to `config.text_config._name_or_path` if missing `config._name_or_path`.

Hotfix for #4323, so the CI becomes green until a definite solution is proposed for the underlying issue.

Issue #4323, opened 2025-10-22
CI fails with dev dependencies: HFValidationError: Repo id must use alphanumeric chars


albertvillanova:
CI fails with dev dependencies: https://github.com/huggingface/trl/actions/runs/18712033739/job/53362520844
> huggingface_hub.errors.HFValidationError: Repo id must use alphanumeric chars, '-', '_' or '.'. The name cannot start or end with '-' or '.' and the maximum length is 96: ''.

Multiple tests:
```python
  FAILED tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm[trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration] - huggingface_hub.errors.HFValidationError: Repo id must use alphanumeric chars, '-', '_' or '.'. The name cannot start or end with '-' or '.' and the maximum length is 96: ''.
  FAILED tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm[trl-internal-testing/tiny-Qwen2VLForConditionalGeneration] - huggingface_hub.errors.HFValidationError: Repo id must use alphanumeric chars, '-', '_' or '.'. The name cannot start or end with '-' or '.' and the maximum length is 96: ''.
  FAILED tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm_beta_non_zero - huggingface_hub.errors.HFValidationError: Repo id must use alphanumeric chars, '-', '_' or '.'. The name cannot start or end with '-' or '.' and the maximum length is 96: ''.
  FAILED tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm_peft - huggingface_hub.errors.HFValidationError: Repo id must use alphanumeric chars, '-', '_' or '.'. The name cannot start or end with '-' or '.' and the maximum length is 96: ''.
  FAILED tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm_and_importance_sampling - huggingface_hub.errors.HFValidationError: Repo id must use alphanumeric chars, '-', '_' or '.'. The name cannot start or end with '-' or '.' and the maximum length is 96: ''.
  FAILED tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm_and_liger - huggingface_hub.errors.HFValidationError: Repo id must use alphanumeric chars, '-', '_' or '.'. The name cannot start or end with '-' or '.' and the maximum length is 96: ''.
[...]

PR #4581, opened 2025-11-26
Revert hotfix Fall back to config.text_config._name_or_path

albertvillanova:
Revert hotfix:
- #4324

for issue:
- #4323

Note that the upstream issue was only present on `transformers` main branch and it has been fixed by:
- https://github.com/huggingface/transformers/pull/41808

### DIFF
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -2026,5 +2026,4 @@ def get_config_model_id(config: PretrainedConfig) -> str:
         `str`:
             The model identifier associated with the model configuration.
     """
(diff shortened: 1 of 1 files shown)


kashif:
@albertvillanova do we need to pin the transformers version?

albertvillanova:
Thanks for your review, @kashif.

No, I don't think so: note that both commits, the one creating the issue and the one fixing it, are not released yet, but only present in the transformers main branch:
- https://github.com/huggingface/transformers/commit/ede7976cd2462ce868a0058c339c6b21baf7fc04
- https://github.com/huggingface/transformers/commit/c79040303967682a6351f3fc01a02b8ce5a13568

PR #4620, opened 2025-12-04
Raise FutureWarning for trainer moved to experimental

albertvillanova:
Raise FutureWarning for trainer moved to experimental.

This PR updates the warning class in the `CPOTrainer` initializer to use a `FutureWarning` and sets the appropriate stack level. This will help users more clearly see the deprecation notice and where it originates in their code.


qgallouedec:
You could do the same for KTO and RLOO as well I think

https://github.com/huggingface/trl/blob/495381d0d95500246e6fdd3137f8397a51487425/trl/trainer/rloo_trainer.py#L246-L251

https://github.com/huggingface/trl/blob/495381d0d95500246e6fdd3137f8397a51487425/trl/trainer/kto_trainer.py#L364-L369

albertvillanova:
Thanks, for your review. Yes, I think it makes sense. Done!

PR #4575, opened 2025-11-25
🚚 Move KTO to trl.experimental

neha222222:
This PR moves KTOTrainer and KTOConfig to the `trl.experimental` submodule.
   
   Part of #[4374](https://github.com/huggingface/trl/issues/4374)
   
   Changes:
   - Move `kto_config.py` and `kto_trainer.py` to `trl/experimental/kto/`
   - Update imports across scripts, examples, tests, and documentation
   - Move KTO documentation from Trainers to Experimental section

PR #4629, opened 2025-12-05
Refactor suppression of warning at experimental import

albertvillanova:
Refactor experimental warning suppression,

This PR refactors how to suppress warnings during imports from `trl.experimental`: instead of setting a temporary environment variable, it uses a localized warning filter.

**Why this is better:**
- Scoped suppression: warnings.catch_warnings() confines changes to the import block, avoiding global or process-wide side effects.
- No environment coupling: Avoids manipulating env variables that can leak into subprocesses, tests, or other modules unintentionally.
- Precision: Filters a specific warning category instead of implicitly affecting code paths that read env variables.
- Thread-safety and reentrancy: Context-manager based filters are safer in concurrent scenarios than shared env state.

**Experimental warning handling improvements:**

* Added a custom warning class `TRLExperimentalWarning` in `trl/import_utils.py` to distinguish experimental API warnings from generic user warnings.
* Replaced the use of the `temporary_env` context manager (resetting the environment variable `TRL_EXPERIMENTAL_SILENCE`) with a new `suppress_experimental_warning` context manager throughout the codebase, including all trainer and config modules. 
* Updated the warning issued in `trl/experimental/__init__.py` to use the new `TRLExperimentalWarning` class instead of the generic `UserWarning`.

### DIFF
--- a/trl/import_utils.py
+++ b/trl/import_utils.py
@@ -112,22 +112,14 @@ class TRLExperimentalWarning(UserWarning):
 
 
 @contextmanager
-def temporary_env(var, value):
-    """
-    Temporarily set an environment variable. Restores the original value (if any) when done.
-    """
-    # Save original value (or None if not set)
-    original = os.environ.get(var)
-    os.environ[var] = value
-    try:
+def suppress_warning(category):
+    with warnings.catch_warnings():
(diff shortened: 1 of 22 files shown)


qgallouedec:
I'm merging this one so that I can use it in #4575

PR #4632, opened 2025-12-05
Preserve truncated tokens in BFD packing

### LINKED ISSUE #4554: Better packing of data with best-fit decrease strategy
Hello,

When using packing with the bfd strategy, it looks like too much truncation is done when the seq_length is smaller than the average length of the sequences we want to pack.

For example : 
```python
from datasets import Dataset
from trl import pack_dataset

examples = {
    "input_ids": [[1, 2, 3, 4], [5, 6], [7, 8, 9], [10]],
    "attention_mask": [[1, 1, 1, 1], [1, 0], [1, 0, 0], [1]],
}

packed_dataset = pack_dataset(dataset, seq_length=3, strategy="bfd")
print(packed_dataset )
```

results in:
```python
{'input_ids': [[1, 2, 3], [7, 8, 9], [5, 6, 10]],
 'attention_mask': [[1, 1, 1], [1, 0, 0], [1, 0, 1]],
 'seq_lengths': [[3], [3], [2, 1]]}
```
So the token '4' is missing from the training tokens.

In a extreme case:
```python
examples_2 = {
    "input_ids": [[0, 0], [1, 2, 3, 4], [5, 6, 7, 8, 9], [10]],
    "attention_mask": [[1, 1], [1, 1, 1, 1], [1, 1, 1, 1, 1], [1]],
}
dataset_2 = Dataset.from_dict(examples_2)
print(pack_dataset(dataset_2, seq_length=1, strategy="bfd")[:])
```
results in:
```python
{'input_ids': [[0], [1], [5], [10]],
 'attention_mask': [[1], [1], [1], [1]],
 'seq_lengths': [[1], [1], [1], [1]]}
```

So here we are basically applying truncation to every sequence instead of having twelve sequences of one token.

If we put ourself in a more usefull setting, when I was finetunning on some very long sequences with a seq_lenfth of 4096, the majority of the tokens was discarded y the bfd packing. On my dataset, the bfd method kept only 0.2% of the total training tokens.

Is the behavior normal ?
I would find it useful to add an option to still have tokens that are deleted in other sequences, even if this is less than ideal. It would be a good compromise between the current versions of bfd and wrapped.

qgallouedec:
This PR updates the **`bfd` packing strategy** so that tokens beyond `seq_length` are **not discarded**.
Instead, truncated fragments are re-queued and packed like any other sequence, preventing unnecessary token loss.

Closes #4554

Untitled-2025-07-22-1600

### Before/After

```python
from datasets import Dataset
from trl import pack_dataset

examples = {
    "input_ids": [[1, 2, 3, 4, 5], [6, 7], [8, 9, 10], [11]],
}
packed_dataset = pack_dataset(dataset, seq_length=4, strategy="bfd")
print(packed_dataset[:])
```

```diff
- {'input_ids': [[1, 2, 3, 4], [8, 9, 10, 11], [6, 7]], 'seq_lengths': [[4], [3, 1], [2]]}
+ {'input_ids': [[1, 2, 3, 4], [8, 9, 10, 5], [6, 7, 11]], 'seq_lengths': [[4], [3, 1], [2, 1]]}
```

## Benchmark

TLDR: a bit slower, but still very fast

```python
import random
import time
from datasets import Dataset
from trl.data_utils import pack_dataset

total_tokens = 10_000_000
seq_length = 2048  # packing target
min_seq_len, max_seq_len = 1024, 3072  # arbitrary input lengths

input_ids = []
tokens_left = total_tokens
while tokens_left > 0:
    n = min(tokens_left, random.randint(min_seq_len, max_seq_len))
    tokens_left -= n
    input_ids.append(list(range(n)))

dataset = Dataset.from_dict({"input_ids": input_ids})

start = time.perf_counter()
packed = pack_dataset(dataset, seq_length=seq_length)
elapsed = time.perf_counter() - start

print(f"Packed {total_tokens} tokens into {len(packed)} examples in {elapsed:.3f}s")
```

```
# Before: Packed 10000000 tokens into 4848 examples in 0.189s
# After:  Packed 10000000 tokens into 4952 examples in 0.255s
```

> [!IMPORTANT]
> This PR was mostly written using Codex. Based on my tests, it works. I think I understand the most of it, but I'm not behind most of the code changes.


trl/data_utils.py
```diff
@@ -647,11 +647,45 @@ def _pack_bfd(examples: pa.Table, seq_length: int) -> pa.Table:
     columns = []
     list_column_idx = None
     for idx, column in enumerate(examples.columns):
+        if isinstance(column, pa.ChunkedArray):
+            column = column.combine_chunks()
         if pyarrow.types.is_list(column.type) or pyarrow.types.is_large_list(column.type):
-            column = pc.list_slice(column, 0, seq_length)
```
qgallouedec, line 651: this was truncating all list columns. We don't want to truncated anymore

trl/data_utils.py
```diff
@@ -647,11 +647,45 @@ def _pack_bfd(examples: pa.Table, seq_length: int) -> pa.Table:
     columns = []
     list_column_idx = None
     for idx, column in enumerate(examples.columns):
+        if isinstance(column, pa.ChunkedArray):
+            column = column.combine_chunks()
```
qgallouedec, line 652: In PyArrow, a `ChunkedArray` is one logical column made of multiple smaller arrays ("chunks") under the hood.
Combining chunks gives one continuous array (leaving it chunked would mean offsets restart in each piece). It allows the code to operate on a single contiguous chunk, which keeps offsets consistent and avoids chunk-boundary surprises.

jiosephlee:
Looking forward to this PR!

trl/data_utils.py
```diff
@@ -644,23 +644,56 @@ def search(self, val):
...
+    expanded_indices: list[int] = []
+    for row_idx, (row_start, row_end) in enumerate(zip(offsets[:-1], offsets[1:], strict=False)):
+        length = row_end - row_start
+        for split_start in range(0, length, seq_length):
+            frag_len = min(seq_length, length - split_start)
+            frag_lengths.append(frag_len)
+            frag_slices.append((row_start + split_start, frag_len))
+            expanded_indices.append(row_idx)
+
+    # Rebuild list column with fragments and duplicate non-list columns accordingly.
```
albertvillanova: > duplicate non-list columns accordingly

Does this function support non-list columns?

qgallouedec: No, but in practice, we only get list columns:

https://github.com/huggingface/trl/blob/d401a42588d1a856aed78b01326eb39048eafbaf/trl/trainer/sft_trainer.py#L1088-L1102

qgallouedec: I added a check in the function to ensure all columns are list

qgallouedec: https://github.com/huggingface/trl/blob/ad82b138d0aefb1aded5e31aa459512725826431/trl/data_utils.py#L650-L654

tests/test_data_utils.py
```diff
@@ -1000,13 +1000,11 @@ class TestPackDatasetBfd(TrlTestCase):
     def test_simple(self):
         examples = {
             "input_ids": [[1, 2, 3], [4, 5, 6, 7], [8]],
-            "attention_mask": [[0, 1, 1], [0, 0, 1, 1], [1]],
         }
         dataset = Dataset.from_dict(examples)
         seq_length = 4
         expected_output = {
             "input_ids": [[4, 5, 6, 7], [1, 2, 3, 8]],
-            "attention_mask": [[0, 0, 1, 1], [0, 1, 1, 1]],
```
qgallouedec, line 1009: In practice, we never all packing with the attention mask + it doesn't add anything to test it here. So I suggest we remove it.

tests/test_data_utils.py
```diff
@@ -1015,44 +1013,53 @@ def test_simple(self):
...
+        examples = {
+            "col1": [[1, -2, 3, -4, 5, -6], [7, -8, 9], [-10, 11, -12], [13, -14, 15, -16]],
+            "col2": [[-1, 2, -3, 4, -5, -6], [-7, 8, -9], [10, -11, 12], [-13, 14, -15, 16]],
+        }
+        dataset = Dataset.from_dict(examples)
+        seq_length = 4
+        expected_output = {
+            "col1": [[1, -2, 3, -4], [13, -14, 15, -16], [7, -8, 9], [-10, 11, -12], [5, -6]],
+            "col2": [[-1, 2, -3, 4], [-13, 14, -15, 16], [-7, 8, -9], [10, -11, 12], [-5, 6]],
+            "seq_lengths": [[4], [4], [3], [3], [2]],
```
qgallouedec, line 1050: Following the finding [here](https://github.com/huggingface/trl/pull/4632#pullrequestreview-3551963666), this new test ensure we have a consistent packing across columns

tests/test_data_utils.py
```diff
@@ -1015,44 +1013,53 @@ def test_simple(self):
...
+            "input_ids": [[1, 2, 3, 4], [8, 9, 10, 11], [6, 7, 5, 12]],
+            "seq_lengths": [[4], [4], [2, 1, 1]],
+        }
+        dataset = pack_dataset(dataset, seq_length, strategy="bfd")
+        assert dataset.to_dict() == expected_output
+
+    def test_with_overlong_two_coluns(self):
+        examples = {
+            "col1": [[1, -2, 3, -4, 5, -6], [7, -8, 9], [-10, 11, -12], [13, -14, 15, -16]],
+            "col2": [[-1, 2, -3, 4, -5, -6], [-7, 8, -9], [10, -11, 12], [-13, 14, -15, 16]],
```
albertvillanova: I think this is causing the CI error.
```suggestion
            "col2": [[-1, 2, -3, 4, -5, 6], [-7, 8, -9], [10, -11, 12], [-13, 14, -15, 16]],
```

trl/data_utils.py
albertvillanova: I meant: if non-list columns are not supported, then why "duplicating non-list columns accordingly"?

tests/test_data_utils.py
qgallouedec: 🤦

trl/data_utils.py
```diff
@@ -644,23 +644,71 @@ def search(self, val):
...
+    expanded_indices: list[int] = []
+    for row_idx, (row_start, row_end) in enumerate(zip(offsets[:-1], offsets[1:], strict=False)):
+        length = row_end - row_start
+        for split_start in range(0, length, seq_length):
+            frag_len = min(seq_length, length - split_start)
+            frag_lengths.append(frag_len)
+            frag_info.append((row_idx, split_start, frag_len))
+            expanded_indices.append(row_idx)
+
+    # Rebuild list columns with fragments and duplicate non-list columns accordingly.
```
albertvillanova: Maybe we can remove the no longer applicable part of the comment:

```suggestion
    # Rebuild list columns with fragments
```

qgallouedec: Right, I can't remember where this comment comes from.

trl/data_utils.py
qgallouedec: https://github.com/huggingface/trl/pull/4632/changes#r2624589843

PR #4635, opened 2025-12-06
Fix KTOTrainer CUDA error for large-vocab models via tensor indexing

bhuvanprakash:
### This​‍​‌‍​‍‌​‍​‌‍​‍‌ PR resolves a CUDA error in KTOTrainer during model training with extremely large vocabularies (e.g., Qwen3-VL ~151k vocab). Hence, it opens up the possibility of KTO training on multimodal models having extended vocabularies like Qwen3-VL or alike.

This problem was initially seen downstream in Unsloth while performing KTO training. After a detailed study, the source was found to be Hugging Face TRL implementation. This PR corrects the root cause upstream so that local patches are no longer necessary for downstream projects (e.g., Unsloth). Tested successfully in Unsloth with Qwen3-VL after this fix.

The error resulted from Python list-based fancy indexing of tensors having shapes similar to:

[batch_size, seq_len, vocab_size]

For extremely large vocab sizes, this might lead to:

CUDA error: invalid configuration argument

Downstream issue reference:

unslothai/unsloth#3675

### Root Cause

In trl/trainer/kto_trainer.py, Python lists are utilized to split the batch and these lists are simultaneously employed for fancy indexing on big tensors:

<pre><code>chosen_idx = [i for i in range(completion_logps.shape[0]) if batch["label"][i]] rejected_idx = [i for i in range(completion_logps.shape[0]) if not batch["label"][i]] </code></pre><pre><code>chosen_logps = completion_logps[chosen_idx, ...] rejected_logps = completion_logps[rejected_idx, ...] chosen_logits = completion_logits[chosen_idx, ...] rejected_logits = completion_logits[rejected_idx, ...] </code></pre>While dealing with extremely large vocab tensors ([batch, seq, vocab]), this makes a CUDA kernel to get launched with an invalid configuration and thus training crashes.

A similar pattern can be found in the reference_logps branch of get_batch_loss_metrics(...).

 ### Fix

Python list indexing is substituted with device-aware tensor indices and torch.index_select that calls the CUDA kernels that are optimized.

<pre><code>forward(...) device = completion_logits.device labels = torch.as_tensor(batch["label"], dtype=torch.bool, device=device) chosen_idx = torch.nonzero(labels, as_tuple=False).view(-1) rejected_idx = torch.nonzero(~labels, as_tuple=False).view(-1) chosen_logps = completion_logps.index_select(0, chosen_idx) rejected_logps = completion_logps.index_select(0, rejected_idx) chosen_logits = completion_logits.index_select(0, chosen_idx) rejected_logits = completion_logits.index_select(0, rejected_idx) </code></pre><pre><code>get_batch_loss_metrics(...) (when using reference_logps) device = batch["reference_logps"].device labels = torch.as_tensor(batch["label"], dtype=torch.bool, device=device) chosen_idx = torch.nonzero(labels, as_tuple=False).view(-1) rejected_idx = torch.nonzero(~labels, as_tuple=False).view(-1) reference_chosen_logps = batch["reference_logps"].index_select(0, chosen_idx) reference_rejected_logps = batch["reference_logps"].index_select(0, rejected_idx) </code></pre>This maintains the semantics to be exactly the same but employs safe CUDA kernels for indexing.

### Why This Works

<ul><li>index_select uses optimized CUDA kernels</li><li>indices are on the correct device</li><li>avoids Python list fancy indexing</li><li>no CPU-GPU sync overhead</li><li>identical behavior for normal vocab sizes</li><li>robust for 150k+ vocab models</li></ul>All existing logic paths remain unchanged for normal vocab scenarios.

Testing Unit Tests

### Executed:

python3.10 -m pytest tests/test_kto_trainer.py -v

All KTO tests pass:

<ul><li> TestKTOTrainer::test_kto_trainer[...]</li><li>test_kto_trainer_with_ref_model_is_model</li><li>test_tokenize_and_process_tokens</li><li>test_kto_trainer_without_providing_ref_model</li><li>test_kto_trainer_generate_during_eval_no_wandb</li><li>test_compute_metrics</li></ul>(LoRA / liger tests skipped as per upstream config.)

Large-Vocab Simulation

Dummy tensor was created:

[4, 256, 151_936]

### Compared:

old: tensor[list, ...]

new: tensor.index_select(0, idx_tensor)

Verification:

torch.allclose(old, new) == True

shapes match

values match

So behavior is unchanged — only the indexing method is safer.

### Backward Compatibility

<ul><li>no API changes</li><li>no behavior changes for normal vocab sizes</li><li>only internal indexing logic updated</li><li>fixes KTO training for large vocab models (e.g. Qwen family)</li></ul>Impact

This allows KTO training for multilingual and multimodal models with extended vocabularies without triggering CUDA kernel launch errors. Downstream frameworks like Unsloth will get this fix automatically when they update TRL; hence, they don't need ​‍​‌‍​‍‌​‍​‌‍​‍‌local

### DIFF
--- a/docs/source/example_overview.md
+++ b/docs/source/example_overview.md
@@ -29,6 +29,7 @@ These notebooks are easier to run and are designed for quick experimentation wit
 
 | Notebook | Description | Open in Colab |
 |----------|-------------|---------------|
+| [`grpo_agent.ipynb`](https://github.com/huggingface/trl/tree/main/examples/notebooks/grpo_agent.ipynb) | GRPO for agent training | Not available due to OOM with Colab GPUs |
 | [`grpo_rnj_1_instruct.ipynb`](https://github.com/huggingface/trl/tree/main/examples/notebooks/grpo_rnj_1_instruct.ipynb) | GRPO rnj-1-instruct with QLoRA using TRL on Colab to add reasoning capabilities | [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/huggingface/trl/blob/main/examples/notebooks/grpo_rnj_1_instruct.ipynb) |
 | [`sft_ministral3_vl.ipynb`](https://github.com/huggingface/trl/tree/main/examples/notebooks/sft_ministral3_vl.ipynb) | Supervised Fine-Tuning (SFT) Ministral 3 with QLoRA using TRL on free Colab | [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/huggingface/trl/blob/main/examples/notebooks/sft_ministral3_vl.ipynb) |
 | [`grpo_ministral3_vl.ipynb`](https://github.com/huggingface/trl/tree/main/examples/notebooks/grpo_ministral3_vl.ipynb) | GRPO Ministral 3 with QLoRA using TRL on free Colab | [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/huggingface/trl/blob/main/examples/notebooks/grpo_ministral3_vl.ipynb) |


qgallouedec:
Very good fix, thanks. I was able to reproduce locally both the bug and the fix

qgallouedec (approved):

PR #4636, opened 2025-12-07
Disable gradient checkpointing during no-grad inference to avoid PyTorch warning

qgallouedec:
This PR adds a context manager to temporarily disable gradient checkpointing when running inference in a `torch.no_grad()` context. This prevents a noisy warning from being raised by PyTorch.

```python
import torch

device = "cuda"

model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen2-0.5B-Instruct").to(device)
model.train()
model.gradient_checkpointing_enable()
input_ids = torch.tensor([[1, 2, 3]]).to(device)

with torch.no_grad():
    output = model(input_ids=input_ids)
```

When 

1. the model is in train model
2. gradient checkpointing is enabled with `use_reentrant=True` (default)
3. the inference is run within a `torch.no_grad()` context

we get the following warning:

```
/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/torch/utils/checkpoint.py:85: UserWarning: None of the inputs have requires_grad=True. Gradients will be None
  warnings.warn(
```

This is a safe warning, because no gradients are expected to be computed within the `torch.no_grad()` context, but it is noisy. To avoid this warning, we can disable gradient checkpointing before entering the `torch.no_grad()` context, and re-enable it afterwards.

```python
import torch
from trl.models.utils import disable_gradient_checkpointing

device = "cuda"

model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen2-0.5B-Instruct").to(device)
model.train()
model.gradient_checkpointing_enable()
input_ids = torch.tensor([[1, 2, 3]]).to(device)

with torch.no_grad(), disable_gradient_checkpointing(model):
    output = model(input_ids=input_ids)
```

### DIFF
--- a/docs/source/example_overview.md
+++ b/docs/source/example_overview.md
@@ -29,6 +29,7 @@ These notebooks are easier to run and are designed for quick experimentation wit
 
 | Notebook | Description | Open in Colab |
 |----------|-------------|---------------|
+| [`grpo_agent.ipynb`](https://github.com/huggingface/trl/tree/main/examples/notebooks/grpo_agent.ipynb) | GRPO for agent training | Not available due to OOM with Colab GPUs |
 | [`grpo_rnj_1_instruct.ipynb`](https://github.com/huggingface/trl/tree/main/examples/notebooks/grpo_rnj_1_instruct.ipynb) | GRPO rnj-1-instruct with QLoRA using TRL on Colab to add reasoning capabilities | [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/huggingface/trl/blob/main/examples/notebooks/grpo_rnj_1_instruct.ipynb) |
 | [`sft_ministral3_vl.ipynb`](https://github.com/huggingface/trl/tree/main/examples/notebooks/sft_ministral3_vl.ipynb) | Supervised Fine-Tuning (SFT) Ministral 3 with QLoRA using TRL on free Colab | [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/huggingface/trl/blob/main/examples/notebooks/sft_ministral3_vl.ipynb) |
 | [`grpo_ministral3_vl.ipynb`](https://github.com/huggingface/trl/tree/main/examples/notebooks/grpo_ministral3_vl.ipynb) | GRPO Ministral 3 with QLoRA using TRL on free Colab | [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/huggingface/trl/blob/main/examples/notebooks/grpo_ministral3_vl.ipynb) |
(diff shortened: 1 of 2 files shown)


trl/models/utils.py
```diff
@@ -522,3 +522,20 @@ def prepare_peft_model(
...
+    Args:
+        model (`PreTrainedModel`):
+            Model for which to temporarily disable gradient checkpointing.
+    """
+    was_enabled = model.is_gradient_checkpointing
+    if was_enabled:
+        model.gradient_checkpointing_disable()
+    yield
+    if was_enabled:
+        model.gradient_checkpointing_enable()
```
albertvillanova: This part of the code will not be called if an exception is raised after the `yield`. I would recommend using a `try...finally` block.

albertvillanova: Just a question: You assume no `gradient_checkpointing_kwargs` was originally passed to enable checkpointing, e.g. `use_reentrant` is `True`. Maybe this assumption is OK with TRL. Just wanted to be sure of this.

trl/models/utils.py
```diff
@@ -522,3 +522,20 @@ def prepare_peft_model(
...
+    Temporarily disable gradient checkpointing, restoring the previous state afterward.
+
+    Args:
+        model (`PreTrainedModel`):
+            Model for which to temporarily disable gradient checkpointing.
+    """
+    was_enabled = model.is_gradient_checkpointing
+    if was_enabled:
+        model.gradient_checkpointing_disable()
+    yield
```
albertvillanova: What about yielding the model? Feel free to ignore this if you think it is OK, but this pattern is widely used 
```python
with func_to_modify_an_instance(instance) as modified_instance:
````

trl/models/utils.py
qgallouedec: I yes, I missed this, thanks!

trl/models/utils.py
qgallouedec: I think this could be a little misleading, since the operation is done in-place within the context manager. As a reader, I would get the impression that we are getting a new instance for which grad checkpointing is disabled.

trl/models/utils.py
qgallouedec: Addressed here https://github.com/huggingface/trl/pull/4636/commits/68bbbc3e7f0025f8dba7f0ce389b49fe34e1cb44

qgallouedec: Addressed here: https://github.com/huggingface/trl/pull/4636/commits/68bbbc3e7f0025f8dba7f0ce389b49fe34e1cb44

qgallouedec:
> Additionally, I have a naive question: what would be the issues if we temporarily set `model.eval()` instead?

`eval` is actually quite different:

| | `model.eval()` | `torch.no_grad()` |
| - | - | - |
| Disables dropout | Yes | No |
| BatchNorm uses running stats | Yes | No |
| Gradient computation | Yes  | No |

(here, we want to disable the gradient computation only.

in short, `eval()` changes layer behavior, `no_grad()` changes autograd behavior.

qgallouedec:
I'll merge this one so that we have it in 0.26

PR #4639, opened 2025-12-08
[CPO/ORPO] Fix handling of different length chosen/rejected prompts.

davmels:
…rompts.

on line 495, zip() function has a flag `strict=True`, and only after that we calculate the num_diff_len, which we can not reach in case the chosen and rejected prompt lengths differ by 1 (as zip raises an exception). Therefore, we should take only the first `prompt_len_input_ids` common parts of the prompts, and find the length differences between the prompts afterwards.

# What does this PR do?

Fixes # issue with handling different length chosen/rejected prompts.

---

> [!NOTE]
> **Low Risk**
> Small change to tokenization validation logic that prevents a premature exception; behavior remains guarded by existing token/length difference checks.
> 
> **Overview**
> Fixes prompt-diff validation in `CPOTrainer` and `ORPOTrainer` tokenization by making the `zip()` comparison non-strict, so examples where chosen/rejected prompts differ in length (e.g., by 1 due to tokenizer merge ops) no longer raise before the explicit length-difference check runs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3cf80bb931ef1b09bbe35615749db72fa0d19129. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/trl/experimental/orpo/orpo_trainer.py
+++ b/trl/experimental/orpo/orpo_trainer.py
@@ -466,7 +466,7 @@ class ORPOTrainer(_BaseTrainer):
             # and length only differs by 1 at most
             num_diff_tokens = sum(
                 a != b
-                for a, b in zip(chosen_tokens["prompt_input_ids"], rejected_tokens["prompt_input_ids"], strict=True)
+                for a, b in zip(chosen_tokens["prompt_input_ids"], rejected_tokens["prompt_input_ids"], strict=False)
             )
             num_diff_len = abs(chosen_prompt_len_input_ids - rejected_prompt_len_input_ids)
             if num_diff_tokens > 1 or num_diff_len > 1:


trl/experimental/cpo/cpo_trainer.py
```diff
@@ -492,7 +492,7 @@ def tokenize_row(self, feature, model: PreTrainedModel | nn.Module | None = None
             # and length only differs by 1 at most
             num_diff_tokens = sum(
                 a != b
-                for a, b in zip(chosen_tokens["prompt_input_ids"], rejected_tokens["prompt_input_ids"], strict=True)
+                for a, b in zip(chosen_tokens["prompt_input_ids"][:prompt_len_input_ids], rejected_tokens["prompt_input_ids"][:prompt_len_input_ids])
```
qgallouedec:
```suggestion
                for a, b in zip(chosen_tokens["prompt_input_ids"], rejected_tokens["prompt_input_ids"], strict=False)
```

davmels: Thanks for the review!
This should also work fine

PR #4643, opened 2025-12-09
Include `generation_config` for tiny model uploads

qgallouedec:
We used to have many of these warnings in the CI, 
```
The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config. The model config and generation config were aligned accordingly, being updated with the tokenizer's values. Updated tokens: {'eos_token_id': 151645, 'pad_token_id': 151645}.
```

it was because the generation config was not properly uploaded


albertvillanova:
It seems the new tiny models break the CI tests. I'm investigating it.

albertvillanova:
I think the issue is caused by this:
```python
from transformers import AutoConfig, GenerationConfig

model_id = "Qwen/Qwen2.5-VL-3B-Instruct"
config = AutoConfig.from_pretrained(model_id)
print(f"{config.temperature=}")
generation_config = GenerationConfig.from_pretrained(model_id)
print(f"{generation_config.temperature=}")
```
```python
config.temperature=1.0
generation_config.temperature=1e-06
```

albertvillanova:
I have proposed a fix:
- #4647

PR #4647, opened 2025-12-09
Overwrite model default generation config used by model.generate

albertvillanova:
Overwrite model default generation config.

This ensures both the trainer and model use the same config object, avoiding the merging conflict.

This PR makes a targeted update to how generation configuration is handled in the trainer class to ensure proper overriding of model defaults with training parameters. The main change is to use the model's existing `generation_config` and update it with training parameters, rather than creating a new `GenerationConfig` object from scratch.

## Problem                                                                                                                                                                                                                                                                                                                                                                       
When using GRPOTrainer with models that have non-standard generation configs (e.g., Qwen models with `temperature=1e-06`), the trainer's generation parameters (like `temperature=1.0` set via `GRPOConfig`) were not being properly applied during generation. This resulted in:
 - ❌ No variance in generated completions                                                                                                                                                                                                                                                                                                                                        
 - ❌ Zero variance in rewards                                                                                                                                                                                                                                                                                                                                                    
 - ❌ GRPO training failing to converge

## Root Cause
The issue stems from transformers' generation config merging logic in `GenerationMixin._prepare_generation_config` (called by `GenerationMixin.generate`): https://github.com/huggingface/transformers/blob/47b0e478f324b54f177ea7998a0791870fdd0324/src/transformers/generation/utils.py#L1776-L1781
```python
                    if (
                        custom_gen_config_value == global_default_value
                        and model_gen_config_value != global_default_value
                    ):
                        modified_values[key] = model_gen_config_value
                        setattr(generation_config, key, model_gen_config_value)
```

**Why this happens:**
- Transformers' global default: `temperature=1.0`
- Model's `generation_config.json`: `temperature=1e-06`
- GRPOTrainer creates: `GenerationConfig(temperature=1.0, ...)`

When `model.generate()` is called, transformers sees that the custom config's `temperature=1.0` matches the global default, so it assumes the user didn't intentionally set it and "helpfully" overwrites it with the model's `1e-06` value.

## Solution
~~Instead of creating a new `GenerationConfig` and passing it to `generate()`, we directly use and update the model's `generation_config` attribute. This ensures both the trainer and model use the same config object, avoiding the merging conflict.~~

This PR implements a clean context manager approach that temporarily overrides generation config only during generation:
1. **Modified shared utility** `unwrap_model_for_generation` in `trl/models/utils.py` to accept optional `generation_config` parameter
2. **Temporary override during generation** using nested context managers: model's config is never permanently modified
3. **Automatic restoration** after generation completes: no manual tracking or save/restore logic needed

This approach benefits all trainers (GRPO, RLOO, etc.), not just GRPO, and follows Python best practices for temporary resource modifications.

## Checklist
- [x] GRPO
- [x] RLOO
- [x] GKD
- [x] GOLD
- [x] Online DPO
- [ ] NashMD
  - [ ] TODO: mixture_model
- [x] XPO
- [x] callbacks: TODO: _generate_completions

### DIFF
--- a/trl/models/utils.py
+++ b/trl/models/utils.py
@@ -151,8 +151,12 @@ def _override_model_generation_config(model, generation_kwargs=None):
         model: The model (typically unwrapped_model) whose generation_config to temporarily override.
         generation_kwargs (dict): Generation kwargs to be used to override model's generation config.
     """
-    # Issue fixed in transformers v5 by PR transformers#42702
-    if Version(transformers.__version__) >= Version("5.0.0") or generation_kwargs is None:
+    if (
+        # Issue fixed in transformers v5 by PR transformers#42702
+        Version(transformers.__version__) >= Version("5.0.0")
+        or generation_kwargs is None
+        or not hasattr(model, "generation_config")
+    ):
         yield model
         return
     # If it is a PEFT model, override the underlying base model


qgallouedec:
Wow nice investigation!

trl/trainer/grpo_trainer.py
```diff
@@ -706,7 +705,9 @@ def cast_outputs_to_original_dtype(module, args, output):
             }
             if args.generation_kwargs is not None:
                 generation_kwargs.update(args.generation_kwargs)
-            self.generation_config = GenerationConfig(**generation_kwargs)
+            # Use the model's generation_config directly to ensure training parameters properly override model defaults
+            self.generation_config = self.model.generation_config
+            self.generation_config.update(**generation_kwargs)
```
qgallouedec: does it mean that, if I train a model, then push it to the hub, depending on the generation kwargs used, I end up with a different pushed generation config?

albertvillanova: I would say you will push `trainer.generation_config` (which equals to the overwritten `trainer.model.generation_config`) as `generation_config.json`.

albertvillanova: For example, in `tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_with_generation_kwargs`, I called `trainer.push_to_hub()` after training and I got this generation_config file: https://huggingface.co/albertvillanova/test_training_with_generation_0/blob/main/generation_config.json
```json
{
  "do_sample": true,
  "eos_token_id": [
    151645,
    151645
  ],
  "length_penalty": -0.1,
  "max_new_tokens": 8,
  "num_beams": 2,
  "pad_token_id": 151643,
  "transformers_version": "4.57.3"
}
```
which equals to
```python
>>> trainer.generation_config
GenerationConfig {
  "do_sample": true,
  "eos_token_id": [
    151645,
    151645
  ],
  "length_penalty": -0.1,
  "max_new_tokens": 8,
  "num_beams": 2,
  "pad_token_id": 151643
}
```

albertvillanova: I see what you mean. Before this PR, in the same test as above, the generation config file pushed to the Hub was: https://huggingface.co/albertvillanova/test_training_with_generation_1/blob/main/generation_config.json
```json
{
  "do_sample": true,
  "eos_token_id": [
    151645,
    151645,
    151643
  ],
  "pad_token_id": 151643,
  "repetition_penalty": 1.05,
  "temperature": 0.7,
  "top_k": 20,
  "top_p": 0.8,
  "transformers_version": "4.57.3"
}
```
which is different from
```python
>>> trainer.generation_config
GenerationConfig {
  "do_sample": true,
  "eos_token_id": 151645,
  "length_penalty": -0.1,
  "max_new_tokens": 8,
  "num_beams": 2,
  "pad_token_id": 151643
}
```
In this case:
```python
>>> trainer.model.generation_config
GenerationConfig {
  "do_sample": true,
  "eos_token_id": [
    151645,
    151645,
    151643
  ],
  "pad_token_id": 151643,
  "repetition_penalty": 1.05,
  "temperature": 0.7,
  "top_k": 20,
  "top_p": 0.8
}
```

albertvillanova: So, when pushing the trained model to the Hub, the pushed generation config is:
- before: the original model generation config (the trainer generation config is ignored)
- now: the trainer generation config (which has overwritten the original model generation config)

On the one hand, I could improve this PR and implement a mechanism to store the original model generation config (different from the training generation config) and then use the original model generation config when saving/pushing the trained model.

Additionally, I am wondering if current transformers' merging logic is correct and/or could be improved. Maybe, I can open an issue in their repo and discuss this.

albertvillanova:
I have opened an issue in `transformers` to discuss their merging logic:
- https://github.com/huggingface/transformers/issues/42762

Their merging logic was introduced in transformers-4.54.1:
- https://github.com/huggingface/transformers/pull/36684

albertvillanova:
In any case, if `transformers` fix the issue on their side, I think we need to fix it as well on our side for older versions of transformers. As I said above, the merging logic was introduced in `transformers-4.54.1`, and we support `transformers>=4.56.1`.

What do you think, @qgallouedec?

qgallouedec:
Totally agree. Any idea how to fix this?

trl/trainer/grpo_trainer.py
qgallouedec: From a user perspective, I think it's more expected that the pushed generation config is that of the initial model. For example, I might want to train with `temperature=1`, but have my model still default to `temperature=0.6`. Do you agree with me @Ed @lewis?

qgallouedec: > implement a mechanism to store the original model generation config (different from the training generation config) and then use the original model generation config when saving/pushing the trained model.


Would this work instead?

```diff
- self.generation_config = self.model.generation_config
+ self.generation_config = copy.deepcopy(self.model.generation_config)
```

albertvillanova: > Would this work instead?

No, I think that would not work because if `model.generation_config` is copied (and not modified in place), their values take precedence over the ones passed to `generate`.

I am working on a workaround: I will push tomorrow so you can review it.

albertvillanova:
I was working on a solution to store the original model generation_config, and restore it when calling `_save` (which saves it and/or pushes it to the Hub).

However, I just thought of another implementation: just modify the original model.generation_config (within a context manager) when generating/sampling.
I'm reworking on this and I will push.

albertvillanova:
After implementing the overwriting of `self.model.generation_config` with `self.generation_config` by using a context manager for the `generate` method, I get these CI failures back: https://github.com/huggingface/trl/actions/runs/20130556478/job/57770710505?pr=4647
```
  FAILED tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm_peft[trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration] - AssertionError: Parameter base_model.model.model.language_model.layers.0.self_attn.q_proj.lora_A.default.weight has not changed.
  FAILED tests/test_rloo_trainer.py::TestRLOOTrainer::test_training_vlm_peft[trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration] - AssertionError: Parameter base_model.model.model.language_model.layers.0.self_attn.q_proj.lora_A.default.weight has not changed.
```
This time only for `_peft`. I am investigating it.

albertvillanova:
I fixed the PEFT issue:
- the model is wrapped in an instance of PeftModel and accessible via `get_base_model()` method.

albertvillanova:
Additionally, I changed the implementation: I integrated the `_override_model_generation_config` context manager within the `unwrap_model_for_generation` context manager.

Still to do: handle the cases where the trainer has 2 models.

albertvillanova:
I changed the implementation and used `generation_kwargs` instead of `generation_config`, so trainers with multiple models can override each of the models' `generation_config` with `generation_kwargs`.

It is still pending to do the implementation for the callbacks:
- these are not currently passed `generation_kwargs`: I could change this and pass it
- however, they are also passed a `model` and a `generation_config`
  - I could use the `generation_kwargs` to override the passed `generation_config`: but which values to override and which not? We get back to the original issue.
  - additionally, what about the case where the passed `generation_config` was not obtained from the `generation_config` of the passed `model`? In that case, the non-default model's generation_config values can have precedence, again getting back to the original issue.

What do you think?

trl/trainer/grpo_trainer.py
albertvillanova: I am fixing the tagged names in this comment :stuck_out_tongue_winking_eye:: https://github.com/huggingface/trl/pull/4647#discussion_r2607926929

CC: @edbeeching @lewtun

trl/experimental/gold/gold_trainer.py
```diff
@@ -1916,7 +1919,9 @@ def training_step(
                 )
                 new_input_ids, new_attention_mask, new_labels, prompt_texts, completion_texts = result
             else:
-                with unwrap_model_for_generation(model, self.accelerator) as unwrapped_model:
+                with unwrap_model_for_generation(
+                    model, self.accelerator, generation_kwargs=self.generation_kwargs
+                ) as unwrapped_model:
                     result = self.generate_on_policy_outputs(
                         unwrapped_model, inputs, self.generation_config, self.processing_class.pad_token_id
```
qgallouedec: When reading the code, it's not clear why we need both a context with `generation_kwargs=self.generation_kwargs` + passing the generation config. Can you possibly add a comment, and maybe a link to the GitHub issue(s)?

trl/trainer/grpo_trainer.py
qgallouedec, line 1499: do we even need to pass `generation_config` here? Since within the context the generation configuration of the model is already the one from the `GRPOConfig`?

albertvillanova:
Thanks for your review, @qgallouedec. I am addressing your comments.

One question is still pending: https://github.com/huggingface/trl/pull/4647#issuecomment-3653556013

> It is still pending to do the implementation for the callbacks:
>- these are not currently passed `generation_kwargs`: I could change this and pass it
>- however, they are also passed a `model` and a `generation_config`
>   - I could use the `generation_kwargs` to override the passed `generation_config`: but which values to override and which not? >We get back to the original issue.
>   - additionally, what about the case where the passed `generation_config` was not obtained from the `generation_config` of the passed `model`? In that case, the non-default model's generation_config values can have precedence, again getting back to the original issue.
>
> What do you think?

More generally about callback design, why these callbacks accept a generation_config as init param, instead of using their trainer.generation_config? If this is an independent argument, then there is no way of knowing the generation_kwargs that should overwrite the model's generation_config.

qgallouedec:
> It is still pending to do the implementation for the callbacks:
> These are not currently passed generation_kwargs: I could change this and pass it

`WinRateCallback` and `LogCompletionsCallback`, right? If so, I think it's okay not to use the trainer's generation parameters for those. In my opinion, users are more likely to expect the actual generation parameters to be used (the ones that are pushed along with the model).

albertvillanova:
After an internal discussion, let me summarize the status of this PR.

First, the bug appears in transformers when calling:
```python
model.generate(..., generation_config=self.generation_config)
```
because of their merging logic, which does not know which `GenerationConfig` parameters were explicitly set to their default value (and should have precedence over the model's value) or were not just explicitly set by the users (and should be ignored and use the model's one instead).

Transformers team is fixing this issue on their side: from v5, non-explicitly set params in `GenerationConfig` will default to None, so their merging logic will know which values should have precedence or not over model's ones.

But what should we do in trl for transformers < v5?

- First, I have tried to keep the code as close as possible to the current one, so once we drop support for transformers < v5, we can continue to use current logic (by passing `self.generation_config` to `model.generate`).
  - So I kept `self.generation_config` and pass it to `model.generate`

- Second, for transformers<5, I have implemented the fixed merging logic on our side, by using a context manager over `model.generate` that uses `self.generation_kwargs` to override `model.generation_config`
  - So I created `self.generation_kwargs` and use them to override `model.generation_config` at each `model.generate` call

trl/experimental/gold/gold_trainer.py
albertvillanova: The self.generation_kwargs passed to the context manager are only used for transformers<5.

For transformers>=5, the self.generation_kwargs are ignored and we just use the passed `generation_config`.

Once we drop support for transformers<5, we will remove the self.generation_kwargs.

trl/trainer/grpo_trainer.py
albertvillanova: For transformers<5, this is a no-operation: both are the same so no overwriting is done.

For transformers>=5, `generation_kwargs` are ignored and we need the passed `generation_config`.

trl/experimental/gold/gold_trainer.py
albertvillanova: I have added explanatory comments.

qgallouedec (approved):
This looks great! Thanks for this important fix. In my opinion this requires a patch release, do you agree?

For the record, at the time of approval, https://github.com/huggingface/transformers/pull/42702 is not merged, so technically

```python
if Version(transformers.__version__) >= Version("5.0.0.dev0") or generation_kwargs is None:
```

isn't enough, but considering that it will mot likely be merged within the next days, we can merge right now as it is!

albertvillanova:
Indeed, I set the `Version(transformers.__version__) >= Version("5.0.0.dev0")` condition mostly to keep in mind that the fix on our side is only necessary for older transformers versions and we can remove it once we drop support for transformers < v5.

Since their fix (https://github.com/huggingface/transformers/pull/42702) is not yet merged on main, I would rather change the condition to: `Version(transformers.__version__) >= Version("5.0.0")`, so our CI is not broken for dev dependencies.

And yes, I agree a patch release is necessary.

PR #58, opened 2022-12-27
`accelerate` integration

younesbelkada:
# What does this PR do?

This PR integrates `trl` with `accelerate` to make it compatible with the tools provided by the library to be able to train models using `PPOTrainer`. This would enable users to train their models in mixed precision, using Data Parallelism etc in a very simple manner. 
Users should design their own training script and run them using `accelerate launch xxx.py` based on the example scripts provided in `examples/scripits`. 
 
This PR also integrates Data Parallelism paradigm, enabling users to benefit from multi-GPU training if they want to speedup training. 

## TODOs

- [x] add example script
- [x] Improve API
- [x] document code
- [x] document example (check `accelerate` examples)
- [x] Input safety checkers (very important)

###  DeepSpeed tests (check where it works)

- [ ] zero-0 
- [x] zero-1
- [x] zero-2
- [ ] zero-3

PR #4654, opened 2025-12-10
Move `AutoModelForCausalLMWithValueHead` and `AutoModelForSeq2SeqLMWithValueHead` to experimental

qgallouedec:
Move these two classes to `trl.experimental.ppo`.

This PR also removes `SUPPORTED_ARCHITECTURES`. This was added early in the project https://github.com/huggingface/trl/pull/58, but was unused since PPO was refactored, and never really intended to be part of the public API. I think it's safe to remove without any deprecation.

It also moves `create_reference_model` to `model.utils`, for two reasons: it doesn't really make sense to have it in `models.modeling_base`; it avoids circular imports


docs/source/model_utils.md
```diff
@@ -11,3 +11,7 @@
 ## disable_gradient_checkpointing
 
 [[autodoc]] models.utils.disable_gradient_checkpointing
+
+## create_reference_model
+
+[[autodoc]] create_reference_model
```
qgallouedec, line 13: moved from the deleted `models.md`

PR #4682, opened 2025-12-11
Fix GRPO config validation in case `num_generations_eval` is specified and different than `num_generations`

apalmas-saifh:
# Fix evaluation batch size validation to use `num_generations_eval` when specified

### Problem:
The validation check in `GRPOConfig.__post_init__` (lines 830-836) (https://github.com/huggingface/trl/blob/1cbd346d072c289c74012c4a5bd67f424a242383/trl/trainer/grpo_config.py#L832)
always validates the evaluation batch size divisibility against `num_generations`, even when `num_generations_eval` is explicitly set.

### Solution:
Modified the check to use `num_generations_eval` when it's set, falling back to `num_generations` only when `num_generations_eval` is None.

### Impact:
  - Fixes incorrect validation errors when users set different generation counts for training vs evaluation
  - Allows proper use of the `num_generations_eval` parameter

###  Changes:
  - Added logic to determine the correct generation count for evaluation validation
  - Updated error message to be more accurate

## Before submitting
      Pull Request section?
      to it if that's the case.


qgallouedec (approved):
thanks!

PR #4684, opened 2025-12-12
Align import utils with transformers

qgallouedec:
Same as https://github.com/huggingface/transformers/pull/41268


albertvillanova (approved):
Thanks.

PR #4689, opened 2025-12-13
feat: DeepSeek V3.2 Off-policy sequence masking

### LINKED ISSUE #4697: feat GRPOTrainer: Off-policy sequence masking from DeepSeek V3.2
### Feature request

Hello,

I proposed implementing the Off-policy sequence masking depicted in the p.7-8 of the DeepSeek V3.2 paper.
See the extract below:

Description of your image

### Motivation

A challenge in hybrid RL trainings, where the rollout and grad updates aren't performed within a single contained framework, is minimizing the drift induced by dual frameworks. Different methods tackle this problem... such as routing replay (in the case of MoEs) mentioned by @qgallouedec in https://github.com/huggingface/transformers/issues/42638.  

Off-policy sequence masking is another method with the same goal and therefore adds more diversity for TRL users to improve the efficiency and stability of their training.

It is also an additional step to mimic the DeepSeek V3.2 GRPO scaling, alongside the Unbiased KL Estimate already implemented in TRL: https://github.com/huggingface/trl/issues/4637

Off-policy sequence masking isn't tied to the specific GRPO loss DeepSeek used, thus enabling users to combine it and experiment with different configurations.

### Your contribution

A PR has been submitted

casinca:
# What does this PR do?

Fixes #4697

This PR aims to implement the Off-policy sequence masking from the DeepSeek V3.2 paper

## Before submitting
      Pull Request section?
      to it if that's the case.

## Changes

- added an helper `_get_off_policy_mask` staticmethod to compute the off-policy mask
- ~~added logic in `_compute_loss` to inject the off-policy mask into the loss mask~~
- added logic in `_compute_loss` to inject the off-policy mask only to the surrogate loss
- added `__post_init__` checks and some tests (including not allowing with Lieger loss atm)
- added docs

## Who can review?

### DIFF
--- a/docs/source/paper_index.md
+++ b/docs/source/paper_index.md
@@ -418,6 +418,7 @@ training_args = GRPOConfig(
 **📜 Paper**: https://huggingface.co/papers/2512.02556
 
 DeepSeek-V3.2 technical report introduces several techniques to enhance the performance of GRPO. In TRL we implement:
+
 - The **Unbiased KL Estimate**, which corrects the K3 estimator (as used in the original GRPO implementation) to obtain an unbiased KL estimate using the importance-sampling
 ratio between the current policy  \\( \pi_\theta \\) and the behavior policy  \\( \pi_{\text{old}} \\).
 


casinca:
- @qgallouedec 👋 I'd like to get your atttention on this commit: https://github.com/huggingface/trl/pull/4689/commits/861b490b4535e0e416a10387c931bbab099d724d

    Initially I had made a mistake by injecting the off-policy mask into the loss mask early. So I had changed the loss mask to be computed upstream, once.
Even if now, with the correct logic, I'm not touching the loss mask, I wanted to understand why is the loss mask recomputed multiple times? I don't see it being modified/overwritten anywhere downstream. In any case I can revert this specific commit, it won't change my implementation.

- Additionally if the PR is good, I could add a metric to log for users, something like a off policy sequence drop ratio, so that they aren't blind and also get a better feel to adjust the delta/`off_policy_treshold` hparam

docs/source/paper_index.md
```diff
@@ -443,6 +444,41 @@ training_args = GRPOConfig(
...
+\end{cases}
+$$
+
+This mask is then applied to the GRPO loss as follows:
+
+$$
+\mathcal{L}_{GRPO}(\theta) = -\mathbb{E} \left[ \frac{1}{G} \sum_{i=1}^G \frac{1}{|o_i|} \sum_{t=1}^{|o_i|} \left( \min \left( \frac{\pi_\theta(o_{i,t} \dots)}{\pi_{\text{old}}(o_{i,t} \dots)} \hat{A}_{i,t}, \text{clip}(\dots) \hat{A}_{i,t} \right) \textcolor{red}{M_{i,t}} - \beta D_{KL} \right) \right]
+$$
+
+To enable this feature, use the `off_policy_mask_threshold` (corresponding to $\textcolor{blue}{\delta}$) in the [`GRPOConfig`]:
```
```suggestion
To enable this feature, use the `off_policy_mask_threshold` (corresponding to  \\( \textcolor{blue}{\delta}$ \\) in the [`GRPOConfig`]:
```

docs/source/paper_index.md
```diff
@@ -443,6 +444,41 @@ training_args = GRPOConfig(
...
+    ...,
+    off_policy_mask_threshold=0.5, 
+)
+```
+
+While the paper doesn't specify a $\textcolor{blue}{\delta}$ value used, a good starting point could be
+$\textcolor{blue}{\delta} = 0.5$. If training seems too conservative or too many sequences are masked, you can increase
+the value.  
+For reference, $\textcolor{blue}{\delta} = 1.0$ corresponds to an average log-ratio divergence of 1 nat per token, i.e.
+on sequences where this threshold is exceeded, the old policy was on average $e^1 \approx 2.7$ times more likely to
```
```suggestion
While the paper doesn't specify a  \\( \textcolor{blue}{\delta} \\) value used, a good starting point could be. \\( \textcolor{blue}{\delta} = 0.5 \\). If training seems too conservative or too many sequences are masked, you can increase
the value.  
For reference, \\( \textcolor{blue}{\delta} = 1.0 \\) corresponds to an average log-ratio divergence of 1 nat per token, i.e.
on sequences where this threshold is exceeded, the old policy was on average  \\( e^1 \approx 2.7 \\) times more likely to
```

tests/test_grpo_trainer.py
```diff
@@ -1956,6 +1956,115 @@ def test_single_reward_model_with_single_processing_class(self):
...
+            GRPOConfig(output_dir="tmp", off_policy_mask_threshold=-0.1)
+
+    def test_liger_kernel_compatibility_with_off_policy_masking(self):
+        """Test that use_liger_kernel and off_policy_mask_threshold are incompatible. At least not at the moment."""
+        with pytest.raises(ValueError, match="does not support"):
+            GRPOConfig(
+                    output_dir="tmp",
+                    use_liger_kernel=True,
+                    off_policy_mask_threshold=0.1
+                )
```
qgallouedec: I think a xfail would be better in this case. Just copy `test_training_with_off_policy_mask` into a `test_training_with_off_policy_mask_with_liger`, and add the xfail decorator:

```python
@require_liger_kernel
@pytest.mark.xfail(reason="Off-Policy Masking isn't compatible with Liger yet.")
def test_training_with_off_policy_mask_with_liger(self):
```

trl/trainer/grpo_trainer.py
```diff
@@ -2152,13 +2153,36 @@ def get_sapo_token_loss(unclipped_token_loss: torch.Tensor, temperature: float)
         sapo_token_loss = sigmoid_smoothed_loss * 4 / temperature
         return sapo_token_loss
 
+    @staticmethod
+    def _get_off_policy_mask(
```
```suggestion
    def get_off_policy_mask(
```

I think it makes more sense to have it public in this case (like `get_sapo_token_loss`)

qgallouedec (commented):
LGTM! Thanks! Just a few nits, but otherwise this looks good to me and I think it's a great addition!

docs/source/paper_index.md
qgallouedec: It's because our renderer use this syntax, and don't support inline `$`

casinca: ah yes my bad, I forgot about that, especially since I encountered this situation in a previous docs PR. Opposite
behavior from the IDE viewer.

casinca:
> LGTM! Thanks! Just a few nits, but otherwise this looks good to me and I think it's a great addition!

Applied all the suggestions, thanks. Let me know if I add something to the log section, like the ratio of sequences being masked in a minibatch `ratio_masked = 1.0 - torch.mean(off_policy_mask.float())`

qgallouedec (approved):
lgtm! Thanks!

LeonEricsson:
As implemented, we rely on `old_per_token_logps` to compute the ratio in Eq. (9), which I do not think matches the paper’s intent. Below is why, and my suggestion for how this method could integrate more generally into TRL. Would appreciate a discussion around this.

According to the paper, DeepSeek uses **off-policy sequence masking** to mitigate off-policy effects caused by both (1) policy staleness and (2) training–inference mismatch:

> “Additionally, inference frameworks used for efficient data generation are often highly optimized, which may differ in implementation details from training frameworks. Such training-inference inconsistency further exacerbates the degree of off-policyness.”

A key detail is that they set $\pi_{\text{old}}$ to the *inference-time sampling distribution*:

> “Note that $\pi_{\text{old}}$ here denotes the sampling probability directly returned by the inference framework, thus the KL divergence between the old and current policy accounts for both sources of off-policyness mentioned above.”

This can be understood as sequence masking based on an importance ratio of the form

$$
\frac{\pi_{\text{train}}(y \mid x)}{\pi_{\text{infer}}(y \mid x)}.
$$

Concretely, if we interpret Eq. (9) as $\pi_{\text{old}} \equiv \pi_{\text{infer}}$ and $\pi_\theta \equiv \pi_{\text{train}}$, and define the usual per-token importance ratio (as in TIS/MIS)

$$
\rho_t \triangleq \frac{\pi_{\text{train}}(o_t \mid q)}{\pi_{\text{infer}}(o_t \mid q)},
$$

then Eq. (9) can be rewritten as

$$
\frac{1}{|o_i|}\sum_{t=1}^{|o_i|}\log\frac{\pi_{\text{old}}(o_t\mid q)}{\pi_\theta(o_t\mid q)}
= \frac{1}{|o_i|}\sum_{t=1}^{|o_i|}\log\rho_t^{-1}
= -\frac{1}{|o_i|}\sum_{t=1}^{|o_i|}\log\rho_t
= -\log\left(\prod_{t=1}^{|o_i|}\rho_t\right)^{1/|o_i|}.
$$

So the statistic being thresholded is the negative log of the **geometric mean** of the per-token importance ratios.

DeepSeek’s masking rule is then a one-sided threshold on this quantity:

$$
-\log\left(\prod_{t=1}^{|o_i|}\rho_t\right)^{1/|o_i|} > \delta.
$$

The main advantage of using the geometric mean is that it captures an *average per-token divergence* that is much less sensitive to sequence length. In contrast, a sequence-level importance ratio (like our implemented sequence-MIS) based on $\prod_t \rho_t$ can scale (or vanish) exponentially with $|o_i|$.

For more background on geometric sequence masking, see: [https://richardli.xyz/post/rl-collapse-part3/](https://richardli.xyz/post/rl-collapse-part3/).

Given the strong relation to existing importance sampling approaches, my suggestion is to treat **geometric sequence masking** as a first-class method alongside existing training–inference importance sampling approaches, and to view the method in this PR as a variant within that family.

casinca:
Thanks @LeonEricsson , this was a good read, the blog is very interesting. DeepSeek didn't link that the arithmetic mean of logprobs is the log of the geometric mean of probs (probably implied). It's funny that they ended up doing the same thing by themselves since the blog was published only 1 month before their paper.

&nbsp;

Concerning your 2 observations, let me start with the 2nd one.

I agree with you, after reading the blog, that OPSM (off policy seq masking) is a unidirectional/single side subvariant of the bidirectional/double side geo masking (Similar vibe to your original MIS implementation and my current IcePoP PR).

So yeah if we consider DeepSeek as doing the forward DKL (old|new), then geo mask is -DKL (new|old)

$$ 
\underbrace{\frac{1}{T} \sum \log \left( \frac{\pi_{\text{old}}}{\pi_{\text{new}}} \right)}_{\text{DeepSeek}} = - \underbrace{\frac{1}{T} \sum \log \left( \frac{\pi_{\text{new}}}{\pi_{\text{old}}} \right)}_{\text{Geo Mask}} 
$$

and the KL I computed can be re-used for the main geo mask for bi directionality/both sides.

&nbsp;

Concerning the 1st problem:

I knew that the numerator was the inference logprobs from the external framework ex: vllm, my mistake was thinking `old_per_token_logps` were the vllm logprobs. Now you made me realize I inverted,`old_per_token_logps` = local logprobs and `sampling_per_token_logps` = vllm ones.

Right now, the implementation is only helping for the 1st source of "off policyness" mentioned in the paper ie when num grad update >1 but not for the 2nd source you mentioned: drift created by hybrid training, which is the main reason for OPSM...

So first I'd like to open a PR solely to fix OPSM and make it compatible whether users use vllm or not, so they can still benefits from OPSM for the 1st source of off policyness (currently implemented).

What I think I'm going to do is pass vllm logprobs to `get_off_policy_mask` if they are not `None`, else i'm passing `old_per_token_logps`
In any case I think I'll need to expose the vllm logprobs `sampling_per_token_logps` in the `inputs` dict because they aren't passed to the `_compute_loss`.

But yeah before refactoring to a more general approach with your 2nd observation, I'd like to fix OPSM first. I can work on Geo mask on a later/2nd separate PR if you're busy.

Note that Deepseek OPSM isn't just about one side geo mask, it's bi conditional with adv $\hat{A}_{i,t}$ <0 too in eq. 9.
In any case I think we can't do Geo Mask inside the MIS/TIS part of the code, because MIS/TIS is using the ratio of old IS (local/ external vllm) but here we need the current/trained model logprobs, so we can't precompute upstream, we have to do it inside the `_compute_loss` I believe, but for now I'll fix OPSM.

LeonEricsson:
I will return to the discussion regarding the second observation later. For now, I agree that an immediate fix to OPSM is appropriate.

Alex-Songs:
hello，
What is the general setting for off_policy_mask_threshold?

LeonEricsson:
> hello， What is the general setting for off_policy_mask_threshold?

refer to the last part of the [paper index docs](https://huggingface.co/docs/trl/paper_index#deepseek-v32-pushing-the-frontier-of-open-large-language-models)

PR #4691, opened 2025-12-14
Fix KeyError with transformers 5.0.0+ where push_to_hub_token is removed

Manodeepray:
## What does this PR do?

Fixes a `KeyError: 'push_to_hub_token'` that occurs when using `SFTTrainer` with transformers 5.0.0+, where `push_to_hub_token` has been completely removed.

## Problem

In transformers 5.0.0rc1 and later, the `push_to_hub_token` parameter has been fully removed from `TrainingArguments`. When `SFTTrainer` converts a `TrainingArguments` object to `SFTConfig`, it tries to pop this non-existent key:
```python
dict_args.pop("push_to_hub_token")  # ❌ KeyError in transformers 5.0.0+
```

## Solution

Use `.pop()` with a default value to safely handle both versions:
```python
dict_args.pop("push_to_hub_token", None)  # Works with all previous versions
```

## Testing

Tested and confirmed working with:
- transformers 5.0.0rc1 (new version where bug occurs)
- transformers 4.44.0 (older version for backward compatibility)
- SFTTrainer initialization works without errors in both versions

This change is fully backward compatible and prepares TRL for the upcoming transformers 5.0.0 stable release.

## Before submitting
- [x] This PR fixes a bug
- [x] Tested with both transformers 4.x and 5.0.0rc1
- [x] Backward compatible

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -22,9 +22,11 @@ from typing import Any
 
 import torch
 import torch.nn as nn
+import transformers
 from accelerate import PartialState
 from accelerate.logging import get_logger
 from datasets import Dataset, IterableDataset
+from packaging.version import Version
 from transformers import (
     AutoProcessor,
     DataCollator,
@@ -60,8 +62,6 @@ from .utils import (
     remove_none_values,
     selective_log_softmax,
 )
-import transformers
-from packaging.version import Version
 
 
 if is_peft_available():


qgallouedec:
Thanks! I opted for an explicit version check instead, which will make it easier for us to know that we can remove this case when we require transformers >=5.0.0.

trl/trainer/sft_trainer.py
```diff
@@ -601,7 +603,8 @@ def __init__(
         elif isinstance(args, TrainingArguments) and not isinstance(args, SFTConfig):
             dict_args = args.to_dict()
             dict_args["hub_token"] = args.hub_token  # to_dict hides the hub_token
-            dict_args.pop("push_to_hub_token")
+            if not Version(transformers.__version__) < Version("5.0.0.dev0"):
+                dict_args.pop("push_to_hub_token")
```
dodziraynard, line 607: The PR descriptions says use pop with default arg but there's no default arg specified here?
This is still causing an issue for me. 
Am I missing something?

albertvillanova: Thanks for reporting, @dodziraynard.

I think a logic error was introduced in: https://github.com/huggingface/trl/pull/4691/commits/7c9c753000e227af62a2b9b6bb9f32a5d0ad6cff

albertvillanova: Proposed fix:
- https://github.com/huggingface/trl/pull/4942

PR #3494, opened 2025-05-26
[GRPO] disabling top_k sampling default

kashif:
# What does this PR do?

Disabling `top_k` in GRPO sampling as a default and setting the sampling temperature default to `1.0`

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #4695, opened 2025-12-15
Fix top_k default value to 0 for disabling top-k filtering

albertvillanova:
Fix top_k default value to 0 for disabling top-k, instead of ambiguous (and non-documented) `None`.

See related comment: 
- https://github.com/huggingface/transformers/pull/42702#issuecomment-3655254057

For context, I was testing this fix in `transformers`:
- https://github.com/huggingface/transformers/pull/42702

and I discovered this edge case for `tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm[trl-internal-testing/tiny-Qwen2VLForConditionalGeneration]`:
- model has `top_k=1` in its `generation_config.json`
- we set `top_k=None` by default, so k-filtering is disabled
- as the latter is `None`, it gets silently overwritten to `1`

Therefore, we need this PR so the transformers fix works as expected once merged.

Not sure about the expected default behavior:
- since now, we were setting default value to `None`, saying this disables top-k filtering
- however, transformers default value is `50`

Should I set 50 instead of 0 as our default?

### DIFF
--- a/trl/experimental/online_dpo/online_dpo_config.py
+++ b/trl/experimental/online_dpo/online_dpo_config.py
@@ -67,7 +67,7 @@ class OnlineDPOConfig(TrainingArguments):
         top_p (`float`, *optional*, defaults to `1.0`):
             Float that controls the cumulative probability of the top tokens to consider. Must be in (0, 1]. Set to
             `1.0` to consider all tokens.
-        top_k (`int`, defaults to `0`):
+        top_k (`int`, *optional*, defaults to `0`):
             Number of highest probability vocabulary tokens to keep for top-k-filtering. If `0`, top-k-filtering is
             disabled and all tokens are considered.
         min_p (`float`, *optional*):
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -81,7 +81,7 @@ class GRPOConfig(TrainingArguments):
         top_p (`float`, *optional*, defaults to `1.0`):
             Float that controls the cumulative probability of the top tokens to consider. Must be in (0, 1]. Set to
             `1.0` to consider all tokens.
-        top_k (`int`, defaults to `0`):
+        top_k (`int`, *optional*, defaults to `0`):
             Number of highest probability vocabulary tokens to keep for top-k-filtering. If `0`, top-k-filtering is
             disabled and all tokens are considered.
         min_p (`float`, *optional*):
--- a/trl/trainer/rloo_config.py
+++ b/trl/trainer/rloo_config.py
@@ -78,7 +78,7 @@ class RLOOConfig(TrainingArguments):
         top_p (`float`, *optional*, defaults to `1.0`):
             Float that controls the cumulative probability of the top tokens to consider. Must be in (0, 1]. Set to
             `1.0` to consider all tokens.
-        top_k (`int`, defaults to `0`):
+        top_k (`int`, *optional*, defaults to `0`):
             Number of highest probability vocabulary tokens to keep for top-k-filtering. If `0`, top-k-filtering is
             disabled and all tokens are considered.
         min_p (`float`, *optional*):


qgallouedec:
> Should I set 50 instead of 0 as our default?

We should use 0, see https://github.com/huggingface/trl/pull/3494


--

Can you also update the docstring? like here

https://github.com/albertvillanova/trl/blob/21cf0a9a9680b9c41128828eb04a827e0565b6c0/trl/trainer/grpo_config.py#L84-L86

albertvillanova:
I am fixing vLLM `top_k`l...

qgallouedec:
After checking vLLM, `0` is now a valid value for `top_k`. The most recent version that disallowed `0` was 0.8.5, see

https://github.com/vllm-project/vllm/blob/3015d5634e74d59704e2b39bab0dbe2e6f86a38a/vllm/sampling_params.py#L411-L413

trl/experimental/online_dpo/online_dpo_trainer.py
```diff
@@ -506,7 +506,7 @@ def __init__(
                 "repetition_penalty": self.repetition_penalty,
                 "temperature": self.temperature,
                 "top_p": self.top_p,
-                "top_k": -1 if self.top_k is None else self.top_k,
+                "top_k": -1 if not self.top_k else self.top_k,
```
qgallouedec: I think can (should) just use `self.top_k`, and don't replace by `-1`

https://github.com/vllm-project/vllm/blob/855b101d75d2fc1fa02a47a6fcfa4053e8541cf0/vllm/sampling_params.py#L393C5-L393C54

albertvillanova: Yes, the vLLM support to consider all tokens by setting top_k=0 (besides -1) was introduced in vllm-0.9.0
- https://github.com/vllm-project/vllm/pull/17773

and we require vllm>=0.10.2.

albertvillanova: On the other hand, what if the users still set `top_k=None`? Maybe we should start a deprecation cycle.

qgallouedec: we could indeed deprecate `None` for 1 version before removal. I'd do it only for GRPO and RLOO though

trl/experimental/online_dpo/online_dpo_config.py
```diff
@@ -67,8 +67,8 @@ class may differ from those in [`~transformers.TrainingArguments`].
         top_p (`float`, *optional*, defaults to `1.0`):
             Float that controls the cumulative probability of the top tokens to consider. Must be in (0, 1]. Set to
             `1.0` to consider all tokens.
-        top_k (`int`, *optional*):
-            Number of highest probability vocabulary tokens to keep for top-k-filtering. If `None`, top-k-filtering is
+        top_k (`int`, defaults to `0`):
```
qgallouedec:
```suggestion
        top_k (`int`, *optional*, defaults to `0`):
```

trl/trainer/grpo_config.py
```diff
@@ -81,8 +81,8 @@ class GRPOConfig(TrainingArguments):
         top_p (`float`, *optional*, defaults to `1.0`):
             Float that controls the cumulative probability of the top tokens to consider. Must be in (0, 1]. Set to
             `1.0` to consider all tokens.
-        top_k (`int`, *optional*):
-            Number of highest probability vocabulary tokens to keep for top-k-filtering. If `None`, top-k-filtering is
+        top_k (`int`, defaults to `0`):
```
qgallouedec:
```suggestion
        top_k (`int`, *optional*, defaults to `0`):
```

trl/trainer/rloo_config.py
```diff
@@ -77,8 +78,8 @@ class RLOOConfig(TrainingArguments):
         top_p (`float`, *optional*, defaults to `1.0`):
             Float that controls the cumulative probability of the top tokens to consider. Must be in (0, 1]. Set to
             `1.0` to consider all tokens.
-        top_k (`int`, *optional*):
-            Number of highest probability vocabulary tokens to keep for top-k-filtering. If `None`, top-k-filtering is
+        top_k (`int`, defaults to `0`):
```
qgallouedec:
```suggestion
        top_k (`int`, *optional*, defaults to `0`):
```

PR #4700, opened 2025-12-16
Refactor vLLM generation [1/N]: Extract vLLM generation

albertvillanova:
Refactor vLLM generation [1/N]: Extract vLLM generation.

This PR introduces a new initialization module for vLLM generation in TRL trainers. The main change is the addition of conditional support for the `VLLMGeneration` backend, which will only be imported and exposed if the `vllm` dependency is available.

### DIFF
--- a/trl/generation/vllm_generation.py
+++ b/trl/generation/vllm_generation.py
@@ -265,17 +265,17 @@ class VLLMGeneration:
                 self.vllm_client.init_communicator(device=torch.cuda.current_device())
 
         elif self.mode == "colocate":
-            # Make sure vllm_tensor_parallel_size group size evenly divides the world size - each group should have
+            # Make sure tensor_parallel_size group size evenly divides the world size - each group should have
             # the same number of ranks
             if not accelerator.num_processes % self.tensor_parallel_size == 0:
                 raise ValueError(
-                    f"vllm_tensor_parallel_size ({self.tensor_parallel_size}) must divide world size "
+                    f"tensor_parallel_size ({self.tensor_parallel_size}) must divide world size "
                     f"({accelerator.num_processes}) evenly."
                 )
 
             if self.tensor_parallel_size > 1:
-                # Create subgroups of ranks for TP, each group with `vllm_tensor_parallel_size` ranks.
-                # For example, if world_size=8 and vllm_tensor_parallel_size=2 → groups: [0,1], [2,3], [4,5], [6,7]
+                # Create subgroups of ranks for TP, each group with `tensor_parallel_size` ranks.
+                # For example, if world_size=8 and tensor_parallel_size=2 → groups: [0,1], [2,3], [4,5], [6,7]
                 self.tp_group, _ = torch.distributed.new_subgroups_by_enumeration(
                     [
                         list(range(i * self.tensor_parallel_size, (i + 1) * self.tensor_parallel_size))


trl/generation/vllm_generation.py
```diff
@@ -0,0 +1,522 @@
...
+                max_model_len = args.vllm_max_model_length
+            elif (
+                hasattr(self.trainer, "max_prompt_length")
+                and self.trainer.max_prompt_length is not None
+                and hasattr(self.trainer, "max_completion_length")
+                and self.trainer.max_completion_length is not None
+            ):
+                max_model_len = self.trainer.max_prompt_length + self.trainer.max_completion_length
+            else:
+                max_model_len = None
```
albertvillanova: This will not be necessary once we merge:
- #4702

trl/generation/vllm_generation.py
```diff
@@ -0,0 +1,522 @@
...
+                "max_num_batched_tokens": 4096,
+                "model_impl": args.vllm_model_impl,
+                "enable_sleep_mode": args.vllm_enable_sleep_mode,
+                "quantization": vllm_quantization,
+            }
+
+            # TODO: improve
+            # Add logprobs_mode only for GRPO (not used in RLOO)
+            if "grpo" in type(self.trainer).__name__.lower():
+                llm_kwargs["logprobs_mode"] = "processed_logprobs"
```
albertvillanova: I am also thinking how to improve this.

trl/generation/vllm_generation.py
```diff
@@ -0,0 +1,522 @@
...
+    from trainers into a separate, testable class.
+    """
+
+    def __init__(self, trainer):
+        """Initialize vLLM generation.
+
+        Args:
+            trainer: Reference to parent trainer for accessing config, model, accelerator, etc.
+        """
+        self.trainer = trainer
```
qgallouedec: I like the idea of having a class dedicated to generation for vLLM now that the amount of code related to generation is starting to saturate the online trainers.

However I'm a bit annoyed by this back-reference. It's not an ideal design choice in my opinion. In this state, I even have trouble seeing how to properly test this.

Conceptually, the generator doesn't need to have a trainer to generation. It only needs the generation parameters, probably the `accelerator`, and an method to update the weights.

Can we think of a alternative design, something like

```python
class VLLMGeneration:
    def __init__(self, model_id, accelerator, mode="server", ...):
        ...

    def sync_weights(self, model):
        ...

    def generate(self, prompts, temperature, ...):
        ...
```

albertvillanova: Thanks for you insightful review! :hugs: 

Yes, I totally agree with you that the back-reference to the trainer is not ideal.

Indeed, this PR is a preliminary step in the complete refactoring and I was planning to change that on a later PR. This was my original plan:
  1. PR 1 (current): Extract the vLLM logic into a separate class (prove the concept, remove duplication)
  2. PR 2: Refine the interface to use explicit parameters instead of trainer reference
  3. PR 3: Add a proper protocol/interface

However, I could fix this now in this PR if you think it is better.

trl/generation/vllm_generation.py
albertvillanova: I have removed it after the merge.

trl/generation/vllm_generation.py
albertvillanova: I think we should pass `logprobs_mode` in the RLOOTrainer as well, after having read the motivation of the PR that introduced this:
- #4163

I am reverting this condition.

albertvillanova: Done.

trl/generation/vllm_generation.py
albertvillanova: I have seen that the `trainer` instance is used by the `rollout_func`. This will need further refactoring to separate (but coordinate) rollout_func and vllm functionalities. :thinking:

albertvillanova: See related PR:
- #4712

qgallouedec: > I have seen that the `trainer` instance is used by the `rollout_func`. This will need further refactoring to separate (but coordinate) rollout_func and vllm functionalities. 🤔

Yes, indeed. The design of `rollout_func` is really not ideal. We implemented it that way initially because we had to move quickly for the release of OpenEnv, but in my opinion it's really something we should completely rethink and refactor.

qgallouedec: > Indeed, this PR is a preliminary step in the complete refactoring and I was planning to change that on a later PR. This was my original plan:

I'm fine with this multi-stage PR for this. It will probably be easier to review this way, and for you to write

albertvillanova: Additionally, I see that the `trainer` instance is also used by the `profiling_context` function. This will need further refactoring as well.

albertvillanova: I created a dedicated PR to decouple profiling from trainer:
- #4717

albertvillanova:
As this PR was not approved, I have finally addressed both issues here:
- trainer reference within rollout_func
- trainer reference within profiling_context

This PR needs merging this PR first though:
- #4717

albertvillanova:
I have integrated PR #4717.
CC: @qgallouedec

trl/generation/vllm_generation.py
```diff
@@ -0,0 +1,552 @@
...
+    """Handles vLLM-based generation for trainers.
+
+    Extracts all vLLM-specific logic (initialization, generation, weight sync)
+    from trainers into a separate, testable class.
+    """
+
+    def __init__(
+        self,
+        model,
+        args,
```
qgallouedec: what do you think of removing args and have instead

- `enable_sleep_mode`
- `gpu_memory_utilization`
- `group_port`
- `guided_decoding_regex`
- `max_model_length`
- `mode`
- `model_impl`
- `server_base_url`
- `server_host`
- `server_timeout`
- `tensor_parallel_size`

in the argument of this function?

you may want to add `max_num_seqs` also

albertvillanova: Thanks, @qgallouedec.

I had something similar in mind in my original plan:
- PR 2: Refine the interface
  - I had in mind to create something like a `VLLMGenerationConfig` class with all present (and open for future) VLLM parameters.

If you prefer, I can do all the refactoring in a single PR?

albertvillanova: @qgallouedec I think I addressed your suggestion above.

trl/trainer/rloo_trainer.py
```diff
@@ -95,22 +89,14 @@
...
 
 if is_vllm_available():
-    import vllm
-    from vllm import LLM, SamplingParams
-
-    if Version(vllm.__version__) <= Version("0.10.2"):
-        from vllm.sampling_params import GuidedDecodingParams
-    else:
-        from vllm.sampling_params import StructuredOutputsParams
+    from ..generation.vllm_generation import VLLMGeneration
```
qgallouedec: `VLLMGeneration` will exist in all cases, I think we can safely import it, even when vllm is not installed

trl/generation/__init__.py
qgallouedec, line 1: not necessarily for this PR, but since we add a generation `submodule`, we could have `vllm_client` in this submodule as well

trl/generation/vllm_generation.py
```diff
@@ -0,0 +1,610 @@
...
+        min_p: float | None = None,
+        repetition_penalty: float | None = None,
+        max_completion_length: int | None = None,
+        # Chat/tool configuration
+        chat_template_kwargs: dict | None = None,
+        tools: list | None = None,
+        chat_template: str | None = None,
+        rollout_func=None,
+    ):
+        """Initialize vLLM generation.
```
qgallouedec: nit, but for consistency with the rest of the lib, instead of

```python
class A:
    """
    This is A.
    """

    def __init__(self, a):
        """
        Init A.

        Args:
            a: This is a.
        """
        ...
```

you could use

```python
class A:
    """
    This is A.

    Args:
        a: This is a.
    """

    def __init__(self, a):
        ...
```

trl/generation/vllm_generation.py
```diff
@@ -0,0 +1,610 @@
...
+
+            # Build LLM initialization kwargs
+            llm_kwargs = {
+                "model": model.name_or_path,
+                "tensor_parallel_size": self.tensor_parallel_size,
+                "gpu_memory_utilization": self.gpu_memory_utilization,
+                "max_num_seqs": self.max_num_seqs,
+                "max_model_len": self.max_model_length,
+                "distributed_executor_backend": "external_launcher",
+                "seed": accelerator.process_index // self.tensor_parallel_size,
```
qgallouedec: imo, this comment is important to keep
```suggestion
                # Feed identical seed for tp groups to ensure sampling results are the same across workers
                "seed": accelerator.process_index // self.tensor_parallel_size,
```

trl/generation/vllm_generation.py
```diff
@@ -0,0 +1,610 @@
...
+        """FSDP2-specific parameter synchronization (lines 1025-1046)."""
+        accelerator = self.accelerator
+        model = self.model
+
+        # For FSDP2, module.state_dict() already covers all parameters, so no need for recursion
+        for name, param in module.state_dict().items():
+            # When using PEFT, we need to recover the original parameter name
+            name = name.removeprefix("base_model.model.").replace(".base_layer", "")
+            # Skip PEFT layers: they don't exist in vLLM, and they are merged already.
+            if is_peft_model(model) and model.prefix in name:
```
qgallouedec: this used to be

```python
if is_peft_model(module) and module.prefix in name:
```

I'm not sure to understand why we have to change `module` -> `model`

qgallouedec: AH, it probably doesn't matter because module=model. Though I think it would be clearer not to mix, and don't access `self.model` in this method

trl/generation/vllm_generation.py
```diff
@@ -0,0 +1,610 @@
...
+                    visited.add(full_name)
+
+                    if self.mode == "server" and accelerator.is_main_process:
+                        self.vllm_client.update_named_param(full_name, param.data)
+                    elif self.mode == "colocate":
+                        llm_model = self.llm.llm_engine.model_executor.driver_worker.model_runner.model
+                        llm_model.load_weights([(full_name, param.data)])
+
+    def _sync_fsdp2_params_to_vllm(self, module):
+        """FSDP2-specific parameter synchronization (lines 1025-1046)."""
```
qgallouedec: Do these line numbers refer to the grpo trainer file before the PR? if so I'd remove any mention of them.

albertvillanova:
Oh, thanks a lot for your detailed review!!! :hugs: 
I will address all your requests.

trl/generation/vllm_generation.py
albertvillanova: I added this and the comment below.

trl/generation/vllm_generation.py
albertvillanova: You are right. Done.

trl/generation/vllm_generation.py
```diff
@@ -0,0 +1,616 @@
...
+                "seed": accelerator.process_index // self.tensor_parallel_size,
+                # Latest vLLM v1 memory profiler is misled by the high default value (i.e., 32768) - thinking there's not enough memory
+                "max_num_batched_tokens": 4096,
+                "model_impl": self.model_impl,
+                "enable_sleep_mode": self.enable_sleep_mode,
+                # Important so temperature scaling/logit tweaking affects the TIS log probs
+                "logprobs_mode": "processed_logprobs",
+                "quantization": quantization,
+            }
+            self.llm = LLM(**llm_kwargs)
```
qgallouedec: nit

instead of 

```python
llm_kwargs = {
    "model": model.name_or_path,
    ...
}
self.llm = LLM(**llm_kwargs)
```

we can also simply do


```python
self.llm = LLM(
    model=model.name_or_path,
    ...,
)
```

PR #4702, opened 2025-12-16
Align use of vllm_max_model_length in RLOOTrainer

albertvillanova:
Align use of `vllm_max_model_length` in RLOOTrainer.

Align it with `GRPOTrainer`, introduced in PR:
- #4300

This alignment is required to make simpler this PR:
- #4700


trl/trainer/rloo_config.py
```diff
@@ -144,6 +144,9 @@ class RLOOConfig(TrainingArguments):
             Control the GPU memory utilization for vLLM. This setting only applies when `vllm_mode` is set to
             `"colocate"`. If you are using `vllm_mode="server"`, this parameter must be passed separately when
             launching the vLLM server via the `--vllm_gpu_memory_utilization` flag.
+        vllm_max_model_length (`int`, *optional*, defaults to `None`):
```
qgallouedec:
```suggestion
        vllm_max_model_length (`int`, *optional*):
```

can you possibly apply the same in grpo config?

PR #4703, opened 2025-12-16
Deprecate max_prompt_length in RLOOTrainer

albertvillanova:
Deprecate max_prompt_length in RLOOTrainer.

Align the deprecation as in GRPOTrainer, introduced in:
- #4300

This alignment is required to make simpler this PR:
- #4700

### DIFF
--- a/trl/trainer/rloo_trainer.py
+++ b/trl/trainer/rloo_trainer.py
@@ -27,15 +27,13 @@ import datasets
 import pandas as pd
 import torch
 import torch.utils.data
-import transformers
-from accelerate import logging
+from accelerate.logging import get_logger
 from accelerate.utils import broadcast_object_list, gather, gather_object, is_peft_model, set_seed
 from datasets import Dataset, IterableDataset
 from torch import nn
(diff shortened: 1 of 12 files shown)


trl/trainer/rloo_trainer.py
```diff
@@ -527,7 +522,7 @@ def __init__(
                     max_num_seqs=self.args.per_device_train_batch_size
                     * self.vllm_tensor_parallel_size
                     * self.args.steps_per_generation,
-                    max_model_len=max_model_len,
+                    # max_model_len=self.args.vllm_max_model_length,  # TODO
```
qgallouedec: You want to add

https://github.com/huggingface/trl/blob/20cc2e1d364684eb1bad5dd305798bc1108da4fb/trl/trainer/grpo_config.py#L150

in this PR or in a next one?

albertvillanova: I did it in a previous PR: :sweat_smile: 
- #4702

See reference: 
- https://github.com/huggingface/trl/pull/4700#discussion_r2624090747

albertvillanova: Merged!

PR #4713, opened 2025-12-17
Disallow `PeftModel` + `peft_config` in trainers

qgallouedec:
## What does this PR do?

This PR prevents a configuration that produces an adapter that cannot be reloaded: training a new PEFT adapter from an already-PEFT-wrapped model while also passing a new PEFT config. In that setup, the trainer merges and unloads the existing adapter

https://github.com/huggingface/trl/blob/157cd63204a36b8220e45ce885ced06232f94fe3/trl/trainer/sft_trainer.py#L699-L702

but never saves the resulting base model, so `AutoPeftModelForCausalLM.from_pretrained` would later fail because the base model is missing.

To avoid this, we must disallow passing a PEFT model to the trainer together with a PEFT config.

**What is the user wants to train a new adapter starting from a PEFT model?**

Users can still train a new adapter starting from a PEFT model, but they must first merge and unload it themselves, save the resulting base model, and then pass that base model to the trainer along with the new PEFT config.

## Reproduce the issue

Asumme already saved `"base_adapter"`

```python
from peft import LoraConfig, get_peft_model

model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen3-0.6B")
lora_config = LoraConfig()
peft_model = get_peft_model(model, lora_config)
peft_model.save_pretrained("base_adapter")
```

Now, let's mimic what a the trainers do:

```python
from peft import LoraConfig, get_peft_model, AutoPeftModelForCausalLM

# The user loads the adapter, and passes it to the trainer along with a PEFT config
model = AutoPeftModelForCausalLM.from_pretrained("base_adapter")
lora_config = LoraConfig()

# This is what the trainer does
# Merge and unload the adapter, and apply a new PEFT config
model.merge_and_unload()
model = get_peft_model(model, lora_config)
# Trains, then saves the new adapter
model.save_pretrained("trained_adapter")
model.push_to_hub("trained_adapter", private=True)
```

Now let's try to load the trained adapter:

```python
from peft import AutoPeftModelForCausalLM
model = AutoPeftModelForCausalLM.from_pretrained("trained_adapter")
```

```
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/peft/auto.py", line 115, in from_pretrained
    base_model = target_class.from_pretrained(base_model_path, revision=base_model_revision, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: PeftModel.from_pretrained() missing 1 required positional argument: 'model_id'
```

Or from the hub:

```python
from peft import AutoPeftModelForCausalLM
model = AutoPeftModelForCausalLM.from_pretrained("qgallouedec/trained_adapter")
```

```
adapter_config.json: 100%|███████████████████████████████████████████████████████████████████████████████████| 1.03k/1.03k [00:00<00:00, 9.71MB/s]
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/peft/auto.py", line 115, in from_pretrained
    base_model = target_class.from_pretrained(base_model_path, revision=base_model_revision, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: PeftModel.from_pretrained() missing 1 required positional argument: 'model_id'
```

In both cases, loading fails for a simple reason: the base model was never saved, so the new adapter does not have a loadable base model.

Because of this, we should disallow passing a PEFT model to the trainer together with a PEFT config.

## In other words

The following table summarizes the different scenarios:

| Model input to trainer | PEFT config? | Before | After | Note |
| --- | --- | --- | --- | --- |
| Model | No |  |  | Nothing special here |
| Model | Yes | Create, train and save/push the adapter | Create, train and push the adapter | Nothing changes here |
| PEFT Model | No | Train and save/push the adapter | Train and save/push the adapter | Nothing changes here |
| PEFT Model | Yes | Merge and unload, create new adapter, train and save/push the new adapter | **Error** | The base model for the new adapter would not exist |


tests/test_dpo_trainer.py
```diff
@@ -589,59 +589,6 @@ def test_dpo_trainer_generate_during_eval_no_wandb(self):
...
-        # train the model
-        trainer.train()
-
-        # save peft adapter
-        trainer.save_model()
-
-        try:
-            AutoModelForCausalLM.from_pretrained(self.tmp_dir)
-        except OSError:
-            pytest.fail("Loading the saved peft adapter failed")
```
qgallouedec, line 643: The reason why this test was passing is actually quite subtle:

It's because 

1. we use the same config for the base adapter and for the trained adapter and
2. when you call (line 606)
    ```python
    model_peft = get_peft_model(model, lora_config)
    ```
    it modifies `lora_config` object and adds `base_model_name_or_path`, see [here](https://github.com/huggingface/peft/blob/89a75aa12a143f82132195b814026bd1aa1a165e/src/peft/mapping_func.py#L66).

So when you "re-apply" the same LoRA config, it mistakenly uses the base model path, while it should be `None` (since the base model actually doesn't exists in this case).

PR #4714, opened 2025-12-18
Fix test assertion for `top_k` parameter in `OnlineDPOTrainer`

qgallouedec:
Fixes a test broken after https://github.com/huggingface/trl/pull/4695


albertvillanova (approved):
Good catch! 😅 Thanks.

PR #4717, opened 2025-12-18
Refactor vLLM generation [3/N]: Decouple profiling from trainer

albertvillanova:
Refactor vLLM generation [3/N]: Decouple profiling from trainer.

Refactor profiling to remove the strong coupling to the trainer instance.

With this refactor, `profiling_context` now acts as a lightweight factory that builds a `ProfilingContext` instance populated with the relevant caller metadata (e.g. the trainer or other higher-level component). This object can be passed explicitly to composite attribute classes or downstream method calls, without requiring access to the full trainer. The callee can then enter the provided `ProfilingContext` as a context manager to measure and report execution duration in a uniform way, while remaining fully decoupled from the trainer implementation.

See relevant comment in:
- https://github.com/huggingface/trl/pull/4700#discussion_r2630637416

Before:
  - `profiling_context` was tightly coupled to `Trainer` class
  - Required passing entire trainer instance
  - Mixed timing and logging concerns

After:
  - Created `ProfilingContext` class that's completely decoupled from `Trainer`
  - Takes only the data it needs via constructor parameters
  - Clean separation of concerns

### DIFF
--- a/trl/extras/profiling.py
+++ b/trl/extras/profiling.py
@@ -154,7 +154,7 @@ def profiling_context(trainer: Trainer, name: str) -> ProfilingContext:
     ```
     """
     context_name = f"{trainer.__class__.__name__}.{name}"
-    step = trainer.state.global_step if hasattr(trainer, "state") else None
+    step = trainer.state.global_step
 
     return ProfilingContext(
         name=context_name,


qgallouedec (approved):
Thanks, in my understanding you don't need [1/N] and [2/N] to be merged before

trl/extras/profiling.py
```diff
@@ -55,27 +153,30 @@ def some_method(self):
...
-    start_time = time.perf_counter()
-    yield
-    end_time = time.perf_counter()
-    duration = end_time - start_time
-
-    profiling_metrics = {f"profiling/Time taken: {trainer.__class__.__name__}.{name}": duration}
-    if "wandb" in trainer.args.report_to and wandb.run is not None and trainer.accelerator.is_main_process:
-        wandb.log(profiling_metrics)
+    context_name = f"{trainer.__class__.__name__}.{name}"
+    step = trainer.state.global_step if hasattr(trainer, "state") else None
```
qgallouedec: maybe quick question here, when doesn't Trainer have a `state`?

albertvillanova: I'm fixing that... Just "forced" the possibility of being None, but it is not necessary indeed.

albertvillanova:
> Thanks, in my understanding you don't need [1/N] and [2/N] to be merged before

No, indeed I will update 1/N once this PR is merged.

PR #4718, opened 2025-12-18
RLOO supports async rewards.

pramodith:
# What does this PR do?
Similar to #4567 but for RLOO.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -362,7 +362,7 @@ class GRPOTrainer(BaseTrainer):
         self._has_async_reward_funcs = any(asyncio.iscoroutinefunction(func) for func in self.reward_funcs)
         if self._has_async_reward_funcs:
             self.async_reward_loop_thread, self.async_reward_loop, self.async_reward_loop_ready_event = (
-                start_event_loop_in_daemon(name="RLOOTrainer-AsyncRewardLoop")
+                start_event_loop_in_daemon(name="GRPOTrainer-AsyncRewardLoop")
             )
             # wait until the event loop is running in the daemon thread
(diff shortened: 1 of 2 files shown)


pramodith:
Not a huge fan of the duplication, ideally I'd like to see `_compute_rewards` become a utility or create something like

```
class RLVRTrainer(BaseTrainer):

     def _compute_rewards(..):
         .....

```

Both `GRPOTrainer` and `RLOOTrainer` can then inherit RLVRTrainer and if the implementations of _calculate_rewards diverge in the future they can override it. Let me know if you'd like to move to this approach instead of what I have currently.

pramodith:
Came up with a better approach aided by the coding agents! In the previous approach we create a new event loop every time since `asyncio.run` is called every time `_compute_rewards` is called.

In the new approach we create a new daemon thread if there any async reward functions. A new event loop set to run forever is now created on the daemon thread.

Every-time we need to run the async reward functions they now run on the daemon thread's event loop. This way we avoid any overhead each time `_compute_rewards`

qgallouedec:
> Both `GRPOTrainer` and `RLOOTrainer` can then inherit RLVRTrainer and if the implementations of _calculate_rewards diverge in the future they can override it. Let me know if you'd like to move to this approach instead of what I have currently.

I understand. We’ll keep the duplication approach for now. It makes it easier to move quickly on a single trainer without having to worry about integration constraints or side effects on all other trainers in the library.

Based on past experience with similar abstractions attempt, the long-term tradeoff tends to be unfavorable: abstraction brings relatively little benefit, while adding complexity and making future refactors harder to untangle if requirements diverge.

PR #4723, opened 2025-12-19
Improve PEFT integration

qgallouedec:
This PR makes four changes:

1. Fixes the type hints and documentation for the `model` argument in `SFTTrainer`, `GRPOTrainer`, `RLOOTrainer`, and `RewardTrainer` to allow `PeftModel`.
2. Uses `is_peft_model` instead of `isinstance(model, PeftModel)`. The two checks are equivalent, but `is_peft_model` is more explicit, so this PR adopts it consistently across the codebase.
3. Introduces `use_adapter`, a context manager that temporarily selects an adapter (this is useful for point 4).
4. For methods that use a `ref_model` (GRPO, RLOO), when the provided `model` is a `PeftModel`, the adapter is cloned under the name `"ref"`. Previously, the reference model was obtained by just disabling the adapter, which is only correct when the adapter is newly initialized—not when it has already been pretrained. An example of this issue can be found in #1193.

Note that we don't apply these changes for DPO as there is currently a refactoring in #3906

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1724,6 +1724,12 @@ class TestGRPOTrainer(TrlTestCase):
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
 
+    @pytest.mark.xfail(
+        Version(transformers.__version__).is_devrelease,  # Tests with dev dependencies
+        reason="Blocked by upstream liger-kernel bug (linkedin/Liger-Kernel#960); "
+        "fixed by linkedin/Liger-Kernel#966 but not yet released (>0.6.4 required)",
+        strict=True,
+    )
     @pytest.mark.parametrize(
         "model_id",
         [


trl/trainer/grpo_trainer.py
```diff
@@ -249,7 +251,7 @@ class GRPOTrainer(BaseTrainer):
 
     def __init__(
         self,
-        model: str | PreTrainedModel,
+        model: "str | PreTrainedModel | PeftModel",
```
edbeeching, line 263:
```suggestion
        model: str | PreTrainedModel | "PeftModel",
```

trl/trainer/rloo_trainer.py
```diff
@@ -224,7 +226,7 @@ class RLOOTrainer(BaseTrainer):
 
     def __init__(
         self,
-        model: str | PreTrainedModel,
+        model: "str | PreTrainedModel | PeftModel",
```
edbeeching, line 240:
```suggestion
        model: str | PreTrainedModel | "PeftModel",
```

trl/trainer/sft_trainer.py
```diff
@@ -580,7 +582,7 @@ class SFTTrainer(BaseTrainer):
 
     def __init__(
         self,
-        model: str | PreTrainedModel,
+        model: "str | PreTrainedModel | PeftModel",
```
edbeeching, line 585:
```suggestion
        model: str | PreTrainedModel | "PeftModel",
```

qgallouedec:
> I believe the type annotations could clearer with a small change, but otherwise LGTM

Agree, but it seems not correct:

```python
from peft import PeftModel

def func(model: str | "PeftModel"): ...
```

```
  File "/fsx/qgallouedec/trl/dem.py", line 3, in <module>
    def func(model: str | "PeftModel"): ...
                    ~~~~^~~~~~~~~~~~~
TypeError: unsupported operand type(s) for |: 'type' and 'str'
```

PR #4729, opened 2025-12-20
🎭 Up to 50% less VRAM during forward with `forward_masked_logits` function

qgallouedec:
## What this PR does?

This PR introduces `forward_masked_logits`, an new utility function that aims to reduces memory peak during the forward pass.

## Why?

Most post-training methods (DPO, GRPO, RLOO, SFT for completion-only training, …) compute the loss on the completion part of the sequence. So basically, doing something close to this:

```python
logits = model(input_ids).logits
logits[~completion_mask.bool()] = 0.0
```

In the above code, we compute the logits for the **full sequence**, and **then** we mask the prompt part.

It could save a lot a VRAM to have a utility function that would skip the `lm_head` based on a mask (it would be the completion mask here)

### Benchmark

With a 4B model:

memory_vs_seq_len

<details>
<summary>Script</summary>

```python
import os
from dataclasses import dataclass

import torch

from trl.trainer.utils import forward_masked_logits


@dataclass
class BenchResult:
    seq_len: int
    mask_ratio: float
    full_peak_gb: float
    masked_peak_gb: float


def _peak_memory(fn) -> float:
    torch.cuda.reset_peak_memory_stats("cuda")
    torch.cuda.synchronize("cuda")
    initial_allocated = torch.cuda.memory_allocated("cuda")
    fn()
    torch.cuda.synchronize("cuda")
    peak_allocated = torch.cuda.max_memory_allocated("cuda")
    peak_delta = max(0, peak_allocated - initial_allocated)
    return peak_delta / 1_000_000_000  # Convert to GB


def main() -> None:
    batch_size = 4
    seq_lens = [128, 256, 512, 1024, 2048, 4096, 8192]
    mask_ratios = [0.2, 0.4, 0.6, 0.8, 1.0]
    output_dir = "benchmarks/forward_masked_logits"

    os.makedirs(output_dir, exist_ok=True)
    model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen3-4B", dtype="auto", device_map="cuda")

    results = []
    base_ratio = mask_ratios[0]
    for seq_len in seq_lens:
        input_ids = torch.randint(0, 1000, (batch_size, seq_len), device="cuda")
        with torch.no_grad():
            full_peak = _peak_memory(lambda: model(input_ids=input_ids))
        for ratio in mask_ratios:
            mask = torch.rand(batch_size, seq_len, device="cuda") < ratio
            with torch.no_grad():
                masked_peak = _peak_memory(lambda: forward_masked_logits(model, mask, input_ids=input_ids))
            results.append(BenchResult(seq_len, ratio, full_peak, masked_peak))

    ratios = sorted({r.mask_ratio for r in results})
    seq_lens = sorted({r.seq_len for r in results})

    plt.figure()
    base_results = [r for r in results if r.mask_ratio == base_ratio]
    base_results.sort(key=lambda r: r.seq_len)
    full_mem = [r.full_peak_gb for r in base_results]
    plt.plot(seq_lens, full_mem, label="full forward", marker="o")
    for ratio in reversed(ratios):
        ratio_results = [r for r in results if r.mask_ratio == ratio]
        ratio_results.sort(key=lambda r: r.seq_len)
        masked_mem = [r.masked_peak_gb for r in ratio_results]
        plt.plot(seq_lens, masked_mem, label=f"masked ratio={ratio}", marker="o")
    plt.xlabel("Sequence length")
    plt.ylabel("Memory for the forward pass (GB)")
    plt.title("Memory vs sequence length (per mask ratio)")
    plt.xscale("log", base=2)
    plt.xticks(seq_lens, [str(x) for x in seq_lens])
    plt.legend()
    plt.grid(True)
    plt.tight_layout()
    plt.savefig(os.path.join(output_dir, "memory_vs_seq_len.png"))
    plt.close()


if __name__ == "__main__":
    main()
```

</details>

### Next step

Use this function in trainers

### DIFF
--- a/scripts/generate_tiny_models.py
+++ b/scripts/generate_tiny_models.py
@@ -18,6 +18,7 @@
 
 import torch
 from huggingface_hub import HfApi, ModelCard
+from peft import LoraConfig, get_peft_model
 from torch import nn
 from transformers import (
     AutoConfig,
@@ -112,8 +113,9 @@ def push_to_hub(model, tokenizer, generation_config, prefix=None, suffix=None, f
         print(f"Model {repo_id} already exists, skipping")
     else:
         model.push_to_hub(repo_id)
-        tokenizer.push_to_hub(repo_id)
         model_card.push_to_hub(repo_id)
+        if tokenizer is not None:
+            tokenizer.push_to_hub(repo_id)
         if generation_config is not None:
             generation_config.push_to_hub(repo_id)
 
@@ -380,3 +382,15 @@ for model_id, model_class, dtype in [
     config = AutoConfig.from_pretrained(model_id, text_config=text_config, vision_config=vision_config, **kwargs)
     model = model_class(config).to(dtype=dtype)
     push_to_hub(model, processor, generation_config, "tiny")
+
+# PEFT models
+model = Qwen3ForCausalLM.from_pretrained("trl-internal-testing/tiny-Qwen3ForCausalLM", dtype="auto")
+model = get_peft_model(model, LoraConfig())
+generation_config = GenerationConfig.from_pretrained("trl-internal-testing/tiny-Qwen3ForCausalLM")
+push_to_hub(model, None, None, "tiny")
+
+# Same model, but different weights
+model = Qwen3ForCausalLM.from_pretrained("trl-internal-testing/tiny-Qwen3ForCausalLM", dtype="auto")
+model = get_peft_model(model, LoraConfig())
+generation_config = GenerationConfig.from_pretrained("trl-internal-testing/tiny-Qwen3ForCausalLM")
+push_to_hub(model, None, None, "tiny", "2")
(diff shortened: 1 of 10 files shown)


pramodith:
Might need to be very careful of how this is used. The library has a lot of logit slicing,shifting, and downstream masking operations that would all be rendered invalid if this new utility function is used.

qgallouedec:
> Might need to be very careful of how this is used. The library has a lot of logit slicing,shifting, and downstream masking operations that would all be rendered invalid if this new utility function is used.

Yes, that's why I first add this utility, and then I'll add it to trainers, where it's possible in follow-up PRs

.github/workflows/tests.yml
```diff
@@ -239,7 +239,7 @@ jobs:
           uv pip install ".[dev]"
           uv pip install accelerate==1.4.0
           uv pip install datasets==3.0.0
-          uv pip install transformers==4.56.1
+          uv pip install transformers==4.56.2
```
qgallouedec, line 242: getter was broken in 4.56.1 (we need it to work for `.get_decoder()`) , see https://github.com/huggingface/transformers/pull/40824 and https://github.com/huggingface/transformers/releases/tag/v4.56.2

edbeeching (approved):

PR #4732, opened 2025-12-20
feat: Bidirectional masked importance sampling ratio (MIS) for IcePop

### LINKED ISSUE #4735: feat GRPOTrainer: Bidirectional MIS from IcePop loss (used in Xiaomi MiMo-V2-Flash and Ring-1T)
### Feature request

Hello,

MIS is already implemented in TRL, but only gives choice to cap for an upper bound.
A recent impressive model [Xiaomi MiMo-V2-Flash](https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash) and [inclusionAI Ring-1T](https://huggingface.co/inclusionAI) used MIS for their trainings but also masking their IS ratios that are below a certain threshold.

This MIS method, in having both an upper and lower bound, was labelled as [IcePop](https://ringtech.notion.site/icepop).

This method is mentioned in the `paper_index.md` but it can't be applied in TRL yet since there's no hparam for a lower bound.

### Motivation

- This is yet another step, similar to other methods, in reducing the logprobs delta between training-infer hybrid RL framework (especially in MoEs) and more stable training.
- Adding a lower bound hparam to mask the IS ratio, gives additional granularity to the users and is more truthful to the original IcePop method.

Which seems pretty effective (at least compared to TIS) from their findings, more details on benchmarks in their blogpost.

Image

&nbsp;

One might think a lower bound is a minor change but a little sweep they did shows some differences:

Image

- Since MIS is already implemented in TRL, this feature should be a minimal change

### Your contribution

A PR has been submitted

casinca:
# What does this PR do?
Please see more details in: resolve #4735

---

**Current TRL:**
image

**IcePop:**
image

---

~~Note: This lower bound hparam `vllm_importance_sampling_min` has been made available only for MIS and not TIS. 
First reason is that I'm following the original paper intentions and 2nd, unlike the upper bound utility with TIS, a lower bound would be undesirable since off policy ratios ex 0.0005 would be truncated to ex 0.5, which mean we would amplify the gradient of a bad sample. Hence the reasons I'm keeping this MIS only.~~

Edit: added for TIS as well, see https://github.com/huggingface/trl/pull/4732#issuecomment-3714895623

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

---

> [!NOTE]
> **Medium Risk**
> Changes default GRPO/vLLM gradient weighting when users set clip_min or migrate from cap; misconfigured bounds are rejected but behavior shifts for off-policy ratio correction during training.
> 
> **Overview**
> **vLLM importance sampling** now uses a **two-sided band** `[C_min, C_max]` instead of a single upper cap `vllm_importance_sampling_cap`, aligning TIS/MIS with IcePop-style bidirectional handling of train–inference ratio outliers.
> 
> `GRPOConfig` adds `vllm_importance_sampling_clip_min` and `vllm_importance_sampling_clip_max` (either may be `None` for one-sided behavior). The old cap is deprecated and mapped to `clip_max` in `__post_init__`, with validation that `clip_min < clip_max` and that truncate modes set at least one bound. **`GRPOTrainer`** (and experimental GFPO / replay-buffer trainers) apply `torch.clamp` for truncate modes and zero out ratios below min or above max for mask modes.
> 
> Docs in `grpo_trainer.md` and `paper_index.md` update formulas and examples; the vLLM IS test is parametrized over max-only, min-only, and both bounds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28952dd770621adb229e6a502a6e26e17d91e92c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


casinca:
I think the logic should be good but I need some directions for the docs.

Currently the upper bound is called `vllm_importance_sampling_cap` or $C$ in the TRL formula, which made sense when there was only an upper bound.
I don't believe calling the lower bound "lower or min cap" grammatically correct. At the moment I've called the var `vllm_importance_sampling_min` but having "min" and "cap" doesn't sound intuitive, imo, for users that these are both sides of the same thing.

And if I add M to the current TRL formula:

image

So I'd like to harmonize and refactor `vllm_importance_sampling_cap` to `vllm_importance_sampling_max`, add a warning deprecation for backward compatibility and in the formula $M_{\min}$ and $M_{\max}$

For reference, in MiMO they call it eps_low and eps_high (but it might be confusing with asymmetric PPO clips). in IcePoP it's alpha and beta. 

So not sure here what to do before proceeding further with the docs.
what do you guys think @qgallouedec @LeonEricsson ?

LeonEricsson:
Sorry for the delay.

Looks good overall, though I’m not convinced that a lower bound is inappropriate for TIS. Intuitively, the same rationale applies on both sides: we still want to use the gradient signal even when the training policy drifts from inference policy. A lower bound would then be a way to preserve sample efficiency, with the caveat that it can amplify unreliable signal if the mismatch reflects genuinely OOD samples rather than minor drift.

Viewed this way, the motivation for double-sided boundaries applies to both TIS and MIS. TIS is essentially trying to extract information from every sample under the assumption that the inference–training mismatch is reasonably controlled, while MIS takes a more conservative stance by rejecting samples or tokens that fall outside a defined trust region.

I’ll defer the naming decision to @qgallouedec, but I agree with the concern that `cap` and `min` don’t read as a matched pair.

casinca:
Thanks for the feedback. I will make the necessary changes for TIS as well then. With the help of the documentation, ultimately it'll give users more combinations to choose on what they want to do/act accordingly.

LeonEricsson:
> ultimately it'll give users more combinations to choose on what they want to do/act accordingly.

Agreed. We should clarify (in paper index) that the referenced TIS paper only employs an upper bound. 

I'm considering extending our repertoire of importance sampling ratios with a geometric mean variant as well, but may hold of until this PR is merged

casinca:
To update on this PR:
- added lower bound for TIS as well based on the discussion with @LeonEricsson 
- took the initiative to refactor (with deprecation in mind) `vllm_importance_sampling_cap` to `vllm_importance_sampling_max` and better align with the lower min bound `vllm_importance_sampling_min`
- updated docs to reflect the changes
- `vllm_importance_sampling_min` is set to 0.0 by default for backward compatibility

I think it's good for review now, and in any case if @qgallouedec wants to rename args, the changes should be minimal now.

trl/trainer/grpo_config.py
```diff
@@ -275,6 +275,9 @@ class GRPOConfig(TrainingArguments):
         vllm_importance_sampling_cap (`float`, *optional*, defaults to `3.0`):
             Importance sampling cap C used by `vllm_importance_sampling_mode`. For `*_truncate` modes, importance
             ratios are clipped from above at C. For `*_mask` modes, ratios larger than C are set to zero.
```
```suggestion
        vllm_importance_sampling_max (`float`, *optional*, defaults to `3.0`):
            Importance sampling C_max used by `vllm_importance_sampling_mode`. For `*_truncate` modes, importance
            ratios are clipped from above at C_max. For `*_mask` modes, ratios larger than C are set to zero.
```

LeonEricsson (commented):
I can't comment on these lines

https://github.com/huggingface/trl/blob/cc346ed34ffd3a80044cea0ebe4b173966edc05b/trl/trainer/grpo_config.py#L266-L274

but they should be updated with C_min / C_max

trl/trainer/grpo_config.py
casinca: Yep thanks, I had also missed `vllm_importance_sampling_mode` docs accounting for both range now `ρ ← clamp(ρ, C_min, C_max))` and switched `min` to `C_min`

trl/trainer/grpo_config.py
```diff
@@ -878,3 +905,12 @@ def __post_init__(self):
 
         if self.delta is not None and self.use_liger_kernel:
             raise ValueError("Liger kernel does not support two-sided GRPO loss yet.")
+
+        if self.vllm_importance_sampling_cap is not None:
+            warnings.warn(
```
LeonEricsson, line 988: need `import warnings`

casinca: Thanks for catching it. I was wondering how this was missed since I remember there were other deprecated arguments with
the one I added.  
I must have messed up the latest merge conflict
https://github.com/huggingface/trl/pull/4732/commits/b30cf1a963110d9f2d0ffa1f15b86509633a65a4 that was from this PR:
https://github.com/huggingface/trl/pull/4847 which removes `warnings` and the other deprecated

The "will be removed in version 0.28.0" is still good for now

casinca:
~~TODO: TRL was updated to 0.28 https://github.com/huggingface/trl/pull/5043
The warning docstring will need to be updated accordingly in `grpo_config.py` to match the dev build version at the time of merging.~~

```python
        if self.vllm_importance_sampling_cap is not None:
            warnings.warn(
                "The `vllm_importance_sampling_cap` argument is deprecated and will be removed in version 0.28.0. You "
                "should instead use `vllm_importance_sampling_max`.",
                FutureWarning,
                stacklevel=2,
            )
```

~~Edit: bumped to 0.29.0~~
~~Edit: bumped to 0.30.0~~

trl/experimental/gfpo/gfpo_trainer.py
```diff
@@ -227,7 +227,7 @@ def _generate_and_score_completions(self, inputs):
             if self.use_vllm and self.vllm_importance_sampling_correction:
                 importance_sampling_ratio = torch.exp(old_per_token_logps - sampling_per_token_logps)
                 importance_sampling_ratio = torch.clamp(
-                    importance_sampling_ratio, max=self.vllm_importance_sampling_cap
+                    importance_sampling_ratio, max=self.vllm_importance_sampling_max
```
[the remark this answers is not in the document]
casinca: This is intended, only GRPO is updated for IcePoP. Other trainers are just being updated for the refactoring of `self.vllm_importance_sampling_cap` to `self.vllm_importance_sampling_max`

**edit:** added `self.vllm_importance_sampling_min` to these experimental trainers for consistency. The default behavior of these trainers do not change.

docs/source/grpo_trainer.md
```diff
@@ -284,13 +284,13 @@ $$
...
 
-The standard way to correct for this distribution shift is **importance sampling (IS)**. We provide two IS variants: [Truncated Importance Sampling (TIS)](paper_index#truncated-importance-sampling) and [Masked Importance Sampling (MIS)](paper_index#masked-importance-sampling). Both variants can be applied either at the token level or at the sequence level.Let  \\( \rho \\) denote the importance weight, for example  \\( \rho_t \\) per token or  \\( \rho_{\text{seq}} \\) per sequence. Under TIS, ratios larger than `vllm_importance_sampling_cap` are clipped,
+The standard way to correct for this distribution shift is **importance sampling (IS)**. We provide two IS variants: [Truncated Importance Sampling (TIS)](paper_index#truncated-importance-sampling) and [Masked Importance Sampling (MIS)](paper_index#masked-importance-sampling). Both variants can be applied either at the token level or at the sequence level.Let  \\( \rho \\) denote the importance weight, for example  \\( \rho_t \\) per token or  \\( \rho_{\text{seq}} \\) per sequence. Under TIS, ratios outside of the range `[vllm_importance_sampling_min, vllm_importance_sampling_max]` are clipped,
```
albertvillanova: Pre-existing typo:
```suggestion
The standard way to correct for this distribution shift is **importance sampling (IS)**. We provide two IS variants: [Truncated Importance Sampling (TIS)](paper_index#truncated-importance-sampling) and [Masked Importance Sampling (MIS)](paper_index#masked-importance-sampling). Both variants can be applied either at the token level or at the sequence level. Let  \\( \rho \\) denote the importance weight, for example  \\( \rho_t \\) per token or  \\( \rho_{\text{seq}} \\) per sequence. Under TIS, ratios outside of the range `[vllm_importance_sampling_min, vllm_importance_sampling_max]` are clipped,
```

trl/trainer/grpo_config.py
```diff
@@ -803,21 +810,42 @@ class GRPOConfig(_BaseConfig):
...
+        default=3.0,
+        metadata={
+            "help": "Importance sampling upper bound C_max used by `vllm_importance_sampling_mode`. For '*_truncate' "
+            "modes, ratios are clipped from above at C_max. For '*_mask' modes, ratios larger than C_max are set to "
+            "zero."
+        },
+    )
+
+    vllm_importance_sampling_min: float = field(
+        default=0.0,
```
albertvillanova: I think this param should default to `None`:
```suggestion
    vllm_importance_sampling_min: float | None = field(
        default=None,
```

trl/trainer/grpo_config.py
```diff
@@ -292,14 +292,21 @@ class GRPOConfig(_BaseConfig):
...
+                - `"token_mask"`: Token-level masked IS. Per-token ratios outside [C_min, C_max] are set to zero.
+                - `"sequence_truncate"`: Sequence-level truncated IS. A single sequence ratio is clipped to
+                [C_min, C_max] and applied to all tokens in the sequence.
+                - `"sequence_mask"`: Sequence-level masked IS. Sequences with ratios outside [C_min, C_max] are masked
+                out.
+        vllm_importance_sampling_max (`float`, *optional*, defaults to `3.0`):
+            Importance sampling upper bound C_max used by `vllm_importance_sampling_mode`. For `*_truncate` modes,
+            importance ratios are clipped from above at C_max. For `*_mask` modes, ratios larger than C_max are set to
+            zero.
+        vllm_importance_sampling_min (`float`, *optional*, defaults to `0.0`):
```
albertvillanova: And then:
```suggestion
        vllm_importance_sampling_min (`float`, *optional*):
```

tests/test_grpo_trainer.py
```diff
@@ -1369,8 +1369,18 @@ def test_train_vllm_structured_outputs(self):
 
     @require_vllm
     @pytest.mark.skip(reason="We should add a mock for the vLLM server.")
-    def test_train_vllm_importance_sampling_correction(self):
-        """Test that training works with vLLM for generation with structured outputs."""
+    @pytest.mark.parametrize(
+        "vllm_importance_sampling_min, vllm_importance_sampling_max",
+        [
+            (0.0, 3.0),  # only cap/max (current default)
```
albertvillanova: And then:
```suggestion
            (None, 3.0),  # only cap/max (current default)
```

trl/trainer/grpo_config.py
```diff
@@ -803,21 +810,42 @@ class GRPOConfig(_BaseConfig):
...
+        default=None,
         metadata={
-            "help": "Importance sampling cap C used by `vllm_importance_sampling_mode`. For '*_truncate' modes, "
+            "help": "Deprecated, use `vllm_importance_sampling_max` instead. "
+            "Importance sampling cap C used by `vllm_importance_sampling_mode`. For '*_truncate' modes, "
             "ratios are clipped from above at C. For '*_mask' modes, ratios larger than C are set to zero."
         },
     )
+
+    vllm_importance_sampling_max: float = field(
```
albertvillanova: Analogously to `vllm_importance_sampling_min`, we could use `None` instead of ``float('inf')`` when no upper bound is set:
```suggestion
    vllm_importance_sampling_max: float | None = field(
```

trl/trainer/grpo_config.py
```diff
@@ -803,21 +810,42 @@ class GRPOConfig(_BaseConfig):
...
+        },
+    )
+
+    vllm_importance_sampling_min: float = field(
+        default=0.0,
+        metadata={
+            "help": "Importance sampling lower bound C_min used by `vllm_importance_sampling_mode`. For `*_truncate` "
+            "modes, ratios are clipped from below at C_min. For `*_mask` modes, ratios below C_min are set to "
+            "zero. To strictly mask ratios below C_min without upper bound, set `vllm_importance_sampling_max` to "
+            "`float('inf')`"
```
albertvillanova: And then:

```suggestion
            "zero. To strictly mask ratios below C_min without upper bound, set `vllm_importance_sampling_max=None`."
```

trl/trainer/grpo_config.py
```diff
@@ -292,14 +292,21 @@ class GRPOConfig(_BaseConfig):
...
+                - `"sequence_mask"`: Sequence-level masked IS. Sequences with ratios outside [C_min, C_max] are masked
+                out.
+        vllm_importance_sampling_max (`float`, *optional*, defaults to `3.0`):
+            Importance sampling upper bound C_max used by `vllm_importance_sampling_mode`. For `*_truncate` modes,
+            importance ratios are clipped from above at C_max. For `*_mask` modes, ratios larger than C_max are set to
+            zero.
+        vllm_importance_sampling_min (`float`, *optional*, defaults to `0.0`):
+            Importance sampling lower bound C_min used by `vllm_importance_sampling_mode`. For `*_truncate` modes,
+            ratios are clipped from below at C_min. For `*_mask` modes, ratios below C_min are set to zero. To strictly
+            mask ratios below C_min without upper bound, set `vllm_importance_sampling_max` to `float('inf')`.
```
albertvillanova: And then:
```suggestion
            mask ratios below C_min without upper bound, set `vllm_importance_sampling_max=None`.
```

trl/trainer/grpo_trainer.py
```diff
@@ -2079,11 +2080,16 @@ def _generate_and_score_completions(
...
                     vllm_importance_sampling_ratio = torch.clamp(
-                        vllm_importance_sampling_ratio, max=self.vllm_importance_sampling_cap
+                        vllm_importance_sampling_ratio,
+                        min=self.vllm_importance_sampling_min,
+                        max=self.vllm_importance_sampling_max,
                     )
                 elif self.vllm_importance_sampling_mode in ["sequence_mask", "token_mask"]:
+                    invalid_mis_mask = (vllm_importance_sampling_ratio < self.vllm_importance_sampling_min) | (
+                        vllm_importance_sampling_ratio > self.vllm_importance_sampling_max
+                    )
```
albertvillanova: This should be updated if default min value is set to `None`.

tests/test_grpo_trainer.py
```diff
@@ -1369,8 +1369,18 @@ def test_train_vllm_structured_outputs(self):
...
+            (0.5, float("inf")),  # only min
```
albertvillanova: And then:
```suggestion
            (0.5, None),  # only min
```

albertvillanova (changes requested):
Hi @casinca, thanks a lot for your contribution and sorry again for the late reply.

I have merged main and solved the conflicts. Some suggestions below: I think we should support `None` values for both new parameters.

Additionally, not a strong opinion, but what about alternative names for the new parameters? I find the `_min/_max` naming could be considered ambiguous: seen as the min/max values of the IS ratio, instead of the lower/upper clipping thresholds. What about `vllm_importance_sampling_clip_min` / `vllm_importance_sampling_clip_max` instead? Other alternatives: `vllm_importance_sampling_cap` (no deprecation needed) / `vllm_importance_sampling_floor` (not sure whether the latter is widely used though; besides it breaks the naming symmetry of min/max)? Any other alternatives? I insist, no strong opinion, but maybe interesting to know your opinion and others' opinions (CC: @qgallouedec).

casinca:
Thanks for the update @albertvillanova  I don't mind at all using `None` for deactivating

Concerning variable names, I remember we discussed it with Leon and we just ended up with these in the meantime while waiting for your opinion.

Some more big labs used it since, from what I've seen in the literature:

- Ring1T,  Kimi K2.5, GLM 5.0: use alpha and beta
- MiMo V2, Nvidia Nemotron: use eps low and eps high

Personally I prefer your first suggestion with suffixes  _clip_min /_clip_max rather than cap/floor

trl/trainer/grpo_trainer.py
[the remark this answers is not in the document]
albertvillanova: This review goes in the line with my comment above: https://github.com/huggingface/trl/pull/4732#discussion_r3346339064
> This should be updated if default min value is set to `None`.

albertvillanova (commented):
To the naming discussion, I would add:
- `alpha` and `beta` are specific to IcePoP notation; 
- the notation we were using in our docs before was "C"; now in our docs we have "C_min" and "C_max", which I think it makes more sense if we call them "clip_min" and "clip_max"
- the "eps" notation is normally used with `1`: `1 - eps`, `1 + eps`

Additionally, I think in the docs we should mention the old and new notations and attribute them properly:
- TIS: Original TIS literature (Munos et al., etc.) uses a single upper-cap C: `min(ρ, C)`
- This PR retroactively rewrites the TIS math with `clip(ρ, C_min, C_max)`; we should attribute the origin of this formula, rather than silently rewriting the classic formula
- The IcePop (2510.18855) explicitly uses alpha and beta; maybe we should mention this and say we use C_min and C_max

trl/experimental/gfpo/gfpo_trainer.py
```diff
@@ -238,7 +238,9 @@ def _generate_and_score_completions(self, inputs):
             if self.use_vllm and self.vllm_importance_sampling_correction:
                 importance_sampling_ratio = torch.exp(old_per_token_logps - sampling_per_token_logps)
                 importance_sampling_ratio = torch.clamp(
-                    importance_sampling_ratio, max=self.vllm_importance_sampling_cap
+                    importance_sampling_ratio,
+                    min=self.vllm_importance_sampling_min,
+                    max=self.vllm_importance_sampling_max,
```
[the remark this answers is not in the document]
casinca: These 2 experimental trainers do not have MIS implemented at all, I'm not sure that's relevant here or related to my PR. The only change done on these 2 exp trainers, was refactoring to include the min bound for their TIS implementations `self.vllm_importance_sampling_min`

albertvillanova:
Hi, just following up on this contribution.

When you have a chance, I'd be interested to hear your thoughts on the points above (renaming and documentation suggestions) so we can move the discussion forward and decide on the best path. Thanks!

casinca:
> Hi, just following up on this contribution.
> 
> When you have a chance, I'd be interested to hear your thoughts on the points above (renaming and documentation suggestions) so we can move the discussion forward and decide on the best path. Thanks!

Sorry @albertvillanova , I misunderstood and was waiting for confirmation or others' input. 

It's fine by me, just a nit: suffix `_clip_min/max` as a general purpose variable is accurate for TIS but for MIS it'll be a slight misnomer since we are not clipping/clamping but completely ignoring/discarding the samples. Just wanted to mention in case, there's also `_bound_min/max` or `range_min/max` as more "general" terms. Will do the changes.

albertvillanova:
Thanks for the detail, @casinca.

Just for context, before this PR, the parameter was named `_cap` both for TIS and MIS.

casinca:
Yes with a single bound Cap was accurate, by the way, there's a bit of overlap between `grpo_trainer.md` and `paper_index.md` I changed the doc in `grpo_trainer.md` but I can somewhat repeat in the `paper_index.md` if needed.

For the deprecation warning I changed the string to version `1.7.0` but not sure when you'd want that.

docs/source/paper_index.md
```diff
@@ -320,27 +322,30 @@ training_args = GRPOConfig(
     use_vllm=True,
     vllm_importance_sampling_correction=True, # default True
     vllm_importance_sampling_mode="sequence_truncate", # or "token_truncate"
-    vllm_importance_sampling_cap=2.0, # hyper-parameter C
+    vllm_importance_sampling_clip_max=2.0, # hyper-parameter C_max
+    vllm_importance_sampling_clip_min=None, # hyper-parameter C_min
```
albertvillanova: I would not show a parameter set to its default in a code example: it may be seen as if it implies the user must set it:
```suggestion
```

docs/source/paper_index.md
```diff
@@ -364,7 +369,8 @@ training_args = GRPOConfig(
     use_vllm=True,
     vllm_importance_sampling_correction=True, # default True
     vllm_importance_sampling_mode="sequence_mask", # or "token_mask"
-    vllm_importance_sampling_cap=2.0, # hyper-parameter C
+    vllm_importance_sampling_clip_max=2.0, # hyper-parameter C_max
+    vllm_importance_sampling_clip_min=0.0, # hyper-parameter C_min
```
albertvillanova: I think this example should either omit clip_min (default None) or show a meaningful value like 0.5 that actually produces two-sided masking. As written, it misleads the user into thinking 0.0 does something.

docs/source/grpo_trainer.md
```diff
@@ -284,13 +284,16 @@ $$
...
 
 $$
-\rho \leftarrow \min(\rho, C).
+\rho \leftarrow \text{clip}(\rho, C_{\min}, C_{\max}).
 $$
 
-Under MIS, ratios larger than `vllm_importance_sampling_cap` are set to zero, so those samples do not contribute to the gradient. In other words, large ratio samples are downweighted under TIS and discarded under MIS. The configuration flag `vllm_importance_sampling_mode` chooses both the IS variant (masking or truncation) and the granularity (token level or sequence level).
+The original [TIS paper](https://huggingface.co/papers/1606.02647) proposed a single upper-bound clipping mechanism, i.e.,  \\( \min(\rho, C_{\max}) \\). The implementation in TRL generalized this by also introducing a lower bound, yielding the two-sided formulation shown above, inspired by the [IcePop](paper_index#masked-importance-sampling) method.  
+Note that in IcePoP, the bounds are labelled as  \\( \alpha \\) and  \\( \beta \\) while in TRL we use  \\( C_{\min} \\) and  \\( C_{\max} \\).
```
albertvillanova: Nit: IcePop capitalization
```suggestion
Note that in IcePop, the bounds are labelled as  \\( \alpha \\) and  \\( \beta \\) while in TRL we use  \\( C_{\min} \\) and  \\( C_{\max} \\).
```

docs/source/grpo_trainer.md
```diff
@@ -284,13 +284,16 @@ $$
...
+
+Under MIS, ratios outside of this range are set to zero, so those samples do not contribute to the gradient. In other words, outliers samples are downweighted under TIS and discarded under MIS. The configuration flag `vllm_importance_sampling_mode` chooses both the IS variant (masking or truncation) and the granularity (token level or sequence level).
```
albertvillanova: Nit:
```suggestion
Under MIS, ratios outside of this range are set to zero, so those samples do not contribute to the gradient. In other words, outlier samples are downweighted under TIS and discarded under MIS. The configuration flag `vllm_importance_sampling_mode` chooses both the IS variant (masking or truncation) and the granularity (token level or sequence level).
```

trl/trainer/grpo_config.py
```diff
@@ -803,21 +810,41 @@ class GRPOConfig(_BaseConfig):
...
+
+    vllm_importance_sampling_cap: float | None = field(
+        default=None,
         metadata={
-            "help": "Importance sampling cap C used by `vllm_importance_sampling_mode`. For '*_truncate' modes, "
+            "help": "Deprecated, use `vllm_importance_sampling_clip_max` instead. "
+            "Importance sampling cap C used by `vllm_importance_sampling_mode`. For '*_truncate' modes, "
             "ratios are clipped from above at C. For '*_mask' modes, ratios larger than C are set to zero."
         },
     )
```
albertvillanova: This parameter should be moved under the "Deprecated parameters" comment.

trl/trainer/grpo_config.py
```diff
@@ -950,3 +977,12 @@ def __post_init__(self):
                 "GRPO requires at least 2 generations per prompt to calculate the advantages. You provided "
                 f"{self.num_generations}, which is less than the minimum required."
             )
+
+        if self.vllm_importance_sampling_cap is not None:
+            warnings.warn(
+                "The `vllm_importance_sampling_cap` argument is deprecated and will be removed in version 1.7.0. You "
+                "should instead use `vllm_importance_sampling_clip_max`.",
```
albertvillanova:
```suggestion
                "The `vllm_importance_sampling_cap` argument is deprecated and will be removed in v2.0.0.  "
                "Use `vllm_importance_sampling_clip_max` instead.",
```

trl/trainer/grpo_config.py
```diff
@@ -292,14 +292,21 @@ class GRPOConfig(_BaseConfig):
             Specifies how Importance Sampling is performed when `vllm_importance_sampling_correction=True`. Possible
             values are:
 
-                - `"token_truncate"`: Token-level truncated IS (default). Per-token ratios are clipped from above at C.
-                - `"token_mask"`: Token-level masked IS. Per-token ratios above C are set to zero.
-                - `"sequence_truncate"`: Sequence-level truncated IS. A single sequence ratio is clipped from above at
-                  C and applied to all tokens in the sequence.
-                - `"sequence_mask"`: Sequence-level masked IS. Sequences with ratios above C are masked out.
-        vllm_importance_sampling_cap (`float`, *optional*, defaults to `3.0`):
```
albertvillanova, line 300: Again, in the docstring, the deprecated param should be moved to the "> Deprecated paramaters section". See `use_transformers_paged` as example.

albertvillanova (changes requested):
Thanks! Just some minor corrections and we will be ready to merge.

docs/source/paper_index.md
casinca: I agree, since I applied your suggestion earlier (removed when the argument is None/default), this time I'll explicitly pass a value to show users the kw exists.

trl/trainer/grpo_trainer.py
```diff
@@ -2079,11 +2080,27 @@ def _generate_and_score_completions(
...
+                        self.vllm_importance_sampling_clip_max
+                        if self.vllm_importance_sampling_clip_max is not None
+                        else math.inf
+                    )
+
+                    invalid_mis_mask = (vllm_importance_sampling_ratio < min_val) | (
+                        vllm_importance_sampling_ratio > max_val
+                    )
                     vllm_importance_sampling_ratio = vllm_importance_sampling_ratio.masked_fill(
-                        vllm_importance_sampling_ratio > self.vllm_importance_sampling_cap, value=0.0
```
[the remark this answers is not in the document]
casinca: Not wrong, technically if an user wants to disable IS correction he would set `self.vllm_importance_sampling_correction=False` but the edge case where someone set both to `None` instead is possible now that *_min is None by default and not 0.0 anymore.

I can add an extra check like below if needed (waiting to confirm)

```diff
@@ -2079,11 +2079,12 @@ class GRPOTrainer(_BaseTrainer):
                 #   sequence_* modes:  (B, 1)  (per-sequence ratio)
 
                 if self.vllm_importance_sampling_mode in ["sequence_truncate", "token_truncate"]:
+                    if self.vllm_importance_sampling_clip_min is not None or self.vllm_importance_sampling_clip_max is not None:
                    vllm_importance_sampling_ratio = torch.clamp(
                        vllm_importance_sampling_ratio,
                        min=self.vllm_importance_sampling_clip_min,
                        max=self.vllm_importance_sampling_clip_max,
                    )
```

trl/trainer/grpo_config.py
```diff
@@ -950,3 +977,12 @@ def __post_init__(self):
...
             )
+
+        if self.vllm_importance_sampling_cap is not None:
+            warnings.warn(
+                "The `vllm_importance_sampling_cap` argument is deprecated and will be removed in v2.0.0.  "
+                "Use `vllm_importance_sampling_clip_max` instead.",
+                FutureWarning,
+                stacklevel=2,
+            )
+            self.vllm_importance_sampling_clip_max = self.vllm_importance_sampling_cap
```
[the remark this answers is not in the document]
casinca: Good catch. Added an extra check + warning to prevent `vllm_importance_sampling_clip_max` from being overwritten if `vllm_importance_sampling_cap` was also modified

trl/trainer/grpo_trainer.py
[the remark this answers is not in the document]
albertvillanova: I think this could be better implemented as a validation check in `GRPOConfig.__post_init__` and raise a ValueError.

trl/trainer/grpo_trainer.py
```diff
@@ -2088,11 +2089,27 @@ def _generate_and_score_completions(
...
```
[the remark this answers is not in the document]
albertvillanova: I also think this could be better implemented as a validation check in `GRPOConfig.__post_init__` and raise a ValueError.

trl/trainer/grpo_config.py
```diff
@@ -338,6 +345,15 @@ class GRPOConfig(_BaseConfig):
 
             </Deprecated>
 
+        vllm_importance_sampling_cap:
+
+            <Deprecated version="1.6.0">
+
+            Parameter `vllm_importance_sampling_cap` is deprecated and will be removed in version v2.0.0. It will be
+            replaced by `vllm_importance_sampling_clip_max`.
```
albertvillanova: I know this aligns the deprecated param just above, but I think this way is more pertinent for this case, where the param was just renamed:
```suggestion
            Parameter `vllm_importance_sampling_cap` is deprecated and will be removed in v2.0.0. Use
            `vllm_importance_sampling_clip_max` instead.
```

albertvillanova (changes requested):

trl/trainer/grpo_trainer.py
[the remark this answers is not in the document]
casinca: I added 2 checks in the `GRPOConfig.__post_init__` for both edge cases

trl/trainer/grpo_config.py
```diff
@@ -950,3 +983,43 @@ def __post_init__(self):
...
+
+        if (
+            self.vllm_importance_sampling_clip_min is not None
+            and self.vllm_importance_sampling_clip_max is not None
+            and self.vllm_importance_sampling_clip_min > self.vllm_importance_sampling_clip_max
+        ):
+            raise ValueError(
+                f"vllm_importance_sampling_clip_min ({self.vllm_importance_sampling_clip_min}) must be less than or "
+                f"equal to vllm_importance_sampling_clip_max ({self.vllm_importance_sampling_clip_max})."
+            )
```
albertvillanova, line 1004: I think allowing `vllm_importance_sampling_clip_min = vllm_importance_sampling_clip_max` is incorrect, because this allow the same kind of issues as if `min > max` (before this fix). Strict < is the mathematically correct definition of a non-degenerate interval and the only version that fully addresses the issue.

trl/trainer/grpo_config.py
```diff
@@ -950,3 +983,43 @@ def __post_init__(self):
...
+
+        if self.vllm_importance_sampling_cap is not None:
+            warnings.warn(
+                "The `vllm_importance_sampling_cap` argument is deprecated and will be removed in v2.0.0.  "
+                "Use `vllm_importance_sampling_clip_max` instead.",
+                FutureWarning,
+                stacklevel=2,
+            )
+            if self.vllm_importance_sampling_clip_max == 3.0:
+                self.vllm_importance_sampling_clip_max = self.vllm_importance_sampling_cap
```
albertvillanova: Also, I think this is incorrect, because we never know if the user just did not explicitly set a value (and left the old default) or explicitly set it to 3. The intent is to distinguish "user left clip_max at its default" from "user explicitly passed clip_max=3.0", but those two situations are indistinguishable from a dataclass perspective: the field value is 3.0 in both cases.

Maybe a solution would be internally (not shown in the docs) to default clip_max to None and resolve the actual 3.0 default in `__post_init__`, after the deprecation migration has run. This repurposes None to mean "not explicitly set" at config-construction time, then resolves it before the value is ever used.

Something like:
```python
# Deprecation migration — runs before the default is applied, so None = "not explicitly set"
if self.vllm_importance_sampling_cap is not None:
    warnings.warn("... use vllm_importance_sampling_clip_max instead ...", FutureWarning, stacklevel=2)
    if self.vllm_importance_sampling_clip_max is not None:
        warnings.warn("Both ... set, ignoring deprecated cap.", UserWarning, stacklevel=2)
    else:
        self.vllm_importance_sampling_clip_max = self.vllm_importance_sampling_cap

# Apply default if neither path set it
if self.vllm_importance_sampling_clip_max is None:
    self.vllm_importance_sampling_clip_max = 3.0
```

albertvillanova (changes requested):
Thanks again for addressing the previous issues. Below, a couple of bugs I think there are in the code.

trl/trainer/grpo_config.py
albertvillanova: I'm thinking that the user may have passed `_max = None` when they want to use only `_min`... So my argument above is wrong!

albertvillanova: Maybe we could just drop the "both set" disambiguation, and always let _cap win. The cap → clip_max migration is a one-way deprecation path used by users who haven't updated their code yet.

```python
if self.vllm_importance_sampling_cap is not None:
    warnings.warn("... use vllm_importance_sampling_clip_max instead ...", FutureWarning, stacklevel=2)
    self.vllm_importance_sampling_clip_max = self.vllm_importance_sampling_cap
```

casinca: I'll rollback to what it was initially then, this is the snippet you wrote

trl/trainer/grpo_config.py
```diff
@@ -950,3 +983,43 @@ def __post_init__(self):
...
+            self.vllm_importance_sampling_correction
+            and self.vllm_importance_sampling_mode in ("token_truncate", "sequence_truncate")
+            and self.vllm_importance_sampling_clip_min is None
+            and self.vllm_importance_sampling_clip_max is None
+        ):
+            raise ValueError(
+                "vllm_importance_sampling_clip_min and vllm_importance_sampling_clip_max cannot both be None when "
+                "using vllm_importance_sampling_correction=True. "
+                "To disable importance sampling correction, set vllm_importance_sampling_correction=False."
+            )
```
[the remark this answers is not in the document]
casinca: `None` logic would crash TIS because of clamp getting no min and max values, for MIS if a user sets both `None`, it results in no masking just like `vllm_importance_sampling_correction=False` because the range is -inf to +inf. The result is: it applies an unbounded correction to the loss.

Although it is unlikely that a user would want to do that, it's not breaking but I can enforce the `ValueError` for MIS too if you prefer @albertvillanova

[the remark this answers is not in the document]
albertvillanova: I agree with you, @casinca. Maybe better aligning the error message with the underlying requirement, assuming the user wants importance sampling correction?

trl/trainer/grpo_config.py
```diff
@@ -950,3 +983,34 @@ def __post_init__(self):
...
+        if (
+            self.vllm_importance_sampling_correction
+            and self.vllm_importance_sampling_mode in ("token_truncate", "sequence_truncate")
+            and self.vllm_importance_sampling_clip_min is None
+            and self.vllm_importance_sampling_clip_max is None
+        ):
+            raise ValueError(
+                "vllm_importance_sampling_clip_min and vllm_importance_sampling_clip_max cannot both be None when "
+                "using vllm_importance_sampling_correction=True. "
+                "To disable importance sampling correction, set vllm_importance_sampling_correction=False."
```
albertvillanova:
```suggestion
                "At least one of `vllm_importance_sampling_clip_min` or `vllm_importance_sampling_clip_max` "
                "must be set when `vllm_importance_sampling_mode` is a `*_truncate` mode."
```

albertvillanova (changes requested):

trl/trainer/grpo_config.py
casinca: Yes, my bad that's probably what triggered BugBot in the first place, thanks

albertvillanova (approved):
Thank you, awesome work!! 🤗

qgallouedec:
good work guys!

PR #4744, opened 2025-12-23
Clarify Accelerate usage in SFTTrainer documentation

Likhita-17:
This change clarifies how `SFTTrainer` relies on Accelerate for device placement and distributed training.

It explains why `device_map="auto"` should not be used during training, and adds a minimal runnable example launched via `accelerate`.

## Before submitting
- [x] This PR improves the documentation.

## Who can review?
@stevhliu

### DIFF
--- a/docs/source/sft_trainer.md
+++ b/docs/source/sft_trainer.md
@@ -25,6 +25,69 @@ trainer.train()
 
 <iframe src="https://trl-lib-trackio.hf.space/?project=trl-documentation&metrics=train*&runs=sft_qwen3-0.6B_capybara" style="width: 100%; min-width: 300px; max-width: 800px;" height="830" frameBorder="0"></iframe>
 
+## Using SFTTrainer with Accelerate
+
+`SFTTrainer` is built on top of the  Transformers `Trainer` and relies on
+ Accelerate for device placement and distributed training.
+
+When training with `SFTTrainer`, models should be loaded normally using
+`from_pretrained`, and training scripts should be launched with
+`accelerate launch`. Device placement, multi-GPU execution, and
+distributed strategies are handled by Accelerate at runtime.
+
+The `device_map="auto"` option is intended for inference-time model sharding
+and should not be used during training, as it can conflict with
+Accelerate-managed execution and lead to runtime errors.
+
+### Minimal example
+
+```bash
+accelerate launch train_sft.py
+```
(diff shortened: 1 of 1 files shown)


qgallouedec:
Thank you, I find that useful!

The fact is that the philosophy of trainers in TRL is that, we recommend passing the model ID to the trainer (i.e. a `str`), rather than the instantiated model itself. If the user passes the model (i.e. a `nn.Module`), we consider this to be advanced usage, and that the user knows what they are doing.

Therefore, I think it would be better to simply mention this briefly in the docstring for `SFTTrainer.model`, here: https://huggingface.co/docs/trl/en/sft_trainer#trl.SFTTrainer.model

Likhita-17:
> Thank you, I find that useful!
> 
> The fact is that the philosophy of trainers in TRL is that, we recommend passing the model ID to the trainer (i.e. a `str`), rather than the instantiated model itself. If the user passes the model (i.e. a `nn.Module`), we consider this to be advanced usage, and that the user knows what they are doing.
> 
> Therefore, I think it would be better to simply mention this briefly in the docstring for `SFTTrainer.model`, here: https://huggingface.co/docs/trl/en/sft_trainer#trl.SFTTrainer.model

Thanks for the clarification — that makes a lot of sense.

I agree that this fits better as a brief note in the `SFTTrainer.model` docstring ,  especially given the distinction between passing a model ID versus an instantiated module.

I’ll update the documentation accordingly and move the guidance there.
Thanks again for the review!

qgallouedec:
Thanks for this, and sorry it stalled after my December comment.

Closing it though, because the `device_map="auto"` advice doesn't match what TRL does:
When you pass a model ID, TRL sets `device_map="auto"` itself on GPU
https://github.com/huggingface/trl/blob/94d6b7ddacc22523044c732bc2f0e311d9417c05/trl/trainer/utils.py#L1179
and clears it to None under multi-GPU and DeepSpeed
https://github.com/huggingface/trl/blob/94d6b7ddacc22523044c732bc2f0e311d9417c05/trl/trainer/sft_trainer.py#L983-L984
So it's handled for you rather than something to avoid.

The example also loads with `from_pretrained` and passes the instance, where we recommend passing the model ID string and letting the trainer load it, which is the point I raised in December.

If you'd like to document the device placement behaviour as it actually works, that would be genuinely useful and I'd review it.

PR #4752, opened 2025-12-26
Fix GRPO `scale_rewards` type specification to fix `__post_init__` validation

apalmas-saifh:
🎄✨🎅
This PR fixes the type annotation for `scale_rewards` parameter in GRPO config 
https://github.com/huggingface/trl/blob/c04fdc044dde5e73200f086fbe8737a842754793/trl/trainer/grpo_config.py#L648-L659
to support both `str` and `bool` types as per documentation.

This fixes the `__post_init__` validation at https://github.com/huggingface/trl/blob/22b386a9570f93697c21df9fdf27fa70bcb260d1/trl/trainer/grpo_config.py#L814
which was silently failing because `bool` values `True` and `False` where being casted into `str`, triggering the fallback case of this `get` method, which in turns was then causing the Exception being raised at
https://github.com/huggingface/trl/blob/c04fdc044dde5e73200f086fbe8737a842754793/trl/trainer/grpo_trainer.py#L2006-L2008
in
https://github.com/huggingface/trl/blob/c04fdc044dde5e73200f086fbe8737a842754793/trl/trainer/grpo_trainer.py#L1992-L2008

A more robust validation for the `scale_rewards` field is also proposed in a separate commit:
```python
    if isinstance(self.scale_rewards, bool):
        self.scale_rewards = "group" if self.scale_rewards else "none"
    elif isinstance(self.scale_rewards, str):
        value = self.scale_rewards.lower()
        if value in {"group", "batch", "none"}:
            self.scale_rewards = value
        else:
            raise ValueError(
                f"Invalid value for scale_rewards: {self.scale_rewards!r}. "
                "Expected one of: True, False, 'group', 'batch', 'none'."
            )
    else:
        raise TypeError(
            f"scale_rewards must be a bool or str, got {type(self.scale_rewards).__name__}"
        )
```
to replace the current one 
https://github.com/huggingface/trl/blob/22b386a9570f93697c21df9fdf27fa70bcb260d1/trl/trainer/grpo_config.py#L814
which could fail in different ways ("gruop" would pass the validation and silently fall inside the computation later on)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec (as previously?)

### DIFF
--- a/trl/experimental/kto/kto_config.py
+++ b/trl/experimental/kto/kto_config.py
@@ -65,9 +65,6 @@ class KTOConfig(TrainingArguments):
         model_init_kwargs (`dict[str, Any]`, *optional*):
             Keyword arguments to pass to `AutoModelForCausalLM.from_pretrained` when instantiating the model from a
             string.
-        ref_model_init_kwargs (`dict[str, Any]`, *optional*):
-            Keyword arguments to pass to `AutoModelForCausalLM.from_pretrained` when instantiating the reference model
-            from a string.
         dataset_num_proc: (`int`, *optional*):
             Number of processes to use for processing the dataset.
         disable_dropout (`bool`, *optional*, defaults to `True`):
@@ -88,7 +85,7 @@ class KTOConfig(TrainingArguments):
             `True`.
     """
 
-    _VALID_DICT_FIELDS = TrainingArguments._VALID_DICT_FIELDS + ["model_init_kwargs", "ref_model_init_kwargs"]
+    _VALID_DICT_FIELDS = TrainingArguments._VALID_DICT_FIELDS + ["model_init_kwargs"]
 
     # Parameters whose default values are overridden from TrainingArguments
     learning_rate: float = field(
@@ -192,13 +189,6 @@ class KTOConfig(TrainingArguments):
         metadata={
             "help": "Keyword arguments to pass to `AutoModelForCausalLM.from_pretrained` when instantiating the model "
             "from a string."
-        },
-    )
-    ref_model_init_kwargs: dict[str, Any] | None = field(
-        default=None,
-        metadata={
-            "help": "Keyword arguments to pass to `AutoModelForCausalLM.from_pretrained` when instantiating the "
-            "reference model from a string."
         },
     )
     dataset_num_proc: int | None = field(
--- a/trl/experimental/kto/kto_trainer.py
+++ b/trl/experimental/kto/kto_trainer.py
@@ -352,17 +352,11 @@ class KTOTrainer(BaseTrainer):
 
         # Reference model initialization
         if isinstance(ref_model, str):
-            ref_model_init_kwargs = args.ref_model_init_kwargs or {}
+            ref_model_init_kwargs = args.model_init_kwargs or {}
             # Distributed training requires device_map=None
             if args.distributed_state.distributed_type in ["MULTI_GPU", "DEEPSPEED"]:
                 ref_model_init_kwargs["device_map"] = None
             ref_model = create_model_from_path(ref_model, **ref_model_init_kwargs)
-        else:
-            if ref_model is not None and args.ref_model_init_kwargs is not None:
-                logger.warning(
-                    "You passed `ref_model_init_kwargs` to the KTOConfig, but your ref_model is already instantiated. "
-                    "The `ref_model_init_kwargs` will be ignored."
-                )
 
         # Initialize this variable to False. This helps tracking the case when `peft_module_casting_to_bf16`
         # has been called in order to properly call autocast if needed.


apalmas-saifh:
@qgallouedec pinging here just to know if this can be useful, I will have a more important PR for robustness in distributed training coming in, but wanted to wait for this first.

qgallouedec:
Although not documented, there is a good reason for having `scale_rewards: str` instead of `scale_rewards: str | bool`. The parser only supports 1 type of value. In other words, `trl grpo ...` won't work if we mix the types in the configuration (see the CI error).

Regarding 

> which was silently failing because bool values True and False where being casted into str, triggering the fallback case of this get method, which in turns was then causing the Exception being raised at

I don't think it's "silently failing", because if you use an invalid value, it does fail (not during the init though):

```
$ trl grpo   --model_name_or_path Qwen/Qwen2.5-0.5B   --dataset_name HuggingFaceH4/Polaris-Dataset-53K   --reward_funcs accuracy_reward --scale_rewards no_valid_value
The following values were not passed to `accelerate launch` and had defaults used instead:
        `--num_processes` was set to a value of `1`
        `--num_machines` was set to a value of `1`
        `--mixed_precision` was set to a value of `'no'`
        `--dynamo_backend` was set to a value of `'no'`
To avoid this warning pass in values for each of the problematic parameters or run `accelerate config`.
Loading weights: 100%|████████| 290/290 [00:00<00:00, 855.84it/s, Materializing param=model.norm.weight]
The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config. The model config and generation config were aligned accordingly, being updated with the tokenizer's values. Updated tokens: {'bos_token_id': None, 'pad_token_id': 151643}.
  0%|                                                                        | 0/159873 [00:00<?, ?it/s]Traceback (most recent call last):
  File "/fsx/qgallouedec/trl/trl/scripts/grpo.py", line 193, in <module>
    main(script_args, training_args, model_args, dataset_args)
  File "/fsx/qgallouedec/trl/trl/scripts/grpo.py", line 162, in main
    trainer.train()
  File "/fsx/qgallouedec/transformers/src/transformers/trainer.py", line 2174, in train
    return inner_training_loop(
           ^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/transformers/src/transformers/trainer.py", line 2541, in _inner_training_loop
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 1168, in training_step
    output = super().training_step(model, inputs, num_items_in_batch)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/transformers/src/transformers/trainer.py", line 3807, in training_step
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/trl/trl/extras/profiling.py", line 199, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 1197, in _prepare_inputs
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 2041, in _generate_and_score_completions
    raise ValueError(
ValueError: Invalid value for scale_rewards: no_valid_value. Must be one of 'batch', 'group', or 'none'.
```

apalmas-saifh:
True, it was failing, but not at initialization through the post_init input validation, which can be inconvenient. 

If I understood well and only `str` values are supported, then I would suggest to, al least, update the documentation/docstring 
https://github.com/huggingface/trl/blob/c04fdc044dde5e73200f086fbe8737a842754793/trl/trainer/grpo_config.py#L195-L204
which states that both types (`str` and `bool`) are supported, leading to the inconsistent behaviour described above (`bool` casted to `str` and then failing) 

as well as reworking the logic at:
https://github.com/huggingface/trl/blob/22b386a9570f93697c21df9fdf27fa70bcb260d1/trl/trainer/grpo_config.py#L814

as no `bool` values will ever arrive there (casting to `str` mentioned above), so no `True` or `False` boolean key will ever be available there.

I can take care of it @qgallouedec if you tell me what you want to support (i.e. only `str` input and alike)

qgallouedec:
> as no bool values will ever arrive there

Actually you can use

```python
GRPOConfig(scale_rewards=True)
```

apalmas-saifh:
I double checked and Hydra zen on our end, combined with the `str`-only annotation you have for that config parameter, was casting to string. 

If you have the one-type-only limitation as you shared above, I guess there is no way to make this more robust for users integrating `trl` under the hood for some use cases as it is for us. 

Thanks anyways, clear how to make sure we are using the lib as expected.

PR #4754, opened 2025-12-26
vllm parameter passthrough for stop sequences

kdubovikov:
# What does this PR do?

I have added a passthrough for vllm stop sequence parameters. This can be very useful when implementing custom rollout functions. For example, for multi-step rollouts with intermediate tool calls.


qgallouedec:
thanks, but I think it's already supported via the argument `generation_kwargs`:

```python
self.client.generate(
    prompts,
    max_tokens=4,
    temperature=0.0,
    generation_kwargs=dict(
        stop=["\n"],
        stop_token_ids=[stop_token_id],
        include_stop_str_in_output=False,
    ),
)
```

unless it's really needed I'd try to avoid adding new parameters to the client/server

PR #4755, opened 2025-12-27
XPO/NashMD perf: avoid `output_hidden_states` when only `last_hidden_state` is used

ciaoyizhen:
# What does this PR do?

Avoid unnecessary `output_hidden_states=True` when only the final layer hidden state is used
to compute reward logits.

Currently, the model enables `output_hidden_states=True` but only consumes
`hidden_states[-1]`. This forces the backbone to store all intermediate layer
activations, increasing memory usage and compute overhead.

This PR switches to `output.last_hidden_state`, which is functionally equivalent
to `hidden_states[-1]`, while keeping the behavior unchanged.

Fixes # (issue)

## Before submitting
- [x] Did you read the contributor guideline,
      Pull Request section?
      to it if that's the case.

## Who can review?

Anyone in the community is welcome to review this PR.

---

> [!NOTE]
> **Low Risk**
> Single-path optimization with an equivalent Transformers API; no auth, data, or API contract changes.
> 
> **Overview**
> **Reward scoring in `get_reward`** no longer requests all layer hidden states from the LM backbone. The forward pass sets `output_hidden_states=False` and feeds `output.last_hidden_state` into `model.score` instead of `hidden_states[-1]`.
> 
> Behavior for reward logits and per-sequence rewards should be unchanged; the path avoids retaining intermediate activations on every reward forward pass, which cuts memory and overhead during RL-style training that calls this helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4142f2c695275397062d42c2f24894e3fb6e753. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ciaoyizhen:
```python
from transformers import AutoModel, AutoTokenizer, AutoModelForSequenceClassification

reward_model_ckpt = "/data1/caoyizhen/caoyizhen/open-source-models/internlm2-1_8b-reward"

model = AutoModel.from_pretrained(reward_model_ckpt, trust_remote_code=True)
tokenizer = AutoTokenizer.from_pretrained(reward_model_ckpt, trust_remote_code=True)
print(model.base_model_prefix)

a = "你好"
input_dict = tokenizer(a, return_tensors="pt")
output = model.model(
    **input_dict,
    return_dict=True,
    output_hidden_states=True,
    use_cache=False,  # otherwise mistral-based RM would error out
)
print(output.hidden_states[-1])


output = model.model(
    **input_dict,
    return_dict=True,
    output_hidden_states=False,
    use_cache=True,  # otherwise mistral-based RM would error out
)
print(output.last_hidden_state)
```


image
It is same.

ciaoyizhen:
image
test_code Error.  
Because 
image
Maybe this is reason. Solve this problem later

verma8076:
Confirmed correct against current main (\`trl/experimental/utils.py:773-783\`).

\`lm_backbone = getattr(model, model.base_model_prefix)\` resolves to the base transformer backbone (not the classification head), so its forward always populates \`last_hidden_state\` regardless of \`output_hidden_states\`. \`output_hidden_states=True\`'s only effect here is additionally materializing the full per-layer tuple in \`hidden_states\`, of which \`hidden_states[-1]\` is by construction the same tensor as \`last_hidden_state\` (every standard HF base-model implementation appends the final layer's output to that tuple and assigns the same value to \`last_hidden_state\`). So this is a real, safe change: same tensor used for the score head, no behavior change, drops an O(num_layers) allocation that was never read.

Been open since 2025-12-27 with zero human review. Small, correct, worth a look.

qgallouedec (approved):
sorry for the delay on this. good catch.

PR #4762, opened 2025-12-29
Sudoku GRPO example script using TextArena

sergiopaniego:
# What does this PR do?

Fixes # (issue)
I also plan on adding the Colab notebook to this PR

The 1.7B model learns to play sudoku super quickly on the easy mode (with strong hints):

Screenshot 2025-12-29 at 16 45 25

It struggles with the medium difficulty and is not able to learn the hard one. Increasing the model size and number of parameters should lead to improvements.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec @albertvillanova @burtenshaw


examples/notebooks/openenv_sudoku_grpo.ipynb
```diff
@@ -0,0 +1,6778 @@
...
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "mHmE7GhRKyJj"
+      },
+      "outputs": [],
+      "source": [
+        "#!pip install -Uq trl[vllm] git+https://github.com/meta-pytorch/OpenEnv.git openenv_core  trackio #liger-kernel\n",
+        "!pip install -Uq trl[vllm] trackio git+https://github.com/meta-pytorch/OpenEnv.git@bf5e968286e0d49cdc03fd904d48faff4b15a437 openenv_core==0.1.1 liger-kernel"
```
albertvillanova: Which line should we keep?

examples/notebooks/openenv_sudoku_grpo.ipynb
```diff
@@ -0,0 +1,6778 @@
...
+        "    gradient_checkpointing=True,                               # Save memory by recomputing activations during backprop\n",
+        "    gradient_checkpointing_kwargs={\"use_reentrant\": False},    # Prevent warnings with certain models\n",
+        "    # chat_template_kwargs={\"enable_thinking\": False},        # Optional template args for model reasoning\n",
+        "\n",
+        "    temperature=0.8,\n",
+        "    top_k=10,\n",
+        "    top_p=None,\n",
+        "\n",
+        "    model_init_kwargs={\n",
+        "      \"use_cache\":\"False\",\n",
```
albertvillanova: Should the `use_cache` value be a string or a boolean instead?
```suggestion
        "      \"use_cache\": False,\n",
```

examples/notebooks/openenv_sudoku_grpo.ipynb
```diff
@@ -0,0 +1,6778 @@
...
+        "    trackio_space_id=output_dir,                               # Trackio space ID to log metrics\n",
+        "\n",
+        "    use_liger_kernel=False,                                    # Enable Liger kernel optimizations for faster training\n",
+        "    gradient_checkpointing=True,                               # Save memory by recomputing activations during backprop\n",
+        "    gradient_checkpointing_kwargs={\"use_reentrant\": False},    # Prevent warnings with certain models\n",
+        "    # chat_template_kwargs={\"enable_thinking\": False},        # Optional template args for model reasoning\n",
+        "\n",
+        "    temperature=0.8,\n",
+        "    top_k=10,\n",
+        "    top_p=None,\n",
```
albertvillanova: I think setting `top_p=None` could be problematic. In most cases, if you want “no nucleus sampling constraint”, the safe default is top_p=1.0 (or better: **simply omit it**).

```suggestion
```

examples/notebooks/openenv_sudoku_grpo.ipynb
```diff
@@ -0,0 +1,6778 @@
...
+        "        row, col, num = match.groups()\n",
+        "        return f\"[{row} {col} {num}]\"\n",
+        "\n",
+        "    # Try without spaces\n",
+        "    match = re.search(r\"\\[(\\d)(\\d)(\\d)\\]\", text)\n",
+        "    if match:\n",
+        "        row, col, num = match.groups()\n",
+        "        return f\"[{row} {col} {num}]\"\n",
+        "\n",
+        "    return \"\"\n",
```
albertvillanova: The `extract_sudoku_move` function can return an empty string if it can't parse its input. And you pass its output directly to the env:
```python
move = extract_sudoku_move(generated_text)
result = env.step(TextArenaAction(message=move))
```
Passing an empty string to an env step, might not be handled properly. Have you considered this edge case?

examples/notebooks/openenv_sudoku_grpo.ipynb
```diff
@@ -0,0 +1,6778 @@
...
+        "            successful_moves=successful_moves,\n",
+        "            failed_moves=failed_moves,\n",
+        "            difficulty=difficulty,\n",
+        "        )\n",
+        "        messages = [\n",
+        "            {\"role\": \"system\", \"content\": system_prompt},\n",
+        "            {\"role\": \"user\", \"content\": user_prompt},\n",
+        "        ]\n",
+        "        prompt_text = tokenizer.apply_chat_template(\n",
+        "            messages, add_generation_prompt=True, tokenize=False, enable_thinking=False\n",
```
albertvillanova: In relation with the `enable_thinking` param to `apply_chat_template`, maybe this can have some compatibility issues depending on transformers (or tokenizer) version. Have you checked this?

examples/notebooks/openenv_sudoku_grpo.ipynb
albertvillanova: Maybe we could explain why the pin and when it can be removed.

sergiopaniego: related: https://github.com/meta-pytorch/OpenEnv/issues/273

examples/notebooks/openenv_sudoku_grpo.ipynb
sergiopaniego: This is automatically handled by the environment and is actually expected. If the environment doesn't receive a properly configured movement, it returns a "wrong movement message" that adds to the rewards.

examples/notebooks/openenv_sudoku_grpo.ipynb
sergiopaniego: I've added a comment. I think it's ok to have that for this notebook since it's an example rather than a super generic solution.

examples/scripts/openenv/sudoku.py
```diff
@@ -0,0 +1,783 @@
...
+        report_to="trackio",
+        # chat_template_kwargs={"enable_thinking": False},
+    )
+
+    grpo_config.run_name = args.run_name or f"run-{timestamp}"
+    grpo_config.project = args.project or f"group-{sanitize_name(args.model_id)}"
+    grpo_config.trackio_space_id = args.trackio_space_id
+    grpo_config.gradient_checkpointing = args.gradient_checkpointing
+
+    def rollout_func(prompts: list[str], trainer: GRPOTrainer) -> dict[str, list]:
```
qgallouedec, line 719: For the next openenv examples, it would be nice to leverage `tools` instead of `rollout_func`, which we may want to remove in a near future

examples/notebooks/openenv_sudoku_grpo.ipynb
```diff
@@ -0,0 +1,6777 @@
...
+              "<IPython.core.display.HTML object>"
+            ]
+          },
+          "metadata": {},
+          "output_type": "display_data"
+        },
+        {
+          "name": "stdout",
+          "output_type": "stream",
+          "text": [
```
qgallouedec: is it possible to make this a bit more readable? this cell is 4k lines long

sergiopaniego: I've truncated it!

PR #4763, opened 2025-12-29
ORPO: Avoid catastrophic cancellation in loss function

hartmans:
I was trying to use ORPO to train models where the differences between the chosen and rejected samples was very small.  The typical use case was trying to correct the behavior of a model instructed to proofread or edit long passages of text. Examples of preference improvements I was trying to correct included avoiding cases where the model skipped forward in the text.

My training runs were blowing up; typically I would get one loss of NAN, then everything would go to +INF or -INF, and of course at that point it was all useless.

I dug into the code and discovered that as the logits became large,  the loss function would blow up.

This fix makes my training much more stable.

I made significant use of ChatGPT in my investigation, but reached a point where I independently understood and verified the following:
* I understand how as the logits increase, we can run into catastrophic cancellation with the original code.
* I am confident that the formulation of the computation I introduced is correct given infinite precision
* I understand how the computation I introduced fixes the specific stability problem I was running into

I have tested the change, but I don't have a good feel for how to reduce my datasets into a form that I would be willing to share and that could be run reasonably in your tests.

# What does this PR do?

Use a numerically more stable form of log(1-exp(x)) to avoid catastrophic cancellation when the probability of a model selecting the chosen or rejected sample is very high.  Also, compute the logs ratio in float32 to increase the stability.

## Before submitting
- [x ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
      to it if that's the case.
- [N/A ] Did you make sure to update the documentation with your changes?

## Who can review?

### DIFF
--- a/docs/source/vllm_integration.md
+++ b/docs/source/vllm_integration.md
@@ -3,7 +3,7 @@
 This document will guide you through the process of using vLLM with TRL for faster generation in online methods like GRPO and Online DPO. We first summarize a tl;dr on how to use vLLM with TRL, and then we will go into the details of how it works under the hood.
 
 > [!WARNING]
-> TRL currently only supports vLLM versions `0.10.2`, `0.11.0`, `0.11.1`, and `0.11.2`. Please ensure you have one of these versions installed to avoid compatibility issues.
+> TRL currently only supports vLLM versions `0.10.2`, `0.11.0`, `0.11.1`, `0.11.2` and `0.12.0`. Please ensure you have one of these versions installed to avoid compatibility issues.
 
 > [!TIP]
 > The following trainers currently support generation with vLLM:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ test = [
     "pytest"
 ]
 vllm = [
-    "vllm>=0.10.2,<0.12.0",
+    "vllm>=0.10.2,<0.13.0",
     "fastapi",
     "pydantic",
     "requests",
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -918,8 +918,8 @@ class TestGRPOTrainer(TrlTestCase):
 
     @require_vllm
     @pytest.mark.skip(reason="We should add a mock for the vLLM server.")
-    def test_training_vllm_guided_decoding(self):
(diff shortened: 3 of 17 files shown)


qgallouedec:
That's quite smart! I did a quick benchmark, and it looks good to me:

```python
import torch

def before(x):
    return torch.log1p(-torch.exp(x))

def after(x):
    # branch at -ln 2 ~ -0.693 to avoid cancellation
    t = -0.6931471805599453
    return torch.where(x < t, torch.log1p(-torch.exp(x)), torch.log(-torch.expm1(x)))

def log_expm1(x):
    return torch.log(-torch.expm1(x))

# plot to verify
if __name__ == "__main__":
    import numpy as np

    # Log-spaced domains excluding zero
    x = -torch.logspace(-7.5, 1.5, 1000)
    y1 = before(x)
    y2 = after(x)
    y3 = log_expm1(x)
    print(y1)

    plt.plot(x.numpy(), y1.numpy(), label="before")
    plt.plot(x.numpy(), y2.numpy(), label="after", linestyle="dashed")
    plt.plot(x.numpy(), y3.numpy(), label="log_expm1", linestyle="dotted")
    plt.xscale("symlog", linthresh=1e-8)
    # plt.yscale("symlog", linthresh=1e-20)
    plt.legend()
    plt.grid()
    plt.title("Numerical stability improvement for log(1 - exp(x))")
    plt.xlabel("x")
    plt.ylabel("log(1 - exp(x))")
    plt.savefig("bench.png")
```

`torch.log(-torch.expm1(x))` not stable below -10

pos

`torch.log1p(-torch.exp(x))` not stable above -1e-6

pos2

trl/experimental/orpo/orpo_trainer.py
```diff
@@ -73,6 +73,12 @@
 logger = logging.get_logger(__name__)
 
 
+def log1mexp(x: torch.FloatTensor) -> torch.FloatTensor:
```
qgallouedec, line 76:
```suggestion
def log1mexp(x: torch.FloatTensor) -> torch.FloatTensor:
    """Numerically stable computation of log(1-exp(x))."""
```

PR #4768, opened 2026-01-02
feat(sft): add generation-based evaluation support to SFTTrainer

### LINKED ISSUE #862: Compute metrics for generation tasks in SFTTrainer
Hi, I want to include a custom generation based compute_metrics e.g., BLEU, to the `SFTTrainer`. However, I have difficulties because:
1. The input, `eval_preds`, into compute_metrics contains a `.predictions` attribute but its value are logits rather than generations. There don't seem to be an attribute for generations.
2. In `Seq2seqTrainingArguments`, there is a `predict_with_generate` argument for the purpose of generation based metrics. But this does not work on the `SFTTrainer`. So I am not able to compute the metrics that I want.

Can anyone advise how I can get the generations into the compute_metrics function, please? The `trl` version I am using is `trl==0.5.0`

Here are my TrainingArgs and SFTTrainer.
```
training_args = Seq2SeqTrainingArguments(
    output_dir=output_dir,
    eval_steps=eval_steps,
    evaluation_strategy=evaluation_strategy,
    gradient_accumulation_steps=gradient_accumulation_steps,
    learning_rate=lr,
    max_steps=max_steps,
    num_train_epochs=num_train_epochs,
    per_device_train_batch_size=batch_size,
    per_device_eval_batch_size=batch_size,
    save_steps=save_steps,
    save_strategy=save_strategy,
    warmup_steps=num_warmup_steps,
    weight_decay=weight_decay,
    save_total_limit=save_total_limit,
    load_best_model_at_end=True,
    metric_for_best_model=metric_for_best_model,
    greater_is_better=metric_greater_is_better,
    generation_config=generation_config,
    predict_with_generate=True,
    generation_max_length=10,
)

trainer = SFTTrainer(
    model,
    args=training_args,
    train_dataset=dataset_train,
    eval_dataset=dataset_eval,
    data_collator=collator,
    compute_metrics=compute_metrics,
)
```

Attributes of eval_preds are:
```
>> dir(eval_preds)
['__class__',
 '__delattr__',
 '__dict__',
 '__dir__',
 '__doc__',
 '__eq__',
 '__format__',
 '__ge__',
 '__getattribute__',
 '__getitem__',
 '__gt__',
 '__hash__',
 '__init__',
 '__init_subclass__',
 '__iter__',
 '__le__',
 '__lt__',
 '__module__',
 '__ne__',
 '__new__',
 '__reduce__',
 '__reduce_ex__',
 '__repr__',
 '__setattr__',
 '__sizeof__',
 '__str__',
 '__subclasshook__',
 '__weakref__',
 'inputs',
 'label_ids',
 'predictions']
```

Thank you in advance!

CodersAcademy006:
This PR adds first-class support for generation-based evaluation in `SFTTrainer` via an override of `prediction_step`, enabling proper autoregressive decoding during evaluation when `predict_with_generate=True`.

This brings `SFTTrainer` behavior in line with `Seq2SeqTrainer` while preserving SFT semantics and avoiding unnecessary logits materialization.

Fixes #862 and mitigates eval-time VRAM spikes described in #4679

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -1155,7 +1155,8 @@ class SFTTrainer(BaseTrainer):
 
         # Request token accuracy from Liger kernel and set token scaling if using DFT loss
         if self.args.use_liger_kernel:
-            # Avoid materializing full logits during eval unless explicitly needed
+            # Avoid materializing full logits during eval unless explicitly needed.
+            # This prevents large VRAM spikes when only loss or metrics are required.
             inputs["skip_logits"] = (
                 self.model.training
                 or self.args.prediction_loss_only
@@ -1216,9 +1217,7 @@ class SFTTrainer(BaseTrainer):
             else:
                 warnings.warn(
                     "liger-kernel did not return token_accuracy when requested. "
-                    "The mean_token_accuracy metric will not be logged. "
-                    "This may indicate an outdated liger-kernel version. "
-                    "Consider upgrading.",
+                    "This may indicate an outdated liger-kernel version.",
                     stacklevel=2,
                 )
         else:


qgallouedec:
Thanks for the PR. Can the same result be obtained by leveraging a callback?

CodersAcademy006:
I don’t think a callback can provide the same result in this case.

The issue here isn’t augmenting evaluation, it’s correcting the semantics of `predict_with_generate` in `SFTTrainer`. Today, that flag implies generation-based evaluation, but `prediction_step` still performs a forward pass and materializes logits. That leads to both incorrect evaluation behavior and unnecessary VRAM usage.

Callbacks execute after `prediction_step` and cannot change whether evaluation uses `model.forward` or `model.generate`, nor can they prevent logits allocation early enough to avoid the memory spike. In other words, they can observe or post-process predictions, but they cannot redefine what a prediction is.

This is why Seq2SeqTrainer implements generation-based evaluation by overriding `prediction_step`. This PR brings `SFTTrainer` into alignment with that established contract while preserving SFT-specific behavior.

If there’s interest, we could factor shared generation-eval logic between trainers, but semantically this needs to live in `prediction_step` rather than a callback.

Please correct me otherwise!

qgallouedec:
> I don’t think a callback can provide the same result in this case.
> 
> The issue here isn’t augmenting evaluation, it’s correcting the semantics of `predict_with_generate` in `SFTTrainer`. 

but `predict_with_generate` isn't an arg of `SFTTrainer`:

```python
>>> from trl import SFTConfig
>>> args = SFTConfig(predict_with_generate=True)
TypeError: SFTConfig.__init__() got an unexpected keyword argument 'predict_with_generate'
```

it's only an arg in `transformers.Seq2SeqTrainer` (`SFTTrainer` doesn't inherit from this one).

> Today, that flag implies generation-based evaluation, but `prediction_step` still performs a forward pass and materializes logits. That leads to both incorrect evaluation behavior and unnecessary VRAM usage.

For the record, it's also the case in `transformers.Seq2SeqTrainer`, where we compute the logits for the generated sequence, see

https://github.com/huggingface/transformers/blob/314f10929a2215b74c2ad6ecf7b2f380c9b7468a/src/transformers/trainer_seq2seq.py#L350

> Callbacks execute after `prediction_step` and cannot change whether evaluation uses `model.forward` or `model.generate`, nor can they prevent logits allocation early enough to avoid the memory spike.

I understand the motivation, in this case, I'd recommend the following, so that you've full control on the generation/evaluation:

```python
from transformers import TrainerCallback

class GenerateEveryNStepsCallback(TrainerCallback):
    def __init__(self, generate_every=100):
        self.generate_every = generate_every

    def on_step_end(self, args, state, control, **kwargs):
        if state.is_world_process_zero and state.global_step % self.generate_every == 0:
            ... # your generation and evaluation code here
```


> but semantically this needs to live in `prediction_step` rather than a callback.

From my perspective, this feature isn’t in very high demand right now. So rather than investing in full support immediately, I’d prefer to guide users toward a callback-based approach. If it becomes more widely requested in the future, we can consider adding a dedicated argument and providing full support for generation-based evaluations.

CodersAcademy006:
Thanks for the clarification, that’s a fair point. You’re right that `predict_with_generate` isn’t currently part of `SFTTrainer`, and my wording earlier made it sound like I was fixing an existing contract rather than proposing new behavior.

My main goal here was to address the eval-time VRAM spike and the `token_accuracy` crash with the Liger kernel, which come from materializing logits during evaluation. I agree that callbacks are the preferred approach for generation right now if we don’t want to commit to generation-based eval as a first-class SFTTrainer feature yet.

I’m happy to drop the generation-related changes and keep this PR focused on the Liger fixes + regression test, or close this and revisit generation support later if there’s more demand. Let me know what you’d prefer.

Saad-Mallebhari:
@qgallouedec @yurkoff-mv @CodersAcademy006 I have made some changes regarding the modification needed please review.

qgallouedec:
> I’m happy to drop the generation-related changes and keep this PR focused on the Liger fixes + regression test, or close this and revisit generation support later if there’s more demand. Let me know what you’d prefer.

Sorry for the late answer, yes I think it's better to keep this PR focused on bug fix for now

alvina-yang:
> > I’m happy to drop the generation-related changes and keep this PR focused on the Liger fixes + regression test, or close this and revisit generation support later if there’s more demand. Let me know what you’d prefer.
> 
> Sorry for the late answer, yes I think it's better to keep this PR focused on bug fix for now

Agree with this. We are getting blocked by the liger kernel bug.

LoganVegnaSHOP:
Hey, Any update on this? I am just going to make a PR with the bug fix if I dont hear back since this is severely affecting us @CodersAcademy006

PR #4769, opened 2026-01-02
Add revision override mechanism for testing tiny models

albertvillanova:
Add revision override mechanism for testing tiny models.

This PR introduces a simple mechanism to test tiny model updates from HF Hub PRs before merging them to main. This enables a safer update workflow where we can validate changes in CI before they affect the entire test suite. This approach provides a systematic way to test tiny model updates while maintaining the simplicity and clarity that aligns with our project's principles.

I've been using this approach locally for testing tiny model updates, and I believe it would benefit the project as a more methodical way to handle model updates in CI.

### Motivation

Currently, tiny models used in CI always load from the `main` branch (implicit default). This makes it challenging to test model updates safely:
  - Changes to tiny models immediately affect all CI runs
  - No way to validate updates before merging
  - Risk of breaking CI if a model update has issues

### Solution

This PR adds a `MODEL_REVISIONS` dict in `tests/conftest.py` that maps model IDs to specific revisions. A pytest fixture automatically injects the revision parameter when loading models.

**Key features:**
  - Zero changes to existing tests: completely transparent
  - Simple hardcoded dict: easy to review in PRs
  - Pythonic implementation using pytest's monkeypatch
  - Git-tracked configuration: PR diffs are self-documenting

### Implementation Details
  - Uses autouse=True pytest fixture with function scope
  - Monkeypatches all AutoModel*.from_pretrained() methods
  - Direct dict lookup by full model ID (no string manipulation)
  - Only injects revision if not already explicitly set
  - ~40 lines of code including documentation

### Testing
I have validated locally with multiple tiny models and test suites. The mechanism correctly:
  - Injects revisions for models in the override dict
  - Leaves other models unaffected
  - Preserves explicitly set revision parameters
  - Works with both direct loading and trainer-based loading

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -46,38 +46,31 @@ def apply_model_revisions(monkeypatch):
     if not MODEL_REVISIONS:
         return
 
-    from transformers import (
-        AutoModel,
-        AutoModelForCausalLM,
-        AutoModelForImageTextToText,
-        AutoModelForSequenceClassification,
-        AutoProcessor,
-        AutoTokenizer,
-    )
+    from transformers import PreTrainedModel, PreTrainedTokenizerBase, ProcessorMixin
 
-    def create_wrapper(original_method):
-        @wraps(original_method)
-        def wrapper(pretrained_model_name_or_path, *args, **kwargs):
-            # Direct lookup - only inject if model_id is in the override dict
+    def create_classmethod_wrapper(original_classmethod):
+        # Extract the underlying function from the classmethod
+        original_func = original_classmethod.__func__
+
+        @wraps(original_func)
+        def wrapper(cls, pretrained_model_name_or_path, *args, **kwargs):
+            # Direct lookup: only inject if model_id is in the override dict
             if pretrained_model_name_or_path in MODEL_REVISIONS:
                 if "revision" not in kwargs:
                     kwargs["revision"] = MODEL_REVISIONS[pretrained_model_name_or_path]
 
-            return original_method(pretrained_model_name_or_path, *args, **kwargs)
+            return original_func(cls, pretrained_model_name_or_path, *args, **kwargs)
 
-        return wrapper
+        # Re-wrap as classmethod
+        return classmethod(wrapper)
 
     # Patch all transformers Auto* classes
     for cls in [
-        AutoModel,
-        AutoModelForCausalLM,
-        AutoModelForSequenceClassification,
-        AutoTokenizer,
-        AutoProcessor,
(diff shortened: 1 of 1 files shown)


qgallouedec:
I think it's useful, thanks! One small suggestion, can you add a `create_pr` arg in https://github.com/huggingface/trl/blob/ae841e35130720fd5e7e218d4c77f13e352bd785/scripts/generate_tiny_models.py#L101, so that when testing/pplaying with generate_tiny_models, we can just add `create_pr=True`.

tests/conftest.py
```diff
@@ -13,11 +13,73 @@
...
+        return wrapper
+
+    # Patch all transformers Auto* classes
+    for cls in [
+        AutoModel,
+        AutoModelForCausalLM,
+        AutoModelForSequenceClassification,
+        AutoTokenizer,
+        AutoProcessor,
+        AutoModelForImageTextToText,
```
qgallouedec:
```suggestion
        AutoModel,
        AutoModelForCausalLM,
        AutoModelForImageTextToText,
        AutoModelForSequenceClassification,
        AutoProcessor,
        AutoTokenizer,
```

tests/conftest.py
```diff
@@ -13,11 +13,73 @@
...
+    ]:
+        if hasattr(cls, "from_pretrained"):
+            monkeypatch.setattr(cls, "from_pretrained", create_wrapper(cls.from_pretrained))
```
qgallouedec: they all have `from_pretrained`

```suggestion
        monkeypatch.setattr(cls, "from_pretrained", create_wrapper(cls.from_pretrained))
```

albertvillanova:
While testing locally, I discovered that the function `create_model_from_path` was not patched.

In my first implementation, I was patching only Auto* classes, which didn't cover `create_model_from_path` when it infers architecture.

Therefore, I changed the implementation to patch the base classes directly, so it works with any `Class.from_pretrained`.:
- `PreTrainedModel` → covers ALL model classes
- `PreTrainedTokenizerBase` → covers ALL tokenizers
- `ProcessorMixin` → covers ALL processors

When patching base class classmethods, we must:
- Extract the underlying function with `.__func__`
- Wrap it with proper `cls` parameter handling
- Re-wrap as `classmethod(wrapper)`

This approach covers:
- Direct loading via `Auto*` classes
- Direct loading via specific classes
- `create_model_from_path` with explicit architecture
- `create_model_from_path` with inferred architecture

qgallouedec:
You're right, good catch!

PR #4773, opened 2026-01-04
make dpo compatible with qwen3vl

flutist:
# What does this PR do?

Fixes # (issue)
make dpo compatible with qwen3vl

## Before submitting
- [yes ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ yes] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [no ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [yes ] Did you make sure to update the documentation with your changes?
- [ no] Did you write any new necessary tests?

## Who can review?


qgallouedec:
Thanks, can you add a test case using `trl-internal-testing/tiny-Qwen3VLForConditionalGeneration` in `TestDPOVisionTrainer` to ensure it works properly?

flutist:
> Thanks, can you add a test case using `trl-internal-testing/tiny-Qwen3VLForConditionalGeneration` in `TestDPOVisionTrainer` to ensure it works properly?

Of course, let me implement one.

flutist:
> Thanks, can you add a test case using `trl-internal-testing/tiny-Qwen3VLForConditionalGeneration` in `TestDPOVisionTrainer` to ensure it works properly?

I have finished the work, and all the test is passed.
image

flutist:
@qgallouedec could you help to do a code review for this request?

flutist:
image
One more thing: these three default values ​​are all quite small.  During actual training, it's easy to overlook setting these values, which can lead to errors. Do they need to be increased?
@qgallouedec

qgallouedec:
> these three default values ​​are all quite small. During actual training, it's easy to overlook setting these values, which can lead to errors. Do they need to be increased? @qgallouedec

`max_prompt_length` and `max_completion_length` should be removed as part of #3906

trl/trainer/dpo_trainer.py
```diff
@@ -184,6 +188,66 @@ def torch_call(self, examples: list[list[int] | Any | dict[str, Any]]) -> dict[s
         return output
 
 
+@dataclass
+class QWEN3VLDataCollatorForPreference(DataCollatorMixin):
```
qgallouedec: do we really need a new collator for this?

flutist: Yes, because qwen3vl adopts a different image stacking method. Using the original image processing approach would cause errors in the model's internal processing, leading to exceptions. In qwen3vl, the 0th dimension of pixel_value must be the image token dimension instead of the patch dimension, and segmentation is then performed using image_grid_thw. This is where it differs from other models, and it is also why qwen3vl introduces image_grid_thw.so i introduce a new collator to cat pixel_value for qwen3vl.

flutist:
> > these three default values ​​are all quite small. During actual training, it's easy to overlook setting these values, which can lead to errors. Do they need to be increased? @qgallouedec
> 
> `max_prompt_length` and `max_completion_length` should be removed as part of #3906

got that, thanks

trl/trainer/dpo_trainer.py
flutist: I think after merging this code, it should be extended from Qwen3VL to the full QwenVL series, i will continue the work.

flutist:
@qgallouedec 
Are there any codes that need further modification?
I need your feedback.
Thanks much.

trl/trainer/dpo_trainer.py
flutist: Do I need to refactor the code here? For example, removing duplicate code and passing the image stacking method as a parameter to the collator class?

qgallouedec: I'm not sure what's the best way to address this right now. Having an additional callback for only one model does not seem like a viable long-term solution to me. I will leave this PR open until we find a better solution.

flutist: Okay, I understand what you mean. I will refactor the code to avoid introducing new collator.

flutist: Thanks so much for the feedback! I’ve gone ahead and updated the code based on your suggestions. I’d really appreciate it if you could take another look whenever you have a chance—I’d love to hear your thoughts on the new version. Thanks again!

flutist: I’ve completed comprehensive testing on the Qwen3VL model using DPO across various batch and multi-GPU setups, and I’m happy to report that everything is running smoothly. We’ve also seen some great accuracy improvements in our internal projects. I’d love to see this integrated into the main branch if you think it’s a good fit, as I’ve put a lot of care into making sure it’s robust. Thanks so much for your time, and I’d be more than happy to hear any further feedback you might have!

flutist: @qgallouedec 
Are there any codes that need further modification thant I can do?
Thanks so much for your time, and I’d be more than happy to hear any further feedback you might have!

qgallouedec:
thanks for the fix, we refactored dpo in #3906, and it seems like we don't need this special case anymore!

```
$ pytest tests/test_dpo_trainer.py::TestDPOTrainer::test_train_vlm[trl-internal-testing/tiny-Qwen3VLForConditionalGeneration]
===================================================================== test session starts =====================================================================
platform linux -- Python 3.13.11, pytest-9.0.2, pluggy-1.6.0
rootdir: /fsx/qgallouedec/trl
configfile: pyproject.toml
plugins: anyio-4.12.1, datadir-1.8.0
collected 1 item                                                                                                                                              

tests/test_dpo_trainer.py .                                                                                                                             [100%]

===================================================================== 1 passed in 45.42s ======================================================================
```

PR #4770, opened 2026-01-02
Hotfix: Set float32 as default dtype for testing tiny models

albertvillanova:
Set float32 as default dtype for testing tiny models, after the merge in `transformers` of this PR:
- https://github.com/huggingface/transformers/pull/42805

Fix #4748.

PR #4778, opened 2026-01-06
Set dtype default to float32

albertvillanova:
Set dtype default to float32.

Follow-up to:
- #4770

### DIFF
--- a/examples/scripts/openenv/sudoku.py
+++ b/examples/scripts/openenv/sudoku.py
@@ -1,4 +1,4 @@
-# Copyright 2020-2025 The HuggingFace Team. All rights reserved.
+# Copyright 2020-2026 The HuggingFace Team. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


trl/trainer/utils.py
```diff
@@ -1136,13 +1136,13 @@ def create_model_from_path(
         kwargs (`dict`):
             Initialization keyword arguments to pass to the model's `from_pretrained` method. When `'dtype'` is
             specified, it can be either a `torch.dtype` or one of the strings: `'bfloat16'`, `'float16'`, `'float32'`,
-            or `'auto'`.
+            or `'auto'`. If not explicitly set, `dtype` defaults to `'float32``.
```
qgallouedec:
```suggestion
            or `'auto'`. If not explicitly set, `dtype` defaults to `'float32'`.
```

qgallouedec:
For the record, QLoRA with DPO will still force the model dtype to be in fp32:

```python
from peft import LoraConfig
from trl import DPOTrainer
from transformers import BitsAndBytesConfig


    dtype="float32",
    quantization_config=BitsAndBytesConfig(load_in_4bit=True),
)
dataset = load_dataset("trl-internal-testing/zen", "standard_preference", split="train")

trainer = DPOTrainer(
    model=model,
    train_dataset=dataset,
    peft_config=LoraConfig(),
)
trainer.train()
```

but in my opinion it's fine, it will be fixed by #3906

albertvillanova:

Although the CI was green, I see you set float32 dtype in some tests, but not in others. I am wondering what criteria you used.

PR #3950, opened 2025-08-25
fix bug when using dataset streaming by accelerate

kaixuanliu:
When running example using cmd line with dataset streaming: 
```
accelerate launch examples/scripts/dpo_vlm.py \
    --dataset_name HuggingFaceH4/rlaif-v_formatted \
    --dataset_streaming \
    --model_name_or_path Qwen/Qwen2.5-VL-3B-Instruct \
    --per_device_train_batch_size 2 \
    --max_steps 100 \
    --gradient_accumulation_steps 32 \
    --dataset_num_proc 32 \
    --output_dir dpo_idefics_rlaif-v \
    --torch_dtype bfloat16 \
    --gradient_checkpointing \
    --use_peft \
    --lora_target_modules=all-linear \
    --report_to wandb
```
It will crash and return error:
```
[rank0]: Traceback (most recent call last):
[rank0]: File "/usr/local/lib/python3.11/dist-packages/accelerate/data_loader.py", line 822, in _fetch_batches
[rank0]: batch = concatenate(batches, dim=0)
[rank0]: ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: File "/usr/local/lib/python3.11/dist-packages/accelerate/utils/operations.py", line 616, in concatenate
[rank0]: return type(data[0])({k: concatenate([d[k] for d in data], dim=dim) for k in data[0].keys()})
[rank0]: ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: File "/usr/local/lib/python3.11/dist-packages/accelerate/utils/operations.py", line 616, in <dictcomp>
[rank0]: return type(data[0])({k: concatenate([d[k] for d in data], dim=dim) for k in data[0].keys()})
[rank0]: ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: File "/usr/local/lib/python3.11/dist-packages/accelerate/utils/operations.py", line 619, in concatenate
[rank0]: return torch.cat(data, dim=dim)
[rank0]: ^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 383 but got size 304 for tensor number 1 in the list.

[rank0]: The above exception was the direct cause of the following exception:

[rank0]: Traceback (most recent call last):
[...]

PR #4784, opened 2026-01-07
Add initial multi-GPU CI tests for distributed training

qgallouedec:
This PR addresses a major CI blind spot: until now we only tested the single-GPU path. It introduces:

* A minimal multi-GPU distributed CI test (smoke test)
* A foundation for expanding distributed test coverage over time

To keep in mind:

* We should not test every possible setting in multi-GPU, only the critical ones. For example, dataset streaming previously caused failures in multi-GPU but not in single-GPU, which is why we include a test for it. We should keep the number of such tests minimal since they are slow to run.

Supersedes #3902

### DIFF
--- a/tests/distributed/test_distributed.py
+++ b/tests/distributed/test_distributed.py
@@ -14,12 +14,11 @@
 
 import os
 import subprocess
-import tempfile
 from pathlib import Path
 
 import pytest
 
-from ..testing_utils import require_torch_multi_accelerator
+from ..testing_utils import TrlTestCase, require_torch_multi_accelerator
 
 
 ROOT = Path(__file__).resolve().parents[2]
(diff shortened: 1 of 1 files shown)


tests/distributed/test_distributed.py
```diff
@@ -0,0 +1,103 @@
...
+                "--output_dir", tmpdir,
+                "--model_name_or_path", "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
+                "--dataset_name", "trl-internal-testing/zen",
+                "--dataset_config", "standard_language_modeling",
+                "--dataset_streaming",
+                "--max_steps", "3",
+            ],
+            os.environ.copy(),
+        )
+        # fmt: on
```
qgallouedec: For the record, this has been broken for a long time, and fixed by #3950.

PR #4785, opened 2026-01-07
Add support for GDPO: Group reward-Decoupled Normalization Policy Optimization for Multi-reward RL Optimization

nbasyl:
# GDPO: Group reward-Decoupled Normalization Policy Optimization for Multi-reward RL Optimization

This PR introduces **GDPO: Group reward-Decoupled Normalization Policy Optimization for Multi-reward RL Optimization** — a new RL training method for multi-reward RL optimization that consistently improves per-reward convergence in multi-reward RL training over GRPO across a wide range of tasks, see [https://nvlabs.github.io/GDPO/](https://nvlabs.github.io/GDPO/) for experiment results. We also provide a easy-to-use, slurm-free TRL training example that enable the community to quickly validate GDPO’s effectiveness over GRPO, see [https://github.com/NVlabs/GDPO/tree/main/trl-GDPO](https://github.com/NVlabs/GDPO/tree/main/trl-GDPO)

---

## 🔍 Motivation

 - “What's wrong with using GRPO for multi-reward RL training?”  We find that the common practice of applying GRPO to multi-reward RL optimization normalize distinct rollout reward combinations causes them to collapse into identical advantage values, reducing the resolution of the training signal and resulting in suboptimal convergence

image

---

## 🧠 Method Overview

- “GDPO comes to the rescue!” To address the fundamental limitation of GRPO, we propose Group reward-Decoupled normalization Policy Optimization (GDPO), which performs group-wise normalization on each reward independently prior to aggregation. This contrasts with GRPO, which applies group-wise normalization directly to the summed rewards

image

## 📦 Files Added / Modified
GDPO is implemented as an extension of GRPO, analogous to enabling DoRA on top of LoRA in HF-PEFT, making the modification relatively straightforward. An apply_gdpo flag is added to GRPOConfig to allow users to enable GDPO when training GRPO with multiple rewards.

docs/source/paper_index.md
trl/trainer/grpo_config.py
trl/trainer/grpo_trainer.py

## 🙌 Request for Review
@qgallouedec Please take a look and let me know if you have any questions. Many thanks! Paper will be submitted to Arxiv tomorrow, and I will update the paper link to the paper_index.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ from trl.rewards import accuracy_reward
 dataset = load_dataset("trl-lib/DeepMath-103K", split="train")
 
 trainer = GRPOTrainer(
-    model="Qwen/Qwen2-0.5B-Instruct",
+    model="Qwen/Qwen2.5-0.5B-Instruct",
     reward_funcs=accuracy_reward,
     train_dataset=dataset,
 )


trl/trainer/grpo_trainer.py
```diff
@@ -397,6 +397,11 @@ def __init__(
         else:
             self.reward_weights = torch.ones(len(reward_funcs), dtype=torch.float32)
 
+        if args.apply_gdpo:
+            self.apply_gdpo = True
+        else:
+            self.apply_gdpo = False
```
pramodith:
```suggestion
self.apply_gdpo = apply_gdpo
```

trl/trainer/grpo_trainer.py
```diff
@@ -2011,38 +2016,67 @@ def _generate_and_score_completions(
...
+                all_reward_advantage.append(each_reward_advantage)
+
+            combined_reward_advantage = torch.stack(all_reward_advantage, dim=1)
+            pre_bn_advantages = (combined_reward_advantage * self.reward_weights.to(device).unsqueeze(0)).nansum(dim=1)
+
+            ## compute batch-wise mean and std
+            bn_advantages_mean = pre_bn_advantages.mean()
+            bn_advantages_std = pre_bn_advantages.std()
+
+            advantages = (pre_bn_advantages - bn_advantages_mean) / (bn_advantages_std + 1e-4)
```
pramodith: I think all of this can be vectorized without using a for-loop.

pramodith (commented):
Cool work! Can you add some test cases too?

nbasyl:
Hi @pramodith, thanks for taking the time to review my PR. I’ve vectorized the GDPO implementation and added a test case in test_grpo_trainer.py. Please take a look when you get a chance!

yu-ki3406:
We have also experimented with a similar approach and confirmed its effectiveness in domains such as machine translation and simulated control. Therefore, we can recommend this method as an excellent technique. 
For details, please refer to https://arxiv.org/abs/2509.22047 (Ichihara et al. 2025).

nbasyl:
Hi @qgallouedec, thanks for the code cleanup! One small suggestion: would you be open to using `multi_objective_aggregation="gdpo"` instead of "normalize_then_sum"? We found the final batch normalization to be important for training stability in the paper, and this would also improve alignment and visibility of our method.

qgallouedec:
Thanks for the feedback!

I replaced the earlier `apply_gdpo` flag with a more general `multi_objective_aggregation` enum, since I expect other options might be proposed in the future (similar to `scale_reward`). From there, it’s not obvious how to name the alternative alongside `gdpo`: calling it `grpo` wouldn’t be quite correct, since the GRPO paper (as far as I remember) doesn’t specify how rewards are aggregated. Mixing `gdpo` with something like `sum_then_normalize` would also be inconsistent, as it would combine paper-specific labels with self-explanatory ones.

In general, I tend to favor descriptive, self-explanatory names when possible, since paper labels can be hard to interpret for users unfamiliar with the reference. That said, we can still revisit the naming if this proves confusing in practice or if we add more variants down the road. We could consider adding an alias or something like that.

qgallouedec:
> We found the final batch normalization to be important for training stability in the paper

Yes, that's indeed how it's implemented, the advantages are normalized per batch (it's not highly explicit in the name, but I tried to explain in more details in the paper index section)

qgallouedec:
@nbasyl Also, I think it makes sense to have `normalize_then_sum ` as the default. Your paper and others show that this is a better way to aggregate than `sum_then_normalize `. If you’re interested, we can change this default in a follow-up PR.

Edit: fixed inversion `normalize_then_sum` and `sum_then_normalize`

nbasyl:
@qgallouedec, got it! Thanks for the effort.

Issue #4786, opened 2026-01-08
KTO refactoring


albertvillanova:
Systematically refactor KTOTrainer to align with modern trainer architecture patterns, improving maintainability, reducing complexity, and adding missing features.

## Problem
KTOTrainer currently has complex code with several architectural issues.

### 1: Foundation & Infrastructure
Establish modern patterns without breaking functionality

- [x] Modernize Model Initialization: #4783
  - [x] Use `create_model_from_path()` helper
  - [x] Simplify dtype and device_map handling
  - [x] Clean up ref_model initialization
- [ ] Improve Config Validation: #4787

### 2: Dataset Processing Modernization
Adopt standard trainers' clean preprocessing patterns

- [ ] Extract `_prepare_dataset` method: #4788

### 3: Reference Model Handling
Simplify ref model management

### 4: Loss Computation & Forward Pass
Reduce duplication in loss computation

### 5: Metrics & Logging
Match standard metrics system

### 6: Evaluation & Generation
Modern evaluation features

### 7: Documentation
Complete, clear documentation

### Phase 8: Advanced Features
Feature parity with standard trainers

### Phase 9: Testing
Production ready

albertvillanova:
After internal discussion with @qgallouedec, we are coordinating to refactor both DPO and KTO. See related:
- #3906

## WIP: Features to Remove

First, remove some features. Benefits:
- Less code to refactor
- Simpler architecture
- Easier DPO alignment
- Clearer core functionality
- Faster refactoring

### A. Remove Encoder-Decoder Support
PR:
- [x] #4792

Why Remove:
  - Adds ~200-300 lines of conditional logic
  - Complicates every dataset processing step
  - Rarely used (most preference learning is causal LM)
  - Different tokenization, label handling, model kwargs
  - Testing burden (2x test coverage needed)

Already broken, minimal user impact.

### B. Separate Prompt/Completion Truncation
PR:
- [x] #4808

Why Remove:
  - Adds ~80 lines of complex truncation logic
  - Confusing UX (users don't know which to set)
  - Most users just want max_length
  - truncation_mode is redundant with tokenizer settings

Replace with simple `max_length` only.

### C. Remove ref_model_init_kwargs
PR:
- [x] #4837

Why?
- Simplify model instantiation

### D. Remove base_model_attribute_name
PR:
- [ ] #4862

Why?
- Simplify base model retrieval logic

albertvillanova:
I am resuming this task: Align KTO with DPO.

[...]

PR #4787, opened 2026-01-08
Refactor KTO [2/N]: Improve config validation in KTOConfig

albertvillanova:
Refactor KTO [2/N]: Improve config validation in KTOConfig.

This PR moves validation logic from `KTOTrainer.__init__()` to `KTOConfig.__post_init__()` for earlier error detection, better user experience, and cleaner separation of concerns.
- Principle: Fail-fast with clear, actionable error messages

Part of:
- #4786

## Problem

**Before**:
  - Validation scattered between config and trainer
  - Errors discovered late (during trainer initialization)
  - Invalid configs could be created and passed around
  - generate_during_eval validated in trainer
  - No validation for loss_type, truncation_mode, beta, weights
  - No validation for max_length relationships

**After**:
  - All validation centralized in `KTOConfig.__post_init__()`
  - Errors discovered immediately at config creation
  - Invalid configs cannot be created
  - Clear, actionable error messages with guidance
  - Comprehensive validation for all critical parameters

### DIFF
--- a/trl/experimental/kto/kto_config.py
+++ b/trl/experimental/kto/kto_config.py
@@ -256,8 +256,10 @@ class KTOConfig(TrainingArguments):
     )
 
     def __post_init__(self):
+        # Set bf16 default
         self.bf16 = not (self.fp16) if self.bf16 is None else self.bf16
 
+        # Handle deprecated use_liger_loss
         if self.use_liger_loss is not None:
             warnings.warn(
                 "The `use_liger_loss` argument is deprecated and will be removed in version 0.28.0. Please use "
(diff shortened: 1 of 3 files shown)


trl/experimental/kto/kto_trainer.py
```diff
@@ -455,12 +454,6 @@ def make_inputs_require_grad(module, input, output):
 
                 model.get_input_embeddings().register_forward_hook(make_inputs_require_grad)
 
-        if args.generate_during_eval and not (is_wandb_available() or is_comet_available()):
-            raise ValueError(
-                "`generate_during_eval=True` requires Weights and Biases or Comet to be installed."
-                " Please install `wandb` or `comet-ml` to resolve."
-            )
-
```
albertvillanova, line 463: What do you think about this change? Before it was checked at trainer instantiation and now at config creation. This has an import time side effect: it imports wandb/comet checkers immediately.

qgallouedec: In general, I prefer to keep all these kinds of checks in one place (in `MyTrainer.__init__`), so that the configuration remains minimal, and we avoid unintended duplication. That said, the codebase isn’t entirely consistent on this point, see for example:

https://github.com/huggingface/trl/blob/1a939718513286684b40d21329b3ddcc0d39a7ea/trl/trainer/grpo_config.py#L845-L888

For this specific argument, I think it could probably be removed. See point 7 here:
https://github.com/huggingface/trl/pull/3906#issue-3326451170

trl/experimental/kto/kto_config.py
```diff
@@ -267,4 +269,60 @@ def __post_init__(self):
...
+            raise ValueError(
+                f"desirable_weight must be positive, got {self.desirable_weight}. "
+                "This weight is used to balance desirable and undesirable examples."
+            )
+
+        if self.undesirable_weight <= 0:
+            raise ValueError(
+                f"undesirable_weight must be positive, got {self.undesirable_weight}. "
+                "This weight is used to balance desirable and undesirable examples."
+            )
```
qgallouedec, line 301: I’m generally in favor of keeping validation as minimal as possible. Users experimenting with parameters are expected to understand the semantics, and overly defensive validation tends to grow without bound. I’d rather keep checks limited to hard invariants (i.e. impossible combinations), like here:

[https://github.com/huggingface/trl/blob/1a939718513286684b40d21329b3ddcc0d39a7ea/trl/trainer/grpo_config.py#L876-L882](https://github.com/huggingface/trl/blob/1a939718513286684b40d21329b3ddcc0d39a7ea/trl/trainer/grpo_config.py#L876-L882)

This one is required by the sampling logic: if you remove the check, you can end up in a failure mode that would be very hard to debug. Beyond that, I’d rather let mistakes fail naturally.

Also, strict value checks can unnecessarily block experimentation. For example, negative (un)desirable weights aren’t mathematically invalid: they change the objective (and are almost certainly not what we want), but I’m not sure we should hard-forbid them at the config level unless they actually cause breakage.

albertvillanova:
Superseded.

PR #4789, opened 2026-01-08
Temporarily Work Around init_communicator self.device Initialization Failure with vLLM-Ascend Server and TRL 0.26.2 in Training

ShareableXue:
When using vllm-ascend as the server and TRL version 0.26.2 for GRPO training, the init_communicator component of the TRL vllm-serve fails to initialize self.device. Temporary workaround.

### DIFF
--- a/trl/scripts/vllm_serve.py
+++ b/trl/scripts/vllm_serve.py
@@ -37,6 +37,8 @@ from trl.import_utils import (
     is_vllm_available,
 )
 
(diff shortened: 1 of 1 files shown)


qgallouedec:
thanks for the PR, can you resolve the conflicts and ensure only the relevant lines are modified? 
Screenshot 2026-01-09 at 9 40 35 AM

PR #4117, opened 2025-09-22
update guided decoding param to structured outputs

jiqing-feng:
The latest vllm changed the `guided_decoding` to `structured_outputs` in this PR [22772](https://github.com/vllm-project/vllm/pull/22772). We also need to update this parameter in the vllm client and vllm serve.

Hi @qgallouedec . Would you please review this PR? Thanks!

PR #4791, opened 2026-01-08
[EXAMPLES] Update wordle to new openenv release

burtenshaw:
This PR make small changes to the wordle openenv example so it's compatible with openenv-0.2.0

- update docstrings with latest install methods
- uses one method for interaction (url)
- fixes vLLM utils @kashif


sergiopaniego (approved):
LGTM! I'll update the rest of the scripts/notebooks in a follow-up PR to use the updated OpenEnv version

trl/experimental/openenv/utils.py
```diff
@@ -23,7 +23,7 @@
 
 if is_vllm_available():
     from vllm import SamplingParams
-    from vllm.sampling_params import StructuredOutputsParams
+    from vllm.sampling_params import GuidedDecodingParams
```
qgallouedec, line 26: We will have to revert this, see https://github.com/huggingface/trl/pull/4117

kashif: ah damn!

burtenshaw:
Sorry about this @qgallouedec. I should have been clearer that I needed those changes reviewed. 🤦

PR #4799, opened 2026-01-09
Remove `DbrxForCausalLM` support

qgallouedec:
This model:
- has been repeatedly broken
- is not very popular: the only models available on the hub are basically tiny versions, see https://huggingface.co/models?other=dbrx&p=2&sort=created
- the "main model" is not public anymore, see https://huggingface.co/databricks/dbrx-instruct/ gives 404
- now it causes the CI to fail with ```RuntimeError: You set `ignore_mismatched_sizes` to `False`, thus raising an error. For details look at the above report!```

Consequently, I suggest that we simply remove the support for this model


albertvillanova (approved):
Thanks, much simpler/cleaner now! I am merging it to fix the CI.

PR #4809, opened 2026-01-12
Refactor _compat module

albertvillanova:
Refactor _compat module.

The refactoring follows Pythonic principles: simple, explicit, readable, with appropriate (not excessive) error handling.

Follow-up to:
- #4807

### Key Improvements

1. Consistent Structure 

   - Organized all patches into separate, named functions
   - Each patch is now independently testable and callable
   - Clear separation between vLLM and transformers patches

2. Error Handling

   - Added try/except blocks around patch logic
   - Patches now emit warnings on failure instead of breaking TRL imports
   - Used stacklevel=2 for proper warning source attribution
   - Specific exceptions for vLLM patches (ImportError, AttributeError)
   - Broad Exception catch for liger patch (version parsing can fail in many ways)

3. Consistent Documentation

   - All patches now follow the same comment format:
     - Issue: <what's broken>
     - Removed/Fixed in: <upstream PR/version>
     - This can be removed when: <condition>
   - Clear, concise docstrings for each function

4. Better Code Organization

   - Patches grouped by dependency (vLLM, transformers)
   - Application logic at the bottom is clean and obvious
   - No more mixing of inline code with functions

5. Improved Maintainability

   - Easy to see all active patches at a glance
   - Simple to add/remove patches as dependencies evolve
   - Clear when each patch can be removed
   - Functions can be tested independently

  6. Defensive Programming

  - Early return in transformers patch if liker-kernel package not available
  - Version checks protected by try/except
  - Import failures won't crash the application

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -52,7 +52,7 @@ from .testing_utils import (
 
 
 if is_peft_available():
-    from peft import LoraConfig, PeftModel
+    from peft import LoraConfig, PeftModel, get_peft_model
 
 
 class TestGetHighEntropyMask(TrlTestCase):
@@ -284,8 +284,8 @@ class TestGRPOTrainer(TrlTestCase):
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
 
     @require_peft
-    def test_training_peft(self):
-        model = AutoModelForCausalLM.from_pretrained("trl-internal-testing/tiny-Qwen2ForCausalLM-2.5")
+    def test_training_peft_config(self):
+        model = AutoModelForCausalLM.from_pretrained("trl-internal-testing/tiny-Qwen2ForCausalLM-2.5", dtype="float32")
         base_param_names = [f"base_model.model.{n}" for n, _ in model.named_parameters()]
         dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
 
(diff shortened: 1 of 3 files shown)


trl/_compat.py
```diff
@@ -20,99 +20,134 @@
...
+
+# Apply vLLM patches
+if is_vllm_available():
+    _patch_vllm_logging()
+    _patch_vllm_disabled_tqdm()
+    _patch_vllm_cached_tokenizer()
 
-_maybe_patch_transformers_hybrid_cache()
+# Apply transformers patches
+_patch_transformers_hybrid_cache()
```
qgallouedec, line 153: tiny consistency improvement

```suggestion
if is_liger_kernel_available():
    _patch_liger_hybrid_cache()
```

albertvillanova: @qgallouedec, I am not sure it is a good idea to rename `_patch_transformers_hybrid_cache` to `_patch_liger_hybrid_cache`.

I am following the naming approach "name WHAT you patch", not "WHY" you patch; and we are patching `transformers`, not `liger_kernel`:
- Describes the code being modified
- "This function patches transformers"

Why I think naming by WHAT is better:
- Consistency with existing code: `_patch_vllm_*`
- Clarity about side effects:
  - When debugging transformers issues, you want to grep for "patch_transformers"
  - The function name should tell you which module gets modified
- Avoids confusion:
  - `_patch_liger_kernel_*` may be understood as if we're modifying liger_kernel's code

If we agree on this, then I would not add the `if is_liger_kernel_available()` condition: we are patching `transformers` and transformers is always available (contrary to the patches above: they patch `vllm` and then only if `vllm`is available).
- I would leave all the intricate "if" conditions inside the patch function (as its responsibility).

qgallouedec: That makes sense

qgallouedec (approved):

Issue #4782, opened 2026-01-07
`RuntimeError: Expected to mark a variable ready only once` when training with PEFT


qgallouedec:
### Reproduction

```python
from datasets import load_dataset
from trl import DPOTrainer
from peft import LoraConfig

if __name__ == "__main__":
    dataset = load_dataset("trl-internal-testing/zen", "standard_preference", split="train")

    trainer = DPOTrainer(
        model="Qwen/Qwen2.5-0.5B",
        train_dataset=dataset,
        peft_config=LoraConfig(),
    )
    trainer.train()
```

outputs:

```
/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/torch/utils/checkpoint.py:85: UserWarning: None of the inputs have requires_grad=True. Gradients will be None
  warnings.warn(
/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/torch/utils/checkpoint.py:85: UserWarning: None of the inputs have requires_grad=True. Gradients will be None
  warnings.warn(
[rank1]: Traceback (most recent call last):
[rank1]:   File "/fsx/qgallouedec/trl/bug_dpo_peft.py", line 13, in <module>
[rank1]:     trainer.train()
[rank1]:   File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/transformers/trainer.py", line 2325, in train
[rank1]:     return inner_training_loop(
[rank1]:            ^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/transformers/trainer.py", line 2674, in _inner_training_loop
[rank1]:     tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
[rank1]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/transformers/trainer.py", line 4071, in training_step
[rank1]:     self.accelerator.backward(loss, **kwargs)
[rank1]:   File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/accelerate/accelerator.py", line 2852, in backward
[rank1]:     loss.backward(**kwargs)
[...]

PR #4811, opened 2026-01-12
Switch gradient checkpointing default to use_reentrant=False (PyTorch recommended)

qgallouedec:
### Set `use_reentrant=False` by default 

This PR defaults gradient checkpointing to `use_reentrant=False` in TRL when no explicit value is provided.

PyTorch now recommends the non-reentrant checkpointing variant, see https://docs.pytorch.org/docs/stable/checkpoint.html. However, Transformers still defaults to `use_reentrant=True` because it was explicitly pinned in the past to silence a PyTorch warning during a transition period, and the default was never updated afterward.

Until this is fixed upstream and released (see https://github.com/huggingface/transformers/pull/43203), TRL aligns with the current PyTorch recommendation by setting `use_reentrant=False` by default, while fully preserving any user-provided value.

### Fixes`Expected to mark a variable ready only once`

This PR fixes an issue that seems unrelated, but is: https://github.com/huggingface/trl/issues/4782

### Remove `ScriptArguments.gradient_checkpointing_use_reentrant`

`ScriptArguments.gradient_checkpointing_use_reentrant` exists but is never used. This is misleading, so this PR removes this argument.

### DIFF
--- a/tests/experimental/test_nash_md_trainer.py
+++ b/tests/experimental/test_nash_md_trainer.py
@@ -178,7 +178,6 @@ class TestNashMDTrainer(TrlTestCase):
         lora_config = LoraConfig(r=8, lora_alpha=16, lora_dropout=0.1, bias="none", task_type="CAUSAL_LM")
         # self.model from setUp is a base AutoModelForCausalLM
         peft_model_instance = get_peft_model(self.model, lora_config)
-        peft_model_instance.enable_input_require_grads()
 
         training_args = NashMDConfig(
             output_dir=self.tmp_dir,
--- a/tests/experimental/test_xpo_trainer.py
+++ b/tests/experimental/test_xpo_trainer.py
@@ -127,7 +127,6 @@ class TestXPOTrainer(TrlTestCase):
     def test_training_pre_pefted_model_implicit_ref(self):
         lora_config = LoraConfig(r=8, lora_alpha=16, lora_dropout=0.1, bias="none", task_type="CAUSAL_LM")
         peft_model_instance = get_peft_model(self.model, lora_config)
-        peft_model_instance.enable_input_require_grads()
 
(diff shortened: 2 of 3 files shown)


trl/trainer/dpo_config.py
```diff
@@ -500,6 +500,14 @@ class DPOConfig(TrainingArguments):
...
     def __post_init__(self):
         self.bf16 = not (self.fp16) if self.bf16 is None else self.bf16
+
+        # Transformers explicitly set use_reentrant=True in the past to silence a PyTorch warning, but the default was
+        # never updated once PyTorch switched to recommending use_reentrant=False. Until that change lands upstream
+        # (see https://github.com/huggingface/transformers/pull/43203) and is released (most likely in 5.0.0), we
+        # default to the recommended non-reentrant behavior here, while preserving any user-provided value.
+        self.gradient_checkpointing_kwargs = self.gradient_checkpointing_kwargs or {}
+        self.gradient_checkpointing_kwargs.setdefault("use_reentrant", False)
+
```
albertvillanova, line 513: First, what about setting an explicit condition on the transformers versions where this setting is necessary? That would help maintainability by making clear when this PR is no longer necessary and can be reverted in the future.

Second, does it make sense to have `self.gradient_checkpointing_kwargs = {"use_reentrant": False}` even if the user passed `gradient_checkpointing=False`?

qgallouedec: Good point, addressed in 269189c35b1c6c47c2c623afe09c6a402c85cb65

qgallouedec:
> Additionally, there are 2 new CI errors that I think we should check if they are caused by this PR and then if they should be fixed as well:

Fixed, `enable_input_require_grads()` is needed not only for reentrant.

PR #4817, opened 2026-01-13
Simplify version checks in compat patches

albertvillanova:
Simplify version checks in compat patches.

This PR refactors the compatibility patch logic to make version checks more robust and maintainable. The main improvement is the introduction of a unified helper function for determining whether a package needs patching, which reduces code duplication and simplifies future updates.

Key improvements:

**Refactoring compatibility checks:**
* Introduced a new `_package_needs_patch` helper function to centralize logic for checking if a package is installed and below a fixed version. This function is now used in all patch functions, replacing repeated version-checking code.

**Patch logic simplification:**
* Updated `_patch_vllm_disabled_tqdm`, `_patch_vllm_cached_tokenizer`, and `_patch_transformers_hybrid_cache` to use `_package_needs_patch`, removing redundant imports and manual version checks from each patch function.
* Removed the conditional check around the patch application at the bottom of the file, so that all vLLM patches are always attempted, with each patch function now internally responsible for deciding if it should run.

### DIFF
--- a/tests/experimental/test_bco_trainer.py
+++ b/tests/experimental/test_bco_trainer.py
@@ -56,7 +56,7 @@ class TestBCOTrainer(TrlTestCase):
         training_args = BCOConfig(
             output_dir=self.tmp_dir,
             remove_unused_columns=False,  # warning raised if not set to False
-            learning_rate=0.1,  # increase the learning rate to speed up the test
+            learning_rate=0.1,  # use higher lr because gradients are tiny and default lr can stall updates
             report_to="none",
         )
 
@@ -92,7 +92,7 @@ class TestBCOTrainer(TrlTestCase):
         training_args = BCOConfig(
             output_dir=self.tmp_dir,
             remove_unused_columns=False,  # warning raised if not set to False
-            learning_rate=0.1,  # increase the learning rate to speed up the test
+            learning_rate=0.1,  # use higher lr because gradients are tiny and default lr can stall updates
             precompute_ref_log_probs=True,
             report_to="none",
         )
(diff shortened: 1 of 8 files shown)


trl/_compat.py
```diff
@@ -25,11 +25,32 @@
...
+        package_name (str): Package name.
+        fixed_in_version (str): Version when the issue was fixed and the patch is no longer necessary.
+    """
+    try:
+        from packaging.version import Version
+        from transformers.utils.import_utils import _is_package_available
+
+        is_available, version = _is_package_available(package_name, return_version=True)
+        return is_available and Version(version) < Version(fixed_in_version)
+    except Exception:
```
qgallouedec: I'm a bit afraid that this could silently not apply the patch. Maybe a warning would make sense here?

qgallouedec: I mean, it's never supposed to fail, right?

trl/_compat.py
```diff
@@ -25,11 +25,32 @@
 from .import_utils import is_vllm_available
 
 
+def _package_needs_patch(package_name: str, fixed_in_version: str) -> bool:
```
qgallouedec: The name `_package_needs_patch` is a bit misleading because it claims semantic intent ("needs a patch") while the function actually performs a simple factual check: whether a package is installed and its version is below a given version. In other words, it encodes how callers currently use the result, not what the function computes.

WDYT of `_is_package_version_below` instead. You'd use it as

```python
needs_patch = _is_package_version_below("vllm", "0.11.1")
```

albertvillanova:
Thanks for your excellent review. I agree with both suggestions. I'm addressing them.

trl/_compat.py
albertvillanova: Done.

trl/_compat.py
albertvillanova: Done.

qgallouedec (approved):
LGTM!

PR #4818, opened 2026-01-13
[GRPO] Add parquet logging for completions with individual rewards

qgallouedec:
Introduce a configuration option for saving completions logs to a Hugging Face Hub repository.

See internal discussion https://huggingface.slack.com/archives/C04EX6W3QSY/p1768297223027059

Initally implemented by @davanstrien

Example:

```python
from trl.rewards import accuracy_reward

dataset = load_dataset("trl-lib/DeepMath-103K", split="train")

trainer = GRPOTrainer(
    reward_funcs=accuracy_reward,
    args=GRPOConfig(
        log_completions=True,
        log_completions_hub_repo="qgallouedec/deepmath-completions-logs",
    ),
    train_dataset=dataset,
)
trainer.train()
```

See https://huggingface.co/datasets/qgallouedec/deepmath-completions-logs

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -695,7 +695,7 @@ class GRPOConfig(TrainingArguments):
             "The clipped weights are then multiplied with the advantages and policy model's log probs. "
             "Individual token losses are aggregated by normalizing with the number of active tokens in "
             "the global accumulated batch. This method was introduced in the "
-            "[MiniMax-M1 paper](https://huggingface.co/papers/2506.13585)."
+            "[MiniMax-M1 paper](https://huggingface.co/papers/2506.13585). "
             "'sapo': Soft Adaptive Policy Optimization loss, as introduced in the "
             "[Soft Adaptive Policy Optimization paper](https://huggingface.co/papers/2506.13585). "
             "Replaces hard clipping with a smooth, temperature-controlled gate that adaptively attenuates "
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -17,7 +17,6 @@ import atexit
 import copy
 import importlib.resources as pkg_resources
 import inspect
-import json
 import os
 import sys
 import textwrap
@@ -36,7 +35,7 @@ import torch
 import torch.utils.data
 import transformers
 from accelerate.logging import get_logger
-from accelerate.utils import broadcast_object_list, gather, gather_object, is_peft_model, set_seed
+from accelerate.utils import gather, gather_object, is_peft_model, set_seed
 from datasets import Dataset, IterableDataset
 from huggingface_hub import CommitScheduler, DatasetCard, DatasetCardData, create_repo
 from packaging.version import Version
@@ -52,7 +51,6 @@ from transformers import (
     PreTrainedTokenizerBase,
     ProcessorMixin,
     TrainerCallback,
-    is_bitsandbytes_available,
     is_trackio_available,
     is_wandb_available,
 )
@@ -64,11 +62,10 @@ from ..data_utils import (
     apply_chat_template,
     is_conversational,
     prepare_multimodal_messages,
-    prepare_multimodal_messages_vllm,
 )
 from ..extras.profiling import profiling_context, profiling_decorator
-from ..extras.vllm_client import VLLMClient
-from ..import_utils import is_jmespath_available, is_liger_kernel_available, is_vllm_available
+from ..generation.vllm_generation import VLLMGeneration
+from ..import_utils import is_jmespath_available, is_liger_kernel_available
 from ..models import prepare_deepspeed, prepare_fsdp, unwrap_model_for_generation
 from ..models.utils import _ForwardRedirection, disable_gradient_checkpointing
 from .base_trainer import BaseTrainer
@@ -78,7 +75,6 @@ from .utils import (
     RepeatSampler,
     create_model_from_path,
     disable_dropout_in_model,
-    ensure_master_addr_port,
     entropy_from_logits,
     get_config_model_id,
     identity,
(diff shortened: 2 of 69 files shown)


kashif:
shall we also log the model name and perhaps grpo config in the data card?

qgallouedec:
@kashif 

done in https://github.com/huggingface/trl/pull/4818/commits/51e20987e8470a4670c74c437d32de045f56ca63 👍

trl/templates/completions_dataset_card.md
```diff
@@ -0,0 +1,41 @@
+---
+{{ card_data }}
+---
+
+# TRL Completion logs
```
lewtun: turbo nit:
```suggestion
# TRL Completion Logs
```

trl/trainer/grpo_trainer.py
```diff
@@ -2461,6 +2492,11 @@ def log(self, logs: dict[str, float], start_time: float | None = None) -> None:
             }
 
             df_base = pd.DataFrame(table)
+            df_base = df_base.astype({"step": "int64"})
```
kashif: shall we set the `step` to `int64` above? the we would not need this?

kashif: is ok i do it?

trl/trainer/grpo_trainer.py
```diff
@@ -2453,14 +2484,18 @@ def log(self, logs: dict[str, float], start_time: float | None = None) -> None:
...
                 "prompt": self._logs["prompt"],
                 "completion": self._logs["completion"],
                 **self._logs["rewards"],
                 "advantage": self._logs["advantages"],
             }
 
             df_base = pd.DataFrame(table)
+            df_base.to_parquet(
+                os.path.join(f"{self.args.output_dir}/completions", f"completions_{self.state.global_step}.parquet")
+            )
```
kashif, line 2183: shall we only do this when `log_completions_hub_repo` is not none?

trl/trainer/grpo_trainer.py
```diff
@@ -781,6 +783,35 @@ def cast_outputs_to_original_dtype(module, args, output):
                         reward_func, evaluation_mode=True, device_placement=True
                     )
 
+        if self.is_world_process_zero() and self.args.log_completions_hub_repo is not None:
```
```suggestion
        if self.accelerator.is_main_process and self.args.log_completions_hub_repo is not None:
```
for consistency?

trl/trainer/grpo_trainer.py
kashif, line 2183:
```suggestion
            if self.args.log_completions_hub_repo is not None:
                df_base.to_parquet(
                    os.path.join(f"{self.args.output_dir}/completions", f"completions_{self.state.global_step}.parquet")
                )
```

kashif:
some comments but looks solid, the same pattern can be added to RLOO and GOLD too?

trl/trainer/grpo_trainer.py
```diff
@@ -781,6 +783,35 @@ def cast_outputs_to_original_dtype(module, args, output):
...
+                template_path=str(template_path),
+                log_completions_hub_repo=self.args.log_completions_hub_repo,
+                hub_model_id=self.args.hub_model_id,
+            )
+            card.push_to_hub(self.args.log_completions_hub_repo)
+            os.makedirs(f"{self.args.output_dir}/completions", exist_ok=True)
+            self.commit_scheduler = CommitScheduler(
+                repo_id=self.args.log_completions_hub_repo,
+                repo_type="dataset",
+                path_in_repo=".",
```
albertvillanova: Defaults to the root folder of the repository:
```suggestion
```

trl/templates/completions_dataset_card.md
```diff
@@ -0,0 +1,41 @@
...
+dataset = datasets.load_dataset("{{ log_completions_hub_repo }}")
+```
+
+You can also load the dataset using Polars:
+
+```python
+import polars as pl
+
+# Login using e.g. `huggingface-cli login` to access this dataset if it's private
+df = pl.read_parquet(f"hf://datasets/{{ log_completions_hub_repo }}/**/*.parquet")
```
albertvillanova: The data files are uploaded at the root of the repo.
```suggestion
df = pl.read_parquet(f"hf://datasets/{{ log_completions_hub_repo }}/*.parquet")
```

trl/trainer/grpo_config.py
```diff
@@ -829,6 +844,13 @@ def __post_init__(self):
 
         self.scale_rewards = {True: "group", False: "none"}.get(self.scale_rewards, self.scale_rewards)
 
+        if self.log_completions_hub_repo is not None and not self.log_completions:
+            warnings.warn(
+                "`log_completions_hub_repo` is set but `log_completions` is False; completions will not be logged.",
+                UserWarning,
+                stacklevel=2,
```
albertvillanova: Shouldn't we raise an error instead?

trl/trainer/grpo_trainer.py
qgallouedec: here it's just saving locally, which I guess is fine, we're not pushing anything. That was my initial intuition

trl/trainer/grpo_config.py
qgallouedec: I was unsure, but Codex made the decision and agrees with you.

qgallouedec:
> Thanks for implementing this useful feature.
> 
> Just some comments below and a few questions in relation with the corresponding internal discussion:
> 
> 1. In the discussion, they commented about using `LogCompletionsCallback`, but your implementation is done within the trainer's `log` method.
>    
>    * Could you explain why?

`LogCompletionsCallback` would generate completion and log them. In online trainers (GRPO, RLOO, etc) we already have completion, because the training involves generation. So instead of re-generating (heavy), we just log the ones that already exist.

> 2. In the discussion, they commented about using Xet feature of appending to the dataset, but you are just writing separate Parquet files per step.
>    
>    * Could you please elaborate on this as well?

Right, I don't have a proper reason here, I guess this is something that can be improved. Do you now if there is an easy way to modify this PR achieve "appending to the dataset"?

>    * If we finally chose the solution with multiple files, aren't they too small? May we generate larger files by using another partitioning criteria?

You know probably better than me on this 😅 Feel free to push on my my branch (subsequent PR is fine as well) to improve this.

qgallouedec:
> some comments but looks solid, the same pattern can be added to RLOO and GOLD too?

@kashif yes the exact same pattern can (should) be added to RLOO. For GOLD I'm less familiar with the implementation, but I guess it's not far

qgallouedec:
I'll proceed with the merging of this PR, @albertvillanova I think we can address 2 and 3 from https://github.com/huggingface/trl/pull/4818#issuecomment-3782396860 in a subsequent PR

PR #4819, opened 2026-01-13
Set packaging as explicit dependency and standardize version comparison

albertvillanova:
Set packaging as explicit dependency.

This PR ensures the `packaging` library is explicitly listed as a dependency. Additionally, it standardizes how version comparisons are performed throughout the codebase by consistently using `packaging.version.Version` instead of `packaging.version` or `packaging.version.parse`. These changes improve code clarity, reliability, and compatibility checks across various modules.

Dependency management:

* Added `packaging>20.0` to the dependencies in `pyproject.toml` to ensure the required version-handling tools are available.

Codebase-wide version comparison standardization:

* Replaced all instances of `packaging.version.parse(...)` and `packaging.version` with direct usage of `Version(...)` from `packaging.version` 

These updates provide a more robust and uniform approach to version checking throughout the project.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -240,7 +240,6 @@ jobs:
           uv pip install accelerate==1.4.0
           uv pip install datasets==3.0.0
           uv pip install transformers==4.56.2
-          uv pip install packaging==20.0
 
       - name: Test with pytest
         run: |


pyproject.toml
```diff
@@ -30,6 +30,7 @@ requires-python = ">=3.10"
 dependencies = [
     "accelerate>=1.4.0",
     "datasets>=3.0.0",
+    "packaging>20.0",
```
qgallouedec, line 33: We currently rely on transitive dependencies in a few places in `trl`. For instance, we import `packaging` directly but don’t declare it, because it’s installed as a dependency of `transformers`. That’s a bit brittle, not good practice and not ideal long-term.

There are similar cases with `pyarrow`, `numpy`, `torch`, etc. I suggest we open a dedicated PR to explicitly list any packages we import at runtime. For `packaging`, we can either keep the dependency addition in this PR or move it to that follow-up. Your call.

qgallouedec:
What do you think of adding

```
uv pip install packaging==20.0
```

in `tests_min_versions`?

I'm not sure it's really needed, as the priority is more to tests against our core deps, like transformers, datasets and accelerate, but suggesting anyway

pyproject.toml
albertvillanova: @qgallouedec, I agree and I am happy to open a dedicated PR with all non-explicitly declared dependencies.

In relation to this PR, I would suggest to keep it because it does not just adds the dependency, but also standardizes how version comparisons are performed (and moves all the import to the module level).

In relation with testing the packaging min version, please note that I already read the docs about the features we need, (like `Version("5.0.0.dev0")`), and PEP 440 parsing for .dev0 is fully stable by packaging 20.0 (Jan 2020).
- On the other hand, I could manually test that nothing is broken with 20.0 before merging this PR
- I agree with you that tests_min_versions is more to test against our core deps

albertvillanova:
CI tests passed with minimum versions and packaging==20.0: https://github.com/huggingface/trl/actions/runs/20966136627/job/60256971197?pr=4819
```python
 Installed 2 packages in 3ms
 - packaging==25.0
 + packaging==20.0
 + pyparsing==3.3.1
```
```python
= 782 passed, 60 skipped, 8 xfailed, 19826 warnings, 1 rerun in 794.69s (0:13:14) =
```
I am reverting the last commit.

PR #4824, opened 2026-01-13
Add support for training with multiple OpenEnv environments

lewtun:
# What does this PR do?

Demo logs: https://lewtun-multienv-grpo.hf.space

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/examples/scripts/openenv/browsergym.py
+++ b/examples/scripts/openenv/browsergym.py
@@ -84,6 +84,7 @@ from __future__ import annotations
 import argparse
 from datetime import datetime
 from pathlib import Path
+from typing import Any
 
 import numpy as np
 from datasets import Dataset
@@ -521,16 +522,17 @@ def main() -> None:
     grpo_config.run_name = args.run_name or f"run-{timestamp}"
     grpo_config.project = args.project or f"group-{sanitize_name(args.model_id)}"
 
-    def rollout_func(prompts: list[str], trainer: GRPOTrainer) -> dict[str, list]:
+    def rollout_func(inputs: list[dict[str, Any]], trainer: GRPOTrainer) -> dict[str, list]:
         episode_prompt_ids: list[list[int]] = []
         episode_completion_ids: list[list[int]] = []
         episode_logprobs: list[list[float]] = []
         completion_rewards: list[float] = []
         episode_images: list[list[Image.Image]] = []
 
-        print(f"\n[DEBUG] rollout_func called with {len(prompts)} prompts")
+        print(f"\n[DEBUG] rollout_func called with {len(inputs)} inputs")
 
-        for i, prompt_text in enumerate(prompts):
+        for i, inp in enumerate(inputs):
+            prompt_text = inp["prompt"]
             print(f"[DEBUG] Processing prompt {i + 1}/{len(prompts)}")
             episode = rollout_once(
                 trainer=trainer,
--- a/examples/scripts/openenv/browsergym.py
+++ b/examples/scripts/openenv/browsergym.py
@@ -87,8 +87,8 @@ from pathlib import Path
 from typing import Any
 
 import numpy as np
-from datasets import Dataset
 from browsergym_env import BrowserGymAction, BrowserGymEnv
+from datasets import Dataset
 from PIL import Image
 from transformers import AutoTokenizer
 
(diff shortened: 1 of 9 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -126,10 +126,11 @@
...
 
-# What we call a rollout function is a callable that takes prompts (list) and the trainer instance as parameters and
-# returns a dict of generation results. Those results must include "prompt_ids", "completion_ids", and "logprobs"
-# fields. Any extra fields (per-completion) are forwarded to the reward functions.
-RolloutFunc = Callable[[list[str], "GRPOTrainer"], dict[str, Any]]
+# What we call a rollout function is a callable that takes the full inputs (list of dicts, each containing at minimum
+# a "prompt" key plus any extra dataset columns) and the trainer instance as parameters and returns a dict of
+# generation results. Those results must include "prompt_ids", "completion_ids", and "logprobs" fields. Any extra
+# fields (per-completion) are forwarded to the reward functions.
+RolloutFunc = Callable[[list[dict[str, Any]], "GRPOTrainer"], dict[str, Any]]
```
lewtun, line 133: This change is needed if we want to route prompts to the appropriate environment via a dataset column (e.g. `"env"`)

trl/trainer/grpo_trainer.py
```diff
@@ -1850,7 +1855,7 @@ def _generate_and_score_completions(
             num_items_in_batch,
             sampling_per_token_logps_list,
             extra_fields,
-        ) = self._generate(prompts)
+        ) = self._generate(prompts, inputs)
```
qgallouedec, line 1859: It's not obvious why we provide both prompts and inputs (prompts are processed from the inputs). I'd add a small comment to help understand.

```suggestion
        # inputs are forwarded to rollout_func if provided; standard generation uses prompts only
        ) = self._generate(prompts, inputs)
```

trl/trainer/grpo_trainer.py
```diff
@@ -1725,14 +1730,14 @@ def _tool_call_loop(self, prompts, prompt_ids, completion_ids, completions, logp
             iteration_num += 1
         return tool_mask, completions, completion_ids, logprobs, tool_call_count, tool_failure_count
 
-    def _generate(self, prompts: list):
+    def _generate(self, prompts: list, inputs: list[dict[str, Any]] | None = None):
```
qgallouedec: nit: it's private, so I think we don't need the default `None` value.

```suggestion
    def _generate(self, prompts: list, inputs: list[dict[str, Any]]):
```

trl/trainer/grpo_trainer.py
```diff
@@ -1287,7 +1289,7 @@ async def _run_async_funcs():
         rewards_per_func = gather(rewards_per_func)
         return rewards_per_func
 
-    def _generate_single_turn(self, prompts: list):
+    def _generate_single_turn(self, prompts: list, inputs: list[dict[str, Any]] | None = None):
```
```suggestion
    def _generate_single_turn(self, prompts: list, inputs: list[dict[str, Any]]):
```

qgallouedec (approved):
I'm fine with the changes in GRPO. I think it more flexible this way.
I'd wait for having feedback from @burtenshaw and @sergiopaniego before merging though, as it's an important change.

trl/trainer/grpo_trainer.py
burtenshaw: I assume that the column could also be an object like 'env state'. I reckon they'll be other use cases for that.

burtenshaw (approved):
Looks good. I would like to return the dx around `rollout_func` shown in examples examples because they're now dealing with `num_gens` too. But we can do that as part of that refactor. 

@lewtun nice work and thanks for updating all examples.

trl/trainer/grpo_trainer.py
lewtun: Indeed, the current proposal is agnostic to the choice of column and content the user wants to propagate to `rollout_func`

examples/scripts/openenv/browsergym.py
```diff
@@ -521,17 +527,18 @@ def main() -> None:
...
-    def rollout_func(prompts: list[str], trainer: GRPOTrainer) -> dict[str, list]:
+    def rollout_func(inputs: list[dict[str, Any]], trainer: GRPOTrainer) -> dict[str, list]:
         episode_prompt_ids: list[list[int]] = []
         episode_completion_ids: list[list[int]] = []
         episode_logprobs: list[list[float]] = []
         completion_rewards: list[float] = []
         episode_images: list[list[Image.Image]] = []
 
-        print(f"\n[DEBUG] rollout_func called with {len(prompts)} prompts")
+        print(f"\n[DEBUG] rollout_func called with {len(inputs)} inputs")
```
lewtun, line 537: Note: should fix to get actual number of prompts

qgallouedec:
Closing this one. The OpenEnv integration has moved on quite a bit since this was opened, and the multi-env story will likely need a fresh design on top of the current state. Happy to revisit if/when we pick this up again.

PR #4839, opened 2026-01-16
Refactor KTO [4/N]: Remove unused padding_value

albertvillanova:
Refactor KTO [4/N]: Remove unused padding_value.

This PR removes the unused `padding_value` parameter from KTO.

Part of:
- #4786

Note:
- Dead code eliminated: The variable was never referenced anywhere else in the codebase
- No functional change: Since it was never used, removing it has zero impact on behavior
- Standard practice: Padding is controlled via tokenizer.pad_token_id
- Simpler API: One less unused parameter cluttering the config

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,9 @@ name: Tests
 
 on:
   push:
-    branches: [ main ]
+    branches:
+      - main
+      - ci-*
   pull_request:
     paths:
       # Run only when relevant files are modified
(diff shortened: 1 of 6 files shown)


qgallouedec:
It's used here:

https://github.com/huggingface/trl/blob/1fe4f3ec573541533ffc38d392a15f107986a204/trl/experimental/kto/kto_trainer.py#L494

albertvillanova:
@qgallouedec, I removed that line as well.

What I meant was that `self.padding_value` is not used anywhere, so it can be safely removed.

qgallouedec:
my bad 😅

for the record, we can also remove this in BCOTrainer

qgallouedec (approved):

albertvillanova:
Yes, first I wanted your validation on the removal.

PR #4844, opened 2026-01-16
Fix _patch_transformers_hybrid_cache for peft

albertvillanova:
Fix `_patch_transformers_hybrid_cache` for peft.

This PR fixes the transformers patch required by peft. In particular, peft import `HybridCache` via:
```python
from transformers import HybridCache
```
whereas liger-kernel imports it via:
```python
from transformers.cache_utils import HybridCache
```
As a result, patching `_LazyModule` is also required to ensure the fix is applied correctly in both cases.

Follow-up to:
- #4820

### DIFF
--- a/trl/_compat.py
+++ b/trl/_compat.py
@@ -180,9 +180,35 @@ def _patch_transformers_hybrid_cache() -> None:
         _is_package_version_below("liger_kernel", "0.6.5") or _is_package_version_below("peft", "0.18.0")
     ):
         try:
-            import transformers.cache_utils as cache_utils
+            import transformers.cache_utils
+            from transformers.utils.import_utils import _LazyModule
 
-            cache_utils.HybridCache = cache_utils.Cache
+            Cache = transformers.cache_utils.Cache
+
+            # Patch for liger_kernel: Add HybridCache as an alias for Cache in the cache_utils module
+            transformers.cache_utils.HybridCache = Cache
+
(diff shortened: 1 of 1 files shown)


trl/_compat.py
```diff
@@ -180,9 +180,35 @@ def _patch_transformers_hybrid_cache() -> None:
...
+
+                    if hasattr(self, "_class_to_module"):
+                        self._class_to_module["HybridCache"] = "cache_utils"
+
+                    if hasattr(self, "__all__") and "HybridCache" not in self.__all__:
+                        self.__all__.append("HybridCache")
+
+                    self.HybridCache = Cache
+
+            _LazyModule.__init__ = _patched_lazy_module_init
```
qgallouedec, line 210: I'm fine with this solution, but any reason not to simply use?

```python
transformers.HybridCache = Cache
```

is it to leverage lazy import?

albertvillanova: @qgallouedec, that was my first approach, but it did not work: after the patch, when peft imports transformers, the transformers module was being replaced with a fresh `_LazyModule` instance that didn't have the `HybridCache` patch applied. This happened because:                                                                                                                                                                                                                                                                                                                                             
  1. Transformers' `__init__.py` replaces itself with a `_LazyModule` at initialization                                                                                                                        
  2. When peft imported transformers, this caused the module to be re-initialized                                                                                                                          
  3. The new `_LazyModule` instance didn't have our `HybridCache` patch

PR #4845, opened 2026-01-16
Add retry strategy to vLLM Client for increased robustness

apalmas-saifh:
### Motivation

  vLLM connection drops can cause training errors and interrupt long-running jobs. This PR allows users to configure retry behavior to make training more robust against transient network issues.

 ### Summary of Changes

  Added configurable HTTP retry parameters for the VLLMClient that can be configured via trainer configs (GRPOConfig, RLOOConfig, OnlineDPOConfig, GOLDConfig)

 #### New Configuration Parameters

  | Parameter                   | Default         | Description                                     |
  |-----------------------------|-----------------|-------------------------------------------------|
  | vllm_retry_total            | 5               | Total number of retries for HTTP requests       |
  | vllm_retry_backoff_factor   | 2.0             | Backoff factor for retry delays                 |
  | vllm_retry_status           | 3               | Number of retries for status codes in forcelist |
  | vllm_retry_status_forcelist | [500, 502, 503] | HTTP status codes that trigger a retry          |
  | vllm_retry_allowed_methods  | ["POST", "GET"] | HTTP methods allowed to be retried              |

Fixes # (issue)

      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec

### DIFF
--- a/trl/extras/vllm_client.py
+++ b/trl/extras/vllm_client.py
@@ -142,7 +142,7 @@ class VLLMClient:
             status=3,  # retry a limited number of times when we receive certain HTTP error responses from the server
             status_forcelist=[500, 502, 503],  # only retry on server-side errors that are usually temporary
             backoff_factor=2,  # exponential backoff between retries (2s, 4s, 8s, ...)
-            allowed_methods=["POST", "GET"],  # allow retries for POST as well, because safe to retry in this context
+            allowed_methods=["POST", "GET"],  # allow POST as well, even though we're not sure it's safe here
         )
 
         adapter = HTTPAdapter(max_retries=retry_strategy)


qgallouedec:
Thank you for your contribution. Instead of configurability, can we have hardcoded values instead? 

That would allow us to just add these lines, and no new parameters:
```python
        # Configure retry strategy
        retry_strategy = Retry(
            total=5,
            connect=5,
            read=5,
            status=3,
            status_forcelist=[500, 502, 503],
            backoff_factor=2.0,
            allowed_methods=["POST", "GET"],
        )
        adapter = HTTPAdapter(max_retries=retry_strategy)
        self.session.mount("http://", adapter)
        self.session.mount("https://", adapter)
```

apalmas-saifh:
> Thank you for your contribution. Instead of configurability, can we have hardcoded values instead?
> 
> That would allow us to just add these lines, and no new parameters:
> 
> ```python
>         # Configure retry strategy
>         retry_strategy = Retry(
>             total=5,
>             connect=5,
>             read=5,
>             status=3,
>             status_forcelist=[500, 502, 503],
>             backoff_factor=2.0,
>             allowed_methods=["POST", "GET"],
>         )
>         adapter = HTTPAdapter(max_retries=retry_strategy)
>         self.session.mount("http://", adapter)
>         self.session.mount("https://", adapter)
> ```

I used/tested it myself with hardcoded values, but I felt you wanted to give this flexibility to the users, plus the default values are what I used/would put as hardcoded values.

A middle ground would be ENV vars with defaults if not found.

Your call @qgallouedec, let me know what you prefer and I will modify accordingly

qgallouedec:
Thanks for testing this 👍
I'd try to keep things simple here. Let's go with hardcoded values for now and avoid adding new configuration parameters. If we later find a real need to tweak retries in different environments, we can revisit configurability (or env vars) then.
I didn't try myself, so I rely on you to suggest the good hardcoded values.
Feel free to update the PR accordingly.

apalmas-saifh:
@qgallouedec Done. Hardcoded values that I have been using in my trainings. 

Let me know if you have any other comment.

trl/extras/vllm_client.py
```diff
@@ -129,6 +131,21 @@ def __init__(
...
+            read=5,
+            status=3,
+            status_forcelist=[500, 502, 503],
+            backoff_factor=2,
+            allowed_methods=["POST", "GET"],
+        )
+
+        adapter = HTTPAdapter(max_retries=retry_strategy)
+        self.session.mount("http://", adapter)
+        self.session.mount("https://", adapter)
```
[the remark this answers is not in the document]
apalmas-saifh: @qgallouedec this is what we discussed already, I guess hardcoding is still your go-to choice, right?

trl/extras/vllm_client.py
```diff
@@ -129,6 +131,21 @@ def __init__(
...
         self.session = requests.Session()
 
+        # Configure retry strategy
+        retry_strategy = Retry(
+            total=5,
+            connect=5,
+            read=5,
+            status=3,
+            status_forcelist=[500, 502, 503],
+            backoff_factor=2,
```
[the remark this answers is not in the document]
apalmas-saifh: @qgallouedec I would keep the backoff factor as it is instead of having it linear, in theory this is not something that should happen often.

trl/extras/vllm_client.py
```diff
@@ -23,8 +23,10 @@
 
 import torch
 import torch.distributed.distributed_c10d as c10d
+from requests.adapters import HTTPAdapter
 from torch import nn
 from transformers import is_torch_xpu_available
+from urllib3.util.retry import Retry
```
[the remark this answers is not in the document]
apalmas-saifh: @qgallouedec On my side this is automated by state of the art ruff / pyright code autoformatting, up to you if you want to change it

trl/extras/vllm_client.py
```diff
@@ -129,6 +131,24 @@ def __init__(
...
+        # failures (network hiccups, temporary 5xx errors, overloaded servers). Without this, such failures could cause
+        # an otherwise healthy training run to fail.
+        retry_strategy = Retry(
+            total=5,  # global cap on the total number of retries across all failure types
+            connect=5,  # retry connection-level failures (DNS issues, refused connections, etc)
+            read=5,  # retry failures while reading the response after the connection was successfully established
+            status=3,  # retry a limited number of times when we receive certain HTTP error responses from the server
+            status_forcelist=[500, 502, 503],  # only retry on server-side errors that are usually temporary
+            backoff_factor=2,  # exponential backoff between retries (2s, 4s, 8s, ...)
+            allowed_methods=["POST", "GET"],  # allow retries for POST as well, because safe to retry in this context
```
qgallouedec: @apalmas-saifh just to double check with you, is `"POST"` really safe to retry here? Why did you need to add it?

apalmas-saifh: @qgallouedec While I did not dig into the extremely low level details of the client-server implementation, I wanted to take the safe side adding POST too assuming that, when you send the updated weights to vLLM you need to use POST method? But I am not sure if this is actually the case. 

This is the configuration that I tested in my experiments.

qgallouedec: ah right, in the case of weight update I think it's safe (though I'm not 100% sure, because we have a `communicator.broadcast` involved.
We also use post for generation, so we may send the same prompts twice.
I'm not sure if it's safe or not in both these cases, but I feel like it will always be better than no retry. Worst case, it will retry and fail because it's not safe. So I agree keeping things like this. I'd just update the comment

```suggestion
            allowed_methods=["POST", "GET"],  # allow POST as well, even though we're not sure it's safe here
```

apalmas-saifh: Perfect, I am aligned!

PR #4855, opened 2026-01-17
feat(rewards): add conditioned_reward based on GDPO

### LINKED ISSUE #4853: Implement hierarchical reward conditioning based on GDPO paper
### Feature request

I propose adding a Conditioned Reward utility in trl/rewards/utils.py. This utility is a high-level wrapper designed to manage hierarchies of objectives in Reinforcement Learning (RL), specifically for trainers like GRPOTrainer.

It implements the Conditioned Reward strategy detailed in the paper ["GDPO: Group reward-Decoupled Normalization Policy Optimization for Multi-reward RL Optimization"](https://huggingface.co/papers/2601.05242).

The utility allows a secondary reward $r_k$​, typically a "format" or "length" constraint) to be strictly conditioned on the success of a primary reward $r_l$, typically "accuracy" or "correctness"). Following Equation 8 of the paper:

```math
\tilde{r}_{k} = \begin{cases} r_{k}, & \text{if } r_{l} \ge t \\ 0, & \text{otherwise} \end{cases}
```

### Motivation

In multi-reward settings, Reinforcement Learning often suffers from "Reward Hacking". This occurs when one objective is significantly easier to optimize than others (e.g., satisfying a format constraint vs. solving a complex math problem). As noted in the Section 3.2 in the GDPO paper:

> If one objective is much easier than the others, the model often focuses on maximizing the reward for that objective regardless of the assigned weights. [...] Therefore, some recent works address such reward hacking by conditioning easier rewards on more difficult rewards

To mitigate this, the paper proposes a hierarchical reward design where a secondary reward (e.g., format) is only granted if a primary reward (e.g., accuracy) meets a certain threshold. By enforcing this hierarchy, we prevent "easy" rewards from dominating the gradient, ensuring the model always prioritizes human-favored outcomes and preventing the primary correctness signal from dropping sharply during training.

This request is related to current implementation of GDPO in TRL, such as Issue #4841 and PR #4785, which focus on the decoupled normalization aspect of the same paper. Adding the ConditionedReward utility would provide a systematic way to implement the reward function design mentioned in Section 3.2 of the paper.

### Your contribution

I propose adding a conditioned_reward utility in trl/rewards/utils.py. This high-level wrapper allows defining a dependency between two reward functions:

```math
\tilde{r}_{secondary} = \begin{cases} r_{secondary}, & \text{if } r_{primary} \ge \text{threshold} \\ 0, & \text{otherwise} \end{cases}
```
​	
 
Features:

- Enforces a strict dependency between rewards to prevent the model from ignoring the primary task.
- Fully compatible with both synchronous and asynchronous reward functions.
- I also propose enhancing GRPOTrainer to support a context dictionary. This allows conditioned_reward to retrieve previously calculated results from the same batch, avoiding redundant and expensive re-executions of the primary reward function.

jlcanta:
# What does this PR do?
Fixes #4853 

This PR introduces the conditioned_reward utility, a hierarchical reward wrapper designed to mitigate "Reward Hacking" in multi-objective Reinforcement Learning. This implementation follows the strategy proposed in the [GDPO paper](https://huggingface.co/papers/2601.05242), specifically Equation 8.

## Changes

- trl/rewards/utils.py: Created the conditioned_reward higher-order function.
- trl/trainer/grpo_trainer.py: Modified _calculate_rewards to initialize and populate a context dictionary for shared batch computation. This allows conditioned_reward to retrieve previously calculated results from the same batch, avoiding redundant and expensive re-executions of the primary reward function.
- trl/rewards/__init__.py: Exported the new utility.
- tests/test_rewards.py: Added comprehensive unit tests covering sync/async logic, tensor vectorization, and context caching.
- docs/source/paper_index.md: Added a section on GDPO reward conditioning with usage examples.

### DIFF
--- a/docs/source/paper_index.md
+++ b/docs/source/paper_index.md
@@ -516,6 +516,46 @@ Note that this method only has an effect when training involve more than one rew
 
 The authors provide a easy-to-use, slurm-free training example that enable the community to quickly validate GDPO’s effectiveness over GRPO, see [Experiment-"Aha" moment](https://github.com/NVlabs/GDPO/tree/main/trl-GDPO).
 
+#### Effective incorporation of priority variation
+
+The authors also introduce a mechanism to handle priority variation among different rewards, specifically addressing "reward hacking" where models might optimize for easier secondary objectives at the expense of primary ones. Section 3.2 of the [GDPO paper](https://huggingface.co/papers/2601.05242) proposes a conditional reward structure:
+
+$$
+\tilde{r}^{(i,j)}_k = \begin{cases}
+r^{(i,j)}_k & \text{if } r^{(i,j)}_l \ge \delta \\
+0 & \text{otherwise}
+\end{cases}
+$$
+
+where $\tilde{r}^{(i,j)}_k$ is the conditioned secondary reward, $r^{(i,j)}_k$ is the original secondary reward, $r^{(i,j)}_l$ is the primary reward, and $\delta$ is the threshold.
+
+To implement this logic in TRL, you can use the `conditioned_reward` utility:
+
+```python
+from trl.rewards import conditioned_reward
+
+def primary_reward(prompts, completions, **kwargs):
+    # check for correctness, reasoning format, etc.
+    ...
+
+def secondary_reward(prompts, completions, **kwargs):
+    # compute length penalty, etc.
+    ...
+
+# Only apply secondary reward if primary reward >= 1.0
+conditioned_secondary = conditioned_reward(primary_reward, secondary_reward, condition=1.0)
+
+trainer = GRPOTrainer(
+    model=...,
+    args=...,
+    train_dataset=...,
+    # Pass the primary reward and the conditioned secondary reward
+    reward_funcs=[primary_reward, conditioned_secondary],
+)
+```
+
+Note that `GRPOTrainer` computes rewards in order. If `primary_reward` is passed before `conditioned_secondary`, its result is cached in a context dictionary. `conditioned_reward` automatically detects and uses this cached result to avoid redundant computation of the primary reward.
+
 ## Direct Policy Optimization
 
 - Papers relating to the [`DPOTrainer`]
--- a/tests/test_rewards.py
+++ b/tests/test_rewards.py
@@ -12,7 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from trl.rewards import accuracy_reward, get_soft_overlong_punishment, reasoning_accuracy_reward, think_format_reward
+import asyncio
+
+from trl.rewards import (
+    conditioned_reward,
+    accuracy_reward,
+    get_soft_overlong_punishment,
+    reasoning_accuracy_reward,
+    think_format_reward,
+)
 
 from .testing_utils import TrlTestCase, require_math_latex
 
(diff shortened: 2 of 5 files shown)


qgallouedec:
At first glance, this seems a bit over-engineered. Why not just

```python
def is_correct(completion, ground_truth):
    return ground_truth in completion
    

def correctness_reward(completions, ground_truths, **kwargs):
    return [float(is_correct(c, g)) for c, g in zip(completions, ground_truths)]


def format_reward(completions, ground_truths, **kwargs):
    are_correct = [is_correct(c, g) for c, g in zip(completions, ground_truths)]
    format_rewards = ["<think>" in c for c in completions]
    rewards = [float(corr and fmt) for corr, fmt in zip(are_correct, format_rewards)]
    return rewards

if __name__ == "__main__":
    completions = [
        "<think>I think the answer is 42.</think><answer>42</answer>",  # Correct and well-formatted
        "The answer is 42.",  # Correct but not well-formatted
        "<think>I think the answer is 24.</think><answer>24</answer>",  # Incorrect but well-formatted
    ]
    ground_truths = [
        "42",
        "42",
        "42",
    ]
    print("Format Rewards:", format_reward(completions, ground_truths))  # Output: [1.0, 0.0, 0.0]
    print("Correctness Rewards:", correctness_reward(completions, ground_truths))  # Output: [1.0, 1.0, 0.0]
```

jlcanta:
Thanks for the review! While I agree your approach is cleaner for simple string-matching rewards, my main concern is computational efficiency for more complex rewards. In many RL pipelines, the primary reward might be an expensive call (e.g., an LLM judge or a heavy reward model). Using the context caching mechanism (as implemented in this PR) prevents evaluating the primary reward twice (once as a standalone reward and once inside the secondary reward).

Additionally, conditioned_reward handles torch.Tensor outputs and device placement automatically, which is common when using trained Reward Models. Implementing this logic via list comprehensions manually for every experiment would be error-prone for users working with tensors or async functions.

I really appreciate your response, keeping things simple and readable is always a priority and I totally see the appeal of the one-liner approach for quick scripts. However, my goal with this utility is to provide a robus tool for the TRL ecosystem that follows the GDPO paper precisely while staying efficient at scale. That said, if you have any suggestions on how to implement this idea in a different way while maintaining these efficiency goal, I’m definitely open to hearing them! :)"

qgallouedec:
> Thanks for the review! While I agree your approach is cleaner for simple string-matching rewards, my main concern is computational efficiency for more complex rewards. In many RL pipelines, the primary reward might be an expensive call (e.g., an LLM judge or a heavy reward model). Using the context caching mechanism (as implemented in this PR) prevents evaluating the primary reward twice (once as a standalone reward and once inside the secondary reward).

I feel like this kind of optimization pertains more to the inference side that trainer side. For example, vLLM has a Automatic Prefix Caching https://docs.vllm.ai/en/stable/design/prefix_caching/, and more generally, you can leverage `functools.lru_cache` if you ant to cache the results of a function.


> Additionally, conditioned_reward handles torch.Tensor outputs and device placement automatically, which is common when using trained Reward Models.

When using reward models (`nn.Module`), you shouldn't wrap it into a function and pass this function to `GRPOTrainer`. Instead, you should pass the reward model (or reward model id) directly to the trainer. It will handle the proper preparation of model and inputs based on the chosen distributed strategy. It's not widely advertised since it's not the primary use case but we have a few example in the tests, see https://github.com/huggingface/trl/blob/5d1c2c1d438d6f3d74671016c932fb576c9e3e8c/tests/test_grpo_trainer.py#L154

jlcanta:
Thanks for the feedback! I understand your point of view. Thanks again for your time!

PR #4857, opened 2026-01-18
fix(DeepSeek OPSM): passing correct (vLLM) logprobs

casinca:
# What does this PR do?

Hello,  
This PR fixes a mistake I made in the DeepSeek Off-Policy Sequence Masking (OPSM) implementation https://github.com/huggingface/trl/pull/4689

&nbsp;

There are 2 sources of "off-policyness" that OPSM aims to improve:
- drift with repetitive num grad updates
- drift with hybrid training

Currently only the 1st point was covered and is relevant whether we use vLLM or not.

But the 2nd point, when using vLLM, was not covered because I was passing `old_per_token_logps` thinking they were the vLLM logprobs. Those are in fact the local pytorch model logprobs and vLLM logprobs are `sampling_per_token_logps`.

So now with the fix, when OPSM is enabled, we prioritize vLLM logprobs if available. If `None`, `old_per_token_logps` are used, so that users still benefit from OPSM for the 1st source of off-policyness (even if they don't use vLLM).

More details in the discussion with @LeonEricsson  https://github.com/huggingface/trl/pull/4689#issuecomment-3764246518

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec since he was on the previous PR

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -2325,7 +2325,7 @@ class GRPOTrainer(BaseTrainer):
             # OPSM should use inference-time logprobs to detect both sources of off-policyness:
             # 1. Drift from gradient updates (always present)
             # 2. Drift from training-inference mismatch (when using vLLM)
-            # If using vLLM, prioritize sampling_per_token_logps as the old policy logprobs; otherwise use old_per_token_logps
+            # When using vLLM, prioritize sampling_per_token_logps, otherwise use old_per_token_logps
             sampling_per_token_logps = inputs.get("sampling_per_token_logps", old_per_token_logps)
 
             off_policy_mask = self.get_off_policy_mask(
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -997,66 +997,6 @@ class TestGRPOTrainer(TrlTestCase):
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
 
-    def test_compute_loss_and_off_policy_mask_with_correct_sampling_per_token_logps(self):
-        """
-        Test that _compute_loss correctly selects sampling_per_token_logps for OPSM if available or fallback to
-        old_per_token_logps otherwise.
-        """
-        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
-
-        training_args = GRPOConfig(
-            output_dir=self.tmp_dir,
-            off_policy_mask_threshold=0.5,
-            report_to="none",
-            bf16=False,
-            fp16=False,
-        )
-
-        trainer = GRPOTrainer(
-            model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
-            reward_funcs="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
-            args=training_args,
-            train_dataset=dataset,
-        )
-
-        # Mock inputs matching the structure for GRPOTrainer
-        batch_size = 2
-        seq_len = 4
-        device = trainer.accelerator.device
-        inputs = {
-            "prompt_ids": torch.zeros((batch_size, seq_len), dtype=torch.long, device=device),
-            "prompt_mask": torch.ones((batch_size, seq_len), dtype=torch.long, device=device),
-            "completion_ids": torch.zeros((batch_size, seq_len), dtype=torch.long, device=device),
-            "completion_mask": torch.ones((batch_size, seq_len), dtype=torch.long, device=device),
-            "advantages": torch.zeros((batch_size,), device=device),
-            "old_per_token_logps": torch.ones((batch_size, seq_len), device=device),
-            "sampling_per_token_logps": torch.ones((batch_size, seq_len), device=device) * 2.0,
-            "ref_per_token_logps": torch.zeros((batch_size, seq_len), device=device),
-            "num_items_in_batch": batch_size,
-        }
-
-        # Mock _get_per_token_logps_and_entropies to avoid running the model
-        per_token_logps = torch.zeros((batch_size, seq_len), device=device)
-        entropies = torch.zeros((batch_size, seq_len), device=device)
-        trainer._get_per_token_logps_and_entropies = lambda *args, **kwargs: (per_token_logps, entropies)
-
-        # Case 1: sampling_per_token_logps is present (vLLM scenario)
-        with patch.object(GRPOTrainer, "get_off_policy_mask", wraps=GRPOTrainer.get_off_policy_mask) as mock_get_mask:
-            trainer._compute_loss(trainer.model, inputs)
-
-            # Verify that get_off_policy_mask received the sampling logps (2.0)
-            _, kwargs = mock_get_mask.call_args
-            torch.testing.assert_close(kwargs["sampling_per_token_logps"], inputs["sampling_per_token_logps"])
-
-        # Case 2: sampling_per_token_logps is absent (local PyTorch scenario)
-        inputs.pop("sampling_per_token_logps")
-        with patch.object(GRPOTrainer, "get_off_policy_mask", wraps=GRPOTrainer.get_off_policy_mask) as mock_get_mask:
-            trainer._compute_loss(trainer.model, inputs)
-
-            # Verify that get_off_policy_mask fell back to old_per_token_logps (1.0)
-            _, kwargs = mock_get_mask.call_args
-            torch.testing.assert_close(kwargs["sampling_per_token_logps"], inputs["old_per_token_logps"])
-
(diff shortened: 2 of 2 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -2228,10 +2230,10 @@ def get_off_policy_mask(
         Computes the Off-Policy Sequence Mask from DeepSeek-V3.2 paper. Returns a (B, 1) tensor where 1.0 indicates
         "Keep" and 0.0 indicates "Drop".
         """
-        # K1 estimator: log(pi_old) - log(pi_theta)
-        k1_divergence = old_per_token_logps - per_token_logps.detach()
+        # forward KL div: log(pi_old) - log(pi_theta)
+        kl_div = old_per_token_logps - per_token_logps.detach()
```
albertvillanova: Nice: Naming `kl_div` instead of `k1_divergence` aligns with the paper's terminology

trl/trainer/grpo_trainer.py
```diff
@@ -2282,10 +2284,14 @@ def _compute_loss(self, model, inputs):
         old_per_token_logps = per_token_logps.detach() if old_per_token_logps is None else old_per_token_logps
 
         if self.off_policy_mask_threshold is not None:
+            # If using vLLM, we use sampling_per_token_logps as the old policy logprobs.
```
albertvillanova: Optionally, the comment could be slightly expanded to mention both drift sources explicitly:

```suggestion
            # OPSM should use inference-time logprobs to detect both sources of off-policyness:                                                                                                                      
            # 1. Drift from gradient updates (always present)                                                                                                                                                        
            # 2. Drift from training-inference mismatch (when using vLLM)                                                                                                                                            
            # If using vLLM, prioritize sampling_per_token_logps as the old policy logprobs; otherwise use old_per_token_logps
```

trl/trainer/grpo_trainer.py
```diff
@@ -2282,10 +2284,14 @@ def _compute_loss(self, model, inputs):
...
+            sampling_per_token_logps = inputs.get("sampling_per_token_logps")
+            old_logps = sampling_per_token_logps if sampling_per_token_logps is not None else old_per_token_logps
```
albertvillanova: Personally, I find the `old_logps` naming a bit semantically ambiguous (`old` has different meanings in this context), but I don't have a strong opinion or a better alternative: maybe `sampling_time_logps` or `pi_old_logps`? Feel free to ignore these suggestions.

albertvillanova (approved):
Thanks for catching this issue, the fix and the constructive discussion, @casinca and @LeonEricsson. 🤗 

This PR correctly fixes the OPSM implementation to handle both sources of off-policyness as intended by the DeepSeek paper. 

Minor suggestion: do you think we could easily implement a test validating OPSM behavior to prevent future regressions?

trl/trainer/grpo_trainer.py
casinca: It's fine I renamed `old_logps` to `pi_old_logps`

casinca: Thanks for the feedback @albertvillanova , I did the changes.  
Concerning the test, I've added one to test for correct `pi_old_logps` logic (if vLLM or not) since I wasn't sure how to best approach it, I've refined it with Gemini. I hope this aligns with what you had in mind.

albertvillanova: Nice test! I'm just simplifying it a bit and adding some explanatory comments, but it correctly validates the PR's core fix! Thank! :hugs:

albertvillanova: We are having an internal discussion about the inconvenience of having a test that checks an implementation detail.

I'll let you know once we make a decision.

LeonEricsson: I prefer `old_logps` over `pi_old_logps`. The `pi*` prefix is not an established convention in the trainer. Perhaps it should be, but if so, I would prefer this to be introduced consistently in a separate PR.. That said, I agree that `old_logps` is not entirely satisfactory either.

Part of the difficulty is that the literature itself is inconsistent. In this case, DS uses $\pi_{old}$ to denote inference log probabilities, whereas we, and what I would consider the more common convention, use $\pi_{old}$ to denote the proximal policy inside the PPO clipping term. TRL should aim to implement methods faithful to the original papers, but also maintain internal consistency and readability. Often, this things end up conflicting.

The only sustainable way I see to navigate this is to commit to a single, consistent, and well-documented notation internally, with documentation that makes translation to diverse notation simple. 

From that perspective, and assuming this framing makes sense, I think it is worth reconsidering the use of `old_per_token_logps` in the context of `get_off_policy_mask`, as it conflates definitions that we have already established elsewhere in the trainer. We explicitly use `old_per_token_logps` to refer to the recomputed behavior policy produced by the trainer framework, i.e. the proximal policy used in PPO clipping. In `get_off_policy_mask`, however, `old_per_token_logps` is now being used to denote both the recomputed trainer log probabilities and the true (inference) behaviour policy

albertvillanova: Thanks a lot for sharing your insights, @LeonEricsson. :hugs: 

As I commented above, I don't have a strong opinion or a better alternative; I just commented that `old` has different meanings in this context.

So the discussion is more than welcome!

qgallouedec: Regarding the convention, what do you think of this order of precedence (this is essentially what the trainer uses today)?

* `sampling_per_token_logps`: log-probs produced by the generation engine
* `old_per_token_logps`: log-probs of the trained policy at sampling time
* `per_token_logps`: log-probs of the trained model at loss-computation time

When two values are identical, we intentionally pick the latter (i.e., the one that comes later in the precedence order). And the former isn't used. For example:

* with `transformers` generation, `sampling_per_token_logps == old_per_token_logps`, so we use `old_per_token_logps`
* when we don’t optimize with minibatches, `old_per_token_logps == per_token_logps` (same policy), so we use `per_token_logps`

As a result, we can implement the precedence rule via a simple fallback:

```python
if self.off_policy_mask_threshold is not None:
    sampling_per_token_logps = inputs.get("sampling_per_token_logps", old_per_token_logps)

    off_policy_mask = self.get_off_policy_mask(
        advantages=advantages,
        per_token_logps=per_token_logps,
        sampling_per_token_logps=sampling_per_token_logps,
        mask=mask,
        off_policy_threshold=self.off_policy_mask_threshold,
    )
```

qgallouedec: I think this aligns with your suggestion @LeonEricsson, right?

> From that perspective, and assuming this framing makes sense, I think it is worth reconsidering the use of `old_per_token_logps` in the context of `get_off_policy_mask`, as it conflates definitions that we have already established elsewhere in the trainer. We explicitly use `old_per_token_logps` to refer to the recomputed behavior policy produced by the trainer framework, i.e. the proximal policy used in PPO clipping. In `get_off_policy_mask`, however, `old_per_token_logps` is now being used to denote both the recomputed trainer log probabilities and the true behavior or sampling policy.

LeonEricsson: > Regarding the convention, what do you think of this order of precedence (this is essentially what the trainer uses today)?
> 
> * `sampling_per_token_logps`: log-probs produced by the generation engine
> * `old_per_token_logps`: log-probs of the trained policy at sampling time
> * `per_token_logps`: log-probs of the trained model at loss-computation time

This looks good and I agree this reflects the current state. I like the clarification that `sampling_per_token_logps` represents the true behavior policy, which makes the proposed rewrite feel very natural:

> ```python
> if self.off_policy_mask_threshold is not None:
>     sampling_per_token_logps = inputs.get("sampling_per_token_logps", old_per_token_logps)
> 
>     off_policy_mask = self.get_off_policy_mask(
>         advantages=advantages,
>         per_token_logps=per_token_logps,
>         sampling_per_token_logps=sampling_per_token_logps,
>         mask=mask,
>         off_policy_threshold=self.off_policy_mask_threshold,
>     )
> ```

casinca: Appreciate the discussion.  
I've refactored to use `sampling_per_token_logps` per @qgallouedec 's update. I guess now it's up to the testcase, feel free to remove it if needed.

qgallouedec: I’m going to remove this test.

In TRL it’s always tricky to write truly behavioral tests, so it’s easy to fall back to "does this internal wiring happen?" kind of checks. (This one is basically asserting which log-prob tensor gets passed around.) The problem with these kind of tests is that they tends to become brittle over time, and complexity any refactoring, while not testing the public API. So by default, I'd prefer not have such test, (even if it means we might catch fewer regressions,) than keep a test that mainly locks in internals.

qgallouedec (approved):
thanks @casinca and @LeonEricsson, that's a very good finding

PR #4858, opened 2026-01-18
Compile entropy from logits.

pramodith:
# What does this PR do?

Compiles the `entropy_from_logits` functions, compiling shows significant speed-ups when dim=0 is up-to 16k tokens

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


pramodith:
Results on a H-100
Screenshot 2026-01-18 200454

Used triton's benchmarking function for obtaining the figures. Code snippet:

```
import triton.testing
import warnings

# Ensure CUDA is available
if not torch.cuda.is_available():
    raise RuntimeError("CUDA is not available. Please enable a GPU runtime.")

# Define the benchmark function using triton.testing.perf_report
@triton.testing.perf_report(
    triton.testing.Benchmark(
        x_names=['M'],  # The x-axis will represent the number of rows
        x_vals=[2**i for i in range(5, 15)],  # Vary M from 32 to 16384 (e.g., 32, 64, ..., 16384)
        line_arg='provider',  # Argument to distinguish different lines in the plot
        line_vals=['eager', 'compiled'],  # Values for provider: 'eager' for original, 'compiled' for Triton-compiled
        line_names=['Eager (Python)', 'Compiled (Triton)'],  # Labels for the lines in the plot
        styles=[('blue', '-'), ('green', '-')],  # Line styles for the plot
        ylabel="ms",  # Label for the y-axis (milliseconds)
        plot_name="Entropy from Logits Benchmark",  # Title of the plot and base file name for saving
        args={'N_CLASSES': 65536, 'CHUNK_SIZE': 128} # Fixed arguments passed to bench_entropy (and then to entropy_from_logits)
    )
)
def bench_entropy(M, N_CLASSES, CHUNK_SIZE, provider):
    # Generate random logits tensor on the GPU with float32 dtype
    logits = torch.randn(M, N_CLASSES, device='cuda', dtype=torch.float32)

    # Select the appropriate function to benchmark based on the 'provider' argument
    if provider == 'eager':
        fn = entropy_from_logits
    elif provider == 'compiled':
        fn = compiled_entropy_from_logits
    elif provider == "eager_softmax":
        fn = selective_log_softmax
    elif provider == "compiled_softmax":
        fn = compiled_selective_softmax
    else:
        raise ValueError(f"Unknown provider: {provider}")

    # Use triton.testing.do_bench to measure the execution time of the function
    # The lambda function ensures that `fn` is called with the generated logits and chunk_size
    ms= triton.testing.do_bench(lambda: fn(logits, chunk_size=CHUNK_SIZE))
    return ms

# Run the benchmark and display the results
# print_data=True will print the raw benchmark data to the console
# show_plots=True will display the generated performance plots
bench_entropy.run(print_data=True, show_plots=True)
```

pramodith:
Also benchmarked memory just to make sure that compile doesn't get rid of the chunking as an optimization.

image

pramodith:
We might need to lower tolerance levels for fp16 dtypes to get the failing test cases to pass.

qgallouedec:
Thanks! I'm not super familiar with torch compile, these are my basic questions:

- as far as I understand, the compilation can recompile if it's called with a new shape, which is expected because the sequence length may vary, am I right?
- depending on the sequence length, the flatten tensor may have different shape. So the number of iterations in the loop may change (even when chunk size is fixed). In my understanding it's not an ideal design for `torch.compile`, is it?

qgallouedec:
cc @kashif who is probably more knowledgeable than me

pramodith:
Ahhh good catch! You're right, ran this to check if a re-compile is triggered when we update the sequence length and it does.

```
import torch._dynamo
torch._dynamo.config.verbose = True

torch._dynamo.reset()  # Clear compilation cache

result1 = entropy_from_logits(torch.randn(10, 20, 100))
result2 = entropy_from_logits(torch.randn(10, 20, 100)) 
result3 = entropy_from_logits(torch.randn(15, 25, 100))
print(torch._dynamo.utils.counters)
```

I think the benchmarking script hid the compile time effects on run-time because of the `warmup` setting which is 20 by default, so offers a chance for the kernel to be compiled for a given shape without showing up in the results.

PR #4861, opened 2026-01-19
Use pytest-datadir for accelerate config files

albertvillanova:
Use pytest-datadir for accelerate config files: Replace `CONFIG_PATHS` dict with `get_config_path` fixture

This PR improves the way test configuration files are accessed in distributed, and updates test dependencies and trainer tests. The main focus is on replacing the static `CONFIG_PATHS` dictionary with a `get_config_path` fixture that uses `lazy_shared_datadir`, and specifying a minimum version for `pytest-datadir` to ensure compatibility with lazy datadirs.

**Test improvements:**

* In `tests/distributed/test_distributed.py`, replaced the static `CONFIG_PATHS` dictionary with a `get_config_path` fixture that uses `lazy_shared_datadir`, and updated all distributed test methods to use this fixture for retrieving config file paths.
* Updated trainer tests (`test_train_with_set_chat_template_from_path`) in both `tests/test_reward_trainer.py` and `tests/test_sft_trainer.py` to use the `lazy_shared_datadir` fixture instead of `shared_datadir` for accessing the chat template file.

**Dependency updates:**

* Set a minimum version (`>=1.7.0`) for `pytest-datadir` in both the `test` and `dev` dependency groups in `pyproject.toml`, with a comment noting the requirement for lazy datadirs.


qgallouedec:
lazy in this context means that nothing is copied until you actually need it., ie, the underlying directory is not copied at fixture setup time?

albertvillanova:
Yes, instead of copying all data files for all test cases requiring the `shared_datadir` fixture, we only copy the data file that is required for that specific test case.

I thought it was smarter. But open to revert to `shared_datadir` because the data files are currently very small and copying them all is fast.

qgallouedec:
> I thought it was smarter. But open to revert to `shared_datadir` because the data files are currently very small and copying them all is fast.

I'm fine with both

albertvillanova:
Let's keep lazy for the moment and see if it works fine: I just discovered it!

We can always revert in case there are problems with the custom `LazyDataDir` and we need a proper `Path` instead.

PR #4862, opened 2026-01-19
Refactor KTO coordinated with DPO [d/N]: Remove base_model_attribute_name

albertvillanova:
Refactor KTO coordinated with DPO [d/N]: Remove `base_model_attribute_name`.

This PR removes the `base_model_attribute_name` parameter from KTO and simplifies the base model retrieval logic.

Part of:
- https://github.com/huggingface/trl/issues/4786#issuecomment-3724124945

Coordinated with DPO refactoring, as discussed with @qgallouedec :
- #3906

### Key Changes: 

1. KTOConfig
    - Removed base_model_attribute_name parameter from docstring 
    - Removed base_model_attribute_name field definition (default was "model")

2. KTOTrainer._compute_loss_liger(): EDIT:
    - ~~Fallback to base_model_prefix (standard attribute)~~
    - Just use `get_decoder` (standard method)

#### Fallback Logic:

EDIT:

Just use `get_decoder` method: the fallback logic is handled internally

~~The simplified fallback now uses:~~
~~1. Primary: get_decoder() method (standard on all models)~~
~~2. Fallback: base_model_prefix attribute (standard attribute, defaults to "model")~~
~~3. Final fallback: The model itself~~
                                                                                                                                                                                                           
This covers all practical cases without requiring user configuration.


qgallouedec (commented):
cc @kashif, in case we missed something

trl/experimental/kto/kto_trainer.py
```diff
@@ -1127,7 +1127,8 @@ def _compute_loss_liger(self, model, batch):
         if hasattr(model, "get_decoder") and model.get_decoder() is not None:
             base_model = model.get_decoder()
         else:
-            base_attr = getattr(model, "base_model_prefix", self.args.base_model_attribute_name)
+            # Fallback to base_model_prefix (standard attribute)
+            base_attr = getattr(model, "base_model_prefix", "model")
             base_model = getattr(model, base_attr, model)
```
qgallouedec: I think this can be replaced by just `get_decoder()`

```suggestion
            base_model = model.get_decoder()
```

trl/experimental/kto/kto_trainer.py
```diff
@@ -1140,7 +1141,8 @@ def _compute_loss_liger(self, model, batch):
         if hasattr(self.ref_model, "get_decoder") and self.ref_model.get_decoder() is not None:
             ref_base_model = self.ref_model.get_decoder()
         else:
-            ref_attr = getattr(self.ref_model, "base_model_prefix", self.args.base_model_attribute_name)
+            # Fallback to base_model_prefix (standard attribute)
+            ref_attr = getattr(self.ref_model, "base_model_prefix", "model")
             ref_base_model = getattr(self.ref_model, ref_attr, self.ref_model)
```
qgallouedec: same here

trl/experimental/kto/kto_trainer.py
albertvillanova: Definitely! I missed that currently the `get_decoder` method already handles those conditional checks and fallback logic internally.

Thanks for the bright insight!

PR #4875, opened 2026-01-20
Refactor KTO coordinated with DPO [e/N]: Remove label_pad_token_id

albertvillanova:
Refactor KTO coordinated with DPO [e/N]: Remove label_pad_token_id.

This PR removes label_pad_token_id parameter from KTO.

The label padding value is now always -100 (PyTorch's standard ignore_index). This is already the default everywhere in the ecosystem.

Part of:
- https://github.com/huggingface/trl/issues/4786#issuecomment-3724124945

Coordinated with DPO refactoring, as discussed with @qgallouedec :
- #3906

### Key Changes:

1. KTOConfig
    - Removed label_pad_token_id parameter from docstring
    - Removed label_pad_token_id field definition

2. KTOTrainer
    - Removed self.label_pad_token_id and args.label_pad_token_id
    - Uses default parameter value (-100):
      - In DPODataCollatorWithPadding initialization
      - In LigerFusedLinearKTOLoss initialization
      - In all `get_batch_logps` calls

3. Tests
    - Removed label_pad_token_id from tests

### Follow-up

If this PR is validated, I can apply the same removal to the other trainers.

### DIFF
--- a/trl/experimental/kto/kto_trainer.py
+++ b/trl/experimental/kto/kto_trainer.py
@@ -78,7 +78,6 @@ if TYPE_CHECKING:
 
 logger = logging.get_logger(__name__)
 
-IGNORE_INDEX = -100
 RUNNING_NAME = "running.pt"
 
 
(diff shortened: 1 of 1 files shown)


qgallouedec:
> If this PR is validated, I can apply the same removal to the other trainers.

sounds good!

trl/experimental/kto/kto_trainer.py
```diff
@@ -78,6 +78,7 @@
 
 logger = logging.get_logger(__name__)
 
+IGNORE_INDEX = -100
```
qgallouedec: using `-100` directly is fine as well, but it's up to you

albertvillanova: What do you think of this? Not sure about the naming...

albertvillanova: :wink: We commented at the same time... https://github.com/huggingface/trl/pull/4875#discussion_r2708809835

qgallouedec: in transformers, they don't use constant, and use `-100` directly. If you feel like it lack clarity, maybe a comment is enough?

PR #4894, opened 2026-01-25
Expose generation index to tool callables in GRPOTrainer

### LINKED ISSUE #4866: Expose generation index to tool callables during GRPOTrainer tool execution
### Feature request

When using GRPOTrainer with tools and `num_generations` > 1, the trainer generates multiple completions per prompt, then executes tool calls for each completion via `_tool_call_loop`. However, tool callables have no way to know which completion triggered them.

Looking at [grpo_trainer.py line 1615](https://github.com/huggingface/trl/blob/main/trl/trainer/grpo_trainer.py#L1615):

`result = self._tool_dict[name](**function["arguments"])`

The tool is called with only the function arguments from the model's output. The `idx_with_tool` variable exists in the loop context (representing the completion index), but it's not exposed to the tool callable.

For reward functions that need to attribute tool behavior to specific completions, having this map is really useful.

Without completion attribution, reward functions must either:

Apply all penalties to completion 0 (incorrect - biases GRPO's relative learning)
Distribute penalties evenly across completions (imprecise but statistically fairer)

Our Current Workaround

We currently distribute penalties evenly when attribution isn't available:

```
# If no completion-specific tracking available, distribute evenly
if gen_severity == 0 and total_severity > 0:
    gen_severity = total_severity / num_generations
```

This preserves the total penalty but loses the ability to correctly rank completions that avoided bad behavior vs those that didn't.

## Proposed Solution

Expose the completion index to tool callables. Options:

Option 1: Context parameter (minimal API change)

```
# In _tool_call_loop, around line 1615:
if "_completion_idx" in inspect.signature(self._tool_dict[name]).parameters:
    result = self._tool_dict[name](**function["arguments"], _completion_idx=idx_with_tool)
else:
    result = self._tool_dict[name](**function["arguments"])
```

Option 2: Thread-local context (no signature change required)

```
import contextvars
_current_completion_idx = contextvars.ContextVar("completion_idx", default=-1)

# In _tool_call_loop:
_current_completion_idx.set(idx_with_tool)
result = self._tool_dict[name](**function["arguments"])
```

### Motivation

For reward functions that need to attribute tool behavior to specific completions, having this map is really useful.

Without completion attribution, reward functions must either:

Apply all penalties to completion 0 (incorrect - biases GRPO's relative learning)
Distribute penalties evenly across completions (imprecise but statistically fairer)

### Your contribution

I am more then happy to make the PR, but wanted others to check my understanding first.

lukehinds:
# What does this PR do?

When using `GRPOTrainer`with tools, the trainer generates multiple completions per prompt (`num_generations`), then executes tool calls for each completion via `_tool_call_loop`. However, tool callables have no way to know which completion triggered them. 

The fix exposes the completion index to tool callables by inspecting each tool's signature at initialization and caching whether it accepts a `_completion_idx` parameter. During tool execution in `_tool_call_loop`, if a tool declares this parameter, the trainer passes the current completion index; otherwise it calls the tool normally. This allows tools to know which of the multiple generated completions triggered them, enabling per-completion attribution for reward functions without breaking existing tools that don't need this information. 

Fixes #4866

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


lukehinds:
hi @qgallouedec do you have enough info to review this now? happy to jump on discord and chat through or fill in any more details on the original issue. I believe its a low risk change as the code is only leveraged when the parameter is passed, so existing users will not be impacted.

lukehinds:
How to get a review folks, I know you have a lot of work - so I am happy to help in anyway if more context is needed.

qgallouedec:
Hi, and sorry for the late reply.

I understand your use case, and I agree that adding the index as you suggested would technically solve the issue. However, since reward functions are not meant to depend on the index, I don’t want to allow that. It would essentially encourage a workaround that goes against the intended design.

That said, I’ve been working on an alternative solution that I believe should work perfectly for your case: environment support. You can take a look here: [https://github.com/huggingface/trl/pull/5093](https://github.com/huggingface/trl/pull/5093).

The idea is that tools are called within the context of a class (method instead of function), which makes it possible to maintain any state for reward computation in a clean and intended way.

qgallouedec:
Closing in favor of #5093

Issue #4879, opened 2026-01-20
Automatic adding of {% generation %} tags to chat_template for training with assistant_only_loss=True


Neelectric:
### Feature request

🤗Tokenizers added support for `{% generation %} {% endgeneration %}` tags, such that when using `tokenizer.apply_chat_template(..., return_assistant_tokens_mask=True)`, all non-assistant tokens are masked. This is necessary for `SFTConfig(..., assistant_only_loss=True)` to work as expected. The problem is that many existing chat templates do not yet have `{% generation %} {% endgeneration %}` tags, requiring the users to edit the chat templates correctly themselves. This feature request suggests implementing functionality such that if `assistant_only_loss=True` is passed, and the chat_template does not yet have  `{% generation %} {% endgeneration %}` tags, the model type is automatically recognised and amended appropriately. This follows suggestions by 🤗 users 'waleko' (https://huggingface.co/Qwen/Qwen3-8B/discussions/14) and 'HarryMayne' (https://github.com/HarryMayne/qwen_3_chat_templates, code below is copied and slightly adapted from HarryMayne.):

```
from tokenizers import AutoTokenizer
from trl.chat_template_utils import add_generation_tags

# Load tokenizer
tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3-8B")

# Automatically add generation tags
tokenizer = add_generation_tags(tokenizer)

# Apply template with assistant token masking
conversation = [
    {"role": "user", "content": "Hello assistant"},
    {"role": "assistant", "content": "Hello user"},
    {"role": "user", "content": "How are you?"},
    {"role": "assistant", "content": "I'm good"},
]

tokenized_output = tokenizer.apply_chat_template(
    conversation,
    return_assistant_tokens_mask=True,
    return_dict=True,
)

# BEFORE
# {'input_ids': [151644, 872, 198, 9707, 17847, 151645, 198, 151644, 77091, 198, 9707, 1196, 151645, 198, 151644, 872, 198, 4340, 525, 498, 30, 151645, 198, 151644, 77091, 198, 151667, 271, 151668, 271, 40, 2776, 1661, 151645, 198], 
#  'attention_mask': [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
[...]

PR #4900, opened 2026-01-26
Automatically add generation tags to chat template for assistant_only_loss=True training (TRL Issue #4879)

Neelectric:
# What does this PR do?
This is a very first step at implementing the feature requested in #4879. For now the approach would be trying to recognise the model type and then replacing the chat template with a 'fixed' version. This would be helpful for the most common models (Qwen, Llama, OLMo etc) but I wonder if there's a more fancy way to go about this that automatically recognises when the assistant starts messaging, which we would want to train on.

Implements #4879 (feature request)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec


trl/chat_template_utils.py
```diff
@@ -275,6 +275,289 @@ def add_response_schema(tokenizer: PreTrainedTokenizer) -> PreTrainedTokenizer:
...
+    ```
+    """
+    # note that the Qwen team actually rejected the request to change their Qwen3 chat templates on a HuggingFace discussion: https://huggingface.co/Qwen/Qwen3-8B/discussions/14
+    # while their response makes sense within their context on a model-wide basis, if the user passes 'assistant_only_loss=True' we probably want to support this
+
+    # Full credit to HuggingFace users 'waleko' and 'HarryMayne' for their progress with Qwen3, much of the code and the templates above is taken from:
+    # https://huggingface.co/Qwen/Qwen3-8B/discussions/14 and https://github.com/HarryMayne/qwen_3_chat_templates
+    if tokenizer.chat_template == qwen3_chat_template:
+        tokenizer.chat_template = qwen3_chat_template_all_assistant
+    # probably want to add support here for most other popular model families like Llama 2/3/3.1, OLMo 2/3/3.1 etc... is there a way to do this fully automatically?
```
qgallouedec, line 549: I don't think we can do it 100% automatically, but I think that there are not thousands of different chat templates out there, if we support the top 5 chat templates, we may covers a good majority of use cases

qgallouedec:
This looks solid so far. A couple things still missing:

* A unit test.
* In `SFTTrainer`, we should call this function when `assistant_only_loss` is enabled (and when `{% generation %}` isn't in the chat template.

Also, in `SFTConfig`, we should explicitly mention that toggling `assistant_only_loss` may modify the tokenizer’s chat template.

---

Ideally, this feature *wouldn’t* mutate the tokenizer: we can avoid that by calling `apply_chat_template` with the `chat_template` argument set to the enhanced template (this is what we do in the GRPO trainer). That said, I’m okay with modifying the chat template for now, it keeps the implementation simpler.

Sirri69:
Dawg this needs to be merged asap, we burned like $500 in training without this in place lmao

qgallouedec:
Thanks @Neelectric! We can't really do it automatically, so we've taken a different approach

PR #4905, opened 2026-01-27
Remove gradient checkpointing option from various training scripts 

qgallouedec:
`gradient_checkpointing` is true by default, there is no need to have it everywhere

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -49,6 +49,8 @@
     title: Liger Kernel
   - local: peft_integration
     title: PEFT
+  - local: ptt_integration
+    title: Post Training Toolkit
   - local: rapidfire_integration
     title: RapidFire AI
   - local: trackio_integration
--- a/docs/source/ptt_integration.md
+++ b/docs/source/ptt_integration.md
@@ -1,0 +1,193 @@
+# Post-Training Toolkit Integration
+
+[Post-Training Toolkit](https://github.com/microsoft/post-training-toolkit) is a diagnostic and observability layer for RLHF training runs. Add one callback to any TRL trainer and get **auto-metrics**, **crash postmortems**, and **literature-backed heuristics**—without writing glue code.
+
+It was built to operationalize the debugging patterns we found most useful when running post-training at scale.
+
+## Usage
+
+1. First, install Post-Training Toolkit:
+
+```bash
+pip install post-training-toolkit
+```
+
+2. Add one callback to your trainer. That's it!
+
+<hfoptions id="trainer">
+<hfoption id="DPO">
+
+```python
+from post_training_toolkit import DiagnosticsCallback
+from trl import DPOTrainer
+
+trainer = DPOTrainer(
+    model=model,
+    args=training_args,
+    callbacks=[DiagnosticsCallback()],  # ← Just add this
+    ...
+)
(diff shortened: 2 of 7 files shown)


tests/test_grpo_trainer.py
```diff
@@ -414,55 +414,6 @@ def test_training_peft_model(self):
             elif "base_layer" not in n and "ref" not in n:  # and the peft params to be different (except base and ref)
                 assert not torch.allclose(param, new_param), f"Parameter {n} has not changed."
 
-    @require_peft
-    def test_training_peft_with_gradient_checkpointing(self):
-        """Test that training works with PEFT and gradient checkpointing enabled."""
```
albertvillanova, line 419: Shouldn't we keep this test? It specifically validates that gradient checkpointing works correctly with PEFT models. Even if gradient checkpointing is enabled by default, this test serves to verify that the integration doesn't break. Consider keeping this test but renaming it to reflect that it tests the default behavior (e.g., `test_training_peft_with_default_gradient_checkpointing`).

tests/test_reward_trainer.py
```diff
@@ -446,42 +446,6 @@ def test_train_moe_with_peft_config_and_gradient_checkpointing(self):
             elif "base_layer" not in n:  # We expect the peft parameters to be different (except for the base layer)
                 assert not torch.allclose(param, new_param), f"Parameter {n} has not changed"
 
-    @require_peft
-    def test_train_with_peft_model_and_gradient_checkpointing(self):
-        # Get the base model parameter names
```
albertvillanova, line 451: The same as above: it validates a specific combination of PEFT + gradient checkpointing. The removal reduces test coverage for this important use case. Consider retaining but updating the docstring to clarify it tests default behavior.

tests/test_dpo_trainer.py
```diff
@@ -1515,7 +1515,6 @@ def test_dpo_peft_model(self, model_id, loss_type, pre_compute_logits, gradient_
             fp16=True,
             logging_strategy="no",
             report_to="none",
-            gradient_checkpointing=True,
```
albertvillanova, line 1518: Maybe adding an explanatory comment about passing `gradient_checkpointing_kwargs` below because gradient_checkpointing is True by default?

tests/test_grpo_trainer.py
qgallouedec: Okay, I understand the reasoning. Let's keep one test that explicitly tests the GC, with a clear comment explaining why we have it.

https://github.com/huggingface/trl/pull/4905/commits/e41654b40fc1493ca888ab6b42307ab218df09cc

tests/test_dpo_trainer.py
qgallouedec: a reverted this change and added a comment in https://github.com/huggingface/trl/pull/4905/commits/e41654b40fc1493ca888ab6b42307ab218df09cc

tests/test_grpo_trainer.py
qgallouedec: oops I missed grpo

qgallouedec: https://github.com/huggingface/trl/pull/4905/commits/77e4e48f0b1202f0f8e43eea6929110567d2b510

PR #4906, opened 2026-01-27
Remove chat template setup in dpo_vlm.py

qgallouedec:
Setting a chat template before a DPO training doesn't really make sense. We should do it in the SFT stage.

### DIFF
--- a/examples/scripts/dpo_vlm.py
+++ b/examples/scripts/dpo_vlm.py
@@ -33,7 +33,7 @@ accelerate launch examples/scripts/dpo_vlm.py \
     --per_device_train_batch_size 2 \
     --gradient_accumulation_steps 32 \
     --dataset_num_proc 32 \
-    --output_dir dpo_idefics_rlaif-v \
+    --output_dir dpo_qwen_2_5_rlaif-v \
     --dtype bfloat16 \
     --gradient_checkpointing \
     --use_peft \
@@ -51,7 +51,7 @@ accelerate launch examples/scripts/dpo_vlm.py \
     --max_steps 100 \
     --gradient_accumulation_steps 32 \
     --dataset_num_proc 32 \
-    --output_dir dpo_idefics_rlaif-v \
+    --output_dir dpo_qwen_2_5_rlaif-v \
     --dtype bfloat16 \
     --gradient_checkpointing \
     --use_peft \


qgallouedec:
> Thanks for the simplification.
> 
> While the cleanup itself is clean and the reasoning is sound in principle, I'm wondering about backward compatibility and whether users who skip SFT or use pre-existing checkpoints will encounter issues.
> 
> The main concern is ensuring that this doesn't break existing workflows where users might run DPO training directly without a prior SFT stage, or where the processor doesn't have a chat template already configured.
>
> Could you clarify the intended usage pattern and confirm that all supported workflows will continue to function after this change?

This change only affects the example, so there isn’t really a backward-compatibility concern in the library itself.

For users who run DPO before SFT (to be precise, before instruction tuning), behavior depends on the base model + dataset format:

* If the base model has a chat template (a bit of an odd setup if the base model was never trained with it, but it’s common—e.g. some Qwen base checkpoints expose one):
  * Conversational data: training won’t error, but DPO is unlikely to learn much because the formatted inputs can be far from the model’s pretraining distribution.
  * Standard (non-chat) data: should work fine.
* If the base model has no chat template:
  * Conversational data: will error when the example tries to apply a chat template (see below).
  * Standard data: should work fine.

The key point is: this PR doesn’t change `DPOTrainer`, so there’s no before/after behavioral difference in the trainer. The only difference is in the example: if someone runs it on a model without a chat template (e.g. LLaVA), it will now fail fast instead of silently "training" on mismatched formatting and effectively not learning.

> Additionally, what about documenting somehow where users should now configure chat templates? Maybe:
> 
> * Add a comment or docstring in `dpo_vlm.py` pointing to the SFT example
> * Update any relevant docs (tutorial or guide) that reference the DPO training workflow
> [...]
> In relation with tests:
> 
> * Do we validate that the script handles missing chat templates gracefully?
>   
>   * Add validation to fail fast with a clear error message

If a user tries to apply DPO with conversational preference data on a base model (before instruction tuning), it will raise this error

```python
from trl import DPOTrainer

model = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-3.2-1B")
dataset = load_dataset("trl-internal-testing/zen-image", "conversational_preference", split="train")
trainer = DPOTrainer(model=model, train_dataset=dataset)
```

```
  File "/fsx/qgallouedec/trl/debug.py", line 7, in <module>
    trainer = DPOTrainer(model=model, train_dataset=dataset)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/trl/trl/trainer/dpo_trainer.py", line 490, in __init__
    train_dataset = self._prepare_dataset(train_dataset, processing_class, args, "train")
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/trl/trl/trainer/dpo_trainer.py", line 659, in _prepare_dataset
    dataset = dataset.map(
              ^^^^^^^^^^^^
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/datasets/arrow_dataset.py", line 562, in wrapper
    out: Union["Dataset", "DatasetDict"] = func(self, *args, **kwargs)
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/datasets/arrow_dataset.py", line 3343, in map
    for rank, done, content in Dataset._map_single(**unprocessed_kwargs):
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/datasets/arrow_dataset.py", line 3675, in _map_single
    for i, example in iter_outputs(shard_iterable):
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/datasets/arrow_dataset.py", line 3649, in iter_outputs
    yield i, apply_function(example, i, offset=offset)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/datasets/arrow_dataset.py", line 3572, in apply_function
    processed_inputs = function(*fn_args, *additional_args, **fn_kwargs)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/trl/trl/data_utils.py", line 378, in maybe_apply_chat_template
    return apply_chat_template(example, tokenizer, tools, **template_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/trl/trl/data_utils.py", line 231, in apply_chat_template
    prompt = tokenizer.apply_chat_template(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/transformers/tokenization_utils_base.py", line 3112, in apply_chat_template
    chat_template = self.get_chat_template(chat_template, tools)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/transformers/tokenization_utils_base.py", line 3294, in get_chat_template
    raise ValueError(
ValueError: Cannot use chat template functions because tokenizer.chat_template is not set and no template argument was passed! For information about writing templates and setting the tokenizer.chat_template attribute, please see the documentation at https://huggingface.co/docs/transformers/main/en/chat_templating
```

This error is mildly informative for beginners, but for average post-trainining practitioner, I think it's informative enough so that they detectsthat it may be training the wrong checkpoint, because they know that an instruct model should have a chat template. I don't think there is a urgent need for a more informative message.

Once again, no difference between/after this PR, we only change an example here.

> * Are we losing these chat templates settings for these specific models? Does the SFT script properly sets up chat templates for these model types?

Yes, we are loosing these chat templates for these models, but we now have a cleaner way to set chat templates, see https://huggingface.co/docs/trl/en/sft_trainer#instruction-tuning-example

PR #4908, opened 2026-01-27
Fix extra EOS appended in DPO preprocessing for conversational data

qgallouedec:
This PR fixes a bug in DPO data preparation where the preprocessing step appending an EOS token for conversational data, but it shouldn't

#### Repro

```python
from trl import DPOTrainer

dataset = load_dataset("trl-internal-testing/zen", "conversational_preference", split="train")

trainer = DPOTrainer(
    model="Qwen/Qwen3-0.6B",
    train_dataset=dataset,
)

print(repr(trainer.processing_class.decode(trainer.train_dataset["chosen_input_ids"][0])))
# Before: "<think>\n\n</think>\n\nBeautiful.<|im_end|>\n<|im_end|>"
# After:  "<think>\n\n</think>\n\nBeautiful.<|im_end|>"
```

This was found when working on #3906.

Note that the solution is probably not perfectly clean nor tested, but that's fine in my opinion, it's just to have a correct DPO to compare against in #3906

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -707,6 +707,9 @@ class DPOTrainer(BaseTrainer):
                 Whether to add special tokens to the sequences. Typically used for encoder-decoder models. If `True`,
                 the prompt sequence will have a bos token prepended and an eos token appended. In any case, the
                 completion sequences will have an eos token appended.
+            is_chat (`bool`):
+                Whether the data is conversational. If `True`, the completion sequences will not have an eos token
+                appended.
 
         Returns:
             `dict[str, list[int]]`:


trl/trainer/dpo_trainer.py
```diff
@@ -732,8 +736,9 @@ def tokenize_row(
                 prompt_input_ids = [tokenizer.bos_token_id] + prompt_input_ids
             if tokenizer.eos_token_id is not None:
                 prompt_input_ids = prompt_input_ids + [tokenizer.eos_token_id]
-        chosen_input_ids = chosen_input_ids + [tokenizer.eos_token_id]
-        rejected_input_ids = rejected_input_ids + [tokenizer.eos_token_id]
+        if not is_chat:
+            chosen_input_ids = chosen_input_ids + [tokenizer.eos_token_id]
+            rejected_input_ids = rejected_input_ids + [tokenizer.eos_token_id]
```
albertvillanova, line 745: What about adding an explanatory comment for future debuggers? Something like:
```suggestion
        # For conversational data, the chat template already includes proper EOS tokens
        if not is_chat:
            chosen_input_ids = chosen_input_ids + [tokenizer.eos_token_id]
            rejected_input_ids = rejected_input_ids + [tokenizer.eos_token_id]
```

trl/trainer/dpo_trainer.py
```diff
@@ -686,6 +689,7 @@ def tokenize_row(
         max_prompt_length: int | None = None,
         max_completion_length: int | None = None,
         add_special_tokens: bool = True,
+        is_chat: bool = False,
```
albertvillanova, line 692: I think you forgot to update the docstring to document this new parameter.

qgallouedec:
> Maybe a naive question: are we sure all chat template include an EOS token to conversational data?

Yes, applying chat template is supposed to add the proper special tokens, see https://huggingface.co/blog/qgallouedec/gotchas-in-tokenizer-behavior#4-applying-the-chat-template-could-add-the-eos-token-but-sometimes-it-doesnt-and-sometimes-it-does-but-not-at-the-end

(the case "Some don’t add an EOS at all, like `databricks/dbrx-instruct`" is actually a bad use of EOS/chat template, we don't want to support this case)

> Maybe we could easily add some regression tests in this case, for tokenize_row and process_row. What do you think?

I won't dedicate too much time on improving the trainer/test, mostly because I'm refactoring DPO in #3906, and in this refacoring I'm dropping `tokenize_row` altogether to be more aligned with the `SFTTrainer/RewardTrainer._prepare_dataset`

trl/trainer/dpo_trainer.py
qgallouedec: indeed, fixed in https://github.com/huggingface/trl/pull/4908/commits/f742ef95a44e2501ec4826e6d245c4caf9bd7791

PR #4909, opened 2026-01-27
`device_map` init consistency in GRPO/RLOO/KTO

qgallouedec:
When training with multi-gpu, `accelerate.prepare` requires the model `device_map` to be `None`.

This is ensured since https://github.com/huggingface/trl/pull/3950

This PR aims to add consistency when instantiating the model, by having the same condition everywhere.

/!\ Note that this PR is not fixing any bug, because the `device_map` is already modified inplace here:

https://github.com/huggingface/trl/blob/5cb03863487da1850d74f4b7602a25f3f8f27cd8/trl/trainer/grpo_trainer.py#L270-L272

and here

https://github.com/huggingface/trl/blob/5cb03863487da1850d74f4b7602a25f3f8f27cd8/trl/trainer/rloo_trainer.py#L244-L246

so technically the `if` condition is never met, but we're just adding a bit of consistency in the code.

### DIFF
--- a/trl/experimental/kto/kto_trainer.py
+++ b/trl/experimental/kto/kto_trainer.py
@@ -339,7 +339,7 @@ class KTOTrainer(BaseTrainer):
         # Model initialization
         if isinstance(model, str):
             model_init_kwargs = args.model_init_kwargs or {}
-            # Distributed training requires device_map=None ("auto" fails with DeepSpeed/FSDP)
+            # Distributed training requires device_map=None ("auto" fails)
             if args.distributed_state.distributed_type in ["MULTI_GPU", "DEEPSPEED"]:
                 model_init_kwargs["device_map"] = None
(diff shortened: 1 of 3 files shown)


trl/experimental/kto/kto_trainer.py
```diff
@@ -339,7 +339,7 @@ def __init__(
         # Model initialization
         if isinstance(model, str):
             model_init_kwargs = args.model_init_kwargs or {}
-            # Distributed training requires device_map=None ("auto" fails with DeepSpeed/FSDP)
```
albertvillanova, line 342: Maybe a naive question: does "FSDP" have a specific type or is it "MULTI_GPU"?

qgallouedec: I discovered the answer while debugging this PR: when FSDP is enabled, `distributed_type==MULTI_GPU` 🤷‍♂️

PR #4914, opened 2026-01-28
Update learning rate comments and add assertions for reference model parameters in GRPO and RLOO tests

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/tests/test_rloo_trainer.py
+++ b/tests/test_rloo_trainer.py
@@ -627,6 +627,7 @@ class TestRLOOTrainer(TrlTestCase):
 
         training_args = RLOOConfig(
             output_dir=self.tmp_dir,
+            beta=0.1,  # ensure ref model is created so sync can update it
             learning_rate=0.1,  # use higher lr because gradients are tiny and default lr can stall updates
             per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
             num_generations=3,  # reduce the number of generations to reduce memory usage


tests/test_grpo_trainer.py
```diff
@@ -387,7 +387,7 @@ def test_training_peft_model(self):
 
         training_args = GRPOConfig(
             output_dir=self.tmp_dir,
-            learning_rate=0.1,  # increase the learning rate to speed up the test
+            learning_rate=0.1,  # use higher lr because gradients are tiny and default lr can stall updates
```
qgallouedec, line 390: see https://github.com/huggingface/trl/pull/4810

tests/test_grpo_trainer.py
```diff
@@ -842,6 +845,8 @@ def test_training_with_sync_ref_model(self):
         for n, param in previous_trainable_params.items():
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
+            new_ref_param = trainer.ref_model.get_parameter(n)
+            assert not torch.equal(previous_ref_params[n], new_ref_param), f"Ref Parameter {n} has not changed."
```
albertvillanova, line 849: Just a naive question: we are sure both models (model and ref_model) have the same parameters and parameter names?

albertvillanova: Anyway, maybe we could make it more explicit with 2 loops like:
```python
# Check trainable params updated

# Check ref params updated
for n, param in previous_ref_params.items():
    new_ref_param = trainer.ref_model.get_parameter(n)
    assert not torch.equal(param, new_ref_param), f"Ref Parameter {n} has not changed."
```

tests/test_rloo_trainer.py
albertvillanova, line 629: You explicitly added this to GRPO but not to RLOO:

```suggestion
            output_dir=self.tmp_dir,
            beta=0.1,  # ensure ref model is created so sync can update it
```

tests/test_rloo_trainer.py
```diff
@@ -652,6 +654,8 @@ def test_training_with_sync_ref_model(self):
         for n, param in previous_trainable_params.items():
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
+            new_ref_param = trainer.ref_model.get_parameter(n)
+            assert not torch.equal(previous_ref_params[n], new_ref_param), f"Ref Parameter {n} has not changed."
```
albertvillanova, line 659: The same as above: maybe 2 loops?

albertvillanova (approved):
Thanks. Just some suggestions below.

tests/test_grpo_trainer.py
qgallouedec: > Just a naive question: we are sure both models (model and ref_model) have the same parameters and parameter names?

Yes, it's a hard requirement, it has to be the same model, only the weights are different.

tests/test_rloo_trainer.py
qgallouedec: `beta>0` by default in RLOO, but you're right, it's probably better to have it explicitly.

tests/test_grpo_trainer.py
qgallouedec: Therefore, I prefer this single-loop option because it clearly indicates that the parameters match, and if for some reason they do not match (and magically, this doesn't make the training fail), the single loop will detect this case, unlike the double loop.

PR #4927, opened 2026-01-29
Set default top_k to 0 in VLLMClient

albertvillanova:
Set default top_k to 0 in VLLMClient

This PR updates the `VLLMClient.generate` and `VLLMClient.chat` methods, changing the default value of the `top_k` parameter from `-1` to `0` to indicate that `0` now means no truncation, and updating the docstring accordingly.

This change improves clarity and consistency, aligning with the rest of the code:
- #4695

### DIFF
--- a/trl/generation/vllm_client.py
+++ b/trl/generation/vllm_client.py
@@ -28,7 +28,7 @@ from torch import nn
 from transformers import is_torch_xpu_available
 from urllib3.util.retry import Retry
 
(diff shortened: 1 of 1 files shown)


qgallouedec (changes requested):
can you apply the same change in `VLLMClient.chat` as well

albertvillanova:
Thanks @qgallouedec: I had missed that one! :sweat_smile: 
Done.

PR #4937, opened 2026-01-30
documentation for modifying chat templates for assistant-only loss

jiosephlee:
# What does this PR do?

While this issue is not super prevalent (most datasets are single-turn interactions and can be addressed with completion-only fine-tuning), some users want to fine-tune on multi-turn interactions or responses with tool-calling.

To my knowledge, this issue can be easily addressed with a proper chat template that adds {% generation %} when appropriate, including executed tool responses, so that the tokenizer returns the masks automatically. If the chat template doesn't do this, there is an ongoing PR to automatically address the issue with chat templates (https://github.com/huggingface/trl/pull/4900), but this doesn't scale well imo. 

An easy solution would be to provide sufficient documentation so users can edit chat templates themselves!


qgallouedec:
Sorry, this one slipped through with no reply at all.

A lot landed in this area since January. There's now a `chat_templates` page in the docs, TRL ships patched training templates under `trl/chat_templates/` and swaps them in automatically when `assistant_only_loss=True` for about thirty model families (Qwen, Llama, Gemma, DeepSeek, GPT-OSS, ...), and `chat_template_utils` documents the helpers. So the framing here is out of date: for supported models it no longer fails, and the Qwen-style `<|im_start|>` example you use is one of the auto-patched families.

What's still missing is the part you wrote well. The chat templates page says "For any other model, you'll need to patch its template yourself" and then never shows how. Your Steps 1 to 3 fill exactly that.

So I'd take this as a short section at the top of `docs/source/chat_templates.md` rather than a new page: locate the template, wrap the assistant branch with the generation tags (with the tool-call vs tool-response distinction, that part is useful), apply it via `chat_template_path`.
Pick a model that isn't auto-patched for the example. Drop the intro and the troubleshooting stub, both are covered on that page already.

I'll close this one as it's too old but feel to open a fresh one from main

PR #4950, opened 2026-02-02
Add test for training with `compute_metrics` in `SFTTrainer`

qgallouedec:
Just adding a test, it seems to be a widely used option in SFTTrainer.

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1252,7 +1252,7 @@ class TestSFTTrainer(TrlTestCase):
         # Train the model
         trainer.train()
 
-        # Check that the training loss is not None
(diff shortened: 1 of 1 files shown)


tests/test_sft_trainer.py
```diff
@@ -1227,6 +1227,34 @@ def test_train_with_multiple_eval_dataset(self):
...
+            args=training_args,
+            train_dataset=dataset["train"],
+            eval_dataset=dataset["test"],
+            compute_metrics=dummy_compute_metrics,
+        )
+
+        # Train the model
+        trainer.train()
+
+        # Check that the training loss is not None
```
```suggestion
        # Check that the custom metric is logged
```

PR #4951, opened 2026-02-02
Fix ZeRO-3 + PEFT + gradient checkpointing

qgallouedec:
This PR does 4 things:

### 1. Move  the `use_reentrant` default overriding inside the trainers

just before calling `super().__init__`. This is required by (2) because we need to know during the init if the user has explicitly pass `use_reentrant=False` or not to raise a warning

### 2. When PEFT is used with DeepSpeed ZeRO-3 and gradient checkpointing in SFTTrainer, force using reentrant

When `use_reentrant` is set to `False` **by the user**, it also raises a warnings:

```
You are using PEFT with DeepSpeed ZeRO-3 and gradient checkpointing with `use_reentrant=False`.
`use_reentrant` is forced to `True` in this configuration to ensure correct training. To remove
this warning, unset `use_reentrant` in `gradient_checkpointing_kwargs` or set it to `True`.
```

### 3. Update the xfail reason of `TestDistributed::test_sft_peft[zero2]`

Now, `TestDistributed::test_sft_peft[zero2]` still fails, but for the same reason as `TestDistributed::test_sft_peft[zero3]`

### 4. Update the xfail condition of `TestDistributed::test_sft_peft[zero2,zero3]` to torch 2.10 only

After investigation, `TestDistributed::test_sft_peft[zero2,zero3]` are failing because of torch 2.10 specifically. For more information see https://github.com/huggingface/transformers/pull/43692.

### DIFF
--- a/docs/source/lora_without_regret.md
+++ b/docs/source/lora_without_regret.md
@@ -195,7 +195,7 @@ hf jobs uv run \
     --output_dir grpo-full-qwen3-0.6b \
     --learning_rate 1.0e-6 \
     --lr_scheduler_type cosine \
-    --warmup_ratio 0.0 \
+    --warmup_steps 0.0 \
     --max_grad_norm 1.0 \
     --beta 0.0 \
     --max_completion_length 4096 \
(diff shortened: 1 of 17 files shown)


albertvillanova (approved):
Thanks. I thinks this change is because we also accept TrainingArguments as args.

PR #4958, opened 2026-02-03
Add test for training with `compute_metrics` in `RewardTrainer`

qgallouedec:
Same as #4950


qgallouedec:
> Thanks. Are you planning to add the same test for other trainers?

yes, for those that support it, ie not GRPO and RLOO, see #4950 for SFT, and for DPO this is covered in #3906

PR #4959, opened 2026-02-03
Add test for tool call data in `RewardTrainer`

qgallouedec:
See the new dataset here:

https://huggingface.co/datasets/trl-internal-testing/toolcall/viewer/preference


albertvillanova (approved):
Thanks.

PR #4962, opened 2026-02-04
Remove padding_value from experimental CPO and use pad_token_id

albertvillanova:
Remove padding_value from CPO and use pad_token_id.

This PR removes the `padding_value` parameter from the `CPOConfig` class and updates the trainer logic to always use the tokenizer's `pad_token_id` for padding. This simplifies configuration and ensures consistent padding behavior throughout the codebase.

Follow-up to:
- #4846
- #4839


trl/experimental/cpo/cpo_trainer.py
```diff
@@ -299,7 +299,7 @@ def make_inputs_require_grad(module, input, output):
 
         self.max_length = max_length
         self.generate_during_eval = args.generate_during_eval
-        self.padding_value = args.padding_value if args.padding_value is not None else processing_class.pad_token_id
+        self.pad_token_id = processing_class.pad_token_id
```
qgallouedec: the issue is that, if the tokenizer doesn't have a padding token (which we should allow), it will fail. We should align with stable trainers:

https://github.com/huggingface/trl/blob/963d046b12219014bf4eba80b3a931707ddaa461/trl/trainer/grpo_trainer.py#L305-L309

albertvillanova: Done.

PR #4963, opened 2026-02-04
Remove max_prompt_length from experimental PRM

albertvillanova:
Remove `max_prompt_length` from experimental PRM.

This PR removes support for the `max_prompt_length` parameter from the PRM trainer code and its associated tests.

Follow-up to:
- #4808
- #4703
- #3906

Key changes include:

**Core code and configuration cleanup:**

* Removed the `max_prompt_length` parameter and its logic from the `PRMConfig` class, the `PRMTrainer.tokenize_row` method, and related configuration and function signatures. 
* Updated the truncation logic in `tokenize_row` removing prompt-specific truncation.

**Testing updates:**

* Removed all uses of `max_prompt_length` from test cases and deleted the test for prompt truncation, as this functionality is no longer supported.

PR #4964, opened 2026-02-04
Remove max_prompt_length from experimental BCO

albertvillanova:
Remove max_prompt_length from experimental BCO.

This PR removes the `max_prompt_length` parameter from the BCO training pipeline, simplifying both the configuration and the tokenization logic. 

Follow-up to:
- #4963
- #4808
- #4703
- #3906

**Configuration and Argument Cleanup**
- Removed the `max_prompt_length` argument from the `BCOConfig` class and its documentation, as well as from all related code paths and warnings. This reduces user confusion and simplifies configuration. 

**Tokenization and Preprocessing Logic**
- Updated the `_process_tokens` function to eliminate prompt truncation based on `max_prompt_length`. Now, only the response is truncated if the total sequence exceeds `max_length`, and prompt tokenization no longer uses a prompt length limit. 
- Removed the passing of `max_prompt_length` in dataset mapping and processing, ensuring consistent use of the new logic across training and evaluation. 

**Example and Script Updates**
- Cleaned up example scripts by removing `--max_prompt_length` from CLI arguments, aligning usage examples with the new configuration. 

**Testing updates:**
- Removed all uses of `max_prompt_length` from test cases.

### DIFF
--- a/trl/experimental/bco/bco_trainer.py
+++ b/trl/experimental/bco/bco_trainer.py
@@ -252,7 +252,7 @@ def _process_tokens(example: dict[str, Any], model: "PreTrainedModel" = None, **
         completion_tokens = kwargs["tokenizer"](
             completion, truncation=True, max_length=kwargs["max_completion_length"], add_special_tokens=True
         )
-        prompt_tokens = kwargs["tokenizer"](prompt, truncation=True, add_special_tokens=True)
+        prompt_tokens = kwargs["tokenizer"](prompt, add_special_tokens=True)
 
         batch[f"{kwargs['prefix']}prompt_input_ids"] = prompt_tokens["input_ids"]
         batch[f"{kwargs['prefix']}prompt_attention_mask"] = prompt_tokens["attention_mask"]


trl/experimental/bco/bco_trainer.py
```diff
@@ -262,9 +252,7 @@ def _process_tokens(example: dict[str, Any], model: "PreTrainedModel" = None, **
         completion_tokens = kwargs["tokenizer"](
             completion, truncation=True, max_length=kwargs["max_completion_length"], add_special_tokens=True
         )
-        prompt_tokens = kwargs["tokenizer"](
-            prompt, truncation=True, max_length=kwargs["max_prompt_length"], add_special_tokens=True
-        )
+        prompt_tokens = kwargs["tokenizer"](prompt, truncation=True, add_special_tokens=True)
```
qgallouedec: you can remove `truncation=True`

albertvillanova: Done.

PR #4969, opened 2026-02-04
Deprecate parameters in `DPOConfig`

qgallouedec:
This PR deprecates some parameters from the `DPOConfig`.

Most importantly, `max_prompt_length` is now None by default, and `force_use_ref_model` is enabled by default

For context, see https://github.com/huggingface/trl/pull/3906#discussion_r2754222616

### DIFF
--- a/trl/trainer/dpo_config.py
+++ b/trl/trainer/dpo_config.py
@@ -182,9 +182,9 @@ class DPOConfig(TrainingArguments):
 
             <Deprecated version="0.28.0">
 
-            The following parameter is deprecated and will be removed in version 0.29.0. In the future the base model
-            will be retrieved via `get_decoder`; if your model does not support this, it will no longer be supported by
-            the [`DPOTrainer`].
+            This parameter is deprecated and will be removed in version 0.29.0. In the future the base model will be
+            retrieved via `get_decoder`; if your model does not support this, it will no longer be supported by the
+            [`DPOTrainer`].
 
             </Deprecated>
         ref_model_init_kwargs (`dict[str, Any]`, *optional*):
@@ -193,8 +193,8 @@ class DPOConfig(TrainingArguments):
 
             <Deprecated version="0.28.0">
 
-            The following parameter is deprecated and will be removed in version 0.29.0. If you need different init
-            kwargs for the reference model, instantiate it yourself and pass it via the `ref_model` argument.
+            This parameter is deprecated and will be removed in version 0.29.0. If you need different init kwargs for
+            the reference model, instantiate it yourself and pass it via the `ref_model` argument.
 
             </Deprecated>
         model_adapter_name (`str`, *optional*):
@@ -203,8 +203,8 @@ class DPOConfig(TrainingArguments):
 
             <Deprecated version="0.28.0">
 
-            The following parameter is deprecated and will be removed in version 0.29.0. Only the default adapter will
-            be supported going forward.
+            This parameter is deprecated and will be removed in version 0.29.0. Only the default adapter will be
+            supported going forward.
 
             </Deprecated>
         ref_adapter_name (`str`, *optional*):
@@ -214,9 +214,9 @@ class DPOConfig(TrainingArguments):
 
             <Deprecated version="0.28.0">
 
-            The following parameter is deprecated and will be removed in version 0.29.0. If you used it to resume
-            training an adapter, you won't need this argument anymore in the next version and can rely on the trainer.
-            For now, it is still the only supported way to do this.
+            This parameter is deprecated and will be removed in version 0.29.0. If you used it to resume training an
+            adapter, you won't need this argument anymore in the next version and can rely on the trainer. For now, it
+            is still the only supported way to do this.
 
             </Deprecated>
         force_use_ref_model (`bool`, *optional*, defaults to `False`):
@@ -225,8 +225,8 @@ class DPOConfig(TrainingArguments):
 
             <Deprecated version="0.28.0">
 
-            The following parameter is deprecated and will be removed in version 0.29.0. There is no need to pass this
-            argument anymore: if you provide a reference model, it will be used automatically.
+            This parameter is deprecated and will be removed in version 0.29.0. There is no need to pass this argument
+            anymore: if you provide a reference model, it will be used automatically.
 
             </Deprecated>
         generate_during_eval (`bool`, *optional*, defaults to `False`):
@@ -235,8 +235,8 @@ class DPOConfig(TrainingArguments):
 
             <Deprecated version="0.28.0">
 
-            The following parameter is deprecated and will be removed in version 0.29.0. Please use a callback instead;
-            see `https://gist.github.com/qgallouedec/a08da3457a3a76c5ca539d4a0b38e482`.
+            This parameter is deprecated and will be removed in version 0.29.0. Please use a callback instead; see
+            `https://gist.github.com/qgallouedec/a08da3457a3a76c5ca539d4a0b38e482`.
 
             </Deprecated>
         label_pad_token_id (`int`, *optional*, defaults to `-100`):
@@ -244,8 +244,8 @@ class DPOConfig(TrainingArguments):
 
             <Deprecated version="0.28.0">
 
-            The following parameter is deprecated and will be removed in version 0.29.0. It will no longer be possible
-            to set this value.
+            This parameter is deprecated and will be removed in version 0.29.0. It will no longer be possible to set
+            this value.
 
             </Deprecated>
         max_prompt_length (`int` or `None`, *optional*, defaults to `512`):
@@ -254,8 +254,8 @@ class DPOConfig(TrainingArguments):
 
             <Deprecated version="0.28.0">
 
-            The following parameter is deprecated and will be removed in version 0.29.0. We recommend filtering
-            overlong prompts from your dataset before passing it to the trainer instead of using this parameter.
+            This parameter is deprecated and will be removed in version 0.29.0. We recommend filtering overlong prompts
+            from your dataset before passing it to the trainer instead of using this parameter.
 
             </Deprecated>
         max_completion_length (`int`, *optional*):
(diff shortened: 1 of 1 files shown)


trl/trainer/dpo_config.py
```diff
@@ -495,6 +521,101 @@ def __post_init__(self):
...
+                "`max_length` instead to control the maximum length of samples."
+            )
+
+        if self.max_prompt_length is not None:
+            warnings.warn(
+                "`max_prompt_length` is deprecated and will be removed in version 0.29.0. We recommend filtering out"
+                "overlong prompts from your dataset before passing it to the trainer instead of using this parameter."
+            )
+        else:  # keep the old default
+            self.max_prompt_length = 512
```
qgallouedec, line 670:
```suggestion
```

trl/trainer/dpo_config.py
```diff
@@ -207,11 +173,57 @@ class DPOConfig(TrainingArguments):
...
+            this flag to `True`. There is no need to pass this argument anymore: if you provide a reference model, it
+            will be used automatically.
         generate_during_eval (`bool`, *optional*, defaults to `False`):
             Whether to generate and log completions from both the model and the reference model to W&B or Comet during
-            evaluation.
+            evaluation. Please use a callback instead; see
+            `https://gist.github.com/qgallouedec/a08da3457a3a76c5ca539d4a0b38e482`.
+        label_pad_token_id (`int`, *optional*, defaults to `-100`):
+            Padding value to use for labels. It will no longer be possible to set this value.
+        max_prompt_length (`int` or `None`, *optional*, defaults to `512`):
```
qgallouedec, line 251:
```suggestion
        max_prompt_length (`int` or `None`, *optional*):
```

trl/trainer/dpo_config.py
albertvillanova: Why this change? We are supposed to keep the old behavior during the deprecation cycle.

I see there are some tests failing before this change, but why? I'm wondering why they were not failing before this PR if this PR is supposed to maintain the old behavior and just raise deprecation warnings.

trl/trainer/dpo_config.py
```diff
@@ -207,11 +173,57 @@ class DPOConfig(TrainingArguments):
             frequently the current policy is synchronized with the reference policy. To use this parameter, you must
             set `sync_ref_model=True`.
 
-        > Parameters that control the logging
+        > Deprecated parameters
```
albertvillanova, line 176: I think we should include the version when the parameters were deprecated and the version when this will take effect. Maybe we could add the `<Deprecated version="X.Y.Z">` tag.

If you think it is not necessary to put the Deprecated tag on every parameter, maybe we could just add a general Deprecated tag just after the `> Deprecated parameters` section title, if that works.

trl/trainer/dpo_config.py
```diff
@@ -475,12 +415,98 @@ class DPOConfig(TrainingArguments):
...
+    label_pad_token_id: int | None = field(
+        default=None,
+        metadata={"help": "Deprecated. It will no longer be possible to set this value."},
+    )
+    max_completion_length: int | None = field(
+        default=None,
+        metadata={"help": "Deprecated. Use `max_length` instead to control the maximum length of samples."},
+    )
+    max_prompt_length: int | None = field(
+        default=None,
```
albertvillanova: As I already asked before (https://github.com/huggingface/trl/pull/4969#discussion_r2767611064), I don't fully understand why we are changing the `max_prompt_length` default value in this PR.

trl/trainer/dpo_config.py
```diff
@@ -495,6 +521,99 @@ def __post_init__(self):
             self.gradient_checkpointing_kwargs = self.gradient_checkpointing_kwargs or {}
             self.gradient_checkpointing_kwargs.setdefault("use_reentrant", False)
 
+        if self.base_model_attribute_name is not None:
+            warnings.warn(
```
albertvillanova, line 616: I think we should emit `FutureWarning` with `stacklevel=2` in all the warnings here.

trl/trainer/dpo_config.py
qgallouedec: `max_prompt_length` allows either `None` (no truncation) or an int (truncation).

To deprecate `max_prompt_length`, we need to know when the user provides a value (-> raise a warning) and when it is left default.

The issue is that, the current (`main`) default is 512. So if we set:

```python
if max_prompt_length == 512:
    warnings.warn(...)
```

it will not catch the case where the user passes this value explicitly:

```python
DPOConfig()  # won't warn 😁
DPOConfig(max_prompt_length=16)  # will warn 😁
DPOConfig(max_prompt_length=512)  # won't warn 😖
```

Usually, we address this issue by changing the default to `None`, and when the value is not `None`, warn:

```python
if my_arg is not None:
    warnings.warn(...)
else:
    my_arg = old_default
```

but in this case, it's not possible, because `None` is actually a valid value for `max_prompt_lenght`, so:

```python
DPOConfig()  # won't warn 😁
DPOConfig(max_prompt_length=16)  # will warn 😁
DPOConfig(max_prompt_length=512)  # will warn 😁
DPOConfig(max_prompt_length=None)  # warn and silently enable truncation with len=512 💀
```

this is why we had these failures: https://github.com/huggingface/trl/actions/runs/21690691119/job/62549401888#step:9:2790

I don't have a better/clean alternative that would satisfy:

```python
DPOConfig()  # won't warn 😁
DPOConfig(max_prompt_length=16)  # will warn 😁
DPOConfig(max_prompt_length=512)  # will warn 😁
DPOConfig(max_prompt_length=None)  # will warn 😁
```

---

The thing is, prompt truncation is very likely detrimental, so in this case it might be ok no to maintain the old behavior. What do you guys think?

trl/trainer/dpo_config.py
qgallouedec: I will try the general tag. I dropped the tag because it seems like they aren't rendered anyway

trl/trainer/dpo_config.py
qgallouedec: answered here https://github.com/huggingface/trl/pull/4969#discussion_r2769777151

trl/trainer/dpo_config.py
qgallouedec: thanks! https://github.com/huggingface/trl/pull/4969/commits/6c8b4c8e862727c01c9cfd891a8bc80ec65c4efa

trl/trainer/dpo_config.py
qgallouedec: https://github.com/huggingface/trl/pull/4969/commits/9abc8486b74803f6de64a76560930e2c2b90885a let's see it it renders. In any case we can keep the tags.

trl/trainer/dpo_config.py
albertvillanova: What about setting a non-valid sentinel value of `-1`? If different from `-1`, then warn; if equals to `-1`, then set to 512.

albertvillanova: We could also use a sentinel object instead: `_UNSET = object()`; or `_MISSING`

trl/trainer/dpo_config.py
```diff
@@ -207,11 +173,61 @@ class DPOConfig(TrainingArguments):
             frequently the current policy is synchronized with the reference policy. To use this parameter, you must
             set `sync_ref_model=True`.
 
-        > Parameters that control the logging
+        > Deprecated parameters
 
+        <Deprecated version="0.28.0">
```
albertvillanova: Something like this would work?
```suggestion
        <Deprecated version="0.28.0">
        
        The following parameters are deprecated and will be removed in version ...
        
        </Deprecated>
```

trl/trainer/dpo_config.py
```diff
@@ -207,11 +173,61 @@ class DPOConfig(TrainingArguments):
...
+            List of tools (callable functions) that will be accessible to the model. If the template does not support
+            function calling, this argument will have no effect. In 0.29 this argument will be ignored; tools should be
+            provided via the dataset instead. For now, `DPOConfig.tools` remains the only supported way to pass tools.
+        use_logits_to_keep (`bool`, *optional*, defaults to `False`):
+            If `True`, only a specified number of logits are computed in the forward pass. This can be useful for
+            saving memory and speeding up training by not computing the logits for all tokens, especially in scenarios
+            when working with very long prompts where labels are ignored (-100). The DPO trainer will no longer use
+            this setting.
+
+        </Deprecated>
```
albertvillanova:
```suggestion
```

trl/trainer/dpo_config.py
qgallouedec: let's try with -1

qgallouedec:
@albertvillanova, even if it's currently broken, (see https://github.com/huggingface/doc-builder/issues/627), it's the documented way to deprecate parameters, so I'll just comply with it.

trl/trainer/dpo_config.py
```diff
@@ -207,11 +173,144 @@ class DPOConfig(TrainingArguments):
...
+        > Deprecated parameters
+
+        base_model_attribute_name (`str`, *optional*, defaults to `"model"`):
+            Name of the attribute in the model that contains the base model. This is used to get the base model from
+            the model when the model does not have a `get_decoder` method in the case when `use_liger_kernel` is
+            `True`.
+
+            <Deprecated version="0.28.0">
 
+            The following parameter is deprecated and will be removed in version 0.29.0. In the future the base model
```
albertvillanova: Nit: "following"? It's indeed the current one... Anyway, we used to set:
> Parameter `base_model_attribute_name` is deprecated...

trl/trainer/dpo_config.py
```diff
@@ -475,12 +502,102 @@ class DPOConfig(TrainingArguments):
...
+        metadata={"help": "Deprecated. It will no longer be possible to set this value."},
+    )
+    max_completion_length: int | None = field(
+        # This default value is used to determine whether the user has set it or not, since `None` is not a valid
+        # value for this parameter. This is overridden in `__post_init__` to preserve the old default value of `None`.
+        default=-1,
+        metadata={"help": "Deprecated. Use `max_length` instead to control the maximum length of samples."},
+    )
+    max_prompt_length: int | None = field(
+        # This default value is used to determine whether the user has set it or not, since `None` is not a valid
```
albertvillanova: Nit:

```suggestion
        # This default value is used to determine whether the user has set it or not, since `None` is a valid
```

PR #4970, opened 2026-02-04
Revert change in GRPO from NeMo-Gym Integration

qgallouedec:
#4848 introduced two changes in GRPO which are not correct. This PR reverts these changes, see PR comments for more details.

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1556,13 +1556,7 @@ class GRPOTrainer(BaseTrainer):
         prompt_ids = pad(prompt_ids, padding_value=self.pad_token_id, padding_side="left")
         prompt_mask = pad(prompt_mask, padding_value=0, padding_side="left")
         completion_ids = [torch.tensor(ids, device=device) for ids in completion_ids_list]
-
-        # Allow custom completion_mask from rollout_func for multi-turn training
-        if "completion_mask" in extra_fields:
-            completion_mask_list = extra_fields.pop("completion_mask")
-            completion_mask = [torch.tensor(m, device=device, dtype=torch.long) for m in completion_mask_list]
-        else:
-            completion_mask = [torch.ones_like(ids, dtype=torch.long) for ids in completion_ids]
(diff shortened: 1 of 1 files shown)


trl/trainer/grpo_trainer.py
```diff
@@ -1556,13 +1556,7 @@ def _generate_and_score_completions(
         prompt_ids = pad(prompt_ids, padding_value=self.pad_token_id, padding_side="left")
         prompt_mask = pad(prompt_mask, padding_value=0, padding_side="left")
         completion_ids = [torch.tensor(ids, device=device) for ids in completion_ids_list]
-
-        # Allow custom completion_mask from rollout_func for multi-turn training
-        if "completion_mask" in extra_fields:
-            completion_mask_list = extra_fields.pop("completion_mask")
-            completion_mask = [torch.tensor(m, device=device, dtype=torch.long) for m in completion_mask_list]
-        else:
-            completion_mask = [torch.ones_like(ids, dtype=torch.long) for ids in completion_ids]
```
qgallouedec, line 1565: this would allow to have a non-continuous attention mask in the forward pass. This would just break the causality (error out?). To mask some tokens in the loss, we should use the "env_mask" extra field.

trl/trainer/grpo_trainer.py
```diff
@@ -1584,10 +1578,7 @@ def _generate_and_score_completions(
 
         # Concatenate prompt_mask with completion_mask for logit computation
         prompt_completion_ids = torch.cat([prompt_ids, completion_ids], dim=1)  # (B, P+C)
-
-        # attend to all non-padding tokens, but mask out user/tool result tokens in loss
-        completion_attention_mask = (completion_ids != self.pad_token_id).long()
-        attention_mask = torch.cat([prompt_mask, completion_attention_mask], dim=1)  # (B, P+C)
```
qgallouedec, line 1590: pad tokens should be masked, but the mask should not be inferred from token values, because a same id can be used for other tokens (usually EOS). In practice, this often leads to unintentionally masking the EOS token. For more details, see: https://huggingface.co/blog/qgallouedec/gotchas-in-tokenizer-behavior#5-when-pad_token-equals-eos_token

PR #4971, opened 2026-02-04
Fix test_train_with_chat_template_kwargs

qgallouedec:
If you test with standard data, the chat template is never applied


qgallouedec:
Actually it's expected: the trainer would sample text (instead of messages). So instead of applying the chat template then tokenize, the trainer directly tokenizes the text.
It just takes another path than the one consuming the chat template kwargs

Issue #4974, opened 2026-02-05
Support multi-turn training in GRPOTrainer


albertvillanova:
Support multi-turn training in GRPOTrainer without breaking attention masking, required for NeMo-Gym integration

See:
- #4970
- #4848

### Context

Multi-turn training scenarios (e.g., agent interactions with tools, environment feedback) require the ability to mask certain tokens in the loss computation (user messages, tool results) while still attending to them during the forward pass.

The previous implementation attempted to support this via custom completion_mask from extra_fields, but introduced two critical bugs:
1. Inconsistent attention masking: Custom completion_mask with non-continuous values (zeros in the middle) was used for attention in loss computation, potentially breaking causality
2. Incorrect padding detection: Attention mask was inferred from token values (completion_ids != pad_token_id), which fails when pad_token_id == eos_token_id and incorrectly masks EOS tokens

### Requirements

The solution must:
  - Allow masking arbitrary tokens in loss computation (e.g., user turns, tool results)
  - Maintain correct causal attention for all non-padding tokens
  - Never infer padding positions from token IDs
  - Work correctly when pad_token_id == eos_token_id
  - Maintain consistency between generation and loss computation phases

hokuyama0106:
@albertvillanova 
Sorry, I understand we cannot currently use the nemo gym integration feature due to this issue in the latest branch.
Is it correct?
The script `trl/scripts/vllm_serve.py‎` also seems not to support the feature since it cannot receive the `/v1/chat/comletion` requests in the latest.

Is anyone working on this issue?
If not, I would like to consider working on this issue since I like to improve the multi-turn performance of our models.

qgallouedec:
Hi @hokuyama0106, sorry for the slow reply, we're catching up on the backlog.

Good news: correct multi-turn GRPO training is supported on current `main`, just not through the original NeMo-Gym integration (#4848 was reverted in #4970). Masking is now handled generally and consistently: attention attends to *all* non-padding tokens (user/tool turns included) via the all-ones `completion_mask`, while the loss masks out external tokens through a *separate* `tool_mask` derived from `env_mask`, so the two never conflict and padding is never inferred from token IDs (this fixes the `pad_token_id == eos_token_id` case). See the `env_mask` entry point at https://github.com/huggingface/trl/blob/44596766101a123f1a3ff70b0aefa54d926179ad/trl/trainer/grpo_trainer.py#L2082

[...]

PR #4976, opened 2026-02-05
Support truncated completions in GRPO multi-turn training

albertvillanova:
Support truncated completions in GRPO multi-turn training:
- Apply mask_truncated_completions to tool_mask.

This PR updates the masking logic in the `_generate_and_score_completions` method in `GRPOTrainer` to ensure that truncated completions are consistently masked for both attention and loss calculations. This improves the reliability of multi-turn training by properly handling truncated completions across all relevant masks.

Masking improvements for truncated completions:

* Updated the masking logic to zero out truncated completions in both `completion_mask` (for attention masking) and `tool_mask` (for multi-turn training consistency).

### Problem

When `mask_truncated_completions=True`, truncated completions (those not ending in EOS or PAD) are masked out by zeroing their completion_mask. However, tool_mask (which comes from env_mask in multi-turn raining) is not updated, causing:
1. Incorrect length reporting: Completion lengths are computed from tool_mask, which still contains non-zero values for truncated completions that are actually excluded from training
2. State inconsistency: completion_mask says "this completion is masked" but tool_mask says "these tokens are active"
3. Potential bugs: Any future code using tool_mask directly would see stale values for truncated completions

Example of the bug:
```python
# Scenario: Multi-turn completion is truncated (no EOS at end)
completion_ids = [1234, 5678, 9012]  # No EOS token - truncated!
completion_mask = [1, 1, 1]  # All tokens are real (not padding)
tool_mask = [1, 0, 1]  # Model tokens=1, env tokens=0
                                                                                                                                                                                                           
# Current behavior:                                                                                                                                                                                      
# Compute lengths BEFORE masking
completion_length = sum(tool_mask) = 2  # Reports 2 model tokens

# Line 1583: Zero out completion_mask
completion_mask = [0, 0, 0]  # Truncated, so mask everything
# tool_mask = [1, 0, 1]  # ❌ BUG: tool_mask NOT zeroed out

# Line 1978: Compute loss mask
loss_mask = [0, 0, 0] * [1, 0, 1] = [0, 0, 0]  # Loss correctly excludes it

# Result: We report 2 tokens trained, but actually 0 tokens used in loss!
```

### Root Cause

In `_generate_and_score_completions`, only completion_mask is zeroed for truncated completions.
- tool_mask is left unchanged!

### Solution

Apply the same masking to tool_mask when mask_truncated_completions is enabled.

### Related

This fix improves the consistency of the multi-turn training feature (env_mask) when used in combination with mask_truncated_completions. It ensures that both masking mechanisms work correctly together.

See:
- #4974

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1570,11 +1570,15 @@ class GRPOTrainer(BaseTrainer):
         else:
             tool_mask = None
 
-        # If mask_truncated_completions is enabled, zero out truncated completions in completion_mask
+        # If mask_truncated_completions is enabled, zero out truncated completions for attention and loss masking
         if self.mask_truncated_completions:
             eos_and_pad = [self.eos_token_id, self.pad_token_id]
             is_truncated = torch.tensor([ids[-1] not in eos_and_pad for ids in completion_ids_list], device=device)
+            # Mask completion_mask for attention masking
             completion_mask = completion_mask * (~is_truncated).unsqueeze(1).int()
+            # Also mask tool_mask for consistency in multi-turn training
+            if tool_mask is not None:
+                tool_mask = tool_mask * (~is_truncated).unsqueeze(1).int()
 
         # Concatenate prompt_mask with completion_mask for logit computation
         prompt_completion_ids = torch.cat([prompt_ids, completion_ids], dim=1)  # (B, P+C)


qgallouedec:
> 1. Incorrect length reporting: Completion lengths are computed from tool_mask


it's not only used for logging, it's also used for the normalization term for cispo and dapo losses. Thanks, great fix!

PR #4977, opened 2026-02-05
Replace `torch.allclose` with `torch.testing.assert_close`

qgallouedec:
See https://github.com/huggingface/transformers/pull/35659


qgallouedec:
Because `Both 'rtol' and 'atol' must be either specified or omitted` 🤷‍♂️ not sure why

qgallouedec:
https://github.com/huggingface/trl/actions/runs/21717128861/job/62636156839#step:9:1733

PR #4983, opened 2026-02-06
Replace `warmup_ratio` with `warmup_steps`

qgallouedec:
`warmup_ratio` is deprecated and will be removed in transformers 5.2.0. We can simply replace it by `warmup_since` since when the value is below 1, it's interpreted as a ratio.


albertvillanova (approved):
Thanks.

PR #4984, opened 2026-02-06
Update dataset configuration name in toolcall dataset loading

qgallouedec:
It's more aligned with how we call the formats

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1170,7 +1170,7 @@ class TestSFTTrainer(TrlTestCase):
 
     def test_train_toolcall_data(self):
         # Get the dataset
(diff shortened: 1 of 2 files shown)


tests/test_sft_trainer.py
```diff
@@ -1170,7 +1170,7 @@ def test_train_with_set_chat_template_from_path(self, lazy_shared_datadir):
 
     def test_train_toolcall_data(self):
         # Get the dataset
-        dataset = load_dataset("trl-internal-testing/toolcall", split="train")
+        dataset = load_dataset("trl-internal-testing/toolcall", "language_modeling", split="train")
```
qgallouedec, line 1173: `messages` used to be the only config

PR #4991, opened 2026-02-06
Assert chat_template is applied in test_train_with_chat_template_kwargs

albertvillanova:
Assert chat_template is applied in test_train_with_chat_template_kwargs.

This PR adds additional assertions to the test suites for both the reward trainer and SFT trainer. These assertions verify that the chat template is correctly applied to the training data, improving test coverage and ensuring template consistency.

The presence of this assertion would have avoided the issue fixed by:
- https://github.com/huggingface/trl/pull/4971#pullrequestreview-3755350437

Test improvements:

* Added assertions in `test_train_with_chat_template_kwargs` in `tests/test_reward_trainer.py` to check that the chat template is properly applied to the training dataset by comparing the start of each input sequence to the expected system prompt tokens.
* Added similar assertions in `test_train_with_chat_template_kwargs` in `tests/test_sft_trainer.py` to verify correct application of the chat template in the SFT trainer's training dataset.

Follow-up to:
- https://github.com/huggingface/trl/pull/4971#pullrequestreview-3755350437

### DIFF
--- a/tests/test_reward_trainer.py
+++ b/tests/test_reward_trainer.py
@@ -507,7 +507,8 @@ class TestRewardTrainer(TrlTestCase):
                 f"<|im_start|>{role}\nYou are Qwen, created by Alibaba Cloud. You are a helpful assistant.<|im_end|>"
             )
             system_prompt_ids = trainer.processing_class(system_prompt)["input_ids"]
-            assert trainer.train_dataset[i]["input_ids"][: len(system_prompt_ids)] == system_prompt_ids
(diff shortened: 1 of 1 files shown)


qgallouedec:
Ok if we test this, what do you think of removing `trainer.train()`?

albertvillanova:
I thought the aim was to test that the trainer can train in this specific setup. But if you think it does not make sense...

qgallouedec (approved):
I'm fine with both feel free to keep or remove

albertvillanova:
Keep it for the moment as a minimal intervention.

PR #5001, opened 2026-02-08
Add `sanitize_logprob` function for NaN handling in vLLM log probabilities

qgallouedec:
I don't know why we didn't apply https://github.com/huggingface/trl/pull/3867 to colocate as well. The PR does this

Note that unsloth does that: https://github.com/unslothai/unsloth/blob/0e90cd026a9f8835a55364ce293098abc442215d/unsloth/models/rl_replacements.py#L345-L352

### DIFF
--- a/trl/generation/vllm_generation.py
+++ b/trl/generation/vllm_generation.py
@@ -48,7 +48,6 @@ def sanitize_logprob(logprob):
     return value
 
 
-
 if TYPE_CHECKING:
     from accelerate import Accelerator
     from peft import PeftModel


trl/generation/vllm_generation.py
```diff
@@ -34,6 +36,20 @@
...
+
+def sanitize_logprob(logprob):
+    value = logprob.logprob
+    if math.isnan(value):
+        logger.warning(f"Generated NaN logprob, token logprob '{logprob}' will be ignored")
+        return None
+
+    return value
+
+    return value
```
LeonEricsson:
```suggestion
```

trl/generation/vllm_generation.py
```diff
@@ -34,6 +36,20 @@
...
+
+
```
kashif:
```suggestion


```

trl/generation/vllm_generation.py
```diff
@@ -34,6 +36,19 @@
...
+def sanitize_logprob(logprob):
+    value = logprob.logprob
+    if math.isnan(value):
+        logger.warning(f"Generated NaN logprob, token logprob '{logprob}' will be ignored")
+        return None
+
+    return value
+
+
+
```
kashif:
```suggestion
    return value


```

LeonEricsson (commented):
good catch. slipped through the cracks. 

lgtm

PR #5015, opened 2026-02-09
Fix CI FutureWarning: tools is deprecated

### LINKED ISSUE #5014: CI FutureWarning: `tools` is deprecated
CI emits FutureWarning: `tools` is deprecated: https://github.com/huggingface/trl/actions/runs/21804947876/job/62906649820
> FutureWarning: `tools` is deprecated and will be removed in version 0.29.0. In 0.29 this argument will be ignored; tools should be provided via the dataset instead but for now, `DPOConfig.tools` remains the only supported way to pass tools.

```python
  tests/test_dpo_trainer.py::TestDPOTrainer::test_dpo_trainer_with_tools
    /__w/trl/trl/tests/test_dpo_trainer.py:1062: FutureWarning: `tools` is deprecated and will be removed in version 0.29.0. In 0.29 this argument will be ignored; tools should be provided via the dataset instead but for now, `DPOConfig.tools` remains the only supported way to pass tools.
      training_args = DPOConfig(
```

albertvillanova:
Fix CI FutureWarning: tools is deprecated:
- Stop testing DPOTrainer with tools

Fix #5014.

Follow-up to:
- #4969

### Rationale
- Our test suite should exercise the current supported public API, so we need to update the tests to the new usage.
- This keeps CI noise-free and ensures we’re testing what users should do now.

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1346,6 +1346,10 @@ class TestSFTTrainer(TrlTestCase):
                         Version(transformers.__version__) < Version("4.57.0"),
                         reason="Qwen3-VL series were introduced in transformers-4.57.0",
                     ),
+                    pytest.mark.xfail(
+                        Version("5.0.0") <= Version(transformers.__version__) < Version("5.1.0"),
+                        reason="Upstream transformers bug (transformers#43334) in 5.0.x; fixed in 5.1.0",
+                    ),
                 ],
             ),
         ],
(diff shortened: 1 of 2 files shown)


qgallouedec:
for this one, it's a bit particular: `DPOCOnfig.tools` is this only valid (yet deprecated) way to train with tools until 0.29. I'd keep this test, and just filter the warning

albertvillanova:
In that case, yes, I totally agree.

PR #5017, opened 2026-02-09
Fix CI FutureWarning: generate_during_eval is deprecated

### LINKED ISSUE #5016: CI FutureWarning: `generate_during_eval` is deprecated
CI emits FutureWarning: `generate_during_eval` is deprecated:
> FutureWarning: `generate_during_eval` is deprecated and will be removed in version 0.29.0. Please use a callback instead. See the example at `https://gist.github.com/qgallouedec/a08da3457a3a76c5ca539d4a0b38e482`.

```python
   tests/test_dpo_trainer.py::TestDPOTrainer::test_dpo_trainer_generate_during_eval_no_wandb
    /__w/trl/trl/tests/test_dpo_trainer.py:564: FutureWarning: `generate_during_eval` is deprecated and will be removed in version 0.29.0. Please use a callback instead. See the example at `https://gist.github.com/qgallouedec/a08da3457a3a76c5ca539d4a0b38e482`.
      training_args = DPOConfig(
```

albertvillanova:
Fix CI FutureWarning: generate_during_eval is deprecated:
- Stop testing generate_during_eval in DPO

Fix #5016.

Follow-up to:
- #4969

### Rationale
- Our test suite should exercise the current supported public API, so we need to update the tests to the new usage.
- This keeps CI noise-free and ensures we’re testing what users should do now.

### DIFF
--- a/tests/test_dpo_trainer.py
+++ b/tests/test_dpo_trainer.py
@@ -40,7 +40,6 @@ from .testing_utils import (
     TrlTestCase,
     require_bitsandbytes,
     require_liger_kernel,
-    require_no_wandb,
     require_peft,
     require_torch_accelerator,
     require_torch_gpu_if_bnb_not_multi_backend_enabled,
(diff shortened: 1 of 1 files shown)


qgallouedec:
What do you think of removing these tests when the DPO actually drops the support? We could just filter the warnings in the CI

albertvillanova:
@qgallouedec, I removed the `test_dpo_trainer_generate_during_eval_no_wandb`, and removed the `generate_during_eval` parameter from `test_dpo_lora_bf16_autocast`.

Which other tests are you suggesting to remove?

albertvillanova:
Oh, I understand, @qgallouedec! I tried to give my reasoning in the Rationale above, but let me explain it better.

Normally, in OSS projects, the deprecation cycle follows a pattern:
1. First, deprecate the params
   - Stop using the deprecated param in “normal” tests: Our test suite should exercise the supported public API. So we need to update the test to the new usage. This keeps CI noise-free and ensures we’re testing what users should do now.
2. After the deprecation cycle, remove the params: 
   - Optionally, add a test asserting the param now errors

This is the standard “best of both worlds”: clean CI, correct user-facing guidance.

qgallouedec:
> Stop using the deprecated param in “normal” tests: Our test suite should exercise the supported public API. So we need to update the test to the new usage. This keeps CI noise-free and ensures we’re testing what users should do now.

Ok sounds good!
In more critical cases, I'd try to ensure that the deprecated path is remains working until removal with a dedicated tst+ignore warning, but for here I guess it's not worth it.

qgallouedec (approved):

PR #5019, opened 2026-02-09
Fix CI FutureWarning: max_prompt_length is deprecated

albertvillanova:
Fix CI FutureWarning: max_prompt_length is deprecated:
- Stop using max_prompt_length in DPO tests

Fix #5018.

Follow-up to:
- #4969

### Rationale
- Our test suite should exercise the current supported public API, so we need to update the tests to the new usage.
- This keeps CI noise-free and ensures we’re testing what users should do now.

PR #5023, opened 2026-02-09
Fix vision model prompt truncation bug in DPOTrainer

albertvillanova:
This PR fixes a bug in `DPOTrainer.process_row()` where prompt truncation for vision models would cause image token/feature mismatches, resulting in errors during training.

### Problem

The `process_row` method (used exclusively for vision models) was truncating `prompt_input_ids` when `max_prompt_length` was set. However, for vision models, the prompt contains image tokens that must exactly match the corresponding image features (`pixel_values`). Truncating these tokens breaks this invariant, causing errors like: https://github.com/huggingface/trl/actions/runs/21820438552/job/62952103518?pr=5019
```python
ValueError: Image features and image tokens do not match: tokens: 1998, features 5184
```

This bug became apparent during the deprecation of `max_prompt_length`:
- #4969
- #5019

Previously, vision tests explicitly set `max_prompt_length=None` to avoid truncation. When this workaround was removed (since the parameter is deprecated), the bug surfaced due to the default value of 512.

### Solution

Remove prompt truncation from `process_row` entirely. Vision models fundamentally cannot support prompt truncation without breaking the image token/feature correspondence. Users should filter overlong prompts from their datasets before training (which is the recommended approach for the deprecated `max_prompt_length` parameter).

Note: This change only affects vision models. Language models use `tokenize_row`, which retains truncation logic during the deprecation period.

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -787,7 +787,6 @@ class DPOTrainer(BaseTrainer):
             warnings.warn(
                 "max_prompt_length is not supported for vision models and will be ignored. "
                 "Truncating prompts would cause image token/feature mismatch errors.",
-                UserWarning,
                 stacklevel=2,
             )
         processor, tokenizer = processing_class, processing_class.tokenizer  # the processing class is a processor


qgallouedec:
It's not a big deal because this param is deprecated anyway, but if a user passes max_prompt_length along with a vlm, they won't know that the trainer actually ignores it. Should we invite the user to explicitly drop this argument in this case?

albertvillanova:
We could emit a UserWarning.

PR #4919, opened 2026-01-28
Fix CI TypeError in llm-blender tests

albertvillanova:
Fix CI TypeError from llm-blender tests by marking them as xfail.

See tracking issue:
- #4918

Hotfix and close #4917.

PR #5024, opened 2026-02-09
Pin transformers < 5 in judges extra due to incompatibility

albertvillanova:
Pin transformers < 5 in judges extra due to incompatibility.

Follow-up to:
- #4918

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,8 @@ deepspeed = [
 ]
 judges = [
     "openai>=1.23.2",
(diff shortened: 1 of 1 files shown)


albertvillanova:
What I was thinking:
- A user installing `trl[judges]` will have the right pin in tansformers
- Our CI (using `trl[dev]`) does not strictly pin transformers version, so we can test any of them
  - However, the tests are protected as xfail, see:
    - #4919

qgallouedec:
that makes perfect sense

PR #5029, opened 2026-02-09
Mark Qwen3VL tests as xfail for transformers 5.0.x

albertvillanova:
Mark Qwen3VL tests as xfail for transformers 5.0.x.

After having thought more carefully about the previous removal, I think the CI is more robust with this PR:
- What if we run the tests with transformers 5.0.x?
- What if we eventually set transformers-5.0.0 as the minimum required version?

Follow-up to:
- #5000

This PR adds test markers to handle a known upstream bug in the `transformers` library version 5.0.x, ensuring that affected tests are expected to fail.

Test robustness improvements:

* Added `pytest.mark.xfail` to relevant tests in `tests/test_sft_trainer.py` and `tests/test_utils.py` to account for the upstream bug in `transformers` 5.0.x, referencing issue `transformers#43334`.

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1346,6 +1346,10 @@ class TestSFTTrainer(TrlTestCase):
                         Version(transformers.__version__) < Version("4.57.0"),
                         reason="Qwen3-VL series were introduced in transformers-4.57.0",
                     ),
+                    pytest.mark.xfail(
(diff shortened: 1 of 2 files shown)


qgallouedec:
That makes sense

> What if we eventually set transformers-5.0.0 as the minimum required version?

Yes, and we will at some point

PR #5031, opened 2026-02-09
Pass vllm_is_ratio to LigerFusedLinearGRPOLoss in compute_liger_loss

yukiu00:
# What does this PR do?

When using `use_vllm=True` with `use_liger_kernel=True`, the vLLM importance sampling correction (`importance_sampling_ratio`) is computed in `_generate_and_score_completions` and stored in the inputs dict, but **never passed** to the Liger GRPO loss in `compute_liger_loss`. This causes ~100x larger `grad_norm` compared to the non-Liger path, leading to training instability or divergence.

**Root cause:** The standard `_compute_loss` path applies `per_token_loss *= inputs["importance_sampling_ratio"]` (line 2056), but `compute_liger_loss` does not pass this value to `self.liger_grpo_loss(...)`.

**Fix:** Add `vllm_is_ratio=inputs.get("importance_sampling_ratio")` to the `self.liger_grpo_loss(...)` call in `compute_liger_loss`. The Liger-Kernel side supports this parameter (added in [linkedin/Liger-Kernel#1088](https://github.com/linkedin/Liger-Kernel/pull/1088)), applying the correction after PPO clipped loss and before KL penalty — matching TRL's standard path exactly.

**This affects all users running `GRPOTrainer` with `use_vllm=True` and `use_liger_kernel=True`.**

Fixes https://github.com/linkedin/Liger-Kernel/issues/1082

## Changes

- **`trl/trainer/grpo_trainer.py`**: Add `vllm_is_ratio=inputs.get("importance_sampling_ratio")` to the `self.liger_grpo_loss(...)` call in `compute_liger_loss` (1-line change)
- **`tests/test_grpo_trainer.py`**: Add `test_compute_liger_loss_passes_vllm_is_ratio` that injects `importance_sampling_ratio` into the training pipeline and verifies it is passed through to the Liger loss as `vllm_is_ratio`. Guarded with `xfail` until the upstream Liger-Kernel release includes the `vllm_is_ratio` parameter.

## Dependencies

- Requires [linkedin/Liger-Kernel#1088](https://github.com/linkedin/Liger-Kernel/pull/1088) to be merged and released (adds `vllm_is_ratio` parameter to `LigerFusedLinearGRPOLoss`)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@qgallouedec @kashif


tests/test_grpo_trainer.py
```diff
@@ -58,6 +58,18 @@
...
+    try:
+        import inspect
+
+        from liger_kernel.chunked_loss import LigerFusedLinearGRPOLoss
+
+        return "vllm_is_ratio" in inspect.signature(LigerFusedLinearGRPOLoss.forward).parameters
+    except Exception:
+        return False
+
+
```
qgallouedec: I prefer an explicit version check than a signature inspection, it's easier to maintain and more explicit. In which version this param is introduced?

tests/test_grpo_trainer.py
```diff
@@ -988,6 +1000,68 @@ def test_training_with_off_policy_mask_with_liger(self):
...
+            model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
+            reward_funcs="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
+            args=training_args,
+            train_dataset=dataset,
+        )
+
+        # Wrap _generate_and_score_completions to inject importance_sampling_ratio
+        original_gen = trainer._generate_and_score_completions
+
+        def gen_with_is_ratio(*args, **kwargs):
```
qgallouedec, line 1018: is it possible to use mocking instead? Like in https://github.com/huggingface/trl/blob/db0d95523e5b8039c94c50df1b6286ff8b7e29ce/tests/test_grpo_trainer.py#L1350-L1362

qgallouedec (commented):
thanks! I have a few initial questions/suggestions
cc @kashif as it involves liger

tests/test_grpo_trainer.py
yukiu00: @qgallouedec 
I've replaced the signature inspection with an explicit version check using `is_liger_kernel_available(min_version="0.6.6")`, following the existing pattern in the codebase (e.g., [line 1907](https://github.com/huggingface/trl/blob/2434ca7a/tests/test_grpo_trainer.py#L1907)).

The upstream PR ([linkedin/Liger-Kernel#1088](https://github.com/linkedin/Liger-Kernel/pull/1088)) that adds the `vllm_is_ratio` parameter hasn't been merged/released yet (latest release is v0.6.5), so I'm using `"0.6.6"` as a provisional version for now. I think we should wait until the upstream PR is merged and released, then update to the correct version and merge this PR. I'll update the version number as soon as the release is out.

tests/test_grpo_trainer.py
yukiu00: I've refactored the test to use `patch.object` context managers instead of manual monkey-patching.

kashif:
@yukiu00 as soon as they make a new release, we should also pin to that version as minimum

yukiu00:
@qgallouedec
liger-kernel v0.7.0 has been released (https://github.com/linkedin/Liger-Kernel/releases/tag/v0.7.0) with the `vllm_is_ratio` support from linkedin/Liger-Kernel#1088.

I've updated the PR to:
- Bump `LIGER_KERNEL_MIN_VERSION` to `"0.7.0"` in `trl/import_utils.py`
- Pin `liger-kernel>=0.7.0` in `pyproject.toml` (both `[liger]` and `[dev]` extras)
- Remove the provisional `xfail` marker on `test_compute_liger_loss_passes_vllm_is_ratio` (no longer needed since v0.7.0 is the minimum)

@kashif This addresses your request to pin to the released version as minimum.

yukiu00:
Hi @qgallouedec @kashif,

Just a gentle bump on this. Since the dependency issue is now resolved, I believe the PR is ready for the final review.

Could you please let me know if there is anything else needed? If possible, I would also appreciate a rough estimate of when this might be reviewed or merged.

Thanks!

kashif:
thanks @yukiu00 checking and reviewing now

kashif (approved):

PR #5038, opened 2026-02-09
Rename AOT loss type 'aot_pair' to 'aot_unpaired' in DPO

qgallouedec:
See https://github.com/huggingface/trl/pull/3906#user-content-3

### DIFF
--- a/trl/trainer/dpo_config.py
+++ b/trl/trainer/dpo_config.py
@@ -742,8 +742,8 @@ class DPOConfig(TrainingArguments):
 
         if "aot_pair" in self.loss_type:
             warnings.warn(
-                "`loss_type='aot_pair'` is deprecated and will be removed in a version 0.29.0. Please use "
-                "`loss_type='aot_unpaired'` instead.",
+                "The loss type 'aot_pair' has been renamed to 'aot_unpaired' and is deprecated. "
+                "It will be removed in version 0.29.0. Please use 'aot_unpaired' in `loss_type` instead.",
                 FutureWarning,
                 stacklevel=3,
             )


albertvillanova (approved):
Thanks.

trl/trainer/dpo_config.py
```diff
@@ -740,4 +740,13 @@ def __post_init__(self):
...
 
+        if "aot_pair" in self.loss_type:
+            warnings.warn(
+                "`loss_type='aot_pair'` is deprecated and will be removed in a version 0.29.0. Please use "
+                "`loss_type='aot_unpaired'` instead.",
+                FutureWarning,
+                stacklevel=3,
+            )
+            self.loss_type = ["aot_unpaired" if lt == "aot_pair" else lt for lt in self.loss_type]
+
```
albertvillanova, line 754: I just realized that some lines above `loss_type` was transformed to `str` if it contains a single element. 
```python
        # Normalize loss_type to string format for internal use
        if hasattr(self.loss_type, "__len__") and len(self.loss_type) == 1:
            self.loss_type = self.loss_type[0]
```

I guess this code line won't work as expected in that case. I'm wondering how it is that tests are passing...

albertvillanova (changes requested):

trl/trainer/dpo_config.py
albertvillanova: Also I think the renaming will not be applied if loss_type is passed as str, instead of list.

albertvillanova (approved):
I think I fixed the previous issues.

PR #5039, opened 2026-02-10
Deprecate FDivergenceType in DPOConfig; update f_divergence_type to use string values

qgallouedec:
See https://github.com/huggingface/trl/pull/3906#user-content-11


albertvillanova (approved):
Thanks. I fixed the handling of `f_divergence_type` as str internally and updated the failing tests accordingly. Additionally, I updated the docstring specifying the supported values, and added choices to its metadata.

PR #5054, opened 2026-02-10
Remove duplicated tests for SFT and add gradient checkpointing tests

qgallouedec:
Removing duplicated tests (both in `TestSFTTrainerSlow` and `TestSFTTrainer`)


tests/test_sft_trainer.py
```diff
@@ -1696,248 +1769,6 @@ def teardown_method(self):
...
 
-    @pytest.mark.parametrize("packing", [True, False])
-    @pytest.mark.parametrize(
-        "model_name",
-        [
-            "trl-internal-testing/tiny-LlamaForCausalLM-3.2",
-            "trl-internal-testing/tiny-MistralForCausalLM-0.2",
-        ],
-    )
-    def test_sft_trainer_str(self, model_name, packing):
```
qgallouedec, line 1707: this test is mostly the same as `test_train_packing`

tests/test_sft_trainer.py
```diff
@@ -1696,248 +1769,6 @@ def teardown_method(self):
...
-
-    @pytest.mark.parametrize("packing", [True, False])
-    @pytest.mark.parametrize(
-        "model_name",
-        [
-            "trl-internal-testing/tiny-LlamaForCausalLM-3.2",
-            "trl-internal-testing/tiny-MistralForCausalLM-0.2",
-        ],
-    )
-    def test_sft_trainer_transformers(self, model_name, packing):
```
qgallouedec, line 1738: I’m not aware of any documented special cases that would justify a test combining packing with passing an instantiated model, so this test should be removed.

tests/test_sft_trainer.py
```diff
@@ -1696,248 +1769,6 @@ def teardown_method(self):
...
-    @pytest.mark.parametrize("packing", [True, False])
-    @pytest.mark.parametrize(
-        "model_name",
-        [
-            "trl-internal-testing/tiny-LlamaForCausalLM-3.2",
-            "trl-internal-testing/tiny-MistralForCausalLM-0.2",
-        ],
-    )
-    @require_peft
-    def test_sft_trainer_peft(self, model_name, packing):
```
qgallouedec, line 1776: same as `test_train_dense_with_peft_config_lora` (except that here we pass the instantiated model)

tests/test_sft_trainer.py
```diff
@@ -1696,248 +1769,6 @@ def teardown_method(self):
...
-    )
-    @pytest.mark.parametrize("packing", [True, False])
-    @pytest.mark.parametrize(
-        "model_name",
-        [
-            "trl-internal-testing/tiny-LlamaForCausalLM-3.2",
-            "trl-internal-testing/tiny-MistralForCausalLM-0.2",
-        ],
-    )
-    def test_sft_trainer_transformers_mp_gc(self, model_name, packing, gradient_checkpointing_kwargs):
```
qgallouedec, line 1860: we already test this in `test_train_with_gradient_checkpointing` (the only difference is bf16 vs fp16)

tests/test_sft_trainer.py
```diff
@@ -1696,248 +1769,6 @@ def teardown_method(self):
...
-    @pytest.mark.parametrize("packing", [True, False])
-    @pytest.mark.parametrize(
-        "model_name",
-        [
-            "trl-internal-testing/tiny-LlamaForCausalLM-3.2",
-            "trl-internal-testing/tiny-MistralForCausalLM-0.2",
-        ],
-    )
-    @require_peft
-    def test_sft_trainer_transformers_mp_gc_peft(self, model_name, packing, gradient_checkpointing_kwargs):
```
qgallouedec, line 1905: we will now test this in `test_train_with_peft_config_and_gradient_checkpointing`

plus, there is no reason to combine with packing

PR #5059, opened 2026-02-10
fix: Use `launch_args` for all trainers

qgallouedec:
I realized that `launch_args` was only properly used for SFT, and ignore for other trainers. This PR fixes it.

### DIFF
--- a/trl/cli.py
+++ b/trl/cli.py
@@ -34,6 +34,12 @@ from .scripts.vllm_serve import make_parser as make_vllm_serve_parser
 logger = logging.getLogger(__name__)
 
 
+def _launch_training_script(script_name: str, launch_args: list[str], training_script_args: list[str]) -> None:
+    training_script = resources.files("trl.scripts").joinpath(script_name)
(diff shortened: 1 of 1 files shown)


trl/cli.py
```diff
@@ -83,63 +89,31 @@ def main():
...
 
     elif args.command == "sft":
-        # Get the path to the training script
-        sft_training_script = resources.files("trl.scripts").joinpath("sft.py")
-
         # This simulates running: `accelerate launch <launch args> sft.py <training script args>`.
-        # Note that the training script args may include launch-related arguments (e.g., `--num_processes`),
-        # but we rely on the script to ignore any that don't apply to it.
-        training_script_args = sys.argv[2:]  # Remove "trl" and "sft"
-        args = launch_command_parser().parse_args(launch_args + [str(sft_training_script)] + training_script_args)
```
qgallouedec, line 141: here, `launch_args` is correctly used, but elsewhere it's not the case.

PR #5060, opened 2026-02-10
Update model from SequenceClassification to CausalLM in `RewardTrainer` tests

qgallouedec:
Although this is fully supported, retraining a reward model as a SequenceClassification model is not very common. Most users train reward models from a causal LM, so I suggest we reflect this in our tests.

### DIFF
--- a/trl/trainer/reward_trainer.py
+++ b/trl/trainer/reward_trainer.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 import contextlib
+import logging
 import os
+import re
 from collections import defaultdict
 from collections.abc import Callable
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


qgallouedec:
requires #5066

tests/test_reward_trainer.py
```diff
@@ -200,14 +200,16 @@ def test_train_model(self):
...
         dataset = load_dataset("trl-internal-testing/zen", "standard_implicit_prompt_preference", split="train")
 
         # Initialize the trainer
         training_args = RewardConfig(output_dir=self.tmp_dir, report_to="none")
         trainer = RewardTrainer(
-            model="trl-internal-testing/tiny-Qwen3ForCausalLM", args=training_args, train_dataset=dataset
+            model="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
+            args=training_args,
+            train_dataset=dataset,
         )
```
qgallouedec, line 213: re https://github.com/huggingface/trl/pull/5060#pullrequestreview-3783866960

we have this test, I think for now it's enough but we can still add more

Issue #4407, opened 2025-10-30
Complete paper index


qgallouedec:
These are the papers mentioned at least one in the codebase.

- [ ] https://huggingface.co/papers/1707.06347
- [x] https://huggingface.co/papers/1909.08593 (only mentioned in notebook, no need to have in paper index)
- [ ] https://huggingface.co/papers/1910.02054
- [ ] https://huggingface.co/papers/1910.10683
- [ ] https://huggingface.co/papers/2106.09685
- [ ] https://huggingface.co/papers/2211.14275
- [ ] https://huggingface.co/papers/2305.10425
- [ ] https://huggingface.co/papers/2305.18290
- [ ] https://huggingface.co/papers/2306.13649
- [ ] https://huggingface.co/papers/2307.09288
- [ ] https://huggingface.co/papers/2309.06657
- [ ] https://huggingface.co/papers/2310.12036
- [ ] https://huggingface.co/papers/2312.00886
- [ ] https://huggingface.co/papers/2312.09244
- [ ] https://huggingface.co/papers/2401.08417
- [ ] https://huggingface.co/papers/2402.00856
- [ ] https://huggingface.co/papers/2402.01306
- [ ] https://huggingface.co/papers/2402.03300
- [ ] https://huggingface.co/papers/2402.04792
- [ ] https://huggingface.co/papers/2402.05369
- [ ] https://huggingface.co/papers/2402.09353
- [ ] https://huggingface.co/papers/2402.14740
- [ ] https://huggingface.co/papers/2403.00409
- [ ] https://huggingface.co/papers/2403.07691
- [ ] https://huggingface.co/papers/2403.17031
- [ ] https://huggingface.co/papers/2404.04656
- [ ] https://huggingface.co/papers/2404.09656
- [ ] https://huggingface.co/papers/2404.19733
- [ ] https://huggingface.co/papers/2405.00675
- [ ] https://huggingface.co/papers/2405.14734
- [ ] https://huggingface.co/papers/2405.16436
- [ ] https://huggingface.co/papers/2405.21046
- [ ] https://huggingface.co/papers/2406.05882
- [ ] https://huggingface.co/papers/2406.08414
- [ ] https://huggingface.co/papers/2406.11827
- [ ] https://huggingface.co/papers/2407.21783
- [ ] https://huggingface.co/papers/2408.06266
- [ ] https://huggingface.co/papers/2409.06411
- [ ] https://huggingface.co/papers/2409.20370
[...]

PR #5062, opened 2026-02-10
docs: Add REINFORCE++ (2501.03262) to Paper Index

behroozazarkhalili:
## Summary
- Added REINFORCE++ (2501.03262) to the RLOO section of the Paper Index
- Included `RLOOConfig` with `normalize_advantages=True` (global advantage normalization) and `num_generations=4` (k=4 from Appendix B.2)

Part of #4407

## Test plan
- [ ] Verify the paper link resolves correctly
- [ ] Verify all config parameters exist in `RLOOConfig`
- [ ] Verify formatting matches existing paper index entries

### DIFF
--- a/docs/source/paper_index.md
+++ b/docs/source/paper_index.md
@@ -34,6 +34,47 @@ trainer = GRPOTrainer(
 )
 ```
 
+### DeepSeek-R1: Incentivizing Reasoning Capability in LLMs via Reinforcement Learning
+
+**📜 Paper**: https://huggingface.co/papers/2501.12948
+
+DeepSeek-R1 achieves reasoning performance comparable to OpenAI-o1 through a multi-stage pipeline that transitions from pure reinforcement learning (RL) to a refined, human-aligned model. Unlike its predecessor, DeepSeek-R1-Zero, which used pure RL on a base model, R1 follows a structured four-stage evolution:
+1. Cold Start: The base model is fine-tuned on a small set of high-quality, long Chain-of-Thought (CoT) data to provide a stable starting point.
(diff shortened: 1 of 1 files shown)


docs/source/paper_index.md
```diff
@@ -925,6 +925,21 @@ training_args = RLOOConfig(
...
+**📜 Paper**: https://huggingface.co/papers/2501.03262
+
+REINFORCE++ is an enhanced variant of the classical REINFORCE algorithm that incorporates key optimization techniques from PPO while eliminating the need for a critic network. It achieves simplicity, enhanced training stability, and reduced computational overhead through global advantage normalization across the entire batch. To approximate the paper's setting with the [`RLOOTrainer`], use this configuration:
+
+```python
+from trl import RLOOConfig
+
+training_args = RLOOConfig(
+    normalize_advantages=True,  # global advantage normalization, core of REINFORCE++
+    num_generations=4,  # k=4 in Appendix B.2 of the paper for reasoning tasks
```
qgallouedec: I think that in the main variant, they use k=1, no?

behroozazarkhalili: You're right — the main REINFORCE++ variant uses k=1 (single sample per prompt). The k=4 value was specific to the Appendix B.2 reasoning task experiments. The paper's abstract confirms: "REINFORCE++, a highly efficient and general algorithm (k ≥ 1)".

Removed the `num_generations` parameter since we can't set k=1 anyway — `RLOOTrainer` requires `num_generations >= 2` for its leave-one-out baseline calculation. Keeping only `normalize_advantages=True` which is the core contribution of the algorithm.

qgallouedec: interesting, maybe we should work in the future on allowing num_generations=1 to fairly mirror REINFORCE++

PR #5066, opened 2026-02-10
fix: Set `num_labels` to 1 in causal model initialization for RewardTrainer

qgallouedec:
This PR fixes a bug in `RewardTrainer` introduced in #4577: when it is initialized from a `CausalLM`, it defaults to `num_labels=2`, whereas a reward model should have a single output.

This issue has gone unnoticed because:

* most tests do not initialize the trainer from a causal LM (even though this is the most common usage in practice; see #5060);
* the `RewardTrainer` implementation does not enforce a single output dimension (except for the optional margin logic), so having a second output dimension does not raise an error, see
https://github.com/huggingface/trl/blob/d934eb757806501a5106b6e4374d920961dc4e9f/trl/trainer/reward_trainer.py#L575-L609

### DIFF
--- a/trl/trainer/reward_trainer.py
+++ b/trl/trainer/reward_trainer.py
@@ -13,9 +13,7 @@
 # limitations under the License.
 
 import contextlib
-import logging
 import os
-import re
 from collections import defaultdict
 from collections.abc import Callable
 from contextlib import contextmanager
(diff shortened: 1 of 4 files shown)


trl/trainer/reward_trainer.py
```diff
@@ -310,6 +310,7 @@ def __init__(
             # Distributed training requires device_map=None ("auto" fails)
             if args.distributed_state.distributed_type in ["MULTI_GPU", "DEEPSPEED"]:
                 model_init_kwargs["device_map"] = None
+            model_init_kwargs["num_labels"] = 1
```
albertvillanova: Maybe adding a comment to explain why?

qgallouedec:
I cannot reproduce the CI failure with min versions locally. I'll merge

PR #5067, opened 2026-02-10
Add validation for conversational prompts in multimodal training

qgallouedec:
There is confusion around whether data should be pre-processed before being passed to `GRPOTrainer`. This adds a clear, actionable error message instead of the cryptic

```
TypeError: string indices must be integers, not 'str'.
```

See https://github.com/huggingface/trl/pull/5064

Closes #4870
Closes #4746
Closes #4451
Closes #5041

Issue #4451, opened 2025-11-05
prepare_multimodal_messages's bug in _generate_and_score_completions


wswaq:
I am working in the main and using the grpo_qwen3_vl.ipynb example.
I find current main does not match this example.Because this example process dataset and return
```python
def make_conversation(example):
    conversation = [
        {
            "role": "system",
            "content": [{"type": "text", "text": SYSTEM_PROMPT}],
        },
        {
            "role": "user",
            "content": [
                {"type": "image", "image": example["image"]},
                {"type": "text", "text": example["problem"]},
            ],
        },
    ]
    prompt = processor.apply_chat_template(conversation, add_generation_prompt=True)
    return {
        "prompt": prompt,
        "image": example["image"],
    }
```
so here prompt is already a text
but in `_generate_and_score_completions`
```python
        if images is not None:
            prompts = [prepare_multimodal_messages(prompt, image_list) for prompt, image_list in zip(prompts, images)]

```
here ,the prompts are treated as conversational, so some problems will happen
```shell
[rank0]:   trainer_stats = trainer.train()
[rank0]:   File ".../transformers/trainer.py", line 2316, in train
[rank0]:     return inner_training_loop(
[rank0]:   File ".../transformers/trainer.py", line 2674, in _inner_training_loop
[rank0]:     tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
[rank0]:   File ".../transformers/trainer.py", line 4014, in training_step
[rank0]:     inputs = self._prepare_inputs(inputs)
[rank0]:   File ".../trl/extras/profiling.py", line 98, in wrapper
[rank0]:     return func(self, *args, **kwargs)
[rank0]:   File ".../trl/trainer/grpo_trainer.py", line 1064, in _prepare_inputs
[rank0]:     generation_batch = self._generate_and_score_completions(generation_batch)
[rank0]:   File ".../trl/trainer/grpo_trainer.py", line 1548, in _generate_and_score_completions
[rank0]:     prompts = [prepare_multimodal_messages(prompt, image_list) for prompt, image_list in zip(prompts, images)]
[rank0]:   File ".../trl/trainer/grpo_trainer.py", line 1548, in <listcomp>
[rank0]:     prompts = [prepare_multimodal_messages(prompt, image_list) for prompt, image_list in zip(prompts, images)]
[rank0]:   File ".../trl/data_utils.py", line 78, in prepare_multimodal_messages
[rank0]:     if message["role"] == "system":
[...]

PR #5073, opened 2026-02-11
Cast multimodal forward_kwargs to compute dtype for bf16/fp16 training

akshan-main:
# What does this PR do?

When training VLMs with bf16=True or fp16=True, pixel_values returned by the processor stay float32 after _prepare_inputs (dtype casting is DeepSpeed-specific).
If the vision encoder weights are bfloat16/float16, this can crash in torch.layer_norm with:
```
RuntimeError: expected scalar type BFloat16 but found Float
```
This is the next failure reported in the #4451 thread after the prompt-format TypeError.

**Note:** Fixed by casting floating-point tensors in forward_kwargs to the compute dtype when bf16=True or fp16=True. This is consistent with how the trainer already handles model dtype casting. If the model is loaded in bf16 via torch_dtype without setting the training flag, this path won't trigger, but neither does the existing model casting.

# Changes made
In the multimodal path, cast only floating-point tensors in forward_kwargs to the active compute dtype (bf16/fp16).
Leave non-floating tensors (for example image_grid_thw) unchanged.

No prompt-format behavior changes (proposed in #5064 and #5067)
No reward-function behavior changes (proposed in #5064)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@qgallouedec @kashif @albertvillanova

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1609,6 +1609,17 @@ class GRPOTrainer(BaseTrainer):
             prompt_inputs = self.processing_class(images=images, text=prompts_text, padding=True, return_tensors="pt")
             prompt_inputs = super()._prepare_inputs(prompt_inputs)
             forward_kwargs = {k: v for k, v in prompt_inputs.items() if k not in ["input_ids", "attention_mask"]}
+            if self.args.bf16:
+                compute_dtype = torch.bfloat16
+            elif self.args.fp16:
+                compute_dtype = torch.float16
+            else:
+                compute_dtype = None
+            if compute_dtype is not None:
+                forward_kwargs = {
+                    k: v.to(compute_dtype) if isinstance(v, torch.Tensor) and torch.is_floating_point(v) else v
+                    for k, v in forward_kwargs.items()
+                }
         else:
             forward_kwargs = {}
 


aditya-ai-architect:
Acknowledged, I will review this.

akshan-main:
@aditya-ai-architect  I'm not sure if you have reviewer authorization for this?

akshan-main:
@qgallouedec hey mate! Can this be reviewed? Let me know if you want any changes.

akshan-main:
@codex review

tests/test_grpo_trainer.py
```diff
@@ -2287,6 +2287,39 @@ def test_single_reward_model_with_single_processing_class(self):
...
+        else:
+            result = forward_kwargs
+
+        if expected_dtype is not None:
+            assert result["pixel_values"].dtype == expected_dtype
+        else:
+            assert result["pixel_values"].dtype == torch.float32
+        assert result["image_grid_thw"].dtype == torch.int64
+
+
```
qgallouedec: what are we testing here? just that the `to` works? if so I think it's out of the scope the trl tests

akshan-main: You're right, I'll drop the test.

akshan-main:
Hey since its been a month, I thought I could ping you @qgallouedec

qgallouedec:
Thanks @akshan-main, and sorry for the long delay here, this got buried in the backlog.

I tried to reproduce the `RuntimeError: expected scalar type BFloat16 but found Float` on current `main` (transformers `5.13`), running GRPO with `bf16=True` and the vision weights actually in bf16, across several tiny VLMs (Qwen2.5-VL, Qwen2-VL, LLaVA, Gemma3). None of them crash anymore.

The reason is that `transformers` now casts `pixel_values` to the vision-tower dtype inside the model's own `forward`, so the trainer doesn't need to. For example, in `Qwen2_5_VL`:

```python
pixel_values = pixel_values.type(self.visual.dtype)
vision_outputs = self.visual(pixel_values, grid_thw=image_grid_thw, **kwargs)
```

see https://github.com/huggingface/transformers/blob/63f32a8782cb70da3365acab16f2b67947737985/src/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py#L1105-L1106

(and the patch embedding additionally does `hidden_states.to(dtype=target_dtype)`); the other VLM architectures do the equivalent. So the bug was real when you filed this against the older stack, but it's since been handled upstream.

Given all that, I think the cleanest outcome is to close this one as obsolete. Really appreciate the effort though, and please do keep the fixes coming! 🙌

akshan-main:
sure! You can feel free to tag me in any issues where contributors are welcome as well, and I'd be glad to take that and contribute @qgallouedec

PR #5076, opened 2026-02-11
Fix BFD packing for SFT datasets

### LINKED ISSUE #5075: The changes made to "bfd" packing in v0.27.0 breaks packing for SFT datasets
### Reproduction

This merge request: #4632  Implemented a new behavior for the "bfd" packing strategy, instead of discarding truncated tokens, truncated fragments are re-queued and packed like any other sequence. 

This approach makes sense for pre-training datasets, in which you can truncate large texts into smaller chunks. However, it breaks packing for SFT datasets.

Here is a quick demonstration: 

```python
from datasets import Dataset
from trl import pack_dataset

examples = {
    "input_ids": [["<user>", "How", "are", "you", "?", "<assistant>", "I", "am", "fine", "thank", "you", "!", "<end_of_turn>"],
    ["<user>", "How", "much", "is", "2", "+", "2", "?", "<assistant>", "4", "<end_of_turn>"]],
}
dex = Dataset.from_dict(examples)
dex = pack_dataset(dex, seq_length=10, strategy="bfd")
print(dex[:])
```

v0.26.2
`{'input_ids': [['<user>', 'How', 'are', 'you', '?', '<assistant>', 'I', 'am', 'fine', 'thank'], ['<user>', 'How', 'much', 'is', '2', '+', '2', '?', '<assistant>', '4']], 'seq_lengths': [[10], [10]]}`
v.028.0
`{'input_ids': [['<user>', 'How', 'are', 'you', '?', '<assistant>', 'I', 'am', 'fine', 'thank'], ['<user>', 'How', 'much', 'is', '2', '+', '2', '?', '<assistant>', '4'], ['you', '!', '<end_of_turn>', '<end_of_turn>']], 'seq_lengths': [[10], [10], [3, 1]]}`

As you can see, the new strategy creates a non-sense sequence with the left-over tokens when the dataset contains conversations longer than the packing `seq_length`. For SFT dataset, chunking the token ids and re-queue should not be done. 

The function should include a flag to allow and disable re-queue of truncated chunks. 

### System Info

Copy-paste the following information when reporting an issue:

- Platform: Linux-5.10.245-245.983.amzn2.x86_64-x86_64-with-glibc2.35
- Python version: 3.12.11
- TRL version: 0.28.0
- PyTorch version: 2.10.0
- accelerator(s): cpu
- Transformers version: 5.1.0
- Accelerate version: 1.12.0
- Datasets version: 4.5.0
- HF Hub version: 1.4.1
- OpenAI version: not installed
- PEFT version: not installed
- vLLM version: not installed

### Checklist

albertvillanova:
Fix BFD packing for SFT datasets:
- Revert 'bfd' strategy to old behavior and default to it
- Add new strategy, named 'bfd-requeue', with new behavior: splitting and re-queuing overflow tokens
- Add `requeue_truncated_sequences=False` parameter to `_pack_bfd`

Fix #5075.

### Problem
1. User training on very long code sequences lost 99.8% of tokens due to truncation: massive data loss for pre-training/long sequence scenarios
    - #4632
2. Re-queuing breaks SFT conversational datasets by creating nonsensical sequences.
    - #5075

### Solution

Given that TRL is primarily for post-training (SFT, RLHF, DPO, etc.), I think the default behavior should be discard overflow, don't re-queue.

This PR adds a new option to the `pack_dataset` function to control how sequences longer than the target length are handled when using the "bfd" packing strategy. By introducing the `requeue_truncated_sequences` parameter, users can now choose between truncating long sequences (default, which is SFT-friendly) or splitting and re-queuing overflow tokens (which preserves all tokens for pre-training or long documents). The changes include updates to the implementation, documentation, and tests to support and verify this new behavior.

### DIFF
--- a/docs/source/paper_index.md
+++ b/docs/source/paper_index.md
@@ -970,6 +970,21 @@ SFTConfig(
 )
 ```
 
+### Exploring the Limits of Transfer Learning with a Unified Text-to-Text Transformer
+
+**📜 Paper**: https://huggingface.co/papers/1910.10683
+
+The T5 paper proposes a unified text-to-text framework for transfer learning and introduces **sequence packing** (Section 3.5.2): grouping multiple short sequences into fixed-length blocks to reduce padding and improve training efficiency. Packing is supported in TRL via [`SFTConfig`] with the [`SFTTrainer`]. To enable packing with TRL, use this configuration:
+
+```python
+from trl import SFTConfig
+
+training_args = SFTConfig(
+    packing=True,  # enable sequence packing (Section 3.5.2 of the paper)
+    max_length=512,  # packed sequence length (Section 3.5.2 of the paper)
+)
+```
+
 ## Parameter-Efficient Fine-Tuning (PEFT)
 
 For general details on using PEFT with TRL, please refer to the [PEFT Integration](peft_integration) guide.
@@ -1065,6 +1080,27 @@ training_args = ORPOConfig(
 
 Papers relating to the [`experimental.cpo.CPOTrainer`]
 
+### Contrastive Preference Optimization: Pushing the Boundaries of LLM Performance in Machine Translation
+
+**📜 Paper**: https://huggingface.co/papers/2401.08417
+
+Introduces Contrastive Preference Optimization (CPO), a preference-based method for machine translation that trains models to avoid adequate-but-imperfect translations instead of mimicking references as in SFT. The paper analyzes limitations of SFT on MT (including reference quality issues) and shows that applying CPO to ALMA with only 22K parallel sentences yields ALMA-R, which matches or exceeds WMT competition winners and GPT-4 on WMT'21–WMT'23. Used in TRL via [`experimental.cpo.CPOTrainer`]. To reproduce the paper's setting, use this configuration:
+
+```python
+from trl.experimental.cpo import CPOConfig
+
+training_args = CPOConfig(
+    loss_type="sigmoid",  # preference learning loss (Section 3 of the paper)
+    cpo_alpha=1.0,  # NLL regularizer weight (Section 3 of the paper)
+    beta=0.1,  # β temperature (Section 4.2 of the paper)
+    learning_rate=1e-4,  # learning rate (official code)
+    lr_scheduler_type="inverse_sqrt",  # scheduler (official code)
+    num_train_epochs=1,  # Section 4.2 of the paper
+    warmup_ratio=0.01,  # warm-up ratio (Section 4.2 of the paper)
+    max_length=512,  # max sequence length (Section 4.2 of the paper)
+)
+```
+
 ### SimPO: Simple Preference Optimization with a Reference-Free Reward
 
 **📜 Paper**: https://huggingface.co/papers/2405.14734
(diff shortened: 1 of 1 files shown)


qgallouedec:
we could also add a new option in `packing_strategy`, it would allow not to add a new param

albertvillanova:
Thanks for your fast review, @qgallouedec! :hugs: 

I was just thinking about this specific aspect! Possible approaches:

1. Add to Trainer Configs (SFTConfig, GOLDConfig, etc.)

   ```python
   # In SFTConfig
   requeue_truncated_sequences: bool = field(default=False, ...)

   # In SFTTrainer
   dataset = pack_dataset(..., requeue_truncated_sequences=args.requeue_truncated_sequences
   ```

   Pros:
   - Users can configure via CLI args or config files
   - Consistent with other packing parameters (packing_strategy)
   - Explicit control for users who need it
   - Easy to discover in documentation

   Cons:
   - Adds another config parameter (config bloat)
   - Most users won't need it (default works for SFT)
   - Need to update multiple trainer configs
   - Increases cognitive load when choosing parameters

3. Keep ONLY in pack_dataset(), Don't Expose in Trainers

   - Leave as-is: parameter exists only in pack_dataset() function
   - Users who need requeuing preprocess manually:
     ```python
     dataset = pack_dataset(dataset, seq_length=4096, requeue_truncated_sequences=True)
     trainer = SFTTrainer(train_dataset=dataset, ...)
     ```

   Pros:
   - Simplest - follows YAGNI principle ("You Aren't Gonna Need It")
   - No config bloat
   - Default behavior works for 95% of TRL users
   - Advanced users can still access it via direct function call
   - Can add to configs later if demand warrants
   - Warning system guides users if they have issues

   Cons:
   - Less convenient for edge cases
   - Requires manual preprocessing step
   - Less discoverable (need to read pack_dataset docs)
   - Can't configure via command-line when using trainers

4. New Packing Strategy Names
    ```python
    packing_strategy: "bfd" | "bfd-requeue" | "wrapped"
    # "bfd" = no requeue (default, SFT-friendly)
    # "bfd-requeue" = with requeue (pre-training friendly)
    ```

   Pros:
   - No new config parameter
   - Self-documenting strategy names
   - Consistent with existing pattern

   Cons:
   - Less flexible than boolean flag
   - More strategies to document
   - Could confuse users ("which bfd do I use?")
   - Proliferates strategy variants

5. Auto-detect Based on Dataset Type

    ```python
    def pack_dataset(dataset, seq_length, strategy="bfd", auto_detect_requeue=True, ...):
        if auto_detect_requeue and is_conversational(dataset[0]):
            requeue = False  # SFT dataset
        else:
            requeue = True   # Assume long sequences
    ```

   Pros:
   - Zero configuration - "just works"
   - Most user-friendly for beginners

   Cons:
   - Heuristics can fail - what if SFT data isn't conversational format?
   - Hidden behavior harder to debug
   - Complex to implement correctly
   - Violates "explicit is better than implicit"
   - Can't handle mixed datasets

qgallouedec:
Nice!

> 1. Add to Trainer Configs (SFTConfig, GOLDConfig, etc.)

Adding a con for this one: This would imply that `SFTConfig(packing_strategy="wrapped", requeue_truncated_sequences=True)` wouldn't be allowed, which is very confusing, because this strategy does requeue truncated sequences.

> 2. Keep ONLY in pack_dataset(), Don't Expose in Trainers

Yes, this is an option but I'm not a fan: this would encourage the user to handle tokenization themselves. This error-prone, and frustrating for user (the don't benefit from the data prep).
And passing a pre tokenized dataset isn't something that I'd like to encourage in general for trl.

> 3. New Packing Strategy Names

I'd like to mitigate a the cons:

>    * Less flexible than boolean flag

not necessarily. Users still get explicit control, just encoded in the strategy name.

>    * More strategies to document

roughly equivalent to option 1, just documented at the strategy level instead of as a separate flag.

>    * Could confuse users ("which bfd do I use?")

fair, but with a good default and a "recommended if you don't know", this should be fine.

> 4. Auto-detect Based on Dataset Type

I think this is premature.
We don’t yet have strong evidence for reliable heuristics. The issue was observed with conversational data, but it might also affect some non-conversational cases, and not all conversational datasets behave the same.
Introducing heuristics now risks hidden behavior and harder debugging. I’d prefer keeping things explicit for now.

PR #5077, opened 2026-02-11
Fix logging warning suppression for transformers 4.56.2

albertvillanova:
Fix logging warning suppression for transformers 4.56.2.

Fix regression introduced in:
- #5058

### Problem

After merging #5058, the test with minimum versions (transformers 4.56.2) fails: https://github.com/huggingface/trl/actions/runs/21908322323/job/63254647774
> NotImplementedError: Cannot copy out of meta tensor; no data! Please use torch.nn.Module.to_empty() instead of torch.nn.Module.to() when moving module from meta to a different device.
```python
FAILED tests/test_reward_trainer.py::TestRewardTrainer::test_train_with_pretokenized_data - NotImplementedError: Cannot copy out of meta tensor; no data! Please use torch.nn.Module.to_empty() instead of torch.nn.Module.to() when moving module from meta to a different device.
FAILED tests/test_reward_trainer.py::TestRewardTrainer::test_train_from_causal_lm - NotImplementedError: Cannot copy out of meta tensor; no data! Please use torch.nn.Module.to_empty() instead of torch.nn.Module.to() when moving module from meta to a different device.
```

**Root cause**: In transformers 4.56.2, using `_keys_to_ignore_on_load_missing` with `device_map=None` causes the score.weight layer to remain as an uninitialized meta tensor, which cannot be moved to a device using the standard `.to()` method. This issue was fixed in transformers 4.57.0.

### Solution

Implements a version-aware approach that:
- **For transformers > 4.56.2**: Uses the new `ignore_seqcls_score_missing_key()` context manager (cleaner, from #5058)
- **For transformers 4.56.2**: Falls back to the old `suppress_from_pretrained_warning()` logging filter approach (backwards compatible)

CC: @qgallouedec

### DIFF
--- a/trl/trainer/reward_trainer.py
+++ b/trl/trainer/reward_trainer.py
@@ -345,6 +345,7 @@ class RewardTrainer(BaseTrainer):
             # Distributed training requires device_map=None ("auto" fails)
             if args.distributed_state.distributed_type in ["MULTI_GPU", "DEEPSPEED"]:
                 model_init_kwargs["device_map"] = None
+            model_init_kwargs["num_labels"] = 1  # the only output of the model is the reward score
             with suppress_seqcls_warning():
                 model = create_model_from_path(model, AutoModelForSequenceClassification, **model_init_kwargs)
         else:


qgallouedec:
thanks, I was investigating which commit in transformers "fixed" it.

qgallouedec (approved):
testing against the min versions is so useful

qgallouedec:
this one https://github.com/huggingface/transformers/pull/40850

albertvillanova:
Thanks for the useful context! :hugs:

PR #5078, opened 2026-02-11
docs: Add TR-DPO paper (2404.09656) to paper index

behroozazarkhalili:
## Summary

- Adds paper index entry for [Learn Your Reference Model for Real Good Alignment](https://huggingface.co/papers/2404.09656) (TR-DPO)
- TR-DPO dynamically updates the reference policy during training via soft updates
- Maps to `DPOConfig` with `sync_ref_model=True`, `ref_model_mixup_alpha`, and `ref_model_sync_steps`
- Method-specific hyperparameters from Table 1: α=0.6, τ=512, β=0.05
- Training config from Table 2: lr=1e-6, 1 epoch, max_length=1024, max_grad_norm=2, warmup_steps=100

## Notes

- `max_grad_norm` and `warmup_steps` are inherited from `TrainingArguments` — valid in `DPOConfig` but with different defaults (`max_grad_norm=1.0` vs paper's `2`, `warmup_steps=0` vs paper's `100`)

Ref: #4407

### DIFF
--- a/docs/source/paper_index.md
+++ b/docs/source/paper_index.md
@@ -722,6 +722,28 @@ training_args = DPOConfig(
 
 For the unpaired version, the user should utilize [`experimental.bco.BCOConfig`] and [`experimental.bco.BCOTrainer`].
 
+### Learn Your Reference Model for Real Good Alignment
+
+**📜 Paper**: https://huggingface.co/papers/2404.09656
+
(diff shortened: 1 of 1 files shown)


behroozazarkhalili:
**Note:** `max_grad_norm` and `warmup_steps` are inherited from `TrainingArguments`, so they are valid parameters in `DPOConfig`. Their defaults differ from the paper's values (`max_grad_norm` defaults to `1.0` in TrainingArguments vs `2` in the paper, `warmup_steps` defaults to `0` vs `100` in the paper).

PR #5100, opened 2026-02-13
Implement Agent Skills [3/N]: Create skills installer

albertvillanova:
Implement Agent Skills [3/N]: Create skills installer.

This PR introduces two new modules to improve the management and installation of TRL skills for AI agents. The changes add a robust system for installing, uninstalling, and listing skills, as well as a registry for defining where different agents look for their skills. This will make it much easier to manage skills across multiple agents and environments.

Part of:
- #5095

**Skill installation and management:**

* Added `SkillInstaller` class in `trl/skills/installer.py` to handle installing, uninstalling, and listing skills in agent directories, supporting both copy and symlink methods.
* Introduced the `InstallMethod` enum to specify installation methods (`copy` or `symlink`).

**Agent registry and targeting:**

* Added `AgentTarget` dataclass and `AGENT_REGISTRY` in `trl/skills/agents.py` to define where each agent (e.g., Claude, Codex, OpenCode) expects its skills, supporting both global (user-level) and project-level directories.
* Provided utility functions `get_agent_target` and `list_agent_names` for querying agent definitions and available agents.

PR #5097, opened 2026-02-13
Implement Agent Skills [2/N]:  Create skills module

albertvillanova:
Implement Agent Skills [2/N]: Create skills module.

This PR introduces a new "skills" utility module to the `trl` package, providing functionality for discovering, listing, and retrieving metadata about agent skills. It also adds comprehensive tests for this new functionality and updates linting rules to allow print statements in the new module.

Part of:
- #5095

**New skills module:**

* Added the `trl/skills/skills.py` file, which provides utilities for:
  - Listing all available agent skills by scanning for directories containing a `SKILL.md` file.
  - Retrieving the path to a specific skill's `SKILL.md`.
  - Parsing YAML frontmatter metadata from a skill's `SKILL.md`.
  - Printing information about all skills or a specific skill, including descriptions and additional metadata.
* Created the `trl/skills/__init__.py` file to expose the new skills utilities as part of the `trl.skills` namespace.

**Testing:**

* Added `tests/test_skills.py` with tests covering skill discovery, metadata parsing, error handling, and output of the print utility.

**Linting configuration:**

* Updated `pyproject.toml` to allow print statements in all files under `trl/skills/`.

Issue #5095, opened 2026-02-13
Implement Agent Skills for TRL


albertvillanova:
### Feature request

As privately discussed, we want to implement Agent Skills support for TRL, following the open standard from https://agentskills.io

Agent Skills are folders of instructions that AI agents can discover and use to learn how to use tools and libraries effectively.

CC: @lewtun @burtenshaw @qgallouedec 

### Motivation

Users who install TRL should be able to make TRL skills accessible to their AI agents, enabling agents to learn how to use TRL's CLI commands for training language models (SFT, DPO, GRPO,...).

### Your contribution

I'm starting to work on a proposal for this feature.

### Plan
- [x] #5096
- [x] #5097 
- [x] #5100
- [x] #5103

PR #5103, opened 2026-02-16
Implement Agent Skills [4/N]: Create skills CLI

albertvillanova:
Implement Agent Skills [4/N]: Create skills CLI.

Part of:
- #5095

This PR requires:
- #5100
- #5097

This PR introduces a new command-line interface (CLI) module for managing TRL skills. The main focus is on providing users with CLI commands to install, uninstall, and list TRL skills for various AI agent directories.

**New CLI for TRL Skills Management:**

* Added a new `trl/skills/cli.py` module that provides CLI commands (`install`, `uninstall`, `list-installed`) for managing TRL skills in agent directories. This includes argument parsing, target resolution, and user-friendly output for skill management tasks.

### DIFF
--- a/trl/skills/cli.py
+++ b/trl/skills/cli.py
@@ -124,18 +124,12 @@ def cmd_install(args):
             print("✓")
             success_count += 1
 
-        except FileNotFoundError as e:
-            print("✗")
-            print(f"  Error: {e}")
         except FileExistsError as e:
             print("✗")
             print(f"  Error: {e}")
             if not args.force:
                 print("  Use --force to overwrite")
-        except ValueError as e:
-            print("✗")
-            print(f"  Error: {e}")
-        except Exception as e:
+        except (FileNotFoundError, ValueError) as e:
             print("✗")
             print(f"  Error: {e}")
 
@@ -159,19 +153,7 @@ def cmd_uninstall(args):
         print(f"\nSkill '{args.skill}' has been removed")
         return 0
 
-    except FileNotFoundError as e:
-        print("✗")
-        print(f"Error: {e}")
-        return 1
-    except PermissionError as e:
-        print("✗")
-        print(f"Error: {e}")
-        return 1
-    except ValueError as e:
-        print("✗")
-        print(f"Error: {e}")
-        return 1
-    except Exception as e:
+    except (FileNotFoundError, PermissionError, ValueError) as e:
         print("✗")
         print(f"Error: {e}")
         return 1
@@ -208,9 +190,6 @@ def cmd_list(args):
         return 0
 
     except ValueError as e:
-        print(f"Error: {e}")
-        return 1
-    except Exception as e:
         print(f"Error: {e}")
         return 1
 


trl/skills/cli.py
```diff
@@ -0,0 +1,220 @@
...
+    uninstall_parser.set_defaults(func=cmd_uninstall)
+
+
+def cmd_install(args):
+    """Handle 'trl skills install' command."""
+    # Validate arguments
+    if not args.skill and not args.all:
+        print("Error: Either provide a skill name or use --all to install all skills")
+        print("Usage: trl skills install <skill> --target <target>")
+        print("   or: trl skills install --all --target <target>")
```
qgallouedec, line 96: I'm not sure to understand why we print and return 1 instead of just raising an exception?

trl/skills/cli.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    Add skills subcommands to the parser.
+
+    This creates nested subcommands under 'trl skills' for managing skill installations.
+
+    Args:
+        subparsers: Subparsers from 'trl skills' command
+    """
+    # Parent parser for common target options
+    target_parser = argparse.ArgumentParser(add_help=False)
```
qgallouedec, line 36: `add_help=False` why?

trl/skills/skills.py
```diff
@@ -55,26 +65,84 @@ def _get_trl_skills_dir() -> Path:
...
+    Args:
+        target (`str | Path`): Agent name (e.g., 'claude', 'codex') or directory path.
+        scope (`str`, defaults to `"project"`):
+            Scope for agent names: 'global' (user-level like ~/.agent/skills/) or 'project' (./agent/skills/).
+
+    Returns:
+        `Path`: Resolved absolute path.
+
+    Raises:
+        `ValueError`: If agent name is not recognized.
```
qgallouedec: it's a keyerror instead, no? and it's the same if scope is not recognized

trl/skills/cli.py
albertvillanova: CLI error-flow is intentional and consistent with parser behavior: `cmd_install` prints and returns 1 because `trl/cli.py` dispatch expects subcommand handlers to return exit codes. Raising directly in cmd_install would either:
- bubble up as traceback (bad UX)
- require a centralized catch+exit wrapper

So "print + return code" is a valid CLI pattern here.

trl/skills/cli.py
albertvillanova: Because `target_parser` is a parent parser reused by `install` and `uninstall`. With help enabled, you risk duplicate `-h/--help` option conflicts. `add_help=False` is standard for shared parent parsers.

trl/skills/skills.py
albertvillanova: Thanks, there are 2 issues here:
1. For public API argument validation, `ValueError` is the better contract. `KeyError` here is currently an implementation detail leak from dict indexing, not a good external API error type.
2. Are those documented errors really raised? Not all of them. I'm fixing this.

trl/skills/cli.py
```diff
@@ -0,0 +1,220 @@
...
+        print(f"Error: {e}")
+        return 1
+    except ValueError as e:
+        print("✗")
+        print(f"Error: {e}")
+        return 1
+    except Exception as e:
+        print("✗")
+        print(f"Error: {e}")
+        return 1
```
qgallouedec: Technically we only need the last one, since we always print the same str

albertvillanova: Yes! I am removing the last branch (`except Exception` is too wide and we should propagate unexpected errors with a traceback; I am also grouping the others into tuples.

PR #5107, opened 2026-02-16
feature: Configurable num logprobs in vLLM generation

LeonEricsson:
### Summary

- Adds a `logprobs` parameter across the vLLM generation stack (client, server, and generation orchestrator), allowing callers to request the top-N log-probabilities per generated token rather than only the sampled token's log-probability.
- `VLLMGeneration::generate` returns a new `logprob_token_ids` field alongside `logprobs`, both with shape `(num_sequences, seq_len, num_logprobs)`, giving downstream consumers access to the full top-k distribution.

### Motivation

This change is a prerequisite for an upcoming PR I’m working on.

More broadly, it is a strict improvement over the original implementation. Previously, `logprobs` was hardcoded to `0`, which discarded the richer top-k information that vLLM already computes internally. I was lazy when I implemented this the first time around because we only needed sampled logps. By exposing the `logprobs` parameter, we incur no additional overhead while unlocking access to information that was previously inaccessible.

Minimal upstream effect, `GRPOTrainer` and `RLOOTrainer` use the default `logprobs=0`

### todo

- [ ] run modified slow test, requires multiple accelerators

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/docs/source/dpo_trainer.md
+++ b/docs/source/dpo_trainer.md
@@ -97,7 +97,7 @@ The [`DPOTrainer`] tokenizes each input using the model's tokenizer.
 
 The loss used in DPO is defined as follows:
 $$
-\mathcal{L}_{\mathrm{DPO}}(\theta) = -\mathbb{E}_{(x,y^{+},y^{-})}\!\left[\log \sigma\!\left(\beta\Big(\log\frac{{\pi_{\theta}(y^{+}\!\mid x)}}{{\pi_{\mathrm{ref}}(y^{+}\!\mid x)}}-\log \frac{{\pi_{\theta}(y^{-}\!\mid x)}}{{\pi_{\mathrm{ref}}(y^{-}\!\mid x)}}\Big)\right)\right]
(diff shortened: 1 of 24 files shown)


qgallouedec (approved):
lgtm, please just wait for #5165 to be merge first

qgallouedec:
btw I've added support for logprobs=None, and explicitly set logprobs=0 in GRPO and logprobs=None in RLOO
also added a small test for the extractor

LeonEricsson:
> lgtm, please just wait for #5165 to be merge first

> btw I've added support for logprobs=None, and explicitly set logprobs=0 in GRPO and logprobs=None in RLOO also added a small test for the extractor

great, thanks

trl/generation/vllm_client.py
```diff
@@ -252,8 +256,11 @@ def generate(
                     List of lists of token IDs representing the tokenized input prompts.
                 - `completion_ids` (`list[list[int]]`):
                     List of lists of token IDs representing the model-generated completions for each prompt.
-                - `logprobs` (`list[list[float]]`):
-                    List of lists of log probabilities for each generated token.
+                - `logprobs` (`list[list[list[float]]]`):
```
winglian, line 259: where is the implementation of the change from list[list[float]] to list[list[list[float]]]? down on line 293, this function is still just getting it out of the response with no modifications.

LeonEricsson: it's on the server side

https://github.com/huggingface/trl/blob/410c00bfaead36b0048921a123739bd0cb4c3e7c/trl/scripts/vllm_serve.py#L647-L654

the client just passes it through

PR #5111, opened 2026-02-17
Fix Qwen3 schema

qgallouedec:
## PR Description

This PR fixes 2 bugs in the Qwen3 schema.

1. Empty `<think>...</think>` blocks were producing a spurious `reasoning_content` field with just a newline.
2. When an assistant message contained a tool call but no reasoning content, the parsed `content` field included a trailing newline.

### Reproducer for 1

```python
>>> from trl.chat_template_utils import add_response_schema, parse_response
>>> tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3-0.6B")
>>> tokenizer = add_response_schema(tokenizer)
>>> messages = [
...     {"role": "user", "content": "What is 3*4?"},
...     {"role": "assistant", "content": "12"},
... ]
>>> prefix = tokenizer.apply_chat_template(messages[:1], add_generation_prompt=True).input_ids
>>> text = tokenizer.apply_chat_template(messages).input_ids
>>> response = text[len(prefix) :]
>>> parse_response(tokenizer, response)
```

Before this change:

```python
{'role': 'assistant', 'content': '12', 'reasoning_content': '\n'}
```

After this change:

```python
{'role': 'assistant', 'content': '12'}
```

#### Root cause

Qwen3 templates can emit an empty thinking block (`<think>\n\n</think>`) for assistant responses.  
The previous schema regex captured any non-empty substring between tags with:

```regex
(?P<reasoning_content>.+?)
```

So a newline-only payload was captured and surfaced as `reasoning_content`.

#### Fix (suggested by GPT-5.3 Codex)

Updated `qwen3_schema["x-regex"]` in `trl/chat_template_utils.py` so:

- Empty/whitespace-only `<think>` content is consumed but **not captured**.
- `reasoning_content` is captured only when it contains at least one non-whitespace character.

Concretely, the reasoning portion now uses:

```regex
(?:(?P<reasoning_content>.*?\S.*?)\n?|[\s]*)
```

inside the `<think>...</think>` block.

### Reproducer for 2

```python
>>> from trl.chat_template_utils import add_response_schema, parse_response
>>> 
>>> tokenizer = AutoTokenizer.from_pretrained("trl-internal-testing/tiny-Qwen3MoeForSequenceClassification")
>>> tokenizer = add_response_schema(tokenizer)
>>> tool_calls = [
...     {"type": "function", "function": {"name": "multiply", "arguments": {"a": 3, "b": 4}}},
... ]
>>> messages = [
...     {"role": "user", "content": "What is 3*4?"},
...     {"role": "assistant", "content": "Let's call the tool.", "tool_calls": tool_calls},
... ]
>>> prefix = tokenizer.apply_chat_template(messages[:1], add_generation_prompt=True).input_ids
>>> text = tokenizer.apply_chat_template(messages).input_ids
>>> response = text[len(prefix) :]
>>> output = parse_response(tokenizer, response)
>>>output["content"]
```

Before this change:

```python
"Let's call the tool.\n"
```

After this change:

```python
"Let's call the tool."
```

#### Root cause

In the Qwen3 template, when assistant content and tool calls coexist, a separator newline is emitted before the first `<tool_call>`:

```jinja
{%- if (loop.first and content) or (not loop.first) %}
    {{- '\n' }}
{%- endif %}
```

The schema regex previously captured this separator as part of `content`.

We update `qwen3_schema["x-regex"]` in `trl/chat_template_utils.py` to consume one newline immediately before `<tool_call>` outside the `content` capture:

From:

```regex
(?P<content>.*?)(?=(?:<tool_call>|<\|im_end\|>|$))
```

To:

```regex
(?P<content>.*?)(?:\n(?=<tool_call>))?(?=(?:<tool_call>|<\|im_end\|>|$))
```

### Validation and new tests

This PR also updates the test cases in `tests/test_chat_template_utils.py`, and includes the reproducer as a test to ensure the fix is effective.

### DIFF
--- a/tests/test_chat_template_utils.py
+++ b/tests/test_chat_template_utils.py
@@ -213,12 +213,175 @@ class TestIsChatTemplatePrefixPreserving:
         assert is_chat_template_prefix_preserving(tokenizer) is False
 
 
+@pytest.mark.parametrize(
+    "tokenizer_name",
+    [
+        pytest.param("trl-internal-testing/tiny-Qwen3MoeForSequenceClassification", id="qwen3"),
+    ],
+)
 class TestGetTrainingChatTemplate:
-    def test_qwen3(self):
-        tokenizer = AutoTokenizer.from_pretrained("trl-internal-testing/tiny-Qwen3MoeForSequenceClassification")
+    def test_new_chat_template_is_prefix_preserving(self, tokenizer_name):
+        tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
         assert is_chat_template_prefix_preserving(tokenizer) is False
         tokenizer.chat_template = get_training_chat_template(tokenizer)
         assert is_chat_template_prefix_preserving(tokenizer) is True
+
+    def test_behavior_unchanged_single_user_no_generation_prompt(self, tokenizer_name):
+        tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
+        messages = [{"role": "user", "content": "What color is the sky?"}]
+        before = tokenizer.apply_chat_template(messages, tokenize=False)
+        new_chat_template = get_training_chat_template(tokenizer)
+        after = tokenizer.apply_chat_template(messages, tokenize=False, chat_template=new_chat_template)
+        assert before == after
+
+    def test_behavior_unchanged_single_user_with_generation_prompt(self, tokenizer_name):
+        tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
+        messages = [{"role": "user", "content": "What color is the sky?"}]
+        before = tokenizer.apply_chat_template(messages, add_generation_prompt=True, tokenize=False)
+        new_chat_template = get_training_chat_template(tokenizer)
+        after = tokenizer.apply_chat_template(
+            messages,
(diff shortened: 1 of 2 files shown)


albertvillanova (approved):
Thanks.

Just a naive question: is it our code which has to fix this? Why not `transformers`?

qgallouedec:
In an ideal world, yes, this would not be fixed in TRL.
Ideally, the schema should live in the model repository itself, since it is tightly coupled to the chat template. However, schema adoption across model repos is currently very low, and Qwen3 does not provide one.
To support tool-calling training today, TRL provides a small set of known schemas (including Qwen3). This is essentially a pragmatic workaround.
Additionally, Transformers is currently redesigning response parsing and plans to drop schema-based parsing entirely (see https://github.com/huggingface/transformers/issues/42249#issuecomment-3670148149). In the meantime, they are not extending or maintaining schema support.
Given this context, fixing the schema in TRL is the most practical solution for now. Once the new parsing backend lands in Transformers, we can remove these schemas and rely on the updated mechanism.

albertvillanova:
Thanks, crystal clear now! :hugs:

PR #5128, opened 2026-02-19
Fix import latency [1/N]: Extract _LazyModule to dedicated module

albertvillanova:
Extract _LazyModule to dedicated module.

This is the 1st PR to fix the TRL import latency I am working on:
- #5130

This PR refactors the handling of the `_LazyModule` class by moving its implementation from `import_utils.py` to a new dedicated file, `_lazy_module.py`. This change improves code organization and separation of concerns. All imports of `_LazyModule` across the codebase are updated to reflect this new location, and the now-redundant code is removed from `import_utils.py`.

### Context

TRL has a _LazyModule class in trl/import_utils.py that enables lazy loading of submodules. However, import_utils.py also contains the _is_package_available function borrowed from transformers, which was introduced by:
- #2064

This is the root cause of some seconds import latency regression (import trl eagerly loads all of transformers).

### Solution

This PR extracts _LazyModule into its own dedicated file (trl/_lazy_module.py) with zero external dependencies. This is a pure refactor with no behavior change.

### Motivation

Moving _LazyModule out of import_utils.py is a clean, reviewable refactor on its own. It also removes the responsibility mismatch: _LazyModule (module loading machinery) had no business living alongside package availability checks.

### DIFF
--- a/trl/__init__.py
+++ b/trl/__init__.py
@@ -17,7 +17,7 @@ from importlib.metadata import PackageNotFoundError, version
 from typing import TYPE_CHECKING
 
 from . import _compat
-from .import_utils import _LazyModule
+from ._lazy_module import _LazyModule
 
 
 try:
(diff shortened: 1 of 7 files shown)


qgallouedec:
Good! I've always been annoyed and surprised by the time trl needs to be imported despite the lazymodule. Agree with these changes

PR #5132, opened 2026-02-19
Fix `SFTTrainer` support for single-image data

qgallouedec:
In TRL, we support both an `image` column (containing a single image) and an `images` column (containing a list of images).

Before this PR, `SFTTrainer` would ignore the `image` column because it was not included in the trainer’s signature. This PR fixes that issue.

I also suggest that for the `zen-image` dataset, we use an `image` column that directly contains the image, rather than a list with a single element. The list-based format is already covered and tested by the `zen-multi-image` dataset.

### DIFF
--- a/tests/test_dpo_trainer.py
+++ b/tests/test_dpo_trainer.py
@@ -1014,9 +1014,7 @@ class TestDPOTrainer(TrlTestCase):
     @require_vision
     def test_train_vlm(self, model_id):
         # Get the dataset
-        dataset = load_dataset(
-            "trl-internal-testing/zen-image", "conversational_preference", split="train"
-        )
+        dataset = load_dataset("trl-internal-testing/zen-image", "conversational_preference", split="train")
 
         # Initialize the trainer
         training_args = DPOConfig(
@@ -1105,9 +1103,7 @@ class TestDPOTrainer(TrlTestCase):
     @pytest.mark.skip(reason="Model google/gemma-3n-E2B-it is gated and requires HF token")
     def test_train_vlm_gemma_3n(self):
         # Get the dataset
-        dataset = load_dataset(
-            "trl-internal-testing/zen-image", "conversational_preference", split="train"
-        )
+        dataset = load_dataset("trl-internal-testing/zen-image", "conversational_preference", split="train")
 
         # Initialize the trainer
         training_args = DPOConfig(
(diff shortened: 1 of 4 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -266,7 +266,7 @@ class DataCollatorForVisionLanguageModeling(DataCollatorMixin):
     image processing on-the-fly to efficiently prepare batches.
 
     Each input example should be a dictionary containing at least:
-    - An `"images"` key holding the image data.
+    - An `"images"` key holding a list of images, or an `"image"` key holding a single image.
```
albertvillanova, line 269: Now that we are simplifying the API for the v1 release, are we sure we want to support both `image` and `images`?

qgallouedec: If we had to support only one format, it would have to be `"images"`, since we don’t want to drop multi-image support.

So the question becomes:

1. Do we completely drop support for the `"image"` column and require users to always provide a list under `"images"`, even for single-image datasets?
2. Or do we allow the `"images"` column to accept either a single image or a list of images for convenience, meaning users currently using `"image"` would just need to rename their column?

I am slightly inclined to keep the API as it is; I don’t have a strong opinion, but my current thinking is:

* Supporting both `"image"` and `"images"` is not particularly burdensome.
* Option (1) could be somewhat annoying for users, since most datasets are single-image. Also, multi-image datasets can be slightly trickier to set up (e.g. requiring explicit `features`, see [here](https://github.com/huggingface/trl/blob/bc4edf6f02e6f07549d43b3543cb54d597cd3d91/scripts/generate_zen_multi_image_dataset.py#L80C151-L80C164)), so forcing everything into a list-based format might add friction for the common case.
* Option (2) feels like unnecessary churn: it doesn’t really simplify the internal implementation (we still need to handle both single and multiple images), and we’d have to infer single vs. multi-image from the sample structure rather than from the column name.

albertvillanova: OK, so let's keep both column names...

qgallouedec:
CI doesn't start. It passed for https://github.com/huggingface/trl/pull/5132/commits/2e3644808c22c5a798cf726e8516d891815b88c8 so I'll go ahead and merge

Issue #5144, opened 2026-02-21
GRPOTrainer tool_mask can become longer than completion_ids after tool-call retokenization


MichalMraz:
### Summary

In `GRPOTrainer`, a tool-call round can retokenize `prompt+completion+tool` so that the resulting completion portion is shorter than the previous completion length.  
In `_tool_call_loop`, `tool_mask` (and optionally `logprobs`) was extended but not truncated in this case, so lengths diverged.

This later fails in `_compute_loss` at:
`mask = completion_mask * inputs["tool_mask"]`
with:
`RuntimeError: The size of tensor a (...) must match the size of tensor b (...) at non-singleton dimension 1`

I will be adding a PR with a bugfix and a minimal regression test that reproduces this path inside `trainer.train()`.

### Reproduction

```python
  #!/usr/bin/env python
"""
Reproduce the GRPO tool-mask shape mismatch *inside* `GRPOTrainer.train()`.

This script intentionally forces a tool-call round where:
- `completion_ids` is effectively shortened by re-tokenization logic, but
- `tool_mask` is only extended (not truncated),
which leads to a mismatch at:
`trl/trainer/grpo_trainer.py`, `_compute_loss`, line with
`completion_mask * inputs["tool_mask"]`.
"""

from __future__ import annotations

import types
from typing import Any

import torch
import torch.distributed.fsdp as fsdp


# Compatibility shims for this environment.
if not hasattr(fsdp, "FSDPModule"):
    fsdp.FSDPModule = fsdp.FullyShardedDataParallel
if not hasattr(torch.backends.mps, "is_macos_or_newer"):
    torch.backends.mps.is_macos_or_newer = lambda major, minor: False

from datasets import load_dataset

import trl.trainer.grpo_trainer as grpo_mod
from trl import GRPOConfig, GRPOTrainer


def multiply_tool(a: int, b: int) -> int:
    """Multiply two integers.

    Args:
        a: First integer.
        b: Second integer.

    Returns:
        Product of `a` and `b`.
    """
    return a * b


def constant_reward(completions: list[Any], **kwargs) -> list[float]:
    del kwargs
    return [0.0] * len(completions)


def main() -> None:
    dataset = load_dataset("trl-internal-testing/zen", "conversational_prompt_only", split="train[:3]")

    args = GRPOConfig(
        output_dir="/tmp/trl_bug_repro_train",
        per_device_train_batch_size=3,
        num_generations=3,
        max_completion_length=128,
        max_steps=1,
[...]

PR #5145, opened 2026-02-21
Fix GRPO tool mask alignment after tool-call retokenization

MichalMraz:
## What does this PR do?

Fixes [#5144](https://github.com/huggingface/trl/issues/5144)

This PR fixes a shape-mismatch bug in `GRPOTrainer` tool-call flow.

Root cause:
- During `_tool_call_loop`, tool-round retokenization can make the completion part shorter than the previous completion.
- `tool_mask`/`logprobs` were only extended, not truncated, so they could become longer than `completion_ids`.
- This later crashes in `_compute_loss` when multiplying `completion_mask * tool_mask`.

Fix:
- In `trl/trainer/grpo_trainer.py`, align `tool_mask` and `logprobs` to computed completion lengths by truncating or padding as needed before appending post-tool tokens.

Tests:
- Added regression test:
  `tests/test_grpo_trainer.py::test_training_with_tools_keeps_masks_aligned_when_retokenization_shortens_completion`
- This test fails on old code with the tensor-size mismatch and passes with this patch.

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -2344,6 +2344,110 @@ class TestGRPOTrainer(TrlTestCase):
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
 
+    @pytest.mark.xfail(
+        condition=Version(transformers.__version__) < Version("5.0.0"),
+        reason="Tool parsing is not supported in transformers versions below 5.0.0",
+        strict=True,
+    )
+    @require_jmespath
+    def test_training_with_tools_keeps_masks_aligned_when_retokenization_shortens_completion(self):
+        dataset = load_dataset("trl-internal-testing/zen", "conversational_prompt_only", split="train[:3]")
+
+        def reward_func(completions, **kwargs):
+            return [0.0] * len(completions)
+
+        training_args = GRPOConfig(
+            output_dir=self.tmp_dir,
+            per_device_train_batch_size=3,
+            num_generations=3,
+            max_completion_length=128,
+            max_steps=1,
+            report_to="none",
+        )
+        trainer = GRPOTrainer(
+            model="trl-internal-testing/tiny-Qwen3MoeForCausalLM",
+            reward_funcs=reward_func,
+            args=training_args,
+            train_dataset=dataset,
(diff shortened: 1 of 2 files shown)


qgallouedec:
> During _tool_call_loop, tool-round retokenization can make the completion part shorter than the previous completion.

shouldn't happen, what model do you use?

MichalMraz:
I originally hit it with Qwen3-32B on transformers==5.0.0.  
The test uses a tiny Qwen model only

qgallouedec:
ok, can you share what are the generated completion ids / tools calls when if occurs?

MichalMraz:
For example this kind of model response causes it
```
assistant_text = (
        "<tool_call>\n"
        "{\n"
        '                                  "name"        :                 "multiply_tool",\n'
        '                                  "arguments"   :                 {\n'
        '                                                    "a"   :  3,\n'
        '                                                    "b"   :  4\n'
        "                                                }\n"
        "}\n"
        f"</tool_call>{tokenizer.eos_token}"
    )
```
The completion_ids are then 
`[151657, 198, 515, 6656, 330, 606, 1, 286, 549, 338, 330, 64648, 22785, 756, 6656, 330, 16370, 1, 256, 549, 338, 341, 6374, 330, 64, 1, 256, 549, 220, 220, 18, 345, 6374, 330, 65, 1, 256, 549, 220, 220, 19, 198, 4569, 456, 532, 151658, 151645]
`
and parsed tool_calls
`[{'type': 'function', 'function': {'name': 'multiply_tool', 'arguments': {'a': 3, 'b': 4}}}]
`

Here

- original completion length = 47
- retokenized completion_tool length = 41

So old logic computes negative delta (41 - 47 = -6).


Yes, it is a somewhat degenerate case but it sometimes happens during unstable training, causing a crash

MichalMraz:
Friendly bump! I recognise it's not a high priority issue but I'd like to see it resolved. Let me know if there's anything I should update or improve to help move this forward.

qgallouedec:
Yep, thanks for the PR. No worry it still a priority PR.

Actually I think the "right" solution is not to retokenize the tool call. We might need some deeper re-factoring of the generation loop here :/

qgallouedec:
surperceded by #5242 thanks again for point this out

PR #5147, opened 2026-02-23
fix: wake up vLLM weights before sync to prevent writes to freed memory

### LINKED ISSUE #5142: [Bug] colocate sleep mode: sync_weights() writes to freed GPU memory before wake_up() on non-CUDA backends
### System Info

- TRL: latest main branch
- vLLM: 0.11.1.dev
- vllm-ascend: 0.11.0rc1
- Hardware: Ascend 910B NPU (issue applies to any non-CUDA backend with strict memory deallocation in sleep mode)
- Training: GRPO with `mode="colocate"`, `enable_sleep_mode=True`

### Bug Description

When using colocate mode with sleep enabled, `sync_weights()` calls `llm_model.load_weights()` while vLLM is still in `sleep(level=2)` state. The weights have been offloaded to CPU by the device memory allocator, so `load_weights()` writes to freed/unmapped device memory.

The call sequence in `_generate_single_turn()` (`grpo_trainer.py:1149`):

```
1. sync_weights()        →  load_weights() into sleeping vLLM  ← BUG: writes to freed memory
2. generate()            →  wake_up(tags=["weights"])           ← too late
```

On CUDA, this doesn't crash because CUDA's memory management tolerates writes to "freed" virtual addresses in sleep mode. On backends with strict physical memory unmapping (e.g., Ascend NPU's `CaMemAllocator` which calls `unmap_and_release` during sleep), this causes a hardware-level DDR out-of-range exception:

```
[AIC_INFO] The DDR address of the MTE instruction is out of range.
The current working operator name is aclnnInplaceCopy.
NPU function error: call failed, error code is 507035
```

### Relevant Code

`trl/generation/vllm_generation.py`:

```python
# sync_weights() - line 407: no wake_up before load_weights()
def sync_weights(self):
    model = self.model  # ← starts immediately, vLLM still sleeping
    ...
    llm_model.load_weights([(name, param.data)])  # ← writes to freed memory

# generate() - line 514: wake_up happens here, AFTER sync_weights already ran
if self.mode == "colocate" and self.enable_sleep_mode:
    torch.cuda.empty_cache()
    self.llm.wake_up(tags=["weights"])  # ← too late
```

### Additional Issues

1. `generate()` hardcodes `torch.cuda.empty_cache()` (line 516), which is a no-op on non-CUDA devices.
2. `generate()` calls `self.llm.collective_rpc("reload_weights")` (line 519), which is not implemented in non-CUDA vLLM backends (e.g., vllm-ascend's `NPUWorkerV1`), causing `NotImplementedError`.

### Expected Behavior

`wake_up(tags=["weights"])` should be called **before** `load_weights()` in `sync_weights()`, not after in `generate()`. This ensures device memory is valid before writing to it, regardless of backend.

### Proposed Fix

1. Move `wake_up(tags=["weights"])` to the beginning of `sync_weights()` for colocate+sleep mode.
2. Add a `_vllm_weights_awake` flag to avoid double `wake_up` (since `generate()` also calls it).
3. Reset the flag after `sleep(level=2)` at the end of `generate()`.
4. Replace `torch.cuda.empty_cache()` with device-agnostic cache clearing.
5. Guard `collective_rpc("reload_weights")` with a try/except or feature check, since not all backends implement it.

```python
def sync_weights(self):
    # Ensure vLLM weights are on device before load_weights()
    if self.mode == "colocate" and self.enable_sleep_mode and not self._vllm_weights_awake:
        torch.cuda.empty_cache()  # or device-agnostic equivalent
        self.llm.wake_up(tags=["weights"])
        self._vllm_weights_awake = True
    ...
```

### Reproduction

Run GRPO training with:
- `use_vllm=True`
- `vllm_mode="colocate"`
- `vllm_enable_sleep_mode=True`
- Any non-CUDA backend with strict memory deallocation in sleep mode (e.g., Ascend NPU via vllm-ascend)

The crash occurs on the first training step when `sync_weights()` is called.

### Checklist

- [x] I have checked that there is no similar issue in the repo.
- [x] I have provided a minimal reproduction of the bug.

bledden:
## Summary

- Fixes a temporal ordering bug where `sync_weights()` calls `load_weights()` while the vLLM engine is still sleeping at level 2, writing to freed/unmapped device memory
- Adds `wake_up(tags=["weights"])` at the start of `sync_weights()` so device memory is re-mapped before weight writes
- Replaces hardcoded `torch.cuda.empty_cache()` with a device-agnostic helper supporting XPU, NPU, and MLU backends
- Guards `collective_rpc("reload_weights")` against `NotImplementedError` on backends that don't implement it

## Context

When using `vllm_mode="colocate"` with `vllm_enable_sleep_mode=True`, the call sequence is:

1. `sync_weights()` → `load_weights()` into sleeping vLLM (**BUG: writes to freed memory**)
2. `generate()` → `wake_up(tags=["weights"])` (**too late**)

On CUDA this is silently tolerated because the virtual addresses remain mapped. On backends with strict physical memory management (e.g., Ascend NPU via vllm-ascend), this causes a hardware exception:

```
[AIC_INFO] The DDR address of the MTE instruction is out of range.
```

The fix moves the `wake_up(tags=["weights"])` to the beginning of `sync_weights()`. The existing `wake_up` in `generate()` is kept as a safety net for the no-sync path (it's idempotent if weights are already awake).

## Test plan

- [ ] Verify no regression on CUDA with `colocate` + `enable_sleep_mode=True`
- [ ] Verify fix resolves the crash on Ascend NPU (requires NPU hardware)
- [ ] Verify `wake_up` is idempotent: calling it twice does not error

Fixes #5142

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,7 +42,7 @@ jobs:
     name: Tests
     strategy:
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
       fail-fast: false
     runs-on:
       group: aws-g4dn-2xlarge
--- a/docs/source/dpo_trainer.md
+++ b/docs/source/dpo_trainer.md
@@ -82,7 +82,7 @@ print(next(iter(dataset["train"])))
 
 ## Looking deeper into the DPO method
 
-Direct Preference Optimization (DPO) is a training method designed to align a language model with preference data. Instead of supervised input–output pairs, the model is trained on pairs of completions to the same prompt, where one completion is preferred over the other. The objective directly optimizes the model to assign higher likelihood to preferred completions than to dispreferred ones, relative to a reference model, without requiring an explicit reward model.
+Direct Preference Optimization (DPO) is a training method designed to align a language model with preference data. Instead of supervised input–output pairs, the model is trained on pairs of completions to the same prompt, where one completion is preferred over the other. The objective directly optimizes the model to widen the margin between the log-likelihoods of preferred and dispreferred completions, relative to a reference model, without requiring an explicit reward model. In practice, this is typically achieved by suppressing the likelihood of dispreferred completions rather than by increasing the likelihood of preferred ones.
 
 This section breaks down how DPO works in practice, covering the key steps: **preprocessing** and **loss computation**.
 
@@ -230,7 +230,7 @@ trainer.train()
 ```
 
 > [!TIP]
-> When training adapters, you typically use a higher learning rate (≈1e‑5) since only new parameters are being learned.
+> When training adapters, you typically use a higher learning rate (≈1e‑5) than full fine-tuning since only new parameters are being learned.
 >
 > ```python
 > DPOConfig(learning_rate=1e-5, ...)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,8 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
-    "Programming Language :: Python :: 3.13"
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14"
 ]
 requires-python = ">=3.10"
 dependencies = [
@@ -108,7 +109,7 @@ dev = [
     # kernels
     "kernels",
     # liger
-    "liger-kernel>=0.7.0",
+    #"liger-kernel>=0.7.0",
     # peft
     "peft>=0.8.0",
     # quality
--- a/scripts/generate_zen_image_dataset.py
+++ b/scripts/generate_zen_image_dataset.py
@@ -15,7 +15,7 @@
 from dataclasses import dataclass, field
 
 import numpy as np
-from datasets import Dataset, Features, Image, List, Value
+from datasets import Dataset, Features, Image, Value
 from transformers import HfArgumentParser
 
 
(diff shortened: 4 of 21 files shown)


trl/generation/vllm_generation.py
```diff
@@ -36,6 +36,21 @@
...
+    if hasattr(torch, "xpu") and torch.xpu.is_available():
+        torch.xpu.empty_cache()
+    elif hasattr(torch, "npu") and torch.npu.is_available():
+        torch.npu.empty_cache()
+    elif hasattr(torch, "mlu") and torch.mlu.is_available():
+        torch.mlu.empty_cache()
+    else:
+        torch.cuda.empty_cache()
+
+
```
qgallouedec: why do we need this?

trl/generation/vllm_generation.py
```diff
@@ -511,12 +534,15 @@ def generate(self, prompts: list, num_generations: int, profiler: ProfilingConte
...
         if self.mode == "colocate" and self.enable_sleep_mode:
-            torch.cuda.empty_cache()  # required to avoid OOM in some cases
+            _empty_device_cache()
             self.llm.wake_up(tags=["weights"])
             # Work around for https://github.com/vllm-project/vllm/issues/29341
-            self.llm.collective_rpc("reload_weights")
+            try:
+                self.llm.collective_rpc("reload_weights")
+            except NotImplementedError:
+                pass
```
qgallouedec, line 549: in which case does it raise a `NotImplementedError`?

qgallouedec: ok the answer is in the issue. I'll just add a comment here.

qgallouedec (approved):
LGTM, I got my answers from the issue!

More generally, thank you for fixing these types of issues related to non-CUDA backends. We develop and test with CUDA, so it is important that external contributors ensure that it also works with other backends.

PR #5156, opened 2026-02-24
Simplify NeMo Gym user experience

cmunley1:
# What does this PR do?

Simplifies NeMo Gym user experience by automating the launch of NeMo Gym servers and using NeMo Gym's RolloutCollectionHelper. 

Also fixes an existing bug in the NeMo Gym integration related to prompt duplication in `rollout_func`.


trl/trainer/grpo_trainer.py
```diff
@@ -2225,7 +2225,8 @@ def log(self, logs: dict[str, float], start_time: float | None = None) -> None:
 
         logs = {**logs, **metrics}
         super().log(logs, start_time)
-        self._metrics[mode].clear()
+        if mode == "train" or "eval_loss" in logs:
```
cmunley1: i found wandb validation metrics were only showing 1 sample, not the full eval. I suspect it is due to my use of `per_device_eval_batch_size=1` along with logging inside of the rollout function https://github.com/cmunley1/trl/blob/9dff29127f294a46df135f440e78a0a5403cdf1c/examples/scripts/nemo_gym/nemo_gym_utils.py#L232
causing metrics to clear after each sample.

This seemed to resolve it, but please let me know if there is a recommended solution existing in TRL! Thanks

cmunley1: ill just simplify things and remove this.

Issue #5090, opened 2026-02-12
RFC: Defining the TRL Public API (towards v1.0)


qgallouedec:
## Motivation

As we approach TRL v1.0, we must explicitly define and freeze the public API surface.

The objective is to clearly distinguish:

* What we commit to supporting long-term
* What remains internal implementation detail

A stable v1 implies a compatibility commitment. Therefore, the public surface must be intentional, minimal, and sustainable.

## What is the "public API"?

For this RFC, the public API is defined as:

> Any function or class that appears in the official API documentation and is intended for direct user use.

Anything importable but not documented in the API reference is considered internal and may change without notice.

In other words, documentation acts as the contract.

## Compatibility Guarantees (v1)

For public symbols:

* Removing them is a breaking change (should wait until v2).
* Signature changes are breaking changes.
* Behavior changes that would reasonably break user workflows should follow a deprecation cycle.

Internal symbols:

* May change, move, or be removed without deprecation.
* Are not covered by stability guarantees.

## Decision Criteria

For each function/class, we should ask:

1. Is this intended for direct user use?
2. Is there a clear, stable, long-term use case?
3. Are we willing to maintain this indefinitely?
4. Would removing or changing it reasonably break user code?
5. Does exposing it meaningfully reduce user friction?

If unclear → default to **internal**.

Public APIs should be:

* Minimal
* Well-motivated
* Stable by design
* Worth their long-term maintenance cost

## Goal of this RFC

Review all currently importable symbols and decide:

* For currently public symbols: keep or remove?
* For currently internal symbols: promote to public or keep internal?

The outcome will define the official TRL v1 API surface.

### Trainers and configs

| Name (function/class) | Primary import path | Currently in public API | Should be in public API | Notes or questions |
| --- | --- | --- | --- | --- |
| `DPOConfig` | `trl.DPOConfig` | ✅ | ✅ | |
| `DPOTrainer` | `trl.DPOTrainer` | ✅ | ✅ | |
| `GRPOConfig` | `trl.GRPOConfig` | ✅ | ✅ | |
| `GRPOTrainer` | `trl.GRPOTrainer` | ✅ | ✅ | |
| `KTOConfig` | `trl.KTOConfig` | ✅ | ✅ | |
| `KTOTrainer` | `trl.KTOTrainer` | ✅ | ✅ | |
| `RewardConfig` | `trl.RewardConfig` | ✅ | ✅ | |
[...]

PR #5158, opened 2026-02-24
Add BaseTrainer and BaseConfig to docs

albertvillanova:
Add BaseTrainer and BaseConfig to docs.

### DIFF
--- a/docs/source/base_trainer.md
+++ b/docs/source/base_trainer.md
@@ -2,9 +2,9 @@
 
 ## BaseTrainer
 
-[[autodoc]] BaseTrainer
+[[autodoc]] trainer.base_trainer.BaseTrainer
     - create_model_card
 
 ## BaseConfig
 
-[[autodoc]] BaseConfig
+[[autodoc]] trainer.base_config.BaseConfig


qgallouedec:
Related to https://github.com/huggingface/trl/issues/5090

I’m slightly hesitant to add `BaseTrainer` and `BaseConfig` to the docs.
My concern is that documenting them implicitly makes them part of the public extension surface. That raises the bar for stability: changes like moving arguments upstream, adjusting defaults at the base level would then become user-visible API changes affecting all subclasses, not just our internal trainers.

Right now, keeping them undocumented gives us more flexibility to refactor or consolidate shared logic without committing to a long-term contract around inheritance.
Maybe we should only document them if we explicitly want to encourage third-party trainers/configs to inherit from them as a supported pattern. Otherwise, it might be safer to keep them as internal scaffolding.

albertvillanova:
Totally agree with you: these classes should be kept as private.

I'm closing this PR and opening another one to address the underlying issue: document the parameters with default values different from `TrainingArguments`:
- #5168

PR #5161, opened 2026-02-24
Add minimal CARLA example script

sergiopaniego:
# What does this PR do?

This PR adds an example for training with the CARLA simulator (autonomous driving) in OpenEnv.
Because CARLA is hardware-intensive, each simulation requires a dedicated GPU. To satisfy the minimum requirement of `num_generations=2`, you need at least two Spaces running in parallel.

OpenEnv PR: https://github.com/meta-pytorch/OpenEnv/pull/378
Spaces:
- https://huggingface.co/spaces/sergiopaniego/carla-env
- https://huggingface.co/spaces/sergiopaniego/carla-env-2

Command:

```
python examples/scripts/openenv/carla.py --env-urls https://sergiopaniego-carla-env.hf.space https://sergiopaniego-carla-env-2.hf.space
```

## Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/examples/scripts/openenv/carla.py
+++ b/examples/scripts/openenv/carla.py
@@ -181,6 +181,7 @@ class CarlaGRPOEnv:
         self.reward = result.observation.rubric_reward or 0.0
         return self._describe(result.observation)
 
+
 def reward_func(completions, environments, **kwargs):
     return [environment.reward for environment in environments]
 


sergiopaniego:
Ready for review! 😄 @qgallouedec

examples/scripts/openenv/carla.py
```diff
@@ -0,0 +1,221 @@
...
+
+dataset = Dataset.from_dict({"prompt": [[{"role": "user", "content": prompt}] for _ in range(1000)]})
+
+
+SIM_TICKS = 10  # Number of simulation steps to advance after each action
+
+
+class CarlaGRPOEnv:
+    def __init__(self):
+        url = next(_env_url_iter)
```
qgallouedec, line 115: One env only accept one connection? did you try to leverage `SUPPORTS_CONCURRENT_SESSIONS` and `max_concurrent_envs=8,`? like in https://huggingface.co/spaces/qgallouedec/wordle/commit/5b665ea4fd09d20e5ebd5e807283cbc257749ec6 and https://huggingface.co/spaces/qgallouedec/wordle/commit/59ad8468a47cc388bb80b9f7797fef7a8810e35d

examples/scripts/openenv/carla.py
```diff
@@ -0,0 +1,221 @@
...
+        return self._describe(result.observation)
+
+    def get_reward(self) -> float:
+        """
+        Get the reward from the last step.
+
+        Returns:
+            The rubric reward value.
+        """
+        return self._reward
```
qgallouedec: If you think that the model shouldn't have access to this, you can 
- remove this method
- `self._reward` -> `self.reward`
- in the `reward_func`: `[environment.reward for environment in environments]`

examples/scripts/openenv/carla.py
```diff
@@ -0,0 +1,221 @@
...
+
+prompt = """You control an autonomous vehicle in an emergency. There are pedestrians ahead and you must \
+decide what to do immediately.
+
+You have the following tools available:
+- `observe`: Advance time and get a new observation of the scene.
+- `emergency_stop`: Apply maximum braking to stop the vehicle.
+- `lane_change(direction)`: Change lane to the left or right. Direction must be "left" or "right".
+
+Observe the scene first, then decide the best course of action to minimize harm."""
```
qgallouedec, line 105: out of curiosity, is this prompt redundant with what's returned by `self._describe(result.observation)`? if so I think you can minimize this prompt to only give what is not in `self._describe(result.observation)`

sergiopaniego: Not in this case since the prompt gives the general idea and the description retrieved from `_describe` is specific for each step. Something like:

```
Speed: 30.0 km/h.
  - pedestrian at 15m
  - pedestrian at 20m
  No collision detected.
```

It could be extended depending on the scenario (or if the model has camera/vision support)

examples/scripts/openenv/carla.py
sergiopaniego: The problem here is that CARLA is a full simulator running inside Unreal Engine 5, so it's not as simple as enabling concurrent sessions. That would require running multiple simulations (each with its own 3D world) on the same GPU simultaneously, which isn't feasible. That's why I use a separate server URL per environment instance.

PR #5163, opened 2026-02-24
Fix CI by removing liger-kernel from dev deps

qgallouedec:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #5164, opened 2026-02-25
Re-add liger-kernel to dev deps

qgallouedec:
This PR reverts #5163. liger-kernel was temporarily removed from dev dependency because its installation caused a lot or permission error in the CI. (Which should be still the case in this CI, until we discover why)


qgallouedec:
Not sure what happened yesterday, but the problem seems solved

PR #5172, opened 2026-02-25
Document parameters with differing default values in experimental configs

albertvillanova:
Document parameters with differing default values in experimental configs.

Follow-up to:
- #5168

PR #5173, opened 2026-02-25
Fix default learning_rate in BCO according to paper

albertvillanova:
Fix default learning_rate in BCO according to paper: `5e-7` instead of `5e-5`
- [Binary Classifier Optimization for Large Language Model Alignment](https://huggingface.co/papers/2404.04656)

Follow-up to:
- #5172


albertvillanova:
The CI failing tests are unrelated. See:
- #5171

qgallouedec:
Same here, please remove the `--learning_rate 1e-6 \` in the bco example. It seems like the value used in the example doesn't match the value from the paper. Unless we have a good reason, I'd use the new default `5e-7` in the examples

PR #5186, opened 2026-02-26
Refactor CLI [5/N]: Refactor TrainingCommand with delayed imports

albertvillanova:
Refactor TrainingCommand for lazy import.

This PR refactors how training commands are registered and executed in the TRL CLI. The changes simplify the code by removing explicit script references and parser imports, making command registration more generic and maintainable. Now, each training command dynamically loads its corresponding script and parser **at runtime**.

Refactoring of training command registration and execution:

* Removed explicit imports of parser functions for each training script from `trl/cli/commands/__init__.py`, reducing manual maintenance and improving modularity.
* Updated the `get_commands()` function to instantiate `TrainingCommand` objects with only the command name, eliminating the need for script filenames and parser functions as arguments.
* Refactored the `TrainingCommand` class in `trl/cli/commands/training.py` to dynamically import the corresponding script module and retrieve its `make_parser()` function at runtime, instead of relying on statically passed parser functions.

These changes make it easier to add new training commands and reduce boilerplate code in the CLI command registration.

PR #5187, opened 2026-02-26
Refactor CLI [6/N]: Refactor env/vllm-serve commands with delayed imports

albertvillanova:
Refactor `env` and `vllm-serve` commands with delayed imports.

The latency of `trl -h` changes from `~7s` to `<0.5s` after the merge of both:
- #5186 
- #5187

This PR refactors import statements in the CLI command modules to move certain imports inside the `run` methods. This change helps avoid unnecessary imports when the command modules are loaded but not executed, which improves startup performance and reduce potential side effects.

**Refactoring of import statements in CLI commands:**

* Moved the import of `print_env` from `trl/scripts/env.py` inside the `run` method of the `env` CLI command to delay the import until execution. 
* Moved the imports of `main` and `make_parser` from `trl/scripts/vllm_serve.py` inside the `run` method of the `vllm_serve` CLI command to delay these imports until execution.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -130,7 +130,8 @@ class TestCLI(TrlTestCase):
         with open(config_path, "w") as f:
             yaml.dump({"model": "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5"}, f)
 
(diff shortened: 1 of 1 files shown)


qgallouedec (approved):
nice!

PR #4519, opened 2025-11-13
[GRPO] switch grpo liger loss to triton version

kashif:
# What does this PR do?

Fixes dapo loss type support, but needs to wait for upstream merge and release 

- [x] https://github.com/linkedin/Liger-Kernel/pull/939
- [x] https://github.com/linkedin/Liger-Kernel/pull/993
- [ ] liger-kernel release

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

---

> [!NOTE]
> **Medium Risk**
> Swaps out the Liger-kernel loss implementation and its call/metric wiring, which can change training behavior and stability across loss types and modalities. Risk is moderated by expanded test coverage around Triton invocation and parameter updates.
> 
> **Overview**
> Switches `GRPOTrainer`’s `use_liger_kernel` path from the old `LigerFusedLinearGRPOLoss` (hidden-state + lm_head) implementation to Liger’s Triton-based `triton_grpo_loss`, computing logits directly and passing through additional loss controls (e.g. `delta`, SA/PO temperatures, bias-corrected KL, importance-sampling ratio).
> 
> Removes the `delta`/two-sided-loss restriction for Liger in `GRPOConfig`, drops the forward-redirection and `_get_last_hidden_state` plumbing, and adjusts metric extraction to match Triton’s returned metric layout.
> 
> Updates tests to patch/verify `triton_grpo_loss` calls (including `vllm_is_ratio` propagation), broadens VLM+Liger coverage across multiple `loss_type`s, and removes assertions tied to the previous fused-loss class.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6dbb0208345467d781b3303eb052416a0811c106. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #5190, opened 2026-02-26
Support sequence sampling in Liger Kernel and pass importance_samplin…

michaelroyzen:
### Summary

- Liger-Kernel's `LigerFusedLinearGRPOLoss` already supports both `"token"` and `"sequence"` importance sampling levels, but TRL's integration was gated behind a stale `NotImplementedError` and never forwarded the setting.
- Removed the blanket rejection of non-token importance sampling when using Liger, replacing it with a validation check for actually unsupported values.
- Passed `importance_sampling_level` through to the `LigerFusedLinearGRPOLoss` constructor so the trainer's setting is actually respected.

### Motivation

`GRPOTrainer` supports `importance_sampling_level="sequence"` (used by e.g. LUSPO loss), but unconditionally raised `NotImplementedError` when combined with `use_liger_kernel=True`. This forced users to choose between Liger's memory/speed benefits and sequence-level importance sampling, even though the underlying Liger loss function handles both levels correctly (see `LigerFusedLinearGRPOFunction.ppo_loss_fn` and the `fused_linear_grpo_loss_sequence` benchmarks in Liger-Kernel).

### Changes

1. **`trl/trainer/grpo_trainer.py` (validation guard):** Replaced the `NotImplementedError` that rejected all non-`"token"` importance sampling with a `ValueError` that only rejects truly invalid values (anything other than `"token"` or `"sequence"`).

2. **`trl/trainer/grpo_trainer.py` (Liger loss construction):** Added `importance_sampling_level=self.importance_sampling_level` to the `LigerFusedLinearGRPOLoss(...)` constructor call, which previously omitted it and silently defaulted to `"token"`.

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -2447,16 +2447,14 @@ class TestGRPOTrainerSlow(TrlTestCase):
             "trl-internal-testing/tiny-MistralForCausalLM-0.2",
         ],
     )
-    @pytest.mark.parametrize("importance_sampling_level", ["token", "sequence"])
     @require_liger_kernel
-    def test_training_with_liger_grpo_kernel(self, model_name, importance_sampling_level):
+    def test_training_with_liger_grpo_kernel(self, model_name):
         training_args = GRPOConfig(
             output_dir=self.tmp_dir,
             per_device_train_batch_size=3,
             num_generations=3,
             use_liger_kernel=True,
             max_completion_length=self.max_length,
-            importance_sampling_level=importance_sampling_level,
             report_to="none",
             logging_strategy="no",
         )
@@ -2494,10 +2492,9 @@ class TestGRPOTrainerSlow(TrlTestCase):
             "trl-internal-testing/tiny-MistralForCausalLM-0.2",
         ],
     )
-    @pytest.mark.parametrize("importance_sampling_level", ["token", "sequence"])
     @require_liger_kernel
     @require_peft
-    def test_training_with_liger_grpo_kernel_and_peft(self, model_name, importance_sampling_level):
+    def test_training_with_liger_grpo_kernel_and_peft(self, model_name):
         from peft import LoraConfig, TaskType
 
         training_args = GRPOConfig(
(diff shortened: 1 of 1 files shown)


michaelroyzen:
@albertvillanova can I get a review please? the changes here are trivial but important to remove an outdated NotImplementedError for sequence-level sampling in GRPO.

kashif:
thanks @michaelroyzen also note that i have been adding other methods too in this comprehensive RP https://github.com/linkedin/Liger-Kernel/pull/993 together with: https://github.com/huggingface/trl/pull/4519/

kashif:
yes i am just checking if there some combinatio of loss type and `importance_sampling_level` that will cause an issue in liger 0.7, @michaelroyzen would you mind adding this `importance_sampling_level` to the set of test params for the liger grpo tests?

kashif:
@albertvillanova we always pin the liger version to the current one that supports the different features, which is i believe 0.7 right now

albertvillanova:
Noted! Thanks, @kashif! :hugs:

michaelroyzen:
Thank you @kashif @albertvillanova. I've added `importance_sampling_level` to the set of test params and appreciate your broader efforts to support more kinds of RL losses in Liger. I agree that this PR is independent because it is limited in scope to only enabling functionality that is already present in Liger via a function that we are already invoking.

Are we ready to merge?

tests/test_grpo_trainer.py
```diff
@@ -2447,14 +2447,16 @@ def teardown_method(self):
...
     @require_liger_kernel
-    def test_training_with_liger_grpo_kernel(self, model_name):
+    def test_training_with_liger_grpo_kernel(self, model_name, importance_sampling_level):
         training_args = GRPOConfig(
             output_dir=self.tmp_dir,
             per_device_train_batch_size=3,
             num_generations=3,
             use_liger_kernel=True,
             max_completion_length=self.max_length,
+            importance_sampling_level=importance_sampling_level,
```
qgallouedec: instead of parametrizing the existing tests, can you instead create one new test dedicated to important sampling?

kashif: good idea! added!

michaelroyzen:
Ready to merge? @albertvillanova @kashif?

PR #5197, opened 2026-02-27
Set CI PYTORCH_ALLOC_CONF env variable to avoid OOM

albertvillanova:
Set CI `PYTORCH_ALLOC_CONF` env variable to avoid OOM: https://github.com/huggingface/trl/actions/runs/22486411322/job/65137047342

> torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 2.60 GiB. GPU 0 has a total capacity of 14.74 GiB of which 2.44 GiB is free. Process 21197 has 366.00 MiB memory in use. Process 21194 has 8.07 GiB memory in use. Process 21200 has 3.40 GiB memory in use. Process 21203 has 484.00 MiB memory in use. Of the allocated memory 7.85 GiB is allocated by PyTorch, and 86.90 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)

### DIFF
--- a/.github/workflows/tests-experimental.yml
+++ b/.github/workflows/tests-experimental.yml
@@ -10,6 +10,7 @@ on:
 env:
   TQDM_DISABLE: 1
   PYTORCH_CUDA_ALLOC_CONF: "expandable_segments:True"
+  PYTORCH_ALLOC_CONF: "expandable_segments:True"
   TRL_EXPERIMENTAL_SILENCE: 1
 
 jobs:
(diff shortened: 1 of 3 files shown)


qgallouedec:
Ok, sharing what I learned:

PyTorch uses a caching allocator for CUDA memory. Instead of freeing memory back to CUDA immediately, it keeps it cached to reuse later. `PYTORCH_ALLOC_CONF` allows to modify how this allocator behaves. `PYTORCH_ALLOC_CONF=expandable_segments:True` Allows allocator segments to grow dynamically.

PR #5204, opened 2026-03-02
CI: Add Qwen 3.5 tiny model to tests

qgallouedec:
# What does this PR do?

This PR adds Qwen 3.5 tiny model to test. 

It was released yesterday, see https://huggingface.co/collections/Qwen/qwen35

Good news is: no modification needed in the code!

### DIFF
--- a/tests/test_data_utils.py
+++ b/tests/test_data_utils.py
@@ -481,6 +481,13 @@ class TestApplyChatTemplate(TrlTestCase):
         "trl-internal-testing/tiny-Phi3ForCausalLM",
         "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
         "trl-internal-testing/tiny-Qwen3ForCausalLM",
+        pytest.param(
+            "trl-internal-testing/tiny-Qwen3_5ForConditionalGeneration",
+            marks=pytest.mark.skipif(
+                Version(transformers.__version__) < Version("5.0.0"),
+                reason="Qwen3.5 tokenizer requires transformers>=5.0.0",
+            ),
+        ),
     ]
 
     conversational_examples = [


scripts/generate_tiny_models.py
```diff
@@ -325,9 +327,10 @@ def init_weights_tiny_model(model):
     ("Qwen/Qwen2-VL-2B-Instruct", Qwen2VLForConditionalGeneration, torch.bfloat16),
     ("Qwen/Qwen2.5-VL-3B-Instruct", Qwen2_5_VLForConditionalGeneration, torch.bfloat16),
     ("Qwen/Qwen3-VL-2B-Instruct", Qwen3VLForConditionalGeneration, torch.bfloat16),
+    ("Qwen/Qwen3.5-0.8B", Qwen3_5ForConditionalGeneration, torch.bfloat16),
 ]:
     processor = AutoProcessor.from_pretrained(model_id)
-    generation_config = GenerationConfig.from_pretrained(model_id)
+    generation_config = GenerationConfig.from_pretrained(model_id) if model_id != "Qwen/Qwen3.5-0.8B" else None
```
qgallouedec, line 333: for some reason, this model doesn't have a generation config

scripts/generate_tiny_models.py
```diff
@@ -371,14 +374,37 @@ def init_weights_tiny_model(model):
...
     config = AutoConfig.from_pretrained(model_id, text_config=text_config, vision_config=vision_config, **kwargs)
     model = model_class(config).to(dtype=dtype)
-    push_to_hub(model, processor, generation_config, "tiny")
+
+    if issubclass(model_class.config_class, Qwen3_5Config):
+        # Qwen3.5 models has some weights in float32, to mirror this in the tiny model we need to convert them to float32 manually.
+        for layer in model.model.language_model.layers:
+            if hasattr(layer, "linear_attn"):  # applies to linear attention layers only
+                layer.linear_attn.A_log.data = layer.linear_attn.A_log.data.float()
+                layer.linear_attn.norm.weight.data = layer.linear_attn.norm.weight.data.float()
```
qgallouedec, line 405: It's the first time we have this: Qwen3.5 models have two specific layers in fp32. Although not super elegant, it's the simplest way to force it at this point.

scripts/generate_tiny_models.py
```diff
@@ -371,14 +374,37 @@ def init_weights_tiny_model(model):
...
+        text_config["full_attention_interval"] = 2
+        # Qwen3.5-VL vision config expects `depth`/`num_heads`, not `num_hidden_layers`/`num_attention_heads`.
+        vision_config.pop("num_hidden_layers", None)
+        vision_config.pop("num_attention_heads", None)
+        vision_config.pop("num_key_value_heads", None)
+        vision_config.pop("embed_dim", None)
+        vision_config["depth"] = 2
+        vision_config["num_heads"] = 4
+        vision_config["intermediate_size"] = 32
+        vision_config["out_hidden_size"] = 16
```
qgallouedec, line 390: TBH, I start to dislike the structure of this file, at some point we should consider refactoring it.

tests/test_data_utils.py
```diff
@@ -481,6 +481,13 @@ class TestApplyChatTemplate(TrlTestCase):
         "trl-internal-testing/tiny-Phi3ForCausalLM",
         "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
         "trl-internal-testing/tiny-Qwen3ForCausalLM",
+        pytest.param(
+            "trl-internal-testing/tiny-Qwen3_5ForConditionalGeneration",
+            marks=pytest.mark.skipif(
+                Version(transformers.__version__) < Version("5.0.0"),
+                reason="Qwen3.5 tokenizer requires transformers>=5.0.0",
```
albertvillanova, line 488: Doesn't it require v5.2.0?

```suggestion
                Version(transformers.__version__) < Version("5.2.0"),
                reason="Qwen3.5 tokenizer requires transformers>=5.2.0",
```

albertvillanova:
Maybe renaming this PR title to mention it is scoped for the CI, something like:
- Title:
```
CI: Add Qwen 3.5 tiny model to tests
```

tests/test_data_utils.py
qgallouedec: To be precise, the tokenizer can be loaded with 5.0.0 and 5.1.0. In practice it's not useful because the model requires 5.2.0, but I think it's more accurate to test like this

PR #5205, opened 2026-03-02
Add support for Qwen3.5 for agent training

qgallouedec:
## Add Qwen3.5 support for agent training
This PR adds end-to-end support for [Qwen3.5](https://huggingface.co/collections/Qwen/qwen35) models in TRL's agent training pipeline (GRPO with tools). It follows #5204.
Qwen3.5 differs from Qwen3 in two key ways that drive the changes:
1. **Different tool call format** — Qwen3 uses JSON inside `<tool_call>` tags, while Qwen3.5 uses nested XML (`<function=name>` + `<parameter=key>value</parameter>`).
2. **Thinking is off by default** — Qwen3 generates `<think>` blocks by default (opt-out with `enable_thinking=False`), while Qwen3.5 suppresses thinking by default (opt-in with `enable_thinking=True`).
### What's included
**Chat template & response parsing (`trl/chat_template_utils.py`)**
- Added `qwen35_schema` for parsing Qwen3.5's XML-style tool call format using `x-regex-key-value` to match `<parameter=key>value</parameter>` pairs.
- Added `qwen35_chat_template` (with vision support) and `qwen35_training_chat_template` — the latter ensures prefix-preserving behavior during training by always including `<think>` blocks on every assistant turn, mirroring the existing Qwen3 approach.
- Wired up `add_response_schema()` and `get_training_chat_template()` to recognize Qwen3.5 templates.
- Relaxed `_validate_tool_calls` to normalize missing/null `arguments` to `{}` instead of raising `ValueError`. This handles Qwen3.5's valid zero-argument tool calls where `<parameter>` tags are simply omitted. The actual failure (unknown tool, missing required args) is deferred to the tool execution step, which is more correct.
**Tiny model generation (`scripts/generate_tiny_models.py`)**
- Added `Qwen3_5ForConditionalGeneration` tiny model generation with appropriate config overrides (explicit `layer_types` to ensure a full-attention layer exists, vision config key remapping, float32 weight conversion for linear attention layers).
**Tests**
- Added `trl-internal-testing/tiny-Qwen3_5ForConditionalGeneration` as a test model across SFT, DPO, GRPO, RLOO, chat template, data utils, and utility tests. All gated behind `transformers>=5.2.0` (or `>=5.0.0` for tokenizer-only tests).
- Updated GRPO malformed tool call test tokens — previously valid-but-incomplete calls now parse correctly under the relaxed validation, so truly malformed tokens (wrong XML tags, unquoted JSON keys) are used instead.
- Added `test_parse_response_tool_call_without_arguments` to cover no-arg tool calls.
- Removed `test_behavior_unchanged_assistant_with_tool_calls_with_string_arguments` since Qwen3.5's template doesn't support string-type `arguments` (it uses `|items` which requires a mapping).
**Documentation (`docs/source/grpo_trainer.md`)**
- Listed Qwen3.5 as a supported model for agent training alongside Qwen3, with links to model collections.
### Key design decisions
- The Qwen3.5 training template is derived from the original via `str.replace()` (same pattern as Qwen3), applying two targeted modifications: (1) require both `<think>` and `</think>` tags to be present before parsing, and (2) always emit `<think>` blocks on all assistant turns, not just the last one. Each replacement target appears exactly once in the template.
- Missing `arguments` in tool calls are normalized to `{}` rather than erroring, since Qwen3.5 legitimately omits them for no-arg functions. This is a behavioral change for all templates — previously these would have raised `ValueError` and fallen back to plain text parsing.

---

<a href="https://app.devin.ai/review/huggingface/trl/pull/5205" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>


tests/test_chat_template_utils.py
```diff
@@ -302,21 +318,6 @@ def test_behavior_unchanged_assistant_with_tool_calls(self, tokenizer_name):
...
-                "role": "assistant",
-                "content": "I will call a tool.",
-                "tool_calls": [{"name": "multiply", "arguments": '{"a": 3, "b": 4}'}],
-            },
-        ]
-        before = tokenizer.apply_chat_template(messages, tokenize=False)
-        new_chat_template = get_training_chat_template(tokenizer)
-        after = tokenizer.apply_chat_template(messages, tokenize=False, chat_template=new_chat_template)
-        assert before == after
-
```
qgallouedec, line 319: This test fails with Qwen3.5.

```
_______________________ TestGetTrainingChatTemplate.test_behavior_unchanged_assistant_with_tool_calls_with_string_arguments[qwen35] _______________________

self = <tests.test_chat_template_utils.TestGetTrainingChatTemplate object at 0x7f73b091be90>
tokenizer_name = 'trl-internal-testing/tiny-Qwen3_5ForConditionalGeneration'

    def test_behavior_unchanged_assistant_with_tool_calls_with_string_arguments(self, tokenizer_name):
        tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
        messages = [
            {"role": "user", "content": "Multiply 3 by 4."},
            {
                "role": "assistant",
                "content": "I will call a tool.",
                "tool_calls": [{"name": "multiply", "arguments": '{"a": 3, "b": 4}'}],
            },
        ]
>       before = tokenizer.apply_chat_template(messages, tokenize=False)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/test_chat_template_utils.py:331: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../transformers/src/transformers/tokenization_utils_base.py:3058: in apply_chat_template
    rendered_chat, generation_indices = render_jinja_template(
../transformers/src/transformers/utils/chat_template_utils.py:555: in render_jinja_template
    rendered_chat = compiled_template.render(
../miniconda3/envs/trl/lib/python3.13/site-packages/jinja2/environment.py:1295: in render
    self.environment.handle_exception()
../miniconda3/envs/trl/lib/python3.13/site-packages/jinja2/environment.py:942: in handle_exception
    raise rewrite_traceback_stack(source=source)
<template>:120: in top-level template code
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

value = '{"a": 3, "b": 4}'

    def do_items(value: t.Union[t.Mapping[K, V], Undefined]) -> t.Iterator[t.Tuple[K, V]]:
        """Return an iterator over the ``(key, value)`` items of a mapping.
    
        ``x|items`` is the same as ``x.items()``, except if ``x`` is
        undefined an empty iterator is returned.
    
        This filter is useful if you expect the template to be rendered with
        an implementation of Jinja in another programming language that does
        not have a ``.items()`` method on its mapping type.
    
        .. code-block:: html+jinja
    
            <dl>
            {% for key, value in my_dict|items %}
                <dt>{{ key }}
                <dd>{{ value }}
            {% endfor %}
            </dl>
    
        .. versionadded:: 3.1
        """
        if isinstance(value, Undefined):
            return
    
        if not isinstance(value, abc.Mapping):
>           raise TypeError("Can only get item pairs from a mapping.")
E           TypeError: Can only get item pairs from a mapping.

../miniconda3/envs/trl/lib/python3.13/site-packages/jinja2/filters.py:249: TypeError
```

After second thought, I don't think it's an important one, the right way to provide arguments is to use a mapping, as in the test just before. And we never rely on the ability to provide arguments as string in the codebase.

trl/chat_template_utils.py
```diff
@@ -143,6 +143,40 @@ def clone_chat_template(
     },
 }
 
+qwen35_schema = {
+    "x-regex": r"^(?:(?:<think>\n?)?(?:(?P<reasoning_content>.*?\S.*?)\n?|[\s]*)</think>\s*)?(?P<content>.*?)(?:\n+(?=<tool_call>))?(?=(?:<tool_call>|<\|im_end\|>|$))(?P<tool_calls>(?:<tool_call>.+?</tool_call>\s*)+)?\s*(?:<\|im_end\|>|$)",
```
qgallouedec, line 147: no idea how it works, but it works. thanks codex

qgallouedec:
@codex review

tests/test_grpo_trainer.py
```diff
@@ -2405,11 +2405,11 @@ def fake_generate(input_ids, **kwargs):
             completion_ids = torch.tensor(
                 [
                     # '<tool_call>\n{"arguments": {"a": 3, "b": 4}}\n</tool_call><|im_end|>'
-                    [151657, 198, 4913, 16370, 788, 5212, 64, 788, 220, 18, 11, 330, 65, 788, 220, 19, 11248, 151658, 151645],
-                    # '<tool_call>\n{"name": "multiply_tool"}\n</tool_call><|im_end|>'
-                    [151657, 198, 4913, 606, 788, 330, 64648, 22785, 16707, 151658, 151645, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643],
-                    # "<tool_call>\n{}\n</tool_call><|im_end|>"
-                    [151657, 198, 16094, 151658, 151645, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643, 151643],
```
qgallouedec, line 2466: because of the changes in `_validate_tool_calls`, `'<tool_call>\n{"name": "multiply_tool"}\n</tool_call><|im_end|>'` and `"<tool_call>\n{}\n</tool_call><|im_end|>"` will both be properly parsed, but it's the calling that will fail (first case: missing args, second case, unknown tool). Which I think it's more correct.

I've changed the tokens to have true malformed tools: `toolcall` instead of `tool_call` and `{a: 3, b: 4}` instead of `{"a": 3, "b": 4}`

trl/chat_template_utils.py
```diff
@@ -397,13 +591,28 @@ def is_chat_template_prefix_preserving(tokenizer: PreTrainedTokenizer) -> bool:
...
+# - {{- '<|im_start|>' + message.role + '\n' + content }}
+# + {{- '<|im_start|>' + message.role + '\n<think>\n' + reasoning_content + '\n</think>\n\n' + content }}
+#   Always include thinking block during training. It's important to have a prefix-preserving template.
+qwen35_training_chat_template = qwen35_chat_template.replace(
+    "{%- if '</think>' in content %}",
+    "{%- if '<think>' in content and '</think>' in content %}",
+).replace(
+    "{{- '<|im_start|>' + message.role + '\\n' + content }}",
+    "{{- '<|im_start|>' + message.role + '\\n<think>\\n' + reasoning_content + '\\n</think>\\n\\n' + content }}",
+)
```
qgallouedec, line 607: 📝 **Info: qwen35_training_chat_template constructed via fragile string replacement**

The `qwen35_training_chat_template` is constructed by chaining two `.replace()` calls on the raw `qwen35_chat_template` string (`trl/chat_template_utils.py:601-607`). I verified that each replacement target appears exactly once in the template, so the replacements are correct today. However, this approach is fragile compared to the Qwen3 training template which is defined as a standalone raw string literal. If the upstream Qwen3.5 chat template ever changes the whitespace or structure of these lines, the `.replace()` would silently become a no-op, producing a training template identical to the original (non-prefix-preserving) one. This could lead to subtle training issues that would be hard to debug.


<a href="https://app.devin.ai/review/huggingface/trl/pull/5205" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

trl/chat_template_utils.py
```diff
@@ -450,6 +659,8 @@ def get_training_chat_template(tokenizer: PreTrainedTokenizer) -> str | None:
 
     if tokenizer.chat_template == qwen3_chat_template:
         return qwen3_training_chat_template
+    if tokenizer.chat_template == qwen35_chat_template:
+        return qwen35_training_chat_template
     else:
         raise ValueError(
             "The tokenizer's chat template is not prefix-preserving and patching is not supported for this template. "
```
qgallouedec, line 666: 📝 **Info: if/if/else pattern in get_training_chat_template is functionally correct but misleading**

The new code at `trl/chat_template_utils.py:660-668` uses `if` / `if` / `else` instead of `if` / `elif` / `else`. The `else` on line 664 is syntactically bound only to the second `if` (line 662), not the first (line 660). This works correctly because: if the first `if` matches, we return immediately; if it doesn't match, we proceed to the second `if`; and the `else` catches any remaining case. However, the pattern is visually misleading — it appears the `else` covers all non-matching cases when in fact it's only paired with the second `if`. Using `elif` would be clearer and prevent a future contributor from accidentally inserting code between the two `if` blocks that would alter control flow.

*(Refers to lines 660-668)*


<a href="https://app.devin.ai/review/huggingface/trl/pull/5205" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

trl/chat_template_utils.py
```diff
@@ -234,6 +269,162 @@ def clone_chat_template(
...
+    {%- endif %}
+{%- endfor %}
+{%- if add_generation_prompt %}
+    {{- '<|im_start|>assistant\n' }}
+    {%- if enable_thinking is defined and enable_thinking is true %}
+        {{- '<think>\n' }}
+    {%- else %}
+        {{- '<think>\n\n</think>\n\n' }}
+    {%- endif %}
+{%- endif %}"""
```
qgallouedec, line 426: 📝 **Info: Qwen3.5 inverts thinking default compared to Qwen3 — verified prefix-preserving logic still works**

Qwen3.5's generation prompt defaults to thinking DISABLED (`<think>\n\n</think>\n\n`), while Qwen3 defaults to thinking ENABLED (no closed think block). This inversion is visible at `trl/chat_template_utils.py:419-425` vs `trl/chat_template_utils.py:265-269`. Despite this difference, the prefix-preserving check in `is_chat_template_prefix_preserving` correctly identifies Qwen3.5 as non-prefix-preserving: `text1` (with generation prompt) includes closed think tags, `text2` (last assistant after last user) also includes think tags (matching), but `text3` (non-last assistant) omits think tags (breaking the prefix property). The training template fix correctly makes both branches emit think tags. The test at `tests/test_chat_template_utils.py:417-418` correctly passes `enable_thinking=True` for the prefix in `test_parse_response_with_reasoning_content`, which is needed because Qwen3.5 defaults to closed thinking in the generation prompt, but this change is also safe for Qwen3 where `enable_thinking=True` produces the same output as not passing it.


<a href="https://app.devin.ai/review/huggingface/trl/pull/5205" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

PR #5209, opened 2026-03-03
Refactor CLI [8/N]: Refactor scripts/utils with delayed imports

albertvillanova:
Refactor scripts/utils with delayed imports

This PR refactors the `trl/scripts/utils.py` file to improve import management and ensure compatibility with type checking and dataset handling. The main focus is on moving imports inside functions to reduce unnecessary dependencies during module load, using string-based type annotations for better type checking, and updating dataset-related code for clarity and correctness.

### Import management and type checking improvements

* Moved imports for `datasets` and `yaml` inside relevant functions to reduce top-level dependencies and speed up module loading. 
* Used string-based type annotation `"DatasetDict"` in the `get_dataset` function signature to improve compatibility with static type checkers and avoid runtime import issues.

### Dataset handling updates

* Updated calls to `concatenate_datasets` and `DatasetDict` to use the `datasets` module namespace, ensuring clarity and correctness in dataset operations. 

### Other minor improvements

* Moved the import for `is_rich_available` inside the `init_zero_verbose` function to further reduce top-level imports and improve modularity.

PR #5210, opened 2026-03-03
Refactor CLI [9/N]: Replace HfArgumentParser from transformers with local

albertvillanova:
Replace HfArgumentParser from transformers with local.

This PR updates the import statements in `trl/scripts/utils.py` to use local versions of argument parsing utilities instead of those from the external `transformers` library. This change helps improving latency within the codebase.

The latency of `import trl.scripts.utils` changes from `7.5s` to `<0.5s` after the merge of:
- #5208
- #5209
- #5210

See upstream issue:
- https://github.com/huggingface/transformers/issues/44273

Dependency management:

* Replaced imports of `HfArgumentParser`, `DataClass`, and `DataClassType` from the `transformers` library with imports from the local `trl.scripts._hf_argparser` module.

---

> [!NOTE]
> **Medium Risk**
> Although intended as a drop-in copy, vendoring core CLI argument parsing code can subtly change parsing behavior and may drift from upstream fixes, impacting script configuration/flag handling.
> 
> **Overview**
> Introduces a vendored copy of `transformers`’ `hf_argparser` as `trl/scripts/_hf_argparser.py`, with `yaml` imported lazily in `parse_yaml_file` to avoid the upstream import-time latency.
> 
> Updates `trl/scripts/utils.py` to import `HfArgumentParser`/`DataClass`/`DataClassType` from the local module instead of `transformers`, keeping CLI parsing behavior local and reducing startup overhead.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2e15b40ff9ca0ac6fce4f2c41304c0466bbcebc1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/trl/scripts/_hf_argparser.py
+++ b/trl/scripts/_hf_argparser.py
@@ -14,6 +14,7 @@
 
 
 # Copied from: https://github.com/huggingface/transformers/blob/3a275d3581c0ecf962f7412aa764c2047331fd6b/src/transformers/hf_argparser.py
+# This avoids an upstream latency issue: https://github.com/huggingface/transformers/issues/44273
 # - Moved yaml import inside function
 
 


qgallouedec:
does it mean that when https://github.com/huggingface/transformers/issues/44273 is solved, we could just revert this commit?

albertvillanova:

Not exactly, because we still support older versions of transformers where the import latency is not fixed.

Instead, we could make a conditional import depending on the transformers version.
What do you think?

qgallouedec:
I'd recommend at least a comment explaining that it's copied from transformers and modified because of the latency (you already did it) + 
- a comment saying that this can be removed when min transformers version contains the fix of https://github.com/huggingface/transformers/issues/44273 OR
- a conditional import (maybe over-engineered here?)

up to you

PR #5227, opened 2026-03-05
Add VLM support when passing raw token IDs to vLLM client

qgallouedec:
## Context

Part of the series to fix the re-tokenization bug in GRPO multi-turn tool calling (see #5224).

When the model generates a completion in a tool-calling loop, the decoded text is re-tokenized via `apply_chat_template`, which can produce different token IDs due to BPE merge ambiguities. To fix this, we need a **token-in / token-out** pipeline: tokenize once, then pass raw token IDs through every subsequent generation call — never decoding and re-tokenizing.

- #5225
- [this one]
- #5232
- #5238
- #5239
- #5240
- #5242

## Changes

- **`vllm_serve.py`**: When `prompts` contains token ID lists, images are now attached as `multi_modal_data` in vLLM's `TokensPrompt` format, matching the existing behavior for text prompts.
- **Tests**: Add `TestVLLMClientServerVLM` test class that verifies generation with token IDs + images using `Qwen/Qwen2.5-VL-3B-Instruct`.

## Backward compatibility

Fully backward compatible. No changes to existing call sites or API signatures.

## Tests

```
$ pytest -v tests/test_vllm_client_server.py
========================================================== test session starts ===========================================================
platform linux -- Python 3.13.11, pytest-9.0.2, pluggy-1.6.0 -- /fsx/qgallouedec/miniconda3/envs/trl/bin/python3.13
cachedir: .pytest_cache
rootdir: /fsx/qgallouedec/trl
configfile: pyproject.toml
plugins: rerunfailures-15.1, anyio-4.12.1, xdist-3.8.0, datadir-1.8.0, cov-7.0.0
collected 42 items                                                                                                                       

tests/test_vllm_client_server.py::TestChunkList::test_even_split PASSED                                                            [  2%]
tests/test_vllm_client_server.py::TestChunkList::test_uneven_split PASSED                                                          [  4%]
tests/test_vllm_client_server.py::TestChunkList::test_more_chunks_than_elements PASSED                                             [  7%]
tests/test_vllm_client_server.py::TestChunkList::test_n_equals_len PASSED                                                          [  9%]
tests/test_vllm_client_server.py::TestChunkList::test_n_is_1 PASSED                                                                [ 11%]
tests/test_vllm_client_server.py::TestChunkList::test_single_element_list PASSED                                                   [ 14%]
tests/test_vllm_client_server.py::TestChunkList::test_any_dtype PASSED                                                             [ 16%]
[...]

PR #5225, opened 2026-03-05
Add support for raw ids in `prompts` in vLLM client and server

qgallouedec:
## Context

Part of the series to fix the re-tokenization bug in GRPO multi-turn tool calling (see #5224).

When the model generates a completion in a tool-calling loop, the decoded text is re-tokenized via `apply_chat_template`, which can produce different token IDs due to BPE merge ambiguities. To fix this, we need a **token-in / token-out** pipeline: tokenize once, then pass raw token IDs through every subsequent generation call — never decoding and re-tokenizing.

- [this one]
- #5227 
- #5232
- #5238
- #5239
- #5240
- #5242

To fix that, we need the ability to pass pre-tokenized prompts directly through the client/server pipeline. This PR adds that capability without changing any existing behavior.

## Changes

- **`VLLMClient.generate()`**: Add support for  `prompts` parameter being tokens. Existing callers using `prompts` are unaffected.
- **`vllm_serve.py`**: `GenerateRequest` now accepts both `prompts` as tokens.
- **Tests**: Add `test_generate_with_token_ids` across all test classes to cover the new code path.

## Backward compatibility

Fully backward compatible.

## Tests

```
$ pytest -v tests/test_vllm_client_server.py
========================================================== test session starts ===========================================================
platform linux -- Python 3.13.11, pytest-9.0.2, pluggy-1.6.0 -- /fsx/qgallouedec/miniconda3/envs/trl/bin/python3.13
cachedir: .pytest_cache
rootdir: /fsx/qgallouedec/trl
configfile: pyproject.toml
plugins: rerunfailures-15.1, anyio-4.12.1, xdist-3.8.0, datadir-1.8.0, cov-7.0.0
collected 41 items                                                                                       

tests/test_vllm_client_server.py::TestChunkList::test_even_split PASSED                                                            [  2%]
tests/test_vllm_client_server.py::TestChunkList::test_uneven_split PASSED                                                          [  4%]
tests/test_vllm_client_server.py::TestChunkList::test_more_chunks_than_elements PASSED                                             [  7%]
tests/test_vllm_client_server.py::TestChunkList::test_n_equals_len PASSED                                                          [  9%]
tests/test_vllm_client_server.py::TestChunkList::test_n_is_1 PASSED                                                                [ 12%]
tests/test_vllm_client_server.py::TestChunkList::test_single_element_list PASSED                                                   [ 14%]
tests/test_vllm_client_server.py::TestChunkList::test_any_dtype PASSED                                                             [ 17%]
[...]

Issue #5224, opened 2026-03-05
Re-tokenization bug in GRPO multi-turn tool calling


qgallouedec:
# Re-tokenization bug in GRPO multi-turn tool calling

## The bug

When `GRPOTrainer` runs multi-turn tool calling, each iteration of `_tool_call_loop` does this:

1. The model generates a completion (producing token IDs)
2. The token IDs are **decoded** to text
3. The text is appended as an assistant message to the conversation
4. The full conversation (prompt + completion + tool result) is re-tokenized via `apply_chat_template`
5. The re-tokenized IDs are passed to the next generation call

The problem is step 4: BPE is not a bijection. Decoding token IDs to text and re-tokenizing can produce different token IDs. For example, the token sequence `["he", "llo"]` decodes to `"hello"`, which re-tokenizes to `["hello"]`: different IDs, same text. This means the log-probabilities used for the policy gradient no longer correspond to the tokens that were actually sampled.

## The fix

**Tokenize once, never re-tokenize.** Build a token-in / token-out pipeline where raw token IDs flow through the entire generation loop without ever being decoded and re-tokenized.

In `_tool_call_loop`, instead of re-tokenizing the full conversation, build the next prompt by concatenation:

```
next_prompt_ids = prompt_ids + completion_ids + tool_suffix_ids
```

The original `prompt_ids` and `completion_ids` are preserved exactly. Only the tool result portion (template formatting + tool output) is freshly tokenized.

## PRs

The fix is split into 7 incremental PRs, each building on the previous:

1. **#5225** — Add `prompt_token_ids` support to vLLM client/server, so vLLM can receive pre-tokenized IDs
2. **#5227** — Extend the token ID path to support VLMs (images alongside token IDs)
3. **#5232** — Move `rollout_func` out of `_generate_single_turn` into `_generate` (prep refactor)
4. **#5238** — Move tokenization from `VLLMGeneration.generate` into `_generate_single_turn`, so vLLM always receives raw token IDs
5. **#5239** — Unify tokenization across all 3 backends (vLLM, paged, regular) at the top of `_generate_single_turn`
6. **#5240** — Extract `_tokenize_prompts()` method, make `_generate_single_turn` accept pre-tokenized inputs
7. **#5242** — The actual fix: replace re-tokenization in `_tool_call_loop` with token ID concatenation

PR #5238, opened 2026-03-07
[GRPO/RLOO] Tokenize before vLLM generation call

qgallouedec:
## Context

- #5225
- #5227
- #5232
- [this one]
- #5239
- #5240
- #5242

This PR moves tokenization out of `VLLMGeneration.generate` and into the trainers' `_generate_single_turn`, so that vLLM always receives raw token IDs instead of text or chat messages.

## Changes

- **`VLLMGeneration.generate()`**: Replace the `prompts` (text/messages) parameter with `prompts` (token ID lists) + `images` (optional PIL image lists). Remove the internal `chat()` / text-tokenization paths — the method now always forwards pre-tokenized IDs to vLLM. For colocate mode, build `{"prompt_token_ids": ids, "multi_modal_data": ...}` dicts. Remove unused imports (`json`, `is_conversational`, `prepare_multimodal_messages_vllm`).
- **GRPO `_generate_single_turn()`**: Tokenize prompts before calling vLLM. For conversational prompts, use `apply_chat_template(tokenize=True)` and extract PIL images from message content blocks. For text prompts, use `processing_class(text=prompts)`. Pass the resulting token IDs and images to `vllm_generation.generate()`.
- **RLOO `_generate_single_turn()`**: Same tokenization pattern applied.

## Why

Previously, `VLLMGeneration.generate` received raw text or chat messages and was responsible for tokenization (server mode) or delegating it to vLLM's `chat()` (colocate mode). This made it impossible to guarantee token-level consistency between the trainer's tokenization and vLLM's — a prerequisite for the token-in/token-out pipeline.

By tokenizing once in `_generate_single_turn` and passing only token IDs downstream, we ensure that:
1. The trainer controls tokenization (chat template, tools, special tokens) in one place.
2. vLLM receives exactly the token IDs the trainer expects, eliminating BPE ambiguity.
3. The VLM case is handled correctly: images are extracted from messages and passed separately via vLLM's `multi_modal_data` mechanism.

## Backward compatibility

The `VLLMGeneration.generate` signature changes from accepting text/messages to accepting token ID lists. This is an internal API — all call sites (GRPO and RLOO trainers) are updated in this PR. No user-facing API changes.

---

> [!NOTE]
> **Medium Risk**
> Touches core generation paths for GRPO/RLOO and changes vLLM input semantics, so regressions would affect sampling behavior and distributed generation (especially multimodal/image batches).
> 
> **Overview**
> Implements a **token-in/token-out** vLLM generation path for `GRPOTrainer` and `RLOOTrainer` by moving all prompt tokenization out of `VLLMGeneration.generate()` and into the trainers’ `_generate_single_turn()`.
> 
> `VLLMGeneration.generate()` now always receives pre-tokenized `prompt_token_ids` (plus optional per-prompt image lists for VLMs), removes the chat/template/tool-handling branches and JSON tool-arg coercion, and adds distributed-safe gathering of `images` to avoid collectives deadlocking when some ranks have no images. Colocate mode now builds vLLM prompt dicts (`prompt_token_ids` + `multi_modal_data`) before calling `llm.generate()`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fee553d557c2153063b528ed238dfbd5c5d3c3c9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/source/vllm_integration.md
+++ b/docs/source/vllm_integration.md
@@ -3,7 +3,7 @@
 This document will guide you through the process of using vLLM with TRL for faster generation in online methods like GRPO and Online DPO. We first summarize a tl;dr on how to use vLLM with TRL, and then we will go into the details of how it works under the hood.
 
 > [!WARNING]
-> TRL currently only supports vLLM versions from `0.10.2` to `0.16.0`. Please ensure you have a version in this range installed to avoid compatibility issues.
+> TRL currently only supports vLLM versions from `0.10.2` to `0.17.0`. Please ensure you have a version in this range installed to avoid compatibility issues.
 
 > [!TIP]
 > The following trainers currently support generation with vLLM:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,7 @@ test = [
     "pytest"
 ]
 vllm = [
-    "vllm>=0.10.2,<=0.16.0",
+    "vllm>=0.10.2,<=0.17.0",
     "fastapi",
     "pydantic",
     "requests",
--- a/trl/_compat.py
+++ b/trl/_compat.py
@@ -89,7 +89,7 @@ def _patch_vllm_disabled_tqdm() -> None:
 
     - Bug introduced in https://github.com/vllm-project/vllm/pull/52
     - Fixed in https://github.com/vllm-project/vllm/pull/28471 (released in v0.11.1)
-    - Since TRL currently supports vLLM v0.10.2-0.16.0, we patch it here
+    - Since TRL currently supports vLLM v0.10.2-0.17.0, we patch it here
     - This can be removed when TRL requires vLLM>=0.11.1
     """
     if _is_package_version_below("vllm", "0.11.1"):
--- a/trl/import_utils.py
+++ b/trl/import_utils.py
@@ -109,9 +109,9 @@ def is_uvicorn_available() -> bool:
 def is_vllm_available() -> bool:
     _vllm_available, _vllm_version = _is_package_available("vllm", return_version=True)
     if _vllm_available:
-        if not (Version("0.10.2") <= Version(_vllm_version) <= Version("0.16.0")):
+        if not (Version("0.10.2") <= Version(_vllm_version) <= Version("0.17.0")):
             warnings.warn(
-                f"TRL currently supports vLLM versions from 0.10.2 to 0.16.0. You have version {_vllm_version} "
+                f"TRL currently supports vLLM versions from 0.10.2 to 0.17.0. You have version {_vllm_version} "
                 "installed. We recommend installing a supported version to avoid compatibility issues.",
                 stacklevel=2,
             )


qgallouedec:
Before vs After

```python
# train_grpo.py
from trl.rewards import accuracy_reward

dataset = load_dataset("BlackBeenie/simple-math", split="train")

dataset = dataset.filter(lambda x: x["level"] == "Level 1")

def extract_solution(example):
    return {"solution": example["reward_model"]["ground_truth"]}

dataset = dataset.map(extract_solution)

training_args = GRPOConfig(
    output_dir="/tmp/grpo_test",
    use_vllm=True,
    vllm_mode="colocate",
    max_steps=50,
    logging_steps=1,
    report_to="trackio",
)

trainer = GRPOTrainer(
    reward_funcs=accuracy_reward,
    train_dataset=dataset,
    args=training_args,
)
trainer.train()
```

https://qgallouedec-trackio.hf.space?project=huggingface&runs=qgallouedec-1772852549,qgallouedec-1772852661&sidebar=hidden&navbar=hidden

everything is exactly the same

chart-161
chart-160
chart-159
chart-158
chart-157

qgallouedec:
@codex review

trl/generation/vllm_generation.py
```diff
@@ -567,9 +573,6 @@ def generate(self, prompts: list, num_generations: int, profiler: ProfilingConte
         min_p = self.min_p
         repetition_penalty = self.repetition_penalty
         max_completion_length = self.max_completion_length
-        chat_template_kwargs = self.chat_template_kwargs
-        tools = self.tools
-        chat_template = self.chat_template
```
albertvillanova, line 572: These attribute (`chat_template_kwargs`, `tools` and `chat_template`) are no longer used by `VLLMGeneration`. I think you could:
- Remove the params/attributes from VLLMGeneration: both in code and docstring
- Update GRPOTrainer and RLOOTrainer accordingly, so they do not pass these params

trl/generation/vllm_generation.py
```diff
@@ -528,13 +526,21 @@ def sync_weights(self):
         elif self.mode == "colocate":
             self.llm.reset_prefix_cache()
 
-    def generate(self, prompts: list, num_generations: int, profiler: ProfilingContext | None = None) -> tuple:
+    def generate(
+        self,
+        prompts: list[list[int]],
+        images: list[list] | None,
```
albertvillanova: Each per-prompt element can itself be None (correctly described in the docstring):
```suggestion
        images: list[list | None] | None,
```

trl/trainer/rloo_trainer.py
```diff
@@ -896,10 +896,41 @@ def _generate_single_turn(self, prompts: list):
...
+                    for message in prompt:
+                        if isinstance(message["content"], list):
+                            for part in message["content"]:
+                                if part["type"] == "image":
+                                    prompt_images.append(part["image"])
+                                    has_images = True
+                    images.append(prompt_images if prompt_images else None)
+                images = images if has_images else None
+
+                tokenized = self.processing_class.apply_chat_template(
```
albertvillanova, line 916: What about adding an explicit comment? Per the codebase's consistency rules, duplicated blocks should carry a comment at the divergence point
```suggestion
                # RLOO does not support tools; omit tools/chat_template args
                tokenized = self.processing_class.apply_chat_template(
```

trl/generation/vllm_generation.py
qgallouedec: good point, done! https://github.com/huggingface/trl/pull/5238/commits/b35f250156e6c21b89f26c46202cdc93a6d0bd4d

PR #5239, opened 2026-03-07
[GRPO/RLOO] Unify tokenization across all generation backends in `_generate_single_turn`

qgallouedec:
## Context

- #5225
- #5227
- #5232
- #5238
- [this one]
- #5240
- #5242

The previous PR moved tokenization out of `VLLMGeneration.generate` so that vLLM always receives raw token IDs. However, the paged and regular generation backends still performed their own independent tokenization. This PR unifies tokenization across **all three backends** (vLLM, paged, regular) by tokenizing once at the top of `_generate_single_turn`.

## Changes

- **GRPO `_generate_single_turn()`**: Move tokenization to the top of the method, before the backend dispatch. For conversational prompts, use `apply_chat_template(tokenize=True, return_dict=True)` and extract images + multimodal fields. For text prompts, use `processing_class(text=prompts)`. All three backends (vLLM, paged, regular) now consume the same `prompt_ids` list. Remove the per-backend tokenization that previously existed in the paged and regular paths.
- **RLOO `_generate_single_turn()`**: Same unification applied.
- **Regular generation path**: Build left-padded tensors from the shared `prompt_ids` using `pad()`. Convert multimodal fields (numpy arrays from `apply_chat_template`) to tensors before passing to `model.generate()`.

## Why

Previously, each generation backend tokenized prompts independently:

- vLLM path: tokenized in `_generate_single_turn` (added in PR4)
- Paged path: called `apply_chat_template` or `processing_class` internally
- Regular path: called `apply_chat_template(return_tensors="pt", padding=True)` or `processing_class(return_tensors="pt")` (we convert to tensor manually now)

This allows in the next PR to move the tokenization outside `_generate_single_turn` and into the trainers' `generate` method.

## Backward compatibility

No user-facing API changes. The only behavioral difference is that the regular and paged paths now use the same tokenization call as the vLLM path, which should produce identical results.

---

> [!NOTE]
> **Medium Risk**
> Changes core generation input preparation across three backends and introduces new tensor conversion/padding behavior (including multimodal fields), which could subtly alter generation outputs or break on processor edge cases.
> 
> **Overview**
> Unifies `_generate_single_turn` in `GRPOTrainer` and `RLOOTrainer` to **tokenize prompts once** and feed the same `prompt_ids` into all generation backends (vLLM, paged `generate_batch`, and standard `model.generate`), removing backend-specific tokenization.
> 
> For conversational/VLM inputs, it now uses `apply_chat_template(..., padding=True)` then *unpads* via `attention_mask`, carries through multimodal fields (e.g. `pixel_values`, `image_grid_thw`), and in the standard generation path manually left-pads `prompt_ids` and converts multimodal outputs to tensors (adding `numpy` as a dependency) before calling `generate`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c098b90827bce88ceaf30681215ad73252283f65. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/trl/trainer/rloo_trainer.py
+++ b/trl/trainer/rloo_trainer.py
@@ -936,33 +936,6 @@ class RLOOTrainer(_BaseTrainer):
                     self.vllm_generation.sync_weights()
                 self._last_loaded_step = self.state.global_step
 
-            # Tokenize prompts and extract images (for VLM) before calling vLLM
-            if is_conversational({"prompt": prompts[0]}):
-                # Extract images from messages for VLM support
-                images = []
-                has_images = False
-                for prompt in prompts:
-                    prompt_images = []
-                    for message in prompt:
-                        if isinstance(message["content"], list):
-                            for part in message["content"]:
-                                if part["type"] == "image":
-                                    prompt_images.append(part["image"])
-                                    has_images = True
-                    images.append(prompt_images if prompt_images else None)
-                images = images if has_images else None
-
-                # RLOO does not support tools; omit tools/chat_template args
-                tokenized = self.processing_class.apply_chat_template(
-                    conversation=prompts,
-                    add_generation_prompt=True,
-                    tokenize=True,
-                    return_dict=True,
-                    **self.chat_template_kwargs,
-                )
-            else:
-                images = None
-
             # Generate using vLLM (note: RLOO doesn't use logprobs from generation, so we ignore them)
             num_generations = self.num_generations if mode == "train" else self.num_generations_eval
             prompt_ids, completion_ids, _, _, _ = self.vllm_generation.generate(


qgallouedec:
@codex review

qgallouedec:
@codex review

trl/trainer/grpo_trainer.py
```diff
@@ -1216,6 +1217,47 @@ def _generate_single_turn(self, prompts: list):
...
+                add_generation_prompt=True,
+                tokenize=True,
+                return_dict=True,
+                padding=True,
+                **self.chat_template_kwargs,
+            )
+            # Unpad input_ids: remove padding tokens using attention_mask to get per-sequence lists
+            prompt_ids = [
+                [tok for tok, m in zip(ids, mask, strict=True) if m]
+                for ids, mask in zip(tokenized["input_ids"], tokenized["attention_mask"], strict=False)
```
albertvillanova: Maybe I'm missing something, but why `strict=False` here? A tokenizer always returns input_ids and attention_mask with the same number of sequences. Using strict=False would silently truncate if they somehow diverged, masking a real bug. Also , it is inconsistent with the inner zip.
```suggestion
                for ids, mask in zip(tokenized["input_ids"], tokenized["attention_mask"], strict=True)
```

trl/trainer/rloo_trainer.py
```diff
@@ -888,6 +889,45 @@ def _generate_single_turn(self, prompts: list):
...
+                add_generation_prompt=True,
+                tokenize=True,
+                return_dict=True,
+                padding=True,
+                **self.chat_template_kwargs,
+            )
+            # Unpad input_ids: remove padding tokens using attention_mask to get per-sequence lists
+            prompt_ids = [
+                [tok for tok, m in zip(ids, mask, strict=True) if m]
+                for ids, mask in zip(tokenized["input_ids"], tokenized["attention_mask"], strict=False)
```
albertvillanova: The same here:

```suggestion
                for ids, mask in zip(tokenized["input_ids"], tokenized["attention_mask"], strict=True)
```

PR #5240, opened 2026-03-07
[GRPO/RLOO] Extract tokenize prompts from `_generate_single_turn`

qgallouedec:
## Context

- #5225
- #5227
- #5232
- #5238
- #5239
- [this one]
- #5242

The previous PR unified tokenization across all generation backends into a single block at the top of `_generate_single_turn`. This PR extracts that tokenization into a dedicated `_tokenize_prompts` method and moves the calls to it into the callers, so that `_generate_single_turn` accepts pre-tokenized inputs.

## Changes

- **New `_tokenize_prompts(prompts)` method** in both GRPO and RLOO: Extracts tokenization logic (image extraction, `apply_chat_template`, multimodal field extraction) into a reusable method. Returns `(prompt_ids, images, multimodal_fields)`.
- **`_generate_single_turn` signature change**: From `(prompts: list)` to `(prompt_ids, images=None, multimodal_fields=None)`. The method is now a pure generation method that accepts pre-tokenized inputs.
- **Updated call sites**: GRPO `_generate`, GRPO `_tool_call_loop`, and RLOO `_generate` now call `_tokenize_prompts` before `_generate_single_turn`.

## Why

This separation is a prerequisite for the token-in/token-out pipeline. In the tool-calling loop, the current code still re-tokenizes the full prompt+completion+tool-result messages on each iteration. With tokenization now happening outside `_generate_single_turn`, future PRs can replace the `_tokenize_prompts` call in the tool loop with direct token ID concatenation — avoiding re-tokenization entirely.

## Backward compatibility

No user-facing API changes. `_generate_single_turn` and `_tokenize_prompts` are internal methods.

---

> [!NOTE]
> **Medium Risk**
> Touches the core generation path and tool-calling loop via a signature change to `_generate_single_turn`, so regressions would show up as incorrect tokenization/generation or broken multimodal/tool flows despite being an internal refactor.
> 
> **Overview**
> Refactors GRPO and RLOO generation to separate prompt preprocessing from decoding/generation by introducing `_tokenize_prompts(prompts)` and changing `_generate_single_turn` to accept pre-tokenized `prompt_ids` plus extracted `images`/`multimodal_fields`.
> 
> Call sites (GRPO `_generate` and tool-calling loop, plus RLOO `_generate`) now tokenize once and pass token IDs through subsequent generation steps, reducing chances of decode/re-tokenize mismatches and preparing for a token-in/token-out tool-calling pipeline.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6c8f55c4ab4ba8fc4614b478acf5c31f92e7df55. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -1553,9 +1553,7 @@ class GRPOTrainer(_BaseTrainer):
             prompt_ids, completion_ids, logprobs = output["prompt_ids"], output["completion_ids"], output["logprobs"]
         else:
             prompt_ids, images, multimodal_fields = self._tokenize_prompts(prompts)
-            prompt_ids, completion_ids, logprobs = self._generate_single_turn(
-                prompt_ids, images, multimodal_fields
-            )
+            prompt_ids, completion_ids, logprobs = self._generate_single_turn(prompt_ids, images, multimodal_fields)
             extra_fields = {}
 
         # Decode completions. It's important to use `parse_response` when possible, because it handles tool calls.


trl/trainer/grpo_trainer.py
```diff
@@ -1257,6 +1254,13 @@ def _generate_single_turn(self, prompts: list):
             prompt_ids = self.processing_class(text=prompts)["input_ids"]
             images = None
             multimodal_fields = {}
+        return prompt_ids, images, multimodal_fields
+
+    def _generate_single_turn(self, prompt_ids, images=None, multimodal_fields=None):
+        device = self.accelerator.device
+        mode = "train" if self.model.training else "eval"
+        if multimodal_fields is None:
+            multimodal_fields = {}
```
albertvillanova: I think this is defensive code. We can remove the defaults because they serve no current purpose: all current callers always pass all three arguments.

```suggestion
    def _generate_single_turn(self, prompt_ids, images, multimodal_fields):
        device = self.accelerator.device
        mode = "train" if self.model.training else "eval"
```

trl/trainer/rloo_trainer.py
```diff
@@ -927,6 +924,13 @@ def _generate_single_turn(self, prompts: list):
             prompt_ids = self.processing_class(text=prompts)["input_ids"]
             images = None
             multimodal_fields = {}
+        return prompt_ids, images, multimodal_fields
+
+    def _generate_single_turn(self, prompt_ids, images=None, multimodal_fields=None):
+        device = self.accelerator.device
+        mode = "train" if self.model.training else "eval"
+        if multimodal_fields is None:
+            multimodal_fields = {}
```
albertvillanova: The same:

```suggestion
    def _generate_single_turn(self, prompt_ids, images, multimodal_fields):
        device = self.accelerator.device
        mode = "train" if self.model.training else "eval"
```

trl/trainer/grpo_trainer.py
qgallouedec: agree I like it

PR #5249, opened 2026-03-09
batch params together in weight sync and async update the weights

winglian:
# What does this PR do?

Rather than firing off 100s of HTTP calls to update weights to vLLM, we can update these in batch. Also, we can async the rpc collective call so the update doesn't block the HTTP call. 

For Owen2 0.5B, this cuts the weight sync from 0.6sec to 0.15sec.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

---

> [!NOTE]
> **Medium Risk**
> Touches distributed weight-sync plumbing (HTTP + NCCL/XCCL ordering, new batching/chunking, and async worker dispatch), so subtle ordering or shape/dtype mismatches could cause sync hangs or incorrect weights despite the change being localized.
> 
> **Overview**
> Speeds up vLLM *server-mode* weight synchronization by introducing a batched update path: the client now can POST parameter metadata once (optionally chunked by total tensor elements) and then broadcast tensors sequentially, and the server loads the received weights in one `load_weights()` call.
> 
> Updates `VLLMGeneration.sync_weights()` to use this batched sync when not using ZeRO-3/FSDP gather paths, and threads a new config/arg (`weight_sync_chunk_size` / `vllm_weight_sync_chunk_size`) from `GRPOConfig` through `GRPOTrainer` into `VLLMGeneration`.
> 
> On the `trl vllm-serve` side, adds a `/batch_update_named_params/` endpoint and switches the existing `init_communicator`/`update_named_param` fan-out to workers to send over pipes concurrently via `asyncio` to avoid blocking the HTTP handler.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b777252b282562ba4da60f8857ceb380489c9a80. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/trl/generation/vllm_client.py
+++ b/trl/generation/vllm_client.py
@@ -511,6 +511,60 @@ class VLLMClient:
             self.communicator.broadcast(weights, src=self.rank)
             self.communicator.group.barrier()
 
+    def batch_update_named_params(self, params: list[tuple[str, torch.Tensor]], chunk_size: int | None = None):
+        """
+        Updates multiple named parameters in a single batch, reducing HTTP round-trips.
+
+        Sends parameter metadata via HTTP POST, then broadcasts each tensor via NCCL in sequence.
+        When chunk_size is set, splits params into chunks whose total element count doesn't exceed
+        the limit, avoiding large HTTP requests regardless of individual parameter sizes.
+
+        Args:
+            params: List of (name, weights_tensor) tuples.
+            chunk_size: Max total elements per HTTP call. None = all in one call. Use smaller
+                        values for large models to avoid HTTP request size limits.
+        """
+        if chunk_size is None:
+            chunks = [params]
+        else:
+            chunks = []
+            current_chunk = []
+            current_elements = 0
+            for name, weights in params:
+                n = weights.numel()
+                if current_chunk and current_elements + n > chunk_size:
+                    chunks.append(current_chunk)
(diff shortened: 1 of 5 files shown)


kashif (approved):

qgallouedec:
thanks for the PR. I'd prefer having a good heuristic (hard-coded value) and avoid adding again a new parameter. What would be a good value?

qgallouedec:
Thanks @winglian, and apologies for the long silence, this slipped through our backlog.

The optimization itself is very welcome: batching the param updates and async-ing the collective RPC is a real win for weight-sync latency. The problem is that `sync_weights` has been substantially refactored since March: weight pushes now go through a single `_push_param_to_vllm` / `self._dist` path, and the FSDP1/FSDP2 branches were unified into `_sync_fsdp_params_to_vllm`, so the branches this PR patches no longer exist and it now conflicts. That makes this a reimplementation rather than a rebase, and the natural home for the batching is now `_push_param_to_vllm` (so RLOO, which shares `VLLMGeneration`, also picks it up along with the `vllm_weight_sync_chunk_size` knob).

I'm going to close this to keep the queue tidy, but we'd gladly review a fresh PR that adds batched sync against the new structure if you're still up for it, thanks again!

PR #5255, opened 2026-03-10
Change default `vllm_mode` to `"colocate"` and add v0→v1 migration guide

qgallouedec:
## Summary

- **Default `vllm_mode` changed from `"server"` to `"colocate"`** across all configs (`GRPOConfig`, `RLOOConfig`, `GOLDConfig`, `OnlineDPOConfig`). Colocate is a better default — it works out of the box without spinning up a separate vLLM server, which is the most common friction point for new users. Users who need server mode can still set `vllm_mode="server"` explicitly.
- **Updated docs and tests** that implicitly assumed the old `"server"` default to explicitly set `vllm_mode="server"` where needed (`vllm_integration.md`, `grpo_trainer.md`, `rloo_trainer.md`, `speeding_up_training.md`, `test_online_dpo_trainer.py`).
- **Added a minimal `MIGRATION.md`** at the repo root for v0 → v1. It should stay small — we have already done most of the breaking work in v0.29 (trainers moved to experimental, model classes removed, etc.), so the guide just highlights what is genuinely new and keeps the v0.29 changes in collapsible sections for anyone migrating from an older version.

---

> [!NOTE]
> **Medium Risk**
> Changes the default vLLM execution mode from external server to in-process across multiple configs, which can alter runtime behavior and GPU memory requirements for existing training setups.
> 
> **Overview**
> **Switches the default vLLM integration mode to `"colocate"`** (from `"server"`) across `GRPOConfig`, `RLOOConfig`, `OnlineDPOConfig`, `GOLDConfig`, and the `VLLMGeneration` backend.
> 
> **Docs and examples are updated** to reflect the new default and to explicitly set `vllm_mode="server"` where server-based workflows are described (`grpo_trainer.md`, `rloo_trainer.md`, `speeding_up_training.md`, `vllm_integration.md`). **Tests are adjusted** to pin server-mode behavior and to assert the new default (`test_online_dpo_trainer.py`).
> 
> Adds a new top-level `MIGRATION.md` highlighting v0→v1 breaking changes, including the `vllm_mode` default change and an `SFTConfig.packing` rename.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d8a0872b49948c03a8156ff6aee2d790d94355d0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -8,3 +8,13 @@ This guide covers the breaking changes introduced in TRL v1 and how to update yo
 | --- | --- | --- | --- | --- |
 | `GRPOConfig` | `vllm_mode` | `"server"` | `"colocate"` | If you use `use_vllm=True` without specifying `vllm_mode`, vLLM will now run in the same process instead of connecting to a separate server. Set `vllm_mode="server"` explicitly if you rely on server mode. |
 | `RLOOConfig` | `vllm_mode` | `"server"` | `"colocate"` | Same as above. |
+
+## Renamed options
+
+| Config | Parameter | v0 value | v1 value | Action needed |
+| --- | --- | --- | --- | --- |
+| `SFTConfig` | `packing` | `"bfd-requeue"` | `"bfd_split"` | Replace `packing="bfd-requeue"` with `packing="bfd_split"`. The old value will still be accepted for a few versions but will be removed in a future release. |
+
+## Migrating from an earlier version
+
+Depending on which version you're migrating from, refer to the [release notes](https://github.com/huggingface/trl/releases) for v0.29 and earlier for version-specific changes.


trl/generation/vllm_generation.py
```diff
@@ -120,13 +120,14 @@ class VLLMGeneration:
 
         > Parameters for vLLM:
 
-        mode (`str`, *optional*, defaults to `"server"`): vLLM mode. Must be one of `"server"` or
-            `"colocate"`.
+        mode (`str`, *optional*, defaults to `"colocate"`):
+            vLLM mode. Must be one of `"colocate"` or `"server"`.
```
albertvillanova, line 122: Shouldn't we follow a deprecation cycle instead? This would give users time to explicitly set the mode until v1 is released.

MIGRATION.md
albertvillanova, line 1: Not sure of fully understanding this: are we creating a migration guide for changes that were already released in the previous minor version?

trl/generation/vllm_generation.py
qgallouedec: I considered a deprecation cycle, but since v1 is right ahead, I’d recommend we skip it. Best practice is usually a deprecation cycle when you’ve got more pre-releases coming, but here, v0.29 was the last one, and v1 should be our stable baseline.
Let’s just set the final defaults at v1, communicate it clearly with a migration guide.

MIGRATION.md
qgallouedec: Yeah, I was a bit hesitant about including those pre-0.29 changes. I realize now it might be confusing, since they’re already effective. I see how that could mislead, so I’ll take them out and keep the guide focused on what’s truly new.

albertvillanova: Thanks for your clarifying explanation: that makes sense to me.

That said, I also understand the original motivation. I think having a place summarizing the recent breaking changes can be useful for users upgrading after skipping a few versions.

What about this compromise?
- Keep MIGRATION.md focused only on the v0 to v1 migration
- Document the v0.29 breaking changes elsewhere
  - For now, we document these in the release notes: is this sufficient? 
  - What about introducing a CHANGELOG.md to track changes across versions?
  - Even though for v1.x series we will try to avoid breaking changes, maybe we could have a clear place to document them in case any become unavoidable
- We could optionally add a short note in the migration guide pointing users to the v0.29 release notes in case they are upgrading from an earlier version

AmineDiro: +1 on the CHANGELOG.md  idea. 

I usually think about migration from a major to another if they were breaking changes. So I think this file is more of "nice to have" for users and mandatory for now ?

qgallouedec: Yes I like the CHANGELOG.md as well, and it will make the writing of the release notes easier. Can you open a dedicated PR?

For this one, I'll just add

```
Depending on which version you're migrating from, refer to the [release notes](https://github.com/huggingface/trl/releases) for v0.29 and earlier for version-specific changes.
```

PR #5286, opened 2026-03-13
Prevent corruption of DPO VLM training if "keep_end" truncation_mode

### LINKED ISSUE #5285: DPOTrainer silently corrupts VLM training with "keep_end" truncation_mode
DPOTrainer silently corrupts VLM training when truncation_mode="keep_end" is set with a non-None max_length.

### Description

When `max_length` is set and `truncation_mode="keep_end"` (the non-default mode) is used with a vision-language model, DPOTrainer silently drops all image tokens from the sequence without any warning or error.

### Root cause

`keep_end` works by flushing padding to the right, taking the last `max_length` tokens, then flushing padding back to the left. In DPO, the sequence layout is [prompt (with image tokens) | completion]. The prompt (and therefore all image tokens) lives at the beginning of the sequence. `keep_end` removes everything from the beginning, so the model is trained without any image information while receiving `pixel_values` for a visual forward pass that is now entirely disconnected from the token sequence.

No error is raised. Training runs to completion and loss decreases, making this failure invisible.

albertvillanova:
Prvent corruption of DPO VLM training if "keep_end" truncation_mode:
- Raise ValueError when truncation_mode="keep_end" is used for VLM training in DPO.

Fix #5285.

This PR addresses a regression related to vision-language models (VLMs) and sequence truncation. It ensures that using the `keep_end` truncation mode with VLMs raises a clear error at initialization, preventing silent corruption of training data. The update includes both a code fix and a regression test.

### Changes

Validation improvements for vision-language models:

* Added a check in the `DPOTrainer.__init__` method to raise a `ValueError` if a vision-language dataset is used with `truncation_mode='keep_end'`, explaining that image tokens would be dropped and recommending alternatives.

Testing enhancements:

* Introduced a regression test (`test_train_vlm_keep_end_raises`) to verify that initializing training with `truncation_mode='keep_end'` for a vision-language model raises the expected error, preventing silent data corruption.

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a defensive init-time validation and a regression test; only affects the VLM + `max_length` + `truncation_mode='keep_end'` configuration by failing fast instead of proceeding.
> 
> **Overview**
> Prevents silent corruption in DPO vision-language training by **failing fast** when a vision dataset is used with `max_length` set and `truncation_mode='keep_end'`, raising a clear `ValueError` during `DPOTrainer` initialization.
> 
> Adds a regression test to ensure VLM trainer construction with `keep_end` truncation reliably errors (fix for #5285).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f36b3c3b0e53ca96e5d67f7e15ad3012bbf6f72a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/trl/experimental/bco/bco_config.py
+++ b/trl/experimental/bco/bco_config.py
@@ -15,8 +15,6 @@
 from dataclasses import dataclass, field
 from typing import Any
 
-from transformers import TrainingArguments
-
 from ...trainer.base_config import _BaseConfig
 
 
@@ -78,7 +76,7 @@ class BCOConfig(_BaseConfig):
     > - `learning_rate`: Defaults to `5e-7` instead of `5e-5`.
     """
 
-    _VALID_DICT_FIELDS = TrainingArguments._VALID_DICT_FIELDS + ["model_init_kwargs"]
+    _VALID_DICT_FIELDS = _BaseConfig._VALID_DICT_FIELDS + ["model_init_kwargs"]
 
     # Parameters whose default values are overridden from TrainingArguments
     learning_rate: float = field(
--- a/trl/experimental/cpo/cpo_config.py
+++ b/trl/experimental/cpo/cpo_config.py
@@ -15,8 +15,6 @@
 from dataclasses import dataclass, field
 from typing import Any
 
-from transformers import TrainingArguments
-
 from ...trainer.base_config import _BaseConfig
 
 
@@ -91,7 +89,7 @@ class CPOConfig(_BaseConfig):
     > - `learning_rate`: Defaults to `1e-6` instead of `5e-5`.
     """
 
-    _VALID_DICT_FIELDS = TrainingArguments._VALID_DICT_FIELDS + ["model_init_kwargs"]
+    _VALID_DICT_FIELDS = _BaseConfig._VALID_DICT_FIELDS + ["model_init_kwargs"]
 
     # Parameters whose default values are overridden from TrainingArguments
     learning_rate: float = field(
--- a/trl/experimental/kto/kto_config.py
+++ b/trl/experimental/kto/kto_config.py
@@ -15,8 +15,6 @@
 from dataclasses import dataclass, field
 from typing import Any
 
-from transformers import TrainingArguments
-
 from ...trainer.base_config import _BaseConfig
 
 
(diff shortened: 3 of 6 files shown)


qgallouedec:
I'm not sure to understand this one. The image tokens are indeed in the beginning, but they are usually not the first tokens. Eg if the data looks like

```
<user><img>What is it?<assistant>A flower<end>
<user><img>What is it?<assistant>A bus<end>
```

What prevents from truncating the first token (`<user>`)?

albertvillanova:
@qgallouedec, even in your very edge case, how could you be sure that you are just removing the first token (and not any image token) with "keep_end" in every example when they have different lengths? `max_length` is a single scalar applied uniformly to the whole dataset.

What this PR is trying to solve is slightly narrower: with `keep_end`, truncation removes the varying-length prefix of the sequence, so in a vision example it can drop the whole prompt prefix including the image placeholder/tokens. In your example, that could indeed mean dropping <user> first, then , then "What is it?", depending on how much needs to be truncated.

The reason I called out image tokens explicitly is that, for VLM inputs, losing them is especially problematic: once the visual tokens are truncated away, the example is no longer a valid multimodal sample and can become semantically inconsistent with the remaining text. By contrast, truncating text-only prefixes is still undesirable, but it is the usual trade-off of sequence truncation and not something specific to vision inputs.

I could improve the wording of the error to make it more precise: the core argument is not that image tokens are necessarily the **very first** tokens, but that they live in the **prefix region** that `keep_end` is designed to discard.
```diff
- "truncation_mode='keep_end' is not supported for vision-language models. Image tokens reside in "
- "the prompt at the beginning of the sequence; keeping the end would drop them. Use "
- "truncation_mode='keep_start' (the default) or set max_length=None."
+ "truncation_mode='keep_end' is not supported for vision-language models. Image tokens reside "
+ "inside the prompt portion of the sequence; depending on the example, keep_end may silently "
+ "drop them, causing pixel_values to be forwarded to the model with no corresponding visual "
+ "tokens in input_ids. Use truncation_mode='keep_start' (the default) or set max_length=None."
```

`keep_start` does not have this problem: as long as `max_length >= prompt_len`, image tokens are always safe.

qgallouedec:
I see. My assumption was that if you modify truncation, you're expected to understand the implications. In other words, if your truncation settings end up affecting images, that’s your responsibility.

In practice, we strongly recommend disabling truncation for VLMs. If you still choose to enable it, it’s not surprising if things break.

In any case, the `truncation_side=“keep_end”` option is generally undesirable; therefore, disallowing it for VLMs—where it is even more undesirable—is also a reasonable solution.

Should we align SFT as well?

albertvillanova:
Thanks for the clarification, @qgallouedec.

My motivation for raising a ValueError here is less about shielding users from all misconfigurations, and more about catching a systematically unsafe configuration. In this case, `keep_end` is not just "suboptimal", but it almost deterministically breaks the multimodal structure by removing the prefix where image tokens live. So it felt closer to an invalid setting than a risky one.

Regarding your last point, yes, I think it would be good to align SFT as well for consistency. If this approach is approved, I will implement the same policy across trainers.

PR #5235, opened 2026-03-06
Update openenv examples to use `environment_factory`

sergiopaniego:
# What does this PR do?

TODO:

- [x] Migrate notebooks
- [x] Update TRL-OpenEnv guide
- [x] Add multi-env example

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes, but they significantly rewrite OpenEnv training guidance and example code; risk is mainly confusing users if the new `environment_factory` patterns or links are incorrect.
> 
> **Overview**
> **Migrates OpenEnv examples to `environment_factory`.** The OpenEnv integration guide is heavily rewritten to center on `GRPOTrainer(environment_factory=...)`, including updated install instructions, environment-class requirements, reward functions consuming the new `environments` argument, and guidance on multi-turn limits and server concurrency.
> 
> **Reorganizes OpenEnv docs/examples.** Adds `OpenEnv` to the main docs toctree under Integrations, restructures `example_overview.md` and `examples/notebooks/README.md` to introduce dedicated *OpenEnv Notebooks*/*OpenEnv Scripts* sections, and updates the Sudoku GRPO notebook to remove the custom rollout pipeline in favor of an `environment_factory`-based `SudokuEnv` plus environment-backed reward functions (and includes a multi-environment training example in the OpenEnv guide).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fee1876f2a350447973c43c5906a8637fd81869b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #5299, opened 2026-03-18
Rollout generation

AmineDiro:
# What does this PR do?

#### Staleness Control (`max_staleness`)

- Added a `max_staleness` to `AsyncRolloutWorker`. **Groups generated** from a model version more than `max_staleness` steps behind the current version are automatically cancelled and dropped.
- A min-heap (`version_heap`) tracks the model version of each in-flight group, enabling O(log n) detection of stale groups.
- When a new model version is pushed, in-flight tasks for stale groups are cancelled via `asyncio.CancelledError` and their slots are immediately reclaimed.
- `update_model_version` now signals an `asyncio.Event` (`_version_updated`) so the generation loop can react immediately without polling.

#### Sequence Length Safety (`max_seq_length`, `max_model_len`)

- Added `max_seq_length` config option: the maximum allowed length for training. Samples exceeding this are dropped before being enqueued.
- At worker startup, `max_model_len` is fetched from the vLLM `/v1/models` endpoint so the worker always knows the server's context window.
- Per-turn `effective_max_tokens` is dynamically clipped to `min(max_completion_tokens, max_seq_length - prompt_len, max_model_len - prompt_len)`, preventing 400 errors from vLLM when prompts are long.
- Prompts that exceed `max_model_len` outright are skipped immediately with a logged warning.
- Both pre-turn and mid-multi-turn length checks are applied, with early stopping when the growing prompt would overflow.
- Added `truncated` and `num_turns` tracking to `RolloutGroup` and `RolloutCompletion`, and exposed them as `rollout/truncated` and `rollout/num_turns` metrics per sample.

#### Pause/Resume Correctness

- `_resume_event` (an `asyncio.Event`) gates the dispatch inner loop: when `pause()` is called, the event is cleared and the generation loop stops dispatching new requests before vLLM is paused — avoiding a race where requests land between the pause signal and the server actually pausing.
- `resume()` sets the event after the HTTP resume call, re-enabling dispatch.

#### Error Handling and Retry Logic

- `_generate_one_turn` now has a bounded retry loop (`max_generation_retry=10`) instead of an infinite loop.
- Retry logic now distinguishes between transient server errors (5xx — retried) and client errors (4xx — re-raised immediately), avoiding wasteful retries on bad requests.
- `_generate_one` now wraps `task.result()` in a try/except: per-sample exceptions are logged and counted as failures rather than crashing the entire generation loop. Groups where all samples fail are dropped with a warning.
- Cancelled tasks (stale cancellation) are handled gracefully via `asyncio.CancelledError`.

#### Initial Weight Sync via Callback

- Removed the inline `_sync_weight()` call from `_inner_training_loop`. Instead, a new `_InitialWeightSyncCallback` fires on `on_train_begin` — after `accelerator.prepare()` has placed parameters on GPU but before the first training step. This ensures the rollout worker always uses the correct (possibly resumed) policy weights.
- The rollout worker is started in `_inner_training_loop` after the sync completes.

#### Return Type Refactor

- `_generate_one` now returns a `RolloutCompletion` dataclass instead of a bare tuple, improving readability and making it easier to extend with new fields.

#### MBPP Example

- Added `examples/scripts/async_grpo_mbpp.py`: an end-to-end example using `AsyncGRPOTrainer` on the MBPP coding benchmark with a tool-use environment that runs Python code and checks test cases.
- Updated `async_grpo.py` launch command to show multi-GPU (FSDP2 + data-parallel vLLM) configuration.

## Who can review?

---

> [!NOTE]
> **Medium Risk**
> Touches core async rollout generation/scoring and weight-sync sequencing, which can affect training stability and throughput if cancellation/length-clipping logic is wrong. No auth/security changes, but concurrency and integration with vLLM make regressions moderately likely.
> 
> **Overview**
> Improves `AsyncGRPO` rollout generation robustness by adding **staleness control** (`max_staleness`) that cancels/drops in-flight groups from old model versions and skips overly stale groups before scoring.
> 
> Adds **sequence-length safety** via new `AsyncGRPOConfig.max_seq_length`, fetching `max_model_len` from vLLM, and dynamically clipping per-request `max_tokens`; overlong prompts/samples are skipped or truncated to avoid vLLM 400s and training-side overflow.
> 
> Refactors rollout results to a structured `RolloutCompletion`/turn record with per-tool timing/failure metrics, improves pause/resume coordination during weight sync, bounds generation retries with better 4xx/5xx handling, and moves initial weight sync to an `on_train_begin` callback. Also adds an `async_grpo_mbpp.py` tool-use example and updates the existing `async_grpo.py` launch instructions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 674073a4d810b0ba5657b2eadf618b1bdf1ee7fc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/trl/experimental/async_grpo/async_rollout_worker.py
+++ b/trl/experimental/async_grpo/async_rollout_worker.py
@@ -183,7 +183,7 @@ class AsyncRolloutWorker:
         max_inflight_tasks: int = 128,
         queue_maxsize: int = 0,
         vllm_server_url: str = "http://localhost:8000",
-        max_completion_tokens: int = 1024,
+        max_completion_length: int = 1024,
         max_seq_length: int | None = None,
         max_staleness: int = 4,
         temperature: float = 1.0,
@@ -254,7 +254,7 @@ class AsyncRolloutWorker:
 
         self.vllm_server_url = vllm_server_url.rstrip("/")
         self.model_update_group = None
-        self.max_completion_tokens = max_completion_tokens
+        self.max_completion_length = max_completion_length
         self.max_seq_length = max_seq_length
         self.max_staleness = max_staleness
         self.temperature = temperature
@@ -736,8 +736,8 @@ class AsyncRolloutWorker:
         # Dynamically clip max_tokens so that prompt + completion fits within both max_seq_length and
         # max_model_len (the vLLM server's context window). Without this, vLLM rejects the request with a 400
         # when len(prompt_ids) + max_tokens > max_model_len.
-        effective_max_tokens = self.max_completion_tokens
-        if self.max_seq_length:
+        effective_max_tokens = self.max_completion_length
+        if self.max_seq_length is not None:
             effective_max_tokens = min(effective_max_tokens, self.max_seq_length - len(prompt_ids))
         if self.max_model_len is not None:
             effective_max_tokens = min(effective_max_tokens, self.max_model_len - len(prompt_ids))
@@ -805,7 +805,7 @@ class AsyncRolloutWorker:
                 )
                 return _build_completion(turns, truncated=True, total_duration=time.monotonic() - t_start)
 
-            effective_max_tokens = self.max_completion_tokens
+            effective_max_tokens = self.max_completion_length
             if self.max_seq_length:
                 effective_max_tokens = min(effective_max_tokens, self.max_seq_length - len(prompt_ids))
             if self.max_model_len is not None:


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -62,6 +62,23 @@ def send_weights(self, iterator: Iterator[tuple[str, torch.Tensor]]) -> None: ..
...
+    """
+
+    def __init__(self, trainer: "AsyncGRPOTrainer"):
+        self._trainer = trainer
+
+    def on_train_begin(self, _args, _state, _control, **_kwargs):
+        self._trainer._sync_weight()
+        self._trainer.remove_callback(type(self))
+
+
```
qgallouedec, line 81: What the motivation to use a callback instead of simply calling the weight sync in the trainer?

examples/scripts/async_grpo_mbpp.py
```diff
@@ -0,0 +1,173 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+CUDA_VISIBLE_DEVICES=2,3,4,5 VLLM_SERVER_DEV_MODE=1 vllm serve Qwen/Qwen3-4B \
+    --data-parallel-size 4 \
+    --weight-transfer-config '{"backend":"nccl"}' \
+    --max-num-seqs 64 \
```
qgallouedec:
```suggestion
```

I checked in the vllm doc, there is no real added value to set this

examples/scripts/async_grpo_mbpp.py
```diff
@@ -0,0 +1,173 @@
...
+    --max-model-len 9216
+
+CUDA_VISIBLE_DEVICES=0,1 accelerate launch  --config_file examples/accelerate_configs/fsdp2.yaml  examples/scripts/async_grpo_mbpp.py
```
qgallouedec:
```suggestion
CUDA_VISIBLE_DEVICES=0,1 accelerate launch --config_file examples/accelerate_configs/fsdp2.yaml  examples/scripts/async_grpo_mbpp.py
```

examples/scripts/async_grpo.py
```diff
@@ -13,11 +13,14 @@
...
 
 """
-CUDA_VISIBLE_DEVICES=1 VLLM_SERVER_DEV_MODE=1 vllm serve Qwen/Qwen3-4B \
+CUDA_VISIBLE_DEVICES=2,3,4,5 VLLM_SERVER_DEV_MODE=1 vllm serve Qwen/Qwen3-4B \
+    --data-parallel-size 4 \
     --weight-transfer-config '{"backend":"nccl"}' \
     --max-model-len 9216
 
-LOG_LEVEL=DEBUG CUDA_VISIBLE_DEVICES=0 accelerate launch examples/scripts/async_grpo.py
+CUDA_VISIBLE_DEVICES=0,1 accelerate launch  --config_file examples/accelerate_configs/fsdp2.yaml  examples/scripts/async_grpo.py
```
qgallouedec:
```suggestion
CUDA_VISIBLE_DEVICES=0,1 accelerate launch --config_file examples/accelerate_configs/fsdp2.yaml  examples/scripts/async_grpo.py
```

examples/scripts/async_grpo_mbpp.py
```diff
@@ -0,0 +1,173 @@
...
+class MBPPEnvironment:
+    """
+    A synchronous environment class designed for `AsyncGRPOTrainer`.
+    Each environment instance handles tracking test cases and exposes the `execute_python_code` tool.
+    """
+
+    def __init__(self):
+        self.test_list = []
+        self.done = False
+
```
qgallouedec, line 52: IMO we can remove this
```suggestion
```

qgallouedec: it's safer because it will fail if we try to step before the reset

examples/scripts/async_grpo_mbpp.py
```diff
@@ -0,0 +1,173 @@
...
+                feedback = result.stderr[-2000:]
+                return f"Execution failed with error:\n{feedback}\nPlease fix the code and try again."
+        except subprocess.TimeoutExpired:
+            return "Execution timeout."
+        finally:
+            if os.path.exists(temp_path):
+                os.remove(temp_path)
+
+    def is_done(self) -> bool:
+        return self.done
```
qgallouedec, line 97: because it's missing the docstring, it won't be exposed as a tool to the model. If it's not meant to be exposed, then I'd recommend `def _is_done`. If it is meant to be exposed as a tool, it requires a docstring.

examples/scripts/async_grpo_mbpp.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def main() -> None:
+    os.environ["WANDB_PROJECT"] = "async_grpo_trl_mbpp"
+    # 1. Load dataset
+    dataset = load_dataset("google-research-datasets/mbpp", split="train+test")
+    dataset = dataset.map(format_sample, remove_columns=dataset.column_names)
+
+    # 2. Config setup
+    config = AsyncGRPOConfig(
```
qgallouedec: nit, for consistency


```suggestion
    training_args = AsyncGRPOConfig(
```

examples/scripts/async_grpo_mbpp.py
```diff
@@ -0,0 +1,173 @@
...
+        logging_steps=1,
+        # trackio
+        project="async_grpo_trl_mbpp",
+        trackio_space_id=None,
+    )
+
+    # 3. Trainer initialization
+    trainer = AsyncGRPOTrainer(
+        model="Qwen/Qwen3-4B",
+        args=config,
```
qgallouedec:
```suggestion
        args=training_args,
```

examples/scripts/async_grpo_mbpp.py
```diff
@@ -0,0 +1,173 @@
...
+        max_steps=100,
+        max_staleness=8,
+        # Logging
+        log_completions=True,
+        num_completions_to_print=2,
+        report_to="wandb",
+        logging_steps=1,
+        # trackio
+        project="async_grpo_trl_mbpp",
+        trackio_space_id=None,
```
qgallouedec:
```suggestion
```

it seems like you use wandb in this example, so you can safely drop this

trl/experimental/async_grpo/async_grpo_config.py
```diff
@@ -34,6 +34,11 @@ class AsyncGRPOConfig(_BaseConfig):
             Number of generations per prompt to sample.
         max_completion_length (`int`, *optional*, defaults to `2048`):
             Maximum number of tokens to generate per completion.
+        max_seq_length (`int`, *optional*):
+            Maximum total sequence length (prompt + completion) for training. When set, generation `max_tokens` is
+            dynamically clipped per-prompt so that `prompt_len + completion_len <= max_seq_length`, and any sample
+            that still exceeds this limit is dropped before training. If `None`, no training-side length limit is
```
qgallouedec, line 40: what if one sequence in the group reaches the limit? The whole group is discarded?

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -368,7 +385,9 @@ def __init__(
                     max_inflight_tasks=self.args.max_inflight_tasks,
                     queue_maxsize=self.args.queue_maxsize,
                     vllm_server_url=self.args.vllm_server_base_url,
-                    max_tokens=self.args.max_completion_length,
+                    max_completion_tokens=self.args.max_completion_length,
```
qgallouedec:
```suggestion
                    max_completion_length=self.args.max_completion_length,
```

if we want to rename , then I'd suggest this instead

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -149,13 +246,20 @@ def __init__(
                 if inspect.iscoroutinefunction(tool):
                     raise ValueError("Asynchronous tools are not supported in AsyncRolloutWorker yet.")
                 self._sync_tool_dicts[i][tool.__name__] = tool
-        self.tools = base_tools + (environment_methods[0] if self.environments is not None else [])
+
+        self.tools = base_tools + (
+            # Pre-convert any bound methods to JSON schema dicts so they pass transformers' `isfunction` check.
+            [get_json_schema(t) for t in environment_methods[0]] if self.environments is not None else []
```
qgallouedec, line 252: you don't need this change, `apply_chat_template` will get the json schema for you

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -419,6 +580,45 @@ async def _generate_loop(self, stop_event: asyncio.Event) -> None:
             except asyncio.QueueFull:
                 pass
 
+    def _cancel_stale_tasks(
```
qgallouedec, line 583: are you sure it's needed? if you set `max_inflight_tasks = max_staleness * per_device_train_batch_size * gradient_accumulation_steps * num_processes` (default), then you shouldn't need this in my understanding

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -127,6 +221,7 @@ def __init__(
         self.num_generations = num_generations
         self.max_inflight_tasks = max_inflight_tasks
         self.environments = None
+        self._is_done_methods = [None] * self.max_inflight_tasks
```
qgallouedec, line 224: I strongly think that we shouldn't add any done signal, see https://github.com/huggingface/trl/pull/5235#discussion_r2931929111

Requiring extra methods unnecessarily narrows the set of valid implementations and reduces substitutability. We should depend on the minimal protocol the class actually needs. There is nothing we *can't* do without `is_done`, so I'd recommend simply not having it. Plus at inference, the model generation won't have access to this flag anyway, so it's useless

examples/scripts/async_grpo_mbpp.py
AmineDiro: it is not a tool function. the `_is_done` means it it a private function but we cant to signal that it needs to be implemented. 
The issue we currently have is that user can't define a stop function to exist the generation loop. For example, when in code environment, we want to exist if the tests success but the current setup keeps rolling until we reach max_turns or token exhaustion

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: The done was added for this case  (cf earlier comment):  

> The issue we currently have is that user can't define a stop function to exit the generation loop. For example, when in code environment, we want to exist if the tests succeed but the current setup keeps rolling until we reach max_turns or token exhaustion

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: I added this because I kept getting error in my `async_grpo_mbpp.py` env. The `isfunction` check fails as it is a method.

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: Yes we still need this to cancel current inflight. `max_inflight_tasks` will limit __fresh__ requests.  If for example there is a generation that takes a long time that gets stale, we need to cancel it. If we don't we'll be wasting vllm's compute resources on a stale requests that will get dropped.

trl/experimental/async_grpo/async_grpo_config.py
AmineDiro: `max_seq_length` is only used in `_generate_one` function to truncate generation or to skip prompt. It is also skipped in scoring loop:
```python
  if self.max_seq_length is not None and seq_len > self.max_seq_length:
      logger.warning(f"Dropping overlong sample (seq_len={seq_len}, max_seq_length={self.max_seq_length})")

```

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: `_sync_weight()` was called at the top of _inner_training_loop, before `super()._inner_training_loop()` ran.  it's super()._inner_training_loop() that performs FSDP wrapping and accelerator.prepare() (which moves the model to GPU). So weights were still on CPU when NCCL tried to broadcast them.

So I had:
```bash
  ┃  [rank0]:   File "/fsx/amine_dirhoussi/trl/.venv/lib/python3.11/site-packages/vllm/distributed/device_communicators/pynccl.py", line
  ┃  363, in broadcast
  ┃  [rank0]:     assert tensor.device == self.device, (
  ┃  [rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ┃  [rank0]: AssertionError: this nccl communicator is created to work on cuda:0, but the input tensor is on cpu
```

examples/scripts/async_grpo_mbpp.py
AmineDiro: I think `test_list` has to be init to [] if I am not mistaken

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: To be precise: `transformers.utils.chat_template_utils.render_jinja_template` uses `inspect.isfunction()` to validate tools, but isfunction() returns False for bound methods. The environment's execute_python_code is collected as a bound method via `inspect.getmembers` so I was getting

```bash
  ┃    File "/fsx/amine_dirhoussi/trl/trl/experimental/async_grpo/async_rollout_worker.py", line 287, in _run_loops
  ┃      await asyncio.gather(
  ┃    File "/fsx/amine_dirhoussi/trl/trl/experimental/async_grpo/async_rollout_worker.py", line 392, in _generate_loop
  ┃      prompt_ids = self.tokenizer.apply_chat_template(
  ┃                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ┃    File "/fsx/amine_dirhoussi/trl/.venv/lib/python3.11/site-packages/transformers/tokenization_utils_base.py", line 1667, in
  ┃  apply_chat_template
  ┃      rendered_chat, generation_indices = render_jinja_template(
  ┃                                          ^^^^^^^^^^^^^^^^^^^^^^
  ┃    File "/fsx/amine_dirhoussi/trl/.venv/lib/python3.11/site-packages/transformers/utils/chat_template_utils.py", line 493, in
  ┃  render_jinja_template
  ┃      raise ValueError(
  ┃  ValueError: Tools should either be a JSON schema, or a callable function with type hints and a docstring suitable for auto-conversion to a schema.
```

Maybe I am missing a quick fix or I defined the Env wrongly?

qgallouedec: you must upgrade transformers to at least 5.2.0

qgallouedec: for ref https://github.com/huggingface/transformers/pull/43968

trl/experimental/async_grpo/async_rollout_worker.py
qgallouedec: ah yes good point

trl/experimental/async_grpo/async_rollout_worker.py
qgallouedec: yes I know, but ideally, the model should learn no to continue calling the tool if it gets an error each time. IMO the right way to handle done would be:

```python
class MyEnv:
    def reset(self, **kwargs):
        self._done = False

    def my_func(self):
        """Some nice documentation"""
        if self._done:
            raise Exception("Session expired, you can't use this tool anymore!")
        if some_condition:
            self._done = True
        return
```

qgallouedec: If the generation stops when the environment is `done`, then the model never learns to know that the [environment is done] implies that [it can't use it anymore].

AmineDiro: Sorry, maybe I didn't explain well the usage of the `is_done` in this portion codebase. The current impl does the following:

```python
async def _generate_one(...):
    while True:
        # .... 
        if tool_calls is None or (max_iterations is not None and iteration_num >= max_iterations):
            return ...
```

Now we can have done() raise an Exception but I am not sure it provides a clean signal to know that this rollout had exception because of some env specific errors vs it genuinely reached end of turn.  If we go with Exception route, I would image that we need to define a library side Exception like `StopGeneration(Exception)` to cleanly catch it and know when env has stopped. But then this is equivalent to defining a done function IMO.

Also having a `done` env defined tool extends tools context for the LLM which can be a good choice in some specific cases but overall maybe unnecesary ??

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: Weird, I am sure I have the correct transformers version in env:
```bash
(trl) amine_dirhoussi@login-node-1 trl ±|rollout-generation ✗|→ uv pip list | rg transformers
transformers                             5.2.0
```
I'll retry this specific parsing function is isolation 👍🏼

trl/experimental/async_grpo/async_rollout_worker.py
qgallouedec: So if we don't use the done flag to stop the generation, what's the point of having it? I can see it's used here, but the reason is not clear:

https://github.com/huggingface/trl/pull/5299/changes#diff-175873f8c2363ec8322c4a0af91437e83da991bb5d388f2d01342b2d31eaf8a3R822

it seems like a proxy the a new `is_truncated` flag, which I don't quite understand either. The truncation is equivalent to "the last token isn't an EOS", why adding this flag here?

But more generally, don't you think it's out of the scope of this PR?

AmineDiro: > So if we don't use the done flag to stop the generation, what's the point of having it? I can see it's used here, but the reason is not clear:

To be precise,  I put the _previous_ generation loop. The changed one does use it to break : [line](https://github.com/huggingface/trl/blob/674073a4d810b0ba5657b2eadf618b1bdf1ee7fc/trl/experimental/async_grpo/async_rollout_worker.py#L822)

```python
 if is_done is not None and is_done():
     return _build_completion(turns, truncated=False, total_duration=time.monotonic() - t_start)
```

> it seems like a proxy the a new is_truncated flag, which I don't quite understand either. The truncation is equivalent to "the last token isn't an EOS", why adding this flag here?

Truncation is separate from done and is mainly for metrics and debugging purposes. It can be computed from "isn't last EOS" but its just a bool and recomputing something that small seems unnecessary ? 
 
 > But more generally, don't you think it's out of the scope of this PR?
 
I think there was some feature creep, but this PR was originally intended to change the `async-grpo` branch directly. All the commits were added to debug the `mbpp` example: metrics, trajectories, the done flag, etc.

I believe the first `async-grpo` PR iteration had metrics, so I viewed this as a logical extension, especially since there wasn’t a strictly defined feature scope for this specific PR

PR #5305, opened 2026-03-18
Fix DPOTrainer collators to truncate sequences before padding

### LINKED ISSUE #5304: DPO sequences longer than max_length cause over-allocation and potential OOM during collation
### Description

When `max_length` is set, `DPOTrainer` truncates sequences inside the forward pass (`_truncate_inputs`), *after* the data collator has already padded the batch. Because `DataCollatorForPreference` and `DataCollatorForVisionPreference` have no knowledge of `max_length`, they pad every batch to the longest raw sequence in the dataset.

If the dataset contains sequences longer than `max_length` (which is the common case for real-world preference datasets) the collator allocates tensors of shape `[2 * batch_size, max_raw_seq_len]`, only for `_truncate_inputs` to immediately discard everything beyond column `max_length`. The wasted factor equals `max_raw_seq_len / max_length`, which can easily be 4 to 8 times.

In practice this causes:
- Significantly higher GPU memory usage than the user expects from setting `max_length`
- OOM crashes that would not occur with the equivalent `SFTTrainer` setup on the same dataset and hardware

### Root cause

`SFTTrainer` stores the full concatenated sequence (`input_ids`) in the dataset and calls `truncate_dataset()` at preprocessing time, so the collator never sees a sequence longer than `max_length`.

`DPOTrainer` stores sequences as three separate columns (`prompt_ids`, `chosen_ids`, `rejected_ids`). Truncation must happen on the *combined* sequence `prompt_ids + chosen_ids`, which only exists after concatenation inside the collator. `truncate_dataset()` cannot be used directly here: it would truncate each column independently, which is wrong (e.g. a 512-token prompt + 800-token completion would not be truncated at all against a `max_length` of 1024).

### Expected behavior

The padded batch tensor should be bounded by `max_length`, matching user expectations and the behaviour of `SFTTrainer`.

albertvillanova:
Fix `DPOTrainer` collators to truncate sequences before padding when `max_length` is set.

This PR adds support for truncating input sequences to a maximum length in both text and vision preference data collators in `DPOTrainer`. This helps prevent oversized tensors when handling very long sequences and allows for configurable truncation behavior. The changes also propagate these new options from the trainer arguments to the collators.

Fix #5304.

### Motivation

Previously, `DataCollatorForPreference` and `DataCollatorForVisionPreference` padded every batch to the longest raw sequence in the dataset. `_truncate_inputs` then sliced the padded tensors down to `max_length` inside the forward pass. This caused the collator to allocate tensors up to N× larger than necessary, where N = max_raw_seq_len / max_length, leading to unexpected OOM errors for users who had set `max_length` precisely to stay within their GPU memory budget.

### Solution

The fix moves truncation to the earliest possible point: right after concatenation in the collator, before any tensor allocation; so padding is always bounded by `max_length`.

`_truncate_inputs` in the forward pass is left unchanged. It remains the correct truncation path for user-supplied custom collators and acts as a no-op (cheap slice on already-bounded tensors) for the built-in collators.

### Changes

**Sequence truncation support:**

* Added `max_length` and `truncation_mode` parameters to `DataCollatorForPreference` and `DataCollatorForVisionPreference`, allowing input sequences to be truncated to a specified maximum length. For text, both `"keep_start"` and `"keep_end"` truncation modes are supported; for vision, only `"keep_start"` is allowed.
* Implemented logic in the collators' `torch_call` methods to perform truncation on input IDs, attention masks, and related fields according to the specified `max_length` and `truncation_mode`.
  - **`DataCollatorForPreference` (text):** After building `prompt_chosen_ids` / `prompt_rejected_ids` as plain Python lists, slice them to `max_length` using `keep_start` (`[:max_length]`) or `keep_end` (`[-max_length:]`) before converting to tensors. The completion masks receive the same slice. `chosen_attention_mask` / `rejected_attention_mask` are derived *after* truncation, so their lengths are always consistent.
  - **`DataCollatorForVisionPreference` (vision):** After `flush_left` assembles the final `[2*batch_size, seq_len]` tensors, slice `input_ids`, `attention_mask`, `completion_mask`, and (where present) `token_type_ids` / `mm_token_type_ids` to `[:, :max_length]`. Only `keep_start` applies here; `keep_end` is already rejected upstream with a `ValueError` before the collator is constructed, because image tokens in the prompt would be silently dropped. `pixel_values` and other spatial image tensors are not touched.

**Trainer integration:**

* Updated the trainer initialization to pass `max_length` and `truncation_mode` arguments from the trainer configuration to the appropriate data collator, ensuring the new functionality is available during training.

---

> [!NOTE]
> **Medium Risk**
> Changes batching/truncation behavior in `DPOTrainer` collators, which can affect training inputs and loss if slicing is incorrect, but scope is limited to data collation and argument plumbing.
> 
> **Overview**
> **Prevents oversized batch tensors in `DPOTrainer` when `max_length` is set** by truncating sequences *inside the collators before padding/tensor allocation*.
> 
> `DataCollatorForPreference` now accepts `max_length` and `truncation_mode` (`keep_start`/`keep_end`) and truncates concatenated `prompt+completion` ids and `completion_mask` prior to building attention masks and padding.
> 
> `DataCollatorForVisionPreference` now accepts `max_length` and slices the post-`flush_left` tensors (`input_ids`, `attention_mask`, `completion_mask`, and optional token-type ids) to `[:, :max_length]`; trainer wiring passes `max_length`/`truncation_mode` into the text collator and `max_length` into the vision collator (with `keep_end` rejected for vision datasets).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 99c194ef282dad2465c3cb3b71568240cf0b13b7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,6 +45,7 @@ Concretely:
 - Do not add layers of indirection (registries, factory patterns, plugin systems). A contributor should be able to read a trainer top to bottom and understand the full flow.
 - Prefer a simple implementation that covers 90% of cases over a complex one that covers 100%. A function that handles the common path in 20 lines is better than a catch-all that handles every edge case in 80.
 - Do not add defensive code, fallback paths, or configuration options "just in case". Only handle cases that actually exist today.
+- Avoid `hasattr` and `getattr`. Their use is almost always a symptom of overly defensive programming or a disguised version check (e.g., "this attribute was added in version X"). Instead, either drop the conditional entirely or express the version check explicitly with a version comparison. There is nearly always a cleaner alternative.
 - When in doubt, prefer less code. Every new function, parameter, or branch is maintenance burden. The best abstraction is often no abstraction.
 
 ## Documentation


qgallouedec:
I think it's very very unlikely we hit OOM at collation time. Even with `seq_len=1M` and `batch_size=8`, that’s only ~8M elements: ~30 MB in int32.

In practice, the memory peak is in the forward/backward pass due to the vocab dimension. For example, with a typical vocab_size of 150k and `seq_len=512`, logits alone are ~1.2 GB in (bf16). To reach 1.2 GB at collation time, you’d need seq_len ≈ 40M tokens, which is unrealistic.

albertvillanova:
@qgallouedec, I agree that the OMM argument was indeed exaggerated. The collation overhead is negligible by comparison with the forward/backward pass.

That said, there are three narrower motivations that still stand:
1. **Consistency with the existing SFT VLM collator**: `DataCollatorForVisionLanguageModeling` in SFT already truncates in the collator. Diverging from that pattern without a technical reason makes the codebase harder to reason about.
   - See: https://github.com/huggingface/trl/blob/435c2aef66d62b2f4a78566c8c0e6ea505f10af8/trl/trainer/sft_trainer.py#L466-L474
2. **Semantic contract of the collator**: after the fix, the tensors emitted by the built-in collators are guaranteed to respect `max_length`. Before the fix, the collator output silently violates `max_length` until the forward pass corrects it, which is confusing to anyone inspecting intermediate batch shapes or reasoning about memory.
3. **Code simplification**: The current `keep_end` path in `_truncate_inputs` operates on an already-padded batch tensor and requires a flush_right → slice → flush_left sequence to correctly extract the trailing tokens from each right-padded sequence. Moving truncation to the collator, where the concatenated sequences are still plain Python lists, reduces `keep_end` to a trivial `ids[-max_length:]` slice, with no batch-level bookkeeping needed.

   - Please note that what I actually had in mind with this PR is that it lays the groundwork for a follow-up where `_truncate_inputs` gets removed entirely. Once both built-in collators truncate internally, the only reason `_truncate_inputs` still exists is as a silent safety net for custom collators: which is arguably worse than no safety net, because it hides the fact that the collator isn't doing its job. The follow-up would make the contract explicit (custom collators must truncate before padding), replace the silent fix-up with a single shape assertion at the system boundary to catch violations loudly and early, and then delete `_truncate_inputs` along with the flush_right/flush_left batch-tensor dance that lives inside it. Net result: less code in the trainer, no hidden behavior, and the `keep_end` complexity vanishes because truncating a plain list before padding is just `ids[-max_length:]`. This PR is a necessary precondition for that: you can't safely remove `_truncate_inputs` until the built-in collators already handle it themselves.

PR #5320, opened 2026-03-20
Cancel stale in-flight rollouts and survive generation failures

AmineDiro:
## What this PR does

This PR adds two robustness improvements to the AsyncGRPO rollout worker. 

### 1. Cancel stale in-flight rollouts

- When the policy version advances, the worker cancels in-flight generations whose group started more than `max_staleness`versions ago.
- This frees the generation slots right away instead of spending vLLM compute on rollouts that would be dropped for staleness later.
- It complements the existing consumer-side drop in `RolloutQueueDataset`:
  - `RolloutQueueDataset` drops already-produced samples that are stale by the time the trainer reads them.
  - This change stops producing those samples in the first place.
- Groups are tracked in a min-heap keyed by the version they started at, so the oldest are found cheaply without scanning every pending group.
- A group only becomes eligible for cancellation once all of its generations have been dispatched. This avoids a case where cancelling a partially dispatched group lets `_repeat_iterator` recreate it with fewer rows, stranding a group that can never reach `num_generations`.

### 2. Keep the worker alive on a generation failure

- A single failed generation no longer re-raises and kills the child process.
- The failure is recorded against its group and the successful siblings are kept.
- When the group finishes:
  - If at least 2 generations succeeded, the group is scored with the survivors.
  - If fewer than 2 succeeded, the group is dropped, because GRPO needs at least 2 completions to form a group relative baseline.
- `reward_kwargs` are trimmed to the number of surviving completions so the reward inputs stay aligned.

---

> [!NOTE]
> **Medium Risk**
> Changes rollout scheduling and which samples reach training (partial groups, cancelled stale work), affecting GRPO data quality and vLLM utilization but not auth or persistence.
> 
> **Overview**
> The async rollout worker now **proactively cancels** in-flight vLLM generations when the policy version advances, using the trainer’s `max_staleness` threshold. A FIFO `version_queue` tracks fully dispatched groups so stale work is dropped and slots freed without scanning every pending group; groups only become cancellable after all `num_generations` tasks are dispatched.
> 
> **Generation failures** no longer tear down the child process: failed completions are counted per group, successful siblings are kept, and groups with **at least two** survivors are still scored (GRPO needs a group baseline). Fewer than two successes drops the group; `_score_group` trims `reward_kwargs` to match surviving completions.
> 
> `AsyncGRPOTrainer` passes `max_staleness` into `AsyncRolloutWorker`, complementing the existing consumer-side stale drop in `RolloutQueueDataset`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce130e76b8f6cc392c671acd4621e96a5bd735a4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/async_grpo/async_grpo_trainer.py
+++ b/trl/experimental/async_grpo/async_grpo_trainer.py
@@ -639,6 +639,7 @@ class AsyncGRPOTrainer(_BaseTrainer):
                     queue_maxsize=self.args.queue_maxsize,
                     vllm_server_url=self.args.vllm_server_base_url,
                     max_tokens=self.args.max_completion_length,
+                    max_staleness=self.args.max_staleness,
                     temperature=self.args.temperature,
                     request_timeout=self.args.request_timeout,
                     chat_template_kwargs=self.args.chat_template_kwargs,
--- a/trl/experimental/async_grpo/async_rollout_worker.py
+++ b/trl/experimental/async_grpo/async_rollout_worker.py
@@ -22,6 +22,7 @@ import threading
 import time
 import traceback
 import warnings
+from collections import deque
 from collections.abc import Awaitable, Callable, Iterator
 from dataclasses import dataclass
 from multiprocessing.queues import Queue as MPQueue
@@ -195,6 +196,7 @@ class _AsyncRolloutLoop:
         queue_maxsize: int = 0,
         vllm_server_url: str = "http://localhost:8000",
         max_tokens: int = 32,
+        max_staleness: int = 4,
         temperature: float = 1.0,
         request_timeout: int = 120,
         chat_template_kwargs: dict[str, Any] | None = None,
@@ -231,6 +233,7 @@ class _AsyncRolloutLoop:
         self.max_inflight_tasks = max_inflight_tasks
         self.queue_maxsize = queue_maxsize
         self.max_tokens = max_tokens
+        self.max_staleness = max_staleness
         self.temperature = temperature
         self.request_timeout = request_timeout
         self.chat_template_kwargs = chat_template_kwargs or {}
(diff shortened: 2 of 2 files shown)


trl/experimental/async_grpo/async_grpo_config.py
```diff
@@ -34,6 +34,11 @@ class AsyncGRPOConfig(_BaseConfig):
             Number of generations per prompt to sample.
         max_completion_length (`int`, *optional*, defaults to `2048`):
             Maximum number of tokens to generate per completion.
+        max_seq_length (`int`, *optional*):
+            Maximum total sequence length (prompt + completion) for training. When set, generation `max_tokens` is
+            dynamically clipped per-prompt so that `prompt_len + completion_len <= max_seq_length`, and any sample
+            that still exceeds this limit is dropped before training. If `None`, no training-side length limit is
+            enforced.
```
qgallouedec: In my opinion, we shouldn't need a new arg for this, we could simply get the max length from the server, eg:

```python
client = OpenAI(base_url=base_url, api_key="EMPTY")
models = client.models.list()
for m in models.data:
    if m.id == model_id:
        max_length = m.max_model_len
```

qgallouedec: there is never a case where a user would need a different max length in its training than the max length in the server

AmineDiro: I see. In this PR I added this function that fetches the max length from vllm :
```python
    def _fetch_max_length_vllm(self, fetch_timeout=5.0):
```

My thinking is that we might need to train on _shorter_ sequences for memory reasons but keep the server spawned with longer sequences for diversity.  Tbh, I don't have a strong opinion on this

qgallouedec: Oh, right, I hadn't noticed that function. I'd recommend going with the simplest solution for now. If, in the future, we need two different values for the maximum length of the server and the trainer, we can add that feature then.

qgallouedec: Wdym by diversity?

AmineDiro: > Wdym by diversity?

Just that we could want to generate sequences with len > training_len while training nothing fancy :)

qgallouedec:
this one went stale, I'm closing it. If it's still needed please open a fresh new pr

PR #5318, opened 2026-03-20
(2/5) Refactor RolloutCompletion in Async Rollout Worker 

AmineDiro:
# What does this PR do?
- Introduces structured dataclasses for the async rollout pipeline: `RolloutCompletion`, `TurnRecord`, `ToolCallRecord`, and `TaggedMessage`.
- Refactors `RolloutGroup` to store a unified list of `RolloutCompletion` objects instead of maintaining five separate parallel arrays (completions, completions_ids, tool_mask, etc.).
- Updates `_generate_one` and `_score_group` to construct and interact with these new objects natively, drastically improving code readability and making it easier to track multi-turn properties (like execution time and tool arguments)   in future PRs.

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

---

> [!NOTE]
> **Medium Risk**
> Moderate refactor of async rollout generation/scoring data flow (including tool-call accounting), which could affect token/logprob masks and metrics if any edge cases are missed; coverage is improved with new unit tests.
> 
> **Overview**
> **Refactors the async GRPO rollout pipeline to use structured dataclasses** (`RolloutCompletion`, `TurnRecord`, `ToolCallRecord`, `TaggedMessage`) instead of parallel arrays for completion text/ids/logprobs/tool masks and tool-call counts.
> 
> `_generate_one` now returns a `RolloutCompletion` with per-turn timing, tool-call records (including args, failures, and durations), and a trajectory that avoids duplicating prompt/context messages across turns; `_score_group` is updated to consume these helpers for `completion_ids`, logprobs/masks, and tool metrics.
> 
> Adds a comprehensive `test_async_rollout_worker.py` suite covering completion flattening/masks, multi-turn trajectory uniqueness, truncation at max tool-calling turns, and tool metric computation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 21360b6a01ba638e4dbd7e6f75fba2ab40b5d1aa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #5321, opened 2026-03-20
(4/5) async grpo break out of generation loop  (is_done)

AmineDiro:
🔴 🔴  **IMPORTANT**:  Depends on: [feature/async-grpo-data-classes](https://github.com/huggingface/trl/pull/5318)

# What does this PR do?
- Dynamically inspects the environment factory for an `is_done` method upon initialization.
- Passes `is_done` into the `_generate_one` method and breaks the multi-turn generation loop early if the environment signals it has completed its objective.
- Includes a small fix using `get_json_schema` to ensure environment tools pass transformers schema validation.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk: changes the rollout data model and multi-turn generation control flow, which can affect reward inputs, token accounting, and tool-calling behavior in async sampling.
> 
> **Overview**
> Adds an environment-driven early-exit path for multi-turn tool-calling: `AsyncRolloutWorker` now discovers an optional `is_done` method per environment instance and stops `_generate_one` once it returns true.
> 
> Refactors rollout outputs from flat lists into structured records (`RolloutCompletion`/`TurnRecord`/`ToolCallRecord`) that track per-turn messages, token ids/logprobs, tool-response suffix ids, and timing; scoring/reward inputs, tool metrics, and completion masks/logprobs are updated to derive from this structure. Also pre-converts environment bound methods to JSON schema via `get_json_schema` so they pass Transformers tool validation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ae9ea3e73ddf738b2a305873e9f27aec06c22ccd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


AmineDiro:
**Discussion continued from [PR #5299](https://github.com/huggingface/trl/pull/5299) regarding the `is_done` flag:**

---

**@qgallouedec:**
I strongly think that we shouldn't add any done signal, see #5235 (comment)

Requiring extra methods unnecessarily narrows the set of valid implementations and reduces substitutability. We should depend on the minimal protocol the class actually needs. There is nothing we can't do without `is_done`, so I'd recommend simply not having it. Plus at inference, the model generation won't have access to this flag anyway, so it's useless

**@AmineDiro:**
The done was added for this case (cf earlier comment):
> The issue we currently have is that user can't define a stop function to exit the generation loop. For example, when in code environment, we want to exist if the tests succeed but the current setup keeps rolling until we reach `max_turns` or token exhaustion.

**@qgallouedec:**
yes I know, but ideally, the model should learn no to continue calling the tool if it gets an error each time. IMO the right way to handle done would be:
```python
class MyEnv:
    def reset(self, **kwargs):
        self._done = False

    def my_func(self):
        """Some nice documentation"""
        if self._done:
            raise Exception("Session expired, you can't use this tool anymore!")
        if some_condition:
            self._done = True
        return
```
If the generation stops when the environment is done, then the model never learns to know that the [environment is done] implies that [it can't use it anymore].

**@AmineDiro:**
Sorry, maybe I didn't explain well the usage of the `is_done` in this portion codebase. The current impl does the following:

```python
async def _generate_one(...):
    while True:
        # .... 
        if tool_calls is None or (max_iterations is not None and iteration_num >= max_iterations):
            return ...
```
Now we can have done() raise an Exception but I am not sure it provides a clean signal to know that this rollout had exception because of some env specific errors vs it genuinely reached end of turn. If we go with Exception route, I would image that we need to define a library side Exception like `StopGeneration(Exception)` to cleanly catch it and know when env has stopped. But then this is equivalent to defining a done function IMO.

Also having a done env defined tool extends tools context for the LLM which can be a good choice in some specific cases but overall maybe unnecesary ??

**@qgallouedec:**
So if we don't use the done flag to stop the generation, what's the point of having it? I can see it's used here, but the reason is not clear:
https://github.com/huggingface/trl/pull/5299/changes#diff-175873f8c2363ec8322c4a0af91437e83da991bb5d388f2d01342b2d31eaf8a3R822

it seems like a proxy the a new `is_truncated` flag, which I don't quite understand either. The truncation is equivalent to "the last token isn't an EOS", why adding this flag here?

But more generally, don't you think it's out of the scope of this PR?

**@AmineDiro:**
> *So if we don't use the done flag to stop the generation, what's the point of having it? I can see it's used here, but the reason is not clear:*

To be precise, I put the previous generation loop. The changed one does use it to break : line
```python
 if is_done is not None and is_done():
     return _build_completion(turns, truncated=False, total_duration=time.monotonic() - t_start)
```
> *it seems like a proxy the a new `is_truncated` flag, which I don't quite understand either. The truncation is equivalent to "the last token isn't an EOS", why adding this flag here?*

Truncation is separate from done and is mainly for metrics and debugging purposes. It can be computed from "isn't last EOS" but its just a bool and recomputing something that small seems unnecessary ?

qgallouedec:
I took another look at this PR, and there’s a pattern here, I’m not sure if it’s intentional. Let me know if I’m understanding this correctly:

The `is_done` check fires after tool execution but before the next generation turn. This means that when the environment terminates, the rollout ends with a tool response as the last message, ie, the model never generates a closing assistant turn.
`[prompt  | tool_call | tool_response]` instead of `[prompt | tool_call | tool_response | assistant_final_message]`

AmineDiro:
> The `is_done` check fires after tool execution but before the next generation turn. This means that when the environment terminates, the rollout ends with a tool response as the last message, ie, the model never generates a closing assistant turn.

yes you are right

qgallouedec:
this one went stale, I'm closing it. If it's still needed please open a fresh new pr

PR #5338, opened 2026-03-22
Fix PPO reward placement at EOS token and missing_eos_penalty stop-token mismatch

### LINKED ISSUE #3750: Two potential issues in PPOTrainer: reward added to PAD token instead of EOS, and missing_eos_penalty may not apply
### Reproduction

Hi, I’m currently reading through the implementation of PPOTrainer in the trl library and found two potential mismatches between the implementation and the intended behavior described in prior work on RLHF, particularly in *“The N+ Implementation Details of RLHF with PPO: A Case Study on TL;DR Summarization”*. 
## First: The position where the scores are added
In the current trl PPO training loop, the reward score is added to the token at the following index:
```python
actual_start = torch.arange(rewards.size(0), device=rewards.device)
actual_end = torch.where(sequence_lengths_p1 < rewards.size(1), sequence_lengths_p1, sequence_lengths)
rewards[[actual_start, actual_end]] += scores

```

where sequence_lengths_p1 is defined as:

```python
sequence_length = first_true_indices(postprocessed_response == processing_class.pad_token_id) - 1
sequence_lengths_p1 = sequence_lengths + 1
    ...
```
This means that in most cases, the reward score (scores) is added to the first **PAD token** in the sequence, rather than the **EOS token**.

However, as explained in the paper mentioned above, and also consistent with the description in the "*From r to Q∗: Your Language Model is Secretly a Q-Function*" paper (Rafailov et al., 2024), the reward score should ideally be assigned at the **EOS token**, since that is where the reward model is expected to backpropagate during training. If the EOS token is missing, the response may be penalized (e.g., with a score of -1), but when EOS is present, the reward should be anchored on it, not on the subsequent padding.

Assigning the score to the PAD token may introduce issues, such as if the PAD position is later masked (which is usually the case), the reward score may not actually contribute to training. 

## Second: The missing_eos_penalty
In the *Response Processing 3. Filter completion. Ensure that the sample contains stop_token_id* :
```python
contain_eos_token = torch.any(postprocessed_responses == self.processing_class.eos_token_id, dim=-1)
```
However, if we explicitly set:

```python
self.policy_model.generation_config.eos_token_id = self.stop_token_id = args.stop_token_id
```
Then generation will **correctly terminate at** the `stop_token_id` (e.g., `<|im_end|>`). Despite that, the above check still looks for the default `eos_token_id` (e.g., `<|endoftext|>`), so the condition fails even though the generation stopped as intended.

I’d like to ask whether this current behavior is intentional—perhaps for implementation simplicity or alignment with certain tokenization behaviors—or whether it might be an unintended bug that could affect training efficacy. Understanding the motivation behind this design choice would be very helpful.

I’d be happy to discuss this further.

Thanks for your great work on this library!

### System Info

None

### Checklist

s-zx:
## Summary

Two related bugs in `trl/experimental/ppo/ppo_trainer.py`:

### Bug 1 — Reward placed at PAD token instead of EOS

The score is accumulated at `actual_end = sequence_lengths_p1 = sequence_lengths + 1`, which points to the **first PAD token after EOS**.  Because the PAD position is later zeroed out by the loss mask, the reward score effectively does not contribute to training.

Per prior RLHF literature (*"N+ Implementation Details of RLHF with PPO"*, Rafailov et al. *"From r to Q*"*), the score should be assigned at the **EOS token** itself.

```python
# Before — score lands on PAD
actual_end = torch.where(sequence_lengths_p1 < rewards.size(1), sequence_lengths_p1, sequence_lengths)

# After — score lands on EOS
actual_end = sequence_lengths  # place score at EOS token, not at the following PAD token
```

`sequence_lengths` is always within bounds: for max-length (unpadded) responses it already equals `response_len - 1`.

### Bug 2 — `missing_eos_penalty` never triggers with custom stop token

```python
contain_eos_token = torch.any(postprocessed_responses == self.processing_class.eos_token_id, dim=-1)
```

When a custom `stop_token_id` is configured that differs from `eos_token_id`, the truncation step replaces tokens after the stop token with PAD, but the EOS-presence check still looks for `eos_token_id`.  If the model never emits `eos_token_id` (only the custom stop token), `contain_eos_token` is always `False` and `missing_eos_penalty` is never applied.

```python
# After — use stop_token_id (falls back to eos_token_id when stop_token_id is None)
_stop_id = self.stop_token_id if self.stop_token_id is not None else self.processing_class.eos_token_id
contain_eos_token = torch.any(postprocessed_responses == _stop_id, dim=-1)
```

Fixes #3750

---

> [!NOTE]
> **Medium Risk**
> Changes PPO reward/penalty logic in `ppo_trainer.py`, which can materially affect training signals and learned behavior, especially for runs using custom stop tokens. Low implementation complexity, but high impact on model optimization dynamics.
> 
> **Overview**
> Fixes two PPO rollout-processing issues in `trl/experimental/ppo/ppo_trainer.py`.
> 
> `missing_eos_penalty` now checks for the configured `stop_token_id` (falling back to `eos_token_id`) so completions that omit the *actual* stop token are penalized correctly.
> 
> The per-sample score is now added to `rewards` at the EOS position (`sequence_lengths`) instead of the following PAD position, ensuring the scalar reward is not masked out during training.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6fab5af5e453d445e2dfc6efbe0ec218456672e1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/trl/experimental/ppo/ppo_trainer.py
+++ b/trl/experimental/ppo/ppo_trainer.py
@@ -750,7 +750,10 @@ class PPOTrainer(_BaseTrainer):
 
                 # Response Processing 3. Filter completion. Ensure that the sample contains stop_token_id
                 # Completions not passing that filter will receive a lower score.
-                contain_eos_token = torch.any(postprocessed_responses == self.processing_class.eos_token_id, dim=-1)
+                # Use stop_token_id (which may differ from eos_token_id) so that missing_eos_penalty
+                # is applied correctly when a custom stop token is configured.
+                _stop_id = self.stop_token_id if self.stop_token_id is not None else self.processing_class.eos_token_id
+                contain_eos_token = torch.any(postprocessed_responses == _stop_id, dim=-1)
                 if self.args.missing_eos_penalty is not None:
                     scores[~contain_eos_token] -= self.args.missing_eos_penalty
                 # accelerator.print(f"{scores=}, {(contain_eos_token.sum() / len(contain_eos_token))=}")
@@ -771,7 +774,7 @@ class PPOTrainer(_BaseTrainer):
                 non_score_reward = -args.kl_coef * kl
                 rewards = non_score_reward.clone()
                 actual_start = torch.arange(rewards.size(0), device=rewards.device)
-                actual_end = torch.where(sequence_lengths_p1 < rewards.size(1), sequence_lengths_p1, sequence_lengths)
+                actual_end = sequence_lengths  # place score at EOS token, not at the following PAD token
                 rewards[actual_start, actual_end] += scores
 
                 # 5. whiten rewards


qgallouedec:
We’re going to close this PR because it appears to be largely auto-generated. While the content may be valid, PRs like this often address non-existent or extremely edge-case issues, which can divert limited reviewer attention away from problems that real users actually face.

If you believe this is a false positive, feel free to open a new PR with a clear reproducer for the issue you’re trying to solve, and make sure to carefully follow the pull request template.

PR #5349, opened 2026-03-23
Add chunked LM head for memory-efficient log-prob computation  for AsyncGRPOTrainer

AmineDiro:
# What does this PR do?

 - Add `chunk_lm_head.py` a custom function that computes per-token log-probs and entropy without materializing the full [N, V] logits tensor, using online logsumexp 
 - Add `chunk_lm_head_size` config parameter to `AsyncGRPOConfig` to enable chunked mode with a configurable chunk size
 - Patch the model's forward pass to skip prompt tokens via `completion_mask` to avoid expensive matmuls on non-completion positions
 - Mutually exclusive with `use_liger_kernel` (both replace the LM head forward pass) !! 

## Tests
 - Unit tests for forward/backward numerical correctness against reference implementation
 - Tests for `bfloat16` gradient accuracy 
 - Tests for completion_mask correctness (masked matches unmasked at completion positions, zero at prompt positions)
 - Backward pass tests with `completion_mask`

## Results

<details>
<summary>Benchmark Results & Script</summary>

----

### Benchmark Script
This script uses `AsyncGRPOTrainer` with a synthetic 8192-token sequence to profile memory usage across different chunk sizes.

```python
"""
Memory profiling script for AsyncGRPOTrainer with a single 8192-token sequence.

Runs three configurations of chunk_lm_head (None, 4096, 8192) and prints
a comparison table of peak memory usage.

Outputs:
  - memory_snapshot_chunk_{none,4096,8192}.pickle  (load in https://pytorch.org/memory_viz)
  - Comparison table of peak allocated/reserved GB and wall time
"""

import gc
import itertools
import os
import pickle
import queue
import tempfile
import time

import torch

from trl.experimental.async_grpo.async_rollout_worker import RolloutSample


def dummy_reward_func(completions, **kwargs):
    return [float(hash(c[0]["content"]) % 100) / 100.0 for c in completions]


class _StubRolloutWorker:
    """Minimal rollout worker stub for testing the trainer in isolation."""

    def __init__(self, tokenizer, dataset, num_generations: int = 8, samples_per_weight_sync: int = 10):
        self.rollout_buffer = queue.Queue()
        self._samples_per_weight_sync = samples_per_weight_sync
        self._model_version = 0
        self._sample_iter = self._make_sample_iter(tokenizer, dataset, num_generations)

    def _make_sample_iter(self, tokenizer, dataset, num_generations):
        import itertools

        import numpy as np

        for row in itertools.cycle(dataset):
            completions = [
                [{"role": "assistant", "content": f"{row['completion'][0]['content']} {idx}"}]
                for idx in range(num_generations)
            ]
            prompt_completions = [row["prompt"] + completion for completion in completions]
            prompt_ids = tokenizer.apply_chat_template(
                row["prompt"], tokenize=True, add_generation_prompt=True, return_dict=False
            )
            prompt_completion_ids = tokenizer.apply_chat_template(
                prompt_completions, tokenize=True, add_generation_prompt=False, return_dict=False
            )
            rewards = np.array(dummy_reward_func(completions))
            advantages = (rewards - rewards.mean()) / (rewards.std() + 1e-8)
            for idx in range(num_generations):
                completion_ids = prompt_completion_ids[idx][len(prompt_ids) :]
                yield RolloutSample(
                    prompt=row["prompt"],
                    completion=completions[idx],
                    input_ids=prompt_ids + completion_ids,
                    completion_mask=[0] * len(prompt_ids) + [1] * len(completion_ids),
                    old_log_probs=[0.0] * len(prompt_ids) + [-0.5] * len(completion_ids),
                    advantage=float(advantages[idx]),
                    model_version=self._model_version,
                    metrics={"reward": float(rewards[idx]), "reward_std": float(rewards.std())},
                )

    def _fill_queue(self):
        for _ in range(self._samples_per_weight_sync):
            self.rollout_buffer.put(next(self._sample_iter))

    def start(self):
        self._fill_queue()

    def update_model_version(self, version):
        self._model_version = version
        self._fill_queue()

    def stop(self):
        pass

    def pause(self):
        pass

    def resume(self):
        pass

    def send_weights(self, iterator):
        pass


class _LongSeqStubRolloutWorker(_StubRolloutWorker):
    """Stub that emits a single synthetic 8192-token sequence per sample."""

    SEQ_LEN = 8192

    def _make_sample_iter(self, tokenizer, dataset, num_generations):
        vocab_size = tokenizer.vocab_size or 32000
        input_ids = [i % vocab_size for i in range(self.SEQ_LEN)]
        completion_mask = [0] * (self.SEQ_LEN // 2) + [1] * (self.SEQ_LEN // 2)
        old_log_probs = [0.0] * self.SEQ_LEN
        prompt = [{"role": "user", "content": "x"}]
        completion = [{"role": "assistant", "content": "y"}]
        for _ in itertools.cycle([None]):
            yield RolloutSample(
                prompt=prompt,
                completion=completion,
                input_ids=input_ids,
                completion_mask=completion_mask,
                old_log_probs=old_log_probs,
                advantage=1.0,
                model_version=self._model_version,
                metrics={"reward": 1.0, "reward_std": 0.0},
            )


def run_profile(chunk_lm_head_value, model_id, dataset, tokenizer):
    """Run a single profiling pass with the given chunk_lm_head setting.

    Returns a dict with label, peak_alloc_gb, peak_reserved_gb, elapsed_ms.
    """
    label = str(chunk_lm_head_value) if chunk_lm_head_value is not None else "none"
    print(f"\n{'=' * 60}")
    print(f"Running with chunk_lm_head = {chunk_lm_head_value}")
    print(f"{'=' * 60}")

    tmp_dir = tempfile.mkdtemp()
    config_kwargs = dict(
        output_dir=tmp_dir,
        max_steps=1,
        num_generations=1,
        max_completion_length=8192,
        report_to="none",
        learning_rate=1e-6,
        logging_steps=1,
    )
    if chunk_lm_head_value is not None:
        config_kwargs["chunk_lm_head_size"] = chunk_lm_head_value

    args = AsyncGRPOConfig(**config_kwargs)

    trainer = AsyncGRPOTrainer(
        model=model_id,
        reward_funcs=lambda completions, **kw: [1.0] * len(completions),
        args=args,
        train_dataset=dataset,
        rollout_worker=_LongSeqStubRolloutWorker(tokenizer, dataset, num_generations=1),
    )

    torch.cuda.memory._record_memory_history(max_entries=100000)

    t0 = time.perf_counter()
    trainer.train()
    torch.cuda.synchronize()
    elapsed = time.perf_counter() - t0

    peak_alloc_gb = torch.cuda.max_memory_allocated() / 1024**3
    peak_reserved_gb = torch.cuda.max_memory_reserved() / 1024**3

    # Save memory snapshot
    snapshot = torch.cuda.memory._snapshot()
    snapshot_path = f"memory_snapshot_chunk_{label}.pickle"
    with open(snapshot_path, "wb") as f:
        pickle.dump(snapshot, f, protocol=4)
    torch.cuda.memory._record_memory_history(enabled=None)
    print(f"Saved: {os.getcwd()}/{snapshot_path}")
    print(
        f"Peak allocated: {peak_alloc_gb:.2f} GB | Peak reserved: {peak_reserved_gb:.2f} GB | Time: {elapsed * 1000:.1f} ms"
    )

    # Cleanup to free GPU memory before next run
    del trainer
    gc.collect()
    torch.cuda.empty_cache()

    return {
        "label": chunk_lm_head_value if chunk_lm_head_value is not None else "None",
        "peak_alloc_gb": peak_alloc_gb,
        "peak_reserved_gb": peak_reserved_gb,
        "elapsed_ms": elapsed * 1000,
    }


def main():
    model_id = "trl-internal-testing/tiny-Qwen2ForCausalLM-2.5"
    dataset = load_dataset("trl-internal-testing/zen", "conversational_prompt_completion", split="train")

    configs = [None, 4096, 8192]
    results = []

    for chunk_val in configs:
        result = run_profile(chunk_val, model_id, dataset, tokenizer)
        results.append(result)

    # Print comparison table
    baseline_alloc = results[0]["peak_alloc_gb"]
    baseline_reserved = results[0]["peak_reserved_gb"]

    print(f"\n{'=' * 95}")
    print("COMPARISON TABLE")
    print(f"{'=' * 95}")
    print(
        f"{'chunk_lm_head':>15} | {'Peak Alloc (GB)':>16} | {'Alloc Reduction':>16} | {'Peak Reserved (GB)':>18} | {'Wall Time (ms)':>15}"
    )
    print(f"{'-' * 15}-+-{'-' * 16}-+-{'-' * 16}-+-{'-' * 18}-+-{'-' * 15}")
    for r in results:
        reduction = f"{baseline_alloc / r['peak_alloc_gb']:.2f}x" if r["peak_alloc_gb"] > 0 else "N/A"
        print(
            f"{str(r['label']):>15} | {r['peak_alloc_gb']:>16.2f} | {reduction:>16} | {r['peak_reserved_gb']:>18.2f} | {r['elapsed_ms']:>15.1f}"
        )
    print(f"{'=' * 95}")


if __name__ == "__main__":
    main()

```

</details>

### COMPARISON TABLE

| chunk_lm_head | Peak Alloc (GB) | Alloc Reduction | Peak Reserved (GB) | Wall Time (ms) |
| :--- | :--- | :--- | :--- | :--- |
| None | 18.55 | 1.00x | 20.89 | 808.7 |
| 4096 | 0.42 | **44.32x** | 4.67 | 459.0 |
| 8192 | 0.76 | **24.34x** | 4.67 | 393.0 |

**Without chunked loss**
image

**After Chunked loss**
image

---

> [!NOTE]
> **Medium Risk**
> Touches core training loss computation by monkey-patching model `forward` to return chunked `log_probs`/`entropy`, which could affect correctness or compatibility across model architectures despite added tests.
> 
> **Overview**
> Adds a chunked LM-head path (`_ChunkedLogProbFunction` + `patch_chunked_lm_head`) that computes per-token log-probs and entropy via online logsumexp without materializing full `[N, V]` logits, with optional filtering via `completion_mask` to skip prompt tokens.
> 
> Updates `AsyncGRPOTrainer` to always patch the loaded model and to use the patched forward (passing `labels`/`completion_mask`) instead of computing `selective_log_softmax`/entropy from full logits; also tweaks weight streaming to ensure tensors are moved to the accelerator device.
> 
> Exports `patch_chunked_lm_head` from `trl.trainer` and adds extensive unit tests covering forward/backward numerical equivalence (including bf16 tolerances) and `completion_mask` behavior on both a dummy LM and multiple tiny HF models.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5d907dbb97370b0153f9ad4e58216d3c28150c1a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1150,8 +1150,20 @@ class _FakeCausalLM(nn.Module):
 _CHUNKED_LM_HEAD_MODEL_IDS = [
     "trl-internal-testing/tiny-CohereForCausalLM",
     "trl-internal-testing/tiny-Cohere2ForCausalLM",
-    "trl-internal-testing/tiny-DeepseekV3ForCausalLM",
-    "trl-internal-testing/tiny-DeepseekV3ForCausalLM-0528",
+    pytest.param(
+        "trl-internal-testing/tiny-DeepseekV3ForCausalLM",
+        marks=pytest.mark.skipif(
+            Version(transformers.__version__) < Version("5.0.0"),
+            reason="DeepseekV3 SDPA attention is broken in transformers < 5.0.0",
+        ),
+    ),
+    pytest.param(
+        "trl-internal-testing/tiny-DeepseekV3ForCausalLM-0528",
+        marks=pytest.mark.skipif(
+            Version(transformers.__version__) < Version("5.0.0"),
+            reason="DeepseekV3 SDPA attention is broken in transformers < 5.0.0",
+        ),
+    ),
     "trl-internal-testing/tiny-Gemma2ForCausalLM",
     "trl-internal-testing/tiny-GemmaForCausalLM",
     "trl-internal-testing/tiny-Glm4MoeForCausalLM",


AmineDiro:
@kashif  

## Benchmark against  liger's `LigerFusedLinearGRPOLoss`. 
- Liger materializes [1, S, V] logits per batch chunk inside chunk_forward. Even though it's transient (freed bygrad_and_value), the peak allocation scales as O(S * V). With V=152064 and S=32768, that's 32768 * 152064 * 4 bytes ~ 18.6 GB just for the float32 log-probs tensor alone.
- Liger is dramatically faster at every sequence length. At S=32k ->  167ms vs 1654ms (~10x).
- The fundamental difference: chunking along V trades compute efficiency for memory efficiency. Liger's approach is compute-optimal (large matmuls  + torch compiled) but memory scales with sequence length. Ours is memory-optimal but has overhead and small-matmul inefficiency.

median_time_vs_seqlen
peak_memory_vs_seqlen

qgallouedec:
That's wonderful!

A few comment comments:
- You can move this function to `trl.trainer.utils`, as it should be usable in the future for DPO, Reward, RLOO, and GRPO. 
- I recommend not necessarily adding a new argument: use a default sweet spot value, and we’ll add an argument in the future if the community requests it.
- Add more tests; vary the models and attention implementations, check [these](https://github.com/huggingface/trl/blob/4fda23e9d9cc2c714631ccfabb638385124a4d77/tests/test_utils.py#L997-L1084) as example

trl/experimental/chunk_lm_head.py
qgallouedec, line 1: can you move this to `trl.trainer.utils` instead? and same for the tests (move out from experimental

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -291,6 +292,22 @@ def __init__(
...
+        if self.args.chunk_lm_head_size is not None:
+            if getattr(model.config, "final_logit_softcapping", None) is not None:
+                logger.warning(
+                    "Model has `final_logit_softcapping` which is not supported "
+                    "by chunked LM head. Disabling `chunk_lm_head_size`."
+                )
+                self.args.chunk_lm_head_size = None
+            else:
+                patch_chunked_lm_head(model, self.args.chunk_lm_head_size, self.temperature)
+
```
qgallouedec:
```suggestion
        if self.args.use_liger_kernel:
            raise NotImplementedError("`use_liger_kernel` is not supported yet.")

        patch_chunked_lm_head(model, chunk_size=8192, temperature=self.temperature)

```

and have something like 

```python
raise NotImplementedError(
    "The model uses `final_logit_softcapping` which is not yet supported. Please open an issue if you "
    "want your model to be supported."
)
```

in `patch_chunked_lm_head`

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -444,13 +461,26 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+                completion_mask=completion_mask,
+                use_cache=False,
+            )
+            log_probs = outputs["log_probs"]
+            entropy = outputs["entropy"]
+        else:
+            outputs = model(input_ids=input_ids, attention_mask=attention_mask, use_cache=False)
+            logits = outputs.logits[:, :-1, :]
+            logits.div_(self.temperature)
+            log_probs = selective_log_softmax(logits, input_ids[:, 1:])
```
qgallouedec:
```suggestion
        outputs = model(
            input_ids=input_ids,
            attention_mask=attention_mask,
            labels=input_ids,
            completion_mask=completion_mask,
            use_cache=False,
        )
        log_probs, entropy = outputs["log_probs"], outputs["entropy"]
```

The reason I'm suggesting all these simplifications (accepting incompatibility with Liger and certain models) is because I think it's okay for this trainer to be highly optimized and less general than `GRPOTrainer`. The problem with trying to support everything from the start is that I think we’ll get bogged down in intricate, complex edge cases that will prevent us from moving forward.

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -485,9 +515,13 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
-            probs = torch.softmax(logits, dim=-1)
-            log_p = torch.log_softmax(logits, dim=-1)
-            entropy = -torch.sum(probs * log_p, dim=-1)
+            if not use_chunked:
+                assert logits is not None, "Logits are only computed on non-chunked forward"
+                probs = torch.softmax(logits, dim=-1)
+                log_p = torch.log_softmax(logits, dim=-1)
+                entropy = -torch.sum(probs * log_p, dim=-1)
+
+            assert entropy is not None
```
qgallouedec:
```suggestion
```

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: I wondered about if we should raise or just warning and thought warning was better from devx, here is my thinking: 
- we default to chunking without the user knowing so by default user doesn't know chunking is happening
- Gemma models exist in our tests and we should support them very soon, so I just disabled the chuking with a warning

wdyt ?

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: I see. 

The reason I kept these under condition is that we might want to fallback to non optimized path for checking numerical stability for example ?

trl/experimental/async_grpo/async_grpo_config.py
```diff
@@ -59,6 +59,13 @@ class AsyncGRPOConfig(_BaseConfig):
         epsilon_high (`float`, *optional*, defaults to `0.2`):
             Upper-bound epsilon value for clipping.
 
+        > Parameters that control the LM head
+
+        chunk_lm_head (`int` or `None`, *optional*, defaults to `8192`):
+            Chunk size for the fused LM head. When set, the lm_head computes log-probs and entropy without
+            materializing the full `[batch, seq, vocab]` logits tensor, processing the vocabulary in chunks of this
+            size instead. Reduces peak memory at the cost of extra matmuls. If `None`, uses the standard full-logits
+            path.
```
qgallouedec:
```suggestion
```

tests/test_chunk_lm_head.py
qgallouedec, line 1: can you just move this to `test/test_utils.py` like


```python

class TestChunkedLogProbFunction:
    N, H, V = 64, 32, 128
    CHUNK_SIZE = 32

    def _reference_logprobs_and_entropy(hidden, weight, labels, temperature): ...

    @pytest.mark.parametrize("temperature", [1.0, 0.7])
    def test_forward(temperature): ...

    @pytest.mark.parametrize("temperature", [1.0, 0.7])
    def test_backward(temperature): ...

    @pytest.mark.parametrize("temperature", [1.0, 0.7])
    def test_backward_bfloat16(temperature): ...


class _FakeTransformerModel(nn.Module): ...

class _FakeCausalLM(nn.Module): ...

class TestPatchChunkedLMHead:
    B, S = 4, 16  # batch size, sequence length (including prompt + completion)

    def _build_model_and_inputs(temperature): ...

    @pytest.mark.parametrize("temperature", [1.0, 0.7])
    def test_dummy_model_chunked_forward_with_completion_mask(temperature): ...

    @pytest.mark.parametrize("temperature", [1.0, 0.7])
    def test_dummy_model_chunked_forward_completion_mask_backward(temperature): ...

    @pytest.mark.parametrize("model_id", [...])
    @pytest.mark.parametrize("temperature", [1.0, 0.7])
    def test_forward(self, model_id, temperature): ...

    @pytest.mark.parametrize("model_id", [...])
    @pytest.mark.parametrize("temperature", [1.0, 0.7])
    def test_backward(self, model_id, temperature): ...
```

qgallouedec:
@codex review

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -292,6 +292,11 @@ def __init__(
         model_name = model
         model = AutoModelForCausalLM.from_pretrained(model, device_map=None, dtype=torch.float32)
 
+        if self.args.use_liger_kernel:
+            raise NotImplementedError("`use_liger_kernel` is not supported yet.")
+
+        patch_chunked_lm_head(model, chunk_size=8192, temperature=self.temperature)
```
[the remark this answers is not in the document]
qgallouedec: yep that's fine for now

trl/trainer/utils.py
```diff
@@ -1219,3 +1220,180 @@ def shutdown_event_loop_in_daemon(
...
+            grad_logits[row_idx, local_idx] += g * in_chunk_cond
+            grad_logits = grad_logits * inv_t
+
+            grad_hidden.add_(grad_logits @ w_chunk.float())
+            grad_weight[start:end].add_(grad_logits.t() @ hidden.float())
+
+        return grad_hidden.to(hidden.dtype), grad_weight.to(weight.dtype), None, None, None, None
+
+
+def patch_chunked_lm_head(model: torch.nn.Module, chunk_size: int, temperature: float) -> None:
```
edbeeching, line 1342: Would we expect this to work with models wrapped by deepspeed / fsdp(2) ?

AmineDiro: Yes this would work with fsdp2/ CP/ TP etc as the lm_head weight is wrapped before entering the patch in FSDP2, and the logits is independant for each token ( it's on the V dim)

edbeeching: That is good to hear, could you / have you validated this?

AmineDiro: Yes tested it using :
```sh
  CUDA_VISIBLE_DEVICES=0,1 accelerate launch --config_file examples/accelerate_configs/fsdp2.yaml examples/scripts/async_grpo.py
```
with num_procs=2.

PR #5315, opened 2026-03-19
Add truncation to SFT DataCollatorForLanguageModeling

albertvillanova:
Add truncation to SFT `DataCollatorForLanguageModeling`.

This PR adds support for sequence truncation to the `DataCollatorForLanguageModeling` class, allowing sequences to be truncated to a specified maximum length from either the start or end. It also introduces comprehensive tests to ensure correct behavior for different truncation modes and edge cases.

This PR aligns SFT `DataCollatorForLanguageModeling` with the existing SFT `DataCollatorForVisionLanguageModeling`, which already truncates inputs.

Follow-up to:
- #5306

### Changes

Enhancements to sequence truncation:

* Added `max_length` and `truncation_mode` parameters to `DataCollatorForLanguageModeling`, enabling truncation of sequences longer than `max_length` with options to keep either the start or end tokens.
* Implemented per-sequence truncation logic in the `torch_call` method, applying the specified truncation mode and ensuring associated masks (such as `completion_mask` and `assistant_masks`) are truncated consistently. A `ValueError` is raised for unsupported truncation modes.

Testing improvements:

* Added multiple test cases covering:
  - Truncation from the start and end.
  - Behavior when no truncation is needed.
  - Truncation with completion masks.
  - Handling of invalid truncation modes.

---

> [!NOTE]
> **Medium Risk**
> Changes how SFT batches are built by introducing truncation in `DataCollatorForLanguageModeling`, which can alter training inputs/labels and raise new configuration errors (especially around `padding_free` and dataset prep). Test coverage mitigates but behavior changes may impact existing training setups relying on untruncated batches.
> 
> **Overview**
> Adds optional per-example truncation to `DataCollatorForLanguageModeling` via new `max_length` and `truncation_mode` (`keep_start`/`keep_end`), ensuring `labels`, `completion_mask`, and `assistant_masks` are truncated consistently and rejecting unsupported modes.
> 
> Wires `SFTTrainer` to pass `max_length`/`truncation_mode` into the text collator (disabled when `padding_free=True`) and introduces a guard that errors when `skip_prepare_dataset=True` is used with `padding_free=True` and a non-`None` `max_length`.
> 
> Extends tests to cover truncation modes, mask interactions, invalid modes, and trainer configuration propagation/validation.
> 
[...]

PR #5359, opened 2026-03-24
Remove post-collation truncation from SFT

albertvillanova:
Remove post-collation truncation from SFT.

Follow-up to:
- #5315

Related to:
- #5350

This PR removes internal dataset truncation logic from the SFT trainer and  enforce that sequence truncation should be handled before padding. This change shifts responsibility for truncation to the dataset preparation or custom collators, making the data processing pipeline more explicit and less error-prone.

### Changes

**Removal of internal truncation logic:**

* Removed the use of the `truncate_dataset` function in the dataset processing pipeline; the trainer no longer performs truncation internally if packing is not enabled.
* Updated comments and logic in the `tokenize_fn` to reflect that only packing (not truncation) is performed during dataset preparation.

**Documentation and enforcement of truncation responsibility:**

* Updated the docstring for the `data_collator` parameter to specify that custom collators must truncate sequences before padding; the trainer will not perform truncation after collation.
* Changed error messaging in the constructor to remove the suggestion to "Disable `skip_prepare_dataset`" and clarify that inputs must be packed/truncated before reaching the collator if `padding_free=True`.

---

> [!NOTE]
> **Medium Risk**
> Changes SFT data-prep semantics by removing automatic dataset truncation when `packing=False`, which can increase sequence lengths and affect memory/accuracy unless callers pre-truncate. Adds stricter validation for `padding_free=True` with `max_length`, potentially breaking existing configs that relied on implicit enforcement.
> 
> **Overview**
> Removes **trainer-side dataset truncation** from `SFTTrainer` when `packing=False`, leaving sequence length control to pre-truncated inputs or custom collators (packing remains the only built-in length enforcement during dataset prep).
> 
> Tightens and clarifies configuration rules by raising an error when `padding_free=True` is used *without packing* alongside a non-`None` `max_length`, and updates the `data_collator` docstring to state that **custom collators must truncate before padding**.
> 
> Updates the related unit test to reflect the new error condition/message for padding-free + no packing + `max_length`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 12f95147ec86cf099ee017965581547916e9706f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1828,7 +1828,12 @@ class TestSFTTrainer(TrlTestCase):
     )
     @pytest.mark.parametrize(
         "dataset_config",
-        ["conversational_language_modeling", "conversational_prompt_completion", "standard_prompt_completion"],
+        [
+            "conversational_language_modeling",
+            "conversational_prompt_completion",
+            "standard_language_modeling",  # Regression test for #5334
+            "standard_prompt_completion",
+        ],
     )
     @require_vision
     def test_train_vlm_text_only_data(self, model_id, dataset_config):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -28,7 +28,6 @@ from trl.trainer.utils import (
     RepeatSampler,
     entropy_from_logits,
     flush_left,
-    flush_right,
     forward_masked_logits,
     generate_model_card,
     get_peft_config,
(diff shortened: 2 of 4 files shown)


tests/test_sft_trainer.py
```diff
@@ -1003,17 +1003,16 @@ def test_skip_prepare_dataset_passes_truncation_to_text_collator(self):
...
         training_args = SFTConfig(
             output_dir=self.tmp_dir,
             max_length=16,
             padding_free=True,
-            dataset_kwargs={"skip_prepare_dataset": True},
             report_to="none",
         )
 
-        with pytest.raises(ValueError, match="must be enforced during dataset preparation or packing"):
+        with pytest.raises(ValueError, match="max_length` is not enforced"):
```
```suggestion
        with pytest.raises(ValueError, match="`max_length` is not enforced"):
```

PR #5369, opened 2026-03-25
Simplify DPO tokenization

albertvillanova:
Simplify DPO `_prepare_dataset`.

This PR refactors the tokenization logic in the `DPOTrainer` to centralize and simplify how examples are tokenized, especially handling differences between conversational and non-conversational inputs and normalizing outputs for vision-language models (VLMs). The main improvement is the introduction of a new `_tokenize` helper method, which reduces code duplication and ensures consistent handling of tokenizer outputs.

Analog to:
- #5363

### Changes

**Tokenization logic centralization:**
* Added a new `_tokenize` method to the `DPOTrainer` class, which abstracts tokenization for both conversational (chat-based) and non-conversational (plain text) inputs, and normalizes outputs for VLMs by removing unnecessary batch dimensions.

**Refactoring of dataset preprocessing:**
* Updated the `tokenize_fn` function to use the new `_tokenize` method for all tokenization steps, removing duplicated logic and manual normalization of outputs for both conversational and non-conversational cases.

---

> [!NOTE]
> **Medium Risk**
> Refactors dataset preprocessing/tokenization, which can subtly change token IDs (especially for conversational/VLM processors) and thus affect training outcomes, though the intent is behavior-preserving.
> 
> **Overview**
> **Simplifies DPO dataset preprocessing tokenization.** Adds a new `DPOTrainer._tokenize` helper that handles both conversational (`apply_chat_template`) and plain-text (`processing_class(text=...)`) inputs and normalizes single-example outputs for VLM processors by unwrapping the extra batch dimension.
> 
> Updates `_prepare_dataset`’s `tokenize_fn` to use `_tokenize` for prompt/chosen/rejected tokenization, removing duplicated branching logic and ad-hoc VLM output-shape fixes while keeping the same prompt-prefix validation warnings.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 860f25525cbc969db9048f93f8901725fa23af85. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


qgallouedec:
thanks, it looks good to me, though I don't understand this comment from cursor:

> Refactors dataset preprocessing/tokenization, which can subtly change token IDs (especially for conversational/VLM processors) and thus affect training outcomes, though the intent is behavior-preserving.

@codex review to ensure the tokenization remains the same

trl/trainer/dpo_trainer.py
```diff
@@ -822,6 +822,40 @@ def __init__(
                 else:
                     self.eval_dataset = self._precompute_ref_logps(self.eval_dataset, "eval", batch_size)
 
+    def _tokenize(
+        self,
+        processing_class: PreTrainedTokenizerBase | ProcessorMixin,
```
qgallouedec, line 827: it's always called with `self.processing_class`, why not simply drop this arg?

qgallouedec: oh I see it's for the map, so that the cache can be used

qgallouedec (approved):

trl/trainer/dpo_trainer.py
albertvillanova: Yes, I had thought the same and had explored that option. Unfortunately, we cannot remove this argument because `_tokenize` is called before `self.processing_class` is initialized. This happens during the `super()__init__` call, and we need to pass the "prepared" `training_dataset` to it.

PR #5307, opened 2026-03-18
Suggest the `Json()` type for tool calling dataset format

lhoestq:
> [!NOTE]
> **Low Risk**
> Low risk documentation-only change; main risk is minor confusion if users rely on the previous legacy `tools` JSON-string recommendation for older `datasets` versions.
> 
> **Overview**
> Updates the tool-calling dataset documentation to treat `get_json_schema()` output as a Python dict (not a JSON string) and to store `tools` as a list of schemas (`[json_schema]`).
> 
> Adds guidance on constructing a `datasets.Dataset` by using the `Json()` feature type (or `on_mixed_types="use_json"`) for arbitrary tool-call arguments, and documents a fallback for `datasets<4.8` where `tools` should be stored via `json.dumps(...)`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 24226bf9b73a0d52b2b2f2241c2df4c4f35bdbac. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #5118, opened 2026-02-17
Update tool handling to support JSON string schemas in trainers

qgallouedec:
## What does this PR do?

Migrates to JSON str as the preferred format for the tools column in tool-calling datasets, while maintaining backward compatibility with `list[dict]`.

### Motivation

The current `list[dict]` format for tools has several drawbacks:

- Data corruption with tabular backends. Arrow/Parquet aligns nested structures across rows. When different rows have tool schemas with different keys, the backend inserts `None` to fill gaps, silently corrupting data. This is handled in `SFTTrainer` and `RewardTrainer` with:
  https://github.com/huggingface/trl/blob/70efa963f1c9bb88ec3144b051db6fdf4ffc10a4/trl/trainer/sft_trainer.py#L948-L951
- Poor readability. The nested dict format renders strangely in dataset viewers and hinders readability (see https://huggingface.co/datasets/trl-internal-testing/toolcall).
  Screenshot 2026-02-17 at 3 21 08 PM
- User confusion. The current format causes confusion for users building on top of TRL (e.g., [Slack thread](https://huggingface.slack.com/archives/C06AF1RUV8F/p1769614169601149)). While the fix is simple (e.g., [functiongemma-tuning-lab#1](https://huggingface.co/spaces/google/functiongemma-tuning-lab/discussions/1)), the format remains misleading.
- Inconsistent with the ecosystem. Most function-calling datasets already use JSON strings:
    - https://huggingface.co/datasets/Salesforce/xlam-function-calling-60k
    - https://huggingface.co/datasets/NousResearch/hermes-function-calling-v1
    - https://huggingface.co/datasets/Locutusque/function-calling-chatml
    - https://huggingface.co/datasets/hiyouga/glaive-function-calling-v2-sharegpt

### Changes

- `SFTTrainer` and `RewardTrainer` now deserialize tools from JSON string at tokenization time (`json.loads`), with an `isinstance` check to preserve backward compatibility with `list[dict]`.
- The dataset generation script (`generate_toolcall_dataset.py`) now serializes tools with `json.dumps`. The PRs 2 and 3 will be merged when this PR gets approved.
- Documentation updated to recommend JSON str as the canonical format, with a note that `list[dict]` is still supported but not recommended.
- Tests added for both the new JSON string format and the legacy `list[dict]` format.

PR #5376, opened 2026-03-26
Require datasets>=4.7.0 for Json dtype to prevent insertion of None values

albertvillanova:
Use datasets `Json` dtype to avoid the insertion of None values.
- Note that this applies not only to tool-calling datasets, but also to multi-modal datasets. See:
  - #5364

CC: @SolarWindRider, @qgallouedec 

This PR updates dataset handling to improve compatibility with mixed data types and simplifies preprocessing in several trainer modules. The most significant changes include upgrading the `datasets` library to support the `Json` new feature, using the `on_mixed_types="use_json"` option when loading datasets, and removing the `remove_none_values` cleaning step from all trainer classes: DPO, Reward, SFT.

Related to:
- #5307
- #5310
- #5311
- #5118

### Changes

**Dataset compatibility and preprocessing improvements:**

* Upgraded the `datasets` library version requirement from `>=3.0.0` to `>=4.7.0` in `pyproject.toml` to enable support for `Json` type fields and the `on_mixed_types="use_json"` option.
* Updated dataset loading calls in `scripts/generate_toolcall_dataset.py` to use `on_mixed_types="use_json"`, ensuring robust handling of mixed-type columns. 

**Trainer codebase simplification:**

* Removed all imports and usage of the `remove_none_values` utility function from `DPOTrainer`, `RewardTrainer`, and `SFTTrainer`, eliminating a redundant data cleaning step now handled by the updated dataset loading logic.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it bumps the minimum `datasets` version and changes trainer preprocessing behavior, which could affect users with older/saved datasets that still contain injected `None` values.
> 
> **Overview**
> Switches TRL to rely on `datasets` v4.7.0+ Json dtype handling for nested/mixed columns (via `on_mixed_types="use_json"` in dataset generation scripts) to prevent Arrow/Parquet from inserting spurious `None` values.
> 
> Removes the trainers’ automatic `remove_none_values` transform in `SFTTrainer`, `DPOTrainer`, and `RewardTrainer`, and documents the new expectation + manual workaround in `MIGRATION.md` for users with legacy datasets that still contain `None`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 27ba6226ca7ee29554ea00ab15284827ce25e719. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -15,6 +15,26 @@ This guide covers the breaking changes introduced in TRL v1 and how to update yo
 | --- | --- | --- | --- | --- |
 | `SFTConfig` | `packing` | `"bfd-requeue"` | `"bfd_split"` | Replace `packing="bfd-requeue"` with `packing="bfd_split"`. The old value will still be accepted for a few versions but will be removed in a future release. |
 
+## Removed automatic `None` stripping from trainer preprocessing
+
+TRL trainers (SFT, DPO, Reward) no longer automatically strip `None` values from dataset examples during preprocessing. Previously, each trainer applied `remove_none_values` via `dataset.with_transform` to work around tabular backends (Arrow/Parquet) inserting `None` for missing keys in nested structures.
+
+This affects datasets that contain `None` values because they were:
+
+- Created before `datasets` v4.7.0, which introduced the Json dtype that preserves nested structures without inserting `None`.
+- Created with `datasets` v4.7.0 or later, but saved without using the Json feature.
+
+**Action needed:** If your dataset falls into one of the above categories and contains `None` values in nested columns, apply the fix manually before training:
+
+```python
+from trl.trainer.utils import remove_none_values
+
+dataset = dataset.with_transform(remove_none_values)
+trainer = SFTTrainer(..., train_dataset=dataset)
+```
+
+Datasets created or re-saved with `datasets` v4.7.0+ using the Json dtype are unaffected.
+
 ## Migrating from an earlier version
 
 Depending on which version you're migrating from, refer to the [release notes](https://github.com/huggingface/trl/releases) for v0.29 and earlier for version-specific changes.


trl/trainer/dpo_trainer.py
```diff
@@ -863,11 +862,6 @@ def _prepare_dataset(
         args: DPOConfig,
         dataset_name: str,
     ) -> Dataset | IterableDataset:
-        # Tabular backends like Arrow/Parquet insert `None` for mismatched keys in nested structures. Clean them from
-        # sampled data.
-        if isinstance(dataset, Dataset):  # IterableDataset does not support `with_transform`
-            dataset = dataset.with_transform(remove_none_values)
-
```
albertvillanova, line 870: I have kept the `remove_none_values` function so we can suggest in the docs using it for datasets created without the `Json` dtype and therefore already containing `None`s.

What do you think?

SolarWindRider: Good !

pyproject.toml
```diff
@@ -30,7 +30,7 @@ classifiers = [
 requires-python = ">=3.10"
 dependencies = [
     "accelerate>=1.4.0",
-    "datasets>=3.0.0",
+    "datasets>=4.7.0",  # Support Json type and on_mixed_types="use_json"
```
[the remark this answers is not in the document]
albertvillanova: I opened an independent PR to remove the `requirements.txt` file:
- https://github.com/huggingface/trl/pull/5377

scripts/generate_harmony_dataset.py
```diff
@@ -91,7 +91,7 @@ def main(test_size, push_to_hub, repo_id):
             {"reasoning_effort": "high", "model_identity": "You are Tiny ChatGPT, a tiny language model."},
             {"reasoning_effort": "low", "model_identity": "You are Tiny ChatGPT, a tiny language model."},
         ]
-    })
+    }, on_mixed_types="use_json")
```
qgallouedec, line 94: here it's because we mix turns with and without "thinking" 👍

scripts/generate_toolcall_dataset.py
```diff
@@ -222,7 +222,7 @@ def get_wind_conditions(city: str, unit: str) -> tuple[int, str]:
             json.dumps([start_timer, create_reminder]),
             json.dumps([get_weather_forecast, get_wind_conditions]),
         ]
-    })
+    }, on_mixed_types="use_json")
```
qgallouedec, line 224: here I'm not sure if it's need, `json.dumps` converts to str

trl/trainer/dpo_trainer.py
qgallouedec: Thanks, so technically it's a breaking change for datasets created without the Json dtype? If so please add it to the migration guide

qgallouedec: > I have kept the remove_none_values

Yes good idea, can you just add a comment above the function so that we remember why it exists but not used in the codebase?

scripts/generate_toolcall_dataset.py
albertvillanova: Oops, I forgot to remove `json.dumps`. :sweat_smile: I'm pushing a new commit for this.

trl/trainer/dpo_trainer.py
albertvillanova: Done!

PR #5381, opened 2026-03-26
Fix DAPO token-level loss to use prompt-level aggregation

### LINKED ISSUE #5375: DAPO token level loss is incorrectly implemented
https://github.com/huggingface/trl/blob/7f1dd11bc0d0951d0ee6e621caeee70b14df0acf/trl/experimental/gspo_token/grpo_trainer.py#L119

It seems like TRL currently acheives the token level loss proposed by DAPO by treating all tokens equally in the batch. If this is indeed true, this is different from what DAPO actually proposes. DAPO treats all tokens across generations for its respective prompt equally, rather than treating all tokens in a given batch (which could include multiple prompts and its generations) equally.

Image

Granted, the paper doesn't make it explicitly clear whether prompts should then be equally averaged or further balanced on a token-level, it seems like follow-up papers such as [ScaleRL](https://arxiv.org/pdf/2510.13786) presume DAPO means token-level aggregation at the prompt level.

Image

matdou:
# What does this PR do?

Fix the DAPO token-level loss aggregation to use prompt-level averaging instead
of batch-level averaging, as specified in the DAPO paper.

The `dapo` loss branch was normalizing over all tokens in the batch globally:
```python
normalizer = inputs["num_items_in_batch"] / self.accelerator.num_processes
loss = (per_token_loss * completion_mask).sum() / normalizer
```

This up-weights longer sequences since a prompt with 6 tokens contributes twice
as much as a prompt with 3 tokens. The DAPO paper specifies prompt-level
averaging: average token losses within each prompt first, then average across
prompts. The fix follows the same pattern already used by the `grpo` loss type
in the same file:
```python
per_prompt_loss = (per_token_loss * completion_mask).sum(-1) / completion_mask.sum(-1).clamp(min=1.0)
loss = per_prompt_loss.mean()
normalizer = self.current_gradient_accumulation_steps if mode == "train" else 1.0
loss = loss / normalizer
```

Fixes #5375

## Before submitting
      to it if that's the case. → #5375
- [x] Did you write any new necessary tests? → Added unit test with unequal prompt
      lengths verifying prompt-level vs batch-level aggregation produce different values.

## Who can review?

---

> [!NOTE]
> **Medium Risk**
> Changes the `dapo` loss computation in multiple trainers, which will alter optimization dynamics and reported losses, especially with variable-length completions. Risk is moderate because it affects core training math but is a localized, well-defined normalization change with a targeted unit test.
> 
> **Overview**
> Fixes `dapo` loss aggregation to match the paper: **average token losses per prompt, then average across prompts**, instead of globally summing tokens and normalizing by `num_items_in_batch` (which over-weights longer completions).
> 
> Applies this prompt-level averaging in `trl/trainer/grpo_trainer.py` and the experimental trainers `gspo_token/grpo_trainer.py`, `papo_trainer.py`, and `dppo_trainer.py`, and adds a unit test (`tests/experimental/test_gspo_token_trainer.py`) that asserts prompt-level vs batch-level averaging diverge for unequal completion lengths.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c763811555e1c0f7ec01b59a1aac7d80b11eb2fa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/experimental/test_gspo_token_trainer.py
+++ b/tests/experimental/test_gspo_token_trainer.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 
+import unittest
+
 import torch
 from datasets import load_dataset
 from transformers.utils import is_peft_available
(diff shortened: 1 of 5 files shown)


qgallouedec:
Thanks for the PR @matdou, and sorry for the slow turnaround.

Digging into the linked issue #5375, we confirmed that TRL's current `dapo` loss already matches the DAPO reference implementation in verl (`loss_agg_mode="token-mean"`: a flat token average across all responses and prompts), so this isn't actually a deviation from the paper.

Separately, the change here averages per *completion* (`(per_token_loss * mask).sum(-1) / mask.sum(-1)).mean()` rather than pooling the G responses within each prompt, which makes `dapo` identical to the existing `grpo` loss type. True prompt-level aggregation would need per-group pooling first. Since `dapo` is the default loss_type, silently redefining it would break reproducibility for existing users; the direction the team is open to is adding the ScaleRL-style aggregation as a *new* opt-in `loss_type` once there's more demand. I'd lean toward closing this in favor of that route. Thanks again!

PR #5390, opened 2026-03-27
feat(grpo): add stop_tool_names for immediate agent loop termination

### LINKED ISSUE #5389: # [Feature Request] GRPOTrainer: support stop-tool termination in agent training loop
# [Feature Request] GRPOTrainer: support stop-tool termination in agent training loop

## Summary

When training agents with `GRPOTrainer` using `tools` + `max_tool_calling_iterations`, there is no way to designate a tool as a **termination signal** — a tool whose invocation means "the agent is done, do not generate any further". This creates a systematic divergence between training and evaluation when using frameworks like [smolagents](https://github.com/huggingface/smolagents) that rely on a `final_answer` tool to end agent runs.

## Current behaviour

The agent loop in `_tool_call_loop` terminates only when:
1. `max_tool_calling_iterations` is reached, or
2. The model generates a turn with no tool calls (EOS).

If the model calls `final_answer(answer=42)`, TRL executes it, appends the result to the context, and then calls `_generate_single_turn` again — the model must now generate *another* turn before the loop can exit. This wastes context budget (potentially hitting `max_completion_length` before EOS), and more importantly, it means the agent's training behaviour differs from its evaluation behaviour.

## Proposed behaviour

Add a `stop_tool_names: list[str] | None` field to `GRPOConfig`. When a sample's tool call matches any name in this list:

1. The tool is executed normally and its result is appended to the context.
2. `_generate_single_turn` is **not** called for that sample — it exits the loop immediately.
3. The `tool_mask` and `completion_ids` are updated to reflect the tool result tokens (masked out of loss), exactly as they would be for a normal tool call.

This mirrors the `final_answer` semantics used in smolagents and similar agentic frameworks.

## Proposed API

```python

def final_answer(answer: int) -> int:
    """Submit the final answer."""
    return answer

config = GRPOConfig(
    ...,
    stop_tool_names=["final_answer"],
    max_tool_calling_iterations=10,
)

trainer = GRPOTrainer(
    ...,
    args=config,
    tools=[tool_a, tool_b, final_answer],
)
```

With this config, as soon as any sample's model calls `final_answer`, that sample exits the agent loop. Other samples in the batch continue their own tool-calling turns as usual.

## Motivation: training/eval alignment

When evaluating with smolagents, `final_answer` is a built-in termination tool. If the model is trained with TRL *without* stop-tool support, it learns to:
- Call `final_answer` → observe the result echoed back → generate more tool calls → eventually hit `max_tool_calling_iterations`

This is a different behaviour from what it will encounter at eval time (call `final_answer` → done). Over many gradient steps, the model's policy diverges from the target behaviour.

With `stop_tool_names`, training and evaluation share the same agent loop semantics.

## Implementation sketch

The change is contained in two files:

**`trl/trainer/grpo_config.py`** — add field:
```python
stop_tool_names: list[str] | None = field(
    default=None,
    metadata={"help": "Tool names that terminate the agent loop immediately when called."},
)
```

**`trl/trainer/grpo_trainer.py`** — in `_tool_call_loop`, after the tool execution block and after the overlong filter, before calling `_generate_single_turn`:
```python
# Remove samples that called a stop tool — finalize their state without further generation
if self.args.stop_tool_names and stop_tool_global_idxs:
    non_stop_mask = [i not in stop_tool_global_idxs for i in idxs_with_tool]
    for local_idx, idx_with_tool in enumerate(idxs_with_tool):
        if idx_with_tool in stop_tool_global_idxs:
            prompt_length = len(prompt_ids[idx_with_tool])
            pct = prompt_completion_tool_ids[local_idx]
            old_completion_length = len(completion_ids[idx_with_tool])
            tool_suffix_length = len(pct) - prompt_length - old_completion_length
            completion_ids[idx_with_tool] = pct[prompt_length:]
            tool_mask[idx_with_tool] += [0] * tool_suffix_length
            if logprobs is not None:
                logprobs[idx_with_tool] += [0.0] * tool_suffix_length
    idxs_with_tool = [i for i, keep in zip(idxs_with_tool, non_stop_mask) if keep]
    prompt_completion_tools = [p for p, keep in zip(prompt_completion_tools, non_stop_mask) if keep]
    prompt_completion_tool_ids = [p for p, keep in zip(prompt_completion_tool_ids, non_stop_mask) if keep]
    if not idxs_with_tool:
        break
```

I'm happy to submit a PR with tests if the approach looks good. Related: #2723, discussion #2704.

lailanelkoussy:
# GRPOTrainer: add `stop_tool_names` for immediate agent loop termination

Fixes #5389

## What this PR does

Adds a `stop_tool_names: list[str] | None` field to `GRPOConfig`. When the agent calls any tool whose name is in this list, that sample exits the `_tool_call_loop` immediately after the tool result is appended — no further model generation occurs for it. Other samples in the batch continue unaffected.

This enables `final_answer`-style termination tools (as used in smolagents and similar frameworks) to work identically during training and evaluation, eliminating a systematic behavioural divergence.

### `trl/trainer/grpo_config.py`
- Added `stop_tool_names: list[str] | None` field with docstring

### `trl/trainer/grpo_trainer.py`
- In `_tool_call_loop`: after tool execution and after the overlong filter, identify samples that called a stop tool, update their `completion_ids` / `tool_mask` / `logprobs` to include the tool result tokens (masked from loss), then remove them from `idxs_with_tool` before `_generate_single_turn`

## Example

```python
def final_answer(answer: int) -> int:
    """Submit the final answer to the task.
    Args:
        answer: The final integer result.
    Returns:
        The submitted answer.
    """
    return answer

config = GRPOConfig(
    max_tool_calling_iterations=15,
    stop_tool_names=["final_answer"],
    ...
)

trainer = GRPOTrainer(
    args=config,
    tools=[tool_a, tool_b, final_answer],
    ...
)
```

## Behaviour

- `stop_tool_names=None` (default): no change from current behaviour.
- When a stop tool is called: the tool result is appended to the conversation (role `"tool"`), those tokens are masked in `tool_mask` (value 0, excluded from loss), and no new completion is generated for that sample.
- If all samples in a batch call a stop tool in the same iteration, the loop exits early via the existing `if not idxs_with_tool: break` path.
- Interaction with `max_tool_calling_iterations`: stop-tool termination takes precedence; the sample exits regardless of how many iterations remain.
- Interaction with overlong filter: a sample that is both overlong and calls a stop tool is handled by the existing overlong path first (it is already removed from `idxs_with_tool`), so the stop-tool path only sees non-overlong samples.

## Testing

```bash
pytest tests/test_grpo_trainer_stop_tool.py -sv
```

The test verifies:
1. A model that calls `final_answer` exits the loop after 1 tool-calling iteration (not `max_tool_calling_iterations`).
2. The completion contains the `final_answer` tool call + result, with no post-tool model tokens.
3. `tool_mask` correctly marks the tool result tokens as 0.
4. With `stop_tool_names=None`, behaviour is unchanged.

---

> [!NOTE]
> **Medium Risk**
> Touches GRPO’s tool-calling generation loop and modifies how `completion_ids`/`tool_mask`/`logprobs` are built and truncated, which can affect training dynamics and memory usage. Behavior is gated behind `stop_tool_names` (default `None`), but incorrect masking/truncation could impact loss/reward calculations when enabled.
> 
> **Overview**
> Adds a new `GRPOConfig.stop_tool_names` option to treat specific tools (e.g. `final_answer`) as *terminal* during GRPO agent training, so samples stop generating immediately after that tool executes.
> 
> Updates `GRPOTrainer._tool_call_loop` to detect stop-tool calls, append the tool-result suffix into `completion_ids`, mask those tool-result tokens out of the loss (`tool_mask=0`), truncate consistently to `max_completion_length` (including `logprobs` alignment), and remove those samples from further post-tool `generate` turns while letting other samples continue.
> 
> Introduces a comprehensive test suite (`tests/test_grpo_trainer_stop_tool.py`) that mocks `model.generate` to assert the reduced generation call count, mixed-batch behavior, reward-visible tool messages, finite loss, and truncation/mask/logprobs alignment, with `xfail` on older `transformers` without tool parsing.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4349c67dfb826c4c310567bbc47c8c8fc45ae1cf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


qgallouedec:
@codex can this be simplified?

qgallouedec:
@albertvillanova I don't know how common is this stop_tool feature, do you know if it's something specific to smolagent or more widely used? if so I'd be happy to add this feature

lailanelkoussy:
@qgallouedec I was also considering proposing a PR to support varying the available tools per rollout (i.e., per datapoint / sampled trajectory used for training). I’m currently fine-tuning a model with tools and validating checkpoints using smolagents, and I think it would benefit from exposing the model to more diverse toolsets across rollouts.
If you think this could be combined with this PR in a clean way, I’d be happy to work on it.

qgallouedec:
Yes I think we should find a way to make the tools vary. I'm sure how yet, maybe with a column in the dataset? But in any case this is a separate concern and should not be combined with this PR

lailanelkoussy:
Makes sense, thanks! I’ll open a separate PR for this.
My current idea is to add a `tools` column to the dataset, where each datapoint specifies the subset of tools (by name) available for that rollout, while keeping `GRPOTrainer(tools=[...])` as the global pool and filtering it per rollout.

qgallouedec:
yep, that would be consistent with SFT/DPO etc, ok please let's continue this discussion in another issue or pr 👍

albertvillanova (commented):
Thanks @lailanelkoussy, and sorry for the long silence on the PR itself — we're catching up on the backlog. 

We answered the design question over on #5389 back in June: the tool-call loop is already an intricate piece of code, and I'm still not convinced termination belongs at the tool level rather than at the response/loop level, so I don't want to lock in an API on a single request. Six weeks after I asked for thumbs-up to gauge demand the thread is still at zero, and the branch has since gone stale against main, so I'm going to close this one. #5389 stays open as the signal thread — if demand shows up there I'll come back to it and we can revive this branch. For whenever that happens: `distillation_trainer.py` carries its own copy of `_tool_call_loop`, so the change would need to land there too.

Issue #4358, opened 2025-10-28
GRPOConfig.max_prompt_length does not truncate prompts for a conversational dataset with vllm (since v0.25.0)


muupan:
### Reproduction

In the latest main branch, specifying `GRPOConfig.max_prompt_length` stopped truncating prompts when vLLM is used with a conversational dataset. As a consequence, it leads to an error when some prompt is larger than `max_model_len`. I think it is a bug newly introduced in the main branch. v0.24.0 seems ok.

## How to reproduce:

The script below is based on the example code in https://huggingface.co/docs/trl/main/grpo_trainer#quick-start, with additional options:
> use_vllm=True, vllm_mode="colocate", max_prompt_length=10, max_completion_length=10

script
```python
from datasets import load_dataset
from trl import GRPOConfig, GRPOTrainer

dataset = load_dataset("trl-lib/ultrafeedback-prompt", split="train")

# Dummy reward function for demonstration purposes
def reward_num_unique_letters(completions, **kwargs):
    """Reward function that rewards completions with more unique letters."""
    completion_contents = [completion[0]["content"] for completion in completions]
    return [float(len(set(content))) for content in completion_contents]

training_args = GRPOConfig(output_dir="Qwen2-0.5B-GRPO", use_vllm=True, vllm_mode="colocate", max_prompt_length=10, max_completion_length=10)
trainer = GRPOTrainer(
    model="Qwen/Qwen2-0.5B-Instruct",
    reward_funcs=reward_num_unique_letters,
    args=training_args,
    train_dataset=dataset,
)
trainer.train()
```

dependencies
```
uv venv && uv pip install "torch==2.8.0" vllm==0.11.0 torchvision git+https://github.com/huggingface/trl.git@5e691d1 deepspeed --torch-backend cu126
```

command
```
CUDA_VISIBLE_DEVICES=0 uv run accelerate launch debug.py
```

output
```
/mnt/shared/fujita/debug_trl_truncate_prompt_tokens/.venv/lib/python3.12/site-packages/trl/import_utils.py:91: UserWarning: TRL currently only supports vLLM version `0.10.2`. You have version 0.11.0 installed. We recommend to install this version to avoid compatibility issues.
  warnings.warn(
INFO 10-28 22:05:01 [__init__.py:216] Automatically detected platform cuda.
[...]

PR #3919, opened 2025-08-19
♻️ Reuse multimodal message preparation from `SFTTrainer` in `GRPOTrainer`

MengAiDev:
- Import DataCollatorForVisionLanguageModeling from sft_trainer
- Replace manual multimodal message preparation with DataCollatorForVisionLanguageModeling.prepare_multimodal_messages

Fixes #3915

PR #5512, opened 2026-04-10
Expose trainer dataset type metadata

### LINKED ISSUE #5511: Expose trainer dataset type metadata
## Feature request

TRL already documents the expected dataset type for each trainer in `docs/source/dataset_formats.md`, but that information is not exposed programmatically.

Today, downstream wrappers and validation layers have to hardcode trainer-name checks like `if trainer_class is GRPOTrainer` or maintain their own registry to figure out whether a trainer expects `prompt-only`, `preference`, `prompt-completion`, etc.

A small metadata API on trainer classes would solve this cleanly:

```python
from trl import GRPOTrainer

GRPOTrainer.dataset_types == ("prompt-only",)
```

For trainers that support more than one type:

```python
from trl import SFTTrainer

SFTTrainer.dataset_types == ("language-modeling", "prompt-completion")
```

## Motivation

This would let wrappers, CLIs, and downstream libraries:
- validate datasets before trainer construction
- choose preprocessing paths without hardcoded class checks
- stay aligned with the dataset-type table already maintained in the docs

## Why now

There seems to be recurring user and maintainer effort around trainer-specific dataset semantics and chat-template behavior, for example:
- #3468
- #3915
- #3919
- #4147
- #4201
- #4358

Those issues are not asking for this exact API, but they do show that trainer dataset semantics matter in practice and are currently discovered indirectly through docs or implementation details.

## Proposed shape

Add a class attribute on trainers:

```python
class _BaseTrainer(Trainer):
    dataset_types: tuple[str, ...] = ()
```

Then set it on the trainers already covered by the dataset-format table.

This would be metadata only, with no behavior change.

JohnGiorgi:
# What does this PR do?

Fixes #5511

TRL already documents the expected dataset type for each trainer, but that information is not exposed programmatically. Downstream wrappers and validation layers currently have to hardcode trainer-name checks or maintain their own registry to decide whether a trainer expects `prompt-only`, `preference`, `prompt-completion`, and so on.

This PR exposes that already-documented information as trainer metadata via a `dataset_types` class attribute.

Related context:
- #3468
- #3915
- #3919
- #4147
- #4201
- #4358

The implementation is metadata only:
- define `dataset_types` on `_BaseTrainer`
- populate it on the trainers already covered by the dataset-format table
- add a focused test covering those mappings
- add a short documentation note showing how to access the metadata programmatically

## Before submitting

## AI writing disclosure

## Who can review?

### DIFF
--- a/docs/source/dataset_formats.md
+++ b/docs/source/dataset_formats.md
@@ -422,6 +422,8 @@ Choosing the right dataset type depends on the task you are working on and the s
 | [`experimental.prm.PRMTrainer`] | [Stepwise supervision](#stepwise-supervision) |
 | [`experimental.xpo.XPOTrainer`] | [Prompt-only](#prompt-only) |
 
+The same metadata is also exposed programmatically on each trainer class through the `dataset_types` attribute. For example, `GRPOTrainer.dataset_types == ("prompt-only",)`.
+
 ## Using any dataset with TRL: preprocessing and conversion
 
 Many datasets come in formats tailored to specific tasks, which might not be directly compatible with TRL. To use such datasets with TRL, you may need to preprocess and convert them into the required format.
--- a/tests/test_trainer_metadata.py
+++ b/tests/test_trainer_metadata.py
@@ -1,0 +1,44 @@
+# Copyright 2020-2026 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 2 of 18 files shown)


qgallouedec:
Thanks for the PR @JohnGiorgi, and I do understand the motivation.

That said, I'm not fully convinced this is solving a real problem, downstream wrappers can maintain their own registry, which isn't a big deal. And since this attribute wouldn't be used internally by trl itself, I'm a bit hesitant to add it to the public surface.

@albertvillanova what do you think?

JohnGiorgi:
> Thanks for the PR @JohnGiorgi, and I do understand the motivation.
> 
> That said, I'm not fully convinced this is solving a real problem, downstream wrappers can maintain their own registry, which isn't a big deal. And since this attribute wouldn't be used internally by trl itself, I'm a bit hesitant to add it to the public surface.
> 
> @albertvillanova what do you think?

Fair enough, no hard feelings if you close it out. Point taken about TRL not using this internally... but I was thinking of it more like exposing a contract that right now only seems to exist in the docs ([Which dataset type to use?](https://huggingface.co/docs/trl/v1.0.0/en/dataset_formats#which-dataset-type-to-use)). We are maintaining code to automatically process a dataset in the right format for a given trainer, and that routing is just based on what we found in the docs which felt a little flimsy.

PR #5517, opened 2026-04-10
Fix `supports_tool_calling` falsely accepting templates that drop assistant `tool_calls`

qgallouedec:
`supports_tool_calling` only checked that a tool *message*'s content survived rendering. Templates that render tool messages correctly but silently drop the assistant's `tool_calls` (notably the basic Llama 3 template, which only reads `message['content']`) were still classified as supporting tool calling:

```python
>>> tok = AutoTokenizer.from_pretrained("trl-internal-testing/tiny-LlamaForCausalLM-3")
>>> tok.apply_chat_template(
...     [
...         {"role": "user", "content": "What is 3*4?"},
...         {
...             "role": "assistant",
...             "content": "",
...             "tool_calls": [{"type": "function", "function": {"name": "multiply", "arguments": {"a": 3, "b": 4}}}],
...         },
...     ],
...     tokenize=False,
... )
'<|begin_of_text|><|start_header_id|>user<|end_header_id|>\n\nWhat is 3*4?<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n<|eot_id|>'
```

The `multiply(a=3, b=4)` call vanished: the model has no way to emit a tool call with this template. Yet `supports_tool_calling` returned `True`.

## Fix

Plant distinct sentinels in the function name, argument key, and argument value (in addition to the existing tool-message-content sentinel) and require *all four* to survive rendering. A template now has to round-trip both halves of the exchange to be accepted.

---

> [!NOTE]
> **Medium Risk**
> Tightens `supports_tool_calling` detection, which may flip capability checks for some models and change downstream code paths that rely on this classification. Logic is simple but affects model/template compatibility gating.
> 
> **Overview**
> Fixes `supports_tool_calling` to **reject chat templates that silently drop assistant `tool_calls`**, not just tool-message content. The check now injects distinct sentinels for tool call name/arguments and tool response content and requires all of them to appear in the rendered prompt.
> 
> Updates tests to reflect the corrected behavior: `tiny-LlamaForCausalLM-3` is now expected to *not* support tool calling, while newer Llama 3.1/3.2 templates are added to the supported set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f6a5bb685f978b74e36a0b551fe1930532c4fb8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_chat_template_utils.py
+++ b/tests/test_chat_template_utils.py
@@ -160,7 +160,8 @@ class TestSupportsToolCalling:
                 ),
             ),
             pytest.param("trl-internal-testing/tiny-GptOssForCausalLM", id="gptoss"),
-            pytest.param("trl-internal-testing/tiny-LlamaForCausalLM-3", id="llama3"),
+            pytest.param("trl-internal-testing/tiny-LlamaForCausalLM-3.1", id="llama3.1"),
+            pytest.param("trl-internal-testing/tiny-LlamaForCausalLM-3.2", id="llama3.2"),
             pytest.param("trl-internal-testing/tiny-Qwen2ForCausalLM-2.5", id="qwen2.5"),
             pytest.param("trl-internal-testing/tiny-Qwen3ForCausalLM", id="qwen3"),
             pytest.param("trl-internal-testing/tiny-Qwen3MoeForCausalLM", id="qwen3moe"),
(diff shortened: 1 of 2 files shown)


AmineDiro (approved):
I wonder if the processing class shouldn't have queryable method to know what part of key/vals will be rendered, this would be easier to check without defensively running

qgallouedec:
> I wonder if the processing class shouldn't have queryable method to know what part of key/vals will be rendered, this would be easier to check without defensively running

It’s mandatory to at least render the tool name, the keys, and the values. The content, however, may vary (I don’t have a concrete example, but I know some models like Gemma4 are trained to omit content when there’s a tool call, even if the chat template allow both to coexist).

Actually, writing this out makes me realize that this "content" check is probably unnecessary, so I’m inclined to remove it. Adding a queryable method for this feels like overkill for now a simple heuristic like this is likely sufficient.

qgallouedec:
Ah no, I’m talking nonsense, we’re checking the content of the tool result, not the content from the tool call. We’re good.

PR #5323, opened 2026-03-20
Support multimodal tool responses in `environment_factory` for VLM training

sergiopaniego:
# What does this PR do?

Enable `environment_factory` tool methods to return multimodal content blocks (images + text) for VLM training.
Previously, tool responses were always converted to `str(result)`, discarding any visual information. Now tools can
return content block lists that include images, and the trainer handles them end-to-end: tokenization, generation,
and training forward pass with proper `pixel_values`.

Also migrates `browsergym.py` (VLM) from `rollout_func` to `environment_factory` with multimodal tool responses.

**Depends on:** #5235 (OpenEnv examples update) and related to https://github.com/huggingface/transformers/pull/44873

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

---

> [!NOTE]
> **Medium Risk**
> Touches core `GRPOTrainer` generation/tool-calling and multimodal forward-pass plumbing, which could affect VLM training stability and sequence-length edge cases across multiple trainers.
> 
> **Overview**
> Adds end-to-end support for *multimodal tool responses* (e.g., image + text content blocks) when using `GRPOTrainer` with `environment_factory`, so tool outputs are no longer stringified and can feed VLMs with `pixel_values` during tool-call loops.
> 
> Updates `GRPOTrainer` internals to track/merge tool-returned images, tokenize tool messages correctly for VLM processors, avoid truncating mid-image token spans, and add safer `mm_token_type_ids`/max-length handling for VLM configs; similar image-handling fixes are applied in `GFPOTrainer`, `GRPOWithReplayBufferTrainer`, and `RLOOTrainer`. Also migrates the OpenEnv `browsergym.py` example to `environment_factory` with screenshot-returning tool methods, adds a new `carla_vlm.py` OpenEnv VLM example, and documents the new example in `example_overview.md`.
> 
[...]

PR #5521, opened 2026-04-11
Drop, don't truncate, overlong tool results in GRPOTrainer

qgallouedec:
Before #5323, GRPO truncated overlong tool results with a plain slice, quite easy to do.

#5323 added support for tools returning images, and we kept truncation, but now it has to be image-token aware: cutting inside a `<|vision_start|>...<|vision_end|>` span desyncs `input_ids` from `pixel_values` and the forward pass errors or silently corrupts embeddings.

Keeping truncation cost us a `_truncate_at_image_boundary` helper, a `_get_vision_token_ids` cache with hardcoded per-family token lists, two post-loop sync blocks, and a two-stage fallback trim. All fragile, all annoying to maintain across VLM families (same experience we had trying to support prompt truncation for VLMs).

## Proposal

Drop overlong tool results instead of truncating them. If `len(completion + tool_suffix) > max_completion_length`, the sample exits the loop with its completion as-is, and the tool messages/images appended this iteration are rolled back via a snapshot taken at the top of the while loop.

**Why the learning signal is probably the same.** The objective is to reach a correct final answer, and a sequence that ran out of budget mid-tool-call rarely gets there — truncated or dropped, it's almost certainly going to be rewarded as a failure. Either way the sample contributes roughly the same signal to the group, so "truncate" vs "drop" is a wash in practice, and drop is dramatically simpler.

## Result

Truncation path goes from ~80 lines of image-aware bookkeeping to ~15 lines of check-and-rollback. Deleted: `_truncate_at_image_boundary`, `vision_start`/`vision_end` lookup, two post-loop sync blocks, post-regen fallback trim.

## Caveats

- `max_model_len` guard removed. Usually `max_completion_length` is the tighter bound; we can add it back if unusual configs hit it.
- Dropped samples end on dangling `tool_calls` (no response). Reward functions should be fine with this.

---

> [!NOTE]
> **Medium Risk**
> Changes tool-calling generation behavior by rolling back (dropping) tool outputs that exceed completion/context budgets, which can affect training trajectories and reward evaluation. Also tweaks VLM token-type reconstruction for tool-injected images, so multimodal edge cases need verification.
> 
> **Overview**
> **Tool-calling in `GRPOTrainer` now *drops* overlong tool results instead of truncating them.** The tool loop snapshots `prompts`/`completions`/`tool_images` each iteration and rolls back any tool messages/images added when the resulting `prompt+completion+tool_suffix` would exceed either `max_completion_length` (completion budget) or the model context limit.
> 
> This removes the previous image-boundary-aware truncation path and post-loop sync bookkeeping, and simplifies post-tool regeneration to only slice model-generated `post_tool_ids` when the completion budget is exceeded.
> 
> For VLM tool responses that inject images into completions, the trainer now resolves image/video placeholder token IDs once at init (e.g. `<|image_pad|>`/`<|image|>`, `<|video_pad|>`) and uses them to rebuild `token_type_ids`/`mm_token_type_ids` during the forward pass.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c78ca99bed0fa92d6032e3d29f442076b7b2bab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
@codex review

qgallouedec:
@codex review

qgallouedec:
@sergiopaniego this would be super cool if you could check if these changes don't break the examples that you created 🙏

albertvillanova (approved):
GREAT! Thanks.

sergiopaniego (approved):
much simpler! thanks
I'm testing the current approach on the examples and will let you know if I find any issue

PR #5544, opened 2026-04-14
Deprecate `use_transformers_paged`

qgallouedec:
Deprecate the `use_transformers_paged` config flag. Setting it to `True` now raises a `FutureWarning`; behavior is unchanged otherwise.

Affected configs: `GRPOConfig`, `RLOOConfig`, and experimental `OnlineDPOConfig`, `GOLDConfig`, `SelfDistillationConfig`.

## Motivation

- **No usage**: the feature is not used in practice inside or outside the repo.
- **Actually worse, not faster**: the original motivation for `use_transformers_paged` was faster generation, but in a small A/B benchmark (Qwen3-0.6B, GRPO, `ultrafeedback-prompt`, `max_completion_length=256`) the paged path is ~20% *slower* than the default padded path (1428 s vs 1189 s) and consumes ~6× more peak VRAM (58.2 GiB vs 9.6 GiB). See table below.
- **Superseded soon**: `transformers` continuous batching support is landing shortly and will replace this path entirely, so maintaining the paged flag in the meantime is dead weight.
- **Maintenance cost**: the paged branch adds a third generation path in each trainer (`vLLM` / `paged` / default) that has to stay aligned across duplicated trainer code.
- **Not on par with the other generation paths**: it seems to be some inconsistency in the way the paged path handles generation resulting in higher grad norm and entropy, not sure exactly why, and if we would support this path long term we would want to investigate and fix this.

Screenshot 2026-04-14 at 11 29 43 AM

## Benchmark

```python

import argparse
import time

import torch

def reward_len(completion_ids, **kwargs):
    return [-abs(20 - len(c)) for c in completion_ids]


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--paged", type=int, default=0)
    args = parser.parse_args()

    paged = bool(args.paged)
    run_name = f"grpo-{'paged' if paged else 'nopaged'}_4"

    dataset = load_dataset("trl-lib/ultrafeedback-prompt", split="train[:64]")

    config = GRPOConfig(
        output_dir=f"/tmp/{run_name}",
        run_name=run_name,
        num_generations=4,
        max_completion_length=256,
        logging_steps=1,
        report_to="trackio",
        use_transformers_paged=paged,
    )

    trainer = GRPOTrainer(
        model="Qwen/Qwen3-0.6B",
        args=config,
        train_dataset=dataset,
        reward_funcs=reward_len,
    )

    t0 = time.perf_counter()
    trainer.train()
    elapsed = time.perf_counter() - t0

    print(f"\n===== {run_name} =====")
    print(f"elapsed:   {elapsed:.1f} s")
    print(f"peak VRAM: {peak_gb:.2f} GiB")
    print(f"samples/s: {trainer.state.log_history[-1].get('train_samples_per_second', 'n/a')}")


if __name__ == "__main__":
    main()
```

|           | No paged | Paged     |
| --------- | -------- | --------- |
| elapsed   | 1188.9 s | 1428.1 s  |
| samples/s | 0.162    | 0.134     |
| peak VRAM | 9.60 GiB | 58.17 GiB |

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it deletes a previously supported generation backend/CLI flag in experimental trainers, which can break workflows that relied on paged attention despite being rarely used. Core trainers keep the flag but now emit deprecation warnings, which may affect noisy logs/tests.
> 
> **Overview**
> This PR **phases out `use_transformers_paged`** by removing the paged-attention generation branch from experimental trainers and configuration surfaces.
> 
> `GOLDTrainer.generate_on_policy_outputs` and `OnlineDPOTrainer` now always use the standard `transformers` `generate()` path when not using vLLM, and associated tests asserting paged behavior are deleted. The `use_transformers_paged` fields/docs are removed from experimental configs (`GOLDConfig`, `OnlineDPOConfig`, `SelfDistillationConfig`), and legacy validation around paged support is dropped in `DPPOConfig`.
> 
> In core configs (`GRPOConfig`, `RLOOConfig`), `use_transformers_paged` is retained but marked **deprecated**: enabling it triggers a `FutureWarning` and the docstrings explicitly note planned removal in v2.0.0.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 909a1ec36402d6caab9b029e2910c5850f38a306. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/experimental/test_gold_trainer.py
+++ b/tests/experimental/test_gold_trainer.py
@@ -550,7 +550,6 @@ def test_gold_trainer_init_defaults_vllm_max_model_length_to_max_length(monkeypa
         top_p=1.0,
         seq_kd=False,
         num_generations=1,
-        use_transformers_paged=False,
         max_completion_length=16,
         top_k=0,
         log_completions=False,
@@ -666,7 +665,6 @@ def test_get_start_and_size_answers_skips_prompt_tokens():
 @pytest.mark.slow
 def test_generate_on_policy_outputs_masks_prompt(llama_tokenizer):
     trainer = GOLDTrainer.__new__(GOLDTrainer)
-    trainer.use_transformers_paged = False
     trainer.processing_class = llama_tokenizer
 
     prompt_text = "<|begin_of_text|><|start_header_id|>user<|end_header_id|>\nHello?<|eot_id|>"
@@ -718,7 +716,6 @@ def test_generate_on_policy_outputs_masks_prompt(llama_tokenizer):
 @pytest.mark.slow
 def test_generate_on_policy_outputs_masks_prompt_smollm(smollm_tokenizer, openr1_examples):
     trainer = GOLDTrainer.__new__(GOLDTrainer)
-    trainer.use_transformers_paged = False
     trainer.processing_class = smollm_tokenizer
 
     collator = DataCollatorForChatML(tokenizer=smollm_tokenizer)
--- a/tests/experimental/test_online_dpo_trainer.py
+++ b/tests/experimental/test_online_dpo_trainer.py
@@ -318,7 +318,6 @@ class TestOnlineDPOTrainer(TrlTestCase):
         assert config.top_k == 0
         assert config.min_p is None
         assert config.repetition_penalty == 1.0
-        assert not config.use_transformers_paged
         assert config.cache_implementation is None
         assert config.generation_kwargs is None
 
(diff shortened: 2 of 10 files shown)


albertvillanova (approved):
Thanks!

AmineDiro (approved):
Should we update the documentation also ?

qgallouedec:
> Should we update the documentation also ?

Documentation should be up to date 👍

---

I just remove paged from experimental (online dpo, gold etc) instead of deprecating it.

PR #5561, opened 2026-04-15
Revert VLM support in `parse_response`

qgallouedec:
`parse_response` previously accepted either a tokenizer or a processor (from #5323) and unwrapped the inner tokenizer on the fly. Now that call sites can easily pass the tokenizer directly, we move that disambiguation to the call sites and keep `parse_response` strictly tokenizer-only. This centralizes the "processor vs tokenizer" logic in one place per trainer and makes `parse_response`'s contract simpler.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes `parse_response`’s public API by no longer accepting a VLM processor, which may break downstream callers that pass processors instead of tokenizers.
> 
> **Overview**
> Simplifies `parse_response`’s contract by making it accept only a `PreTrainedTokenizerBase` (removing the internal “processor vs tokenizer” unwrapping).
> 
> Updates `TestParseResponse` to explicitly pass `processing_class.tokenizer` for VLM processors so response parsing continues to work for multimodal models.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cdc6c9084f4fca5d80ad8c9b30fc9594176d9aba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova (commented):
I'm not sure of this PR, as it seems to overlap quite a bit with what I'm currently addressing in #5489. The approach I’m taking there is intentionally incremental:
- First, ensure that we can consistently pass only tokenizer instances to `parse_response` (by introducing `self._tokenizer` across trainers).
- Then, in a follow-up step, simplify parse_response to only accept tokenizers:  [Make parse_response accept only tokenizer](https://github.com/huggingface/trl/pull/5489/commits/2242686ead877dc8ec6b9400c55f5c50f5fed1ae)

Given that, this PR feels somewhat like duplicated effort. Would it make sense to wait for #5489 to land instead?

For context, I’m already working through the relevant discussion here: https://github.com/huggingface/trl/pull/5489#discussion_r3084804195
> `parse_response` only needs a tokenizer instance but it had to handle both because we did not have a simple way to pass only tokenizer. Once we implement `self._tokenizer` in all trainers, `parse_response` could be simplified to accept only tokenizer instances.

and here: https://github.com/huggingface/trl/pull/5489#discussion_r3084921980
> More broadly, the underlying goal of this PR is to centralize the processor/tokenizer disambiguation within processing_class in a single place, so that the rest of the code can rely on a well-defined and consistent interface, with a clear expected class instance.
>
> In that sense, the current change in calling `parse_response` is an intermediate step toward that simplification, rather than a deviation from it.

qgallouedec:
Ah yes ok, Lgtm @albertvillanova

albertvillanova (approved):
Thanks.

PR #5572, opened 2026-04-16
Fix empty-target self-distillation loss to stay connected to model graph

walawalagoose:
# What does this PR do?

This PR fixes an edge case in the experimental SDPO/self-distillation path where an empty self-distillation batch returns a standalone zero tensor instead of a zero loss connected to the model forward graph.

In `distillation_only` mode, a batch can legitimately contain no valid self-distillation targets, for example when no rollout in the current batch exceeds `success_reward_threshold` and no environment feedback is used. In the current implementation, this path returns:

```python
torch.tensor(0.0, device=completion_ids.device, requires_grad=True)
```

Although this tensor has `requires_grad=True`, it is not connected to model parameters. Under DeepSpeed ZeRO-2, calling `backward()` on such a graph-disconnected loss can fail during gradient reduction.

This PR fixes the issue by keeping the self-distillation loss connected to the student/teacher forward graph even when the effective target mask is empty, so the loss remains numerically zero but backward stays safe.

This behavior is also closer to the original verl SDPO implementation, where an all-zero mask yields a zero loss through masking rather than through an early return.

A representative failure mode is:

- `SDPOTrainer`
- `sdpo_policy_loss_mode="distillation_only"`
- `include_environment_feedback=False`
- current batch has no successful rollouts
- DeepSpeed ZeRO-2 enabled

In that setup, DeepSpeed can fail during backward/all-reduce because the returned zero loss is graph-disconnected.

## Before submitting

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **Medium Risk**
> Changes loss computation behavior in the experimental self-distillation path by always running model forwards even when the target mask is empty, which can affect training performance/metrics but should improve backward stability (e.g., under ZeRO).
> 
> **Overview**
> Fixes an edge case in `SelfDistillationMixin._compute_self_distillation_loss` where batches with an all-zero `response_mask` previously returned a standalone `torch.tensor(0.0)`.
> 
> The loss is now always computed via the normal masked aggregation so it stays connected to the student/teacher forward graph for safe `backward()`/reduction, and a new `self_distillation/empty_target_batch` metric is logged to track when this happens.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee61598d7bca01f0070944154fbb6703d34f42b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:
Thanks for the fix @walawalagoose, and sorry for the slow turnaround. 

Since this PR was opened, the experimental self-distillation code was refactored in #5862 (merged Jun 2): `self_distillation_mixin.py` no longer exists, and the logic now lives in `trl/experimental/sdpo/sdpo_trainer.py`. That refactor already covers exactly the case you describe — the empty-batch path now returns `distillation_logits.student_logits.sum() * 0.0`, a zero loss that stays attached to the student graph, so backward() is safe under DeepSpeed ZeRO. It also keeps the early-return short-circuit, so it avoids the extra full student+teacher forward that removing the guard (as this PR does) would incur on every empty batch. 

Since the target file is gone and the bug is fixed on main, I think the cleanest path is to close this one. If you'd still like the extra `empty_target_batch` counter metric, that's a trivial one-liner we'd happily take as a fresh PR against the new location — thanks again!

PR #5575, opened 2026-04-17
Chunked cross-entropy loss for SFT (up to –50% VRAM)

qgallouedec:
# What does this PR do?
 
sft_memory (1)

## Choosing the right chunk size -> 256

We care more about peak memory than wall time (the whole point of chunked CE is to fit workloads that wouldn't otherwise fit). So IMO "optimal" = the smallest `chunk_size` whose time stays within 50% of the sweep minimum.

chunk_size_pareto

Fortunately, the sweet spot is **consistent across model size and sequence length**: `chunk_size = 256` wins in all 5 configs tested.

| shape                      | ref memory | knee memory      |     | ref time | knee time       |
| -------------------------- | ---------- | ---------------- | --- | -------- | --------------- |
| small model                | 8.29 GB    | 2.07 GB (-×4.0)  |     | 82 ms    | 151 ms (×1.84)  |
| small model, long sequence | 30.25 GB   | 2.14 GB (-×14.1) |     | 323 ms   | 604 ms (×1.87)  |
| medium model               | 9.70 GB    | 4.84 GB (-×2.0)  |     | 193 ms   | 337 ms (×1.74)  |
| big model                  | 12.04 GB   | 9.47 GB (-×1.3)  |     | 376 ms   | 658 ms (×1.75)  |
| big model, long sequence   | 34.29 GB   | 9.84 GB (-×3.5)  |     | 1456 ms  | 2642 ms (×1.81) |

Picking 512 or 1024 would trim wall time by only a few percent at the cost of meaningfully higher memory. Not worth it given the goal. Setting `_CHUNKED_LM_HEAD_CHUNK_SIZE = 256`.

<details>
<summary>Benchmarking script</summary>

eg:

```
python benchmark_chunked_nll_chunk_size.py \
    --hidden_size 1024 --n_valid 4096 --output /tmp/bench.json
```

```python
# benchmark_chunked_nll_chunk_size.py
"""
Micro-benchmark `_chunked_cross_entropy_loss` across chunk sizes.

Runs forward + backward on synthetic `(hidden, weight, labels)` tensors and reports peak GPU
memory + wall time. The un-chunked baseline (full `[N_valid, V]` logits) is also timed as an
upper bound.
"""

import argparse
import json
import time

import torch

from trl.trainer.sft_trainer import _chunked_cross_entropy_loss


VOCAB_SIZE = 151936  # Qwen2.5 / Qwen3 family
CHUNK_SIZES = [64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384]
ITERS = 10
DTYPE = torch.bfloat16


def make_inputs(n_valid, hidden_size):
    # Shift means `n_valid + 1` tokens yield exactly `n_valid` valid (non-ignored) positions.
    hidden = torch.randn(1, n_valid + 1, hidden_size, device="cuda", dtype=DTYPE, requires_grad=True)
    weight = torch.randn(VOCAB_SIZE, hidden_size, device="cuda", dtype=DTYPE, requires_grad=True)
    labels = torch.randint(0, VOCAB_SIZE, (1, n_valid + 1), device="cuda")
    return hidden, weight, labels


def bench(fn, n_valid, hidden_size):
    hidden, weight, labels = make_inputs(n_valid, hidden_size)
    try:
        fn(hidden, weight, labels).backward()  # warmup
        torch.cuda.synchronize()
        start = time.perf_counter()
        for _ in range(ITERS):
            hidden.grad = weight.grad = None
            fn(hidden, weight, labels).backward()
        torch.cuda.synchronize()
        return torch.cuda.max_memory_allocated() / 1e6, (time.perf_counter() - start) * 1000 / ITERS
    except torch.cuda.OutOfMemoryError:
        return None, None
    finally:
        del hidden, weight, labels
        torch.cuda.empty_cache()


def reference_loss(h, w, lbl):
    """What the standard nll path does: materialize full logits."""
    shift_h = h[..., :-1, :].reshape(-1, h.size(-1))
    shift_l = lbl[..., 1:].reshape(-1)
    return F.cross_entropy(shift_h.float() @ w.float().t(), shift_l, ignore_index=-100)


def chunked_loss(chunk_size):
    return lambda h, w, lbl: _chunked_cross_entropy_loss(h, w, lbl, chunk_size=chunk_size)[0]


def main():
    parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
    parser.add_argument("--hidden_size", type=int, required=True)
    parser.add_argument("--n_valid", type=int, required=True)
    parser.add_argument("--output", type=str, default=None)
    args = parser.parse_args()

    torch.manual_seed(0)
    print(f"N_valid={args.n_valid}, vocab={VOCAB_SIZE}, hidden={args.hidden_size}, dtype=bfloat16, iters={ITERS}")
    print(f"{'chunk_size':>12} {'peak_MB':>10} {'time_ms':>10}")

    ref = None
    peak, ms = bench(reference_loss, args.n_valid, args.hidden_size)
    if peak is None:
        print(f"{'reference':>12} {'OOM':>10} {'—':>10}")
    else:
        print(f"{'reference':>12} {peak:>10.1f} {ms:>10.2f}")
        ref = {"peak_mb": peak, "time_ms": ms}

    results = []
    for cs in CHUNK_SIZES:
        if cs > args.n_valid:
            continue
        peak, ms = bench(chunked_loss(cs), args.n_valid, args.hidden_size)
        if peak is None:
            print(f"{cs:>12} {'OOM':>10} {'—':>10}")
            results.append({"chunk_size": cs, "oom": True})
        else:
            print(f"{cs:>12} {peak:>10.1f} {ms:>10.2f}")
            results.append({"chunk_size": cs, "peak_mb": peak, "time_ms": ms})

    if args.output:
        shape = {"vocab_size": VOCAB_SIZE, "hidden_size": args.hidden_size, "n_valid": args.n_valid, "dtype": "bfloat16"}
        payload = {"shape": shape, "iters": ITERS, "reference": ref, "results": results}
        with open(args.output, "w") as f:
            json.dump(payload, f, indent=2)
        print(f"\nwrote {args.output}")


if __name__ == "__main__":
    main()
```

</details>

<details>
<summary>Plotting script</summary>

```python
# /// script
# dependencies = ["matplotlib", "numpy"]
# ///
import argparse
import json
import pathlib

import numpy as np
from matplotlib.lines import Line2D


def find_knee(chunk_sizes, time_ms, tol=0.50):
    """Smallest chunk_size whose time is within `tol` of the minimum (memory-priority)."""
    threshold = time_ms.min() * (1 + tol)
    for cs, t in zip(chunk_sizes, time_ms):
        if t <= threshold:
            return int(cs), float(t)
    return int(chunk_sizes[-1]), float(time_ms[-1])


def describe(shape, all_shapes):
    """Human-readable label with 'small/medium/big model' and 'long sequence' tags."""
    h, n = shape["hidden_size"], shape["n_valid"]
    hs = {s["hidden_size"] for s in all_shapes}
    ns = {s["n_valid"] for s in all_shapes}
    tags = []
    if len(hs) > 1:
        tags.append({max(hs): "big model", min(hs): "small model"}.get(h, "medium model"))
    if len(ns) > 1 and n == max(ns):
        tags.append("long sequence")
    return f"H={h}, N_valid={n}" + (f"  ({', '.join(tags)})" if tags else "")


def load_runs(dir_path):
    datas = []
    for p in sorted(pathlib.Path(dir_path).glob("*.json")):
        data = json.loads(p.read_text())
        rows = [r for r in data["results"] if not r.get("oom")]
        if rows:
            datas.append((data, rows))
    datas.sort(key=lambda d: (d[0]["shape"]["hidden_size"], d[0]["shape"]["n_valid"]))
    shapes = [d[0]["shape"] for d in datas]
    return [
        {
            "label": describe(data["shape"], shapes),
            "reference": data.get("reference"),
            "chunk_sizes": np.array([r["chunk_size"] for r in rows]),
            "peak_gb": np.array([r["peak_mb"] for r in rows]) / 1024,
            "time_ms": np.array([r["time_ms"] for r in rows]),
        }
        for data, rows in datas
    ]


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--dir", default="/tmp/chunk_size_bench")
    parser.add_argument("--out", default="chunk_size_pareto.png")
    args = parser.parse_args()

    runs = load_runs(args.dir)
    if not runs:
        raise SystemExit(f"No JSON files found in {args.dir}")

    colors = [plt.get_cmap("tab10")(i) for i in range(len(runs))]
    fig, ax = plt.subplots(figsize=(9.5, 7))
    fig.suptitle("Chunked CE: memory ↔ time trade-off", fontsize=14)

    for r, c in zip(runs, colors):
        ax.plot(r["time_ms"], r["peak_gb"], "-o", color=c, markersize=6, linewidth=2, label=r["label"])
        for cs, mem, t in zip(r["chunk_sizes"], r["peak_gb"], r["time_ms"]):
            ax.annotate(str(int(cs)), (t, mem), textcoords="offset points", xytext=(6, 4), fontsize=8, color=c)

        knee_cs, knee_t = find_knee(r["chunk_sizes"], r["time_ms"])
        knee_mem = r["peak_gb"][r["chunk_sizes"] == knee_cs][0]
        ax.scatter([knee_t], [knee_mem], color=c, marker="*", s=320, edgecolor="black", linewidth=1.2, zorder=6)

        if r["reference"] is not None:
            ref_t, ref_mem = r["reference"]["time_ms"], r["reference"]["peak_mb"] / 1024
            ax.scatter([ref_t], [ref_mem], color=c, marker="s", s=100, edgecolor="black", linewidth=1, zorder=5)
            ax.annotate("ref", (ref_t, ref_mem), textcoords="offset points", xytext=(8, -3),
                        fontsize=9, fontweight="bold", color=c)

    ax.set(xscale="log", yscale="log", xlabel="time per step (ms)", ylabel="peak memory (GB)")
    ax.set_yscale("log", base=2)
    ax.grid(True, which="both", alpha=0.3)

    shape_handles = [Line2D([0], [0], marker="o", linestyle="-", color=c, label=r["label"])
                     for r, c in zip(runs, colors)]
    glyph_handles = [
        Line2D([0], [0], marker="*", color="gray", linestyle="", markeredgecolor="black", markersize=14,
               label="knee (within 50% of min time)"),
        Line2D([0], [0], marker="s", color="gray", linestyle="", markeredgecolor="black", markersize=9,
               label="un-chunked reference"),
    ]
    ax.add_artist(ax.legend(handles=shape_handles, loc="upper left", fontsize=10, title="shape"))
    ax.legend(handles=glyph_handles, loc="upper right", fontsize=10, title="markers")

    print("\nknee per shape:")
    for r in runs:
        knee_cs, knee_t = find_knee(r["chunk_sizes"], r["time_ms"])
        knee_mem = r["peak_gb"][r["chunk_sizes"] == knee_cs][0]
        print(f"  {r['label']}  →  chunk_size = {knee_cs}  ({knee_mem:.2f} GB, {knee_t:.1f} ms)")

    plt.tight_layout(rect=[0, 0, 1, 0.96])
    plt.savefig(args.out, dpi=140, bbox_inches="tight")
    print(f"\nwrote {args.out}")


if __name__ == "__main__":
    main()
```
</details>

---

> [!NOTE]
> **Medium Risk**
> Adds a new `loss_type='chunked_nll'` path that monkey-patches `model.forward` and changes how loss/metrics are computed, which could affect training correctness or performance across model families and distributed configs. Guardrails and extensive numerical-equivalence tests reduce risk, but this touches the core SFT training loop.
> 
> **Overview**
> Adds a new SFT loss mode, `loss_type="chunked_nll"`, that reduces peak activation memory by avoiding full `[batch×seq×vocab]` logits: ignored-label tokens are dropped before the `lm_head` matmul and the remaining tokens’ cross-entropy is computed in checkpointed chunks (default chunk size `256`).
> 
> Implements this by patching the model’s `forward` to run the decoder directly and return loss + aggregated `num_correct_tokens`/`entropy_sum` (and MoE `aux_loss` when `output_router_logits=True`), while falling back to the original forward when labels aren’t provided to keep generation/eval behavior unchanged.
> 
> Updates `SFTConfig`/docs to document the new option and its constraints (not compatible with Liger, PEFT, or VLM; FSDP2 guidance), adds an FSDP2 performance warning, and introduces tests covering numerical equivalence (forward/backward), ignore-index edge cases, and end-to-end training with `chunked_nll`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d87972c1b8a33322160b7f0f7061af1bd475b41c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
## Numerical equivalence: `chunked_nll` vs `nll` under realistic SFT configs

End-to-end SFT training (bf16), two sequential runs from the same seed, one with `loss_type="nll"` and one with `loss_type="chunked_nll"`, comparing per-step losses over **100 steps**.

At 100 steps on Capybara + Qwen3-1.7B, the bf16 noise floor produces occasional outlier batches where the two paths' accumulated weight drift manifests as ~1e-1 loss spikes (the median step diff is ~2e-3; max is dominated by 1–2 outliers). Tolerance of **`5e-2`** catches the median behavior; the max excursion is ~1.2e-1 for baseline-like configs and is not specific to `chunked_nll`.

### Status

| #   | Config                                                                    | Tests                                                       | Status | max \|Δloss\| | Δmem   | Δtime  | Trackio                                                                                                                                                                                                |
| --- | ------------------------------------------------------------------------- | ----------------------------------------------------------- | ------ | ------------- | ------ | ------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| 1   | Baseline (`Qwen3-1.7B`)                                                   | tied embeddings, bf16, default settings                     | ✅ PASS | 1.22e-01      | -33.0% | -2.6%  | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=0902d984d937418893a4645617870822%2C90d05c9498d54daaa76e1de8294aca7b&sidebar=hidden&navbar=hidden&smoothing=0) |
| 2   | Gradient accumulation (`--gradient_accumulation_steps 8`)                 | `num_items_in_batch` plumbing across micro-batches          | ✅ PASS | 1.32e-02      | -33.4% | -1.4%  | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=13d8a7108bbc45608932bbe446cd155f%2Ce25a0cf6fc924ea2899238ef50553105&sidebar=hidden&navbar=hidden&smoothing=0) |
| 3   | Packing (`--packing`)                                                     | variable-length packed sequences                            | ✅ PASS | 3.22e-02      | -24.9% | +13.4% | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=dca4090adea4431cb8df19bfe0e31d62%2C65ed8910045140769d05eccc7fe41fcf&sidebar=hidden&navbar=hidden&smoothing=0) |
| 4   | No gradient checkpointing (`--no-gradient_checkpointing`)                 | chunked CE's per-chunk `checkpoint` without outer grad-ckpt | ✅ PASS | 1.22e-01      | -22.0% | -9.4%  | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=d77b88d1c40e49e1bf6a656e0320c4b6%2Cd7aef29b359f4949b149ec13ab48039c&sidebar=hidden&navbar=hidden&smoothing=0) |
| 5   | Flash attention 2 (`--attn_implementation kernels-community/flash-attn2`) | attention kernel orthogonality                              | ✅ PASS | 5.11e-02      | -33.0% | -9.5%  | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=ebe7868877b74e67b95ef30e11a6d916%2Cd1cd6968093543379102087231fd0ff5&sidebar=hidden&navbar=hidden&smoothing=0) |
| 6   | Long context (`--max_length 8192 --batch_size 1`)                         | `N_valid` >> `chunk_size`                                   | ✅ PASS | 1.33e-01      | +4.3%  | +6.2%  | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=5fbc716f60114ab8a2eee70f17bc0c0d%2C6c5a3fe41d4f4d7b9ad83f8431485ddd&sidebar=hidden&navbar=hidden&smoothing=0) |
| 7   | Untied embeddings (`microsoft/Phi-3-mini-4k-instruct`)                    | separate lm_head weight                                     | ✅ PASS | 8.21e-02      | -0.0%  | -0.0%  | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=9262fdbac92346fba5b32d655f688f3b%2C91e07aba455140799187d02b26560212&sidebar=hidden&navbar=hidden&smoothing=0) |
| 8   | DDP × 4 GPUs (`--gres=gpu:4`)                                             | `DistributedDataParallel` wrapping, gradient all-reduce     | ✅ PASS | 2.24e-02      | -28.5% | -23.2% | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=166f0f3ae53a4ef1819c29c1ba6f07f1%2Ce195cfb1c0bd4abb954bde3b7fb260a7&sidebar=hidden&navbar=hidden&smoothing=0) |
| 9   | FSDP2 × 4 GPUs                                                            | sharded params + grads + optimizer                          | ✅ PASS | 9.70e-03      | -53.1% | +49.7% | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=ffac47664e964667b4b5d7e4332378d8%2Cc1f490523b724c01bb638a6c29dde135&sidebar=hidden&navbar=hidden&smoothing=0) |
| 10  | FSDP2 × 4 GPUs, `--fsdp_reshard_after_forward false`                      | same as 9. but keeps the un-wrapped `lm_head` resident      | ✅ PASS | 7.80e-03      | -47.3% | -6.3%  | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=a412a4efd36c4bab8721930e9c7e26b8%2C440601d68f074481be55b4baaab28039&sidebar=hidden&navbar=hidden&smoothing=0) |
| 11  | fp32 (`--no-bf16`)                                                        | tight tolerance — catches any real correctness drift        | ✅ PASS | 9.50e-03      | -30.6% | -5.7%  | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=efbc60881f4f445d8c177296f812b06a%2C46c777937e1b443c967eb1427ed66fea&sidebar=hidden&navbar=hidden&smoothing=0) |
| 12  | ALST / Ulysses SP × 4 GPUs (`sp_size=2 × dp_shard_size=2`)                | `shift_labels` path under sequence parallelism              | ✅ PASS | 1.50e-01      | -0.1%  | -1.4%  | [link](https://qgallouedec-chunked-nll-benchmark-2.hf.space/?project=huggingface&run_ids=d8ce3d9e36364816a4369aca1d6ebbe0%2C81c7d863e79849d7a9a36e55d3b64989&sidebar=hidden&navbar=hidden&smoothing=0) |

**Note on row 9 (FSDP2 default):** the +49.7% time cost comes from chunked CE triggering a per-chunk all-gather of `lm_head.weight` during backward — with `chunk_size=256`, that's ~32 redundant gathers per step. Setting `--fsdp_reshard_after_forward false` (row 10) keeps the (unwrapped) `lm_head` resident across forward/backward, eliminating the repeated gathers; memory savings drop slightly (-47% vs -53%) but time flips from +50% to ~neutral. `SFTTrainer` now emits a `logger.warning` when it detects `loss_type='chunked_nll'` + FSDP2 + `reshard_after_forward=True` so users don't hit this unknowingly.

They all look mostly like this (almost perfect overlap), check out the Trackio links for detailed per-step loss comparisons:

Screenshot 2026-04-20 at 11 16 34 PM

<details>
<summary>benchmark_chunked_nll.py</summary>

```python
# Copyright 2020-2026 The HuggingFace Team. All rights reserved.
# Licensed under the Apache License, Version 2.0 (the "License");
# you may not use this file except in compliance with the License.
# You may obtain a copy of the License at
#     http://www.apache.org/licenses/LICENSE-2.0

# /// script
# dependencies = ["trl", "trackio"]
# ///

"""
Train one SFT run with a given `loss_type` and report peak memory + wall time.

Invoke twice (once with `--loss_type nll`, once with `--loss_type chunked_nll`) using the same
`--benchmark_id` to produce a comparable pair. Both runs log to the same trackio project under
different run names (`{benchmark_id}-nll` vs `{benchmark_id}-chunked_nll`). Running each loss
type as a separate process avoids cross-contamination from residual CUDA allocations.
"""

import argparse
import os
import time

import torch

def main() -> None:
    parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
    parser.add_argument("--loss_type", choices=["nll", "chunked_nll"], required=True)
    parser.add_argument("--benchmark_id", required=True, help="Shared id for the nll + chunked_nll pair.")
    parser.add_argument("--model", default="Qwen/Qwen3-1.7B")
    parser.add_argument("--dataset", default="trl-lib/Capybara")
    parser.add_argument("--max_steps", type=int, default=100)
    parser.add_argument("--batch_size", type=int, default=8)
    parser.add_argument("--gradient_accumulation_steps", type=int, default=1)
    parser.add_argument("--max_length", type=int, default=1024)
    parser.add_argument("--bf16", action=argparse.BooleanOptionalAction, default=True)
    parser.add_argument("--gradient_checkpointing", action=argparse.BooleanOptionalAction, default=True)
    parser.add_argument("--packing", action=argparse.BooleanOptionalAction, default=False)
    parser.add_argument("--attn_implementation", default=None, help="e.g. 'sdpa', 'eager', 'flash_attention_2'")
    parser.add_argument("--output_dir", default="/tmp/chunked_nll_bench")
    args = parser.parse_args()

    dataset = load_dataset(args.dataset)
    # Keep only single-turn rows → prompt-completion format so `completion_only_loss` kicks in.
    dataset = dataset.filter(lambda ex: len(ex["messages"]) == 2)
    dataset = dataset.map(
        lambda ex: {"prompt": [ex["messages"][0]], "completion": [ex["messages"][1]]},
        remove_columns=["messages"],
    )

    training_args = SFTConfig(
        output_dir=f"{args.output_dir}/{args.benchmark_id}/{args.loss_type}",
        loss_type=args.loss_type,
        run_name=f"{args.benchmark_id}-{args.loss_type}",
        max_length=args.max_length,
        per_device_train_batch_size=args.batch_size,
        gradient_accumulation_steps=args.gradient_accumulation_steps,
        max_steps=args.max_steps,
        logging_steps=1,
        save_strategy="no",
        report_to="trackio",
        trackio_space_id=os.getenv("TRACKIO_SPACE_ID"),
        seed=1234,
        data_seed=1234,
        bf16=args.bf16,
        gradient_checkpointing=args.gradient_checkpointing,
        packing=args.packing,
        model_init_kwargs={"attn_implementation": args.attn_implementation} if args.attn_implementation else None,
    )

    trainer = SFTTrainer(model=args.model, args=training_args, train_dataset=dataset["train"])

    torch.cuda.synchronize()
    start = time.perf_counter()
    trainer.train()
    torch.cuda.synchronize()
    train_time = time.perf_counter() - start

    if torch.distributed.is_initialized():
        t = torch.tensor(peak_gb, device="cuda")
        torch.distributed.all_reduce(t, op=torch.distributed.ReduceOp.MAX)
        peak_gb = t.item()
        if torch.distributed.get_rank() != 0:
            raise SystemExit(0)

    print(
        f"benchmark_id={args.benchmark_id} loss_type={args.loss_type}"
        f" peak={peak_gb:.2f} GB time={train_time:.2f} s"
    )


if __name__ == "__main__":
    main()
```

</details>

<details>
<summary>benchmark_chunked_nll.slurm (single-GPU / DDP)</summary>

```bash
#!/bin/bash
#SBATCH --job-name=chunked-nll-bench
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=1
#SBATCH --gres=gpu:1
#SBATCH --partition=hopper-prod
#SBATCH --time=2:00:00
#SBATCH --output=/fsx/qgallouedec/logs/%x-%j.out
#SBATCH --qos=normal

set -x
source ~/.bashrc
conda activate trl
echo "START TIME: $(date)"
cd /fsx/qgallouedec/trl
export TRACKIO_SPACE_ID="qgallouedec/chunked-nll-benchmark-2"

accelerate launch --main_process_port $((29500 + RANDOM % 1000)) benchmark_chunked_nll.py "$@"

echo "END TIME: $(date)"
```

</details>

<details>
<summary>benchmark_chunked_nll_fsdp.slurm (FSDP2)</summary>

```bash
#!/bin/bash
#SBATCH --job-name=chunked-nll-bench-fsdp
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=1
#SBATCH --gres=gpu:4
#SBATCH --partition=hopper-prod
#SBATCH --time=2:00:00
#SBATCH --output=/fsx/qgallouedec/logs/%x-%j.out
#SBATCH --qos=normal

set -x
source ~/.bashrc
conda activate trl
echo "START TIME: $(date)"
cd /fsx/qgallouedec/trl
export TRACKIO_SPACE_ID="qgallouedec/chunked-nll-benchmark-2"

accelerate launch \
    --main_process_port $((29500 + RANDOM % 1000)) \
    --mixed_precision bf16 \
    --use_fsdp \
    --fsdp_version 2 \
    --fsdp_auto_wrap_policy TRANSFORMER_BASED_WRAP \
    --fsdp_transformer_layer_cls_to_wrap Qwen3DecoderLayer \
    benchmark_chunked_nll.py "$@"

echo "END TIME: $(date)"
```

</details>

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,172 @@
...
+    valid = labels != -100
+    hidden = hidden[valid]
+    labels = labels[valid]
+    n_valid = hidden.size(0)
+
+    loss = hidden.new_zeros((), dtype=torch.float32)
+    correct = hidden.new_zeros((), dtype=torch.float32)
+    entropy_sum = hidden.new_zeros((), dtype=torch.float32)
+    if n_valid == 0:
+        return loss, correct, entropy_sum
```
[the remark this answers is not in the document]
qgallouedec: not correct

trl/trainer/sft_trainer.py
```diff
@@ -1285,23 +1483,38 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+
+                per_token_entropy = entropy_from_logits(shift_logits)
+                mask = shift_labels != -100
+                total = mask.sum()
+                if total > 0:
+                    entropy = (per_token_entropy * mask).sum() / total
                 else:
-                    raise ValueError("Expected 'attention_mask' or 'position_ids' in inputs.")
+                    entropy = per_token_entropy.new_zeros(())
                 entropy = self.accelerator.gather_for_metrics(entropy).mean().item()
```
qgallouedec, line 1574: these changes actually change the way the entropy is computed.

before: computed on the full sequence, including tokens masked in the loss -> requires decoding the full sequence
after: computed on the unmasked tokens

this allows
- to be closer to what user expect IMO
- to be comparable with liger and chunked nll returned entropy

I will isolate this change in a dedicated PR

docs/source/reducing_memory_usage.md
```diff
@@ -180,6 +180,20 @@ training_args = GKDConfig(..., use_liger_kernel=True)
 </hfoption>
 </hfoptions>
 
+## Chunked cross-entropy for reducing peak memory usage
+
+At large vocabulary sizes, the `[batch × seq_len × vocab]` logits tensor produced by the LM head is one of the dominant activations held in memory across forward and backward. `loss_type="chunked_nll"` in [`SFTTrainer`] avoids materializing it all at once: positions with `labels == -100` are dropped *before* the `lm_head` matmul, and the cross-entropy is computed in chunks of tokens using gradient checkpointing, so peak activation memory scales with `chunk_size × vocab_size` instead of `(batch × seq_len) × vocab_size`.
```
lewtun, line 185: Can the `chunk_size` be set by the user or is this not needed in practice? If the former, an example on how to change the default value would be good

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,176 @@
...
+    logit_scale: float = 1.0,
+    final_logit_softcapping: float | None = None,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """
+    Memory-efficient next-token cross-entropy over hidden states and an `lm_head` weight.
+
+    The full `lm_head` projection is never materialized. Positions where `labels == -100` are dropped before the
+    matmul, and the remaining tokens are processed in chunks of `chunk_size`. Each chunk's `[chunk_size, vocab_size]`
+    logits tensor is kept alive only during its own forward/backward pass via gradient checkpointing, so peak
+    activation memory is `chunk_size * vocab_size` instead of `N_valid * vocab_size`.
```
lewtun: What is `N_valid`? Did you mean `(batch_size * seq_len)`?

trl/trainer/sft_config.py
```diff
@@ -256,10 +264,10 @@ class SFTConfig(_BaseConfig):
...
         default="nll",
         metadata={
-            "help": (
-                'Type of loss to use. Possible values are `"nll"` (negative log-likelihood, default) and `"dft"` '
-                "(Dynamic Fine-Tuning, as described in https://huggingface.co/papers/2508.05629)."
-            )
+            "help": "Type of loss to use. Possible values are `'nll'` (negative log-likelihood, default), `'dft'` "
+            "(Dynamic Fine-Tuning, https://huggingface.co/papers/2508.05629), and `'chunked_nll'` (same math as "
+            "`'nll'` but skips the `'lm_head'` matmul on ignored tokens and chunks the CE to reduce peak memory; not "
+            "compatible with Liger, PEFT, VLM, or `'final_logit_softcapping'`)."
```
lewtun: You say here it's not compatible with `final_logit_softcapping` but looking at the code for `_chunked_cross_entropy_loss` it seems it actually is?

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,176 @@
...
+        chunk_loss = F.nll_loss(log_p, lbl, reduction="sum")
+        chunk_correct = (logits.argmax(dim=-1) == lbl).sum().float()
+        chunk_entropy = -(log_p.exp() * log_p).sum(dim=-1).sum()
+        return chunk_loss, chunk_correct, chunk_entropy
+
+    for start in range(0, n_valid, chunk_size):
+        h_chunk = hidden[start : start + chunk_size]
+        lbl_chunk = labels[start : start + chunk_size]
+        chunk_loss, chunk_correct, chunk_entropy = torch.utils.checkpoint.checkpoint(
+            _chunk, h_chunk, lm_head_weight, lbl_chunk, use_reentrant=False
```
lewtun: Could be nice to add a comment on why we pick `use_reentrant=False` here. I'm never quite sure which one to use under which conditions :)

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,176 @@
...
+        )
+        return _ChunkedCELMHeadOutput(
+            loss=loss,
+            logits=None,
+            past_key_values=outputs.past_key_values,
+            hidden_states=outputs.hidden_states,
+            attentions=outputs.attentions,
+            token_accuracy=token_accuracy,
+            entropy=entropy,
+        )
```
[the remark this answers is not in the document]
lewtun: This is a good point and we should make sure MoE training still works @qgallouedec

trl/trainer/sft_trainer.py
```diff
@@ -1010,6 +1193,21 @@ def __init__(
 
         self.aux_loss_enabled = getattr(model.config, "output_router_logits", False)
 
+        # Under FSDP2 with `reshard_after_forward=True` (accelerate's default), the chunked CE path triggers a
```
lewtun, line 1250: More of a question for `accelerate`, but is their default needed in general? If not, we could set `reshard_after_forward=True` in our example FSDP2 config too.

lewtun (changes requested):
Very nice feature! Implementation-wise it looks good to me, but I'd like to see some small experiments done with:

* MoEs
* ZeRO-3 + Ulysses and FSDP2 + CP

to make sure we understand whether there's any material difference for these architectures and long context plugins. They can be fixed in follow-up PRs, but in that case we should add a note in the docs that they are not currently supported

docs/source/reducing_memory_usage.md
sergiopaniego, line 185: We could mention some number here of how much memory saving is expected and how it affects speed. We have a large list of possible improvements in the docs and could be difficult to know which one to use or if a user should combine them

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,176 @@
...
+    The patched forward calls the base decoder directly (`model.model`) to obtain hidden states, skips the `lm_head`
+    matmul on positions with `labels == -100`, and computes the cross-entropy in chunks of `chunk_size` valid tokens.
+    It returns a [`_ChunkedCELMHeadOutput`] with `loss` set, `logits` set to `None`, and `token_accuracy` / `entropy`
+    fields set to the mean values over non-ignored tokens. When `labels` is `None`, the original `lm_head` path is used
+    so generation / evaluation without labels keeps working.
+
+    Not supported yet: VLM / multimodal models whose forward injects visual tokens outside the base decoder, and
+    PEFT-wrapped models.
+    """
+    final_logit_softcapping = getattr(model.config, "final_logit_softcapping", None)
```
albertvillanova, line 221: As @lewtun mentioned (https://github.com/huggingface/trl/pull/5575#discussion_r3115640738), here `final_logit_softcapping` is read from `model.config`

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,176 @@
...
+        return loss, correct, entropy_sum
+
+    loss = hidden.new_zeros((), dtype=torch.float32)
+
+    def _chunk(h, w, lbl):
+        logits = h.float() @ w.float().t()
+        if logit_scale != 1.0:
+            logits = logits * logit_scale
+        if final_logit_softcapping is not None:
+            logits = final_logit_softcapping * torch.tanh(logits / final_logit_softcapping)
```
albertvillanova, line 175: And here, `final_logit_softcapping` is applied.

trl/trainer/sft_config.py
albertvillanova: The help string will mislead users running any model whose config has a non-None `final_logit_softcapping` field:
- Either add a validation that raises a NotImplementedError to match the documentation
- Or remove `final_logit_softcapping` from the incompatibility list

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,176 @@
...
+    if n_valid == 0:
+        # Keep the loss connected to the autograd graph so `.backward()` works when an entire
+        # micro-batch is masked (can happen with completion-only loss + truncation).
+        loss = (hidden_states.float().sum() + lm_head_weight.float().sum()) * 0.0
+        return loss, correct, entropy_sum
+
+    loss = hidden.new_zeros((), dtype=torch.float32)
+
+    def _chunk(h, w, lbl):
+        logits = h.float() @ w.float().t()
```
albertvillanova, line 169: `lm_head.bias` is silently ignored here. Most modern causal LMs have `bias=False`, but models that do have a bias (some older or custom models) will silently produce incorrect logits, loss, and gradients.

There's no guard, no check at patch time, and none of the tested models in _CHUNKED_CE_MODEL_IDS would expose this.

At minimum, the patch function should assert model.lm_head.bias is None and raise a clear error if not.

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,176 @@
...
+    def _chunked_ce_forward(
+        self: torch.nn.Module,
+        input_ids: torch.Tensor | None = None,
+        attention_mask: torch.Tensor | None = None,
+        labels: torch.Tensor | None = None,
+        num_items_in_batch: torch.Tensor | int | None = None,
+        **kwargs,
+    ) -> CausalLMOutputWithPast:
+        kwargs.pop("use_cache", None)
+        outputs: BaseModelOutputWithPast = self.model(
```
albertvillanova: Does `self.model` exist for all supported models? Otherwise, it will raise an AttributeError.

Should we check at the top of `_patch_chunked_ce_lm_head` that `model.model` exists and raising a NotImplementedError with a clear message instead?

trl/trainer/sft_trainer.py
```diff
@@ -1285,23 +1483,38 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+            self._metrics[mode]["entropy"].append(entropy)
+        elif not self.args.use_liger_kernel:  # liger doesn't return logits
             with torch.no_grad():
-                per_token_entropy = entropy_from_logits(outputs.logits)
-                # When using Prompt Tuning, skip the virtual tokens in logits before entropy computation, since they
-                # do not correspond to actual input tokens.
+                # Shift logits/labels so entropy is averaged over the same set of tokens the loss is computed on
+                # (positions with `labels == -100` are skipped). This keeps the metric comparable across
+                # `completion_only_loss`, packing, and `chunked_nll`, which otherwise disagree on which
+                # positions to include.
```
albertvillanova, line 1551: This is a silent behavior change for `nll` path: it is an unannounced behavioral change to a logged metric, and it will affect existing users' dashboards without warning

docs/source/reducing_memory_usage.md
qgallouedec: Technically yes, it could be exposed, but I'd rather not. The default (256) was picked from a Pareto sweep across chunk sizes and sits right at the knee of the memory/throughput curve, so in practice there isn't much to gain by tuning it.

More generally, I don't think making every internal knob configurable is a good default. Each new config field is a long-term maintenance cost and most of them end up unused. I'd rather start minimal and let real user feedback drive the addition: if someone hits a case where 256 is clearly wrong for them, they'll tell us, and we can add the flag then with a concrete motivation.

For the records, you'd need this to change the chunk size:

```python
from trl.trainer import sft_trainer
sft_trainer._CHUNKED_LM_HEAD_CHUNK_SIZE = 512
```

trl/trainer/sft_trainer.py
qgallouedec: It's the number of tokens that aren't masked in the loss

eg

```python
loss_mask = [[-100, -100,  123,  456,  789],
             [-100,  101,  213,  141,  516]]
```

N_valid = 7

qgallouedec: > Did you mean (batch_size * seq_len)?

Ah yes you're right I think it should be `(batch_size * seq_len)` here!

qgallouedec: batch_size * seq_len * vocab_size actually

trl/trainer/sft_trainer.py
qgallouedec: `use_reentrant=False`: PyTorch recommended. So always use this unless you know what you're doing ;) see https://github.com/huggingface/transformers/pull/43203

qgallouedec: so follow-up question, why passing `False` explicitly if it's the default?

It's because it's not the default, nor `True` is:  PyTorch requires this argument to be passed (otherwise, it fails starting with PyTorch version 2.9)

trl/trainer/sft_config.py
qgallouedec: good catch, it was a stale comment, fixed in https://github.com/huggingface/trl/pull/5575/commits/bca6c537d7a68adcfb17b12245f76f28f101559a

trl/trainer/sft_trainer.py
qgallouedec: fixed in https://github.com/huggingface/trl/pull/5575/commits/2a9024d1e03f601f97d9acb0cee3531e980a6c07

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,172 @@
...
+            If set, applies `softcap * tanh(logits / softcap)` to each chunk's logits before the cross-entropy,
+            matching the `final_logit_softcapping` behavior of Gemma-style models. Applied after `logit_scale`.
+
+    Returns:
+        `tuple[torch.Tensor, torch.Tensor, torch.Tensor]`: scalar loss, scalar token accuracy, and scalar mean Shannon
+        entropy (in nats) — all taken over non-ignored tokens in the local batch. When there are no valid tokens, all
+        three are `0.0`.
+    """
+    hidden = hidden_states[..., :-1, :].reshape(-1, hidden_states.size(-1))
+    labels = labels[..., 1:].reshape(-1)
```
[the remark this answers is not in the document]
qgallouedec: fix in https://github.com/huggingface/trl/pull/5575/commits/2a9024d1e03f601f97d9acb0cee3531e980a6c07

trl/trainer/sft_trainer.py
qgallouedec: Yes, as mentioned in https://github.com/huggingface/trl/pull/5575/changes#r3114879186, I'll decouple this part from this PR

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,172 @@
...
+        outputs: BaseModelOutputWithPast = self.model(
+            input_ids=input_ids, attention_mask=attention_mask, use_cache=False, **kwargs
+        )
+        hidden_states = outputs.last_hidden_state
+
+        if labels is None:
+            return CausalLMOutputWithPast(
+                loss=None,
+                logits=self.lm_head(hidden_states),
+                past_key_values=outputs.past_key_values,
```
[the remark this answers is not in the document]
qgallouedec: https://github.com/huggingface/trl/pull/5575/commits/946c782d5a36f6e493934aa2541a7c106afca58f

trl/trainer/sft_trainer.py
qgallouedec: #5617

docs/source/reducing_memory_usage.md
qgallouedec: thanks, done https://github.com/huggingface/trl/pull/5575/commits/9a96483a33997a3af598eebfa7360b9d1cd1cb70

trl/trainer/sft_trainer.py
qgallouedec: #5617

trl/trainer/sft_trainer.py
qgallouedec: Accelerate's FSDP2 default is `reshard_after_forward=True`, applied uniformly to every `fully_shard()` call. Interestingly, it's overriding PyTorch's `None` default (which would leave the root unit's params, including `lm_head`, resident). Any idea why @SunMarc?

In my understanding, in general, `True` is fine: the backward pass re-gathers each unit's params exactly once, overlappable with compute. It only bites when the same unsharded params are used in multiple forward passes within a step, which is what `chunked_nll` does (one `lm_head.weight` all-gather per chunk).

trl/trainer/sft_trainer.py
qgallouedec: good catch, fixed here: https://github.com/huggingface/trl/pull/5575/commits/befd8fe93bd26532e55b0409f08931449cfc5e70

trl/trainer/sft_trainer.py
qgallouedec: right! the right way is `self.get_decoder()` instead of `self.model`. Fixed in https://github.com/huggingface/trl/pull/5575/commits/941887de257679eb658e57e9b3a8803eea5e5e9c

trl/trainer/sft_trainer.py
```diff
@@ -65,10 +68,194 @@
...
+    fields set to the mean values over non-ignored tokens. Also accepts pre-shifted `shift_labels` in place of
+    `labels`, for the context / sequence parallelism path. When both are `None`, the original forward is invoked so
+    generation and labels-free evaluation preserve any per-model logits post-processing (e.g. `logit_scale`,
+    `final_logit_softcapping`).
+
+    Not supported yet: VLM / multimodal models whose forward injects visual tokens outside the base decoder, and
+    PEFT-wrapped models.
+    """
+    final_logit_softcapping = getattr(model.config, "final_logit_softcapping", None)
+    logit_scale = getattr(model.config, "logit_scale", 1.0)
```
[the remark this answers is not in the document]
qgallouedec: I think it's the only way here.

trl/trainer/sft_trainer.py
SunMarc: I think this was more to copy FSDP setup, but FSDPv2 api is indeed a bit more flexible since you can `fully_shard` directly the modules you want. A quick fix to that and this will probably match what torchtitan do is to just set it to `reshard_after_forward =None` when sharding the remaining modules that we have. 

One caveat is that usually we should reshard the embedding unless it is tied with the `lm_head`. But for now, let's just not reshard it and we can deal with this case later if needed. The new FSDPv2 impl from @3outeille should have better defaults.

SunMarc: can you try this ? https://github.com/huggingface/accelerate/pull/4015

qgallouedec: Way better, though is seems like `reshard=False` is still a faster, which is expected in my opinion

| Config | Δmem | Δtime |
| -- | -- | -- |
| Row 9 — old default (reshard=True) | −53.1 % | +49.7 % |
| Row 10 — explicit reshard=False override | −47.3 % | −6.3 % |
| accelerate#4015 | −53.1 % | +16.4 % |

SunMarc: Yes, that's there is a tradeoff between memory and communication. We can still advertise users to set it to False if they want faster fsdp.

albertvillanova (approved):
Thanks.

trl/trainer/sft_trainer.py
```diff
@@ -976,8 +1213,22 @@ def __init__(
                         "passing a `compute_loss_func` is not allowed."
                     )
                 compute_loss_func = dft_loss
+            elif args.loss_type == "chunked_nll":
+                # Same math as `"nll"` but the `lm_head` matmul is skipped on ignored tokens and the CE is computed in
+                # chunks of tokens. Implemented by patching the model's forward before `super().__init__` so accelerate
+                # wraps the patched forward.
+                if self._is_vlm:
+                    raise NotImplementedError("`loss_type='chunked_nll'` is not supported for VLM models yet.")
+                if peft_config is not None or is_peft_model(model):
```
BenjaminBossan, line 1222: Nice to see this in trl. I don't understand why it doesn't work with PEFT though, is there anything we can do to enable it @qgallouedec?

qgallouedec: Probably nothing is blocking. But the PR was already big, so I wanted to separate the concerns.

As for VLM I will dedicate a PR to PEFT x chunked loss.

qgallouedec: In other words, it's not "not working with peft" it's just not tested, and I'll do it soon

BenjaminBossan: Ah nice, looking forward to it. After a first glance, I don't think anything should conflict with PEFT, but LMK if you find any issues.

jiosephlee:
some questions:
(1) is the liger baseline the fused variant? 
(2) since the optimization comes from dropping ignored labels, I imagine the benefits are highly dependent on the ratio between masked and unmasked tokens? 
(3) wouldn't this approach be compatible with liger kernels which optimize the CE (or lm_head + CE for fused) kernels?

qgallouedec:
(1) yes, chunked and fused
(2) partially yes. actually, two sources for the optimization: 1. chunking reduces the peak (even with no mask token) and 2. no forward / backward on masked tokens
(3) I think this is already integrated in liger. the reason to have it here is for the user to benefit from this optimization without having to install an additional dep

jiosephlee:
@qgallouedec Thanks for the quick response. Following-up on a few things:

(1) On my own ablations, liger is showing a +14GB improvement (144GB vs 158GB) on B200s (length: 4096, FA2 padding-free) over chunked_nll on Qwen3-8B. Maybe this warrants further examination so we can inform users more accurately on when chunked_nll does better and does not? 

(3) Looking at the code now, it doesn't seem to do this https://github.com/linkedin/Liger-Kernel/blob/3bb3b3fae6d0b2356116034a7f0ee1dde0ea71ea/src/liger_kernel/ops/fused_linear_cross_entropy.py#L99

But it's my first pass of the code so I might have looked past something.

Issue #5207, opened 2026-03-03
CI often fails for TestDPOTrainer::test_train_toolcall_data: torch.OutOfMemoryError: CUDA out of memory


albertvillanova:
CI often fails for `TestDPOTrainer::test_train_toolcall_data`: https://github.com/huggingface/trl/actions/runs/22581457191/job/65414409040
> torch.OutOfMemoryError: CUDA out of memory

```python
  FAILED tests/test_dpo_trainer.py::TestDPOTrainer::test_train_toolcall_data - torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 76.00 MiB. GPU 0 has a total capacity of 14.74 GiB of which 42.19 MiB is free. Process 21219 has 284.00 MiB memory in use. Process 21222 has 3.40 GiB memory in use. Process 21213 has 258.00 MiB memory in use. Process 21216 has 10.77 GiB memory in use. Of the allocated memory 10.59 GiB is allocated by PyTorch, and 42.18 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

Stacktrace:
```python
   >       trainer.train()
  
  tests/test_dpo_trainer.py:852: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  .venv/lib/python3.14/site-packages/transformers/trainer.py:1412: in train
      return inner_training_loop(
  .venv/lib/python3.14/site-packages/transformers/trainer.py:1742: in _inner_training_loop
      tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  trl/trainer/dpo_trainer.py:1422: in training_step
      return super().training_step(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  .venv/lib/python3.14/site-packages/transformers/trainer.py:1951: in training_step
      loss = self.compute_loss(model, inputs, num_items_in_batch=num_items_in_batch)
[...]

PR #5680, opened 2026-04-29
Fix OOM in CI by reducing image size of tiny Gemma3 model

albertvillanova:
Fix OOM in CI by reducing image size of tiny Gemma3 model.

This PR introduces a targeted adjustment for the `google/gemma-3-4b-it` model in the `generate_tiny_models` script to address memory usage issues related to image processing.

Partial fix for:
- #5207

### Motivation

The `tiny-Gemma3ForConditionalGeneration` model was generated with the default SigLIP image size of 896×896, which produces 4,096 patches per image. During training, the vision encoder attention maps have shape `[batch, heads, 4096, 4096]`, consuming ~1 GB per layer. With 2 vision layers and backpropagation, a single Gemma3 test consumes 5–7 GiB of GPU memory. Two such tests running concurrently on a 14.74 GiB GPU caused CUDA out-of-memory errors in all other parallel workers.

### Solution

Override `image_size=224` (256 patches) when generating the tiny Gemma3 model. This is consistent with `mm_tokens_per_image=256` in the Gemma3 config: the projector's `AvgPool2d` gets `kernel_size=1` (identity), which is architecturally valid. The processor's image processor size is updated to match so that test inputs are also resized to 224×224.

### Changes

Model-specific configuration:

* For the `google/gemma-3-4b-it` model, sets `vision_config["image_size"]` and `processor.image_processor.size` to 224x224 (instead of the default 896x896) to reduce memory consumption during training by limiting the number of image patches and ensuring the projector's average pooling layer acts as an identity function.

---

> [!NOTE]
> **Low Risk**
> Targeted, model-specific config tweak in a test-only model generation script; low blast radius beyond the tiny Gemma3 artifacts and related CI tests.
> 
> **Overview**
> Reduces memory usage when generating the tiny `google/gemma-3-4b-it` VLM by overriding its vision settings in `scripts/generate_tiny_models.py`.
> 
> For this model only, the script now forces a smaller vision `image_size` (224) and scales down both text/vision `intermediate_size` to keep activations small, and updates the processor’s image resize settings to match so CI/unit tests don’t OOM.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ad45e25b6e367f4e30997903bf44ebe7246b519. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/generate_tiny_models.py
+++ b/scripts/generate_tiny_models.py
@@ -359,9 +359,12 @@ for model_id, model_class, dtype in [
     kwargs = {}
 
     if model_id == "google/gemma-3-4b-it":
-        # Gemma3 SigLIP processes images at 896×896 → 4,096 patches, producing 1 GB+ attention matrices per layer
-        # during training. Use 224×224 → 256 patches instead; this matches mm_tokens_per_image=256, so the
-        # projector's AvgPool2d degenerates to kernel_size=1 (identity) and memory stays manageable.
+        # Gemma3 SigLIP defaults to image_size=896 → 4,096 patches. With production-scale intermediate_size
+        # (~4,304), the vision FFN saves [batch, 4,096, 4,304] activations for backprop (~134 MB/layer).
+        # Use image_size=224 → 256 patches (matching mm_tokens_per_image=256, so the projector's AvgPool2d
+        # gets kernel_size=1, i.e. identity) and scale down intermediate_size to keep activations tiny.
+        text_config["intermediate_size"] = 32
+        vision_config["intermediate_size"] = 32
         vision_config["image_size"] = 224
         processor.image_processor.size = {"height": 224, "width": 224}
 
--- a/scripts/generate_tiny_models.py
+++ b/scripts/generate_tiny_models.py
@@ -358,6 +358,13 @@ for model_id, model_class, dtype in [
     }
     kwargs = {}
 
+    if model_id == "google/gemma-3-4b-it":
+        # Gemma3 SigLIP processes images at 896×896 → 4,096 patches, producing 1 GB+ attention matrices per layer
+        # during training. Use 224×224 → 256 patches instead; this matches mm_tokens_per_image=256, so the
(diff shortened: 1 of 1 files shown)


qgallouedec:
thanks! can you just confirm with a forward + gpu peak memory measurement for old vs new?

albertvillanova:
Thanks for your sensible suggestion: the difference is small. I continue investigating...

albertvillanova:
I think the `intermediate_size` should also be reduced: from 4304 to 32.

albertvillanova:
Here are the measurements. The profile script runs a single forward+backward pass in isolation (no trainer overhead, no other tests sharing the GPU):
- image_size=896, intermediate_size=4304 (main):  peak = 8.205 GiB
- image_size=224, intermediate_size=32   ([pr/10](https://huggingface.co/trl-internal-testing/tiny-Gemma3ForConditionalGeneration/discussions/10/files)): peak = 6.677 GiB
- saving: 1.528 GiB

Running the actual CI test (test_train_vlm[tiny-Gemma3ForConditionalGeneration]) with a per-test delta measurement gives a smaller visible saving:
- main:  peak GPU (delta) = 9.343 GiB
- PR:    peak GPU (delta) = 9.215 GiB
- saving: 0.128 GiB

The discrepancy is expected. The true peak in a full training run is reached during the loss computation, not the vision encoder: `vocab_size=262208` was never scaled down for the tiny model, so the logits tensor `[8, 279, 262208]` in float32 dominates at ~6–7 GiB for both models equally. The 1.528 GiB saving from the vision encoder fixes is real but occurs at an earlier, lower peak that is never the maximum.

I therefore opened a companion PR:
- #5687 

The fixes in this PR remain valuable: they make the model genuinely tiny as intended.

qgallouedec:
Thanks!

qgallouedec (approved):

PR #5319, opened 2026-03-20
(1/5) Add callback to sync weights before training begins

AmineDiro:
# What does this PR do?
- Introduces the _InitialWeightSyncCallback class to handle the first weight sync.
- Replaces the manual self._sync_weight() call inside AsyncGRPOTrainer._inner_training_loop with this dedicated callback.
- Registers the callback in the trainer's __init__.
- This ensures vLLM is guaranteed to use the current policy weights (after FSDP wrapping and moving parameters to the GPU) before producing the first batch of samples.

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

---

> [!NOTE]
> **Medium Risk**
> Changes the training startup sequence by moving the first vLLM weight sync and rollout worker start into an `on_train_begin` callback, which could affect distributed initialization/order and rollout generation timing.
> 
> **Overview**
> Ensures the vLLM rollout worker sees the correct policy weights *before* the first training step by adding `_InitialWeightSyncCallback`, which runs a one-time `_sync_weight()` on `on_train_begin`, starts the rollout worker on the main process, and then unregisters itself.
> 
> Removes the previous manual initial sync/start logic from `AsyncGRPOTrainer._inner_training_loop`, keeping periodic weight syncing via `StepIntervalCallback` unchanged.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4bf921153cfc5c9b90392c7c33dd7ce7168ce680. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #5749, opened 2026-05-11
feat: move async rollout worker to separate process

AmineDiro:
# What does this PR do?

Rollout generate and score loops now run in a spawned child process instead of a thread inside the trainer. The trainer's autograd engine no longer competes with `recursive_parse` / `accuracy_reward` for the GIL.

* `AsyncRolloutWorker` (parent): spawns the child rollout process. owns the shared `mp.Queue` / `mp.Value` / `mp.Event`s.
* `_AsyncRolloutLoop` handles the rollout logic (private, child-only): tokenizer, dataset iter, reward funcs, asyncio loops.
* `WeightTransferClient` (new, `weight_transfer.py`): NCCL group with vLLM + `/pause`, `/resume`, `/init_weight_transfer_engine`, `/update_weights`. The rollout worker only talks to `/v1/completions`.

Two callbacks register in `on_train_begin` (post-`accelerator.prepare`), guarded with `_fired` flags:

* `_InitialWeightSyncCallback`: `init_weight_transfer()` + cold `_sync_weight()`
* `_StartRolloutWorkerCallback`: `rollout_worker.start()` (registration order enforces "sync before start")

`remove_callback(type(self))` (PR #5319's pattern) drops the next callback in line when used with two callbacks, because `CallbackHandler.call_event` iterates the list directly. `_fired` is safe.

Two correctness fixes folded in because they touch the same `_AsyncRolloutLoop` methods this PR moves (would conflict otherwise):

- **widen aiohttp retry**: the narrow `(ServerDisconnectedError, ClientConnectionError, ClientResponseError)` catch in `_generate_one_turn` and `_post` missed `ClientPayloadError` (which vLLM's async engine fires when `asyncio.CancelledError` interrupts a long request mid-stream), killing the trainer on a single transient network blip. Broadened to `(aiohttp.ClientError, asyncio.TimeoutError, TimeoutError, ConnectionResetError)` with bounded exponential backoff. Shared `_retry_on_http_error` helper used by both call sites.
- **preserve NaN in `_score_group`**: `np.nansum` on an all-NaN column silently returns 0, so a completion where every reward func returned `None` (gold unparseable by `math_verify` ~30% of DeepMath / OpenR1-Math rows) ended up with reward 0 and a real advantage signal that pushed the policy away from actually-correct text answers. Mark all-NaN columns NaN, compute advantage on the scorable subset only, unscorable advantages = 0.

## Why

Training a Qwen3-30B-A3B @ 16k completion length, rollout's per-completion Python work holds the GIL in 1-5s bursts. Rank 0's autograd engine starves especially with deepspeed backend so ranks 1-7 timeout on the next NCCL collective. 

I used both`py-spy` and NCCL flight recorder info to confirm `MainThread` idle on `_engine_run_backward` while `recursive_parse` ran active and holds the GIL.

Also moved the weight init to a  callback structure to also fixes the DS-Z2 multi-node crash where the NCCL StatelessProcessGroup's IPC pages were bound before `Stage1And2ZeroOptimizer.__init__ → empty_cache()`, as I was getting`cudaErrorIllegalAddress`.

## Validation

| Config | Steps | `train_runtime` |
|---|---|---|
| Qwen3-30B-A3B + DS-Z2 + EP=8 + 16k DAPO, 2 trainer nodes + 1 vLLM | 5 | 924.3s |
| Qwen3-4B + OpenR1-Math (`examples/scripts/async_grpo.py`), 1 node 2 GPUs | 10 | 229.8s |

No `cudaErrorIllegalAddress`, no SIGBUS, no NCCL watchdog timeouts.

## Test plan

* [x] Multi-node DS-Z2 + EP=8 + 30B end-to-end
* [x] Single-node `examples/scripts/async_grpo.py` end-to-end
* [x] `tests/experimental/test_async_grpo_trainer.py` against stub worker (stub trimmed to new protocol)

---

> [!NOTE]
> **High Risk**
> Changes distributed training lifecycle (NCCL init timing, multiprocessing, weight sync split) and reward/advantage semantics; affects large-scale async GRPO runs where GIL and multi-node stability were the motivation.
> 
> **Overview**
> **Async GRPO** now runs rollout generation and scoring in a **spawned child process** instead of a background thread, so heavy reward work (e.g. `recursive_parse`) no longer contends with training on the GIL.
> 
> **vLLM NCCL weight sync** moves out of the rollout worker into a new **`WeightTransferClient`** on the trainer. The rollout child only calls **`/v1/completions`**. Training begins with idempotent callbacks: initial NCCL setup + cold sync, then **`rollout_worker.start()`** (after `accelerator.prepare`).
> 
> **`RolloutWorkerProtocol`** drops `pause` / `resume` / `send_weights`. Docs warn that **`reward_funcs` / `tools` / `environment_factory` must be picklable** and the child runs **CPU-only** (`CUDA_VISIBLE_DEVICES=""`). **`get_soft_overlong_punishment`** returns a picklable callable class.
> 
> **Correctness:** broader **aiohttp retries** with backoff; **all-NaN rewards** stay NaN with advantages computed only on scorable completions (unscorable → 0); trainer metrics use **NaN-aware** aggregation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 077d6af00fee65993d586df1d09a64acb8733392. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wadeKeith (commented):
Withdrawn; this review was too cursory.

AmineDiro:
[347b9b2](https://github.com/huggingface/trl/pull/5749/commits/347b9b29e443da55d8fd3d686ecf1554eb4d8829)

this commit fixes : 

- **widen aiohttp retry**: the narrow `(ServerDisconnectedError, ClientConnectionError, ClientResponseError)` catch in `_generate_one_turn` and `_post` missed `ClientPayloadError` (which vLLM's async engine fires when `asyncio.CancelledError` interrupts a long request mid-stream), killing the trainer on a single transient network blip. Broadened to `(aiohttp.ClientError, asyncio.TimeoutError, TimeoutError, ConnectionResetError)` with bounded exponential backoff. Shared `_retry_on_http_error` helper used by both call sites.
- **preserve NaN in `_score_group`**: `np.nansum` on an all-NaN column silently returns 0, so a completion where every reward func returned `None` (gold unparseable by `math_verify` ~30% of DeepMath / OpenR1-Math rows) ended up with reward 0 and a real advantage signal that pushed the policy away from actually-correct text answers. Mark all-NaN columns NaN, compute advantage on the scorable subset only, unscorable advantages = 0.

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -752,17 +679,140 @@ async def _score_group(self, group: RolloutGroup) -> list[RolloutSample]:
...
+        if self._process is not None:
+            logger.warning("AsyncRolloutWorker.start() called but child process is already running; ignoring.")
+            return
+        # Reset so spawn-import latency (~tens of seconds) doesn't immediately trip check_health.
+        self._heartbeat_value.value = time.time()
+        loop_kwargs_pkl = pickle.dumps(self._loop_kwargs)
+        self._process = self._mp_ctx.Process(
+            target=_child_main,
+            args=(
+                loop_kwargs_pkl,
```
qgallouedec: Can we pass `self._loop_kwargs` directly? `spawn` pickles the args in `Process.start()` anyway, so the manual `dumps` + `loads` in `_child_main` looks redundant

qgallouedec (approved):
Thanks, and sorry for the late review. Overall lgtm.
One concern: moving rollout to a `spawn` child means everything forwarded to it gets pickled, which silently narrows what `reward_funcs` / `tools` / `environment_factory` can be. (Tokenizer, scalars, and datasets are fine: a memory-mapped dataset pickles as a ~0-byte file reference.) The risk is in the callables: They must now be be picklable:

Module-level `def`s, `functools.partial`, and callable instances work; but lambdas and closures fail at `start()` with a raw pickle traceback. E.g. `trl.rewards.get_soft_overlong_punishment` returns a closure and breaks. It's also a regression vs. sync `GRPOTrainer`, where
closures work (reward funcs are called in-process, never pickled). `environment_factory` is usually a class, so it's only at risk if passed as a lambda.

Also, less important: GPU reward models also stop working. `_scrub_child_env()` sets `CUDA_VISIBLE_DEVICES=""`, so a reward model pickles fine but runs CPU-only in the child silently slow/wrong, no error. Under the old thread model it shared the trainer's GPU. But I don't think it's a very common setup. So this could be ignored and just documented.

Suggestions:

1. Fail fast: `try: pickle.dumps(self._loop_kwargs)` in `start()`, re-raise with a message naming the constraint (must be picklable, not lambda/closure).
2. Make `get_soft_overlong_punishment` picklable (module-level callable class with a `__name__`), so a first-party reward isn't broken.
3. Document that reward funcs run in a separate CUDA-free process.

Not blocking but worth addressing before merging.

---

**Secondary: all-None reward fix vs. main.** The folded-in `_score_group` fix (all-NaN rows excluded from the advantage baseline) is correct, but main GRPO/RLOO still `nansum`→0 and only warn. Impact is mostly masked by per-group normalization (whole-group-unscorable → advantage ≈ 0 anyway); the real bug is within-group variation (some completions unparseable, siblings not). Worth a follow-up to align main.

qgallouedec:
and I think https://github.com/huggingface/trl/pull/5749#discussion_r3329939125 and https://github.com/huggingface/trl/pull/5749#discussion_r3225357886 are valid, worth simplifying.

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: 100% right. I was going to use cloudpickle to have larger pickling support but that's not needed for now.

AmineDiro:
> Also, less important: GPU reward models also stop working. `_scrub_child_env()` sets `CUDA_VISIBLE_DEVICES=""`, so a reward model pickles fine but runs CPU-only in the child silently slow/wrong, no error. Under the old thread model it shared the trainer's GPU. But I don't think it's a very common setup. So this could be ignored and just documented.

Oh thats a very good catch! I wonder if running a reward model in rollout worker is a common usecase 🤔 

Thx for the review, all valid points/suggestions.  Implemented the suggestions in 0bae8ff0. 

> **Secondary: all-None reward fix vs. main.** The folded-in `_score_group` fix (all-NaN rows excluded from the advantage baseline) is correct, but main GRPO/RLOO still `nansum`→0 and only warn. Impact is mostly masked by per-group normalization (whole-group-unscorable → advantage ≈ 0 anyway); the real bug is within-group variation (some completions unparseable, siblings not). Worth a follow-up to align main.

Good catch. I'll open a PR to fix 👍🏼

qgallouedec:
> I wonder if running a reward model in rollout worker is a common usecase

Let's assume it's not common for now

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -53,13 +54,31 @@ def reset(self, **kwargs) -> str | None: ...
...
+        start():
+            Begin producing rollouts. Called once on train begin, after the initial weight sync.
+        stop():
+            Stop the worker and release its resources. Called on train end.
+        update_model_version(version):
+            Tell the worker which policy version is now live, so it can tag/discard stale samples.
+        check_health(stale_after_s):
+            Raise if the worker has crashed or stopped producing within `stale_after_s` seconds.
+    """
+
```
qgallouedec, line 67: it doesn't need to be fixed here, but for information, the method section doesn't render properly:

Image

qgallouedec: https://moon-ci-docs.huggingface.co/docs/trl/pr_5749/en/async_grpo_trainer#trl.experimental.async_grpo.async_grpo_trainer.RolloutWorkerProtocol

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -388,21 +458,21 @@ def __init__(
...
             self.rollout_queue = self.rollout_worker.rollout_buffer
         else:
             self.rollout_queue = None
             self.rollout_worker = None
+            self.weight_transfer = None
 
-        # Add callbacks
+        # Add callbacks. Registration order matters: weight sync first, then worker start.
+        self.add_callback(_InitialWeightSyncCallback(self))
+        self.add_callback(_StartRolloutWorkerCallback(self))
```
qgallouedec, line 476: I don't think it's a blocker, but for the record, I'm really not a fan of this back-reference pattern. If `transformers.Trainer` doesn't set `self.trainer`, it's probably because it's not the desired format. But let's keep going like this and see how it goes. We can re-visit later

trl/rewards/other_rewards.py
qgallouedec, line 1: good!

qgallouedec (approved):
ok let's go

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: Yeah, I agree but the hook based system need acess to self. 
Thinking about it should really be injected by the hook itself. The alternative inject self by capturing it in the closure  if I recall correctly 🤔

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: image
Good catch. Fixed in  220c4e8

trl/experimental/async_grpo/async_grpo_trainer.py
qgallouedec: yes no worry, it's not a blocker, we have more important questions to solve

PR #5751, opened 2026-05-12
async_grpo don't return on queue.Empty

AmineDiro:
# What does this PR do?

Don't return None from Queue as it can propagate None through the dataloader.

---

> [!NOTE]
> **Medium Risk**
> Changes async training data loading and failure detection; mis-tuned `heartbeat_stale_after_s` could abort long pauses, but default behavior is safer than silent None batches.
> 
> **Overview**
> **Async GRPO** no longer ends the rollout dataloader epoch when the rollout queue stays empty for a long time. On `queue.Empty`, `RolloutQueueDataset` now **polls** (default 5s) and calls **`check_health`** instead of returning, which avoids **`None`** batches propagating through Accelerate’s `dispatch_batches` path.
> 
> Training adds **`heartbeat_stale_after_s`** (default 300s) on `AsyncGRPOConfig`. The rollout worker exposes **`check_health`**: it raises if the background thread failed (with chained exception) or if generation/scoring loops have not updated **`_last_heartbeat`** within the threshold. Heartbeats are refreshed in `_generate_loop` and `_score_loop`; thread failures are recorded via `_failed` / `_exception`, and the heartbeat is reset on **`start()`** so a slow init does not false-positive.
> 
> The test stub implements a no-op **`check_health`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 883fd4d0b6185db7b874f073c212b4e6bee17f0c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -93,8 +96,18 @@ def __iter__(self):
...
+                    raise RuntimeError(
+                        f"Rollout queue empty for {self.max_empty_waits * self.timeout:.0f}s "
+                        f"({self.max_empty_waits} consecutive {self.timeout}s waits); rollout worker is dead."
+                    ) from None
+                logger.warning(
+                    f"Rollout queue empty for {self.timeout}s "
+                    f"(empty waits: {empty_waits}/{self.max_empty_waits}); continuing to wait."
+                )
+                continue
+            empty_waits = 0
```
qgallouedec: we need both this and `safe_float`?

qgallouedec:
@codex review

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: safe_float is for the advantage if all reward returns None. This is more for waiting for rollouts to land and not iterating and yielding `None` values if I'm not mistaken

qgallouedec: What real code path produces a `None` in `sample.metrics`?

`AsyncRolloutWorker._score_group` already normalizes `None → float("nan")` at 

https://github.com/huggingface/trl/blob/23201dd7989ddde855de28a2c1d4523c06aa9d7a/trl/experimental/async_grpo/async_rollout_worker.py#L674

and every value stored in `sample.metrics` (`reward`, `reward_std`, `rewards/*`, `tools/*`, generation throughput, …) is wrapped in `float(...)`. The all-None reward case from the description ends with `advantage = 0.0` and `reward = 0.0` via `np.nansum` (never `None`). `torch.tensor([nan, …], dtype=torch.float32)` works without `_safe_float`.

```python
from trl.experimental.async_grpo.async_grpo_trainer import DataCollatorForRollout
ex = [{"input_ids":[1,2], "completion_mask":[0,1], "old_log_probs":[0.0,-1.0],
       "advantage":0.0, "metrics":{"reward": float("nan")}}]
DataCollatorForRollout(pad_token_id=0).torch_call(ex)  # OK without _safe_float
# Replacing nan with None *does* crash, but nothing in the current pipeline produces that.
```

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -78,13 +78,16 @@ def on_step_end(self, _args, state, _control, **_kwargs):
 
 
 class RolloutQueueDataset(torch.utils.data.IterableDataset):
-    def __init__(self, rollout_queue, model_version_fn, max_staleness=3, timeout=120.0):
+    def __init__(self, rollout_queue, model_version_fn, max_staleness=3, timeout=120.0, max_empty_waits=10):
```
qgallouedec: it seems like we aren't actually detecting a dead worker. It just waits 20 min before assuming. A worker that crashed on the first `_generate_one` call still takes 20 min to surface. The direct fix for the root cause would be to check the worker thread's liveness on each empty wait:

```python
except queue.Empty:
    if not self._worker_alive_fn():  # pass in worker.is_alive
        raise RuntimeError("Rollout worker thread is dead.")
    logger.warning(...)
    continue
```

would this work? If so it would probably be a better fix

qgallouedec:
I've thought about it, and I believe a better long-term fix is the exception channel + heartbeat pattern. The empty-queue signal can't distinguish slow / crashed / hung producers, so a retry budget can't either; surface health from the worker directly.

1. On the worker: add `_exception`, `_failed: Event`, `_last_heartbeat: float`. Wrap `_run` in `try/except BaseException` that stashes the exception and sets the event. Bump `_last_heartbeat` at the top of `_generate_loop` and `_score_loop` (orchestrator-level — ticks fast in healthy ops, stays stale during hangs).

2. Add `check_health(stale_after_s)` that raises with `__cause__=self._exception` if failed, or with a stale-heartbeat message past the threshold. Add to `RolloutWorkerProtocol`.

3. `RolloutQueueDataset`: short poll (e.g. 5 s), call `check_health_fn` on every empty wait. Plumb `self.rollout_worker.check_health` from `get_train_dataloader`.

Crashes surface immediately with the original traceback; hangs surface within `stale_after_s`; slow-but-alive producers don't trigger false alarms. One principled knob instead of `max_empty_waits × timeout`.

tell me what you think

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: I agree, but this, in my opinion, should not be defensive. The worker thread should always be running and stop cleanly ( like a classical service). That's why we should catch all exceptions inside the loop.
The worker should only stop at the trainer's signal.

AmineDiro:
> tell me what you think

yeah I think you're right. Looking at the worker today, it already raises in multiple places  so it's already a fallible process, not a service. Making it truly never-raise like a service will probably swallow real bugs (like reward func raising every call, .. ) and hide vLLM hangs/crashes. 👍🏼 

I'll change the impl to use heartbeat and the channel but I think it should be a notify like channel because the worker is a subprocess now, so event won't really work + exception might not be able to pickle correctly.

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: Oh my bad, this was a local change due to having none before I added due to having a local change that produces None. 

```python
  def _sanitize_metric(v: float) -> float | None:
      if isinstance(v, float) and (np.isnan(v) or np.isinf(v)):
          return None
      return v
 metrics={
      "reward":      _sanitize_metric(float(reward)),
      "reward_std":  _sanitize_metric(reward_std),
      **{f"rewards/{name}": _sanitize_metric(float(func_reward)) for ...},
      **tm,
  },
 ```

AmineDiro:
@qgallouedec  implemented in [1b14b33](https://github.com/huggingface/trl/pull/5751/commits/1b14b332a97476cd36668b74b6afb40c743b30f8) but will probably need different impl in https://github.com/huggingface/trl/pull/5749

EDIT: Impl in [1d80480](https://github.com/huggingface/trl/pull/5749/commits/1d8048033bfa713d9571be6999ff5238c3c8722c)

qgallouedec (approved):
lgtm!

qgallouedec:
@codex review

PR #5757, opened 2026-05-12
Tighten old_per_token_logps recomputation check in GRPO

### LINKED ISSUE #5770: Optimize `old_per_token_logps` recomputation in GRPOTrainer: per-rollout window check instead of static modulo
### Feature request

Replace the static modulo check that gates `old_per_token_logps` recomputation in `GRPOTrainer` with a precise per-rollout window check.

Currently, when `gradient_accumulation_steps` is not a multiple of `steps_per_generation × num_iterations`, the trainer recomputes `old_per_token_logps` for **every** rollout — even when the specific rollout's consumption window does not cross an `optimizer.step()` boundary.

The proposed change replaces:

```python
if self.args.gradient_accumulation_steps % generate_every != 0:
    old_per_token_logps = recompute(...)
```

with a per-rollout time-domain check:

```python
a = self._step
b = ((a // generate_every) + 1) * generate_every - 1
weights_change_mid_rollout = (a // grad_accum) != (b // grad_accum)
if weights_change_mid_rollout:
    old_per_token_logps = recompute(...)
```

This skips unnecessary forward passes when the rollout window stays within a single grad-accum bucket, while preserving identical behaviour in the default (aligned) configuration.

Related PR: #5757 

### Motivation

In non-aligned configurations (where `gradient_accumulation_steps` is not a multiple of `generate_every`), the current static modulo check is over-conservative — it triggers a full `_get_per_token_logps_and_entropies` forward pass for every rollout, regardless of whether the model weights actually changed mid-window.

Concrete example (`spg=2`, `num_iter=1`, `grad_accum=3`, `generate_every=2`):
- Only 2 out of every 6 rollouts truly need recomputation
- The current check recomputes for all 6 — wasting 4 forward passes

This adds up to significant unnecessary compute in large-scale GRPO training runs with non-default accumulation settings. The fix is correctness-preserving and introduces zero regression for the default config path.

### Your contribution

I have a working implementation ready and will submit a PR shortly #5757 . The change is minimal (~10 lines in `grpo_trainer.py`) and correctness-preserving — identical behaviour in the default config, with fewer unnecessary forward passes in non-aligned configurations. I will also include a unit test to verify the recomputation decisions for representative non-aligned settings (e.g. `spg=2, num_iter=1, grad_accum=3`).

wengeezhang:
# What does this PR do?

Fix #5770 

Refines the condition that gates `old_per_token_logps` recomputation in `GRPOTrainer` from a coarse static modulo check to a precise per-rollout window check. The result is fewer unnecessary forward passes in configurations where `gradient_accumulation_steps` is not a multiple of `steps_per_generation × num_iterations`, with **identical behaviour** in the default (aligned) configuration.

## Background

GRPO consumes one rollout across `generate_every = steps_per_generation × num_iterations` micro-steps. The PPO ratio requires `π_θ_old` to be the policy at *rollout time*. If `optimizer.step()` fires **strictly within** a rollout's consumption window, the model has changed mid-window — `old_per_token_logps` must be snapshotted (cannot use `.detach()` of the current logp). When no `optimizer.step()` fires inside the window, `old == current` and `.detach()` is mathematically equivalent and saves a full forward pass.

## Why the existing check is over-conservative

The current condition

```python
if self.args.gradient_accumulation_steps % generate_every != 0 or (
    self.use_vllm and self.vllm_importance_sampling_correction
):
    old_per_token_logps = recompute(...)
```

is sufficient but not necessary: whenever `grad_accum` is not a multiple of `generate_every`, it triggers recomputation **for every rollout**, regardless of whether the specific rollout's consumption window actually crosses an `optimizer.step()` boundary.

**Concrete example** (`spg=2`, `num_iter=1`, `grad_accum=3`, so `generate_every=2`, `grad_accum % generate_every == 1`):

| `_step` (rollout) | window `[a, b]` | bucket of `a` | bucket of `b` | needs recompute? | current code |
|---|---|---|---|---|---|
| 0  | [0, 1] | 0 | 0 | **No**  | Yes (wasted) |
| 2  | [2, 3] | 0 | 1 | Yes     | Yes |
| 4  | [4, 5] | 1 | 1 | **No**  | Yes (wasted) |
| 6  | [6, 7] | 2 | 2 | **No**  | Yes (wasted) |
| 8  | [8, 9] | 2 | 3 | Yes     | Yes |

In this configuration, only 2 out of every 6 rollouts truly need recomputation, but the existing check recomputes for all of them — wasting one full `_get_per_token_logps_and_entropies` forward pass per unnecessary trigger.

## What this PR changes

Replace the static modulo check with a per-rollout time-domain check:

```python
generate_every = self.args.steps_per_generation * self.num_iterations
a = self._step                                            # rollout generation step
b = ((a // generate_every) + 1) * generate_every - 1      # last consumption step
grad_accum = self.args.gradient_accumulation_steps
weights_change_mid_rollout = (a // grad_accum) != (b // grad_accum)
if weights_change_mid_rollout or (self.use_vllm and self.vllm_importance_sampling_correction):
    old_per_token_logps, _ = self._get_per_token_logps_and_entropies(...)
```

**Key insight:** `optimizer.step()` fires only at micro-step indices where `(s + 1) % grad_accum == 0`, i.e., at grad-accum bucket boundaries. The rollout's window `[a, b]` crosses such a boundary iff `(a // grad_accum) != (b // grad_accum)`. An optimizer step at exactly `s == b` is harmless (the rollout is already fully consumed by then), so the boundary case is correctly excluded.

The vLLM importance-sampling-correction branch is preserved unchanged: when the sampling distribution differs numerically from the training engine even at identical weights, a snapshot is always required.

## Equivalence in the default config

For the default path (`steps_per_generation == gradient_accumulation_steps`, `num_iterations == 1`, so `generate_every == grad_accum`):
- Every rollout starts at a multiple of `grad_accum`, and `b = a + grad_accum - 1`
- Both `a` and `b` fall in the same grad-accum bucket → `weights_change_mid_rollout == False`
- Old check: `grad_accum % generate_every == 0` → also `False`

→ Identical recomputation decisions in the common path; **no regression**.

## Net effect

| Config | Old behaviour | New behaviour |
|---|---|---|
| `grad_accum == generate_every` (default) | Use `.detach()` | Use `.detach()` (same) |
| `grad_accum` is integer multiple of `generate_every` (`k ≥ 2`) | Use `.detach()` | Use `.detach()` (same) |
| `grad_accum` is not a multiple of `generate_every` | Recompute every rollout | Recompute only when the window crosses a sync boundary |
| `use_vllm` + `vllm_importance_sampling_correction` | Always recompute | Always recompute (same) |

The optimization is correctness-preserving and only affects non-aligned configurations.

## Why also `disable_gradient_checkpointing`

Gradient checkpointing inside a `torch.no_grad()` block is a no-op for its primary memory-saving purpose (no backward graph exists), but in some setups it can still trigger checkpoint hook registration/teardown overhead. Wrapping the block in `disable_gradient_checkpointing` makes the intent explicit and avoids any related per-call overhead during these pure forward passes.

## Before submitting
- [x] Was this discussed/approved via a GitHub issue?
- [x] Did you make sure to update the documentation with your changes? (Internal optimization, no public API change.)

> **TODO before merge:** add a unit test asserting that for `spg=2, num_iter=1, grad_accum=3`, `old_per_token_logps` is computed at `_step ∈ {2, 8, 14, ...}` and skipped at `_step ∈ {0, 4, 6, 10, 12, ...}`. Happy to add in a follow-up commit on this branch.

## AI writing disclosure
- [x] AI-assisted: derivation of the per-rollout precise condition was developed with an AI assistant; the code change and review were done by a human.

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -2037,17 +2037,32 @@ class GRPOTrainer(_BaseTrainer):
         # torch.no_grad() block triggers a harmless PyTorch warning ("None of the inputs have requires_grad=True").
         # Temporarily disable checkpointing to avoid this warning during inference.
         with torch.no_grad(), disable_gradient_checkpointing(self.model, self.args.gradient_checkpointing_kwargs):
-            # If the generation and optimization steps are misaligned—i.e., if generation does not occur at the end of
-            # a full optimizer step (when gradient_accumulation_steps is not a multiple of generate_every)—then the
-            # samples may come from an earlier version of the model. In that case, we need to track old_per_token_logps
-            # for importance sampling. If the steps are aligned, importance sampling isn't necessary and we set
-            # old_per_token_logps to None.
-            # When using vLLM, we always compute old_per_token_logps for importance sampling, it was shown that the
-            # distribution mismatch between vLLM and the training model can be large and harm the training.
+            # Decide whether to compute old_per_token_logps:
+            # 1. "Time-domain old": if optimizer.step() fires *during* this rollout's consumption window
+            #    (i.e., after some micro-step strictly before the last one), the model weights change
+            #    while the rollout is being reused, so we need a logπ snapshot at sampling time.
+            # 2. "Engine-domain old": with vLLM + importance sampling correction, the vLLM sampling
+            #    distribution differs numerically from the training model's distribution even at
+            #    identical weights, so we always need the snapshot.
+            #
+            # For (1) we judge per-rollout instead of relying on the coarser
+            # `gradient_accumulation_steps % generate_every == 0` modulo check. The rollout will be
+            # consumed at micro-steps [a, b]: a = self._step at generation time, b = the last
+            # micro-step before the next generate_every boundary. An optimizer.step() fires after
+            # micro-step s iff (s + 1) % grad_accum == 0; it is harmless when s == b (the rollout is
+            # already fully consumed), so old==current iff [a, b-1] contains no such s, i.e. iff
+            # ⌊a / grad_accum⌋ == ⌊b / grad_accum⌋ (the window does not cross a sync boundary).
+            #
+            # The b formula `((a // generate_every) + 1) * generate_every - 1` reduces to
+            # `a + generate_every - 1` whenever a is aligned to a generate_every boundary (the
+            # default path), and shrinks correctly in partial-cycle cases (e.g., forks that restore
+            # _step from checkpoint while _buffered_inputs is None forces mid-cycle regeneration).
             generate_every = self.args.steps_per_generation * self.num_iterations  # generation frequency
-            if self.args.gradient_accumulation_steps % generate_every != 0 or (
-                self.use_vllm and self.vllm_importance_sampling_correction
-            ):
+            a = self._step
+            b = ((a // generate_every) + 1) * generate_every - 1
+            grad_accum = self.args.gradient_accumulation_steps
+            weights_change_mid_rollout = (a // grad_accum) != (b // grad_accum)
+            if weights_change_mid_rollout or (self.use_vllm and self.vllm_importance_sampling_correction):
                 old_per_token_logps, _ = self._get_per_token_logps_and_entropies(
                     self.model,
                     prompt_completion_ids,


qgallouedec:
closing see https://github.com/huggingface/trl/issues/5770#issuecomment-5093324023

PR #5758, opened 2026-05-12
Add telemetry to trainers

qgallouedec:
Adds a single anonymous ping per trainer instantiation via `huggingface_hub.send_telemetry`, hooked from `_BaseTrainer.__init__` so it covers every TRL trainer without per-trainer changes.

**Topic:** `trl/<TrainerClassName>` (e.g. `trl/SFTTrainer`, `trl/GRPOTrainer`)

**Payload (user-agent):** `model_arch`, `peft`, `distributed` (`deepspeed`/`fsdp`/`ddp`/`none`), `world_size`, plus `trl` version.

No user data, dataset names, model paths, or hyperparameter values are sent.

## Why

TRL currently emits no telemetry, so we have no signal on which trainers, model families, or distributed backends are actually used. This blocks prioritization decisions.

## Docs

Short opt-out note added to `docs/source/installation.md`.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new outbound network call (`huggingface_hub.send_telemetry`) on every TRL trainer initialization, which could affect privacy expectations and environments with restricted networking if opt-out isn’t set.
> 
> **Overview**
> Adds anonymous usage telemetry across all TRL trainers by hooking a single `send_telemetry` ping into `_BaseTrainer.__init__`, emitted only on the main process and skipped in CI/offline/telemetry-disabled modes.
> 
> The ping reports **bucketed** runtime metadata (TRL version, allowlisted trainer name/model architecture, PEFT usage, distributed backend, world-size bucket, device type, and GPU model when available) and introduces new docs (`usage_stats.md`) plus a toctree entry explaining what’s collected and how to opt out.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 514a7ae2a65089e933ec297599a4cfa16ec09b58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/trainer/base_trainer.py
+++ b/trl/trainer/base_trainer.py
@@ -51,12 +51,16 @@ class _BaseTrainer(Trainer):
         else:
             distributed = "none"
         device = self.accelerator.device.type
-        # Each accelerator backend (cuda, xpu, npu, mlu, …) exposes its own `torch.<device>.get_device_name`. MPS and
-        # CPU do not, hence the broad except.
-        try:
-            gpu = getattr(torch, device).get_device_name(0)
-        except (AttributeError, RuntimeError):
-            gpu = ""
+        if device == "cuda":
+            gpu = torch.cuda.get_device_name(0)
+        elif device == "xpu":
+            gpu = torch.xpu.get_device_name(0)
+        elif device == "npu":
+            gpu = torch.npu.get_device_name(0)
+        elif device == "mlu":
+            gpu = torch.mlu.get_device_name(0)
+        else:
+            gpu = "other"
         # Bucketed to avoid fingerprinting individual deployments by their exact cluster size.
         n = self.accelerator.num_processes
         world_size = "1" if n == 1 else "2-8" if n <= 8 else "9-64" if n <= 64 else "65+"
--- a/trl/trainer/base_trainer.py
+++ b/trl/trainer/base_trainer.py
@@ -14,8 +14,11 @@
 
 import os
 
+from accelerate.utils import is_peft_model
+from huggingface_hub.utils import send_telemetry
 from transformers import Trainer, is_wandb_available
 
+from .. import __version__
 from .utils import generate_model_card, get_comet_experiment_url, get_config_model_id, get_trackio_space_url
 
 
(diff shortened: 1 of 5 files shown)


trl/trainer/base_trainer.py
```diff
@@ -29,6 +33,44 @@ class _BaseTrainer(Trainer):
...
+        else:
+            distributed = "none"
+        device = self.accelerator.device.type
+        # Each accelerator backend (cuda, xpu, npu, mlu, …) exposes its own `torch.<device>.get_device_name`. MPS and
+        # CPU do not, hence the broad except.
+        try:
+            gpu = getattr(torch, device).get_device_name(0)
+        except (AttributeError, RuntimeError):
+            gpu = ""
+        send_telemetry(
```
AmineDiro, line 74: I think we should do this in a separate thread and probably async just to have minimal footprint

AmineDiro (commented):
Great to have this in the base_trainer 👍🏼 

Just a comment on where we should send telemetry. 

We'll probably need continuous telemetry for monitoring usage not one off send. So maybe we'll need a separate thread with async loop

trl/trainer/base_trainer.py
qgallouedec: it's already how `send_telemetry` works https://huggingface.co/docs/huggingface_hub/package_reference/utilities#huggingface_hub.utils.send_telemetry

> Telemetry collection is run in a separate thread to minimize impact for the user.

AmineDiro: but we spawn a thread each time we send the data 🤔 ?

qgallouedec: Yes, but we report once: at init

AmineDiro: yes, to clarify was just a comment for the future structure. We will probably need to report more than once (for example, monitoring exception to report), so that's why I asked if we should put it in a separate thread. But we can do that once we need to 👍🏼

AmineDiro (approved):

trl/trainer/base_trainer.py
qgallouedec: Ah ok I see!
I think we should be careful about the amount of data we get from the telemetry. But if we decide to get more data at some point, yes we should make sure to do it efficiently

PR #5765, opened 2026-05-13
feat(grpo): replace deprecated `use_transformers_paged` with transformers continuous batching

sergiopaniego:
# What does this PR do?

Replaces the deprecated `use_transformers_paged` with proper transformers continuous batching support + a new training example script. The old branch set `logprobs = None`, silently bypassing importance-sampling correction. This PR captures logprobs from `output.logprobs` and exposes `ContinuousBatchingConfig` for KV cache tuning.

Existing `use_transformers_paged=True` configs continue to work and forward to the new flag with a `FutureWarning`.

Requires `transformers>=5.8.0` (two training-mode bugs in `generate_batch()` fixed in huggingface/transformers#45943). The 5.8.0 could actually need an update if these changes land in a follow-up version.

## Usage

```python
GRPOConfig(
    ...
    use_transformers_continuous_batching=True,
    transformers_continuous_batching_config={
        "use_cuda_graph": False,
        "max_memory_percent": 0.4,  # leave ~60% free for training
    },
    ...
)
```

Use CB over the default `generate()` when N≥32 with variable completion lengths (e.g. math reasoning). Use vLLM when maximum throughput or multi-GPU tensor parallelism is the priority.

## Performance

Benchmark (A100 80GB, Llama-3.2-1B-Instruct). Full script: https://gist.github.com/sergiopaniego/740a9708289e8f64cacd0d087d17d162

| Scenario | Gens | Max tokens | Default | CB | Speedup | VRAM delta |
|---|---|---|---|---|---|---|
| GSM8K | 8 | 2048 | 18.24 ± 12.81 s | 18.16 ± 8.12 s | 1.00x | +17.86 GB |
| GSM8K | 32 | 2048 | 37.06 ± 8.63 s | 29.91 ± 10.09 s | **1.24x** | +7.39 GB |
| GSM8K | 64 | 2048 | 41.12 ± 12.69 s | 32.84 ± 7.00 s | **1.25x** | **-16.66 GB** |

CB pulls ahead at N≥32 with variable completion lengths. At N=64 the VRAM delta inverts: default `generate()` eagerly allocates KV cache for all 64×2048 sequences (~40 GB) while CB pre-allocates a fixed fraction of free VRAM (~25 GB).

cb_benchmark_results

`RLOOTrainer` also supported

## Before submitting

## AI writing disclosure

## Who can review?

@qgallouedec @AmineDiro @kashif @remi-or 

---

> [!NOTE]
> **Medium Risk**
> Changes the online RL generation path and training VRAM behavior for users switching from paged generation; gated on a new transformers version and blocks VLMs on this path.
> 
> **Overview**
> Replaces deprecated **`use_transformers_paged`** with **`use_transformers_continuous_batching`** and **`transformers_continuous_batching_config`** on `GRPOConfig` / `RLOOConfig`. Legacy `use_transformers_paged=True` still works via a `FutureWarning` and auto-enables the new flag.
> 
> **`GRPOTrainer`** and **`RLOOTrainer`** now call `generate_batch(..., continuous_batching_config=...)`, require **transformers≥5.8.0**, default **`max_memory_percent` to 0.5** for KV cache headroom during backward, and reject multimodal processors when continuous batching is on.
> 
> Adds a **GSM8K example** (`grpo_continuous_batching.py`), **GRPO docs** for the in-process alternative to vLLM, **contract tests** for the transformers CB API, and renames the GPU integration test from paged to continuous batching.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86b33bd2c3b1ee6ab391fbcd23ddb577f8fb70d0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kashif (approved):

trl/trainer/grpo_trainer.py
```diff
@@ -557,7 +562,23 @@ def __init__(
...
         self.repetition_penalty = args.repetition_penalty
-        self.use_transformers_paged = args.use_transformers_paged
+        self.use_transformers_continuous_batching = args.use_transformers_continuous_batching
+        if self.use_transformers_continuous_batching:
+            if not Version(transformers.__version__) >= Version("5.8.0"):
+                raise ImportError(
+                    "Using `use_transformers_continuous_batching` requires transformers>=5.8.0. "
+                    "Please upgrade with `pip install --upgrade transformers`."
+                )
+            from transformers.generation import ContinuousBatchingConfig
```
AmineDiro, line 585: inlined import maybe we can move it ?

trl/trainer/grpo_trainer.py
```diff
@@ -557,7 +562,23 @@ def __init__(
...
+
+            cb_kwargs = dict(args.transformers_continuous_batching_config or {})
```
AmineDiro, line 587: I wonder if there are some good defaults that we can set ? @sergiopaniego did you test different config to get a sense of throughput, latency, memory ?

trl/trainer/grpo_trainer.py
AmineDiro, line 1430: I wonder if this `unwrap_model_for_generation` yields a correct model in all cases (fsdp, deepspeed, tp etc ) for the continuous batching ? Maybe a small test with fsdp can tell us if it's robust.

AmineDiro (commented):
Great work 👏🏼 

Just wondering if the unwrapped model passed to continous batching  still works with fsdp etc ..

trl/trainer/grpo_trainer.py
remi-or: I would be interested in this as well! 
Usually if no config is given, continuous batching adapts to the situation and sets smarts default values for the parameters itself, but if you see something different please let me know!
BTW, compile is turned off by default, because it adds a lot of warmup overhead. But if you re-use the continuous batching manager for several generations (you have to set `persistent_manager=True` so you don't need to re-setup the CB manager, and it is `False` by default) it might be worth it. Lmk if you have any question!

trl/trainer/grpo_trainer.py
sergiopaniego: It's intentional since it was introduced in transformers 5.4.0

trl/trainer/grpo_trainer.py
sergiopaniego: thanks! I've tested it and once it does yield the correct model, it's surfaced another bug caused by using `with torch.inference_mode():` so I've updated it

trl/trainer/grpo_trainer.py
sergiopaniego: `max_memory_percent=0.5` is the only parameter TRL overrides (transformers defaults to 0.9). The lower value reserves more VRAM headroom for the backward pass, which matters on larger models. The rest (allow_block_sharing, block_size) inherit transformers defaults directly. Running some benchmarks with the latest changes:

cb_config_sweep

sergiopaniego:
The `transformers` version needs to be updated from `5.8.0` once https://github.com/huggingface/transformers/pull/45943 (already merged) lands in a new version

qgallouedec:
Thanks for the PR. High-level for now:

+1 on retiring `use_transformers_paged`.

On CB itself: the part that's a bit of a shame is that in sync GRPO the batch can't span training steps — CB's intra-batch refill helps (hence the 1.24× in your bench), but the bigger win, keeping the batch continuously full across the entire training run, is left on the table. That's the win async GRPO would unlock; I'm prototyping it in #5781.

Not a blocker: sync+CB still has a clear niche, and the speedup is real. But I'd love your thoughts on framing: for a user purely chasing performance, what's the sync+CB sweet spot vs. async+CB?

sergiopaniego:
> for a user purely chasing performance, what's the sync+CB sweet spot vs. async+CB?

I think we'd need to run some small benchmarks 🤔 but theoretically, async should win if we consider pure performance

qgallouedec (approved):
nice work @sergiopaniego

PR #5780, opened 2026-05-15
Drop unjustified `model.visual.` skip in GRPO / RLOO Qwen2.5-VL tests

qgallouedec:
Five tests parametrized only on `tiny-Qwen2_5_VLForConditionalGeneration` were skipping every parameter under `model.visual.` from the "params changed during training" assertion, with the comment:

> *Because of the way the tiny models are initialized, the gradient does not flow properly through the vision parts of the model, so we skip them.*

Empirically false: across 51 optimizer steps of the existing test config, **every** `model.visual.*` parameter (patch_embed, blocks, merger) updates. The skip was masking a real assertion for nothing.

---

> [!NOTE]
> **Low Risk**
> Test-only change that strengthens assertions around parameter updates during VLM training; primary risk is increased CI flakiness if upstream model behavior changes.
> 
> **Overview**
> Removes overly broad vision-parameter exclusions in GRPO/RLOO VLM training tests so the “parameters changed during training” assertion now covers `model.visual.*` (and other previously skipped vision modules) for `tiny-Qwen2_5_VLForConditionalGeneration`.
> 
> Keeps a narrow skip only for `model.vision_tower.*` in the general VLM training test, making the suite more likely to catch regressions where vision weights fail to update.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1be7105974c1afa4ad94ec87bec1a3b118c64119. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -1985,13 +1985,8 @@ class TestGRPOTrainer(TrlTestCase):
         assert trainer.state.log_history[-1]["train_loss"] is not None
 
         # Check that the params have changed
-        # Because of the way the tiny models are initialized, the gradient does not flow properly through the
-        # vision parts of the model, so we skip them. Ideally, we should fix the init of these models.
         params_to_skip = (
             "model.vision_tower.",
-            "model.multi_modal_projector.",
-            "model.visual.",
-            "model.image_newline",
         )
         for n, param in previous_trainable_params.items():
             if n.startswith(params_to_skip):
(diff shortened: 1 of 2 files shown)


albertvillanova (approved):
Thanks.

Just wondering why these were first added. Maybe that made sense for a previously supported transformers version? 🤔

qgallouedec:
> Just wondering why these were first added. Maybe that made sense for a previously supported transformers version?

I can't remember tbh, maybe the first version of these tiny models were not properly built, and the issue was fixed in the meantime. We would have known if we had a `xfail` instead of a `skip`

Issue #3858, opened 2025-08-06
when use GRPO+ deepspeed_zero3 + ds3_gather_for_generation=False, stuck in the training stage, step is still 0 after an hour


nstl-zyb:
### Reproduction

```python
    training_args = GRPOConfig(
        # use_vllm = True, # use vLLM for fast inference!
        # vllm_mode="colocate",
        # vllm_tensor_parallel_size=8,
        # vllm_server_base_url='http://127.0.0.1:8000',
        learning_rate = 5e-6,
        adam_beta1 = 0.9,
        adam_beta2 = 0.99,
        weight_decay = 0.1,
        warmup_ratio = 0.1,
        lr_scheduler_type = "cosine",
        optim = "paged_adamw_8bit",
        logging_steps = 1,
        bf16=True,
        # bf16=is_bfloat16_supported(),
        # fp16=not is_bfloat16_supported(),
        per_device_train_batch_size = 1,
        gradient_accumulation_steps = 1, # Increase to 4 for smoother training
        num_generations = 2, # Decrease if out of memory
        max_prompt_length = max_seq_length,
        max_completion_length = max_completion_length,
        # num_train_epochs = 1, # Set to 1 for a full training run
        max_steps = 250,
        save_steps = 250,
        max_grad_norm = 0.1,
        report_to = "none", # Can use Weights & Biases
        output_dir = save_path,
        # deepspeed="./config_file/deepspeed/ds_z3_offload_config.json",
        auto_find_batch_size=False,
        ds3_gather_for_generation=False
    )
    trainer = GRPOTrainer(
        model = model_path,
        # processing_class = tokenizer,
        reward_funcs =  reward_json_and_answer,
        args = training_args,
        train_dataset = grpo_dataset,
    )
    trainer.train()

```

```deepspeed_zero3
compute_environment: LOCAL_MACHINE
debug: false
deepspeed_config:
  deepspeed_multinode_launcher: standard
  offload_optimizer_device: cpu
  offload_param_device: none
  zero3_init_flag: true
  zero3_save_16bit_model: true
  zero_stage: 3
  activation_checkpointing:
    partition_activations: true
  bf16:
    enabled: true
    loss_scale: 0
    loss_scale_window: 1000
distributed_type: DEEPSPEED
downcast_bf16: 'no'
machine_rank: 0
main_training_function: main
mixed_precision: bf16
num_machines: 1
num_processes: 8
rdzv_backend: static
same_network: true
tpu_env: []
[...]

Issue #3678, opened 2025-07-02
OOM when training qwen3-0.6B model by GRPO trainer on 20G VRAM GPU


wa008:
## Problem
OOM appear when training qwen3-0.6B model by GRPO trainer with max-prompt=256 and max-completions=256 on 20G VRAM GPU, possible main reason is that model-forward and backward consumed too much memory, using flash-attention may reduce consumed memory, I haven't try it yet. 

I'm not sure whether this's reasonable. 

## not related with old verison 
I test it on 0.14.0, 0.16.0, 0.17.0 and 0.19.0 version, it shows that every version didn't work well

## Possible main reason
I found that the most consuming GPU point is 
```
logits = model(
                input_ids=input_ids_batch,
                attention_mask=attention_mask_batch,
                logits_to_keep=logits_to_keep + 1,
            ).logits
```
[code](https://github.com/huggingface/trl/blob/e04f7eb3b993b5278733a793853ed3a85d1c69c8/trl/trainer/grpo_trainer.py#L879)

and, model init by `model = AutoModelForCausalLM.from_pretrained(model, **model_init_kwargs)`

when doing model forward, consumed memory increased from 8B into 20G, which result in OOM in the next training step, so, the main reason why trl GRPO consume so much memory is related with model-forward(with backward) function rather than trl repository. 

## Reproduce 
model: Qwen3-0.6B
GPU: 20G memory
data: longest input is 168

full data: 
```
[
	{
        "prompt": [
            {
                "role": "user",
                "content": "what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning?what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning? what is machine learning?"
            }
        ]
    },
    {
        "prompt": [
            {
                "role": "user",
                "content": "why sky is blue?"
            }
        ]
    }
]
```

code:
```
from datasets import load_dataset
from trl import GRPOConfig, GRPOTrainer
import os

# dataset = load_dataset("/root/.cache/modelscope/hub/datasets/okwinds/tldr/data", split="train")
[...]

Issue #3172, opened 2025-03-28
Training GRPO (Qwen/Qwen2.5-7B) with LoRA on 8XH100 deepspeed zero3 gives OOM


thesillystudent:
### Reproduction

```
CUDA_VISIBLE_DEVICES=0,1,2,3 trl vllm-serve --model Qwen/Qwen2.5-7B --tensor_parallel_size 4

CUDA_VISIBLE_DEVICES=4,5,6,7 HF_HUB_ENABLE_HF_TRANSFER=1 ACCELERATE_USE_FSDP=false accelerate launch --config_file test_grpo_ds.yaml 3094.py
```

```python
import torch
from datasets import load_dataset
from trl import GRPOTrainer, GRPOConfig, get_peft_config
from peft import LoraConfig, get_peft_model
from transformers import AutoTokenizer, AutoModelForCausalLM

dataset = load_dataset("trl-lib/tldr", split="train")


# Dummy reward function: count the number of unique characters in the completions
def reward_num_unique_chars(completions, **kwargs):
    return [len(set(c)) for c in completions]

model_name = "Qwen/Qwen2.5-7B"


lora_config = LoraConfig(
    r=16,
    lora_alpha=32,
    target_modules=["q_proj", "k_proj", "v_proj", "o_proj", "up_proj", "down_proj", "gate_proj"],
    task_type="CAUSAL_LM",
    lora_dropout=0.05
)

model = AutoModelForCausalLM.from_pretrained(
    model_name,
    torch_dtype=torch.bfloat16,
    attn_implementation="flash_attention_2",
    device_map=None
).to("cuda")

tokenizer = AutoTokenizer.from_pretrained(model_name)
tokenizer.pad_token = tokenizer.eos_token

training_args = GRPOConfig(
    output_dir="3094-compiled",
    use_vllm=True,
    bf16=True,
    gradient_checkpointing=True,
    logging_steps=10,
    torch_compile=True,
)

trainer = GRPOTrainer(
    model=get_peft_model(model, lora_config),
    processing_class=tokenizer,
    args=training_args,
    reward_funcs=reward_num_unique_chars,
    train_dataset=dataset,
)
trainer.train()

```

```
compute_environment: LOCAL_MACHINE
debug: false
deepspeed_config:
  deepspeed_multinode_launcher: standard
  offload_optimizer_device: none
  offload_param_device: none
  zero3_init_flag: true
  zero3_save_16bit_model: true
  zero_stage: 3
[...]

PR #5794, opened 2026-05-20
docs: document expandable_segments allocator config for memory tuning

akshansh47:
# What does this PR do?

Adds a new section to `docs/source/reducing_memory_usage.md` documenting the PyTorch caching-allocator environment variables that materially reduce OOM rates on long-context and online-RL training, but which are currently not surfaced in the central memory-reduction guide.

The new section covers:

- `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` (and the `PYTORCH_ALLOC_CONF` canonical name in PyTorch >= 2.11) — typically reclaims 2–5 GB/GPU on long-context training (especially online RL methods like GRPO / RLOO / Online DPO, where rollout, log-prob, and update phases allocate at different sizes within each iteration), at no measurable throughput cost.
- vLLM `CuMemAllocator` incompatibility warning — avoids a class of startup-time crashes when `use_vllm=True`.
- Optional `garbage_collection_threshold:0.85` for chunked-kernel fragmentation cases.

The variable is referenced exactly once in the TRL docs today (in `docs/source/lora_without_regret.md`, as a recipe-specific aside) but is missing from the central memory-reduction guide that users land on when searching for OOM mitigations. This addition complements existing advice (gradient checkpointing, activation offloading, `ds3_gather_for_generation=False`, etc.) rather than replacing any of it; same class of advice as the runtime-level sections already in the doc.

This advice has been most-helpfully called out in user-reported OOM threads such as #3678, #3172, and #3858.

Fixes # (no specific issue — docs improvement)

## Before submitting

## AI writing disclosure

## Who can review?

@qgallouedec @lewtun

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Low Risk**
> Documentation-only change adding guidance on PyTorch allocator environment variables; no runtime or API behavior is modified.
> 
> **Overview**
> Adds a new **PyTorch caching allocator** section to `docs/source/reducing_memory_usage.md` recommending `expandable_segments:True` (via both `PYTORCH_CUDA_ALLOC_CONF` and the newer `PYTORCH_ALLOC_CONF`) to reduce fragmentation/OOMs in long-context and online-RL training.
> 
> Documents a *vLLM `CuMemAllocator` incompatibility* warning and an optional `garbage_collection_threshold` tweak for persistent fragmentation, linking to upstream PyTorch environment variable docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e68ef8267a45c1f11db8ff8027dd007cd484706. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/reducing_memory_usage.md
+++ b/docs/source/reducing_memory_usage.md
@@ -270,6 +270,30 @@ training_args = RewardConfig(..., pad_to_multiple_of=2048)
 </hfoption>
 </hfoptions>
 
+## PyTorch caching allocator
+
+PyTorch's caching allocator can fragment over long training runs — separate caches for forward activations, optimizer states, KL reference logits, and FlashAttention workspace each grow and shrink at different rates, leaving holes that cannot satisfy a new large allocation even though aggregate free memory is high. This is especially common in online RL methods (GRPO, RLOO, Online DPO) where the rollout, log-prob, and update steps allocate at different sizes within each iteration.
+
+Setting `expandable_segments:True` lets the allocator grow existing segments instead of fragmenting into fresh fixed blocks, typically reclaiming 2–5 GB per GPU on long-context training with no measurable throughput cost.
+
+```bash
+export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True
+export PYTORCH_ALLOC_CONF=expandable_segments:True   # canonical name in PyTorch >= 2.11
+```
+
+Set both names: `PYTORCH_ALLOC_CONF` is the canonical knob in PyTorch ≥ 2.11; older versions only read the legacy `PYTORCH_CUDA_ALLOC_CONF` alias.
+
+> [!WARNING]
+> `expandable_segments:True` is **incompatible with vLLM's `CuMemAllocator`**. If you use vLLM as the generation backend (e.g. `use_vllm=True` in [`GRPOConfig`]), clear the variable in the worker environment before vLLM initializes, or vLLM will assert on startup.
+
+For runs where fragmentation persists despite this knob (typically chunked-LM-head and other custom kernels that allocate transient ~GB tensors), append `garbage_collection_threshold:0.85` to trigger the allocator's defragmentation pass earlier:
+
+```bash
+export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,garbage_collection_threshold:0.85
+```
+
+This is a defragmentation-only knob; it does not change peak memory or training dynamics. See the [PyTorch CUDA semantics docs](https://pytorch.org/docs/stable/notes/cuda.html#environment-variables) for the full list of options.
+
 ## Disabling model gathering for generation in online methods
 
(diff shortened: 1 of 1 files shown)


qgallouedec:
Thanks @akshansh47, this is worth documenting, and it matches what we already do ourselves: `.github/workflows/tests.yml` sets both `PYTORCH_CUDA_ALLOC_CONF` and `PYTORCH_ALLOC_CONF` to `expandable_segments:True`.

Two things to fix first.

1. The vLLM warning is too broad: `CuMemAllocator` is only built when sleep mode is on (every `get_instance()` call site in `vllm/v1/worker/gpu_worker.py` sits behind `enable_sleep_mode`) so plain `use_vllm=True` is unaffected, and as written the warning tells users that a default-safe setup will crash on startup. Please scope it to `vllm_enable_sleep_mode=True` and cross-link the existing "vLLM sleep mode" section rather than leaving the warning 47 lines above it.
2. Worth adding there too: vLLM's assert reads only `PYTORCH_CUDA_ALLOC_CONF`, so setting just `PYTORCH_ALLOC_CONF` keeps expandable segments on while silently bypassing the guard. Finally, please drop or source the "2–5 GB per GPU" and "no measurable throughput cost" figures, and double-check the "PyTorch >= 2.11" boundary for the rename.

Thanks again for the contrib, sorry for the time needed before we reviewed it

akshansh47:
Thanks for the review @qgallouedec — all four points addressed in 125ed04:

1. **Warning scoped to sleep mode**: it now states the conflict only applies with `vllm_enable_sleep_mode=True` (plain `use_vllm=True` unaffected) and cross-links the [vLLM sleep mode](https://github.com/huggingface/trl/blob/main/docs/source/reducing_memory_usage.md#vllm-sleep-mode) section. I also corrected the mechanism while I was there: current vLLM doesn't assert — `CuMemAllocator` detects the setting and temporarily disables expandable segments while its pool is active (`vllm/device_allocator/cumem.py`), so the practical hazard is the bypass in point 2 rather than a startup crash.
2. **Alias-bypass note added**: both in the prose (general form — vLLM and possibly other libraries inspect only `PYTORCH_CUDA_ALLOC_CONF`) and in the warning (setting only `PYTORCH_ALLOC_CONF` bypasses vLLM's detection, so prefer the legacy name or both when sleep mode is on).
3. **Figures dropped**: replaced the 2–5 GB / no-throughput-cost claims with a qualitative statement.
4. **Boundary corrected to 2.10**: verified against the docs — [2.10+](https://docs.pytorch.org/docs/stable/cuda_environment_variables.html) lists `PYTORCH_ALLOC_CONF` as primary with `PYTORCH_CUDA_ALLOC_CONF` as backward-compat alias, while the [2.9 table](https://docs.pytorch.org/docs/2.9/cuda_environment_variables.html) only has `PYTORCH_CUDA_ALLOC_CONF`.

Issue #5613, opened 2026-04-21
Feature Request: Allow passing dataset-provided sample weights to DPOTrainer


sirano1004:
### Feature request

I propose adding support for an optional dataset column (e.g., sample_weight) to DPOTrainer that scales the unreduced per-example DPO loss before batch aggregation. Currently, DPOTrainer supports trainer-computed weighting (use_weighting=True), but cannot consume a precomputed per-example weight column from the dataset.

**Proposed Solution:**
 1. Allow DPOConfig to accept an optional sample_weight_key (defaulting to None).
 2. If present in the dataset, multiply the unreduced per-example loss by this value before reduction.
 3. Restrict support to loss types where per-example weighting is well-defined (e.g., disable for aot / aot_unpaired).

**Example usage:**
```python
training_args = DPOConfig(
    ...,
    sample_weight_key="pair_weight",
)

trainer = DPOTrainer(
    model=model,
    ref_model=ref_model,
    args=training_args,
    train_dataset=train_dataset,  # includes "pair_weight"
)

```

### Motivation

This provides a general mechanism for offline preference optimization workflows where different preference pairs should contribute differently to training. Examples include confidence-based weighting, curriculum schemes, filtered/relabeled preference data, and research methods like [Margin-Adaptive DPO (MADPO)](https://openreview.net/pdf?id=nkulfhphPr) that rely on precomputed per-pair weights.

This keeps weighting logic outside the trainer, making it highly reusable across different methods without adding algorithm-specific trainers. It aligns perfectly with DPOTrainer's existing pattern of consuming optional dataset fields like precomputed reference log probabilities.

### Your contribution

I am happy to implement this and submit a PR if the maintainers agree with this dire ill ensure it follows the guidelines in CONTRIBUTING.md.

qgallouedec:
Hey @sirano1004, thanks for the proposal and sorry for the late reply. The change itself is small and slots cleanly next to the existing `use_weighting` hook, with carve-outs needed for aot/aot_unpaired and the Liger path.

I'd like to hold on merging for a bit though. The concrete consumer today is your MADPO paper; the broader use cases (confidence/curriculum/filtered data) are plausible but hypothetical so far. I'll open a draft PR (#5804) with the implementation so anyone interested can try it out; if a few users chime in here with a +1 or a use case, happy to revisit and merge.

Not a rejection, just waiting for signal. Thanks again.

sirano1004:
[...]

PR #5804, opened 2026-05-21
MADPO

qgallouedec:
Adds optional per-pair sample weighting to `DPOTrainer`: if the dataset contains a `"weight"` column, the per-pair loss is scaled by it before reduction.

Proposed in #5613. Opening as a draft so users with a concrete need can try it; **not planning to merge until we see more community signal** (additional +1s or use cases on the issue).


qgallouedec:
It's been a while and no community signal, I'm going to close the PR, but the branch remains accessible if necessary

PR #5816, opened 2026-05-22
Align KTO with DPO: Align _compute_loss_liger flow

albertvillanova:
Align KTO with DPO: Align `_compute_loss_liger` flow.

This PR refactors the Liger kernel loss computation in the `KTOTrainer` class to streamline metric tracking, clarify input handling, and simplify the interface. The most significant changes focus on how the Liger loss is computed and integrated, removing the previous dictionary-based output in favor of direct metric accumulation and returning only the loss value.

Part of:
- #4786

### Solution
- Store `self.use_liger_kernel = args.use_liger_kernel` as an instance attribute.
- Rewrite `_compute_loss_liger` as a fully standalone method: new signature `(model, inputs, return_outputs)`, appends directly to `self._metrics[mode]`, returns a scalar loss. 
  - Previously it returned a dict and was called from inside `_compute_loss`.
- Remove the `if self.args.use_liger_kernel` dispatch branch from `_compute_loss`, leaving it as a pure non-liger path.
- Update `compute_loss` to dispatch between the two paths: `if self.use_liger_kernel: return self._compute_loss_liger(...) else: return self._compute_loss(...)`.

### Changes

**Liger Kernel Loss Refactoring and Integration:**

* The `_compute_loss_liger` method is rewritten to:
  - Accept `inputs` and a `return_outputs` flag, raising an error if outputs are requested (since logits are not materialized by the Liger loss).
  - Directly accumulate metrics (such as KL, rewards, and log probabilities) into the trainer's internal `_metrics` dictionary, instead of returning them in a dictionary.
  - Return only the scalar loss value, not a dictionary of outputs.

* The main `_compute_loss` method is updated to:
  - Remove the branching logic that handled dictionary outputs from the Liger loss; now, only the standard loss path remains for non-Liger cases.

**Configuration and Input Handling:**

* The `use_liger_kernel` flag is now stored as an instance variable for consistent access, and the kernel import logic is updated accordingly.
* The `compute_loss` method now directly dispatches to `_compute_loss_liger` if the Liger kernel is enabled, passing through the `return_outputs` flag and enforcing the new interface.

---

> [!NOTE]
> **Medium Risk**
> Refactors the Liger-kernel training loss path and metric collection logic; while mostly structural, it affects core training/eval loss computation and logging behavior.
> 
> **Overview**
> Refactors the Liger-kernel KTO loss integration so `compute_loss` dispatches directly to `_compute_loss_liger(model, inputs, return_outputs)`, while `_compute_loss` becomes a single non-Liger path (removing the previous in-method branch and dict-based return handling).
> 
> `_compute_loss_liger` now returns only the scalar loss, explicitly errors on `return_outputs=True`, and records KL/reward/logp/logit metrics by appending directly into `self._metrics` (including auxiliary loss being added into the returned loss), bringing its flow in line with the standard KTO loss computation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a222f282e6d72e3251ca7f5d84c7405b807695b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/kto/kto_trainer.py
+++ b/trl/experimental/kto/kto_trainer.py
@@ -1052,73 +1052,60 @@ class KTOTrainer(_BaseTrainer):
         num_chosen = labels.sum().to(self.accelerator.device)
         num_rejected = (len(labels) - num_chosen).to(self.accelerator.device)
 
-        if self.args.use_liger_kernel:
-            model_output = self._compute_loss_liger(model, batch)
-            losses = model_output["loss"]
-            policy_chosen_logits = model_output["chosen_logits_sum"]
-            policy_rejected_logits = model_output["rejected_logits_sum"]
-            policy_chosen_logps = model_output["chosen_logps_sum"]
-            policy_rejected_logps = model_output["rejected_logps_sum"]
-            chosen_rewards = model_output["chosen_rewards_sum"]
-            rejected_rewards = model_output["rejected_rewards_sum"]
-            kl = model_output["kl"]
-            if self.aux_loss_enabled:
-                aux_loss = model_output["aux_loss"]
-        else:
-            forward_output = self.forward(model, batch)
-            (
-                policy_chosen_logps,
-                policy_rejected_logps,
-                policy_chosen_logits,
-                policy_rejected_logits,
-                policy_KL_logps,
-            ) = forward_output[:5]
-            if self.aux_loss_enabled:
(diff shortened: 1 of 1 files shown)


qgallouedec (approved):
thanks

PR #5826, opened 2026-05-24
Fix NaN loss when completions are fully truncated

### LINKED ISSUE #5662: TPOTrainer.evaluate() returns NaN eval_loss while training loss is finite
### Reproduction

# TPOTrainer.evaluate() returns NaN eval_loss while training loss is finite

## Description

`TPOTrainer.evaluate()` returns `eval_loss: nan` even though training runs normally and `train_loss` is finite.

I tested the experimental `TPOTrainer` from `trl.experimental.tpo` using the public dataset `tpo-alignment/triple-preference-ultrafeedback-40K` and the model `Qwen/Qwen3-0.6B`.

The dataset has the expected TPO columns:

```python
['prompt', 'reference', 'chosen', 'rejected']
```

I converted the dataset to conversational/chat format before passing it to `TPOTrainer`:

```python
{
    "prompt": [
        {"role": "user", "content": prompt}
    ],
    "chosen": [
        {"role": "assistant", "content": chosen}
    ],
    "rejected": [
        {"role": "assistant", "content": rejected}
    ],
    "reference": [
        {"role": "assistant", "content": reference}
    ],
}
```

Training works and produces finite losses, but `trainer.evaluate()` returns `nan` both before and after training.

## Reproduction

```python
import os
os.environ["TRL_EXPERIMENTAL_SILENCE"] = "1"
os.environ["TOKENIZERS_PARALLELISM"] = "false"

import torch
from peft import LoraConfig
from trl.experimental.tpo import TPOConfig, TPOTrainer

MODEL_NAME = "Qwen/Qwen3-0.6B"
DATASET_NAME = "tpo-alignment/triple-preference-ultrafeedback-40K"

TRAIN_SAMPLES = 1000
EVAL_SAMPLES = 100
MAX_LENGTH = 384
MAX_STEPS = 100

raw_train = load_dataset(DATASET_NAME, split="train")
raw_eval = load_dataset(DATASET_NAME, split="test")

raw_train = raw_train.select(range(TRAIN_SAMPLES))
raw_eval = raw_eval.select(range(EVAL_SAMPLES))

print("Raw train columns:", raw_train.column_names)
print("Raw eval columns:", raw_eval.column_names)
print("Raw train size:", len(raw_train))
print("Raw eval size:", len(raw_eval))


def clean_str(x):
    if x is None:
        return ""
    return str(x).strip()


def to_chat_format(example):
    return {
        "prompt": [
            {"role": "user", "content": clean_str(example["prompt"])}
        ],
        "chosen": [
            {"role": "assistant", "content": clean_str(example["chosen"])}
        ],
        "rejected": [
            {"role": "assistant", "content": clean_str(example["rejected"])}
        ],
        "reference": [
            {"role": "assistant", "content": clean_str(example["reference"])}
        ],
    }


train_ds = raw_train.map(
    to_chat_format,
    remove_columns=raw_train.column_names,
)

eval_ds = raw_eval.map(
    to_chat_format,
    remove_columns=raw_eval.column_names,
)

print("Processed columns:", train_ds.column_names)
print("Example prompt:", train_ds[0]["prompt"])

    MODEL_NAME,
    trust_remote_code=True,
    use_fast=True,
)

tokenizer.padding_side = "right"

print("pad_token:", tokenizer.pad_token)
print("eos_token:", tokenizer.eos_token)

peft_config = LoraConfig(
    r=8,
    lora_alpha=16,
    lora_dropout=0.05,
    bias="none",
    task_type="CAUSAL_LM",
    target_modules="all-linear",
)

args = TPOConfig(
    output_dir="Qwen3-0.6B-TPO-LoRA-fast-test",

    per_device_eval_batch_size=1,
    gradient_accumulation_steps=4,
    gradient_checkpointing=True,

    max_steps=MAX_STEPS,
    learning_rate=1e-5,
    warmup_steps=10,
    lr_scheduler_type="cosine",

    max_length=MAX_LENGTH,
    loss_type="sigmoid",
    beta=0.01,
    tpo_alpha=1.0,

    fp16=True,
    bf16=False,

    logging_steps=5,
    eval_strategy="steps",
    eval_steps=25,

    save_strategy="no",
    remove_unused_columns=False,
    report_to="none",
    dataloader_num_workers=2,
)

trainer = TPOTrainer(
    model=MODEL_NAME,
    args=args,
    train_dataset=train_ds,
    eval_dataset=eval_ds,
    processing_class=tokenizer,
    peft_config=peft_config,
)

print("Eval before training:")
print(trainer.evaluate())

train_result = trainer.train()
print(train_result.metrics)

print("Eval after training:")
print(trainer.evaluate())
```

outputs:

```text
================================================================================
Torch: 2.10.0+cu128
CUDA: True
GPU: Tesla T4
================================================================================
Raw train columns: ['prompt', 'reference', 'chosen', 'rejected']
Raw eval columns: ['prompt', 'reference', 'chosen', 'rejected']
Raw train size: 40000
Raw eval size: 1910
Fast train size: 1000
Fast eval size: 100
Converting train dataset: 100% 1000/1000
Converting eval dataset: 100% 100/100
Processed columns: ['prompt', 'reference', 'chosen', 'rejected']
Example prompt: [{'role': 'user', 'content': 'Write about the importance of self-reflection and self-improvement during workplace conflicts.'}]
pad_token: <|endoftext|>
eos_token: <|im_end|>
Loading weights: 100% 311/311
The tied weights mapping and config for this model specifies to tie model.embed_tokens.weight to lm_head.weight, but both are present in the checkpoints, so we will NOT tie them. You should update the config with `tie_word_embeddings=False` to silence this warning
Tokenizing train dataset: 100% 1000/1000
Tokenizing eval dataset: 100% 100/100
================================================================================
EVAL BEFORE TRAINING
================================================================================
The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config. The model config and generation config were aligned accordingly, being updated with the tokenizer's values. Updated tokens: {'bos_token_id': None, 'pad_token_id': 151643}.
{'eval_loss': nan, 'eval_model_preparation_time': 0.0189, 'eval_runtime': 25.1482, 'eval_samples_per_second': 3.976, 'eval_steps_per_second': 3.976}

================================================================================
START FAST TRAINING
================================================================================

Step    Training Loss    Validation Loss    Model Preparation Time
25      4.514127         nan                0.018900
50      3.317068         nan                0.018900
75      2.523718         nan                0.018900
100     3.386643         nan                0.018900

================================================================================
TRAIN RESULT
================================================================================
{'train_runtime': 412.4126, 'train_samples_per_second': 0.97, 'train_steps_per_second': 0.242, 'total_flos': 1452732883402752.0, 'train_loss': 3.5000829219818117}

================================================================================
EVAL AFTER TRAINING
================================================================================
{'eval_loss': nan, 'eval_model_preparation_time': 0.0189, 'eval_runtime': 24.9824, 'eval_samples_per_second': 4.003, 'eval_steps_per_second': 4.003}
```

## Actual behavior

`trainer.evaluate()` returns:

```python
{'eval_loss': nan, ...}
```

This happens both before and after training.

During training, validation loss is also reported as `nan`:

```text
Step 25   Training Loss 4.514127   Validation Loss nan
Step 50   Training Loss 3.317068   Validation Loss nan
Step 75   Training Loss 2.523718   Validation Loss nan
Step 100  Training Loss 3.386643   Validation Loss nan
```

However, training itself appears to work because the training losses are finite and the final training metrics are finite:

```python
{
    'train_runtime': 412.4126,
    'train_samples_per_second': 0.97,
    'train_steps_per_second': 0.242,
    'total_flos': 1452732883402752.0,
    'train_loss': 3.5000829219818117,
}
```

## Expected behavior

`trainer.evaluate()` should return a finite `eval_loss`, or at least expose a clear reason why eval loss cannot be computed for the current batch/data configuration.

Since training loss is finite and the same dataset format is used for training and evaluation, I expected evaluation loss to also be finite.

## Additional observations

The issue does not seem to be caused by missing columns. The dataset has the expected TPO fields:

```python
['prompt', 'reference', 'chosen', 'rejected']
```

I initially saw tokenizer prefix mismatch warnings when using plain string format. Those warnings disappeared after converting the examples to chat format. After the conversion, tokenization completes successfully and training starts normally.

The following warning appears, but training continues:

```text
The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config.
The model config and generation config were aligned accordingly.
Updated tokens: {'bos_token_id': None, 'pad_token_id': 151643}.
```

I also manually checked that training itself is not failing: the model trains and produces coherent generations after 100 steps. The NaN seems specific to `TPOTrainer.evaluate()` / validation loss reporting rather than a full training failure.

No Python traceback is produced because the script does not crash. The issue is that `TPOTrainer.evaluate()` returns `{'eval_loss': nan, ...}` while training continues with finite losses. The full relevant output is included in the `outputs:` block above.

## Environment

```text
TRL: 1.3.0
Torch: 2.10.0+cu128
CUDA available: True
GPU: Tesla T4
Model: Qwen/Qwen3-0.6B
Dataset: tpo-alignment/triple-preference-ultrafeedback-40K
Precision: fp16=True, bf16=False
PEFT: LoRA, target_modules="all-linear"
```

## Question

Is this a bug in `TPOTrainer.evaluate()` / evaluation loss aggregation, or is there an additional configuration required for evaluation with `TPOTrainer`?

If this is expected behavior for the current experimental implementation, it would be helpful to document the limitation or provide recommended evaluation metrics for TPO, such as chosen-vs-rejected preference accuracy or reward margin.

### System Info

Platform: Linux-6.6.113+-x86_64-with-glibc2.35
Python: 3.12.13 (main, Mar  4 2026, 09:23:07) [GCC 11.4.0]
PyTorch: 2.10.0+cu128
CUDA available: True
GPU: Tesla T4
CUDA version: 12.8
Transformers: 5.0.0
TRL: 1.3.0
Accelerate: 1.13.0
Datasets: 4.8.4
PEFT: 0.18.1

### Checklist

matdou:
# Fix NaN eval_loss when completions are fully truncated

Fixes #5662

## What does this PR do?

When `max_length` is smaller than the prompt length, all completion tokens get 
truncated away. The completion mask ends up entirely `False`, and `F.cross_entropy` 
on an empty tensor returns `NaN`, which is what was causing the issue in #5662.

Two files touched:

- `trl/experimental/tpo/tpo_trainer.py`: guard on the NLL cross-entropy and the 
entropy mean, plus a `logger.warning` so users know their `max_length` is too small
- `trl/trainer/dpo_trainer.py`: same issue in the `sft` loss branch, fixed the 
same way as suggested by @qgallouedec

Regression tests added for both.

## Before submitting

## AI writing disclosure

## Who can review?

@qgallouedec

---

> [!NOTE]
> **Medium Risk**
> Changes core loss computation in DPO and TPO trainers; behavior is limited to the edge case of all-false completion masks, with warnings and tests.
> 
> **Overview**
> Fixes **NaN `eval_loss`** when every completion token is truncated (e.g. `max_length` shorter than the prompt), where masked `F.cross_entropy` on an empty selection used to return NaN (fixes #5662).
> 
> **`DPOTrainer`** (`loss_type="sft"`): only runs chosen-completion cross-entropy when the mask has at least one token; otherwise logs a warning and uses a **zero loss tied to the logits** so `backward()` still works. **`TPOTrainer`**: same guard for the reference NLL term, and entropy averaging no longer calls `.mean()` on an empty masked tensor.
> 
> Regression tests cover DPO SFT (finite loss + backward) and TPO `evaluate()` with a tiny `max_length`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d85b4dfe7c67fc88fda8b24b630b1ca9b1d34ddf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


behroozazarkhalili:
I looked into #5662 independently and arrived at the same root cause and fix as this PR: `max_length` smaller than the prompt truncates all completion tokens, the reference (TPO) / chosen (DPO `sft`) completion mask becomes entirely `False`, and `F.cross_entropy` on the resulting empty tensor returns `NaN`. The `if ref_mask.any()` guard and the matching DPO `sft`-branch guard are the right shape, and the `max_length=5` regression tests reproduce the real trigger. Two things before it can land:

1. **Cursor's grad-fn finding on `dpo_trainer.py` is real and worth fixing.** When `loss_type="sft"` is the only loss and `chosen_mask` is all-`False`, the zero fallback (`torch.zeros(...)` / `chosen_logits.sum() * 0.0`) produces a tensor with no `grad_fn`, so `loss.backward()` raises during **training**. The current tests only call `evaluate()`, so they don't catch it. A fallback that stays connected to the graph (e.g. derived from the logits without a magnitude that can overflow in fp16) plus a training-step test for the all-truncated case would close this. The TPO side avoids the issue by skipping the add entirely and leaving `loss` as the contrastive term, which already requires a grad.

2. **The branch is out of date with `main` (`mergeable_state: dirty`).** A rebase is needed to resolve the conflict before the review can be completed.

Happy to see this merge;  it fixes a real silent-NaN. Flagging in case the rebase + the training-path fix help move it forward.

behroozazarkhalili:
@matdou this PR is `dirty` (out of date with `main`), and the conflict is small and purely in `tests/test_dpo_trainer.py` — the source files (`dpo_trainer.py`, `tpo_trainer.py`) rebase cleanly. I resolved it locally; here's the resolution so you can apply it on a rebase.

**The conflict:** `main` added a new `TestDPOTrainerVLM` class right where your `test_evaluate_no_nan_when_sft_completions_truncated` was inserted. The naive rebase drops your test *inside* `TestDPOTrainerVLM` (between `test_train_vlm` and `test_train_vlm_keep_end_raises`) — which is wrong, since your test is a non-VLM `sft`-loss test and `TestDPOTrainerVLM` is `@require_vision`-gated (it would be skipped whenever vision deps are absent).

**Correct resolution:** your test belongs at the end of `TestDPOTrainer` (the non-VLM class), and the whole `@require_vision class TestDPOTrainerVLM` block from `main` is kept as-is. After `git rebase main`, the test file diff vs `main` should be exactly:

```diff
@@ class TestDPOTrainer(TrlTestCase): (after test_peft_with_quantization) @@
+    def test_evaluate_no_nan_when_sft_completions_truncated(self):
+        # Regression test: when max_length is smaller than the prompt length, all completion
+        # tokens are truncated away and chosen_mask is entirely False for the 'sft' loss branch.
+        # Previously, F.cross_entropy on the resulting empty tensor returned NaN which propagated
+        # into eval_loss.
+        prompts = ["Tell me all about the history of computing and its wide impact."] * 4
+        chosen = ["Great."] * 4
+        rejected = ["Bad."] * 4
+        dataset = Dataset.from_dict({"prompt": prompts, "chosen": chosen, "rejected": rejected})
+
+        training_args = DPOConfig(
+            output_dir=self.tmp_dir,
+            loss_type="sft",
+            max_length=5,  # prompt tokenizes to more than 5 tokens; completions are fully truncated
+            report_to="none",
+        )
+        trainer = DPOTrainer(
+            model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
+            args=training_args,
+            train_dataset=dataset.select(range(2)),
+            eval_dataset=dataset.select(range(2, 4)),
+        )
+
+        results = trainer.evaluate()
+        assert not torch.isnan(torch.tensor(results["eval_loss"])), (
+            "eval_loss must be finite when completions are truncated"
+        )
```

(plus the existing `from datasets import Dataset, load_dataset` import you already added.)

I rebased your commit onto current `main` with this resolution and verified the DPO `sft` fix works: with `max_length=5` and a longer prompt, the guard fires (`Skipping 'sft' loss for a batch where all chosen completion tokens were truncated`) and `eval_loss` is `0.0`, not `NaN`. `ruff check` and `ruff format` both pass on the resolved file.

Two small follow-ups while you're rebasing:
- `ruff format` wants to reflow the long one-line `assert` in `tests/experimental/test_tpo_trainer.py` (the TPO regression test) to the parenthesized multi-line form — CI's format check will flag it.
- The `@require_vision` decorator on `test_train_vlm_keep_end_raises` is redundant (its class is already `@require_vision`), but it's pre-existing on `main`, so leave it.

Happy to help further if useful.

matdou:
Thank you @behroozazarkhalili, your review basically wrote the fix.
               
  - Rebased with your resolution: `test_evaluate_no_nan_when_sft_completions_truncated` at the end of `TestDPOTrainer`, `TestDPOTrainerVLM` untouched
  - Fallback is now `chosen_logits[:, 0, :].softmax(dim=-1).mean(dim=-1) * 0.0`; softmax keeps values in [0, 1], no fp16 overflow, stays in the graph
  - Added `test_train_no_error_when_sft_completions_truncated`: calls `trainer.train()` with `max_steps=1`, `backward()` completes without error

behroozazarkhalili:
I went looking for gaps in this fix and did not find one, so recording the negative result in case it saves someone else the same pass. Not an approval, just what I checked.

The concern I started with was that the guards fix the loss paths while leaving the metric paths able to produce `NaN` for a fully masked batch, specifically the masked logit reductions:

```python
total_chosen_logits = chosen_logits[chosen_mask.bool()].mean(-1).sum()      # tpo_trainer.py:712
total_rejected_logits = rejected_logits[rejected_mask.bool()].mean(-1).sum() # :714
```

That turns out to be safe. With an all-`False` mask the selection is `(0, vocab)`, so `.mean(-1)` reduces over the non-empty vocab dimension across zero rows and yields an empty tensor rather than `NaN`, and `.sum()` on it is `0.0`:

```
selection shape : (0, 50)
.mean(-1) shape : (0,)
.mean(-1).sum() : 0.0   isnan=False
```

The division is then guarded at `:720-721` by `if total_chosen_tokens > 0 else 0.0`, and token accuracy is guarded the same way at `:734`. So the reduction order plus those guards covers it, and the `ref_mask.any()` and `masked_entropy.numel() > 0` guards this PR adds handle the two paths that genuinely would have gone `NaN`.

One adjacent thing worth flagging, since it is the same failure class and this PR does not cover it: `SFTTrainer` has no equivalent guard. When every label in a batch is masked, the forward loss is `NaN` but the run reports `loss: 0.0` and exits normally, with no warning. I filed #6668 with a reproduction. The difference from the DPO behaviour here is that this PR emits a warning alongside its zero, so the user is told why the value is zero, whereas the SFT case is entirely silent. If a shared guard ends up being preferable to per-trainer ones, that issue is probably the place to decide it.

Thanks for the fix @matdou.

behroozazarkhalili:
@matdou the branch is conflicting with `main` again, and both conflicts are single import lines in the test files; the source changes merge clean. Keeping both sides resolves them:

```python
# tests/experimental/test_tpo_trainer.py
from datasets import Dataset, DatasetDict, IterableDatasetDict, load_dataset

# tests/test_dpo_trainer.py
from datasets import Dataset, DatasetDict, IterableDatasetDict, load_dataset
```

Verified on a local merge of the current `main` (1e3f69cd) into this branch: only those two hunks conflict, and `trl/trainer/dpo_trainer.py` and `trl/experimental/tpo/tpo_trainer.py` apply without conflict. Once pushed this is back to the state @qgallouedec asked for on #5662 (guards on the NLL and entropy means, the fully truncated warning, both trainers).

PR #5832, opened 2026-05-25
feat(grpo): tool response/observation training 

LeonEricsson:
## What does this PR do?

Add an optional tool observation prediction loss, allowing the trainer to supervise the model on the textual feedback returned by tool calls.

When enabled, the trainer adds next-token cross-entropy on the textual observations returned by tools. In tool-using rollouts, the model takes an action by emitting a tool call, the trainer executes that call, and the tool result is inserted back into the conversation before generation continues. This gives the model a simple world-model signal from tool feedback: given the action and prior context, predict the observation returned by the tool.

The feature is disabled by default through `tool_observation_loss_weight=0.0`. When enabled, the trainer adds a next-token prediction loss over positions marked by `tool_observation_mask`, scaled by the configured weight.

## Implementation

The internal tool loop now builds a `tool_observation_mask` alongside `tool_mask`. `tool_mask` continues to represent which completion tokens are model-generated versus externally inserted tool tokens. `tool_observation_mask` is narrower, marking the subset of inserted tool tokens that should receive the auxiliary prediction loss.

For custom rollout functions, the rollout provider must return `tool_observation_mask` when `tool_observation_loss_weight` is configured.

The loss itself is standard next-token cross-entropy over the marked tool observation positions.

#### building `tool_observation_mask`

Constructing an exact tool observation mask is the main design issue. The trainer receives a structured tool message with textual content, but that message is not fed to the model as raw text. It passes through the tokenizer chat template, which may add role headers, tool delimiters, separators, generation prompts, or model-specific formatting around the tool response. The auxiliary loss should train on the **actual tool response tokens**, not on the surrounding chat-template scaffolding.

We take an opinionated approach: treat the raw tool text as a content signature and look for that signature inside the fully rendered tool suffix. When the rendered suffix contains the raw content token sequence, those positions are marked as trainable tool observation tokens. If the same content appears more than once, all exact occurrences are marked. If the content cannot be found, it is left unmarked.

This keeps the implementation aligned with the existing chat-template path without pretending to have token-level provenance through arbitrary templates. The method assumes the template wraps tool content without transforming the content tokens themselves. Templates that rewrite, escape, normalize, or otherwise alter the content may prevent a match; the conservative behavior in that case is to skip the auxiliary loss for that content rather than train on template scaffolding.

## reference

https://github.com/anadim/anadim.github.io/blob/master/papers/echo.pdf

## Before submitting

## AI writing disclosure

## Who can review?

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -260,6 +260,9 @@ class GRPOConfig(_BaseConfig):
             When enabled, truncated completions are excluded from the loss calculation, preventing them from being
             incorrectly penalized and introducing noise during training. According to the
             [DAPO](https://huggingface.co/papers/2503.14476) paper, this is a good practice for training stability.
+        tool_observation_loss_weight (`float`, *optional*, defaults to `0.0`):
+            Weight for the auxiliary tool observation prediction loss. When greater than zero, the trainer adds a
+            next-token prediction loss on tool observation tokens.
         sync_ref_model (`bool`, *optional*, defaults to `False`):
             Whether to synchronize the reference model with the active model every `ref_model_sync_steps` steps, using
             the `ref_model_mixup_alpha` parameter. This synchronization originates from the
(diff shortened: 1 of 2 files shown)


LeonEricsson:
@qgallouedec @albertvillanova is this feature interesting? 

This still needs docs and cleanup, but want to gauge interest before finishing. I really like the method, its a simple way to give the model extra signal about the tool/environment during RL. 

Unfortunately, identifying the exact observation tokens is not generally possible because the tool response passes through chat templating. Without template-level provenance or an explicit mask from the rollout, we need to rely on a heuristic

PR #5840, opened 2026-05-25
Remove NeMo Gym Integration Guide (broken)

cmunley1:
# What does this PR do?

Remove broken nemo gym integration for now, can be reintroduced when a better integration is figured out!

Fixes # (issue)

## Before submitting

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **Low Risk**
> Documentation and example-only removal with no changes to core trainers or runtime behavior.
> 
> **Overview**
> This PR **removes the NeMo Gym integration** from TRL because it was broken; the PR description notes it can return when a better approach exists.
> 
> The **docs** drop the `nemo_gym` page and its entries in `_toctree.yml` and `example_overview.md`. The **OpenEnv** guide no longer links to NeMo Gym when describing `rollout_func` for external agent servers.
> 
> The **`examples/scripts/nemo_gym/`** tree is deleted, including `train_multi_environment.py` (GRPO + custom `rollout_func` against NeMo Gym agents), configs, Slurm/DeepSpeed launch scripts, and the README.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea7ec32cf7f00e09ce0dbb03de4681cccada9a30. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
Cc @sergiopaniego @albertvillanova @adithya-s-k

adithya-s-k (approved):

sergiopaniego (approved):
@cmunley1 thanks and please let us know how we can work towards the correct integration 🙏

cmunley1:
Yup! Just want to take this down as people are trying to use it and I do not think it works today.

NeMo Gym needs to control full rollout e.g. through rollout_func or similar (maybe a single call to step() in env factory, I haven't looked close recently) to support custom agent harness such as hermes agent. I think this is doable today via rollout_func, but there was also a note it might be replaced by env factory i think.

It also needs something like async vllm server chat completions endpoint that can support preserving original token ids across multi turn rollouts, whether through a routine like in NeMo RL (https://github.com/NVIDIA-NeMo/RL/blob/main/nemo_rl/models/generation/vllm/vllm_worker_async.py#L41), a token in token out endpoint, or something else. This seems to be the challenge with use with main today. Rather than chat completions it can likely go through existing trl vllm server endpoint such as generate using a Gym side change (a new model server), though I am not sure how to prevent retokenization drift (ie preserve generated token ids) in the current state. 

Let me know if you have any pointers.

qgallouedec:
Thanks for the context, agreed on taking it down.

I want to push back on the framing though. I think that the blocker isn't missing infrastructure, it's a category mismatch with on-policy RL.

Hard invariant: the `completion_ids` the loss sees must be bit-identical to the IDs the sampler emitted, with the sampler's logprobs. One reformatted tool-call JSON, one stripped whitespace, one re-applied template on the harness side, and per-token alignment silently drifts. Loss still computes; gradient is wrong; no error fires. Agent harnesses like hermes are *message*-abstracted by construction: putting one between sampler and optimizer almost guarantees this breaks.

The viable shape is `environment_factory`: Gym exposes the env as a class whose methods are invoked as tools during generation. TRL owns the model loop, tokenizes env responses once trainer-side, appends as IDs. The invariant is preserved by construction. It's like TRL-guaranteed.

`rollout_func` is an escape hatch for cases env factory can't cover. It's for power-users. It's token-IDs-shaped (`prompt_ids`/`completion_ids`/`logprobs`), but doesn't enforce that those IDs came from the sampler unaltered. Unless you're very careful, you're almost certainly breaking the gradient, silently (we did it ourselves, a lot). We're unlikely to extend it in the "plug arbitrary harness" direction, because it would send the message that you can plug arbitrary harnesses in without consequences, which is false.

If Gym's design fundamentally requires the harness/env to own the generation loop, that's not a TRL gap to close, it's incompatible with on-policy RL, unless you use the `rollout_func` escape hatch and accept the risk of silent gradient bugs.

PR #5842, opened 2026-05-26
Padding-free invariance test

qgallouedec:
Adds a `sft_fa2` equivalence class to the invariance suite that compares padded vs padding-free SFT under FA2 + bf16. Catches divergences between the two paths.

## Dependencies (must merge first)
- #5843 — `cli-accept-none` (needed for `--max_length None` on the padfree member)
- #5844 — `invariant-per-field-tolerances` (grad_norm needs its own tolerance under FA2+bf16)
- #5845 — `refresh-invariance-snapshots` + the matching transformers release

## Why

Padding-free is a subtle execution path with its own kernel (FA varlen) and label conventions. Without a direct invariance check, regressions silently shift trajectories. This PR adds the check.

## Calibration

See https://github.com/huggingface/trl/pull/5842#issuecomment-4539190615 for the deep investigation of the residual grad_norm noise and why `tol.grad_norm = 1.0` is correct (not papering over a bug).

---

> [!NOTE]
> **Low Risk**
> Test-only changes (new JSON reference and equivalence-class config); no production training or CLI logic in this diff.
> 
> **Overview**
> Adds an **`sft_fa2`** invariance equivalence class so padded SFT and **`padding_free=True`** SFT are compared under **Flash Attention 2 + bf16** (same model, dataset, seeds, 50 steps). The canonical run **`sft_fa2`** gets a new reference snapshot in `tests/invariant/references/sft_fa2.json`; **`sft_fa2_padfree`** must match it within class-specific **`tol`** / **`residual_tol`** (tighter on **loss**, much looser on **grad_norm** for FA2/bf16 kernel noise).
> 
> Docs in `test_invariant.py` are expanded to describe per-field tolerances and the `--report` workflow for calibrating them.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 805a7ea14ed6a8b2a076e9daa5fa68ebd7e1621e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
# Why the padding-free invariance test has a loose `grad_norm` tolerance

## Setup

The `sft_fa2` equivalence class in `tests/invariant/test_invariant.py` compares two SFT configs that should be mathematically identical:

- **padded**: FA2 + bf16, `padding_free=False`
- **padfree**: FA2 + bf16, `padding_free=True`

Loss matches very tightly (`max |Δ| ≈ 7e-3`, `mean Δ ≈ 2e-5`). But `grad_norm` still diverges over 50 steps: `max |Δ| ≈ 0.56`, `mean Δ ≈ -0.025`.

This was investigated thoroughly. The verdict: **genuine bf16 + FA-varlen-kernel noise, not a bug.** The tolerance is real headroom for real noise.

## Step 1 is essentially bit-identical

Side-by-side, with the same weights and the same batch:

| | padded | padfree |
|---|---|---|
| `num_items_in_batch` | 1535 | 1535 |
| logits | bit-identical | bit-identical |
| loss | bit-identical | bit-identical |
| `dL/dlogits` | bit-identical | bit-identical |
| grads (288 / 290 params) | bit-identical | bit-identical |
| global grad_norm | differs by `1.7e-9` | |

Two parameters carry the entire difference:
- `layers.1.post_attention_layernorm.weight` — one element differs by exactly **1 bf16 ULP**.
- `layers.23.self_attn.k_proj.bias` — `Δ = 1.2e-10` (pure fp32 roundoff).

## Why that one RMSNorm element differs

RMSNorm's weight gradient is

```
dL/dw_i = Σ_t (dL/dy_{t,i} · x̂_{t,i})
```

a reduction over the `(B, L)` token dimension. Padded mode reduces over shape `(2, 1182, 896)`; padfree over `(1, 1239, 896)`. The fp32 partial-sum *order* differs between those two shapes, and for one specific output column (index 201) the running sum crosses a rounding boundary differently. The bf16 representation of the final value tips by one ULP — a true tie-break artifact of mixed-precision accumulation, not a logic error.

## Why it grows over 50 steps

Adam stores momentum buffers in fp32. The 1-ULP gradient noise gets folded into the running averages every step. After ~2 optimizer steps the buffers have shifted enough to flip a bf16 parameter bit. From step 3 onward the two trainings are exploring genuinely different points in parameter space.

The divergence is then amplified by gradient clipping: with `max_grad_norm=1.0` and grad_norm ~10 every step, **every** update is at the clip. After clipping, the update direction is `-lr · grad / grad_norm` — so a tiny change in gradient direction (or magnitude) maps to a tiny change in update direction. Compounded over steps, this is a chaotic feedback loop.

## The "one-sided" mean Δ is path-dependent

We measured at `data_seed=42, 50 steps` a `mean Δ grad_norm = -0.025`. That looks like a systematic bias, but it isn't:

| data_seed | 30-step mean Δ grad_norm |
|---|---|
| 42 | -0.003 |
| 100 | +0.119 |
| 200 | +0.021 |

The sign flips. The −0.025 is one realization of a random walk, not a bias. Different seeds will land on different signs and magnitudes.

## What this means for the test

- `tol.grad_norm = 1.0` covers the worst step in a 50-step run with comfortable margin.
- `residual_tol.grad_norm = 5e-2` covers the path-dependent drift at the current seed.
- If you need cross-seed/cross-hardware robustness, bump these by another ~3× — the worst we saw across three seeds in a 30-step run was `max |Δ| = 2.74`.

The loose `grad_norm` tolerance is not papering over a bug. It's the noise floor of mixed-precision training with chaotic feedback from clipping and Adam momentum.

albertvillanova (approved):
Thanks

PR #5846, opened 2026-05-26
Default SFT loss to chunked_nll

qgallouedec:
Switch the default `loss_type` in [`SFTConfig`] from `"nll"` to `"chunked_nll"`, with a safe automatic fallback to `"nll"` when `use_liger_kernel=True`.

Contains the changes from #5857!

Concretely:

- `loss_type` field default is now `None`.
- `SFTConfig.__post_init__` resolves it to `"chunked_nll"` by default, or to `"nll"` when `use_liger_kernel=True` (the two paths are incompatible — explicit `ValueError` already exists in the trainer).
- Docstring, help string, and user-facing docs in `sft_trainer.md` / `reducing_memory_usage.md` are updated to reflect the new default and the resolution rule.
- The FSDP2 caveat (`fsdp_reshard_after_forward=False`) and the coverage caveat (standard causal LM / VLM with a top-level `lm_head`) are documented alongside the default.

## Why

This is motivated entirely by the priority of providing **good defaults**.

`chunked_nll` is mathematically equivalent to `nll`, but trades a tiny amount of forward bookkeeping for a large reduction in peak activation memory at the `lm_head`. Benchmarks in [#5575](https://github.com/huggingface/trl/pull/5575) show **~30 % less peak VRAM on average, up to ~50 %** on large-vocab models, with wall time typically neutral or slightly faster.

That is the kind of win that *should not* require users to read the memory-optimization page and flip a flag. Users who never touch `loss_type` are today silently paying a memory cost for no benefit, the old default is strictly dominated for the supported configurations.

By auto-resolving in `__post_init__` rather than hard-coding `"chunked_nll"` as the field default, we avoid breaking the one population that genuinely cannot use it: Liger users. Everything else (FSDP2 wall-time, non-standard heads) is already covered by the existing warnings / errors and is documented.

---

> [!NOTE]
> **Medium Risk**
> Default training behavior changes for anyone who omits `loss_type` (new lm_head patching path with existing PEFT/VLM/FSDP2 caveats), though loss math matches `nll` for supported setups.
> 
> **Overview**
> **`SFTConfig` now resolves unset `loss_type` to `"chunked_nll"`** instead of `"nll"`, giving most `SFTTrainer` runs the memory-efficient chunked cross-entropy path without an explicit flag. When **`use_liger_kernel=True`**, the same unset field still resolves to **`"nll"`** because the two options are incompatible.
> 
> The TRL 1.7 **`FutureWarning`** about this default change is removed; docstrings, CLI help, **`sft_trainer.md`**, and **`reducing_memory_usage.md`** now describe chunked NLL as the default and document opting out with **`loss_type="nll"`**. Tests that targeted the standard loss path are renamed and pass **`loss_type="nll"`** explicitly; Liger unit tests do the same so **`model.forward`** stays unpatched.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed49e49dbe81004300f0251603bafa58806fca60. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/trainer/sft_config.py
```diff
@@ -309,3 +317,7 @@ def __post_init__(self):
                 stacklevel=3,
             )
             self.packing_strategy = "bfd_split"
+
+        # When unset, default to "chunked_nll" unless `use_liger_kernel=True`, in which case default to "nll".
+        if self.loss_type is None:
+            self.loss_type = "nll" if self.use_liger_kernel else "chunked_nll"
```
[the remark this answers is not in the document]
qgallouedec: #5857

trl/trainer/sft_config.py
```diff
@@ -98,17 +98,21 @@ class SFTConfig(_BaseConfig):
             Whether to compute loss only on the assistant part of the sequence. If set to `True`, loss is computed only
             on the assistant responses, which is supported only for [conversational](#conversational) datasets. If
             `False`, loss is computed on the entire sequence.
-        loss_type (`str`, *optional*, defaults to `"nll"`):
-            Type of loss to use. Possible values are:
+        loss_type (`str`, *optional*, defaults to `"chunked_nll"`):
```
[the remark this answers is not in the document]
qgallouedec: that's intentional. Although not 100% accurate, that actually way more informative for user that the default loss type is chunked. ```loss_type (`str`, *optional*, defaults to `None`)``` would be quite misleading: I read "No loss type used??"
On the other hand, the only special case is liger, which also uses a chunked loss. So technically it's always kinda true.

docs/source/reducing_memory_usage.md
qgallouedec, line 183: I wasn't sure if we should just remove this section now that's it's default.

tests/test_sft_trainer.py
```diff
@@ -911,9 +911,13 @@ def test_train_with_liger(self):
     def test_compute_loss_skip_logits_on_eval_without_metrics_with_liger(self):
         dataset = load_dataset("trl-internal-testing/zen", "standard_language_modeling", split="train[:1]")
 
+        # Init with `use_liger_kernel=False` to skip Liger model wrapping, then flip the flag after to exercise the
+        # Liger branch of `compute_loss`. `loss_type="nll"` so the chunked path isn't patched in (incompatible with
+        # Liger), keeping `model.forward` unmodified.
```
qgallouedec, line 958: not a fan of these test and these workarounds, but they exist, so removing/refactoring them is out of the scope of this PR

albertvillanova (commented):
Thanks for working on this!

One thought: would it make sense to introduce this through a deprecation cycle instead of changing the default immediately? Configuration defaults are part of the public API. We could warn users that the default will change from "nll" to "chunked_nll" in a future release, and recommend setting loss_type explicitly if they want to preserve the current behavior. This would make the transition more predictable for users who rely on implicit defaults.

That said, there is also a reasonable counterargument: we could also consider "better default for the same semantics" as an improvement rather than a breaking change.

qgallouedec:
Both would make sense, wdyt @kashif  @AmineDiro?

docs/source/reducing_memory_usage.md
AmineDiro: No I think it's good to be clear and explicite that it is the default now 👍🏼

AmineDiro:
@qgallouedec @albertvillanova 
Ran a small test to profile comms for FSDP2 (rank 1) 
image


All look good 👍🏼 . We call `all_gather` **ONCE** for the forward and backward and it's only fetched once for all the chunks.  For details :
  - `lm_head` is  `Linear(2048, 151936)`, `bf16`, `fully_shard`  into the root FSDP group (mirrors accelerate's default wrapping).
  - 8192 valid tokens, chunk_size=256. SO we have 32 gradient-checkpoint chunks with `use_reentrant=False`.
  - 8× H100 (world=8)

Declared as :
```python
class LMHeadModel(torch.nn.Module):
    """Stand-in for the SFT model: the chunked-CE forward accesses `self.lm_head.weight`
    directly (never calls `self.lm_head.__call__`), exactly like `_chunked_ce_forward`."""

    def __init__(self):
        super().__init__()
        self.lm_head = torch.nn.Linear(HIDDEN, VOCAB, bias=False, dtype=torch.bfloat16)
        self.seen_shape = None

    def forward(self, hidden, labels):
        w = self.lm_head.weight
        self.seen_shape = tuple(w.shape)  # what the chunked path sees during forward
        loss = hidden.new_zeros((), dtype=torch.float32)
        for start in range(0, hidden.size(0), CHUNK_SIZE):
            h_chunk = hidden[start : start + CHUNK_SIZE]
            lbl_chunk = labels[start : start + CHUNK_SIZE]
            chunk_loss = torch.utils.checkpoint.checkpoint(
                _chunk, h_chunk, w, lbl_chunk, use_reentrant=False
            )
            loss = loss + chunk_loss
        return loss / hidden.size(0)
```

trl/trainer/sft_config.py
```diff
@@ -98,23 +98,21 @@ class SFTConfig(_BaseConfig):
...
             - `"dft"`: Dynamic Fine-Tuning, as described in
               [this paper](https://huggingface.co/papers/2508.05629).
             - `"chunked_nll"`: same math as `"nll"`, but the `lm_head` projection is computed on non-ignored tokens
               only (positions with `labels == -100` are dropped before the matmul) and the cross-entropy is processed
               in chunks of tokens to reduce peak activation memory. Not compatible with `use_liger_kernel`. Under
               FSDP2, set `fsdp_reshard_after_forward false` in the accelerate config — the chunked path otherwise
-              re-gathers `lm_head.weight` per chunk during backward, adding noticeable wall-time.
+              re-gathers `lm_head.weight` per chunk during backward, adding noticeable wall-time. The patched
+              `lm_head` path covers standard causal LMs and VLMs whose language model exposes a top-level `lm_head`;
+              architectures with a non-standard head are not supported.
```
qgallouedec:
```suggestion
              re-gathers `lm_head.weight` per chunk during backward, adding noticeable wall-time.
```

trl/trainer/sft_config.py
qgallouedec:
```suggestion
```

qgallouedec:
thanks for the reviews guys, I'll merge it, I want to let some time before merging and release of 1.7

PR #5854, opened 2026-05-26
Padding-free training in AsyncGRPO

qgallouedec:
Makes padding-free the only training mode for `AsyncGRPOTrainer`. (No new flag it replaces the previous `(B, max_len)` padded forward pass.)

## What changed

- `compute_loss` unpacks each rank's padded slice into a single concatenated `(1, total_real_tokens)` sequence using `attention_mask.bool()`, FlashAttention derives `cu_seq_lens` from the position resets (`position_ids`).
- The collator is unchanged: it still pads to `(B, max_len)` so `DataLoaderDispatcher` can slice rows across ranks. The flattening happens post-dispatch.
- Model is loaded with `attn_implementation="kernels-community/flash-attn2"`

## Why

Padding tokens are pure wasted compute in the forward/backward. With variable-length completions across a batch, the previous local-max-len truncation only mitigated cross-rank padding, within-rank padding remained. Concatenating real tokens removes both.

---

> [!NOTE]
> **High Risk**
> Core training forward/loss and distributed batch layout changed, with a hard dependency on FlashAttention3 and stricter GPU requirements.
> 
> **Overview**
> **AsyncGRPOTrainer** switches to **padding-free** training: each rank’s rollout samples are concatenated into one packed row with **`position_ids`** resetting per sequence and **per-token advantages**, while rectangular batches are kept only for **`DataLoaderDispatcher`** scatter (inter-rank padding is stripped in **`compute_loss`**).
> 
> The model is loaded with **`attn_implementation="kernels-community/flash-attn3"`** because packed sequences rely on FlashAttention **`cu_seq_lens`** from position resets; SDPA/eager are no longer supported on this path. **`DataCollatorForRollout`** gains **`num_processes`**, rank-strided grouping, **`position_ids`**, NaN-padded per-rank **metrics**, and **`global_n_tokens`** shaped for dispatch. **`compute_loss`** flattens valid tokens via **`attention_mask`**, forwards with **`position_ids`** (not **`attention_mask`**), and applies GRPO with expanded advantages; logging uses **`position_ids == 0`** for sample counts and packed length for **`train_seq_len`**.
> 
> Integration tests skip when Flash Attention is unavailable (**Ampere+** GPU or **XPU**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 151bbbe30bcadfd3d9706553fc4eea59dcbef3ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -300,7 +300,11 @@ def __init__(
 
         # Model
         model_name = model
-        model = AutoModelForCausalLM.from_pretrained(model, device_map=None, dtype=torch.float32)
+        # FlashAttention is required: training runs in padding-free mode, where sequences are concatenated into a
+        # single row and `cu_seq_lens` are derived from `position_ids` resets. SDPA/eager can't handle this.
+        model = AutoModelForCausalLM.from_pretrained(
+            model, device_map=None, dtype=torch.float32, attn_implementation="kernels-community/flash-attn2"
+        )
```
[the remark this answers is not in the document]
qgallouedec: we don't care it's experimental

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -300,7 +300,11 @@ def __init__(
 
         # Model
         model_name = model
-        model = AutoModelForCausalLM.from_pretrained(model, device_map=None, dtype=torch.float32)
+        # FlashAttention is required: training runs in padding-free mode, where sequences are concatenated into a
+        # single row and `cu_seq_lens` are derived from `position_ids` resets. SDPA/eager can't handle this.
+        model = AutoModelForCausalLM.from_pretrained(
+            model, device_map=None, dtype=torch.float32, attn_implementation="kernels-community/flash-attn2"
```
AmineDiro: why not FA3 directly ? It is faster than FA2 from my benchmark. Is it because it's only on amper GPUs?

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -474,11 +488,11 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
 
         completion_mask = completion_mask[:, 1:]
```
AmineDiro, line 591: If I understand correctly, the shift is applied to the whole row, so each sequence boundary yields log-prob (predicting seq k+1's first token from seq k's context).  if the first token of each sequence has `completion_mask==0 `(prompt ) it shouldn't be a problem. 
Maybe we should assert `completion_mask` is 0 at each seq_offsets index, since a prompt-free sample would leak a garbage term into the loss ?

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro, line 1: I think I have some concerns on putting the padding-free code in `compute_loss`:

- **Packing is per-rank, post-dispatch.** Each rank packs its own arbitrary dim-0 slice, after the collator already padded to the global max and dispatched the pad columns. So we pad → dispatch → unpad, when we could pack once where the data lives. This could be problematic in setups where each rank has 1 batch.
- **No per-rank microbatch balancing (straggler).** Per-rank token counts are whatever the slice happened to contain; one rank can draw all the long sequences and stall everyone at the all-reduce. A length-balanced distribution (or real bin-packing) on rank 0 would fix this and is safe, given the point above.
-  This packing/`position_ids`/`per_token_advantages` reshaping doesn't belong in `compute_loss` IMO. `compute_loss` should receive already-clean, pad-free sequences and just compute the loss. the flattening/segmenting is a data-layout responsibility (collator / a dedicated pre-dispatch step) or in the RolloutQueueDataset as post_process. Right now, the loss function is doing batch surgery, which makes it hard to test and couples the math to the layout.
- No per-microbatch token budget, The seq length is whatever the queue served, worst case `p·(max_prompt+max_completion)`. There's no dynamic batch sizing to bound peak memory ?

trl/experimental/async_grpo/async_grpo_trainer.py
qgallouedec: let's use fa3

qgallouedec:
```suggestion
            model, device_map=None, dtype=torch.float32, attn_implementation="kernels-community/flash-attn3"
```

trl/experimental/async_grpo/async_grpo_trainer.py
qgallouedec: about 3. I think I agree. And this is what we do in SFT. The thing that we need to be aware though: `get_train_dataloader` uses `dispatch_batches=True`: the collator runs once on the main process and emits a rectangular `(global_B, max_len)` batch; `DataLoaderDispatcher` slices it on dim=0 (rows) and scatters rows to each rank. The padding exists so there are rows to slice. So the flatten must happen after dispatch → per-rank.

Flattening in the collator into one row `(1, total_global_len)` gives the dispatcher nothing to split → breaks

OR: collator pre-groups into exactly `num_processes` flattened rows `(num_processes, max_rank_total_len)`, scattered row `i` → rank `i`. But that couples the collator to Accelerate's dispatch internals, only trades per-sample padding for inter-rank padding. And we still need to remove this inter-rank padding in the `compute_loss`

trl/experimental/async_grpo/async_grpo_trainer.py
qgallouedec: Yes exactly. I'd rather not add a defensive assert here. All chat templates have role markers, system prompt etc, there is no real scenario of a prompt-free sample. I'm not even sure that vllm would allow empty list input.

trl/experimental/async_grpo/async_grpo_trainer.py
qgallouedec: About 2.  
Packing now happens in the collator (pre-dispatch), so length-balanced assignment is easy to add here. I've kept the simple strided split for now

https://github.com/huggingface/trl/blob/4ad0a568188f4a138f8cfaa0bff5a2021563ea4b/trl/experimental/async_grpo/async_grpo_trainer.py#L206

 it doesn't balance token count, so stragglers are still possible. Happy to add greedy length-balancing as a follow-up if profiling shows rank idle time at the all-reduce.

qgallouedec: About 4. 
right, there's no dynamic micro-batching, each rank runs a single forward over its whole slice. That's unchanged by this PR though: the previous padded path also did one forward over `(B_local, local_max_len)`, so the worst-case bound is identical before and after.

Padding-free lowers the typical tokens-per-forward but doesn't change that worst case. A token-budgeted chunk + grad-accumulation loop to bound peak memory would be a good follow-up, but it's orthogonal to this change imo

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -545,11 +559,11 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
         old_log_probs = old_log_probs[:, 1:]
-        advantages = advantages.unsqueeze(1)
+        per_token_advantages = per_token_advantages[:, 1:]
         log_ratio = log_probs - old_log_probs
         ratio = torch.exp(log_ratio)
         clipped = torch.clamp(ratio, 1 - self.epsilon_low, 1 + self.epsilon_high)
-        per_token_loss = -torch.min(ratio * advantages, clipped * advantages)
+        per_token_loss = -torch.min(ratio * per_token_advantages, clipped * per_token_advantages)
```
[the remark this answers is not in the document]
qgallouedec: https://github.com/huggingface/trl/pull/5854#discussion_r3397247812

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -184,33 +184,70 @@ def __iter__(self):
...
+            completion_mask.append(
+                torch.tensor([m for example in group for m in example["completion_mask"]], dtype=torch.long)
+            )
+            old_log_probs.append(
+                torch.tensor([lp for example in group for lp in example["old_log_probs"]], dtype=torch.float32)
+            )
+            position_ids.append(torch.cat([torch.arange(n) for n in seq_lengths]))
+            advantages.append(
+                torch.cat([torch.full((n,), example["advantage"]) for example, n in zip(group, seq_lengths)])
+            )
```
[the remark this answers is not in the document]
qgallouedec: this is not possible, the batch size takes into account the world size

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -184,33 +184,72 @@ def __iter__(self):
...
+    num_processes: int = 1
     return_tensors: str = "pt"
 
     def torch_call(self, examples: list[dict[str, Any]]) -> dict[str, Any]:
-        input_ids = [torch.tensor(example["input_ids"], dtype=torch.long) for example in examples]
-        attention_mask = [torch.ones(len(ids), dtype=torch.long) for ids in input_ids]
-        completion_mask = [torch.tensor(example["completion_mask"], dtype=torch.float32) for example in examples]
-        old_log_probs = [torch.tensor(example["old_log_probs"], dtype=torch.float32) for example in examples]
-        advantages = torch.tensor([example["advantage"] for example in examples], dtype=torch.float32)
+        groups = [examples[i :: self.num_processes] for i in range(self.num_processes)]
```
AmineDiro, line 206: I think there is a subtle bug in using `self.num_processes` but I am not sure how accelerate handles replicating batches. Because is should actually be something like `self.dp_rank`. Because the groups need to be replicated for TP.

AmineDiro: Also `examples[i :: self.num_processes]` does strided groupping. Any reason to group this way and not grouop successive groups together ? I think it doesnt matter but just asking out of curiosity if I am missing something ?? maybe having better balance lengths across groups ?

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: Thanks very clear responses 
> Padding-free lowers the typical tokens-per-forward but doesn't change that worst case. A token-budgeted chunk + grad-accumulation loop to bound peak memory would be a good follow-up, but it's orthogonal to this change imo

Very clear, and I do agree that its not really part of this PR just something to keep in mind 👍🏼 


> it doesn't balance token count, so stragglers are still possible. Happy to add greedy length-balancing as a follow-up if profiling shows rank idle time at the all-reduce.

No I think that's good for now. A good balancing would probably be some heuristic based on the  _real_ attention cost 🤔 

>  Padding-free lowers the typical tokens-per-forward but doesn't change that worst case. A token-budgeted chunk + grad-accumulation loop to bound peak memory would be a good follow-up, but it's orthogonal to this change imo

Yesss this could be great ! There some subtelty making this work in distributed settings imho. Thinking ahead a bit, I think we should have the same number of mbs per rank so we need to have some sort of algorithm, really upstream, probably not in the collator ( because collator only has global_batch_size examples already). To make the per_rank local micro-batch agree on token count and not cause deadlock in FSDP settings.

Anyway, great work !!

AmineDiro (approved):

PR #5573, opened 2026-04-16
refactor: self distillation trainers (sdpo/sdft/...) 

LeonEricsson:
_fyi see #5609 for a concrete extension (SD-Zero) on top of this PRs refactor_

### Summary

Major internal restructuring of the experimental self-distillation trainers. The core change: unify shared lifecycle logic in `BaseSelfDistillationTrainer`, while concrete trainers (`SDFTTrainer`, `SDPOTrainer`) own their algorithmic recipes.

The previous design mixed generic self-distillation lifecycle with SDPO-specific reward/scoring machinery in the same inheritance stack, meaning the base implicitly assumed every method needed reward-based optimization. This PR corrects that with a clean split: base owns infrastructure, trainers own semantics. 

No intended change in functionality.

### What changed

**Base class now owns the common lifecycle:**
model/tokenizer/generation setup, student-side rollout generation (transformers + vLLM), and teacher synchronization.

**Concrete trainers own the method recipe:**
rollout interpretation, teacher input construction, algorithm-specific metadata (rewards, advantages, etc.), and loss composition.

**SDPO owns its identity.**
Reward scoring and policy optimization are no longer baked into the base. SDPO explicitly manages its own reward functions, normalization, and hybrid loss.

**Teacher regularization**
is now a single semantic config parameter (`teacher_model_kind`: `"live"` / `"base"` / `"ema"`), with the base handling PEFT/non-PEFT details internally.

### Removed

I’d like to touch on a few abstractions I’ve cleared out and the reasoning behind those calls. In short, I felt that **`PromptTokenizer`**, **`SelfDistillationMixin`**, and **`OnlineRolloutMixin`** were "leaky"- masking the logic they were supposed to define and making the architecture harder to follow. 

`PromptTokenizer` logic now lives directly in the base, next to the rollout pipeline it serves. The two mixins were splitting orchestration, rollout mechanics, and algorithm-specific tasks (like reward scoring) across too many pieces. That responsibility is now split cleanly: **`BaseSelfDistillationTrainer`** handles the "heavy lifting" of the distillation machinery and rollout mechanics. Meanwhile, concrete trainers have full ownership over their specific batch finalization and loss logic.

### Caveat

- `generate_from_teacher` isn't ported; the base currently assumes student-sourced rollouts. There's still hanging code that references it, so we need to decide whether this is a capability worth porting before this PR can land.

[...]

PR #5862, opened 2026-05-27
[1/2] refactor: decoupled self distillation trainers (sdpo, sdft, ...)

LeonEricsson:
# What does this PR do?

This is an alternative refactoring to #5573, which removes the Base class abstraction and instead implements SDFT and SDPO as independent trainers. 

From 5573:

_"As per discussion on slack, this PR [5573] still retains the Base class contract, forcing trainer (SDPO,SDFT) coupling. While this works well for SDPO, SDFT and SD Zero, long term stability of these contracts is uncertain. An alternate path is to drop the Base altogether, let each trainer live separately, and move common functionality into pure helpers"._

This is that alternate path.

### roadmap

- [x] refactor SDPO and SDFT into self contained trainers
- [x] refactor test suites
- [x] update docs
- [x] `generate_from_teacher`?

## Before submitting

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **High Risk**
> Large refactor of RL/self-distillation training paths with breaking config renames (`full_logit_distillation`, `sync_ref_model`, `teacher_regularization`) and duplicated trainer logic; incorrect teacher sync or loss dispatch would silently change training behavior despite new tests.
> 
> **Overview**
> This PR **decouples SDPO and SDFT** from the shared experimental self-distillation base (mixin/config) into **standalone trainers** with their own configs, training loops, and per-package `loss_utils` helpers. Shared behavior is duplicated as pure functions rather than inherited contracts.
> 
> **Config/API changes:** `full_logit_distillation` is replaced by **`distillation_mode`** (`sampled_token`, `full_logits`, `topk_logits`). Teacher handling moves from **`teacher_regularization` / `sync_ref_model` / `ref_model_*`** to **`teacher_model_kind`** (`base`, `live`, `ema`) with **`teacher_sync_steps`** and **`teacher_update_rate`**. `SDFTConfig` and `SDPOConfig` now extend `_BaseConfig` directly and own generation/vLLM fields previously centralized.
> 
> **SDFT:** Inlines rollout buffering, transformers/vLLM generation, teacher setup (including PEFT EMA via `teacher_sync.py`), and distillation loss (top-k / full / sampled-token) plus optional **Liger** fused JSD for `full_logits`. Drops separate `ref_model` in the CLI script. Docs now describe continual-learning SDFT and document vLLM.
> 
> **SDPO:** Same teacher and distillation-mode model; configs gain explicit online-RL fields; examples/docs use `distillation_mode` instead of `full_logit_distillation`.
> 
> **Tests/docs:** Trainer tests are refactored and expanded (`test_self_distillation_trainer_behavior`, Liger parity tests, EMA renamed); paper index and trainer markdown updated for the new flags.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit efe6cd1e08e3a6b2130e7eb156fdc69f7b6141ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/source/sdft_trainer.md
qgallouedec: sync_ref_model doesn't exist anymore

qgallouedec: same here, doesn't exist anymore

qgallouedec (commented):
A few questions:

If my understanding is correct, SDFT default teacher changed semantics: OLD default ran with teacher_model=None → live student as teacher (self_distillation_mixin.py:211-215). NEW default teacher_model_kind="base" (sdft_config.py:190-196) deep-copies the student into a frozen teacher. Right? Is it expected?

What about moving these into a follow-up PR (not hard request, just a suggestion, happy to merge as is if you think it's better):

- `generate_from_teacher` for SDFT, net-new feature added in the last commit, currently no test coverage and the prompt-truncation behavior deserves a closer look on its own.
- `topk_logits` distillation mode (+ `distillation_add_tail`): the bool → enum rename is great, but the third mode is new math with no tests yet. Could ship as `distillation_mode` ∈ {`sampled_token`, `full_logits`} here and add topk_logits after.
- `teacher_model_kind="live"`: base and ema are direct renames of existing paths, but live is a new mode. Same idea: drop it here, add it later.
- `sdpo_policy_loss_mode="policy_only"`: replaces the previous silent fall-back when `distillation_weight=0`. Reasonable change but a new mode; would be cleaner as its own.
- FSDP summon_full_params in the transformers generation path: looks like a fix rather than a refactor?

That would leave this PR as a pure refactor with identical defaults, which is easier to sign off on. The remaining items then land as small, focused follow-ups.

LeonEricsson:
> A few questions:
> 
> If my understanding is correct, SDFT default teacher changed semantics: OLD default ran with teacher_model=None → live student as teacher (self_distillation_mixin.py:211-215). NEW default teacher_model_kind="base" (sdft_config.py:190-196) deep-copies the student into a frozen teacher. Right? Is it expected?
> 

For completeness, let me give some context on the change. On main, the teacher was selected implicitly by sync_ref_model × PEFT-ness

| main | resulting teacher |
|---|---|
| sync_ref_model=False + PEFT | adapter disabled → base weights |
| sync_ref_model=False + non-PEFT | teacher_model=None → returns model → live student |
| sync_ref_model=True + PEFT | PEFTAdapterEMACallback "teacher" adapter → EMA |
| sync_ref_model=True + non-PEFT | deepcopy + sync callback → EMA |

Note the mismatching behavior under sync_ref_model=False (depending on PEFT-ness). sync_ref_model=False is the default on main

I've unified the behavior under `teacher_model_kind`. This is mostly just to make the existing choice explicit, and importantly **decouple teacher-kind from PEFT-ness**.  Here's the mapping 

| combo | teacher | main equivalent |
|---|---|---|
| live + non-PEFT | live student | sync_ref_model=False, non-PEFT |
| live + PEFT | live student w/ adapter | none |
| base + PEFT | adapter-disabled base weights | sync_ref_model=False, PEFT |
| base + non-PEFT | frozen deepcopy of initial student | none |
| ema + PEFT (pure-lora) | EMA "teacher" adapter | sync_ref_model=True, PEFT |
| ema + non-PEFT | EMA deepcopy | sync_ref_model=True, non-PEFT |

There are two "new" path's which aren't reachable on main, but I argue these aren't new behavior. it's mostly a restructure.

**conclusion/verdict**  Has the SDFT default changed? We'll main's default behavior depends on PEFT-ness. This PR, with "live", implements the non-PEFT branch of that default behavior. So no, I would argue it hasn't changed. Splitting this revamp into multiple PRs is difficult, I'd propose we merge as is or push the whole revamp into a follow up.

>  * `generate_from_teacher` for SDFT, net-new feature added in the last commit, currently no test coverage and the prompt-truncation behavior deserves a closer look on its own.

`generate_from_teacher` retains the same functionality on as on main. generate completions from the teacher privileged context, then train on student prompt. I also think the prompt-truncation behavior is the same, both main and the PR apply left truncation on the tokenized prompts. 

There's two tests that cover `generate_from_teacher` in `test_sdft_trainer.py`, one that checks that the student prompts for training don't contain the hint, and one that checks that old_per_token_logps is skipped.

**conclusion/verdict** keep as is

>  * `topk_logits` distillation mode (+ `distillation_add_tail`): the bool → enum rename is great, but the third mode is new math with no tests yet. Could ship as `distillation_mode` ∈ {`sampled_token`, `full_logits`} here and add topk_logits after.

The distillation mode enum is only a refactor to make the objective more explicit. It implements the same function as main. On main there are three config fields:

  - distillation_topk: int|None (default 100)
  - full_logit_distillation: bool (default False)
  - distillation_add_tail: bool (default False)

which combined resolved into some code path implementing a type of distillation objective.  

| main path | condition | new enum value |
|---|---|---|
| top-k branch  | topk is not None and (full_logit or _allow_topk…()) | topk_logits |
| full-vocab branch | full_logit_distillation=True | full_logits |
| token-level reverse-KL else branch | otherwise | sampled_token |

all I've done is lift this into a single field `distillation_mode`. So `topk_logits` is the exact same code as main's top-k branch. Also SDPO and SDFT retain their defaults, SDFT uses top-k and SDPO uses the sampled token.

There's specific topk math tests in `test_self_distillation_trainer_behavior.py`, and an integration test in `test_sdpo_trainer.py`. 

**conclusion/verdict** no new functionality, could be pushed to a follow up but I would recommend keeping it

> 
>  * `teacher_model_kind="live"`: base and ema are direct renames of existing paths, but live is a new mode. Same idea: drop it here, add it later.

answered above

>  * `sdpo_policy_loss_mode="policy_only"`: replaces the previous silent fall-back when `distillation_weight=0`. Reasonable change but a new mode; would be cleaner as its own.

**conclusion/verdict** Yes, we could push this change into a follow up.

edit: moved

>  * FSDP summon_full_params in the transformers generation path: looks like a fix rather than a refactor?

**conclusion/verdict** This could also be moved to a follow up. It is indeed a bug fix. 

edit: moved

docs/source/sdft_trainer.md
LeonEricsson: ye all docs needs updating, fixing asap

trl/experimental/sdpo/sdpo_trainer.py
```diff
@@ -367,23 +1126,154 @@ def _warn_on_inactive_self_distillation(self, mode: str) -> None:
...
-                return base_policy_loss
+    def _compute_hybrid_loss(self, model, inputs) -> torch.Tensor:
+        distillation_logits = self._compute_teacher_student_logits(model, self.teacher_model, inputs)
+        policy_loss = self._compute_policy_loss(inputs, distillation_logits.student_logits)
+        weighted_distillation_loss = self._compute_weighted_self_distillation_loss(
+            model,
+            inputs,
+            distillation_logits,
+        )
+        return policy_loss + weighted_distillation_loss
```
LeonEricsson, line 1522: On main (retained in this PR), the hybrid loss treats distillation as an additive auxiliary term:

`policy + λ·distillation`

https://github.com/huggingface/trl/blob/51c6d3ca31fb4cc80ff719c0844bbdfcd0feeefb/trl/experimental/sdpo/sdpo_trainer.py#L383

What's the motivation behind this structure? I couldn't find this variant in the paper or repo, so I'm assuming it's an empirical choice.

Section 4.5 of the paper instead uses a convex combination of policy and distillation:

`(1-w)·policy + w·distillation`

Should we consider switching to this variant in a follow-up PR? I don't think there is a "correct" answer, but I've seen the convex version more commonly used in recent RL+Distillation work

cc @kashif @cmpatino

trl/experimental/sdft/sdft_config.py
```diff
@@ -57,6 +260,64 @@ class SDFTConfig(SelfDistillationConfig):
...
+        if self.distillation_mode == "topk_logits":
+            if self.distillation_topk is None:
+                raise ValueError("`distillation_mode='topk_logits'` requires `distillation_topk` to be set.")
+        elif self.distillation_topk is not None:
+            raise ValueError("`distillation_topk` is only valid when `distillation_mode='topk_logits'`.")
+        if self.distillation_mode == "sampled_token" and self.distillation_alpha != 1.0:
+            raise ValueError(
+                "sampled-token distillation requires `distillation_alpha=1.0`. "
+                "Set `distillation_mode='full_logits'` or `distillation_mode='topk_logits'` to use other divergence settings."
+            )
```
qgallouedec: For the record, I'm really not a fan of this kind of validation. Because once you start writing them, the number of possible misconfigurations is just endless. For example, why don't we also check that `vllm_gpu_memory_utilization` is between 0.0 and 1.0? And why value errors but not type errors?

I think we should only check when there’s real added value for the user. For example, it’s pretty tricky to know how the generation batch size interacts with gradient accumulation, etc. (see below); in that case, it makes sense to have tests.

trl/experimental/sdft/loss_utils.py
```diff
@@ -0,0 +1,178 @@
...
+    if loss_type == "grpo":
+        loss = (per_token_loss * response_mask).sum(-1) / response_mask.sum(-1).clamp(min=1.0)
+        return loss.mean()
+    if loss_type == "bnpo":
+        return (per_token_loss * response_mask).sum() / response_mask.sum().clamp(min=1.0)
+    if loss_type == "dr_grpo":
+        return (per_token_loss * response_mask).sum() / (per_token_loss.size(0) * max_completion_length)
+    if loss_type == "dapo":
+        return (per_token_loss * response_mask).sum() / response_mask.sum().clamp(min=1.0)
+    raise ValueError(f"Unsupported loss_type: {loss_type}")
```
qgallouedec: I don't understand why we have different loss types here. what the original paper uses? all of them?

trl/experimental/sdft/sdft_trainer.py
```diff
@@ -52,28 +53,47 @@
...
+from ..self_distillation.loss_utils import (
+    aggregate_loss,
+    apply_importance_sampling_clipping,
+    compute_full_logit_self_distillation_loss,
+    compute_sampled_token_self_distillation_loss,
+    compute_topk_self_distillation_loss,
+    select_token_log_probs,
+)
+from ..self_distillation.prompt_utils import extract_last_user_text
+from ..self_distillation.teacher_sync import PEFTAdapterEMACallback, SyncTeacherModelCallback, is_pure_lora_training
```
qgallouedec: ideally, one experimental submodule should not depend on another experimental submodule. I'd recommend code duplication to avoid this kind of pattern

trl/experimental/sdft/sdft_config.py
```diff
@@ -36,10 +45,204 @@ class SDFTConfig(SelfDistillationConfig):
...
+        metadata={"help": "Number of optimization iterations per generated batch."},
+    )
+    loss_type: str = field(
+        default="dapo",
+        metadata={"help": "Policy loss aggregation. Supported: `grpo`, `bnpo`, `dr_grpo`, `dapo`."},
+    )
+    mask_truncated_completions: bool = field(
+        default=False,
+        metadata={"help": "Whether to exclude truncated completions from the loss."},
+    )
```
qgallouedec: same here. This is a technique introduced from grpo, why do we have it in self-distillation trainer?

trl/experimental/self_distillation/generation.py
```diff
@@ -0,0 +1,171 @@
...
+
+@dataclass
+class GenerationOutput:
+    """Token ids produced by a self-distillation generation backend."""
+
+    prompt_ids: list[list[int]]
+    completion_ids: list[list[int]]
+
+
+class Generation:
```
qgallouedec: I'm not convinced that we need this abstraction. A simple if/else (like in grpo) is probably sufficient

trl/experimental/sdft/sdft_config.py
LeonEricsson: i agree.

trl/experimental/sdft/loss_utils.py
LeonEricsson: The original code/paper uses grpo style normalization

`loss = ((loss * mask).sum(-1) / mask.sum(-1).clamp(min=1.0)).mean()`

If we want to match the paper exactly, we'd hardcode the GRPO-style normalization.

I chose to expose the variants because loss normalization has been a contentious topic, and it isn't inherent to GPRO. Any masked loss needs a reduction, and these self-distill completions share the same varying completion length property that motivates GRPO's variants. Although I don't think inheriting the names from GRPO as-is, is right

trl/experimental/sdft/sdft_config.py
LeonEricsson: Agree, doesn't make sense in the SDFT context. 

Does apply to SDPO though, which has an onpolicy RL component.

trl/experimental/sdft/sdft_trainer.py
LeonEricsson: gotchu. I've had the wrong tendency of wanting to find shared abstractions and methods instead of duplicating and isolating the trainers.

moved this into sdft+sdpo modules instead; inlining some stuff directly in the trainers

trl/experimental/self_distillation/generation.py
LeonEricsson: removed since the shared `self_distillation` module no longer exists. back to being inlined in each respective trainer.

trl/experimental/sdft/loss_utils.py
qgallouedec: I see, thanks. I lean towards keeping the original equation as it appears on paper, and will only extend it to other forms if we receive a request to do so.

> Although I don't think inheriting the names from GRPO as-is, is right

yes, that is definitely misleading, and that would be annoying to revert once adopted by the users

LeonEricsson: got it! removed and hardcoded the paper's loss aggregation variant.

currently tests fail because liger loss variant uses "bnpo" style normalization, which doesn't match with the now hardcoded non liger path. 

**edit: tests fixed**

trl/experimental/sdpo/sdpo_trainer.py
kashif: it was prob. a typo... we could make it convex or in a followup as you suggest after running experiments

LeonEricsson:
Thanks @kashif for ZeRO-3 fixes and Liger loss impl.

trl/experimental/sdft/sdft_trainer.py
```diff
@@ -409,88 +635,445 @@ def _generate_completion_ids(self, prompts: list[Any]) -> tuple[torch.Tensor, to
...
+        """
+        logits_to_keep = inputs["completion_ids"].size(1)
+        response_mask = self._build_self_distillation_response_mask(
+            inputs["completion_mask"], inputs.get("self_distillation_mask")
+        )
+
+        unwrapped_student = self.accelerator.unwrap_model(model)
+        unwrapped_teacher = self.accelerator.unwrap_model(self.teacher_model)
+
+        self.teacher_model.eval()
```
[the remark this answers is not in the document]
LeonEricsson: Confirmed. For PEFT setups the student forward pass could run in eval mode. Fixing by restoring `train()`-mode after teacher forward. 

EDIT: already fixed by kashif

kashif:
So next @LeonEricsson will test out the vllm teacher PR https://github.com/LeonEricsson/trl/pull/5  but can integrate it later

kashif (approved):

qgallouedec (approved):
thanks!

PR #5866, opened 2026-05-27
Align KTO with DPO: Support remove_unused_columns

albertvillanova:
Align KTO with DPO: Support remove_unused_columns.

This PR simplifies the handling of the `remove_unused_columns` argument in the KTO trainer and updates related test cases. The main change is to remove unnecessary settings of `remove_unused_columns`, ensuring that the argument is not modified internally and that the expected signature columns are set explicitly when needed.

Part of:
- #4786 

### Motivation

KTO was overriding `remove_unused_columns` to `False` at runtime (warning users to set it explicitly), because it lacked the plumbing to declare which dataset columns its collator needs. DPO handles this correctly via `_set_signature_columns_if_needed`, letting `remove_unused_columns` default to `True` and drop unused columns after preprocessing.

### Changes

**Trainer logic simplification:**

* Removed internal override from `KTOTrainer` that forcibly set `remove_unused_columns=False` and issued a warning, aligning the trainer's behavior with other trainers.

**Signature columns handling:**

* Added the `_set_signature_columns_if_needed` method to explicitly define the required signature columns for the KTO trainer when `remove_unused_columns` is `True`. This declares the six columns `DataCollatorForUnpairedPreference` reads: `prompt_ids`, `completion_ids`, `KL_completion_ids`, `label`, `reference_logps`, `reference_KL_logps`.

**Test cleanup and consistency:**

* Removed the `remove_unused_columns=False` argument from all test cases for KTO, as it is no longer required. 

---

> [!NOTE]
> **Medium Risk**
> Behavior change for users who relied on the old forced False or default True without signature columns; training could break if remove_unused_columns strips collator inputs, though tests cover the default path.
> 
> **Overview**
> **KTO** now respects `remove_unused_columns` like **DPO** instead of forcing it off at init.
> 
> `KTOTrainer` no longer sets `args.remove_unused_columns = False` or logs a warning when using the default unpaired collator. It adds `_set_signature_columns_if_needed` so that when column dropping is enabled, the six fields the collator needs (`prompt_ids`, `completion_ids`, `KL_completion_ids`, `label`, `reference_logps`, `reference_KL_logps`) are kept. KTO tests drop explicit `remove_unused_columns=False` from `KTOConfig` because that workaround is no longer required.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 931f064d95c233a14cdf0f33c1b83113ede3afca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (approved):
Good!

PR #5867, opened 2026-05-27
Raise if precompute_ref_log_probs with vision datasets in DPO

albertvillanova:
Raise for precompute_ref_log_probs with vision datasets in DPO.

This pull request adds a new validation to prevent the use of `precompute_ref_log_probs=True` with vision datasets in the `DPOTrainer`. It also introduces a corresponding unit test to ensure an informative error is raised in this scenario.

### Motivation

DPOTrainer silently accepted precompute_ref_log_probs=True together with a vision dataset. The trainer would initialize without complaint, run a full forward pass of the reference model over the entire dataset, and only then crash at the first training step with a cryptic `KeyError: 'ref_chosen_logps'`: after all the precomputation time had already been spent!

This combination is not supported: for vision-language models, all data processing is intentionally deferred to the collator rather than run upfront on the full dataset (`skip_prepare_dataset = self._is_vision_dataset`). This PR applies the same principle to reference log-prob precomputation and raises a clear ValueError at init, consistent with the existing guards for the Liger+precompute and keep_end+VLM combinations.

### Changes

Validation for vision datasets:

* Added a check in the `DPOTrainer` constructor to raise a `ValueError` if `precompute_ref_log_probs=True` is used with a vision dataset, since precomputing reference log probabilities is not supported for vision-language models.

Testing improvements:

* Added the test `test_precompute_ref_log_probs_raises_for_vision` for DPO to verify that the appropriate error is raised when this unsupported configuration is used.

### Question

Should `precompute_ref_log_probs=True` be supported for vision datasets in the future? If this is desirable, I am happy to open a follow-up PR.

CC: @qgallouedec 

---

> [!NOTE]
> **Low Risk**
> Adds an early config validation and a unit test; no change to supported training paths.
> 
> **Overview**
> **DPOTrainer** now rejects **`precompute_ref_log_probs=True`** when the training data is a vision dataset (image/images keys), raising a clear **`ValueError` at construction** instead of running reference precomputation and failing later with a missing **`ref_chosen_logps`** key.
> 
> This matches how VLMs defer tokenization and batching to the vision collator rather than upfront dataset preparation. A vision DPO test asserts the new error message.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 719fbd5216865336e9ee6a03b7ffaa3718bf2800. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_dpo_trainer.py
```diff
@@ -1184,3 +1184,20 @@ def test_train_vlm_keep_end_raises(self):
...
             )
+
+    @require_vision
+    def test_precompute_ref_log_probs_raises_for_vision(self):
+        dataset = load_dataset("trl-internal-testing/zen-image", "conversational_preference", split="train")
+        training_args = DPOConfig(
+            output_dir=self.tmp_dir,
+            max_length=None,
+            report_to="none",
+            use_cpu=True,
```
qgallouedec: I don't think we need this


```suggestion
```

qgallouedec (approved):
useful!

albertvillanova:
@qgallouedec, should `precompute_ref_log_probs=True` be supported for vision datasets in the future? If this is desirable, I am happy to open a follow-up PR.

qgallouedec:
I don't see how that could be support. The precomputation requires preprocessing, which takes up a huge amount of disk space. Unless we preprocess the data on the fly for the precomputation?

PR #5869, opened 2026-05-27
Fix: specify encoding="utf-8" when reading .jinja chat templates on Windows

ColebyPearson:
## Problem

On Windows, the default text-mode encoding is cp1252. Several of the bundled `.jinja` chat templates contain non-ASCII bytes that cp1252 cannot decode — most reproducibly `trl/chat_templates/deepseekv3.jinja`, which has byte `0x81` at offset 932. As a result, any user who runs `from trl import SFTTrainer` (or anything that transitively imports `trl.chat_template_utils`) on Windows without setting `PYTHONUTF8=1` hits:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 932: character maps to <undefined>
```

which then surfaces as:

```
RuntimeError: Failed to import trl.trainer.sft_trainer because of the following error (look up to see its traceback): 'charmap' codec can't decode byte 0x81 in position 932: character maps to <undefined>
```

I hit this while following the HuggingFace [smol-course](https://github.com/huggingface/smol-course) Unit 1 SFT recipe on a stock Windows 11 + Python 3.13 + TRL 1.5.0 install. `PYTHONUTF8=1` is a known workaround, but most users won't know to reach for it from the error message.

## Fix

Pass `encoding="utf-8"` explicitly to every `.read_text()` call in `trl/chat_template_utils.py` (38 in total — all the `*_chat_template` and `*_training_chat_template` module-level constants). UTF-8 is the encoding the `.jinja` files are actually saved as, so this is a behaviour-preserving no-op on Linux/macOS — the only effect is removing the Windows-default-encoding trap.

## Verification

- `git diff --stat` shows exactly 38 insertions + 38 deletions in `trl/chat_template_utils.py`, all of the form `).read_text() → ).read_text(encoding="utf-8")`.
- Locally reproduces: with the current `main`, `from trl import SFTTrainer` on Windows-default-encoding fails as above; with this patch applied, the import succeeds (verified with a Windows 11 / Python 3.13 / cp1252 default install).
- No new dependencies, no API changes.

---

> [!NOTE]
> **Low Risk**
> Encoding-only file reads at import time; no API or logic changes beyond fixing Windows decode failures.
> 
> **Overview**
> All bundled `.jinja` chat templates in `trl/chat_template_utils.py` are now loaded with **`encoding="utf-8"`** on every `Path.read_text()` call (inference and training template constants).
> 
> On Windows, the default locale encoding (e.g. cp1252) could raise **`UnicodeDecodeError`** when importing TRL (e.g. `SFTTrainer`) because some templates contain UTF-8 bytes that are not valid in that encoding. Linux/macOS behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3d44da847ca0dc97350b21e788968dbb4661154. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/chat_template_utils.py
+++ b/trl/chat_template_utils.py
@@ -574,13 +574,19 @@ qwen2_5_vl_training_chat_template = (_CHAT_TEMPLATES_DIR / "qwen2_5_vl_training.
 
 qwen3_training_chat_template = (_CHAT_TEMPLATES_DIR / "qwen3_training.jinja").read_text(encoding="utf-8")
 
-qwen3_instruct_2507_training_chat_template = (_CHAT_TEMPLATES_DIR / "qwen3_instruct_2507_training.jinja").read_text(encoding="utf-8")
+qwen3_instruct_2507_training_chat_template = (_CHAT_TEMPLATES_DIR / "qwen3_instruct_2507_training.jinja").read_text(
+    encoding="utf-8"
+)
 
 qwen3_vl_training_chat_template = (_CHAT_TEMPLATES_DIR / "qwen3_vl_training.jinja").read_text(encoding="utf-8")
 
-qwen3_5_nothink_training_chat_template = (_CHAT_TEMPLATES_DIR / "qwen3_5_nothink_training.jinja").read_text(encoding="utf-8")
+qwen3_5_nothink_training_chat_template = (_CHAT_TEMPLATES_DIR / "qwen3_5_nothink_training.jinja").read_text(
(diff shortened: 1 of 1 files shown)


qgallouedec:
we don't officially support Windows but this is mostly safe to merge

qgallouedec (approved):

PR #5873, opened 2026-05-27
Fix `loss_type="chunked_nll"` under DeepSpeed ZeRO-3

qgallouedec:
The chunked-CE path in `SFTTrainer` reads `lm_head.weight` (and `lm_head.bias`) directly to do its own per-chunk matmul, bypassing the DeepSpeed pre-forward hooks that normally allgather ZeRO-3 partitioned params. On non-owning ranks the weight is a 0-element shard and the matmul crashes.

To see the traceback, see https://github.com/huggingface/trl/actions/runs/26475700871/job/77960573388?pr=5846#step:9:60

This adds a tiny `_maybe_gather_lm_head_ctx(w, b)` helper that wraps the matmul in `deepspeed.zero.GatheredParameters` when the param is ZeRO-3-partitioned, and is a no-op otherwise (DDP, FSDP2, single-GPU, or when the param is already `AVAILABLE`, e.g. tied embeddings where `embed_tokens` shares the weight). 

With `use_reentrant=False` the checkpoint recompute re-enters the context during backward, so the gather happens on both passes.

Also gathers in the `n_valid == 0` fallback branch which reads `lm_head_weight.float().sum()` for the dummy loss.

---

> [!NOTE]
> **Medium Risk**
> Touches distributed training loss math and DeepSpeed parameter gathering; scope is narrow to chunked_nll but affects multi-GPU ZeRO-3 SFT runs.
> 
> **Overview**
> Fixes **`loss_type="chunked_nll"`** under **DeepSpeed ZeRO-3** by gathering partitioned **`lm_head`** weights before the custom chunked matmul.
> 
> The chunked cross-entropy path reads **`lm_head.weight`** (and bias) directly, so ZeRO-3 ranks can see empty shards and crash. A new **`_maybe_gather_lm_head_ctx`** wraps those reads in **`deepspeed.zero.GatheredParameters`** when needed; it stays a no-op for DDP, FSDP2, single GPU, or params already **`AVAILABLE`** (e.g. tied embeddings).
> 
> The same gather is applied in the **all-masked** dummy-loss branch so backward still touches head params.
> 
> Adds distributed **`test_sft_chunked_nll`** (DDP, ZeRO-2/3, FSDP2) mirroring **`test_sft`** with **`--loss_type chunked_nll`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 666c92cb52f4896d1778476424ef473b83ef4e24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -86,10 +86,27 @@ class _ChunkedCELMHeadOutput(CausalLMOutputWithPast):
     aux_loss: torch.Tensor | None = None
 
 
+def _maybe_gather_lm_head_ctx(w, b):
+    # Allgather ZeRO-3 partitioned `lm_head` weight/bias for the chunked matmul. No-op if not ZeRO-3, or if the
+    # param is already gathered (tied embeddings: `embed_tokens` shares the weight and keeps it `AVAILABLE`, so
+    # partitioning on our exit would collide with its active-submodule tracking).
+    params = [w] if b is None else [w, b]
+    if not any(hasattr(p, "ds_id") for p in params):
+        return contextlib.nullcontext()
+    from deepspeed.runtime.zero.partition_parameters import ZeroParamStatus
+
+    if all(p.ds_status == ZeroParamStatus.AVAILABLE for p in params):
+        return contextlib.nullcontext()
+    import deepspeed
+
+    return deepspeed.zero.GatheredParameters(params)
+
+
 def _chunk(h, w, b, lbl, logit_scale, final_logit_softcapping):
-    logits = h.float() @ w.float().t()
-    if b is not None:
-        logits = logits + b.float()
+    with _maybe_gather_lm_head_ctx(w, b):
+        logits = h.float() @ w.float().t()
+        if b is not None:
+            logits = logits + b.float()
     if logit_scale != 1.0:
         logits = logits * logit_scale
     if final_logit_softcapping is not None:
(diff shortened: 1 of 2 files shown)


trl/trainer/sft_trainer.py
```diff
@@ -86,10 +86,29 @@ class _ChunkedCELMHeadOutput(CausalLMOutputWithPast):
...
+    if not is_deepspeed_zero3_enabled():
+        return contextlib.nullcontext()
+
+    import deepspeed
+    from deepspeed.runtime.zero.partition_parameters import ZeroParamStatus
+
+    params = [w] if b is None else [w, b]
+    if all(p.ds_status == ZeroParamStatus.AVAILABLE for p in params):
+        return contextlib.nullcontext()
+    return deepspeed.zero.GatheredParameters(params)
```
AmineDiro, line 104: how does this work in backward ?

qgallouedec: No gathered weight has to survive forward→backward. The chunk matmul runs inside `torch.utils.checkpoint.checkpoint(_chunk, ..., use_reentrant=False)` https://github.com/huggingface/trl/blob/70d9081d0ef5e38d54be29f9e3de4a11b9dd4828/trl/trainer/sft_trainer.py#L212-L221
and `_maybe_gather_lm_head_ctx` wraps the matmul *inside* `_chunk`. With `use_reentrant=False` the checkpoint **recomputes `_chunk` in backward**, so the gather context re-enters at backward time and the weight is in the right state exactly when the chunk's gradient is computed.

The re-entered context is correct in both states:
- **resident** (tied embeddings / already engine-gathered) → no-op.
- **partitioned** → re-gathers. `GatheredParameters` releases on exit, so a weight gathered in a forward chunk is `NOT_AVAILABLE` again by backward

I confirmed on Qwen3-0.6B (ZeRO-3, 2 ranks): the gather context fires in both forward and backward and training runs fine (tied head, so backward hits the `AVAILABLE` no-op)

And the partitioned path in isolation (bare `zero.Init`, weight `NOT_AVAILABLE`) is confirmed as well.

kashif:
Smoke tested this on 2 GPUs. Ran sft.py with `--loss_type chunked_nll` across ddp, zero2, zero3 and fsdp2, and all four train to the same loss now (~11.9). zero3 was the one crashing before so this sorts it out. Gather context matches what we already do for the grpo/liger lm_head path. LGTM 👍

PR #5882, opened 2026-05-29
`DPOTrainer`: eagerly delete intermediate logits tensors to reduce peak memory

flutist:
### PR Title

**`DPOTrainer`: eagerly delete intermediate logits tensors to reduce peak memory**

### PR Description

**What does this PR do?**

Adds 4 `del` statements in `DPOTrainer._compute_loss` to eagerly release large intermediate tensors as soon as they are no longer needed, reducing peak GPU memory during training.

The logits tensor shape is `[2*batch_size, seq_len, vocab_size]` — for a model with vocab_size ~150k this is the single largest allocation in `_compute_loss`. The original code keeps multiple copies alive simultaneously (policy `outputs`, `shift_logits`, `ref_outputs`, `ref_shift_logits`), even though each is only needed briefly.

**Changes:**

- `del outputs` — after extracting `shift_logits` via `.contiguous()` (skipped when `return_outputs=True`)
- `del ref_outputs` — after extracting `ref_shift_logits` via `.contiguous()`
- `del ref_shift_logits` — after `selective_log_softmax`
- `del shift_logits` — after metrics logging is complete

**Memory benchmark**

Benchmark script (not included in this PR):

```python
import gc
import torch

if not torch.cuda.is_available():
    print("ERROR: No CUDA GPU available.")
    exit(1)

gc.collect()
torch.cuda.empty_cache()

dataset = load_dataset("trl-lib/ultrafeedback_binarized", split="train[:200]")

args = DPOConfig(
    output_dir="/tmp/dpo_peak_bench",
    max_steps=3,
    max_length=1024,
    gradient_checkpointing=True,
    bf16=True,
    report_to="none",
    logging_steps=1,
)

trainer = DPOTrainer(model="Qwen/Qwen2.5-0.5B-Instruct", args=args, train_dataset=dataset)
trainer.train()

peak_mb = torch.cuda.max_memory_allocated() / 1024**2
print(f"\nPeak allocated: {peak_mb:.1f} MB")
```

Results (Qwen2.5-0.5B-Instruct, batch_size=4, max_length=1024, bf16, gradient_checkpointing=True):

| | Peak allocated |
|---|---|
| Before | 28751 MB |
| After | 22807 MB |
| **Saved** | **5944 MB (−20.7%)** |

**No functional regression** — loss and all metrics are consistent across both runs:

| Step | Loss (before) | Loss (after) |
|------|--------------|-------------|
| 1 | 0.6931 | 0.6931 |
| 2 | 0.6824 | 0.6558 |
| 3 | 0.6605 | 0.6978 |

(Minor differences are expected due to non-deterministic operations; step 1 is identical, confirming no computational change.)

---

> [!NOTE]
> **Low Risk**
> Lifecycle-only memory tweaks in the standard loss path; `return_outputs=True` still retains `outputs` for eval/prediction_step.
> 
> **Overview**
> **`DPOTrainer._compute_loss`** now drops large intermediate tensors as soon as they are no longer needed, so policy/reference forward outputs and full **vocab-sized logits** are not all resident at once during a training step.
> 
> After building **`shift_logits`**, the policy **`outputs`** object is **`del`**’d when **`return_outputs`** is false (training path). On the live reference forward path, **`ref_outputs`** and **`ref_shift_logits`** are freed right after **`selective_log_softmax`**. **`shift_logits`** is removed only after entropy, optional WPO weighting, and chosen/rejected logit metrics have finished using it.
> 
> This is a **peak-memory** change only: loss math and logging order stay the same, with **`outputs`** kept when eval needs **`return_outputs=True`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a57a63633e1185f2fd39d94d43b0dd86bcd709e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


flutist:
@qgallouedec Hi! Could you take a look at this PR when you have a chance?

qgallouedec:
Thanks, interesting results! I don't quite like playing with del but in this case it seems useful! I'll reproduce the results myself before merging!

flutist:
@qgallouedec  is there anything I can do to help make the review easier? For example, adding more tests, providing additional data, or clarifying something in the code. Just let me know — happy to help.

albertvillanova (commented):
Thanks for your contribution and for looking into DPO memory usage: this is a real and impactful area to optimize.

That said, I'd like to suggest a different approach before merging this. Explicit `del` statements are not idiomatic PyTorch memory management: their effect depends entirely on whether the deleted name is the last reference to the underlying storage. 

In this PR, the `del shift_logits` on line 1443 is a good illustration of the problem: `chosen_logits` and `rejected_logits` are views created via `.detach().chunk()`, so they share the same storage as `shift_logits`. Deleting the name frees nothing. 

The `del ref_shift_logits` and `del outputs` changes are more defensible: the `.contiguous()` calls there do create independent copies. However, their correctness is silently coupled to those `.contiguous()` calls remaining in place, a fragile invariant with no mechanism to enforce it.

The root cause of the excess allocation is not that tensors are held too long, but that `.contiguous()` forces unnecessary materialized copies of what would otherwise be cheap views. PR #5926 targets exactly that.

I would recommend closing this PR and either contributing to #5926 or waiting for it to land. If after that merge there are still specific allocation hotspots you have identified (for instance, the `ref_outputs` object holding unnecessary buffers beyond just logits), a focused follow-up with a benchmark and a narrow diff would be very welcome.

PR #5890, opened 2026-05-30
[GRPO] Match the Liger dapo/cispo/vespo normalizer to the non-Liger path

### LINKED ISSUE #5619: DAPO loss is not normalized by gradient_accumulation_steps
Hi, I noticed that the DAPO loss is not normalized by current_gradient_accumulation_steps and am not sure why.  

```

        if self.loss_type in ["grpo", "sapo"]:
            loss = ((per_token_loss * mask).sum(-1) / mask.sum(-1).clamp(min=1.0)).mean()
            normalizer = self.current_gradient_accumulation_steps if mode == "train" else 1.0  # no accum in eval
            loss = loss / normalizer
        elif self.loss_type in ["cispo", "dapo", "vespo"]:
            normalizer = inputs["num_items_in_batch"] / self.accelerator.num_processes
            loss = (per_token_loss * mask).sum() / normalizer

```
in https://github.com/huggingface/trl/blob/a08e7139f933b770177fc2abc0b43118e26b260b/trl/trainer/grpo_trainer.py#L2547-L2562

kashif:
The Liger and non-Liger GRPO paths normalize the dapo/cispo/vespo loss differently.

The non-Liger path divides by `num_items_in_batch / num_processes` (the total tokens across the whole generation batch) and leaves gradient accumulation alone. The Liger path didn't pass `num_items_in_batch` at all, so it fell back to the current micro-batch's mask, and then divided again by `gradient_accumulation_steps`. So the two drift apart once you have grad accumulation with uneven completion lengths.

This passes `num_items_in_batch` through to the Liger loss and drops the extra grad-accum division for those three loss types. Everything else keeps dividing as before.

On top of that, when `steps_per_generation != gradient_accumulation_steps` the non-Liger path now rescales by `current_gradient_accumulation_steps / steps_per_generation` so the accumulated loss lands on the per-window token-mean (fixes #5619). The latest commit applies the same rescale to the Liger path so both paths stay in parity, per behroozazarkhalili's review on #6024.

It's the last bit of the parity work tracked in linkedin/Liger-Kernel#1082 (the IS correction, tool_mask, delta and bias-correction KL were already sorted).

Added a test that runs dapo with grad accumulation > 1 and checks the arg gets forwarded and the loss is rescaled by `steps_per_generation / gradient_accumulation_steps`.

Waiting for #6768 to be merged first
 

---

> [!NOTE]
> **Medium Risk**
> Changes gradient scaling for Liger + DAPO/CISPO/VESPO training (including grad accumulation and uneven completion lengths), which can shift effective learning rates; other loss types keep the prior normalizer.
> 
> **Overview**
> **Aligns Liger GRPO loss scaling with the existing `_compute_loss` behavior for `dapo`, `cispo`, and `vespo`.**
> 
> The Liger kernel call now receives **`num_items_in_batch`** so token-mean normalization uses the full generation window instead of the current micro-batch mask. For those three loss types, **`compute_liger_loss` no longer divides only by gradient accumulation steps**; it divides by **`current_gradient_accumulation_steps / steps_per_generation`** in training (unchanged for other loss types), matching the per-window rescale in the non-Liger path and addressing **#5619**.
> 
> A new **`test_compute_liger_loss_dapo_normalizer`** (Liger-only) trains with `loss_type="dapo"` where `steps_per_generation` ≠ `gradient_accumulation_steps` and asserts each step’s final loss equals the raw Liger output times **`steps_per_generation / gradient_accumulation_steps`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1657d0ab809a740155044a0ce9a7312b007a438b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kashif:
Pushed the per-window rescale per behroozazarkhalili's review on #6024. The Liger and non-Liger paths now end at the same effective normalizer:

`num_items_in_batch / num_processes * current_gradient_accumulation_steps / steps_per_generation`

The test asserts the rescale (`steps_per_generation / gradient_accumulation_steps`) and now uses `steps_per_generation=4, gradient_accumulation_steps=2` so the rescale is non-trivial. Merging the latest main in the process (resolved the liger-pin conflict by keeping `>=0.8.1` — the upstream temp hotfix `<0.8.1` from #6517 is the broken-pin workaround that this PR makes obsolete once linkedin/Liger-Kernel#1316 lands).

pyproject.toml
```diff
@@ -57,7 +57,7 @@ kernels = [  # transformers renamed the "hub-kernels" extra to "kernels" in 5.1.
     "transformers[hub-kernels]",  # transformers < 5.1.0
 ]
 liger = [
-    "liger-kernel>=0.8.0,!=0.8.1"  # upstream PR https://github.com/linkedin/Liger-Kernel/pull/1316
+    "liger-kernel>=0.8.1"
```
kashif:
```suggestion
    "liger-kernel>=0.8.2"
```

pyproject.toml
```diff
@@ -111,7 +111,7 @@ dev = [
     "transformers[kernels]",  # transformers >= 5.1.0
     "transformers[hub-kernels]",  # transformers < 5.1.0
     # liger
-    "liger-kernel>=0.8.0,!=0.8.1",  # upstream PR https://github.com/linkedin/Liger-Kernel/pull/1316
+    "liger-kernel>=0.8.1",
```
kashif:
```suggestion
    "liger-kernel>=0.8.2",
```

trl/import_utils.py
```diff
@@ -21,7 +21,7 @@
 from packaging.version import Version
 
 
-LIGER_KERNEL_MIN_VERSION = "0.8.0"
+LIGER_KERNEL_MIN_VERSION = "0.8.1"
```
kashif:
```suggestion
LIGER_KERNEL_MIN_VERSION = "0.8.2"
```

qgallouedec (approved):
The fix looks good.

I would `drop test_compute_liger_loss_dapo_normalizer` though. It patches two internals to capture intermediate values and then asserts the same formula the implementation uses, so it can only fail on a refactor, not on a real regression.

If we want a guard here, the meaningful check is liger vs non-liger parity: run dapo with `use_liger_kernel` True and False on the same seed and compare per-step losses, which could be folded into test_train_loss_types. Fine as a follow-up, or without a new test at all, but the easiest is to just remove imo.

kashif:
will follow up

PR #5893, opened 2026-05-31
`AsyncGRPOTrainer`: add `model_init_kwargs` support

rycerzes:
# What does this PR do?

Adds `model_init_kwargs` support to `AsyncGRPOTrainer`, matching the existing behavior in `GRPOTrainer`.

Currently, `AsyncGRPOTrainer` hardcodes `AutoModelForCausalLM.from_pretrained(model, device_map=None, dtype=torch.float32)`, which prevents passing custom kwargs like `attn_implementation` or `dtype`. This PR:

- Adds `model_init_kwargs` field to `AsyncGRPOConfig` (with `_VALID_DICT_FIELDS` for dict serialization support)
- Replaces direct `AutoModelForCausalLM` loading with `create_model_from_path(model, **model_init_kwargs)`, which infers the correct architecture class from the model config (important for models like Qwen3.5 that use `Qwen3_5ForConditionalGeneration` instead of a standard CausalLM)
- Enforces `device_map=None` for FSDP2 compatibility

**Changes:**
- `async_grpo_config.py`: add `model_init_kwargs` field + docstring + `_VALID_DICT_FIELDS`
- `async_grpo_trainer.py`: replace `AutoModelForCausalLM` with `create_model_from_path`, update imports

**Tests:**
- `test_model_init_kwargs`: verifies that `model_init_kwargs={"dtype": "bfloat16"}` results in bfloat16 model parameters

Closes part of https://github.com/huggingface/trl/issues/5831

## Before submitting

- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request), Pull Request section?
- [x] Was this discussed/approved via a GitHub issue? Please add a link to it if that's the case.
- [x] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

## AI writing disclosure

- [ ] No AI usage: the PR was written entirely by a human.
- [x] AI-assisted: some parts were suggested or improved by AI, but the PR was written and reviewed by a human.
- [ ] AI-generated: the PR was mostly or fully generated by an AI tool.

## Who can review?

@qgallouedec

---

> [!NOTE]
> **Low Risk**
> Small, localized change to model instantiation with an existing shared helper and a focused unit test; no auth or rollout pipeline logic changes.
> 
> **Overview**
[...]

Issue #5831, opened 2026-05-25
Add missing AsyncGRPOTrainer features present in GRPOTrainer


rycerzes:
### Feature request

Fill feature gaps in `AsyncGRPOTrainer` that are already present in `GRPOTrainer` but missing from the async variant.

### Motivation

The sync `GRPOTrainer` already supports several features that `AsyncGRPOTrainer` lacks. These gaps limit the usability of `AsyncGRPOTrainer`:

1. **ProcessorMixin handling** - `GRPOTrainer` uses `AutoProcessor` and extracts the tokenizer from `ProcessorMixin` instances, supporting VLM-style models without vision inputs. `AsyncGRPOTrainer` only uses `AutoTokenizer`.
2. **PEFT/LoRA support** - `GRPOTrainer` has a `peft_config` parameter, wraps the model with `get_peft_model`, and handles gradient checkpointing for PEFT models. `AsyncGRPOTrainer` has none of this.
3. **Sampling parameters** - `GRPOTrainer` supports `top_p`, `top_k`, `min_p`, `repetition_penalty` from config. `AsyncGRPOTrainer` does not plumb these to the vLLM rollout worker.
4. **`model_init_kwargs`** - `GRPOTrainer` supports `model_init_kwargs` for passing custom kwargs (e.g., `attn_implementation`, `dtype`) when loading the model from a string. `AsyncGRPOTrainer` lacks this.

### Your contribution

I have implemented all the changes in this [branch](https://github.com/rycerzes/trl/tree/feat/async-grpo-parity). Will draft a PR soon.

qgallouedec:
thanks! I think the best approach would be to decouple these changes. They all come with their own complexity, and reviewing one by one will be way easier

rycerzes:
Cool! will decouple and raise the PRs

qgallouedec:
Hey @rycerzes, thanks for splitting this up so cleanly, the decoupling worked great.

Quick status so we can retire this umbrella issue in favor of the individual PRs: 
1. `model_init_kwargs` already landed on main independently `AsyncGRPOConfig` has the field so #5893 was closed as done.
2. The remaining three are live and just waiting on review: #5894 (sampling params), #5895 (`ProcessorMixin` handling), #5896 (PEFT/LoRA).

Since every item now has its own tracking PR, I'll close this issue and we'll follow the rest there. Thanks again!

PR #5895, opened 2026-05-31
`AsyncGRPOTrainer`: add `ProcessorMixin` handling

rycerzes:
# What does this PR do?

Adds `ProcessorMixin` support to `AsyncGRPOTrainer`, matching the existing behavior in `GRPOTrainer`.

Currently, `AsyncGRPOTrainer` only uses `AutoTokenizer` when no `processing_class` is provided, and only accepts `PreTrainedTokenizerBase` instances. This fails for models like Qwen3.5 that use a processor (e.g., `Qwen3_5ForConditionalGeneration` returns a `ProcessorMixin` from `AutoProcessor`). 
#### This PR:
- Uses `AutoProcessor.from_pretrained()` instead of `AutoTokenizer.from_pretrained()` when no `processing_class` is provided
- Accepts both `PreTrainedTokenizerBase` and `ProcessorMixin` as `processing_class`
- Extracts the tokenizer from `ProcessorMixin` instances (`processing_class.tokenizer`)
- Passes the extracted tokenizer (not the processor) to `super().__init__()` and `AsyncRolloutWorker`
- Normalizes pad token on the tokenizer (`pad_token ← eos_token` when not set)

**Changes:**
- `async_grpo_trainer.py`: update imports (`AutoProcessor`, `ProcessorMixin`), update `processing_class` type hint and docstring, add processor/tokenizer extraction logic

**Tests:**
- `test_processor_mixin_handling`: passes an `AutoProcessor` instance as `processing_class` and verifies the trainer extracts a `PreTrainedTokenizerBase`
- `test_auto_processor_when_none`: omits `processing_class` entirely and verifies the trainer auto-loads via `AutoProcessor` with a valid pad token

Closes part of https://github.com/huggingface/trl/issues/5831

## Before submitting

## AI writing disclosure

## Who can review?

@qgallouedec

---

> [!NOTE]
> **Medium Risk**
> Touches default preprocessing and tokenizer wiring for all AsyncGRPO runs (AutoProcessor vs AutoTokenizer), which could affect text-only models if processor loading behaves differently, though rollout paths still use the extracted tokenizer.
> 
> **Overview**
> **`AsyncGRPOTrainer`** now accepts vision/multimodal processors the same way as **`GRPOTrainer`**, so models like Qwen3.5 that ship an **`AutoProcessor`** instead of a bare tokenizer can train without manual wiring.
> 
> When **`processing_class`** is omitted, the trainer loads **`AutoProcessor.from_pretrained`** (with left truncation/padding) instead of **`AutoTokenizer`**. Callers may pass either **`ProcessorMixin`** or **`PreTrainedTokenizerBase`**; the trainer keeps the full processor on **`processing_class`** for the base trainer while routing rollout, collation, and pad-token setup through an internal **`_tokenizer`** extracted from **`processing_class.tokenizer`** when needed. **`AsyncRolloutWorker`** and **`DataCollatorForRollout`** receive **`pad_token_id`** from that tokenizer, not the processor object.
> 
> Tests add **`test_train_vlm_text_only_data`**: end-to-end async GRPO on a tiny Qwen3.5 VLM with text-only prompts, asserting **`processing_class`** stays a processor, training produces loss, language weights update, and **`model.visual`** weights stay frozen. The rollout stub accepts processors and handles prompt-only datasets without a **`completion`** column.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ba094a15fad857917fc185583cb9d91b48b3766. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/experimental/test_async_grpo_trainer.py
+++ b/tests/experimental/test_async_grpo_trainer.py
@@ -21,9 +21,11 @@ import queue
 import numpy as np
 import pytest
 import torch
+import transformers
 from accelerate import PartialState
 from datasets import load_dataset
-from transformers import AutoTokenizer
+from packaging.version import Version
+from transformers import AutoProcessor, AutoTokenizer, ProcessorMixin
 from transformers.testing_utils import torch_device
 
 import trl.experimental.async_grpo.async_rollout_worker as worker
@@ -46,7 +48,7 @@ from trl.experimental.async_grpo.async_rollout_worker import (
     _SampleBuilder,
 )
 
-from ..testing_utils import TrlTestCase, is_ampere_or_newer
+from ..testing_utils import TrlTestCase, is_ampere_or_newer, require_vision
 
 
 def dummy_reward_func(completions, **kwargs):
@@ -57,20 +59,26 @@ class _StubRolloutWorker:
     """Minimal rollout worker stub for testing the trainer in isolation."""
 
     def __init__(
-        self, tokenizer, dataset, num_generations: int = 8, samples_per_weight_sync: int = 10, fork_k: int = 1
+        self, processing_class, dataset, num_generations: int = 8, samples_per_weight_sync: int = 10, fork_k: int = 1
     ):
         self.rollout_buffer = queue.Queue()
         self._samples_per_weight_sync = samples_per_weight_sync
         self._model_version = 0
         self._fork_k = fork_k
+        tokenizer = processing_class.tokenizer if isinstance(processing_class, ProcessorMixin) else processing_class
         self._sample_iter = self._make_sample_iter(tokenizer, dataset, num_generations)
 
     def _make_sample_iter(self, tokenizer, dataset, num_generations):
         for group_id, row in enumerate(itertools.cycle(dataset)):
-            completions = [
-                [{"role": "assistant", "content": f"{row['completion'][0]['content']} {idx}"}]
-                for idx in range(num_generations)
-            ]
+            if "completion" in row:
+                completions = [
+                    [{"role": "assistant", "content": f"{row['completion'][0]['content']} {idx}"}]
+                    for idx in range(num_generations)
+                ]
+            else:
+                completions = [
+                    [{"role": "assistant", "content": f"completion {idx}"}] for idx in range(num_generations)
+                ]
             prompt_completions = [row["prompt"] + completion for completion in completions]
             prompt_ids = tokenizer.apply_chat_template(
                 row["prompt"], tokenize=True, add_generation_prompt=True, return_dict=False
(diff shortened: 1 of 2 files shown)


qgallouedec:
at this point I don't know if it's very useful to integrate processor, a the asynchronous grpo trainer doesn't support vlm yet.

tests/experimental/test_async_grpo_trainer.py
```diff
@@ -137,3 +137,46 @@ def test_train(self):
...
+        trainer = AsyncGRPOTrainer(
+            model=model_id,
+            reward_funcs=dummy_reward_func,
+            args=training_args,
+            train_dataset=dataset,
+            rollout_worker=_StubRolloutWorker(AutoTokenizer.from_pretrained(model_id), dataset, num_generations=8),
+        )
+
+        assert isinstance(trainer.processing_class, PreTrainedTokenizerBase)
+        assert trainer.processing_class.pad_token is not None
```
qgallouedec: in this case, it would be better to have a test that looks like a mix of  https://github.com/huggingface/trl/blob/bf80d705b4976ffd929573ce69dd461854f55c8d/tests/test_sft_trainer.py#L1820-L1842

and 

https://github.com/huggingface/trl/blob/bf80d705b4976ffd929573ce69dd461854f55c8d/tests/test_grpo_trainer.py#L1959-L1984

qgallouedec: something like this


```python
def test_train_vlm_text_only_data(self):
        model_id = "trl-internal-testing/tiny-Qwen3_5ForConditionalGeneration-Think"
        dataset = load_dataset("trl-internal-testing/zen", "conversational_prompt_only", split="train")

        training_args = AsyncGRPOConfig(
            output_dir=self.tmp_dir,
            learning_rate=0.1,  # use higher lr because gradients are tiny and default lr can stall updates
            per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
            max_completion_length=8,  # reduce the completion length to reduce memory usage
            vllm_server_timeout=5.0,  # short timeout so test fails fast if queue runs dry
            report_to="none",
        )
        trainer = AsyncGRPOTrainer(
            model=model_id,
            reward_funcs=dummy_reward_func,  # unused: the stub pre-computes rewards, but the trainer requires this argument
            args=training_args,
            train_dataset=dataset,
            rollout_worker=_StubRolloutWorker(AutoProcessor.from_pretrained(model_id), dataset, num_generations=3),
        )

        trainer.train()

        assert trainer.state.log_history[-1]["train_loss"] is not None

        # Check that the params have changed
            if n.startswith("model.visual"):
                torch.testing.assert_close(param, new_param, rtol=1e-12, atol=1e-12, msg=f"Param {n} is updated")
            else:
                assert not torch.equal(param, new_param), f"Param {n} is not updated"
        
```

rycerzes: Yes, I will add the qwen 3.5 test as suggested, but since it depends on the model-loading change from the model_init_kwargs PR #5893, I will be stacking this PR on top of it.

rycerzes:
> at this point I don't know if it's very useful to integrate processor, a the asynchronous grpo trainer doesn't support vlm yet.

Yes I agree, this isn’t for full VLM support. The scope is text-only training for processor-backed models like Qwen3.5

Updated to stack on #5893, tests passing :)

qgallouedec:
Thanks @rycerzes, and sorry for the long silence, you answered back in June and this sat.

Your text-only framing is right and I do want this one, but it needs a rebase: the async module moved a lot since your base (38 commits, +2840/-603) and all three files conflict.

Please drop the `model_init_kwargs`/`create_model_from_path` half on the way: the config field landed on main independently (that's why I closed #5893), and `create_model_from_path` would drop the explicit `attn_implementation="kernels-community/flash-attn3"` that padding-free async training needs.
Keep the processor block as-is (it mirrors `grpo_trainer.py` exactly) but match main's call site: `AutoProcessor.from_pretrained(get_config_model_id(model.config), truncation_side="left", padding_side="left", trust_remote_code=args.trust_remote_code)`,
and switch the two remaining `processing_class` uses (`async_grpo_trainer.py:820` rollout worker, `:893` `pad_token_id`) to `self._tokenizer`.

One thing to work through: with `AutoModelForCausalLM` retained, does `tiny-Qwen3_5ForConditionalGeneration-Think` still load for your new test? If not we need the architecture-inference path that preserves the flash-attn3 kwarg (the separable piece I flagged on #5893).

Ignore the Bugbot note about `super().__init__()`: your last commit already keeps the processor on `self.processing_class`, which is what we want.

rycerzes (commented):
@qgallouedec Rebased, dropped the `model_init_kwargs` half. `async_grpo_config.py` untouched; trainer keeps `AutoModelForCausalLM` so `attn_implementation="kernels-community/flash-attn3"` survives — main already threads `model_init_kwargs` through `from_pretrained`. Processor block matches `grpo_trainer.py`'s call site; both remaining `processing_class` uses go through `self._tokenizer`.

Re-applied on top of main instead of rebasing through conflicts. Same two commits.

Your open question: yes, it loads. `AutoModelForCausalLM` → `Qwen3_5ForCausalLM`, `AutoProcessor` → `Qwen3VLProcessor`. No architecture inference needed, #5893 stays out of scope.

Two test fixes:

1. Missing `weight_transfer=_StubWeightTransfer()` and `token_budget=256` (without it the trainer hits `wait_for_server_ready()` against a real vLLM server).
2. Hang in `queue.get()`: `_StubRolloutWorker` buffers 10 and refills only on weight sync, but `TokenBudgetBatcher` wants ~`token_budget / sample_length` per row — ~11 for Qwen3.5's ~23-token samples, one over the buffer. Set `samples_per_weight_sync=64`.

PR #5896, opened 2026-05-31
`AsyncGRPOTrainer`: add PEFT/LoRA support

rycerzes:
# What does this PR do?

Adds PEFT/LoRA support to `AsyncGRPOTrainer`, matching the existing behavior in `GRPOTrainer`.

Currently, `AsyncGRPOTrainer` has no `peft_config` parameter and cannot train with LoRA adapters. This PR adds full PEFT support including the weight sync merge/unmerge cycle required for correct vLLM weight transfer.

**Changes:**
- `async_grpo_trainer.py`:
  - Add `peft_config` parameter to `__init__` with validation
  - Wrap model with `get_peft_model()` when `peft_config` is provided
  - Enable input gradients for gradient checkpointing compatibility with PEFT (workaround for [transformers#42489](https://github.com/huggingface/transformers/issues/42489))
  - PEFT-aware weight metadata collection: iterate base model params (`model.base_model.model`), strip `.base_layer` suffixes, skip adapter-only params — so vLLM receives base model weight names
  - PEFT-aware `_streaming_iter`: iterate base model params after merge so vLLM gets full (merged) weights
  - Merge/unmerge cycle in `_sync_weight`: `merge_adapter()` before weight transfer, `unmerge_adapter()` after — so vLLM gets merged weights while the trainer continues training with separate adapters
- `async_grpo_config.py`: add `model_init_kwargs` field (this PR is stacked on PR #5893)

**Note:** This PR includes the `model_init_kwargs` commit from PR #5893 as a base, since PEFT support requires `create_model_from_path` for correct model loading. Once PR #5893 merges, this PR will be rebased onto `main`.

**Tests:**
- `test_train_peft_config`: full training loop with `LoraConfig()` — verifies LoRA params change while base params stay frozen
- `test_peft_weight_metadata_excludes_adapters`: monkeypatches the worker to capture `weight_names` and verifies no `lora`-prefixed names leak into vLLM weight metadata

Closes part of https://github.com/huggingface/trl/issues/5831

## Before submitting

## AI writing disclosure

## Who can review?

@qgallouedec

---

> [!NOTE]
> **Medium Risk**
> Changes the async training hot path (weight sync merge/unmerge and parameter naming for vLLM); mistakes could desync rollouts from the trainer, but behavior is covered by new PEFT tests.
> 
> **Overview**
> Adds **PEFT/LoRA** training to `AsyncGRPOTrainer`, aligned with `GRPOTrainer`, plus optional **`model_init_kwargs`** on `AsyncGRPOConfig` for `from_pretrained` when the model is passed as a hub id.
> 
> **Trainer:** New `peft_config` wraps the base model via `get_peft_model`, with validation and gradient-checkpointing input-grad workaround. Loading switches from direct `AutoModelForCausalLM.from_pretrained` to `create_model_from_path` (honoring `model_init_kwargs`, forcing `device_map=None` for FSDP2). QLoRA paths upcast trainable adapter weights to **bfloat16**.
> 
> **vLLM weight sync:** Weight metadata and `_streaming_iter` strip PEFT prefixes, skip LoRA-only and `original_module` tensors, and normalize `modules_to_save` names so vLLM sees base checkpoint names. `_sync_weight` **merges** adapters before transfer and **unmerges** in a `finally` block so rollouts get merged weights while training stays on separate adapters.
> 
> **Tests:** `@require_peft` coverage for LoRA training (frozen base), `model_init_kwargs` dtype, QLoRA bf16 casting, and metadata/streaming name hygiene.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00477b193eda9b8e994cf9fc6c543b15de3804f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -601,6 +644,10 @@ def _sync_weight(self):
         self.accelerator.wait_for_everyone()
         t_barrier = time.time()
 
+        # Merge LoRA before syncing so vLLM gets full weights
+        if is_peft_model(self.model):
+            self.model.merge_adapter()
```
AmineDiro, line 764: I think we need to check if `merge_adapter` model.merge_adapter() works in distributed settings. If the base weights  are still DTensor shards how does the merge operate 🤔  ??

AmineDiro (commented):
Thanks a lot @rycerzes  for PR 🤗 ! LoRA support is definitely needed for AsyncGRPO. 

Looking at the PEFT integration in the GRPO module. I think the PR is missing support for  `modules_to_save` for example:

```
  LoraConfig(target_modules=[...], modules_to_save=["lm_head", "embed_tokens"])
```
Peft renames the target modules, so if we send them as is to vLLM it will probably error.  So in GRPO we do something like [this](https://github.com/huggingface/trl/blob/5f0e83d2bbbad79d6c231164ee09c3492cd27707/trl/generation/vllm_generation.py#L424).

rycerzes:
Thanks for the review @AmineDiro

> **On `modules_to_save`:**

I'll add the same name cleanup that GRPOTrainer does in `_sync_fsdp2_params_to_vllm`, skipping `original_module` entries and stripping `modules_to_save.default.` in both weight metadata collection and `_streaming_iter`.

> **On `merge_adapter()` in distributed settings:**

I will be running some integration tests for this to verify the behavior and then re-request for a review :)

rycerzes (commented):
Thanks for flagging this @AmineDiro, tested the PEFT sync path more directly and can confirm:

- **Dense LoRA + FSDP2** works: both ranks participate in `full_tensor()` and the merged weights sync to vLLM correctly.
- **QLoRA + FSDP2** does not work with the current merge-before-sync path. PEFT/bnb `merge_adapter()` expects bitsandbytes params with quantization metadata (`quant_state`, etc.), but FSDP2 wraps/shards those as `DTensor`, so the merge fails.

I've added an explicit guard for the `AsyncGRPO + PEFT + QLoRA + FSDP2` combination in the current full-weight vLLM sync path, with a clear error message pointing users toward the supported alternatives (dense LoRA + FSDP2, or QLoRA without FSDP2).

Tracked in #5975.

Longer term, the right fix would be  an **adapter-only vLLM LoRA sync** path, load the frozen base model in vLLM with LoRA support enabled and sync only the adapter tensors, rather than calling `merge_adapter()` on quantized FSDP2 base weights. This is what Axolotl does with `trl.vllm_lora_sync: true` and what verl supports with `lora.merge=False`. Avoids the DTensor/bnb metadata problem entirely and keeps the sync payload small.

AmineDiro:
Closing this in favour of #7017, which lands PEFT/LoRA support for `AsyncGRPOTrainer` together with the adapter-only sync path — thank you for the work here, and for the review round on `modules_to_save` and `merge_adapter()` in distributed settings, which is a good part of why the final version handles both.

Where #7017 differs: the sync mode is derived from a PEFT model plus a server reporting `lora_config` rather than being a config field, and the decision is made on rank 0 and broadcast The adapter save runs on every rank because materializing a sharded parameter is a collective. Each sync publishes a new versioned adapter name rather than reloading in place, since vLLM keys its prefix cache on the adapter name alone (vllm#42125) and would otherwise serve the previous policy's KV blocks.

PR #5906, opened 2026-06-02
Replace custom numpy cache in precompute_ref_logps with native datasets

albertvillanova:
Replace custom numpy cache in precompute_ref_logps with native datasets.

This PR updates the caching mechanism for reference log probabilities in the `DPOTrainer` class to store and load cached data using the HuggingFace `datasets` library, rather than using NumPy `.npz` files. This improves compatibility and simplifies cache management.

### Motivation

`DPOTrainer._precompute_ref_logps` had a hand-rolled cache mechanism: a custom fingerprint → `.npz` path derived by stripping the `.arrow` suffix from `_get_cache_file_path` and appending `.npz`, then saving and loading the two logp arrays with `numpy`.

This reimplemented what `datasets` already does natively. This PR replaces it with the standard `Dataset.map` caching.

### Changes

**Caching and Loading Improvements:**

* Changed the cache file handling in `_precompute_ref_logps` to use HuggingFace dataset serialization instead of NumPy `.npz` files, allowing for seamless integration with the rest of the dataset pipeline.
* When loading cached reference log probabilities, the code now uses `concatenate_datasets` to merge the original and cached datasets along columns, rather than loading NumPy arrays. 
* When saving reference log probabilities, the code now uses the `map` function to add columns directly to the dataset and writes the result to disk, instead of saving NumPy arrays.

**Dependency Updates:**

* Added the `concatenate_datasets` import from `datasets` and removed the unused `numpy` import, reflecting the switch away from NumPy-based caching.

---

> [!NOTE]
> **Medium Risk**
> Changes DPO training data preparation and invalidates prior .npz caches; multi-GPU cache write still relies on main-process map plus barrier before load.
> 
> **Overview**
> **`DPOTrainer._precompute_ref_logps`** no longer uses a custom NumPy `.npz` sidecar. Reference log-prob caches now use Hugging Face **`datasets`** Arrow files at the fingerprint path from `_get_cache_file_path`, written via **`Dataset.map`** and read with **`Dataset.from_file`**.
> 
> On a cache hit or after writing, the method returns **`concatenate_datasets([dataset, cached], axis=1)`** so `ref_chosen_logps` / `ref_rejected_logps` are merged as columns instead of **`add_column`** from in-memory NumPy arrays. The **`numpy`** import is removed; **`concatenate_datasets`** is added.
> 
> Existing **`.npz`** caches from the old path suffix hack will not be picked up automatically; the first run after upgrade may recompute reference log probs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd0208544700c992414b33e639ca82b768edef19. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -29,7 +29,7 @@ import transformers
 from accelerate import PartialState
 from accelerate.logging import get_logger
 from accelerate.utils import is_peft_model, tqdm
-from datasets import Dataset, IterableDataset, IterableDatasetDict
+from datasets import Dataset, IterableDataset, IterableDatasetDict, concatenate_datasets
 from datasets.fingerprint import Hasher
 from packaging.version import Version
 from torch.utils.data import DataLoader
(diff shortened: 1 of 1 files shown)


qgallouedec:
I'm almost sure that I tried a similar approach, but ended up choosing numpy for some reason. Let me dig in the history of #3906 to understand, before approving.

albertvillanova:
Sure! That is the professional approach! :hugs: 

For more context, I also benchmarked the performance:
- the new approach is negligibly slower to write and measurably faster to read, at comparable file sizes
  - Save: new approach is slower in absolute terms, but the absolute numbers are tiny: 57 ms for 50k samples. This is a one-time cost, and it's dwarfed by the ref model forward pass.
  - Load: new approach (`from_file` + `concatenate_datasets`) is consistently ~2× faster than `np.load` + two `add_column` calls

qgallouedec:
Ok, I can't find why. Maybe I didn't try after all.

I probably would have written a comment. Because in situations like this, when I suspect that my future self will wonder why my past self did things this way and not another, I leave a comment. 😅

I can also see that we don't need the manual conversion ~to~ from bf16 anymore, nice! (I wrote a comment for this one at least)

qgallouedec (approved):
To be applied in KTO as well

PR #5911, opened 2026-06-02
Fix AsyncGRPO checkpoint resume triggering unnecessary vLLM inference

### LINKED ISSUE #5863: [AsyncGRPO] Loading from checkpoint recomputes all vLLM past requests
### Reproduction

Loading from a checkpoint, with default value of `ignore_data_skip=False` successfully skips the previous batches and does not apply any backward but it **does perform vLLM inference for all of the already-seen batches**.

    trainer = AsyncGRPOTrainer(
        model=args.model_path,
        ...
        train_dataset=dataset,
    )
    trainer.train(resume_from_checkpoint=...)

Can add more details later, I think this should be straight-forward for me to reproduce

Thanks in advance,

### System Info

- Python version: 3.12.13
- TRL version: 1.4.0
- Transformers version: 5.9.0
- Accelerate version: 1.13.0
- Datasets version: 4.8.5
- vLLM version: 0.19.0

### Checklist

AmmarHassona:
# What does this PR do?

When resuming from a checkpoint, the base Trainer's skip-and-replay loop pulls and discards batches from the DataLoader to fast-forward to the saved step. For `AsyncGRPOTrainer`, each discarded batch requires the rollout worker to call vLLM which wastes inference before training resumes.

Fixes #5863

**Changes:**

1. `AsyncGRPOTrainer.__init__` forces `ignore_data_skip=True` (with a warning if the user had it set otherwise) so the base Trainer's skip-and-replay loop never runs against the live rollout queue. `ignore_data_skip` is also added to `AsyncGRPOConfig` with `default=True` so the warning only fires when a user explicitly sets it to `False`.
2. A new `_SaveRolloutStateCallback` writes the rollout worker's current prompt position to `rollout_state.json` alongside each checkpoint. The position is computed as the low-water-mark of `_trained_groups` (the first untrained group id) plus `dataset_start_index`, so it reflects the training cursor rather than the generation cursor.
3. On resume, `_inner_training_loop` reads `rollout_state.json` and injects the saved index into the worker's spawn kwargs as `dataset_start_index` before `_StartRolloutWorkerCallback` fires. This is skipped for `IterableDataset` since streaming datasets have no `len()`. If `rollout_state.json` is absent, a warning is logged and the worker restarts from prompt 0.
4. `_AsyncRolloutLoop.__init__` accepts `dataset_start_index` and uses an O(1) Arrow slice (`dataset.select(range(start, len(dataset)))`) to fast-forward the dataset iterator before rollouts begin.

**Tests:**
Added `test_resume_from_checkpoint` and `TestRolloutStateCheckpoint` (3 unit tests, no GPU needed) covering the callback write, iterator skip, and `_inner_training_loop` assignment.

## Before submitting

- [x] Was this discussed/approved via a GitHub issue? https://github.com/huggingface/trl/issues/5863

## AI writing disclosure

## Who can review?

@AmineDiro @albertvillanova

---

> [!NOTE]
> **Medium Risk**
> Changes checkpoint/resume and epoch-stop accounting for long async GRPO runs; behavior is localized and covered by new tests, but incorrect cursor logic could skip or repeat prompts.
> 
> **Overview**
> Fixes checkpoint resume for **AsyncGRPO** so resuming no longer runs the base Trainer’s skip-and-replay loop against the live rollout queue (each skipped batch would otherwise trigger wasteful vLLM generation).
> 
> **`ignore_data_skip`** now defaults to **`True`** in `AsyncGRPOConfig` and is forced on in the trainer. Resume position is tracked separately via **`rollout_state.json`** written in **`_save_checkpoint`** (before the parent save/Hub upload) with **`prompt_index`** = trained cursor (`dataset_start_index` + first untrained `group_id`, including gaps when stale drops leave holes). On resume, **`_inner_training_loop`** loads that file into the worker’s **`dataset_start_index`** and **`_groups_before_resume`** so epoch stopping still counts prompts trained before the checkpoint; **`IterableDataset`** / missing state fall back to prompt 0 with warnings. The rollout loop fast-forwards with an Arrow **`dataset.select`** slice.
> 
> Docs describe the checkpoint/resume model; tests cover save ordering, hole handling, iterator skip, resume wiring, and end-to-end resume smoke.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e5fe37de9a110821e8528c80ccab1fb7317def1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AmmarHassona:
Fixed the missing `next()` after iterator reset on `StopIteration`

AmineDiro (changes requested):
Thanks for tackling this 🤗 
I think general checkpointing, start/resume from AsyncGRPO needs to get tightened a bit, so cool to see a PR, but  I'd like to push back on the layer the fix operates on, though, because I think it's working around base-class behavior and not the general design.

In AsyncGRPO, the dataloader pulls from a live queue that vLLM fills on the fly so there's no fixed batch N. It's just whatever's in the queue at that moment. So the base Trainer's skip-and-replay-batches logic doesn't apply to AsyncGRPO at all. 
This PR makes the replay cheap with dummy samples, but it's still working around a behavior that shouldn't run at all because we inherit it from base class.

IMO, the correct fix is to: 
-  Set `ignore_data_skip=True`. That completely turns off the replay loop, which is the right thing for streaming data. 
- The only thing left to do is tracking the dataset/prompt position in the worker: which prompt we're on. Right now the worker restarts at prompt 0 on resume, so we retrain on the first prompts. If we care about that, the fix belongs in the worker; we should just save the **prompt index** when checkpointing and skip the dataset iterator forward on resume.

AmmarHassona:
Thanks for the feedback! I completely understand your point and it makes sense, the base class skip logic doesn't apply and `ignore_data_skip=True` is the right fix.

Should I update the PR to apply that fix? Also, should prompt-position tracking in the worker be a follow-up issue, or is that also in scope here?

AmineDiro:
@AmmarHassona  you can implement in the same PR! 
more than happy ot review your work !

AmmarHassona:
Sounds good, I'll implement them in the PR! Thanks!

AmmarHassona:
Hi @AmineDiro, just checking in when you get a chance. Happy to address any additional feedback or comments you might have!

tests/experimental/test_async_grpo_trainer.py
```diff
@@ -134,3 +135,56 @@ def test_train(self):
...
+            report_to="none",
+        )
+
+        # First run: train for 2 steps, which saves a checkpoint at step 1.
+        trainer = AsyncGRPOTrainer(
+            model=model_id,
+            reward_funcs=dummy_reward_func,
+            args=training_args,
+            train_dataset=dataset,
+            rollout_worker=_StubRolloutWorker(tokenizer, dataset, num_generations=3),
```
albertvillanova, line 226: I think the position-tracking feature is never exercised by the test:
- You are passing as `rollout_worker` an instance of `_StubRolloutWorker`
- And the resume logic in `AsyncGRPOTrainer` is guarded by `isinstance(self.rollout_worker, AsyncRolloutWorker)`

The test only asserts `ignore_data_skip is True` and that resume doesn't crash with a stub.

However, the `global_step → prompts_sent` math, the `_loop_kwargs["dataset_start_index"]` assignment, and the worker's skip loop are all untested. The `dataset_start_index` doesn't appear anywhere in the test file.

Do you think at least a unit test could be implemented?

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -770,6 +774,20 @@ def _sync_weight(self):
...
+        # We set dataset_start_index on _loop_kwargs before start() is called by the callback.
+        resume_from_checkpoint = kwargs.get("resume_from_checkpoint")
+        if resume_from_checkpoint is not None and isinstance(self.rollout_worker, AsyncRolloutWorker):
+            state_file = os.path.join(resume_from_checkpoint, "trainer_state.json")
+            if os.path.isfile(state_file):
+                global_step = TrainerState.load_from_json(state_file).global_step
+                if global_step > 0:
+                    batch_size = self.args.per_device_train_batch_size * self.accelerator.num_processes
+                    samples_seen = global_step * self.args.gradient_accumulation_steps * batch_size
+                    prompts_sent = samples_seen // self.args.num_generations
```
albertvillanova: I think this position is an approximation, not the saved cursor @AmineDiro asked for:
- @AmineDiro asked to "save the prompt index when checkpointing and skip forward on resume"
- This PR instead derives it: `prompts_sent = global_step * grad_accum * per_device_bs * num_proc // num_generations`

I think this diverges under staleness drops: `RolloutQueueDataset.__iter__` drops stale samples without counting them as trained. So by the time `global_step` is reached , the worker has produced more prompts than `samples_seen // num_generations`. The derived index lands behind the worker's true position, and the resumed run re-trains some already-seen prompts.

I would point out that this approach trades exactness for simplicity. Not a strong opinion though: either accept it with a comment noting the approximation, or persist the worker's actual group cursor via shared state at checkpoint time. 

@AmineDiro, what would you suggest?

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -770,6 +774,20 @@ def _sync_weight(self):
...
+                    if isinstance(self.train_dataset, Dataset):
+                        self.rollout_worker._loop_kwargs["dataset_start_index"] = prompts_sent % len(self.train_dataset)
```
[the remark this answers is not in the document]
albertvillanova: I think this is right: `IterableDataset` resume silently restarts at prompt 0.

I think current implementation is defensible: `len()` and exact repositioning aren't generally possible on streaming data, and the main bug is still fixed. However, there's no comment explaining the deliberate choice. 

What about adding a one-line comment on the isinstance guard so a future maintainer doesn't read it as an oversight?

trl/experimental/async_grpo/async_grpo_trainer.py
albertvillanova: Additionally, there's no enforced divisibility between the global batch and `num_generations`, so `samples_seen` need not be a multiple of `num_generations`. The floor division then drops a partial group.

I would say this is a minor issue: one prompt may be re-generated. However, combined with my comment above (https://github.com/huggingface/trl/pull/5911/changes#r3473503158), it's another small source of drift.

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -471,6 +472,9 @@ def __init__(
                 f"(max_staleness={self.args.max_staleness} × samples_per_step={samples_per_step})"
             )
 
+        # The base Trainer's data-skip replay loop doesn't apply to AsyncGRPO's live rollout queue.
+        self.args.ignore_data_skip = True
```
albertvillanova, line 931: This silently overrides a user-set value. What about a note in the `AsyncGRPOConfig` docstring (or a warning or debug log) so users aren't surprised their setting is ignored?

albertvillanova (commented):
Thanks for contribution, @AmmarHassona.

Some comments/questions below.

Additionally, I think the PR description must be updated: it documents the abandoned skip_samples mechanism and a _StubRolloutWorker.dataset/_dataset_iter/skip_samples assertion that no longer exist.

tests/experimental/test_async_grpo_trainer.py
AmmarHassona: I agree yes that a unit test should be implemented. I will hold off on the unit test until the approximation concern is resolved, then I can write a test for the logic we decide on.

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: Those are very good points. In the general case and (at the limit), I think this needs to be a user's decision. To go deeper: For environments where we don't have a dataset to pass to the trainer, we still need to have some state that could be checkpointed. 

This probably won't be correct like  @albertvillanova  mentions.  Now what I suggested is to save the exact index of the prompt in the same checkpoint as I think it's a bit clearner

tests/experimental/test_async_grpo_trainer.py
AmmarHassona: I added a `TestRolloutStateCheckpoint` with 4 unit tests covering the callback write, iterator skip, counter tracking, and the `_inner_training_loop` assignment. They do not need a GPU to run.

trl/experimental/async_grpo/async_grpo_trainer.py
AmmarHassona: The child now tracks rows consumed in a shared `mp.Value` and a new `_SaveRolloutStateCallback` writes to `rollout_state.json` on each checkpoint. On resume, the trainer reads the file directly.

trl/experimental/async_grpo/async_grpo_trainer.py
AmmarHassona: The warning was already in place in a previous commit, but re-reading your comment I thought it made sense to also add a note to the `AsyncGRPOConfig` docstring so it's visible before users even run the code.

trl/experimental/async_grpo/async_grpo_trainer.py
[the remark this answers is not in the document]
AmmarHassona: Noted - I added an inline comment on the `isinstance` guard explaining the `IterableDataset` skip is deliberate.

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -167,6 +169,19 @@ def on_train_begin(self, _args, _state, _control, **_kwargs):
...
+    """Saves the current prompt index to rollout_state.json on each checkpoint."""
+
+    def __init__(self, trainer: "AsyncGRPOTrainer"):
+        self._trainer = trainer
+
+    def on_save(self, args, state, _control, **_kwargs):
+        if self._trainer.accelerator.is_main_process and isinstance(self._trainer.rollout_worker, AsyncRolloutWorker):
+            checkpoint_dir = os.path.join(args.output_dir, f"checkpoint-{state.global_step}")
+            with open(os.path.join(checkpoint_dir, "rollout_state.json"), "w") as f:
+                json.dump({"prompt_index": self._trainer.rollout_worker.prompt_index}, f)
```
albertvillanova: This is the point I raised in https://github.com/huggingface/trl/pull/5911#discussion_r3473503158, and I think it's still open: the rewrite to `mp.Value` + `rollout_state.json` changed *how* the index is obtained, but not *which* index it is. Bugbot restated it on 07-09 ("Resume skips untrained prompts").

`_prompt_index_value` is incremented in `_repeat_iterator` the moment a row is **pulled for generation**, and that point runs well ahead of training:

- `_generate_loop` pulls eagerly to keep all `max_inflight_tasks` slots busy (`async_rollout_worker.py#L490-L491`)
- `_score_loop` then buffers up to `queue_maxsize` *samples* (`async_rollout_worker.py#L686`, default 1024)

So the saved index leads the trained position by roughly `(queue_maxsize + max_inflight_tasks) / num_generations` prompts — with the defaults (1024, `4 * samples_per_step`, `num_generations=8`) that's on the order of 160 prompts — **plus** every group that `RolloutQueueDataset.__iter__` dropped as stale (`async_grpo_trainer.py#L326`, bare `continue`).

Those buffered and in-flight samples die with the child process, so on resume the worker skips past prompts that were generated but **never trained**. The error is one-directional: always skip-untrained, never re-generate. I think that means the PR currently trades `main`'s *wasteful but complete* resume for an *efficient but lossy* one — and for a fix whose whole point is avoiding wasted inference, silently dropping training data feels like the wrong side to err on.

`_trained_groups` is the trained truth, and it's already maintained in the collator on the main process only (`async_grpo_trainer.py#L489`) — the same guard this callback already uses. `_repeat_iterator` also makes the mapping exact: group `g` is dataset row `dataset_start_index + g`. So something like:

```python
def on_save(self, args, state, _control, **_kwargs):
    if self._trainer.accelerator.is_main_process and isinstance(self._trainer.rollout_worker, AsyncRolloutWorker):
        # Persist the *trained* cursor, not the worker's generation cursor: the worker runs ahead by the rollout
        # queue depth plus in-flight tasks, and those samples die with the child process. Group `g` is dataset row
        # `dataset_start_index + g` (see `_repeat_iterator`), so resume at the first group never trained --
        # stale-dropped groups are re-generated rather than silently skipped.
        trained = self._trainer._trained_groups
        first_untrained = next(g for g in itertools.count() if g not in trained)
        prompt_index = self._trainer.rollout_worker._loop_kwargs["dataset_start_index"] + first_untrained
        ...
```

`max(trained) + 1` is the simpler variant, but it skips the holes left by stale drops; the low-water mark above costs one generator expression and never loses data. Either is a large improvement over the generation cursor — I'd just like the choice made explicitly, with a comment.

One nice side effect: with this, the `mp.Value` plumbing through `_child_main` / `_AsyncRolloutLoop` is no longer needed at all, and only `dataset_start_index` still has to reach the child. That's a decent chunk of this diff removed.

@AmineDiro does this match what you had in mind with "save the exact index of the prompt"?

tests/experimental/test_async_grpo_trainer.py
```diff
@@ -188,6 +193,59 @@ def test_train(self):
...
+            num_generations=3,
+            max_steps=2,
+            save_steps=1,
+            max_completion_length=8,
+            vllm_server_timeout=5.0,
+            report_to="none",
+        )
+
+        # First run: train for 2 steps, which saves a checkpoint at step 1.
+        trainer = AsyncGRPOTrainer(
```
albertvillanova, line 221: I ran this on an Ampere+ box and it can't pass as written — it fails at trainer construction, before reaching any resume logic:

```
tests/experimental/test_async_grpo_trainer.py:216: in test_resume_from_checkpoint
    trainer = AsyncGRPOTrainer(
trl/experimental/async_grpo/async_grpo_trainer.py:822: in __init__
    self.weight_transfer = WeightTransferClient(
E   ImportError: vLLM >= 0.22.0 is required to use WeightTransferClient
```

Two omissions relative to `test_train`:

1. no `weight_transfer=_StubWeightTransfer()`, so a real `WeightTransferClient` is built (and `_InitialWeightSyncCallback` would then try to reach a real vLLM server);
2. no `token_budget=256`, so `get_train_dataloader` calls `vllm_client.wait_for_server_ready()` — `test_train` sets it with exactly that comment ("the stub worker has no real vLLM server to query for max_model_len").

Adding both brings this test to parity with `test_train` on my machine, so those two omissions are the whole test-side defect.

CI never caught it because `TestAsyncGRPOTrainer` is `skipif(not is_ampere_or_newer() and torch_device != "xpu")` and `tests-experimental.yml` runs on `aws-g4dn-2xlarge` (T4, Turing), so the entire class is skipped. Could you run this one on an Ampere+ GPU before the next push?

Separately: even once it's fixed, this test doesn't exercise the new feature — `_StubRolloutWorker` isn't an `AsyncRolloutWorker`, so both `_SaveRolloutStateCallback.on_save` and the resume branch in `_inner_training_loop` stay inert. That's fine now that `TestRolloutStateCheckpoint` covers the pieces; I'd just narrow the docstring/comment so the test doesn't read as asserting more than `ignore_data_skip is True` plus "resume doesn't crash".

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -13,8 +13,10 @@
 # limitations under the License.
 
 
+import json
```
albertvillanova: `Check code quality` is failing on this commit (both the `ruff check` and `ruff format` hooks) — `import json` landed before `import contextvars`, which trips `I001`. There's a second occurrence in the test file: the new `from trl.experimental.async_grpo.async_rollout_worker import AsyncRolloutWorker, RolloutSample, _AsyncRolloutLoop` should be merged into the existing import block from the same module just below it.

`make precommit` fixes both.

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -309,11 +311,19 @@ def __init__(
...
         num_completions_to_print: int | None = None,
+        dataset_start_index: int = 0,
+        prompt_index_value: MPValue | None = None,
         fork_threshold_tokens: int = 1024,
     ):
         self.model_name = model_name
         self.dataset = dataset
         self._dataset_iter = iter(dataset)
+        if dataset_start_index > 0:
+            for _ in range(dataset_start_index % len(dataset)):
```
albertvillanova: Two small things here:

- This runs `next()` `dataset_start_index % len(dataset)` times inside `_AsyncRolloutLoop.__init__`, i.e. before `child_ready_event.set()`, which the parent waits on with a 300s `child_ready_timeout`. For a large dataset resumed deep into an epoch that's a lot of row decoding on the startup path. `iter(dataset.select(range(dataset_start_index % len(dataset), len(dataset))))` is an O(1) Arrow slice, and `_repeat_iterator` already does `iter(self.dataset)` on wrap-around, so epoch 2 onwards is unaffected either way.
- Inserting `dataset_start_index` / `prompt_index_value` before `fork_threshold_tokens` reorders the existing signature. They're always passed by keyword, so it's harmless, but appending would keep the diff smaller.

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -1144,6 +1169,23 @@ def _sync_weight(self):
...
+            self.rollout_worker._loop_kwargs["dataset_start_index"] = 0
+        resume_from_checkpoint = kwargs.get("resume_from_checkpoint")
+        if resume_from_checkpoint is not None and isinstance(self.rollout_worker, AsyncRolloutWorker):
+            rollout_state_file = os.path.join(resume_from_checkpoint, "rollout_state.json")
+            # IterableDataset is skipped deliberately: streaming datasets have no len() and can't be repositioned.
+            if os.path.isfile(rollout_state_file) and isinstance(self.train_dataset, Dataset):
+                with open(rollout_state_file) as f:
+                    prompt_index = json.load(f)["prompt_index"]
+                self.rollout_worker._loop_kwargs["dataset_start_index"] = prompt_index
+            elif not os.path.isfile(rollout_state_file):
```
albertvillanova: Thanks for adding the comment — that resolves my earlier point for a maintainer reading the code.

For a *user*, though, this branch shape means the `IterableDataset` case logs nothing at all: the file exists, the `isinstance` check fails, and the run silently regenerates from row 0. What about making the guard warn as well, so a streaming resume says so out loud? Something like `elif not isinstance(self.train_dataset, Dataset)` with its own one-line warning.

albertvillanova (changes requested):
Thanks for pushing this through @AmmarHassona, and sorry for the slow turnaround. The diagnosis and the overall shape are right, and I'd like this PR to be the one that lands the fix. But I think the value being persisted is the wrong cursor, and CI is red on this commit, so I'm marking it as changes requested. Details inline; two adjacent points here.

### Epoch/duration state is still not restored (pre-existing, not introduced here)

`_trained_groups` is also what `_EpochStopCallback` counts against `_epoch_stop_groups = math.ceil(num_train_epochs * len(train_dataset))`. It isn't persisted either, so a resumed run still trains a *full* `num_train_epochs` starting from the resume point. That's already true on `main`, so it's not a regression from this PR — but if the trained cursor ends up being persisted anyway (see my inline comment), the count lands in the same place for almost no extra code. Otherwise let's open a follow-up issue so it isn't forgotten.

### Smaller things

- `os.path.join(args.output_dir, f"checkpoint-{state.global_step}")` diverges from `Trainer._get_output_dir(trial)` under hyperparameter search. Unrealistic for AsyncGRPO, so I don't think it needs changing — noting it for completeness.
- The trainer reaching into `rollout_worker._loop_kwargs` from outside is a bit hidden, given the worker already exposes a public `prompt_index` property. Low priority, and I don't feel strongly about it.
- Please **don't** act on Bugbot's "malformed `rollout_state.json` crashes resume" comment. `AGENTS.md` explicitly rules out defensive code, and a `KeyError` on a corrupt checkpoint file is the right behaviour here.
- Test hygiene: the two `_MockWorker(AsyncRolloutWorker)` classes that skip `__init__` could just be `MagicMock(spec=AsyncRolloutWorker)`, which passes `isinstance`.

The 4 new unit tests in `TestRolloutStateCheckpoint` do address my earlier request for unit coverage — they pass locally in ~22s with no GPU. Thanks for adding them.

AmmarHassona:
Thanks for the thorough review @albertvillanova! Here's what was addressed in this push:

- Fixed import ordering (`contextvars` before `json`) and merged the duplicate import block in the test file
- Added an `elif not isinstance(self.train_dataset, Dataset)` warning so `IterableDataset` resumes logs explicitly that the worker restarts from prompt 0
- Replaced the `next()` skip loop with `dataset.select(range(...))` for an O(1) Arrow slice
- Added `token_budget=256` and `weight_transfer=_StubWeightTransfer()` to both trainer instances in `test_resume_from_checkpoint` bringing it to parity with `test_train`. Narrowed the docstring to reflect what the test actually checks. I unfortunately don't have access to Ampere hardware to verify locally but the fix matches exactly what you described.
- Replaced both `_MockWorker(AsyncRolloutWorker)` classes with `MagicMock(spec=AsyncRolloutWorker)`

The cursor change (switching from `mp.Value` to `_trained_groups`) and the epoch/duration state point are pending the response of @AmineDiro. I'll address those together in the next push once he weighs in his opinion.

verma8076:
Thanks for the ping @albertvillanova. Dug into the `prompt_index` vs. trained-state gap you flagged, and I think it's a real problem, not just a theoretical one.

`prompt_index` is bumped in [`_repeat_iterator`](https://github.com/AmmarHassona/trl/blob/ddcf4227f36dd4390279634fb97eca80c1c536ab/trl/experimental/async_grpo/async_rollout_worker.py#L717-L718), right after a row is pulled off the dataset — i.e. at *dispatch-for-generation* time. The actual "this group completed a training step" signal is [`self.groups_trained.update(...)`](https://github.com/AmmarHassona/trl/blob/ddcf4227f36dd4390279634fb97eca80c1c536ab/trl/experimental/async_grpo/async_grpo_trainer.py#L489), which only fires once a batch reaches the collator.

Between those two points sits the whole rollout pipeline: `rollout_buffer`'s queue, whose [`queue_maxsize` defaults to 1024](https://github.com/AmmarHassona/trl/blob/ddcf4227f36dd4390279634fb97eca80c1c536ab/trl/experimental/async_grpo/async_grpo_config.py#L290), plus anything still mid-generation. So by default, `prompt_index` can legitimately sit up to ~1000+ prompts ahead of `groups_trained`. Saving `prompt_index` and resuming with it as `dataset_start_index` would silently skip that entire gap on every resume — prompts that were already generated (and cost real vLLM time) but never actually trained on, gone for good. Given the default queue depth, this isn't an edge case, it's the buffer's normal operating range.

One more wrinkle on top of "use `groups_trained` instead": it's a `set`, and with concurrent generation there's no guarantee groups complete and land in the queue in dispatch order. So the correct resume point is the *lowest untrained group id* (the first gap), not `max(groups_trained)` — otherwise a straggler group stuck mid-pipeline at checkpoint time gets skipped forever even though everything after it did get trained.

Happy to help work through a fix if useful — the shape I'd reach for is tracking the low-water-mark of contiguous trained group ids (rather than the raw set) and checkpointing that instead of `prompt_index`, but curious if you all see a cleaner way to thread it through the queue/child-process boundary.

albertvillanova (changes requested):
Thanks for the quick turnaround @AmmarHassona, and thanks @verma8076 for the follow-up analysis; it's exactly right, and it settles the open question.

I verified my previous comments have been addressed.

#### Still blocking: the persisted value is the generation cursor

Unchanged in this push, and @verma8076's numbers in #5863 match mine: _prompt_index_value is bumped in _repeat_iterator (async_rollout_worker.py:718) at dispatch-for-generation time, while the trained truth is groups_trained.update(...) in the collator (async_grpo_trainer.py:489). Between them sit queue_maxsize samples (default 1024) plus max_inflight_tasks in flight (default max_staleness × samples_per_step = 4 × samples_per_step), so with num_generations=8 and, say, samples_per_step=64, the saved index leads the trained position by ~160 prompts — plus every group RolloutQueueDataset.__iter__ dropped as stale. Those samples die with the child, so resume skips prompts that were generated but never trained. The error only ever goes one way: skip-untrained, never re-generate.

#### New: the ignore_data_skip warning fires on every run

TrainingArguments.ignore_data_skip defaults to False, so if not self.args.ignore_data_skip (async_grpo_trainer.py:792) is true for every AsyncGRPO construction, not just when a user set it. I hit it in my own test run:
> WARNING trl.experimental.async_grpo.async_grpo_trainer:async_grpo_trainer.py:793 [RANK 0] `ignore_data_skip` is forced to `True` for AsyncGRPO because ...

from a test that never touches ignore_data_skip. That's my earlier request coming back to bite; sorry. The clean fix is the pattern the config already uses: add it under # Parameters whose default values are overridden from TrainingArguments (async_grpo_config.py:170), next to learning_rate / lr_scheduler_type.

#### Minor

  - The if / elif not isfile / elif not isinstance chain in _inner_training_loop re-tests both conditions to pick a message. Reads easier as the negative cases first, then the load in the else. Same line count, less backtracking.
  - The double isinstance(self.rollout_worker, AsyncRolloutWorker) in adjacent statements (:1195, :1198) can nest under one guard.
  - docs/source/async_grpo_trainer.md has no resume/checkpoint section, and the PR's own docs checkbox is unchecked. A short paragraph under "How it differs from [GRPOTrainer]" — no replay loop, rollout position restored from rollout_state.json, streaming datasets restart at prompt 0 — would be the right home. Low priority relative to the above.
  - Still standing from last round: please don't act on Bugbot's "malformed rollout_state.json crashes resume" comment. AGENTS.md rules out defensive code; a KeyError on a corrupt checkpoint file is the behaviour we want.

AmmarHassona:
Thanks again @albertvillanova, and thanks @verma8076 for the follow-up analysis that settled the cursor question.

Here's what was addressed in this push:

  - Switched `_SaveRolloutStateCallback.on_save` from `mp.Value` (generation cursor) to `_trained_groups` + `itertools.count()` low-water-mark — resumes now land at the first untrained prompt, correctly handling stale-dropped groups with gaps
  - Removed all `mp.Value` plumbing: `_prompt_index_value` from `AsyncRolloutWorker`, `prompt_index` property, `prompt_index_value` param from `_AsyncRolloutLoop.__init__` and `_child_main`, and the counter increment in `_repeat_iterator`
  - Added `ignore_data_skip` to `AsyncGRPOConfig` with `default=True` so the warning only fires when a user explicitly sets it to `False`
  - Restructured the `if/elif` chain in `_inner_training_loop` to put negative cases first and load in the `else`, and nested the two `isinstance` guards under one check
  - Added a checkpoint/resume paragraph to `docs/source/async_grpo_trainer.md`
  
  Also updated the PR description to reflect the new cursor approach and corrected the docs checkbox.

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -1167,7 +1183,45 @@ def _sync_weight(self):
...
+            trained = self._trained_groups
+            first_untrained = next(g for g in itertools.count() if g not in trained)
+            rollout_state = {
+                "prompt_index": self.rollout_worker._loop_kwargs["dataset_start_index"] + first_untrained,
+                # This counts prompts trained, which is what an epoch is measured in
+                "groups_trained": self._groups_before_resume + len(trained),
+            }
+            with open(os.path.join(checkpoint_dir, "rollout_state.json"), "w") as f:
+                json.dump(rollout_state, f)
+        super()._save_checkpoint(model, trial)
```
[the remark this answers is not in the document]
AmineDiro: what I dont see any  tmp checkpoint staging in `Trainer._save_checkpoint` ?!?!

AmineDiro (approved):

AmineDiro:
[435ee4a](https://github.com/huggingface/trl/pull/5911/commits/435ee4a952ff4a8373289e5b47eb8b19db007dc4) 

Added a few things and fixed missing stuff, ran an experiment on 800 steps, on [hf jobs](https://huggingface.co/docs/huggingface_hub/en/guides/jobs). Checkpoint every 200 steps.  Looks good .
image


Repro: 
- [async_grpo_sanity.py](https://gist.github.com/AmineDiro/2636ec39768586d1f8a8351d7cb6a524#file-async_grpo_sanity-py)
- [hfjob_async_grpo_smoke.sh](https://gist.github.com/AmineDiro/1b13a7f28bc459a3d6b176b9949a53b1#file-hfjob_async_grpo_smoke-sh)

AmmarHassona:
Thanks for validating @AmineDiro!

albertvillanova (approved):
Thanks @AmmarHassona, and thanks @AmineDiro for taking this over the line. All three of my blocking points are addressed, CI is green, and I'm happy for this to land.

Verified addressed:
- The cursor is now the trained one
- The mp.Value plumbing is gone
- ignore_data_skip is now a real config default
- Epoch state is restored
- Docs paragraph added
- TestRolloutStateCheckpoint: 5 passed locally (~109s, no GPU). test_save_checkpoint_skips_holes_left_by_stale_drops is exactly the hole-aware test I asked for.

One thing I want to call out as better than what I proposed: moving from an on_save callback to a _save_checkpoint override. I'd suggested on_save; @AmineDiro is right that it's wrong.

AmmarHassona:
Thanks @albertvillanova!

ggcr:
Thanks for the effort, folks

PR #5914, opened 2026-06-02
Align KTO with DPO: Replace completion_labels/get_batch_logps with completion_mask

albertvillanova:
Align KTO with DPO: Replace completion_labels/get_batch_logps with completion_mask.

Part of:
- #4786

This PR refactors how prompt and completion tokens are masked and processed in the KTO trainer and data collator. The main change is replacing the use of label masking (`-100` for prompt tokens) with an explicit binary `completion_mask` to distinguish prompt and completion tokens. This leads to more consistent handling of masked tokens throughout the codebase and simplifies log probability computations.

### Solution

This PR replaces KTO's `-100`-based logprob masking approach with the `completion_mask` binary tensor pattern used by DPO, eliminating the `get_batch_logps` static method.

- **Before:** The collator emitted `completion_labels` (prompt tokens masked with `-100`, completion tokens as actual IDs). Logprobs were computed via a dedicated static method `get_batch_logps` that derived the mask by testing `labels != -100`.

- **After:** The collator emits `completion_mask` (0 for prompt/padding, 1 for completion tokens), matching DPO's `DataCollatorForPreference`. Logprob computation is inlined at each call site using `selective_log_softmax` + `per_token_logps[mask == 0] = 0.0`, exactly as DPO does.

### Changes

**Data Collator and Batch Preparation:**
- The data collator now generates a binary `"{prefix}_mask"` (e.g., `completion_mask`) that marks completion tokens with `1` and prompt/padding tokens with `0`, instead of using label masking with `-100`. 
- The test for tokenization and processing is updated to check the correctness of the new `completion_mask` instead of label masking.

**Log Probability and Loss Computation:**
- All log probability computations now use the new `completion_mask` to select completion tokens, setting non-completion token log-probs to zero before summing. The old `get_batch_logps` method (which relied on `-100` label masking) is removed.
- The loss computation for Liger uses the shifted `completion_mask` to set the target tensor, masking out prompt and padding tokens with `-100`.

---

> [!NOTE]
> **Medium Risk**
> Changes how sequence log-probabilities are aggregated for KTO training and reference precomputation; behavior should match prior masking intent but any mask/shift mismatch would affect rewards and loss.
> 
> **Overview**
> KTO’s unpaired preference collator and trainer now follow the same **completion masking** pattern as DPO: batches expose binary `completion_mask` / `KL_completion_mask` (0 for prompt and padding, 1 for completion tokens) instead of `completion_labels` with `-100` on the prompt.
> 
> **Collator:** `DataCollatorForUnpairedPreference` builds and pads `{prefix}_mask` tensors and documents the new batch keys; `completion_labels` padding is removed.
> 
> **Log-probs:** Reference, policy, and KL log-probability paths no longer call the removed static `get_batch_logps`. They use causal shifting, `selective_log_softmax`, zero out non-completion positions via the shifted mask, then sum per sequence—matching DPO’s inline style.
> 
> **Liger path:** Targets for the fused KTO loss are built from shifted `completion_input_ids` with prompt/padding positions set to `-100` using the shifted `completion_mask`, instead of slicing `completion_labels`.
> 
> **Tests:** `test_tokenize_and_process_tokens` asserts mask structure (zeros before first one, both 0 and 1 present) rather than label alignment with `input_ids`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 845858d75b92fcf771f6bf0470903b36bb5819b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/experimental/test_kto_trainer.py
+++ b/tests/experimental/test_kto_trainer.py
@@ -138,17 +138,13 @@ class TestKTOTrainer(TrlTestCase):
         batch = trainer.data_collator([example])
         # completion_input_ids ends with EOS
         assert batch["completion_input_ids"][0, -1].item() == self.tokenizer.eos_token_id
-        # completion_labels: prompt prefix masked with -100, answer+EOS unmasked and matching input_ids
-        completion_input_ids = batch["completion_input_ids"][0].tolist()
-        completion_labels = batch["completion_labels"][0].tolist()
-        first_unmasked = next(i for i, lbl in enumerate(completion_labels) if lbl != -100)
-        assert first_unmasked > 0  # at least the prompt is masked
-        assert completion_labels[first_unmasked:] == completion_input_ids[first_unmasked:]
-        # completion_mask: binary mask consistent with completion_labels (-100 ↔ 0, non-(-100) ↔ 1)
+        # completion_mask: prompt tokens are 0, completion tokens are 1; at least the prompt is masked
         assert "completion_mask" in batch
         completion_mask = batch["completion_mask"][0].tolist()
-        expected_mask = [0 if lbl == -100 else 1 for lbl in completion_labels]
-        assert completion_mask == expected_mask
+        assert 0 in completion_mask and 1 in completion_mask
+        first_completion = next(i for i, m in enumerate(completion_mask) if m == 1)
+        assert first_completion > 0  # at least the prompt is masked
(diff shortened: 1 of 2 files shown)


qgallouedec:
I think it looks good

@codex review

qgallouedec (approved):

PR #5897, opened 2026-06-01
SFTTrainer: merge entropy and accuracy computation to eliminate redundant logits copy

flutist:
**PR Title**

`SFTTrainer: merge entropy and accuracy computation to eliminate redundant logits copy`

**PR Description**

## What does this PR do?

Merges the entropy and token accuracy computations in `SFTTrainer.compute_loss` into a single `torch.no_grad()` block, eliminating a redundant `.contiguous()` copy of the full logits tensor.

When `loss_type="nll"` (the default) and Liger kernel is not used, `compute_loss` previously materialized `shift_logits = outputs.logits[..., :-1, :].contiguous()` **twice** — once for entropy calculation and once for accuracy calculation. The logits tensor shape is `[batch_size, seq_len, vocab_size]` — for a model with `vocab_size` ~152k (e.g. Qwen2.5), this is one of the largest allocations in `compute_loss`. The second copy is entirely redundant since both metrics operate on the same shifted logits.

### Changes:

- Moved `predictions = shift_logits.argmax(dim=-1)` and accuracy computation into the existing entropy `torch.no_grad()` block, reusing the same `shift_logits` tensor
- Removed the separate accuracy `else` branch that performed a second `.contiguous()` call
- Net reduction of ~20 lines of code

## Memory benchmark

Benchmark script (not included in this PR):

```python
import gc
import torch
from datasets import load_dataset
from trl import SFTConfig, SFTTrainer

if not torch.cuda.is_available():
    print("ERROR: No CUDA GPU available.")
    exit(1)

gc.collect()
torch.cuda.empty_cache()
torch.cuda.reset_peak_memory_stats()

dataset = load_dataset("trl-lib/Capybara", split="train[:200]")

args = SFTConfig(
    output_dir="/tmp/sft_peak_bench",
    per_device_train_batch_size=4,
    max_steps=3,
    max_length=1024,
    gradient_checkpointing=True,
    bf16=True,
    report_to="none",
    logging_steps=1,
    loss_type="nll",
)

trainer = SFTTrainer(model="Qwen/Qwen2.5-0.5B-Instruct", args=args, train_dataset=dataset)
trainer.train()

peak_mb = torch.cuda.max_memory_allocated() / 1024**2
print(f"\nPeak allocated: {peak_mb:.1f} MB")
```

Results (Qwen2.5-0.5B-Instruct, batch\_size=4, max\_length=1024, bf16, gradient\_checkpointing=True):

| | Peak allocated |
|---|---|
[...]

PR #5926, opened 2026-06-03
Remove redundant .contiguous() calls in DPOTrainer to reduce peak memory

flutist:
## PR Title

`Remove redundant .contiguous() calls in DPOTrainer to reduce peak memory`

## PR Body

**What does this PR do?**

Removes unnecessary `.contiguous()` calls on logits, labels, and completion masks in `DPOTrainer.compute_ref_log_probs` and `DPOTrainer._compute_loss`.

These slicing operations (`[..., :-1, :]`, `[..., 1:]`) produce non-contiguous views. Calling `.contiguous()` forces a full copy — for the logits tensor `[B, S, V]` with Qwen's 151k vocab, each copy is ~580 MB in bf16. In `_compute_loss` without `precompute_ref_log_probs`, both model and ref logits are copied, wasting ~1.16 GB.

The downstream consumer `selective_log_softmax` only uses `torch.gather` and `torch.logsumexp`, neither of which requires contiguous input. The same optimization was already applied to `SFTTrainer` in #5897.

The Liger loss path (`_compute_loss_liger`) is left unchanged since fused Liger kernels may rely on contiguous memory.

**Benchmark** (Qwen2.5-0.5B-Instruct, A100 44GB, bf16, batch\_size=4, max\_length=512, `precompute_ref_log_probs=False`):

| | Before | After | Δ |
|---|---|---|---|
| Peak memory | 18289 MB | 15531 MB | **−2758 MB (−15.1%)** |
| Train time (3 steps) | 2.00 s | 2.00 s | — |
| Loss / metrics | identical | identical | — |

<details>
<summary>Reproduce</summary>

```python
# Save as bench.py, then run:
#   python bench.py                                          # after
#   git stash && python bench.py && git stash pop            # before
import gc, tempfile, time, torch

gc.collect(); torch.cuda.empty_cache(); torch.cuda.reset_peak_memory_stats()
dataset = load_dataset("trl-lib/ultrafeedback_binarized", split="train[:100]")
args = DPOConfig(
    output_dir=tempfile.mkdtemp(), save_strategy="no",
    per_device_train_batch_size=4, max_steps=3, max_length=512,
    gradient_checkpointing=True, bf16=True, report_to="none",
    logging_steps=1, precompute_ref_log_probs=False,
)
start = time.perf_counter()
trainer = DPOTrainer(model="Qwen/Qwen2.5-0.5B-Instruct", args=args, train_dataset=dataset)
trainer.train()
peak_mb = torch.cuda.max_memory_allocated() / 1024**2
print(f"Peak: {peak_mb:.0f} MB | Time: {time.perf_counter()-start:.1f}s")
```

</details>

**Before submitting**

- [x] This PR fixes a bug or adds a non-invasive improvement (no new features or API changes)
- [x] A benchmark is included demonstrating the memory reduction

---

> [!NOTE]
> **Low Risk**
> Narrow memory optimization on tensor views with unchanged numerics; Liger path untouched and behavior matches the prior SFTTrainer change.
> 
> **Overview**
> **DPOTrainer** no longer calls `.contiguous()` on shifted `logits`, `input_ids`, and `completion_mask` in `compute_ref_log_probs` and `_compute_loss`. Slicing already yields views; forcing contiguity duplicated large `[B, S, V]` tensors and inflated peak VRAM during reference and policy log-prob computation.
> 
> Downstream `selective_log_softmax` only uses `torch.gather` and `torch.logsumexp`, so contiguity was unnecessary—the same change was already made for **SFTTrainer** in #5897. The **Liger** loss path (`_compute_loss_liger`) is unchanged because fused kernels may still expect contiguous buffers.
> 
> Reported impact: materially lower peak memory with identical loss/metrics and no training-time regression in the included benchmark.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 528bc898e32466b9164a9f6d9a0f58b9057e782b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -1074,9 +1074,9 @@ class DPOTrainer(_BaseTrainer):
 
         input_ids = inputs["input_ids"]
         completion_mask = inputs["completion_mask"]
-        shift_labels = input_ids[..., 1:].contiguous()
-        shift_completion_mask = completion_mask[..., 1:].contiguous()
-        ref_shift_logits = ref_outputs.logits[..., :-1, :].contiguous()
+        shift_labels = input_ids[..., 1:]
+        shift_completion_mask = completion_mask[..., 1:]
+        ref_shift_logits = ref_outputs.logits[..., :-1, :]
         ref_per_token_logps = selective_log_softmax(ref_shift_logits, shift_labels)
         ref_per_token_logps[shift_completion_mask == 0] = 0.0
 
@@ -1182,9 +1182,9 @@ class DPOTrainer(_BaseTrainer):
 
         input_ids = inputs["input_ids"]
         completion_mask = inputs["completion_mask"]
-        shift_logits = outputs.logits[..., :-1, :].contiguous()
-        shift_labels = input_ids[..., 1:].contiguous()
-        shift_completion_mask = completion_mask[..., 1:].contiguous()
+        shift_logits = outputs.logits[..., :-1, :]
+        shift_labels = input_ids[..., 1:]
+        shift_completion_mask = completion_mask[..., 1:]
         per_token_logps = selective_log_softmax(shift_logits, shift_labels)
         per_token_logps[shift_completion_mask == 0] = 0.0  # mask out non-completion tokens
         if self.ld_alpha is None:
@@ -1219,7 +1219,7 @@ class DPOTrainer(_BaseTrainer):
                 else:
                     ref_outputs = self.ref_model(**model_kwargs)
 
-            ref_shift_logits = ref_outputs.logits[..., :-1, :].contiguous()
+            ref_shift_logits = ref_outputs.logits[..., :-1, :]
             ref_per_token_logps = selective_log_softmax(ref_shift_logits, shift_labels)
             ref_per_token_logps[shift_completion_mask == 0] = 0.0  # mask out non-completion tokens
             if self.ld_alpha is None:


flutist:
@qgallouedec Hi! Could you take a look at this PR when you have a chance?

qgallouedec:
cool! super useful! can you make sure that this same inconsistency is also removed from the other trainers (some actually use the same pattern)

qgallouedec:
I did a test myself:

```python
import torch

from trl import DPOTrainer

dataset = load_dataset("trl-lib/ultrafeedback_binarized", split="train[:200]")

trainer = DPOTrainer(model="Qwen/Qwen3-1.7B", train_dataset=dataset)
trainer.train()
print(f"Peak GPU memory: {torch.cuda.max_memory_allocated() / 1024**3:.2f} GiB")
```

| | Peak GPU memory |
| --- | --- |
| Before | 69.29 GiB |
| After | 67.40 GiB |

qgallouedec (approved):

flutist:
> cool! super useful! can you make sure that this same inconsistency is also removed from the other trainers (some actually use the same pattern)

@qgallouedec Thanks for the review! I checked the other trainers:

SFTTrainer: SFT's .contiguous() is in the metrics computation block (inside torch.no_grad()), which runs after the forward + backward pass. At that point, peak memory has already been determined by the activations during forward/backward — even if there's a logits copy during the metrics phase, it won't exceed the previous peak.

qgallouedec:
ok no worry, we'll investigate case by case in subsequent PRs, thanks!

albertvillanova (commented):
Thank you for this contribution.

Before we merge, I would like to ask for two things.

1. Extend to the other trainers with the identical pattern

This repository has a strong convention that when duplicated logic is fixed in one trainer, the same fix must be applied to all others that carry the same code. The pattern you removed  (.contiguous() on a shifted slice immediately before selective_log_softmax) appears in several more places.

We would prefer to handle all of these in a single PR rather than a series of follow-ups; the experimental trainers can be included in this PR or in an immediate follow-up.

2. Minor: adjust the PR title

"with 15% save" is accurate for the specific benchmark configuration (Qwen2.5-0.5B, batch size 4, precompute_ref_log_probs=False), but the real saving is proportional to vocab size and sequence length.

Thank you again for the clear write-up. Once the scope is expanded to the other trainers, this will be ready to merge.

qgallouedec:
Oh sorry I merged before your review @albertvillanova

flutist:
> Oh sorry I merged before your review @albertvillanova

dont worry, i will fix it, finish the remaining work

PR #5930, opened 2026-06-03
Remove unnecessary explicit `.contiguous()` before `entropy_from_logits`

qgallouedec:
`entropy_from_logits` never assumes a contiguous input. Its first operation is `logits.reshape(-1, num_classes)`  and reshape (unlike `view`) accepts any stride layout, returning a view when it can and silently making a contiguous copy when it can't. Everything after that (`split`, `F.log_softmax` etc) operates on those chunks and likewise handles arbitrary strides.

Note that we don't expect the peak VRAM to be reduced: The input is now `outputs.logits[..., :-1, :]`, which gets sliced on the sequence dim, so it is non-contiguous, so `reshape` copies anyway. The copy doesn't disappear; it just moves from the explicit `.contiguous()` into reshape

---

> [!NOTE]
> **Low Risk**
> Metrics-only path under `torch.no_grad()`; downstream ops typically accept non-contiguous views, with only a small chance of kernel-specific issues.
> 
> **Overview**
> Stops forcing **contiguous** copies of `shift_logits` and `shift_labels` in `SFTTrainer.compute_loss` when logging entropy and token accuracy from full logits (non–`chunked_nll`, non–Liger path). Slicing and CP/SP `shift_labels` paths now use the tensors as returned instead of calling `.contiguous()` first, which avoids extra allocations and lowers peak VRAM during metric computation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d0e79965d46be6006e8898040968fd6c433f1b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova (commented):
I would say this is a duplicate of #5926. 😅 

Could we ask the original contributor to implement the solution in the rest of the trainers instead?

qgallouedec:
> Could we ask the original contributor to implement the solution in the rest of the trainers instead?

It seems that it's not as simple as just removing all the contiguous, I'll split into multiple PRs

albertvillanova (approved):
Thanks.

PR #5936, opened 2026-06-04
Align KTO with DPO: Inline _compute_logps into _compute_loss

albertvillanova:
Align KTO with DPO: Inline _compute_logps into _compute_loss.

This PR removes the `_compute_logps` helper and inlines its logic directly into  `_compute_loss`, making KTO's loss computation self-contained like DPO's. This change improves code clarity and efficiency, especially in the handling of chosen/rejected log probabilities and logits for both the policy and reference models.

Part of:
- #4786

### Solution

**Before:** `_compute_loss` delegated to `_compute_logps`, which bundled the KL logps call, the completion forward pass, the per-token logp computation, and the chosen/rejected split into a single method called in two places (policy pass and on-the-fly ref pass).

**After:** All of that logic lives directly in `_compute_loss`. The flow is now readable top-to-bottom without jumping to a helper method:
  1. KL logps (`_compute_kl_logps`, kept separate: also used by `_compute_loss_liger`)
  2. Policy forward pass + logp computation + chosen/rejected split
  3. Reference logps (precomputed or on-the-fly)
  4. `kto_loss` call + metrics

The `precompute_ref_logps` branch is also simplified: it now reuses `chosen_idx`/`rejected_idx` computed during the policy pass instead of recomputing them from the batch.

### Changes

**Refactoring and code simplification:**

* Removed the `_compute_logps` method from `kto_trainer.py`, consolidating its logic into the `_compute_loss` method for more direct and clear computation of log probabilities and logits.
* Inlined the computation of `policy_chosen_logps`, `policy_rejected_logps`, `policy_chosen_logits`, and `policy_rejected_logits` directly in `_compute_loss`, reducing indirection and improving readability.

**Reference model computation improvements:**

* Refactored reference model log probability computation to mirror the policy logic, explicitly handling both PEFT and non-PEFT cases and ensuring correct adapter usage and device placement.
* Improved handling of tensor indexing and CUDA operations by consistently using `index_select` with boolean masks for chosen/rejected splits. 

**General improvements:**

* Added clearer error handling for mismatches between batch size and label count, and ensured auxiliary loss logic remains intact after refactoring.

---

> [!NOTE]
> **Medium Risk**
> Touches core KTO training math paths (policy/ref forwards and chosen/rejected indexing); behavior should be equivalent but regressions would affect alignment training.
> 
> **Overview**
> **Refactors KTO loss computation** to match DPO’s pattern by removing `_compute_logps` and folding policy/reference log-probability work into `_compute_loss`.
> 
> `_compute_loss` now runs top-to-bottom: `_compute_kl_logps` (unchanged helper, still shared with `_compute_loss_liger`), policy forward + per-token logps + chosen/rejected split via `index_select`, then reference logps (cached `ref_logps` or an on-the-fly forward under `no_grad` with explicit PEFT `ref` adapter handling). The precomputed-reference path **reuses** `chosen_idx` / `rejected_idx` from the policy pass instead of rebuilding boolean masks on `ref_logps.device`.
> 
> On-the-fly reference scoring no longer calls a bundled helper; it mirrors the policy path with a separate completion forward and the same masking/sum logic before splitting chosen vs rejected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ad2dbd1907de448e3d7c44d8cbd20867519a6f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (approved):
yes!

PR #5937, opened 2026-06-04
Delta weight sync for AsyncGRPO (sparse patches over  HF Bucket)

AmineDiro:
## What does this PR do?

Adds **delta (sparse) weight synchronization** to the experimental `AsyncGRPO`. Instead of broadcasting the full policy to vLLM on every weight sync, the trainer detects which bf16 weights actually changed after the optimizer step, encodes **only those** as a sparse safetensors patch, pushes it to an **HF Storage Bucket**. On the inference engine side, vLLM applies it **in place**  no full-model broadcast, we use the _NEW_ [`SparseWeightPatch` from vllm](https://github.com/vllm-project/vllm/pull/40096) (still waiting for a release)

At steady state, a delta is ~1–10% of the model, and sparsity rises as the LR decays.

## How it works

**Trainer side**
- `LowByteChangeDetector` hooks into the optimizer and snapshots only the **low byte** of each weight's bf16 pattern (1 B/elem, half a full clone). A flipped low byte ⊆ a changed bf16 value, so precision is 1.0 by construction; recall (measured) is 1.0 in normal training. Misses cause inference drift, bounded by periodic full **anchors**.
- `delta_codec.py` does the sparse extraction on the **GPU**: `extract_sparse_batched` runs a single`nonzero` (to device sync) split across all changed params (instead of one `nonzero`/param)  (~16× less than the old dense-D2H path).
- Index encodings (main difference with the previous implementation): `raw` (int32), `gap_delta` (uint16 gaps, 2×), [`nvcomp_cascaded` ](https://developer.nvidia.com/nvcomp#section-performance)(uses GPU Cascaded delta+bitpack, ~3×, optional added a dep).
- `DeltaWeightTransferEngine.upload` writes one safetensors patch (anchor = full tensors; delta =`{name}.idx` + `{name}.val`) and pushes it to the bucket. Self-describing format: names from the `.val` keys, encoding from a global header field, gap-delta width from the index dtype.

**Lifecycle**: the trainer drives vLLM's `start_weight_update` / `update_weights` /`finish_weight_update` HTTP routes; the change detector is created in `compute_loss` before the first `optimizer.step`. 
The NCCL path is untouched.

## Requirements/constraints

- vLLM with sparse weight transfer (**vllm-project/vllm#40096**) merged to `main` after v0.22.0,  not in a release yet; install from nightly. New `delta_weight_sync` extra added to `pyproject.toml`.
- Serve with **`--model-impl transformers`**  and **`VLLM_USE_V2_MODEL_RUNNER=0`** (`apply_sparse_weight_patches` exists only on the V1 runner 😢 ). Example:
  ```bash
  CUDA_VISIBLE_DEVICES=1 VLLM_SERVER_DEV_MODE=1 VLLM_USE_V2_MODEL_RUNNER=0 \
  vllm serve Qwen/Qwen3-1.7B \
        --model-impl transformers \
        --worker-extension-cls trl.experimental.async_grpo.delta_engine.DeltaWorkerExtension \
        --weight-transfer-config '{"backend":"delta"}' \
        --max-model-len 2560
  ```

>  🔴🔴  Sparse apply is **TP=1 / PP=1** (enforced by vLLM). Dense/small models today; sharded (TP>1 / EP /
  fused-MoE) is future work.

## Tests

- **Low-byte detector**: recall 1.0 / precision 1.0 vs the full bf16 diff
- **Codec/file round-trip**: bit-exact for `raw` / `gap_delta` / `nvcomp_cascaded`, and  `extract_sparse_batched`
- [**End to end AsyncGRPO** ](https://huggingface.co/buckets/aminediroHF/async-grpo-delta-demo) (Qwen3-1.7B, GSM8K): sparse deltas apply with **0 failures** (~91→99% sparse),   reward improves **0.10 → 0.50** through the delta path; receiver timing anchor download-bound,  vllm side`decode+apply` ~1 s.

## References

- vLLM sparse weight transfer: https://github.com/vllm-project/vllm/pull/40096
- PULSE (sparse delta / selective overwrite): https://huggingface.co/papers/2602.03839

## AI writing disclosure
- [x] AI-assisted: parts were suggested/iterated with an AI tool, written and reviewed by a human.

---

> [!NOTE]
> **High Risk**
> Changes how every training step pushes policy weights to inference; mis-sync or sparse/MoE misconfiguration can leave vLLM on stale weights while training continues.
> 
> **Overview**
> **AsyncGRPO** no longer assumes a full-policy NCCL broadcast on every sync. It now defaults to **`weight_sync_mode="sparse"`**, sending only bf16 weights that changed in the last optimizer step. Changes are found by **inverting the AdamW update** from resident optimizer moments (no pre-step snapshot). Periodic **anchors** and **`weight_sync_mode="full"`** remain for non-AdamW optimizers and MoE (sparse is rejected for MoE at init).
> 
> Sync is pluggable via **`weight_sync_backend`**: **`nccl`** (in-place sparse broadcast on a shared group) or **`bucket`** (safetensors patches to an HF Storage Bucket, applied on vLLM through **`HFBucketWorkerExtension`** / `hf_bucket`). New supporting code includes GPU sparse extraction and index encodings (`gap_delta`, optional `nvcomp_cascaded`), plus config validation (e.g. sparse requires `weight_sync_steps=1`).
> 
> **vLLM ≥ 0.23.0** is required; docs and examples are updated (`async_grpo.py`, `async_grpo_delta.py`, disaggregated **`async_grpo_buckets`**). The weight-transfer API is refactored to **`init` / `sync`** transports instead of the old pause/send/resume client.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8bbc35f4bc7f030053d39063979af6df89a6c925. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AmineDiro:
I've run a micro-benchmark (no GRPO loop): the policy is held on GPU 0, vLLM serves on GPU 1 (on the same node), and we time N weight syncs to compare FULL sync NCCL vs using buckets for small models. 

| model      | full size | **NCCL (full)** | delta 0.90 | delta 0.99 | delta 0.999 | delta full-anchor |
| ---------- | --------- | --------------- | ---------- | ---------- | ----------- | ----------------- |
| Qwen3-4B   | 8.04 GB   | **0.17 s**      | 12.7 s     | 3.0 s      | 1.3 s       | 29.9 s            |
| Qwen2.5-7B | 15.23 GB  | **0.15 s**      | 27.2 s     | 4.5 s      | 1.8 s       | 46.2 s            |
| gemma-2-9b | 18.48 GB  | **0.19 s**      | 27.5 s     | 5.3 s      | 2.1 s       | 53.9 s            |

AmineDiro:
Discussing with @qgallouedec  and running some benchmarks. Even with 1 byte/param as a diff, we can't keep it on the GPU's VRAM, especially for big models. So we have to offload to CPU. But to reduce D2H sync time, I just send the low bits and use buckets in [cd83506](https://github.com/huggingface/trl/pull/5937/commits/cd8350624e40619f9dd746f608eb52b5d0c8e47d) and use pinned memory to do async transfer.

Mode & Sparsity | Baseline | Optimized H | Speedup | Peak GPU Overhead | Retained GPU
-- | -- | -- | -- | -- | --
CPU (Sparsity 0.90) | 5.147s | 1.652s | 3.11× | 1.94 GB | 0.0 GB
CPU (Sparsity 0.99) | 4.668s | 1.631s | 2.86× | 1.94 GB | 0.0 GB
CPU (Sparsity 0.999) | 5.097s | 1.632s | 3.12× | 1.94 GB | 0.0 GB

AmineDiro:
Implemented in 4a4bf3b1fea3a8f6ed05c03fa1ea965201a80238  default to sparse delta weight sync (AdamW-inversion), with pluggable transports. Now the sparse delta path is the default, with a clean per-transport structure.

We now use `AdamWInversionChangeDetector` to recover which bf16 weights changed by inverting the AdamW step from the moments already resident in the optimizer state.

 We also now have two transports, one clear path each :
  - `NCCLWeightTransfer`:  Sends a dense full policy (anchor / mode="full") or a sparse update_kind="sparse_flat" patch applied in place on vLLM via apply_sparse_weight_patches (vLLM #40096).
  
  - `HFBucketWeightTransfer`: a two-phase sync; routes the same patch through an HF Bucket for cross-host setups. Registered as the `hf_bucket` vLLM backend via `--worker-extension-cls`.

Each transport owns its full sync() (including the non-main-rank collective walk), so the trainer's `_sync_weight` is
backend-agnostic.

 #### E2E test.
 Tested end-to-end (Qwen3-0.6B, GSM8K), all three paths: clean training (importance ratio ≈ 1.0) with per-sync times≈  0.1 s for NCCL sparse/full and ≈ 10–12 s for the bucket path (Xet transfer-bound). `extract_sparse_batched` is chunked so  the nonzero buffer stays bounded on multi-billion-param models.

#### Next step:

~~FSDP and distributed optimizer.~~

AmineDiro:
Support for FSDP sharding implemented in : [1562e53](https://github.com/huggingface/trl/pull/5937/commits/1562e5362d14e76038ce7740b3885441bc11f960)
## 8B end-to-end: FULL vs SPARSE weight sync

**Setup:** Qwen3-8B, FSDP2 — 2 trainer GPUs + 1 vLLM

| mode | sync kind | n | median/sync |
|------|-----------|---|-------------|
| SPARSE | sparse delta | 10 | 0.1 s |
| SPARSE | anchor (every 5th = full) | 3 | 0.3 s |
| FULL | full broadcast (every sync) | 13 | 0.3 s |

**Takeaway:** 
- At 8B, a sparse delta sync (~0.1 s) is ~3× faster than a full broadcast (~0.3 s). 
-  The gap widened vs 4B (where it was ~2×, 0.1 vs 0.2 s), as expected: the full broadcast scales with model size (16 GB at 8B) while the sparse delta scales with the ~1% that changed. 

Tested on more models: 

| model | sparse | full |
|-------|--------|------|
| 1.7B | ~0.1 s | ~0.2 s |
| 4B | ~0.1 s | ~0.2 s |
| 8B | ~0.1 s | ~0.3 s |

Sparse stays ~flat (~0.1 s, dominated by fixed per-param overhead at high sparsity) while full grows with model size, so the sparse advantage compounds as models get bigger.

FSDP was tested using the following config :

```yaml
distributed_type: FSDP
fsdp_config:
  fsdp_version: 2
  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
  fsdp_cpu_ram_efficient_loading: true
  fsdp_offload_params: false
num_machines: 1
num_processes: 2
mixed_precision: bf16
```

start vllm : 
```sh
CUDA_VISIBLE_DEVICES=2 VLLM_SERVER_DEV_MODE=1 VLLM_USE_V2_MODEL_RUNNER=0 vllm serve "$MODEL" \
  --model-impl transformers --port 8000 --max-model-len 2048 --gpu-memory-utilization 0.85 \
  --logprobs-mode processed_logprobs \
  --weight-transfer-config '{"backend":"nccl"}' 
```

Spawn trainer on 2GPUs using accelerate:
```sh
CUDA_VISIBLE_DEVICES=0,1 accelerate launch --config_file dev_delta_v2/fsdp2_2proc.yaml \
  dev_delta_v2/e2e_sparse_nccl_driver.py 
```

`e2e_sparse_nccl_driver.py` main config: 
```python
def main():
    dataset = load_dataset("openai/gsm8k", "main", split="train")
    dataset = dataset.map(format_sample, remove_columns=dataset.column_names)

    config = AsyncGRPOConfig(
        save_strategy="no",
        num_generations=8,
        max_completion_length=512,
        max_steps=10,
        learning_rate=1e-5,
        logging_steps=1,
        bf16=True,
        report_to="none",
        log_completions=True,
        weight_sync_anchor_interval=5,
    )
    trainer = AsyncGRPOTrainer(
        model=os.environ.get("MODEL", "Qwen/Qwen3-0.6B"),
        args=config,
        train_dataset=dataset,
        reward_funcs=accuracy_reward,
    )
    trainer.train()
```

qgallouedec (approved):
Thanks for this. Really impressive work. It's been open for a while, and we've to move forward so I'll approve it.

To be upfront: I'm not against merging it. My only hesitation is surface area: it's a big, deep feature (new detector, codec, two transports, a vLLM-side engine, plus a dependency specific vllm version) that *has* to keep working alongside the rest of the features, including the ones to come. That's a real long-term maintenance commitment.

So: if you feel you can own this and keep it working over the long run, feel free to merge. You know this code better than anyone, and I'd rather have you as its maintainer than see it merged and bit-rot. 🙏

One correctness thing worth fixing first, though: the AdamW inversion only recovers the last optimizer step, so with weight_sync_steps > 1 the sparse delta silently misses weights changed in the earlier steps of the interval. Since sparse is the default, worth either guarding against weight_sync_steps > 1 or accumulating masks across the interval.

AmineDiro:
> To be upfront: I'm not against merging it. My only hesitation is surface area: it's a big, deep feature (new detector, codec, two transports, a vLLM-side engine, plus a dependency-specific vLLM version) that _has_ to keep working alongside the rest of the features, including the ones to come. That's a real long-term maintenance commitment.

Thanks for the review 🙏🏼 . I do agree with your take, this is a lot of stuff to land, and will definitely need continuous maintenance. But I think that's a manageable risk with a cutting-edge feature :
-  most of the "surface area" is sandboxed. All long-term maintenance is in one file: delta_engine.py 
- the sparse/bucket feature carries its own `delta_weight_sync` flag. So It doesn't force the whole package's dependency resolution for a user, and we can always default to the "stable" full weight sync

So in my view, we can land it

> One correctness thing worth fixing first, though: the AdamW inversion only recovers the last optimizer step, so with weight_sync_steps > 1 the sparse delta silently misses weights changed in the earlier steps of the interval. Since sparse is the default, worth either guarding against weight_sync_steps > 1 or accumulating masks across the interval.

That's a very nice catch ! I'll fix it in the config for now ( check that:
```python 
if self.weight_sync_mode == "sparse" and self.weight_sync_steps > 1:
```

AmineDiro:
Before merging, ran some experiments on `vllm==0.23.0` to be sure there isn't any blind spots. I tested both dense and MoE models in different configs.

Here is the table of full vs sparse, MoE are just not supported with sparse updates: 

| sync   | dense TP=1  |     dense TP>1      | dense DP |      MoE (EP=1)      |         MoE + EP         |
| ------ | :---------: | :-----------------: | :------: | :------------------: | :----------------------: |
| full   |     ✅      |       ✅         |  ✅   |       ✅ (IE)        |      ✅ (IE, EP=4)       |
| sparse | ✅ baseline | ❌ vLLM refuses (C) |  ✅ (H)  | ❌ fused-expert name | ❌ same   fused-expert|

IE: HF inference endpoints for newer CUDA drivers

sergiopaniego (commented):
great effort!
some nits I think we can address before merging:
- `examples/scripts/async_grpo_delta.py` is missing the # /// script PEP 723 dependency block
- `weight_diff.py` and `delta_codec.py` have a duplicated copyright header
- `async_grpo_delta.py` isn't linked anywhere in the docs.
- example_overview.md doesn't list any async_grpo script (not async_grpo.py, async_grpo_delta.py, nor async_grpo_buckets/)
- examples/scripts/async_grpo_buckets link in async_grpo_trainer.md points to /tree/main/…

qgallouedec:
Re-reviewing this since my approval predates the rewrite, the branch is essentially a new PR now.

Good call building on vLLM's native sparse transfer from 0.23.0 rather than rolling your own transport.

Two things I would like before merging, both about failing loudly:

- The docs say FSDP2 only and that ZeRO is not supported, but nothing enforces it. The param matching goes through `data_ptr()`, and under ZeRO the optimizer holds flattened partitions, so nothing matches, `compute_masks()` returns an empty dict and no weights get synced at all. The only trace is an info log saying `matched 0/N`. Since rollouts then run on a stale policy with no error, can you raise at init instead?

- `isinstance(optimizer, torch.optim.AdamW)` still passes with `amsgrad=True`, where the denominator uses `max_exp_avg_sq`. The inversion is then wrong and the mask drops real changes. A check on the state dict should be enough.

Also, could the Dockerfiles and Space apps under `examples/scripts/async_grpo_buckets/` go in a follow-up? It is a lot of deployment scaffolding riding along with a correctness sensitive change.

This pr is old now, if it easier for you to just restart from scratch from main, it's also an option, your call.

qgallouedec:
Closing this one for now to keep the queue tidy. The branch is conflicting, and the last update is a month ago.

The measured work is the valuable part and it doesn't expire though.

Happy to see it again as a fresh branch off main whenever it's the right time, with the ZeRO guard and the `amsgrad` check in, and the Dockerfiles and Space apps split out.

PR #5943, opened 2026-06-04
Support vision datasets for Liger in DPO

albertvillanova:
Support vision datasets for Liger in DPO.

This PR introduces support and testing for vision-language models (VLMs) using the Liger kernel in the DPOTrainer. The main changes include updates to the Liger loss computation to ensure proper handling of vision inputs by invoking the full model backbone instead of just the decoder, and a new test for VLM training with Liger.

Follow-up to review comment: https://github.com/huggingface/trl/pull/5939#discussion_r3356605875

### Motivation

The Liger loss path in `_compute_loss_liger` was calling `model.get_decoder()`, which on VLMs returns only the text decoder and bypasses the vision encoder entirely. Visual inputs (`pixel_values`, `image_grid_thw`, etc.) were also not forwarded.

### Solution

This PR replaces `model.get_decoder()` with `model.model`, the backbone attribute that, for VLMs, includes the vision encoder before the language decoder. This is the same pattern`GRPOTrainer._get_last_hidden_state` already uses, with an explicit comment explaining why `get_decoder()` is wrong for VLMs.

VLM keys are now forwarded using the same exclusion-based `model_kwargs` construction that `_compute_loss` already uses, so any key the processor emits (`pixel_values`, `image_grid_thw`, `pixel_attention_mask`, `image_sizes`, `image_position_ids`, etc.) is forwarded automatically.

### Changes

**Liger Loss Computation Updates:**

* Modified `_compute_loss_liger` in `DPOTrainer` to call the full model backbone (`model.model`) instead of just the decoder, ensuring that vision inputs (e.g., `pixel_values`) are processed correctly for VLMs. This change is applied both for the policy and reference models.
* Cleaned up input handling by filtering out non-model keys and passing all relevant inputs (including vision-related ones) to the model.

**Vision-Language Model (VLM) Support:**

* Added a new test `test_train_vlm_liger` for DPO to verify that VLMs can be trained with the Liger kernel, including checks that model parameters are updated after training.

---

> [!NOTE]
> **Medium Risk**
> Changes core Liger loss forward path for all models (backbone selection and kwargs forwarding); behavior for text-only LMs should be equivalent but VLM and transformers version branches add regression surface.
> 
> **Overview**
> **DPO Liger** now supports vision-language preference training by fixing how hidden states are computed for VLMs.
> 
[...]

PR #5949, opened 2026-06-05
Fix backbone access in GRPO by aligning with SFT

albertvillanova:
Fix backbone access in GRPO aligning with SFT.

This PR aligns `GRPOTrainer` with `SFTTrainer` by fixing how the last hidden state is retrieved from models in the `_get_last_hidden_state` method. This improves compatibility with different versions of the `transformers` library and better handling of vision-language models (VLMs).

Follow-up to this review comment: https://github.com/huggingface/trl/pull/5943#discussion_r3361794937

### Motivation

In `GRPOTrainer._get_last_hidden_state`, the backbone is accessed via `unwrapped_model.model`: a hardcoded attribute name that assumes `base_model_prefix == "model"`. 

`SFTTrainer` uses the more robust `model.base_model`, which is a `PreTrainedModel` property that resolves to `getattr(self, self.base_model_prefix)` and works correctly across all model families.

### Solution

This PR aligns `GRPOTrainer` with `SFTTrainer`'s established pattern:
- Use `unwrapped_model.base_model` in the general case (text LMs and VLMs on transformers ≥ 5.0).
- Fall back to `unwrapped_model.model` only for pre-5.0 VLMs, where `base_model_prefix = ""` makes `base_model is self` (which would re-run `lm_head`).

The comment explaining the choice is taken verbatim from SFTTrainer.

### Changes

Model compatibility and robustness improvements:

* Updated the logic to select the correct model backbone (`base_model` or `model`) depending on whether the model is a VLM and the version of `transformers`, ensuring that the correct module is used for both language models and VLMs, and avoiding issues with pre-5.0 transformers where `base_model_prefix` is set to an empty string.
* Added explanatory comments clarifying why `base_model` or `model` is chosen and the pitfalls of using `get_decoder()` with VLMs.

---

> [!NOTE]
> **Medium Risk**
> Changes which submodule runs during GRPO/Liger loss hidden-state extraction, so wrong selection would skew policy gradients; scope is limited to backbone routing, not reward or generation logic.
> 
> **Overview**
> **GRPO** `_get_last_hidden_state` no longer hardcodes `unwrapped_model.model` for hidden states. It now picks the backbone the same way as **SFT**: `unwrapped_model.base_model` by default, and `unwrapped_model.model` only for VLMs on **transformers &lt; 5.0** (where `base_model` would alias the full model and re-run `lm_head`). The same inline comment as SFT documents why `get_decoder()` is avoided for VLMs.
> 
> **SFT** only renames the local `decoder` variable to `backbone` and tightens the comment wording—behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f9d2efb99f64d71bfe689bd26fd3976858f7460. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -997,7 +997,7 @@ class GRPOTrainer(_BaseTrainer):
 
         model_inputs["use_cache"] = False  # only used in generation; set False to suppress warnings
 
-        # `base_model` gives the inner module (skipping `lm_head`) — text decoder for LMs, multimodal wrapper for
+        # `base_model` gives the backbone model (skipping `lm_head`) — text decoder for LMs, multimodal wrapper for
         # VLMs (so vision-token injection runs before the text decoder). `get_decoder()` won't do: on VLMs it
         # returns just the text stack and feeds image-placeholder IDs through it.
         # Pre-5.0 transformers VLMs set `base_model_prefix = ""` so `base_model is self` (re-runs `lm_head`).
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -997,7 +997,16 @@ class GRPOTrainer(_BaseTrainer):
 
         model_inputs["use_cache"] = False  # only used in generation; set False to suppress warnings
 
-        last_hidden_state = unwrapped_model.model(**model_inputs).last_hidden_state
+        # `base_model` gives the inner module (skipping `lm_head`) — text decoder for LMs, multimodal wrapper for
+        # VLMs (so vision-token injection runs before the text decoder). `get_decoder()` won't do: on VLMs it
+        # returns just the text stack and feeds image-placeholder IDs through it.
+        # Pre-5.0 transformers VLMs set `base_model_prefix = ""` so `base_model is self` (re-runs `lm_head`).
+        # Fall back to `.model` there.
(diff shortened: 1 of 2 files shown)


qgallouedec (approved):
Thanks!!

trl/trainer/grpo_trainer.py
```diff
@@ -997,7 +997,16 @@ def _get_last_hidden_state(
...
+        # `base_model` gives the inner module (skipping `lm_head`) — text decoder for LMs, multimodal wrapper for
+        # VLMs (so vision-token injection runs before the text decoder). `get_decoder()` won't do: on VLMs it
+        # returns just the text stack and feeds image-placeholder IDs through it.
+        # Pre-5.0 transformers VLMs set `base_model_prefix = ""` so `base_model is self` (re-runs `lm_head`).
+        # Fall back to `.model` there.
+        if self._is_vlm and Version(transformers.__version__) < Version("5.0.0"):
+            backbone = unwrapped_model.model
+        else:
+            backbone = unwrapped_model.base_model
+        last_hidden_state = backbone(**model_inputs).last_hidden_state
```
qgallouedec, line 1009: should we use `backbone` everywhere?

albertvillanova: I agree we should align. Doing it.

PR #5988, opened 2026-06-09
[CI] Check that training chat templates keep the stop token in the loss mask

kashif:
A chat template can pass the prefix-preservation check and still attribute the assistant's end-of-turn token to the *following* message. With `assistant_only_loss=True` that token then falls outside the assistant mask, and the model never learns to stop. SFT runs fine, loss goes down, and the bug only shows up at inference as endless generations.

This PR adds a check for that property, gates the shipped training templates on it in CI, and fixes the one template the gate caught:

- `is_chat_template_stop_token_trained`: renders an assistant turn followed by a user message and checks that the assistant's masked span ends on an end-of-turn token (added vocab) rather than on content. Lives next to `is_chat_template_prefix_preserving`, same calling convention.
- A new test in `TestGetTrainingChatTemplate` runs the check against every supported family. It found a real case: GLM closes the assistant turn with the next role marker (`<|user|>`/`<|observation|>`), so the terminator was never trained.
- `glm4moe_training.jinja` is fixed accordingly: the following message's role marker is emitted inside the assistant's `{% generation %}` block (and suppressed at the start of the next message), so it lands in the loss mask. The rendered text is byte-identical to the original template, and prefix preservation still holds.
- Known processor limitations (Llava, Gemma3, Qwen2-VL, Qwen2.5-VL don't propagate the assistant tokens mask) are `xfail(strict=True)`: CI stays green today but flips red the moment something is fixed upstream, so the markers can't go stale.
- SFT now logs a warning when `assistant_only_loss=True` and the resolved template has this problem.

Related to #4879 (auto-adding `{% generation %}` tags): this is the complementary verification side. Once a template has generation markers, this checks they actually cover the stop token.

---

> [!NOTE]
> **Medium Risk**
> Changes SFT chat-template selection/warnings and alters GLM-4-MoE training template masking behavior; impact is limited to assistant-only loss and template patching paths, not general inference templates.
> 
> **Overview**
> Adds **`is_chat_template_stop_token_trained`** to verify that, with `assistant_only_loss`, the assistant’s end-of-turn token is inside the `{% generation %}` loss mask—not attributed to the next message (a failure mode that can cause endless generation at inference).
> 
> **`has_generation_markers`** replaces brittle `{% generation %}` string checks (including `{%- generation %}` variants) in `get_training_chat_template` and SFT setup.
> 
> **CI** gains unit tests plus a per-family check that shipped training templates pass the stop-token gate; **GLM-4-MoE** `glm4moe_training.jinja` is fixed so the following role marker is emitted inside the assistant generation block (rendered text unchanged). **SFT** logs a warning when `assistant_only_loss=True` and the resolved template still fails the check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3275bfaa6969a610afa6a710d0e94b68794f31b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_chat_template_utils.py
+++ b/tests/test_chat_template_utils.py
@@ -684,6 +684,64 @@ class TestGetTrainingChatTemplate:
             pytest.skip("Template does not support tool calling; prefix-preservation check is not applicable.")
         assert is_chat_template_prefix_preserving(tokenizer) is True
 
+    @pytest.mark.parametrize("append_role", ["user", "system"])
+    def test_new_chat_template_is_prefix_preserving_for_appended_role(self, tokenizer_name, append_role, request):
+        # Multi-turn rollouts can append user or system messages mid-conversation, and the same prefix-preservation
+        # property applies per role. Templates that reject a role outright are skipped (failing loud is safe);
+        # templates that silently rewrite earlier turns are known breaks, marked as strict xfail so CI flags them as
+        # soon as a template gains the surface. None are fixable byte-identically: the rewrites are inherent to the
+        # original templates.
+        known_breaks = {
+            ("trl-internal-testing/tiny-GptOssForCausalLM", "user"): (
+                "appending demotes the final assistant turn and rewrites its <|return|> terminator to <|end|>"
+            ),
+            ("trl-internal-testing/tiny-GptOssForCausalLM", "system"): (
+                "appending demotes the final assistant turn and rewrites its <|return|> terminator to <|end|>"
+            ),
+            ("trl-internal-testing/tiny-Phi3ForCausalLM-3", "user"): (
+                "the final render carries a trailing eos token that appending removes"
+            ),
+            ("trl-internal-testing/tiny-Phi3ForCausalLM-3", "system"): (
+                "the final render carries a trailing eos token that appending removes"
+            ),
+            ("trl-internal-testing/tiny-Phi3ForCausalLM-3.5", "user"): (
+                "the final render carries a trailing eos token that appending removes"
+            ),
+            ("trl-internal-testing/tiny-Phi3ForCausalLM-3.5", "system"): (
+                "the final render carries a trailing eos token that appending removes"
+            ),
+            ("trl-internal-testing/tiny-Cohere2ForCausalLM", "system"): (
+                "the system message is hoisted into the developer preamble at the front of the render"
+            ),
+            ("trl-internal-testing/tiny-DeepseekV3ForCausalLM", "system"): (
+                "the system message is hoisted to the front of the render"
+            ),
+            ("trl-internal-testing/tiny-LlavaForConditionalGeneration", "user"): (
+                "the role markers are plain text, and the trailing space merges with them under BPE"
+            ),
+            ("trl-internal-testing/tiny-LlavaForConditionalGeneration", "system"): (
+                "the role markers are plain text, and the trailing space merges with them under BPE"
+            ),
+            ("trl-internal-testing/tiny-LlavaNextForConditionalGeneration", "user"): (
+                "the role markers are plain text, and the trailing space merges with them under BPE"
+            ),
+            ("trl-internal-testing/tiny-LlavaNextForConditionalGeneration", "system"): (
+                "the role markers are plain text, and the trailing space merges with them under BPE"
+            ),
+        }
+        reason = known_breaks.get((tokenizer_name, append_role))
+        if reason:
+            request.node.add_marker(pytest.mark.xfail(strict=True, reason=f"{tokenizer_name}: {reason}"))
+        tokenizer = self._load(tokenizer_name)
+        chat_template = get_training_chat_template(tokenizer)
+        if chat_template is not None:
+            tokenizer.chat_template = chat_template
+        try:
+            preserved = is_chat_template_prefix_preserving(tokenizer, append_role=append_role)
+        except TemplateError:
+            pytest.skip(f"Template does not support appending a {append_role} message.")
+        assert preserved is True
+
     def test_new_chat_template_trains_stop_token(self, tokenizer_name, request):
         # Known failures, marked as strict xfail rather than skip so that CI flags them (and gains coverage) as soon
         # as the underlying processor is fixed.
--- a/tests/test_chat_template_utils.py
+++ b/tests/test_chat_template_utils.py
@@ -677,7 +677,7 @@ class TestGetTrainingChatTemplate:
     def test_new_chat_template_is_prefix_preserving(self, tokenizer_name):
         tokenizer = self._load(tokenizer_name)
         new_chat_template = get_training_chat_template(tokenizer)
-        if chat_template is not None:
+        if new_chat_template is not None:
             tokenizer.chat_template = new_chat_template
         # Prefix-preservation is only meaningful for templates that actually support tool messages — the check
         # itself renders one. Skip the assertion for tool-less templates (e.g. Gemma).
--- a/tests/test_chat_template_utils.py
+++ b/tests/test_chat_template_utils.py
@@ -556,6 +556,11 @@ class TestIsChatTemplateStopTokenTrained:
         {%- endif %}""")
         assert is_chat_template_stop_token_trained(tokenizer) is False
 
+    def test_template_error_returns_false(self):
+        tokenizer = AutoTokenizer.from_pretrained("trl-internal-testing/tiny-Qwen3MoeForCausalLM")
+        tokenizer.chat_template = "{{ raise_exception('probe rejected') }}"
+        assert is_chat_template_stop_token_trained(tokenizer) is False
+
 
 @pytest.mark.parametrize(
     "tokenizer_name",
--- a/tests/test_chat_template_utils.py
+++ b/tests/test_chat_template_utils.py
@@ -691,7 +691,7 @@ class TestGetTrainingChatTemplate:
             "trl-internal-testing/tiny-LlavaNextForConditionalGeneration",
         ):
             reason = f"{tokenizer_name}: the processor returns an all-zero assistant tokens mask"
-            request.node.add_marker(pytest.mark.xfail(strict=True, reason=reason))
+            request.node.add_marker(pytest.mark.xfail(strict=False, reason=reason))
         tokenizer = self._load(tokenizer_name)
         new_chat_template = get_training_chat_template(tokenizer)
         assert is_chat_template_stop_token_trained(tokenizer, chat_template=new_chat_template) is True
--- a/tests/test_chat_template_utils.py
+++ b/tests/test_chat_template_utils.py
@@ -676,9 +676,9 @@ class TestGetTrainingChatTemplate:
 
     def test_new_chat_template_is_prefix_preserving(self, tokenizer_name):
         tokenizer = self._load(tokenizer_name)
-        chat_template = get_training_chat_template(tokenizer)
+        new_chat_template = get_training_chat_template(tokenizer)
         if chat_template is not None:
-            tokenizer.chat_template = chat_template
+            tokenizer.chat_template = new_chat_template
         # Prefix-preservation is only meaningful for templates that actually support tool messages — the check
         # itself renders one. Skip the assertion for tool-less templates (e.g. Gemma).
         if not supports_tool_calling(tokenizer):
--- a/tests/test_chat_template_utils.py
+++ b/tests/test_chat_template_utils.py
@@ -25,6 +25,7 @@ from trl.chat_template_utils import (
     add_response_schema,
     get_training_chat_template,
     is_chat_template_prefix_preserving,
+    is_chat_template_stop_token_trained,
     parse_response,
     supports_tool_calling,
 )
@@ -513,6 +514,48 @@ class TestIsChatTemplatePrefixPreserving:
         assert is_chat_template_prefix_preserving(processor) is True
 
 
+class TestIsChatTemplateStopTokenTrained:
+    def test_stop_token_trained(self):
+        tokenizer = AutoTokenizer.from_pretrained("trl-internal-testing/tiny-Qwen3MoeForCausalLM")
+        # The assistant turn is closed by <|im_end|> inside the generation span, so the end-of-turn token is masked
+        # in and the model is trained to stop.
+        # docstyle-ignore
+        tokenizer.chat_template = textwrap.dedent(r"""
+        {%- for message in messages %}
+        {%- if message.role == 'user' %}
+            {{- '<|im_start|>user\n' + message.content + '<|im_end|>\n' }}
+        {%- elif message.role == 'assistant' %}
+            {{- '<|im_start|>assistant\n' }}
+            {%- generation %}{{- message.content + '<|im_end|>' }}{%- endgeneration %}
+            {{- '\n' }}
+        {%- endif %}
+        {%- endfor %}
+        {%- if add_generation_prompt %}
+            {{- '<|im_start|>assistant\n' }}
+        {%- endif %}""")
+        assert is_chat_template_stop_token_trained(tokenizer) is True
+
+    def test_stop_token_not_trained(self):
+        tokenizer = AutoTokenizer.from_pretrained("trl-internal-testing/tiny-Qwen3MoeForCausalLM")
+        # GLM-style: the assistant's end-of-turn token is emitted as the prefix of the following message, so the
+        # generation span covers content only and the model is never trained to stop.
+        # docstyle-ignore
+        tokenizer.chat_template = textwrap.dedent(r"""
+        {%- for message in messages %}
+        {%- if message.role == 'user' %}
+            {{- '<|im_start|>user\n' + message.content + '<|im_end|>\n' }}
+        {%- elif message.role == 'assistant' %}
+            {{- '<|im_start|>assistant\n' }}
+            {%- generation %}{{- message.content }}{%- endgeneration %}
+            {{- '<|im_end|>\n' }}
+        {%- endif %}
+        {%- endfor %}
+        {%- if add_generation_prompt %}
+            {{- '<|im_start|>assistant\n' }}
+        {%- endif %}""")
+        assert is_chat_template_stop_token_trained(tokenizer) is False
+
+
 @pytest.mark.parametrize(
     "tokenizer_name",
     [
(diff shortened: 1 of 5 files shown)


kashif:
the CI found an issue with the GLM template, fixing that next

trl/chat_template_utils.py
```diff
@@ -563,6 +563,72 @@ def is_chat_template_prefix_preserving(processing_class: PreTrainedTokenizerBase
...
+    input_ids = output["input_ids"]
+    assistant_masks = output["assistant_masks"]
+    if is_vlm:
+        input_ids = input_ids[0]
+        assistant_masks = assistant_masks[0]
+
+    # The model stops by emitting an end-of-turn token, which is part of the added vocabulary rather than produced by
+    # the base tokenizer's merges. Ignoring trailing whitespace, the last masked token must be that terminator; if it
+    # is plain content, the end-of-turn token was attributed to the following message and is never trained.
+    tokenizer = getattr(processing_class, "tokenizer", processing_class)
```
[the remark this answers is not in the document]
kashif: Done in 8471ac8 — replaced the getattr fallback with an explicit `processing_class.tokenizer if is_vlm else processing_class` branch.

trl/chat_templates/glm4moe_training.jinja
```diff
@@ -49,7 +53,8 @@ For each function call, output the function name and arguments within the follow
     {%- endif %}
 {%- endfor %}
 {% for m in messages %}
-{%- if m.role == 'user' -%}<|user|>
+{%- if m.role == 'user' -%}
+{{- '<|user|>' if loop.first or messages[loop.index0 - 1].role != 'assistant' }}
```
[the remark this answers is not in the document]
kashif: Good catch — fixed in e3c4bb6: the assistant block now suppresses its leading `<|assistant|>` when the previous message is an assistant, same as the other roles. Added `test_behavior_unchanged_consecutive_assistant_messages` covering this shape for every supported family.

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -256,6 +257,17 @@ def __init__(
...
+        # A prefix-preserving template can still attribute the assistant's end-of-turn token to the next message,
+        # leaving it out of the loss mask so the model is never trained to stop. Warn if the template we resolved
+        # has that issue.
+        if self.chat_template is not None and not is_chat_template_stop_token_trained(
+            self.tokenizer, chat_template=self.chat_template
+        ):
+            logger.warning(
+                "The chat template does not include the assistant turn's end-of-turn token in the loss mask; "
+                "the model may not learn to stop."
+            )
```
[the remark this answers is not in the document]
kashif: Fixed in e3c4bb6: the check now also runs against the tokenizer's default template when no training template was swapped in, gated on the template defining `{% generation %}` markers (without them the assistant mask is empty and the check is meaningless).

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -256,6 +257,19 @@ def __init__(
...
             self.chat_template = None
 
+        # A prefix-preserving template can still attribute the assistant's end-of-turn token to the next message,
+        # leaving it out of the loss mask so the model is never trained to stop. Warn if the template we resolved
+        # has that issue. The check only applies to templates with `{% generation %}` markers; without them the
+        # assistant mask is empty and the check is meaningless.
+        chat_template = self.chat_template or self.tokenizer.chat_template
+        if "{% generation %}" in chat_template and not is_chat_template_stop_token_trained(
+            self.tokenizer, chat_template=chat_template
+        ):
```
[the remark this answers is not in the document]
kashif: Right, the shipped templates use the trim variants (`{%- generation %}`, `{%- generation -%}`). Fixed in 864cb7b with a regex that matches all of them.

trl/chat_template_utils.py
```diff
@@ -563,6 +563,75 @@ def is_chat_template_prefix_preserving(processing_class: PreTrainedTokenizerBase
...
+    try:
+        output = processing_class.apply_chat_template(
+            messages,
+            tokenize=True,
+            return_dict=True,
+            return_assistant_tokens_mask=True,
+            chat_template=chat_template,
+        )
+    except (TypeError, ValueError):
+        return False
```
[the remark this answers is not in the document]
kashif: Fixed in 72dc836: `TemplateError` is now caught alongside `TypeError`/`ValueError`, with a regression test.

trl/chat_template_utils.py
```diff
@@ -502,33 +511,44 @@ def supports_tool_calling(processing_class) -> bool:
     return all(s in rendered for s in (_name_sentinel, _arg_key_sentinel, _arg_val_sentinel, _content_sentinel))
 
 
-def is_chat_template_prefix_preserving(processing_class: PreTrainedTokenizerBase | ProcessorMixin) -> bool:
+def is_chat_template_prefix_preserving(
+    processing_class: PreTrainedTokenizerBase | ProcessorMixin, append_role: str = "tool"
+) -> bool:
```
qgallouedec: I don't understand this change, why do we need this?

trl/chat_template_utils.py
```diff
@@ -25,6 +26,14 @@
 _CHAT_TEMPLATES_DIR = Path(__file__).parent / "chat_templates"
 
 
+def has_generation_markers(chat_template: str) -> bool:
+    """
+    Check whether the chat template defines `{% generation %}` markers, accounting for whitespace-trim variants such as
+    `{%- generation %}` and `{%- generation -%}`.
+    """
+    return re.search(r"\{%-?\s*generation\s*-?%\}", chat_template) is not None
```
qgallouedec, line 34: instead of this, why not consistently use `"{% generation %}"`. ie we would modify the training templates

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -256,6 +258,21 @@ def __init__(
...
 
+        # A prefix-preserving template can still attribute the assistant's end-of-turn token to the next message,
+        # leaving it out of the loss mask so the model is never trained to stop. Warn if the template we resolved
+        # has that issue. The check only applies to templates with `{% generation %}` markers; without them the
+        # assistant mask is empty and the check is meaningless.
+        chat_template = self.chat_template or self.tokenizer.chat_template
+        if (
+            chat_template
+            and has_generation_markers(chat_template)
+            and not is_chat_template_stop_token_trained(self.tokenizer, chat_template=chat_template)
```
qgallouedec: I don't understand. do we use assistant mask from the chat template anywhere in grpo? I think it's only in sft

qgallouedec (commented):
thanks! a few questions

trl/chat_template_utils.py
kashif: the property turns out to be role-specific: gpt-oss for example is prefix-preserving for tool messages but not for user messages (appending one rewrites the final `<|return|>` to `<|end|>`). The new test documents which surfaces hold per family, with the known breaks as strict xfails. Happy to split it into a follow-up PR if you prefer?

trl/chat_template_utils.py
kashif: tried that first, but the trim markers are load-bearing: rewriting `{%- generation %}` to `{% generation %}` in `qwen3_training.jinja` inserts a stray newline (`assistant\n\n<think>` instead of `assistant\n<think>`), so the render is no longer byte-identical to the original template. The check also runs on user templates from the Hub where we do not control the marker style.

trl/experimental/async_grpo/async_rollout_worker.py
kashif: you are right, only SFT uses the assistant mask. Moved the warning to sft_trainer (gated on assistant_only_loss) and dropped it from the worker in 7fdc008.

trl/chat_template_utils.py
qgallouedec: you're right, but we never need to be prefix-preserving for another role than tool. That's why I'm not sure if adding this is actually needed

tests/test_chat_template_utils.py
```diff
@@ -635,6 +684,79 @@ def test_new_chat_template_is_prefix_preserving(self, tokenizer_name):
             pytest.skip("Template does not support tool calling; prefix-preservation check is not applicable.")
         assert is_chat_template_prefix_preserving(tokenizer) is True
 
+    @pytest.mark.parametrize("append_role", ["user", "system"])
```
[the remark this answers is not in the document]
kashif: good catch, fixed in a84f2f7: the template is only swapped when the helper returns one, so the stock template gets checked as is when it is already training-ready.

trl/chat_template_utils.py
kashif: fair enough, the trainers only ever append tool messages. Dropped the role parameter and the extra test in 8d8beb5, so the PR is back to just the stop-token check and the GLM fix. Can revisit if we ever grow a user-appending surface.

trl/chat_template_utils.py
qgallouedec: fair enough

tests/test_chat_template_utils.py
```diff
@@ -628,13 +677,30 @@ def _load(self, model_name):
...
+    def test_new_chat_template_trains_stop_token(self, tokenizer_name, request):
+        # Known failures, marked as strict xfail rather than skip so that CI flags them (and gains coverage) as soon
+        # as the underlying processor is fixed.
+        if tokenizer_name in (
+            "trl-internal-testing/tiny-LlavaForConditionalGeneration",
+            "trl-internal-testing/tiny-Gemma3ForConditionalGeneration",
+            "trl-internal-testing/tiny-Qwen2VLForConditionalGeneration",
+            "trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration",
+        ):
+            reason = f"{tokenizer_name}: processor does not propagate the assistant tokens mask"
```
qgallouedec: oh no, really?? :/

kashif: yeah :( the mask comes back all-zero through these processors. The newer ones (Qwen3-VL, Qwen3.5, Qwen3.6) propagate it fine, so it is only the older generation. It is latent for now since we block `assistant_only_loss` for VLMs anyway, and the strict xfail flips red the moment transformers fixes it. Can open an issue upstream if useful.

tests/test_chat_template_utils.py
```diff
@@ -676,6 +742,39 @@ def test_behavior_unchanged_single_user_and_final_assistant_plain_content(self,
         after = tokenizer.apply_chat_template(messages, tokenize=False, chat_template=new_chat_template)
         assert before == after
 
+    def test_behavior_unchanged_consecutive_assistant_messages(self, tokenizer_name):
```
qgallouedec: I don't think this is something we support nor we should support.

tests/test_chat_template_utils.py
qgallouedec: are you sure?

```python

from trl.chat_template_utils import get_training_chat_template

# A VLM processor whose default chat template has no `{% generation %}` markers
processor = AutoProcessor.from_pretrained("Qwen/Qwen2.5-VL-3B-Instruct")
assert "{% generation %}" not in processor.chat_template

# Patch in a training-compatible template that adds the markers
chat_template = get_training_chat_template(processor)
assert "{% generation %}" in chat_template

messages = [
    {"role": "user", "content": [{"type": "text", "text": "Hi"}]},
    {"role": "assistant", "content": [{"type": "text", "text": "Hello there"}]},
]

out = processor.apply_chat_template(
    messages,
    tokenize=True,
    return_dict=True,
    return_assistant_tokens_mask=True,
    chat_template=chat_template,
)

ids, mask = out["input_ids"][0], out["assistant_masks"][0]
tokens = processor.tokenizer.convert_ids_to_tokens(ids)

print("keys:", list(out.keys()))
print("assistant tokens:", [t for t, m in zip(tokens, mask) if m == 1])
```

```
keys: ['input_ids', 'attention_mask', 'mm_token_type_ids', 'assistant_masks']
assistant tokens: ['Hello', 'Ġthere', '<|im_end|>', 'Ċ']

Comment (qgallouedec) [reply tests/test_chat_template_utils.py]: Ah it seems to be only when images are provided, strange:

```python
from PIL import Image

from trl.chat_template_utils import get_training_chat_template, prepare_multimodal_messages

# A VLM processor whose default chat template has no `{% generation %}` markers
processor = AutoProcessor.from_pretrained("Qwen/Qwen2.5-VL-3B-Instruct")
assert "{% generation %}" not in processor.chat_template

# Patch in a training-compatible template that adds the markers
chat_template = get_training_chat_template(processor)
assert "{% generation %}" in chat_template


def assistant_tokens(messages):
    out = processor.apply_chat_template(
        messages,
        tokenize=True,
        return_dict=True,
        return_assistant_tokens_mask=True,
        chat_template=chat_template,
    )
    ids, mask = out["input_ids"][0], out["assistant_masks"][0]
    tokens = processor.tokenizer.convert_ids_to_tokens(ids)
    return [t for t, m in zip(tokens, mask) if m == 1]


# 1) Text-only: the assistant tokens mask is propagated correctly
text_only = [
    {"role": "user", "content": [{"type": "text", "text": "Hi"}]},
    {"role": "assistant", "content": [{"type": "text", "text": "Hello there"}]},
]
print("text-only :", assistant_tokens(text_only))

# 2) With an image: the same call returns an EMPTY mask (upstream processor bug)
with_image = prepare_multimodal_messages(
    [
        {"role": "user", "content": "Hi"},
        {"role": "assistant", "content": "Hello there"},
    ],
    images=[Image.new("RGB", (8, 8))],
)
print("with image:", assistant_tokens(with_image))
```

```
text-only : ['Hello', 'Ġthere', '<|im_end|>', 'Ċ']
with image: []
```

kashif: yes, exactly: text-only works, with images it breaks. And it gets stranger: on Qwen2.5-VL the mask comes back all-zero, but on Qwen3-VL it comes back shifted onto two `<|image_pad|>` tokens, so the mask is computed before image expansion and never re-aligned. Silently wrong is worse than zero. I changed the check in 3c3913f to probe VLMs text-only so it tests the template rather than the processor; with that only llava/llava_next still fail (all-zero even without images).

tests/test_chat_template_utils.py
kashif: removed in 3c3913f, along with the template handling for that shape: consecutive assistant turns now render exactly like the original template.

trl/chat_template_utils.py
```diff
@@ -563,6 +572,75 @@ def is_chat_template_prefix_preserving(processing_class: PreTrainedTokenizerBase
...
+    that the assistant's masked span ends on an end-of-turn token rather than on content.
+
+    The template must define `{% generation %}` / `{% endgeneration %}` markers (see [`get_training_chat_template`]),
+    otherwise the assistant mask is empty and this returns `False`.
+
+    Args:
+        processing_class (`PreTrainedTokenizerBase` or `ProcessorMixin`):
+            Tokenizer or processor instance to check.
+        chat_template (`str`, *optional*):
+            Chat template to check. Defaults to the one attached to `processing_class`.
```
qgallouedec, line 594: wouldn't it be easier to just access `tokenizer.chat_template`? this would be aligned with `is_chat_template_prefix_preserving`

qgallouedec: Ah, actually you can't, because in sft you want to test a different chat template.

tests/test_chat_template_utils.py
```diff
@@ -628,13 +676,26 @@ def _load(self, model_name):
...
+    def test_new_chat_template_trains_stop_token(self, tokenizer_name, request):
+        if tokenizer_name in (
+            "trl-internal-testing/tiny-LlavaForConditionalGeneration",
+            "trl-internal-testing/tiny-LlavaNextForConditionalGeneration",
+        ):
+            reason = f"{tokenizer_name}: the processor returns an all-zero assistant tokens mask"
+            request.node.add_marker(pytest.mark.xfail(strict=True, reason=reason))
+        tokenizer = self._load(tokenizer_name)
+        chat_template = get_training_chat_template(tokenizer)
+        assert is_chat_template_stop_token_trained(tokenizer, chat_template=chat_template) is True
```
qgallouedec:
```suggestion
        new_chat_template = get_training_chat_template(tokenizer)
        assert is_chat_template_stop_token_trained(tokenizer, chat_template=new_chat_template) is True
```

nit, for consistency

tests/test_chat_template_utils.py
```diff
@@ -628,13 +676,26 @@ def _load(self, model_name):
 
     def test_new_chat_template_is_prefix_preserving(self, tokenizer_name):
         tokenizer = self._load(tokenizer_name)
-        tokenizer.chat_template = get_training_chat_template(tokenizer)
+        chat_template = get_training_chat_template(tokenizer)
+        if chat_template is not None:
+            tokenizer.chat_template = chat_template
```
qgallouedec:
```suggestion
        new_chat_template = get_training_chat_template(tokenizer)
        if chat_template is not None:
            tokenizer.chat_template = new_chat_template
```

tests/test_chat_template_utils.py
```diff
@@ -628,13 +676,26 @@ def _load(self, model_name):
...
             pytest.skip("Template does not support tool calling; prefix-preservation check is not applicable.")
         assert is_chat_template_prefix_preserving(tokenizer) is True
 
+    def test_new_chat_template_trains_stop_token(self, tokenizer_name, request):
+        if tokenizer_name in (
+            "trl-internal-testing/tiny-LlavaForConditionalGeneration",
+            "trl-internal-testing/tiny-LlavaNextForConditionalGeneration",
+        ):
+            reason = f"{tokenizer_name}: the processor returns an all-zero assistant tokens mask"
+            request.node.add_marker(pytest.mark.xfail(strict=True, reason=reason))
```
qgallouedec:
```suggestion
            request.node.add_marker(pytest.mark.xfail(strict=False, reason=reason))
```

nit: I think here we want strict=False, when they fix it in transformers, we don't want the fix to break the CI, just to be reported with a xpass

qgallouedec (commented):
thanks, just a few suggestion. For ref, there is an open pr for the vlm masking issue: https://github.com/huggingface/transformers/issues/44521

trl/chat_template_utils.py
kashif: right, that is the reason: sft resolves a training template and we want to check it without mutating the tokenizer. Also fixed the leftover rename from the suggestion in the prefix test (the guard still read the old variable name).

qgallouedec (approved):

kashif:
thanks for iterating this is great!

PR #5989, opened 2026-06-09
SDFT/SDPO: live teacher logprobs from the vLLM server

kashif:
Follow up to #5883.

In the self distillation trainers the teacher is the student itself, and with `teacher_model_kind="live"` the vLLM generation server already holds the current student weights (they are synced every step for rollouts). This PR adds `use_teacher_server=True` which scores the teacher log probabilities on that same server instead of running a separate local teacher forward, removing the teacher from the training step entirely.

Supported scope: `teacher_model_kind="live"` with `vllm_mode="server"` and `distillation_weight=1.0`. `distillation_mode` can be `sampled_token` (reverse KL on the realized token) or `topk_logits`, where the server returns the teacher's own top k so the loss distills over the teacher's support (the same approach as the distillation trainer's server path). When buffered batches are reused across optimizer steps (`num_iterations > 1`) the weights are re synced before scoring so the teacher never scores stale.

Tested with the unit suites plus distributed smoke runs (zero3 and fsdp2, both modes, PEFT, and `num_iterations > 1` where the server log shows the extra sync session firing).

Thanks @LeonEricsson for the reviews on the fork side ([LeonEricsson#5](https://github.com/LeonEricsson/trl/pull/5)).

---

> [!NOTE]
> **Medium Risk**
> Touches training loss paths and distributed vLLM weight sync; incorrect masking or stale server weights could affect gradients, though behavior mirrors the existing distillation server path and is covered by new edge-case tests.
> 
> **Overview**
> Adds **`use_teacher_server`** to **SDFT** and **SDPO** so that with `teacher_model_kind="live"`, vLLM **server** mode, and pure distillation, teacher log-probs come from the same vLLM server used for rollouts (via `VLLMClient.get_sequence_logprobs`) instead of a local teacher forward—student logits still run on the training process.
> 
> `compute_loss` branches into `_compute_server_distillation_loss` for `sampled_token` or `topk_logits` (not `full_logits` / Liger); weights are re-synced to the server before scoring when buffered batches reuse steps (`num_iterations > 1`), with a cross-rank barrier. **SDPO** additionally requires `distillation_weight=1.0`. Config validation now enforces `distillation_alpha=1.0` when `distillation_mode="sampled_token"`.
> 
> Docs cover setup and constraints; unit tests assert finite loss/gradients under fully masked or ragged padded server responses.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a15ec5de0b222c5d416fefccbace618bedbe5db1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/sdpo/sdpo_trainer.py
```diff
@@ -1188,6 +1261,138 @@ def _compute_self_distillation_loss(
...
+            model=model,
+            input_ids=student_input_ids,
+            attention_mask=student_attention_mask,
+            logits_to_keep=logits_to_keep,
+        )
+
+        mode = "train" if model.training else "eval"
+        if loss_mask.sum() == 0:
+            self._log_self_distillation_metric(mode, 0.0)
+            return student_logits.sum() * 0.0
```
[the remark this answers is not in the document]
kashif: right, fixed. dropped the early return so every rank runs the full path and joins the gathers, an all masked rank just flows through with a zero loss (the masked reductions and clamps keep everything finite). note the local _compute_self_distillation_loss from #5883 has the same early return before its gather, probably worth a follow up there too.

trl/experimental/sdft/sdft_trainer.py
```diff
@@ -846,6 +917,138 @@ def _compute_self_distillation_loss(
...
+            model=model,
+            input_ids=student_input_ids,
+            attention_mask=student_attention_mask,
+            logits_to_keep=logits_to_keep,
+        )
+
+        mode = "train" if model.training else "eval"
+        if loss_mask.sum() == 0:
+            self._log_self_distillation_metric(mode, 0.0)
+            return student_logits.sum() * 0.0
```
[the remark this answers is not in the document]
kashif: right, fixed. dropped the early return so every rank runs the full path and joins the gathers, an all masked rank just flows through with a zero loss (the masked reductions and clamps keep everything finite). note the local _compute_self_distillation_loss from #5883 has the same early return before its gather, probably worth a follow up there too.

trl/experimental/sdft/sdft_trainer.py
```diff
@@ -846,6 +917,144 @@ def _compute_self_distillation_loss(
...
+        student_logits = self._forward_logits(
+            model=model,
+            input_ids=student_input_ids,
+            attention_mask=student_attention_mask,
+            logits_to_keep=logits_to_keep,
+        )
+
+        mode = "train" if model.training else "eval"
+        # No early return for an all-masked rank: the gathers below are collective, so every rank must run the full
+        # path (the masked reductions resolve to a zero loss on such a rank).
```
LeonEricsson: the code does not have an early return anymore, so explaining its absence is a bit confusing. 

also the placement of this comment is quite unintuitive in the context of the function.

trl/experimental/sdft/sdft_trainer.py
```diff
@@ -846,6 +917,144 @@ def _compute_self_distillation_loss(
...
+
+        teacher_per_token_logps = teacher["actual_logprobs"]
+        if (required & ~torch.isfinite(teacher_per_token_logps)).any():
+            raise ValueError("Teacher server returned no logprob for a required completion token.")
+        teacher_per_token_logps = torch.where(
+            required, teacher_per_token_logps, torch.zeros_like(teacher_per_token_logps)
+        )
+
+        # Mean realized-token logprob gap between the local student and the context-conditioned teacher on the server.
+        # It shrinks as distillation progresses; a sudden jump points at a weight-sync or numerics problem.
```
LeonEricsson: this feels a bit too "predictive", is monotonic shrinkage guaranteed? how about:

```suggestion
        # Diagnostic for disagreement between local student scores and server teacher scores on realized tokens.
        # Sudden jumps can indicate stale server weights or numerical drift.
        
```

trl/experimental/sdft/sdft_trainer.py
```diff
@@ -846,6 +917,144 @@ def _compute_self_distillation_loss(
...
+
+        The student is forwarded locally (grad) for its logits; the teacher logprobs are fetched from the server (no
+        local teacher forward). `sampled_token` distills the realized token (reverse KL); `topk_logits` distills over
+        the teacher's own top-k support (the server cannot score the student's top-k indices).
+        """
+        # Buffered batches are reused across optimizer steps (`num_iterations > 1`), so the server weights can lag the
+        # live student; re-sync before scoring.
+        if self.state.global_step != self._last_loaded_step:
+            self.vllm_generation.sync_weights()
+            self._last_loaded_step = self.state.global_step
```
LeonEricsson, line 934: Weight syncing to the vLLM server is only performed by the main rank. Is there a risk that non-main ranks start teacher scoring before the server update has completed?

In server mode, non-main ranks can return from sync_weights() early because they don't sync, then proceed to get_sequence_logprobs() while rank 0 is still pushing parameters. This isn't a problem for the other weight sync (during generation) because only the main rank calls the vLLM server after syncing and then broadcasts the results to other ranks.

tests/experimental/test_sdft_trainer.py
```diff
@@ -60,6 +63,10 @@ def _assert_any_trainable_param_changed(model, previous_trainable_params):
             for name, previous_param in previous_trainable_params.items()
         )
 
+    def test_config_sampled_token_requires_reverse_kl_alpha(self):
+        with pytest.raises(ValueError, match="distillation_alpha=1.0"):
+            SDFTConfig(output_dir=self.tmp_dir, distillation_mode="sampled_token", distillation_alpha=0.5)
```
LeonEricsson:
```suggestion

```

tests/experimental/test_sdft_trainer.py
```diff
@@ -439,3 +446,68 @@ def test_train_reuses_buffered_generation_batches(self):
...
+        student_sparse = add_tail_bucket(student_topk)
+        teacher_sparse = add_tail_bucket(teacher_topk)
+        assert torch.isfinite(student_sparse).all()
+
+        per_token_loss = compute_divergence(student_sparse, teacher_sparse, 0.5)
+        loss = (per_token_loss * loss_mask).sum() / loss_mask.sum()
+        assert torch.isfinite(loss)
+
+        loss.backward()
+        assert torch.isfinite(raw_student.grad).all()
```
LeonEricsson: this doesn't test any actual trainer behavior, its testing a hand written copy of the implementation. will it provide any regression coverage for SDFTrainer?

trl/experimental/sdft/sdft_trainer.py
kashif: fair, dropped the comment.

trl/experimental/sdft/sdft_trainer.py
kashif: applied, and mirrored the same wording to sdpo.

trl/experimental/sdft/sdft_trainer.py
kashif: good catch, added accelerator.wait_for_everyone right after the sync so every rank waits for the push to finish before scoring.

tests/experimental/test_sdft_trainer.py
kashif: applied, removed the sdpo one too.

tests/experimental/test_sdft_trainer.py
kashif: right, replaced these with a test that drives the path through compute_loss on a real trainer with a fake teacher client, so the masking and reductions under test are the real ones and nothing private is called.

LeonEricsson (approved):

PR #6002, opened 2026-06-11
Support multiple environments [2/2]: Per-example environment selection

qgallouedec:
Builds on PR #6001. Adds the actual feature; no behavior change for single-callable usage.

`environment_factory` now also accepts a `dict[str, factory]`. Each dataset example selects its environment via an `environment` field, and **only that environment's tools are exposed in the example's prompt**, so a coding task and a game can be trained together in one run without leaking each other's tools.

A single callable keeps working: it's the special case of one unnamed environment shared by every example. Both go through one code path — environment instances are pooled and reused (reset between steps), and the per-rollout tool dicts + per-example tool schema are built per batch, right before generation.

Same change applied to GRPO and async-GRPO. Adds a "Multiple environments" docs section and tests (multi-env GRPO train, async loop wiring, single-factory behavior unchanged).

---

> [!NOTE]
> **Medium Risk**
> Changes rollout generation, tool dispatch, and prompt tokenization across GRPO-family trainers; behavior for existing single-factory users should be equivalent but mixed-env batches add new data-contract requirements (`environment` must match factory keys).
> 
> **Overview**
> **`environment_factory`** can now be a **`dict[str, factory]`** in addition to a single callable. Dataset rows pick an environment with an **`environment`** column; each prompt is tokenized with **only that environment’s tools** in the chat template, so mixed tasks (e.g. coding vs. game) can train in one run without cross-leaking tool schemas.
> 
> Single-callable usage is unchanged (normalized to one shared unnamed environment). Implementation refactors pooling to **`_environment_pool` / `_env_tools` per name**, validates unknown `environment` values with a clear **`ValueError`**, and strips **`environment`** from kwargs passed to **`reset`**. The same wiring is applied to **GRPO**, **async GRPO** (rollout loop), **DPPO**, and **GRPO with replay buffer**, including per-example tool lists for VLM prompt processing.
> 
> Docs add a **“Multiple environments”** section; tests cover multi-env GRPO training, async rollout wiring, and unknown-environment errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c6a4cd04d69ed8410c08ef38a0cf6149cc6922da. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -319,13 +337,17 @@ async def _generate_loop(self, stop_event: asyncio.Event) -> None:
                 self._heartbeat_value.value = time.time()
                 while free_slots and not stop_event.is_set():
                     group_id, row = next(work_iter)
+                    # The environment is the same for all generations of a group; only its tools are exposed in the
+                    # example's prompt. When there are no environments, every example shares the standalone tools.
+                    name = row["environment"] if self._multi_environment else None
```
adithya-s-k: No check that `environment` maps to a real factory — a missing or typo'd value is a bare `KeyError` mid-rollout (here, and at `self._env_tools[name]`). Worth validating the dataset's `environment` values against `environment_factories.keys()` up front. Same on the GRPO side (`grpo_trainer.py:1895`).

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -350,13 +372,16 @@ async def _generate_loop(self, stop_event: asyncio.Event) -> None:
...
-                        environment = (
-                            self._environment_pool.pop() if self._environment_pool else self.environment_factory()
+                    if self.environment_factories is not None:
+                        pool = self._environment_pool[name]
+                        environment = pool.pop() if pool else self.environment_factories[name]()
+                        reset_kwargs = (
+                            {k: v for k, v in row.items() if k != "environment"} if self._multi_environment else row
                         )
-                        environment.reset(**row)
+                        environment.reset(**reset_kwargs)
```
adithya-s-k: Async drops `reset()`'s return here and generates from `row["prompt"]` as-is, but GRPO folds it into the prompt (`grpo_trainer.py:1939`: `prompt[-1]["content"] += observation`). So an environment that hands the task back through `reset()` (empty dataset prompt + instruction) trains blind on async but not on GRPO. Intentional split (tracked in #6067), or should async mirror GRPO here so the two stay consistent?

adithya-s-k: Non-blocking follow-up: `reset()` runs unguarded in `_generate_loop`, so one bad sandbox raises out of the loop and takes down the whole worker (heartbeat-stale). Pre-existing, but more environment types = more surface — a per-rollout try/except that drops just that rollout would harden long mixed-environment runs.

adithya-s-k (commented):
Nice — collapsing the single callable into `{None: factory}` so both modes share one path is clean, and doing it the same way for GRPO + async is the right call. Tool isolation checks out (per-example `tools` flow into tokenization at `async_rollout_worker.py:543`, and the tool-response delta is schema-independent so it's unaffected). A few inline notes.

On tests: happy path is well covered — worth adding an error case (example missing / unknown `environment`) and a multi-turn multi-env rollout, to pin that the per-example tools stay consistent across tool-response turns.

trl/experimental/async_grpo/async_rollout_worker.py
qgallouedec: ok, done in https://github.com/huggingface/trl/pull/6002/commits/d500220987340814ee53f7da659291beb8b747ab

```python
#repro.py
import os

os.environ["TRL_EXPERIMENTAL_SILENCE"] = "1"

class CounterEnvironment:
    def reset(self, **kwargs): ...

    def increment(self, step: int) -> int:
        """Increment the counter.

        Args:
            step: Value to add.

        Returns:
            The updated value.
        """
        return step


class EchoEnvironment:
    def reset(self, **kwargs): ...

    def shout(self, text: str) -> str:
        """Shout the text.

        Args:
            text: Text to shout.

        Returns:
            The text in upper case.
        """
        return text.upper()


dataset = load_dataset("trl-internal-testing/zen", "conversational_prompt_only", split="train")

trainer = GRPOTrainer(
    model="trl-internal-testing/tiny-Qwen3MoeForCausalLM",
    args=GRPOConfig(
        output_dir="/tmp/grpo-guard-demo",
        num_generations=2,
        report_to="none",
    ),
    train_dataset=dataset,
    environment_factory={"counter": CounterEnvironment, "echo": EchoEnvironment},
)

bad_inputs = [dict(x, environment="calulator") for x in dataset.select(range(4))]
trainer._generate_and_score_completions(bad_inputs)
```

before

```
$ python/fsx/qgallouedec/trl/repro.py
  File "/fsx/qgallouedec/trl/repro.py", line 58, in <module>
    trainer._generate_and_score_completions(bad_inputs)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 2037, in _generate_and_score_completions
    pool = self._environment_pool[name]
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'calulator'
```

after

```
$ python/fsx/qgallouedec/trl/repro.py
...
  File "/fsx/qgallouedec/trl/repro.py", line 64, in <module>
  File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 2037, in _generate_and_score_completions
    raise ValueError(
    ...<2 lines>...
    )
ValueError: Example has `environment='calulator'`, which is not among the environments passed to `environment_factory`. Expected one of: ['counter', 'echo'].
```

trl/experimental/async_grpo/async_rollout_worker.py
qgallouedec: Yes the split is intentional for now.
The async worker discarding reset()'s return is pre-existing (single-env #6001) and independent of this PR's per-example selection. Mirroring GRPO means moving the reset ahead of tokenization and making the prompt per-rollout instead of per-group, so I'd rather keep it out of scope here and track it in #6067.

trl/trainer/grpo_trainer.py
```diff
@@ -1450,6 +1473,34 @@ def _tokenize_prompts(self, prompts: list):
...
+                    for k, v in tokenized.items():
+                        if k not in ("input_ids", "attention_mask"):
+                            multimodal_fields.setdefault(k, []).append(v)
+                # Merge the per-prompt fields so the result matches the single batched `apply_chat_template` call used
+                # when there are no environments: image fields (pixel_values, image_grid_thw, ...) are flattened over
+                # patches/images and concatenated; per-token fields (e.g. token_type_ids) stay batched per prompt.
+                multimodal_fields = {
+                    k: torch.cat(v) if isinstance(v[0], torch.Tensor) else [row for prompt_v in v for row in prompt_v]
+                    for k, v in multimodal_fields.items()
+                }
```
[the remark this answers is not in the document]
qgallouedec: The per-example path is structurally equivalent to the batched call here. The merge only `torch.cat`s image-**indexed fields** (`pixel_values`, `image_grid_thw`, where concatenating over images is correct) while per-token fields (`token_type_ids`, `mm_token_type_ids`) come back as lists and are kept per-prompt via the else-branch, exactly as the downstream forward (`_get_per_token_logps_and_entropies`) expects.

I verified this against Gemma3 and Qwen2VL with mixed-length prompts: the merged fields match the single batched apply_chat_template call byte-for-byte, e.g.

```python
import torch
from PIL import Image

IMG = Image.new("RGB", (64, 64), (128, 30, 200))
def m(t): return [{"role":"user","content":[{"type":"image","image":IMG},{"type":"text","text":t}]}]
prompts = [m("hi"), m("a considerably longer prompt with more text tokens here")]

MODELS = [
    "tiny-Gemma3ForConditionalGeneration", "tiny-Idefics2ForConditionalGeneration",
    "tiny-Idefics3ForConditionalGeneration", "tiny-LlavaForConditionalGeneration",
    "tiny-LlavaNextForConditionalGeneration", "tiny-Qwen2VLForConditionalGeneration",
    "tiny-Qwen2_5_VLForConditionalGeneration", "tiny-Qwen3VLForConditionalGeneration",
    "tiny-SmolVLMForConditionalGeneration",
]
def desc(x):
    if x is None: return "MISSING"
    if isinstance(x, torch.Tensor): return f"tensor{tuple(x.shape)}"
    return f"list[{len(x)}]"

for name in MODELS:
    print(f"\n===== {name} =====")
    try:
        proc = AutoProcessor.from_pretrained(f"trl-internal-testing/{name}", trust_remote_code=True)
        batched = proc.apply_chat_template(conversation=prompts, add_generation_prompt=True, tokenize=True, return_dict=True)
        mm = {}
        for p in prompts:
            d = proc.apply_chat_template(conversation=[p], add_generation_prompt=True, tokenize=True, return_dict=True)
            for k, v in d.items():
                if k not in ("input_ids", "attention_mask"):
                    mm.setdefault(k, []).append(v)
        merged = {k: (torch.cat(v) if isinstance(v[0], torch.Tensor) else [r for pv in v for r in pv]) for k, v in mm.items()}
        for k in sorted(merged):
            ok = "SAME" if desc(batched.get(k)) == desc(merged.get(k)) else "DIFFER <<<"
            print(f"  {k:20s} {desc(batched.get(k)):18s} vs {desc(merged.get(k)):18s} {ok}")
    except Exception as e:
        print(f"  (skip: {type(e).__name__}: {str(e)[:80]})")
```

```
===== tiny-Gemma3ForConditionalGeneration =====
  pixel_values         list[2]            vs list[2]            SAME
  token_type_ids       list[2]            vs list[2]            SAME
===== tiny-Idefics2ForConditionalGeneration =====
  pixel_attention_mask tensor(2, 5, 378, 378) vs tensor(2, 5, 378, 378) SAME
  pixel_values         tensor(2, 5, 3, 378, 378) vs tensor(2, 5, 3, 378, 378) SAME
===== tiny-Idefics3ForConditionalGeneration =====
  pixel_attention_mask list[2]            vs list[2]            SAME
  pixel_values         list[2]            vs list[2]            SAME
===== tiny-LlavaForConditionalGeneration =====
  pixel_values         list[2]            vs list[2]            SAME
===== tiny-LlavaNextForConditionalGeneration =====
  image_sizes          list[2]            vs list[2]            SAME
  pixel_values         list[2]            vs list[2]            SAME
===== tiny-Qwen2VLForConditionalGeneration =====
  image_grid_thw       tensor(2, 3)       vs tensor(2, 3)       SAME
  mm_token_type_ids    list[2]            vs list[2]            SAME
  pixel_values         tensor(32, 1176)   vs tensor(32, 1176)   SAME
===== tiny-Qwen2_5_VLForConditionalGeneration =====
  image_grid_thw       tensor(2, 3)       vs tensor(2, 3)       SAME
  mm_token_type_ids    list[2]            vs list[2]            SAME
  pixel_values         tensor(32, 1176)   vs tensor(32, 1176)   SAME
===== tiny-Qwen3VLForConditionalGeneration =====
  image_grid_thw       tensor(2, 3)       vs tensor(2, 3)       SAME
  mm_token_type_ids    list[2]            vs list[2]            SAME
  pixel_values         tensor(512, 1536)  vs tensor(512, 1536)  SAME
===== tiny-SmolVLMForConditionalGeneration =====
```

The earlier `v[0]`-slice concern was on a prior revision; the current merge no longer slices. So no mismatch in practice. The only residual is that the tensor-vs-list check is an implicit proxy for image- vs sample-indexed fields, which holds for every processor tested.

trl/experimental/async_grpo/async_rollout_worker.py
qgallouedec: Agree it's worth noting, but I'd keep `reset()` unguarded: a failing reset is an environment contract violation, and catching it would silently drop rollouts (and diverge from the sync GRPO path, which also resets unguarded). Retry/robustness belongs in the environment. Leaving as-is.

trl/experimental/dppo/dppo_trainer.py
```diff
@@ -851,14 +851,27 @@ def _generate_and_score_completions(
 
         prompts = [x["prompt"] for x in inputs]
 
-        # Draw one reusable instance per rollout from the pool, creating more only when this batch needs more concurrent
-        # instances than exist.
-        if self.environment_factory is not None:
+        # Resolve each example's environment and draw one reusable instance per rollout from the pool, creating more
```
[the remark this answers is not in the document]
qgallouedec: Again, that's why I think we should avoid inheritance as much as possible for trainers in this repo.

Anyways it's safe to ignore, these two trainers aren't used (as per our telemetry), we will likely remove them soon)

trl/experimental/grpo_with_replay_buffer/grpo_with_replay_buffer_trainer.py
```diff
@@ -71,14 +71,27 @@ def _generate_and_score_completions(
...
-                    self._environment_pool.append(self.environment_factory())
-                self.environments.append(self._environment_pool[i])
+            pool_cursor = {}  # how many instances of each environment have been handed out so far this batch
+            for name in self._batch_environments:
+                pool = self._environment_pool[name]
+                index = pool_cursor.get(name, 0)
+                if index == len(pool):
+                    pool.append(self.environment_factories[name]())
+                pool_cursor[name] = index + 1
+                self.environments.append(pool[index])
```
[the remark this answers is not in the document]
qgallouedec: https://github.com/huggingface/trl/pull/6002#discussion_r3508610978

qgallouedec:
@adithya-s-k I'm going to merge without review/approval. It' been a while since it was opened and keeping the branch aligned with main is taking too much time. We need to move forward on the multi-env aspect.

PR #6036, opened 2026-06-12
test: don't hard-code bf16=True on devices that lack bf16 support

### LINKED ISSUE #3616: Latest default config of bf16=True breaks test cases when run on CPU.
### Reproduction

#3515 causes all test cases to break when they are run on a cpu. Test cases need to be updated to account for the underlying device and device type being used.

### System Info

Github codespaces.

### Checklist

behroozazarkhalili:
## What

Test cases that hard-code `bf16=True` raise `"Your setup doesn't support bf16/gpu"` when run on a CPU or a pre-Ampere GPU (e.g. T4), because the default `bf16=True` was introduced in #3515 and transformers' `TrainingArguments` validation rejects `bf16=True` on such devices.

## How

- Add `is_bf16_supported()` to `tests/testing_utils.py`, mirroring the exact condition `TrainingArguments` accepts for `bf16=True` (`is_torch_bf16_gpu_available() or is_torch_xla_available()`), as a sibling to the existing `is_ampere_or_newer()`.
- Use `bf16=is_bf16_supported()` at the one opportunistic GRPO site (`test_vlm_processor_vllm_colocate_mode`, where `bf16` is a memory optimization, not a flash-attention-2 requirement).
- The flash-attention-2 sites that genuinely require `bf16` (`test_train_padding_free` in SFT/DPO, `test_vlm_training` in GRPO) are already guarded by `@skipif(not is_ampere_or_newer() and torch_device != "xpu")`, so they are left unchanged.
- The invariant FA2 members (`sft_fa2`, `sft_fa2_padfree`) require `bf16` for the bf16-only FA2 kernels, so they now skip on a non-bf16 device instead of erroring.

This follows the approach suggested in the issue (a small `is_bf16_supported()` helper threaded through the affected tests).

## Verification

- On CPU, `is_bf16_supported()` returns `False`, matching the device condition `TrainingArguments` validates.
- The invariant FA2 members previously errored on CPU (`bf16=True`); they now skip with a clear reason.
- `ruff check` and `ruff format --check` pass on the changed files.

Resolves #3616

---

> [!NOTE]
> **Low Risk**
> Test-only guards and one config flag change; no production training or library behavior is modified.
> 
> **Overview**
> Fixes CI failures on CPU and pre-Ampere GPUs where `TrainingArguments` rejects `bf16=True` with "Your setup doesn't support bf16/gpu".
> 
> Adds **`is_bf16_supported()`** in `tests/testing_utils.py`, aligned with transformers' hardware check (`is_torch_bf16_gpu_available()` or XLA). **Invariant** FA2 configs (`sft_fa2`, `sft_fa2_padfree`) that need bf16-only kernels now **skip** when the helper is false instead of erroring. **`test_vlm_processor_vllm_colocate_mode`** sets **`bf16=False`** because it only constructs the trainer (no `train()`), so bf16 was unnecessary and triggered validation on unsupported devices.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 808a13354b456ed077a796d9dc74a66154618527. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_grpo_trainer.py
```diff
@@ -3195,7 +3196,7 @@ def test_vlm_processor_vllm_colocate_mode(self):
             use_vllm=True,  # Enable vLLM
             vllm_mode="colocate",  # Use colocate mode to avoid server dependency
             vllm_gpu_memory_utilization=0.05,  # Use minimal GPU memory (5%)
-            bf16=True,  # Use bfloat16 to reduce memory
+            bf16=is_bf16_supported(),  # bfloat16 to reduce memory, when the device supports it
```
qgallouedec: wouldn't it be easier to just remove bf16=True?

behroozazarkhalili:
@qgallouedec 

Good call, you're right that `is_bf16_supported()` is overkill here. I pushed a simpler version.

One detail on why I went with `bf16=False` rather than just deleting the line: `bf16` defaults to `None` and is resolved in `_BaseConfig.__post_init__` as

```python
self.bf16 = not (self.fp16) if self.bf16 is None else self.bf16
```

`fp16` is unset in this test, so dropping the line entirely resolves `bf16` back to `True` (`not False`), which is the original value. `bf16=False` makes it explicit and device-safe instead.

It doesn't actually change what this test exercises either way: `test_vlm_processor_vllm_colocate_mode` is construction-only (it never calls `train()`), and the only place `args.bf16` is read is the `use_transformers_paged` branch, which this `use_vllm=True` test doesn't take. The 4-bit weights are already loaded with `bnb_4bit_compute_dtype=torch.bfloat16` regardless, so the flag has no effect on the colocate path. `bf16=False` just avoids the hard-coded-`True`-on-an-unsupported-device footgun the PR is about.

I kept `is_bf16_supported()` for `tests/invariant/test_invariant.py`, where it gates a real skip for the FA2/bf16 configs.

tests/test_grpo_trainer.py
behroozazarkhalili: Done in 2a2d3b51 — went with explicit `bf16=False` rather than deleting the line, since `bf16` defaults to `None` and `_BaseConfig.__post_init__` resolves it to `not fp16` = `True`, so a bare delete would keep it `True`. (`args.bf16` is unused on this construction-only vLLM path anyway.)

tests/test_grpo_trainer.py
albertvillanova: Stale docstring `bfloat16 precision`: the bf16 compute dtype is already implied by the "4-bit quantization with BitsAndBytesConfig" bullet, and the trainer no longer runs in bf16
```suggestion
```

albertvillanova (approved):
Thanks for the clean, well-scoped fix here! 🤗 

One tiny comment: the `test_vlm_processor_vllm_colocate_mode` docstring still lists "bfloat16 precision" as a memory optimization, which now reads a bit oddly next to `bf16=False`.

Also, it would be nice to merge the upstream main branch into this PR to be sure everything continues green.

PR #6025, opened 2026-06-12
fix(sft): drop examples with no trainable tokens after max_length truncation

0xadvait:
# What does this PR do?

`assistant_only_loss=True` (or completion-only training) combined with `max_length` truncation can silently train on nothing.

The all-zero `assistant_masks` check in `tokenize_fn` runs **before** truncation, but truncation happens later, in `DataCollatorForLanguageModeling`. An example whose assistant/completion tokens all lie beyond `max_length` (e.g. the prompt alone reaches `max_length`) passes the check, then gets every label set to -100 by the collator. Since fully-masked micro-batches return a graph-connected zero loss (for DDP/FSDP safety), a batch made of such examples trains with no error and no warning, on current main:

```
{'loss': 0.0, 'grad_norm': nan, 'mean_token_accuracy': 0.0, 'num_tokens': 256.0, ...}
{'loss': 0.0, 'grad_norm': nan, 'mean_token_accuracy': 0.0, 'num_tokens': 512.0, ...}
```

(repro: 8 conversational examples with a ~200-token user turn and a short assistant reply, `max_length=64`, `assistant_only_loss=True`; `num_tokens` keeps counting while nothing is learned)

When only some examples are affected, they silently dilute training; the failure surfaces, if ever, as a confusing downstream error.

## The fix

This implements the approach discussed in #3930 -- drop the affected examples at dataset preparation time and log how many were dropped -- without the dataset-wide trainable-token counting that stalled that PR:

- after tokenization (and also for already-tokenized datasets), when `max_length` is set and packing is disabled, a batched `filter` keeps only examples with at least one trainable token within the truncation boundary
- a token counts as trainable when every applicable mask is 1 at its position: `assistant_masks` is always applied by the collator when present, while `completion_mask` is only applied when `completion_only_loss` resolves to `True` -- the filter mirrors exactly that, so examples are never dropped based on a mask the collator won't use
- both `truncation_mode="keep_start"` and `"keep_end"` are handled with the same slices the collator uses
- if some examples are dropped: a warning with the dropped count and the relevant settings
- if **all** examples are dropped: an actionable `ValueError` (previously: silent zero-loss training, or a cryptic downstream failure)
- for `IterableDataset`, the same filter applies lazily; the counting/warning applies to map-style datasets where `len()` is available

No new dependencies, and a single extra batched `filter` pass (`num_proc`-parallel) that only runs when a mask column is present and truncation is active.

## Tests

[...]

PR #6037, opened 2026-06-12
refactor(sft): build labels during dataset preparation instead of collation

0xadvait:
# What does this PR do?

Part 1 of the plan discussed in #6025 (toward fixing #3927): move label construction out of the collator and into dataset preparation, so that "what's trainable" is defined in exactly one place.

## Changes

- **Dataset preparation builds the labels.** After tokenization, a single batched map builds a `labels` column: a token keeps its ID when every applicable mask is 1 at its position, else `-100`. "Applicable" mirrors the previous collator rules exactly: `assistant_masks` always applies when present; `completion_mask` only when `completion_only_loss` resolves to `True`. The step is unified, so it also covers **pre-tokenized datasets** that carry mask columns but no labels (they previously relied on the collator's mask path).
- **Datasets that already provide `labels` are taken as is**, even when mask columns are also present.
- **Datasets with no applicable mask don't get a labels column** (plain LM stays storage-neutral); the collator keeps defaulting to the input IDs.
- **The collator no longer consumes the mask columns at all** (per review): it takes `labels` as provided (defaulting to `input_ids`), truncates and pads them, and that's it. The now-unused `completion_only_loss` field is removed from the collator; the `SFTConfig` option is untouched and drives label construction at preparation time. The structural padding-free masking (`position_ids == 0 -> -100`, document boundaries that only exist after flattening) stays in the collator.
- **`skip_prepare_dataset` guard**: a dataset that skips preparation while carrying mask columns without labels would now silently train on the full sequence; when using the default collator, this is rejected with an actionable error (provide `labels` or drop `skip_prepare_dataset`).
- **Packing carries `labels` instead of the mask columns** (the masks are baked into the labels by that point).
- **Liger column selection** (`collator_expected_keys`) includes `labels`.
- **Shared truncation window**: the `keep_start`/`keep_end` slice is extracted into a module-level `_truncation_slice` helper, so the part-2 filter can use the exact same window as the collator and can't drift.
- **VLM path untouched**, as suggested: vision examples aren't tokenized at preparation time, so the vision collators keep building their own labels.

With this in place, the part-2 drop-and-warn filter (#6025) reduces to checking `labels[truncation_slice]` for any non `-100` entry, with no mask-precedence logic.

## Tests

- Collator: prebuilt labels used as is (mask columns ignored); truncated with the same window as input IDs and padded with `-100`; padding-free with prebuilt labels (structural first-of-document masking preserved). The previous mask-based collator tests are converted to prebuilt-labels equivalents.
- Preparation: labels built correctly for assistant-only (conversational) and completion-only (prompt-completion) datasets, cross-checked position-by-position against the mask columns; existing `labels` respected verbatim; labels built for pre-tokenized datasets with masks; no labels column for plain LM; packing carries labels (and drops the masks) with per-row length equality.
- `skip_prepare_dataset`: masks without labels raise the explicit error; labels-only datasets pass through untouched.

## Before submitting

- [x] Was this discussed/approved via a GitHub issue? Please add a link to it if that's the case. -> #6025 (and #3927)
- [x] Did you make sure to update the documentation with your changes? (collator docstring and examples rewritten for the labels-based contract)

## AI writing disclosure

## Who can review?

@qgallouedec (as discussed in #6025)

---

> [!NOTE]
> **Medium Risk**
> Refactors the SFT loss-mask pipeline (when tokens get -100) and packing column contract; behavior is intended to match prior mask rules but misconfigured skip-prepare datasets now fail instead of training wrong.
> 
> **Overview**
> **SFT label masking is centralized in dataset preparation** instead of being split between `_prepare_dataset` and `DataCollatorForLanguageModeling`, so trainable tokens are defined before truncation/packing (toward fixing #3927).
> 
> After tokenization, `SFTTrainer` maps a **`labels`** column when missing: token IDs stay where every applicable mask is 1 (`assistant_masks` when present; `completion_mask` only when `completion_only_loss` is on), otherwise `-100`. Existing **`labels`** are unchanged; plain LM still omits **`labels`** and the collator defaults to **`input_ids`**. **Packing** and Liger column selection keep **`input_ids` + `labels`** (masks dropped once baked in).
> 
> **`DataCollatorForLanguageModeling`** only truncates/pads provided **`labels`** (or defaults to **`input_ids`**), keeps padding-free document-boundary masking, and **no longer reads mask columns** or **`completion_only_loss`**.
> 
> With **`skip_prepare_dataset=True`**, datasets that have masks but no **`labels`** now **raise** at init/eval instead of silently training on the full sequence. Vision collators are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 641e9a97a613c5a3b7f3f0f071df7792c51e2272. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/trainer/sft_trainer.py
```diff
@@ -464,26 +476,25 @@ class DataCollatorForLanguageModeling(DataCollatorMixin):
     def torch_call(self, examples: list[dict[str, Any]]) -> dict[str, Any]:
         input_ids = [example["input_ids"] for example in examples]
         batch_seq_lengths = [example["seq_lengths"] for example in examples] if "seq_lengths" in examples[0] else None
+        has_labels = "labels" in examples[0]
```
qgallouedec: we could also remove label construction from collator (ie, constrain the input to contain labels)

0xadvait: Done in 6fa4494c. The collator no longer reads the mask columns at all: it takes `labels` as provided (defaulting to `input_ids`), truncates and pads them, and that's it. Also removed the now-unused `completion_only_loss` field from the collator; the `SFTConfig` option is untouched and drives label construction at prep time.

Two notes on where I drew the line:

- I kept the `input_ids` default rather than hard-requiring a labels column, so plain LM datasets (nothing to mask) don't need to duplicate `input_ids` in the cache. If you'd rather strictly require labels, prep can emit them unconditionally.
- The `position_ids == 0 -> -100` overlay in padding-free mode stays in the collator: that one is structural (document boundaries only exist after flattening), not content masking.

trl/trainer/sft_trainer.py
```diff
@@ -1557,6 +1526,42 @@ def tokenize_fn(example, processing_class, dataset_text_field, assistant_only_lo
...
+                                ]
+                            )
+                        return {"labels": labels}
+
+                    dataset = dataset.map(
+                        build_labels,
+                        batched=True,
+                        fn_kwargs={"mask_columns": mask_columns},
+                        **map_kwargs,
+                    )
```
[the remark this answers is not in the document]
0xadvait: Real gap, thanks. Fixed in 95a735e0: when dataset preparation is skipped and the default collator is used, a dataset carrying mask columns without labels now raises an actionable error (provide `labels` or drop `skip_prepare_dataset`) instead of silently training on the full sequence. Custom collators that consume the masks themselves are not affected by the check. The VLM half of this finding doesn't apply: vision examples aren't prepared at all and the vision collators build their own labels (and `assistant_only_loss` is already rejected for VLMs).

trl/trainer/sft_trainer.py
```diff
@@ -515,10 +499,6 @@ def torch_call(self, examples: list[dict[str, Any]]) -> dict[str, Any]:
             input_ids = [torch.cat(input_ids, dim=0)]
             labels = [torch.cat(labels, dim=0)]
             position_ids = [torch.cat(position_ids, dim=0)]
-            if completion_mask is not None:
-                completion_mask = [torch.cat(completion_mask, dim=0)]
-            if assistant_masks is not None:
-                assistant_masks = [torch.cat(assistant_masks, dim=0)]
```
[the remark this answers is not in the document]
0xadvait: Intentional: the standalone mask fallback existed in the first commit, and the maintainer review asked to remove label construction from the collator entirely (see the thread above). The PR description claiming unchanged standalone behavior was stale -- updated it to describe the labels-based contract, and the docstring/examples were rewritten in 6fa4494c.

trl/trainer/sft_trainer.py
```diff
@@ -1557,6 +1553,42 @@ def tokenize_fn(example, processing_class, dataset_text_field, assistant_only_lo
...
+                                [
+                                    token_id if all(mask[pos] for mask in masks) else -100
+                                    for pos, token_id in enumerate(input_ids)
+                                ]
+                            )
+                        return {"labels": labels}
+
+                    dataset = dataset.map(
+                        build_labels,
+                        batched=True,
```
qgallouedec: I don't think batched=True is useful here. We have a loop in build_labels

0xadvait: Right, switched to a per-example map in df3ac8c5; matches `tokenize_fn`'s style.

albertvillanova (approved):
Thanks for addressing the underlying issue: the direction is correct and the implementation is logically sound.

Do you think it would make sense to add a regression test for the original bug? There is no test demonstrating that the issue from #3927 is now surfaceable: e.g., assistant_only_loss=True, max_length shorter than the assistant response → the prepared labels contain real token IDs, but after the collator truncates them, the surviving slice is all -100. Such a test would:
- guard against regression where labels-before-truncation would be wrong

A minimal form:
```python
def test_labels_all_masked_after_truncation(self):
    """When the assistant response lies beyond max_length, the truncated labels should be all -100"""
    dataset = ...  # assistant response beyond max_length tokens
    trainer = SFTTrainer(..., args=SFTConfig(max_length=short, assistant_only_loss=True, ...))
    # labels in the full dataset contain some non-(-100) entries
    assert any(l != -100 for l in trainer.train_dataset[0]["labels"])
    # but after collation-truncation all surviving labels are -100
    batch = collator([trainer.train_dataset[0]])
    assert batch["labels"].eq(-100).all()
```

qgallouedec:
Thanks @0xadvait — direction's right and the implementation is clean. A few changes on top:

- `build_labels` now always runs, so every prepared dataset has a `labels` column. Simplifies downstream: packing selects `["input_ids", "labels"]`, Liger keys + `_signature_columns` trimmed accordingly.
- Rewrote label construction as one `zip(input_ids, *masks) comprehension (0..N masks; no mask → `labels == input_ids`).
- Trimmed the `skip_prepare_dataset` guard; inlined the single-use `_truncation_slice`.
- Tests: collator tests now pass labels in (matching the new contract), dropped redundant prebuilt-mask variants, and added @albertvillanova's regression test `test_labels_all_masked_after_truncation`, confirmed it fails on main.

qgallouedec (approved):

PR #6045, opened 2026-06-13
Remove redundant `.contiguous()` calls

qgallouedec:
Drops `.contiguous()` calls that are provably no-ops, in cases where the surrounding code already guarantees a contiguous result:

- **After advanced indexing**: integer/boolean indexing (`x[index_tensor]`) already returns a fresh contiguous tensor.
- **Before `.reshape(...)`**: `reshape` copies to a contiguous layout itself when the input is non-contiguous, so a preceding `.contiguous()` is dead.

---

> [!NOTE]
> **Low Risk**
> Layout-only cleanup with no logic changes; behavior should match prior code while slightly reducing unnecessary copies in training loops.
> 
> **Overview**
> This PR strips **redundant `.contiguous()`** calls across several training hot paths where they do not change tensor layout or memory behavior.
> 
> In **distillation** and **GOLD** Liger JSD paths, `true_labels` is built with `masked_input_ids[:, 1:].reshape(-1)` only—dropping `.contiguous()` before reshape because reshape already handles non-contiguous inputs when needed.
> 
> In **GFPO**, after group filtering via integer indexing (`group_global_indices`, `local_input_indices_to_keep`), `.contiguous()` is removed from rewards, per-token logprob tensors, and prompt/completion batches because advanced indexing already yields contiguous views.
> 
> In **`trl/trainer/utils.py`**, flattened hidden states and label targets for chunked logprob computation no longer call `.contiguous()` immediately before `.reshape()`.
> 
> **No algorithm or API changes**—only fewer redundant memory copies in loss and rollout scoring code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27cf0a4b198c80db6fbbd53a97a30e4796c7e666. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/distillation/distillation_trainer.py
+++ b/trl/experimental/distillation/distillation_trainer.py
@@ -1564,7 +1564,7 @@ class DistillationTrainer(_BaseTrainer):
 
         labels_mask = inputs["labels"] != -100
         masked_input_ids = torch.where(labels_mask, inputs["input_ids"], torch.full_like(inputs["input_ids"], -100))
-        true_labels = masked_input_ids[:, 1:].contiguous().reshape(-1)
+        true_labels = masked_input_ids[:, 1:].reshape(-1)
 
         student_head = unwrapped_student.get_output_embeddings()
         teacher_head = unwrapped_teacher.get_output_embeddings()
--- a/trl/experimental/gfpo/gfpo_trainer.py
+++ b/trl/experimental/gfpo/gfpo_trainer.py
@@ -313,8 +313,8 @@ class GFPOTrainer(_GRPOTrainer):
             group_global_indices = group_row_offsets + group_local_indices
             group_global_indices = group_global_indices.flatten()
 
-            rewards = rewards[group_global_indices].contiguous()
-            rewards_per_func = rewards_per_func[group_global_indices, :].contiguous()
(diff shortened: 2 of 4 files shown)


trl/experimental/distillation/distillation_trainer.py
```diff
@@ -1564,7 +1564,7 @@ def _compute_liger_loss(self, model, inputs, num_items_in_batch=None):
 
         labels_mask = inputs["labels"] != -100
         masked_input_ids = torch.where(labels_mask, inputs["input_ids"], torch.full_like(inputs["input_ids"], -100))
-        true_labels = masked_input_ids[:, 1:].contiguous().reshape(-1)
+        true_labels = masked_input_ids[:, 1:].reshape(-1)
```
AmineDiro, line 1567: Is this equivalent because of `reshape` ? right ?

AmineDiro (approved):

trl/experimental/distillation/distillation_trainer.py
qgallouedec: Yes, reshape does 
- .contiguous then .view if the input isn't contiguous
- just .view if the input is contiguous

PR #6054, opened 2026-06-13
fix(sft): reject transformed datasets during preparation

### LINKED ISSUE #6039: `SFTTrainer` silently breaks datasets that use `Dataset.with_transform`
### Use case

`with_transform` is the natural tool for augmentation that must re-randomize on every access rather than bake in once. Example: fine-tuning a function-calling LLM, I want it to read the tool docs in the system prompt instead of memorizing an intention --> function-name mapping. So each access renames the functions randomly but consistently within the example (schema, call, docs). Applying this once with .map() would just make the model memorize the new fixed names — it only works if every access differs. Identifier renaming, paraphrasing, noise injection all share this shape.

### Summary

If `train_dataset` has a transform set via `with_transform(...)`, there is no configuration in which `SFTTrainer` behaves correctly:

1. **`_prepare_dataset` corrupts the data.** Each internal prep step (`add_eos`, `tokenize_fn`, …) uses `Dataset.map`, and `Dataset.map` reads its input rows through the dataset's transform — so the user's augmentation fires once _per prep step_, each time on top of the previous step's already-transformed output. By the time tokenization runs, `input_ids` is frozen to this arbitrary stacked realization (in the repro below, the second application even lands _after_ the EOS token).
2. **Default config crashes at train time.** `Trainer._remove_unused_columns` drops the underlying raw column (e.g. `text`), but the transform survives → `KeyError: 'text'` on the first batch.
3. **With `remove_unused_columns=False`, augmentation is silently dead.** `text` keeps re-randomizing on access, so augmentation *looks* live — but the collator only reads `input_ids`, which is frozen. Every epoch trains on the same corrupted sample from (1).
### Reproduction — failure modes 1 and 3

```python
import tempfile, random
from datasets import Dataset

random.seed(0)
ds = Dataset.from_dict({"text": ["hello world"] * 2})

def augment(batch):
    # stand-in for real augmentation, e.g. random function renaming for tool-calling SFT
    batch["text"] = [t + f" <TAG{random.randint(0, 999)}>" for t in batch["text"]]
    return batch

ds = ds.with_transform(augment)

    trainer = SFTTrainer(
        train_dataset=ds,
        args=SFTConfig(output_dir=tmp_dir, report_to="none", use_cpu=True, bf16=False),
    )

tok = trainer.processing_class
for epoch in range(3):
    item = trainer.train_dataset[0]
    print(f"epoch {epoch}: text={item['text']!r}")
    print(f"         input_ids -> {tok.decode(item['input_ids'])!r}")
```

Output:

```
epoch 0: text='hello world <TAG911><|im_end|> <TAG654>'
         input_ids -> 'hello world <TAG911><|im_end|> <TAG41>'
epoch 1: text='hello world <TAG911><|im_end|> <TAG114>'
         input_ids -> 'hello world <TAG911><|im_end|> <TAG41>'
epoch 2: text='hello world <TAG911><|im_end|> <TAG25>'
         input_ids -> 'hello world <TAG911><|im_end|> <TAG41>'
```

**Reading the output:** `<TAG911>` got baked into stored `text` during prep's `add_eos` map, then `<|im_end|>` was appended after it. On each access, the transform only _appends_ a new tag to this stored string — so only the trailing `<TAG654>`/`<TAG114>`/`<TAG25>` changes; everything before it, including `<TAG911>`, is frozen literal data. `input_ids` was tokenized from a separate draw (`<TAG41>`) during the `tokenize_fn` map and never changes afterward, since the transform doesn't touch `input_ids`.

### Reproduction — failure mode 2 (default config crashes)

Same setup, but call `next(iter(trainer.get_train_dataloader()))` (or `trainer.train()`) instead of accessing `trainer.train_dataset[0]`:

```python
import tempfile, random
from datasets import Dataset

random.seed(0)
ds = Dataset.from_dict({"text": ["hello world"] * 2})

def augment(batch):
    batch["text"] = [t + f" <TAG{random.randint(0, 999)}>" for t in batch["text"]]
    return batch

ds = ds.with_transform(augment)

    trainer = SFTTrainer(
        train_dataset=ds,
        args=SFTConfig(output_dir=tmp_dir, report_to="none", use_cpu=True, bf16=False),
    )
```

Full traceback:

```
  File "C:\Users\my-user\Documents\trl-transform-repro\repro_crash.py", line 22, in <module>
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\.venv\Lib\site-packages\accelerate\data_loader.py", line 585, in __iter__
                    ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\.venv\Lib\site-packages\torch\utils\data\dataloader.py", line 718, in __next__
    data = self._next_data()
           ^^^^^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\.venv\Lib\site-packages\torch\utils\data\dataloader.py", line 778, in _next_data
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\.venv\Lib\site-packages\torch\utils\data\_utils\fetch.py", line 52, in fetch
    data = self.dataset.__getitems__(possibly_batched_index)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\.venv\Lib\site-packages\datasets\arrow_dataset.py", line 3162, in __getitems__
    batch = self.__getitem__(keys)
            ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\.venv\Lib\site-packages\datasets\arrow_dataset.py", line 3158, in __getitem__
    return self._getitem(key)
           ^^^^^^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\.venv\Lib\site-packages\datasets\arrow_dataset.py", line 3140, in _getitem
    formatted_output = format_table(
                       ^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\.venv\Lib\site-packages\datasets\formatting\formatting.py", line 658, in format_table
    return formatter(pa_table, query_type=query_type)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\.venv\Lib\site-packages\datasets\formatting\formatting.py", line 415, in __call__
    return self.format_batch(pa_table)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\.venv\Lib\site-packages\datasets\formatting\formatting.py", line 541, in format_batch
    return self.transform(batch)
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\my-user\Documents\trl-transform-repro\repro_crash.py", line 10, in augment
    batch["text"] = [t + f" <TAG{random.randint(0, 999)}>" for t in batch["text"]]
                                                                    ~~~~~^^^^^^^^
KeyError: 'text'
```

A transform that returns only the changed column (`return {"text": [...]}` — the other common idiom) is worse still: a custom transform's output *replaces* the row, so the prepared dataset's `__getitem__` yields no `input_ids` at all.

### Proposed fix — one of:

- **Minimal:** detect a custom transform at the top of `_prepare_dataset` and raise, pointing to `dataset_kwargs={"skip_prepare_dataset": True}`. Proceeding is never correct today.
- **Proper:** when a custom transform is detected, replace the `.map()` calls with lazy transform chaining (`user_transform → add_eos → tokenize_fn → …` composed via `with_transform`), so fresh augmentation reaches `input_ids` on every access. This requires raising on `packing=True` (packing can't be lazy) and forcing `remove_unused_columns=False` (the Arrow-level columns stay raw).

Happy to help with a PR for either approach if there's interest

### System Info

- Platform: Windows-10-10.0.26100-SP0
- Python version: 3.11.13
- TRL version: 1.7.0.dev0
- PyTorch version: 2.12.0
- accelerator(s): cpu
- Transformers version: 5.12.0
- Accelerate version: 1.14.0
- Datasets version: 5.0.0
- HF Hub version: 1.19.0
- PEFT version: not installed
- vLLM version: not installed

### Checklist

he-yufeng:
Fixes #6039.

SFTTrainer currently lets `Dataset.with_transform()` datasets enter the automatic preparation pipeline. That path calls `Dataset.map()` for EOS insertion and tokenization, and `map()` reads rows through the active custom transform. For stateful or random transforms, that can bake one transform realization into the prepared Arrow columns while later accesses still run a different transform.

This PR adds a guard before SFT dataset preparation: if the dataset has a custom transform, fail with a clear error explaining why automatic preparation is unsafe and pointing users to either `dataset_kwargs={"skip_prepare_dataset": True}` with already-tokenized examples, or materializing the transform with `Dataset.map()` before constructing the trainer.

I kept this as the conservative fix from the issue. It does not attempt lazy transform composition or packing support in this PR.

## Before submitting

- [x] Was this discussed/approved via a GitHub issue? Please add a link to it if that's the case. #6039

## AI writing disclosure

## Tests

- `python -m ruff check trl\\trainer\\sft_trainer.py tests\\test_sft_trainer.py`
- `.\\.venv\\Scripts\\python.exe -m py_compile trl\\trainer\\sft_trainer.py tests\\test_sft_trainer.py`
- `.\\.venv\\Scripts\\python.exe -m pytest tests\\test_sft_trainer.py::TestSFTTrainer::test_dataset_with_transform_requires_skip_prepare_dataset -q`

One nearby skip-prepare test was also attempted in this local Windows venv, but it fails before reaching trainer logic because the current CPU-only environment rejects the default bf16 settings. I did not change existing tests to hide that environment issue.

---

> [!NOTE]
> **Low Risk**
> Small, early validation in the dataset prep path; only blocks an unsafe pattern and does not change successful training flows.
> 
> **Overview**
> **SFTTrainer** now refuses datasets created with **`Dataset.with_transform()`** when automatic dataset preparation runs. Preparation uses **`Dataset.map()`**, which materializes rows through the lazy transform and can freeze one random or stateful augmentation into Arrow columns while later reads still apply a different transform.
> 
> At the start of **`_prepare_dataset`**, the trainer checks for a custom dataset format and raises **`ValueError`** with guidance to use **`dataset_kwargs={'skip_prepare_dataset': True}`** with trainer-ready (e.g. tokenized) examples from the transform, or to **`map()`** deterministic transforms before building the trainer.
> 
> A regression test asserts that **`SFTTrainer`** construction fails on a **`with_transform`** dataset with the expected message.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27def356a338de514eeb1b370cf45ca8df2b495a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -23,7 +23,7 @@ import torch
 import torch.nn.functional as F
 import transformers
 from accelerate.utils.memory import release_memory
-from datasets import load_dataset
+from datasets import Dataset, load_dataset
 from packaging.version import Version
 from packaging.version import parse as parse_version
 from transformers import (
@@ -1099,6 +1099,21 @@ class TestSFTTrainer(TrlTestCase):
         assert isinstance(trainer.data_collator, DataCollatorForLanguageModeling)
         assert trainer.data_collator.max_length == 16
         assert trainer.data_collator.truncation_mode == "keep_end"
+
+    def test_dataset_with_transform_requires_skip_prepare_dataset(self):
+        dataset = Dataset.from_dict({"text": ["hello world"]})
+
+        def add_suffix(batch):
+            batch["text"] = [text + " <AUG>" for text in batch["text"]]
+            return batch
+
+        dataset = dataset.with_transform(add_suffix)
+        training_args = SFTConfig(output_dir=self.tmp_dir, report_to="none")
+
+        with pytest.raises(ValueError, match=r"Dataset\.with_transform\(\).*skip_prepare_dataset"):
+            SFTTrainer(
+                model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5", args=training_args, train_dataset=dataset
+            )
 
     def test_padding_free_without_packing_and_max_length_raises(self):
         dataset = load_dataset("trl-internal-testing/zen", "standard_language_modeling", split="train[:2]")
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1110,7 +1110,10 @@ class TestSFTTrainer(TrlTestCase):
         dataset = dataset.with_transform(add_suffix)
         training_args = SFTConfig(output_dir=self.tmp_dir, report_to="none")
 
-        with pytest.raises(ValueError, match=r"Dataset\.with_transform\(\).*skip_prepare_dataset"):
+        with pytest.raises(
+            ValueError,
+            match=r"Dataset\.with_transform\(\).*skip_prepare_dataset.*trainer-ready",
+        ):
             SFTTrainer(
                 model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5", args=training_args, train_dataset=dataset
             )
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -1421,8 +1421,9 @@ class SFTTrainer(_BaseTrainer):
             raise ValueError(
                 "SFTTrainer cannot prepare a dataset that uses `Dataset.with_transform()`. The preparation pipeline "
                 "calls `Dataset.map()`, which reads through the transform and can bake a random or stateful transform "
-                "into the tokenized columns. Pass `dataset_kwargs={'skip_prepare_dataset': True}` and provide already "
-                "tokenized examples, or materialize the transform with `Dataset.map()` before constructing the trainer."
+                "into the tokenized columns. Pass `dataset_kwargs={'skip_prepare_dataset': True}` and make the "
+                "transform return trainer-ready examples, including tokenized fields, or materialize deterministic "
+                "transforms with `Dataset.map()` before constructing the trainer."
             )
 
         # If the dataset is already preprocessed (tokenized), skip the processing steps.
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -1417,6 +1417,14 @@ class SFTTrainer(_BaseTrainer):
         formatting_func: Callable[[dict], str] | None,
         dataset_name: str,
     ) -> Dataset | IterableDataset:
+        if isinstance(dataset, Dataset) and dataset.format["type"] == "custom":
+            raise ValueError(
+                "SFTTrainer cannot prepare a dataset that uses `Dataset.with_transform()`. The preparation pipeline "
+                "calls `Dataset.map()`, which reads through the transform and can bake a random or stateful transform "
+                "into the tokenized columns. Pass `dataset_kwargs={'skip_prepare_dataset': True}` and provide already "
+                "tokenized examples, or materialize the transform with `Dataset.map()` before constructing the trainer."
+            )
+
         # If the dataset is already preprocessed (tokenized), skip the processing steps.
         column_names = get_dataset_column_names(dataset)
         is_processed = "input_ids" in column_names


trl/trainer/sft_trainer.py
```diff
@@ -81,6 +81,10 @@
     from peft import PeftConfig, PeftModel, PeftType, get_peft_model
 
 
+def _dataset_has_custom_transform(dataset: Dataset | IterableDataset) -> bool:
```
albertvillanova: I would inline this function: I think it creates an unnecessary indirection for a one-liner that names a well-understood condition.

trl/trainer/sft_trainer.py
```diff
@@ -81,6 +81,10 @@
...
+    return getattr(dataset, "format", {}).get("type") == "custom"
```
albertvillanova: This guard is defensive programming and should be removed.

tests/test_sft_trainer.py
```diff
@@ -1093,6 +1093,21 @@ def test_skip_prepare_dataset_passes_truncation_to_text_collator(self):
...
 
+    def test_dataset_with_transform_requires_skip_prepare_dataset(self):
+        dataset = Dataset.from_dict({"text": ["hello world"]})
+
+        def add_suffix(batch):
+            batch["text"] = [text + " <AUG>" for text in batch["text"]]
+            return batch
+
+        dataset = dataset.with_transform(add_suffix)
+        training_args = SFTConfig(output_dir=self.tmp_dir, report_to="none", use_cpu=True, bf16=False)
```
albertvillanova: The guards for use_cpu and bf16 are not appropriate.

trl/trainer/sft_trainer.py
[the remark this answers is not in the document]
albertvillanova: There is no public API path that puts a custom transform on an IterableDataset. 

I think this would need an `isinstance(dataset, Dataset)` guard to make this explicit and close the concern permanently.

albertvillanova (changes requested):
Thanks for addressing the underlying issue. Below my concerns and suggested changes.

he-yufeng:
Addressed the review in `cee507b9`:

- inlined the custom-transform check
- limited it explicitly to `datasets.Dataset`; streaming `IterableDataset` is unaffected
- removed the defensive `getattr` / `.get` path
- removed the test's `use_cpu` and `bf16` overrides
- rebased onto current `main`

Validation:

```text
python -m ruff check trl/trainer/sft_trainer.py tests/test_sft_trainer.py
python -m ruff format --check trl/trainer/sft_trainer.py tests/test_sft_trainer.py
git diff --check
```

All passed. The focused pytest now reaches the test but this local Windows machine cannot construct the repository-default `SFTConfig` because it has no bf16-capable GPU. I left the device-specific overrides removed as requested and will rely on the repository CI environment for that test.

trl/trainer/sft_trainer.py
```diff
@@ -1417,6 +1417,14 @@ def _prepare_dataset(
...
+        if isinstance(dataset, Dataset) and dataset.format["type"] == "custom":
+            raise ValueError(
+                "SFTTrainer cannot prepare a dataset that uses `Dataset.with_transform()`. The preparation pipeline "
+                "calls `Dataset.map()`, which reads through the transform and can bake a random or stateful transform "
+                "into the tokenized columns. Pass `dataset_kwargs={'skip_prepare_dataset': True}` and provide already "
+                "tokenized examples, or materialize the transform with `Dataset.map()` before constructing the trainer."
+            )
+
         # If the dataset is already preprocessed (tokenized), skip the processing steps.
         column_names = get_dataset_column_names(dataset)
```
albertvillanova, line 1444: I think the error message is misleading: it says "provide already tokenized examples", which implies the user must pre-tokenize outside the transform. The correct pattern (transform that augments AND tokenizes) should be stated explicitly.

The second suggested workaround ("materialize the transform with Dataset.map() before constructing the trainer") is correct only for deterministic transforms; for the reported use case (random augmentation) it defeats the point and should be qualified or dropped.

albertvillanova (changes requested):
Thanks for addressing the suggested changes. Just one additional suggestion below.

Additionally, would you mind not force-pushing to a public branch? That makes the review process more difficult because it it not possible to check only the modification to the previous reviewed PR code (by reading only the specific lines have been modified), and instead forces to review the entire PR.

trl/trainer/sft_trainer.py
he-yufeng: Agreed. I updated the message in `70cc219f` so it no longer implies users must pre-tokenize outside the transform.

It now says to use `skip_prepare_dataset=True` and make the transform return trainer-ready examples, including tokenized fields. The `Dataset.map()` workaround is now qualified to deterministic transforms only.

Validated locally:

```text
python -m py_compile trl\trainer\sft_trainer.py tests\test_sft_trainer.py
python -m ruff check trl\trainer\sft_trainer.py tests\test_sft_trainer.py
python -m ruff format --check trl\trainer\sft_trainer.py tests\test_sft_trainer.py
git diff --check origin/main..HEAD
```

The focused pytest still fails before reaching this assertion on my Windows machine because the repository-default `SFTConfig` requires bf16 GPU support here. I kept the previous review request and did not reintroduce `use_cpu` / `bf16` test overrides.

albertvillanova (approved):
Thanks again for your contribution and for addressing all suggestions. It looks good to me.

Issue #4501, opened 2025-11-07
NaN metric propagation in GRPOTrainer with sparse multi-reward training


shriramc1:
## Bug Description

When using sparse multi-reward training with `logging_steps > 1`, metrics accumulate across steps. If a reward function is not active in some steps (returns `None`), it appends NaN values to the metrics list. The current implementation in `GRPOTrainer.log()` at line 1920 uses `sum(val) / len(val)` to compute the mean, which returns NaN if ANY value in the list is NaN, causing all valid metrics to be lost.

## Bug Location

https://github.com/huggingface/trl/blob/20760ba3ac7092432f67a5e3e9b2c624f168fdb0/trl/trainer/grpo_trainer.py#L1920

## Proposed Fix
```python
metrics = {
    key: torch.tensor(val).nanmean().item() if len(val) > 0 else 0.0
    for key, val in self._metrics[mode].items()
}
```

## Concrete Example

Consider a multi-task training setup with `logging_steps=4`:
- Step 1: Task A active, Task B inactive → metrics: `[0.5, NaN]`
- Step 2: Task A inactive, Task B active → metrics: `[0.5, NaN, NaN, 0.8]`
- Step 3: Both active → metrics: `[0.5, NaN, NaN, 0.8, 0.6, 0.7]`
- Step 4 (logging): `sum([0.5, NaN, NaN, 0.8, 0.6, 0.7]) / 6 = NaN` ❌

All valid reward values are lost due to NaN propagation.

## Minimal Reproducible Example

```python
from trl import GRPOTrainer, GRPOConfig
from datasets import Dataset

# Create dataset with mixed tasks
dataset = Dataset.from_dict({
    "prompt": ["task_a question", "task_b question", "task_a question", "both tasks"]
})

# Multi-reward functions that return None for non-applicable samples
def reward_func_a(prompts, completions, **kwargs):
    return [1.0 if "task_a" in p else None for p in prompts]

def reward_func_b(prompts, completions, **kwargs):
    return [0.8 if "task_b" in p else None for p in prompts]

trainer = GRPOTrainer(
    model="Qwen/Qwen2-0.5B-Instruct",
    reward_funcs=[reward_func_a, reward_func_b],
    args=GRPOConfig(
        output_dir="./output",
        logging_steps=4,  # Accumulate metrics across 4 steps
    ),
    train_dataset=dataset,
)

trainer.train()
# After 4 steps, logged metrics will be NaN even though valid values exist

### System Info

[...]

PR #6055, opened 2026-06-14
Align SDPO with GRPO/RLOO: drop NaN values when averaging logged metrics

anshulkulhari7:
# What does this PR do?

`SDPOTrainer.log()` averages the metrics accumulated over a logging window with `sum(v) / len(v)`. When a reward function is inactive for a whole batch (it returns `None` for every sample), that batch's per-function reward mean is recorded as `NaN` (via `rewards_per_func.nanmean(dim=0)` in `_record_reward_diagnostics`). With `logging_steps > 1`, a single `NaN` in the accumulation list then contaminates the average and wipes the valid values logged for the other batches.

This is the same bug class reported for `GRPOTrainer` in #4501. It was already fixed in `GRPOTrainer` and `RLOOTrainer` (and is inherited by the GRPO subclasses `gfpo`/`dppo`/`grpo_with_replay_buffer`), but the fix was never propagated to `SDPOTrainer`, which has the same reward-`None`→`nan` path and its own `log()` method. Per the consistency guideline in `AGENTS.md` ("a fix in GRPO often implies the same fix in RLOO … Not propagating a change is a bug"), this brings `SDPOTrainer.log()` in line with the GRPO/RLOO implementation: filter out `NaN` values before averaging, and only emit `None` when no valid values remain.

I audited the other reward-based trainers for the same pattern, and `sdpo` is the only remaining one affected:
- `online_dpo` aggregates rewards with `nansum`, so its logged scalars are never `NaN`.
- `a2po` / `tpo` have no `None`→`nan` reward path.
- `dpo` / `sft` / reward-model / `kto` / distillation trainers never log `NaN`-able metrics, so their `sum(v)/len(v)` is correct and is left untouched.

### Test

Added `test_log_filters_nan_from_accumulated_metrics`, which injects a `NaN` into an accumulated metric list and asserts the valid values survive the averaging (fails on `main`, passes with this change). The full `tests/experimental/test_sdpo_trainer.py` suite passes locally (13 passed, 1 skipped — the GPU-only Liger test).

Related to #4501 (the `GRPOTrainer` instance of this bug, already fixed on `main`; this propagates the same fix to `SDPOTrainer`).

## Before submitting

- [x] Was this discussed/approved via a GitHub issue? Please add a link to it if that's the case. — Related to #4501.
- [ ] Did you make sure to update the documentation with your changes? — N/A (no user-facing API or documentation change).

## AI writing disclosure

## Who can review?

@qgallouedec @pramodith — you both engaged on #4501; this propagates that fix to `SDPOTrainer`.

---

> [!NOTE]
> **Low Risk**
> Logging-only change with no training, reward, or gradient logic touched; behavior aligns with an existing fix in GRPO/RLOO.
> 
> **Overview**
> **`SDPOTrainer.log()`** now matches **GRPO/RLOO**: accumulated metrics are averaged only over **non-NaN** values, and a key becomes **`None`** when every value in the window was NaN (instead of propagating NaN across batches when `logging_steps > 1`).
> 
> This addresses the case where a reward function returns **`None` for every sample** in a batch, which records **`NaN`** for that batch’s per-reward metric via **`rewards_per_func.nanmean`**. The change adds **`import math`** and uses **`logs.update(metrics)`** with the same eval-key prefixing as the other trainers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b78e36333113bc23f6dcef319d4e6a9686c1b435. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (approved):
thanks!

Issue #6028, opened 2026-06-12
AsyncGRPOTrainer: vision-language (*ForConditionalGeneration) checkpoints can't be trained (weight-sync key mismatch)


adithya-s-k:
# `AsyncGRPOTrainer`: vision-language (`*ForConditionalGeneration`) checkpoints can't be trained (weight-sync key mismatch)

### Environment

- `trl` `main` (`trl.experimental.async_grpo`)
- `transformers` 5.11, `vllm` 0.22.0 (native 4-phase RL weight transfer; needs
  [#5892](https://github.com/huggingface/trl/pull/5892) for the `WeightTransferClient` 0.22 path)
- Layout: vLLM server (`--data-parallel-size 2`) + trainer FSDP2, weights synced over NCCL

### What happens

Training a vision-language checkpoint (e.g. `Qwen/Qwen3.5-4B`, `google/gemma-4-E4B-it`) — even for text-only
RL — crashes at the **first weight sync**:

```
ValueError: There is no module or parameter named 'model' in Qwen3_5ForConditionalGeneration.
The available parameters are: {'language_model.model.layers...', 'visual.blocks...'}
```

### Root cause

`async_grpo_trainer.py` loads the policy with a hardcoded:

```python
model = AutoModelForCausalLM.from_pretrained(model, ...)
```

For a VL checkpoint this yields the **text tower** (weight keys `model.*`). But vLLM serves the model under
its config architecture `*ForConditionalGeneration` (keys `language_model.model.*` + `visual.*`). The NCCL
weight transfer broadcasts the trainer's `model.*` keys; vLLM's loader can't map them → `ValueError`. The two
sides disagree on the model's parameter namespace.

### Proposed fix

Load the model class that matches what vLLM serves. When the checkpoint's `config.architectures` is a
conditional-generation / image-text-to-text type, load via `AutoModelForImageTextToText` and freeze the
vision tower (text-only RL), so the trainer's keys (`language_model.model.*` + `visual.*`) match vLLM:

```python
archs = getattr(AutoConfig.from_pretrained(model), "architectures", None) or []
if any("ConditionalGeneration" in a or "ImageTextToText" in a for a in archs):
    model = AutoModelForImageTextToText.from_pretrained(model, ...)
    for name, p in model.named_parameters():
        if "visual" in name or "vision" in name:
            p.requires_grad = False
else:
    model = AutoModelForCausalLM.from_pretrained(model, ...)
```

Text models are unaffected.

[...]

Issue #6027, opened 2026-06-12
AsyncGRPOTrainer: environment_factory rewards are never passed to reward_funcs


adithya-s-k:
# `AsyncGRPOTrainer`: `environment_factory` rewards are never passed to `reward_funcs`

### Environment

- `trl` `main` (`trl.experimental.async_grpo`)
- `transformers` 5.11, `vllm` 0.22.0 (native 4-phase RL weight transfer; needs
  [#5892](https://github.com/huggingface/trl/pull/5892) for the `WeightTransferClient` 0.22 path)
- Layout: vLLM server (`--data-parallel-size 2`) + trainer FSDP2, weights synced over NCCL

### What happens

`AsyncGRPOTrainer` accepts `environment_factory` and exposes each env instance's methods as tools (great).
But the reward path scores **only from completions + dataset-row fields** — it never reads anything off the
env. A reward function that reads env state (the standard `GRPOTrainer` pattern — e.g. a verifier score
computed after the rollout) raises:

```
TypeError: reward_func() missing 1 required positional argument: 'environments'
```

In `GRPOTrainer`, reward functions receive the live env instances as `environments=`; the async worker has
no equivalent. So environment-/verifier-based rewards — the main reason to use `environment_factory` — can't
be expressed in async; the env can only act as a tool provider.

### Root cause

`async_rollout_worker.py::AsyncRolloutWorker._score_group` builds the reward kwargs from the rollout group
only:

```python
kwargs = dict(
    completions=group.completions,
    prompt=group.prompt,
    prompts=[group.prompt] * len(group.completions),
    completion_ids=group.completions_ids,
    **group.reward_kwargs,   # <- dataset-row fields only
)
```

The env pool (`self.environments[slot]`) is reset per rollout and its methods are used as tools, but the
env's post-rollout reward is never captured before the slot is reused, and never reaches `_score_group`.

### Proposed fix

Capture each rollout's env reward while its slot still holds the rollout state (the slot is only reset on the
next scheduling iteration), off the event loop since reading it may run a blocking verifier, and pass it to
scoring as a dedicated kwarg:

```python
# in the done-task handling, after appending the completion:
if self.environments is not None:
    env = self.environments[slot]
    group.env_rewards.append(await asyncio.to_thread(lambda: env.reward))

# in _score_group:
if group.env_rewards:
    kwargs["environment_reward"] = group.env_rewards
```

[...]

Issue #5912, opened 2026-06-02
RFC: Environment-owned reward


qgallouedec:
# RFC: Environment-owned reward

**Scope:** `GRPOTrainer` + `AsyncGRPOTrainer`. Same pattern applies to any online trainers with environment support.
**Out of scope:** Environment-owned dataset (RFC #5903)

## Summary

Let the environment own the reward.
Today the trainer owns reward computation: you pass `reward_funcs` to the trainer, and the trainer scores each completion. This RFC lets the environment return the reward instead.

In the RL-native formulation, where the environment that runs the episode is also the one that scores it.

This matters for multi-turn / stateful environments: the reward is usually a function of the environment's internal state (did the game end in a win? was the word guessed?), which the trainer does not have. Forcing `reward_funcs` then means leaking env state back out just so an external function can recompute what the env already knows.

## Other data points

- verifiers: the environment owns the reward, via a `rubric` (`score_rollout` / `score_group`).
- OpenEnv: gym-like — `step()` returns the reward directly. The environment owns it by design.

### Before

<img width="1729" height="1040" alt="Image" src="https://github.com/user-attachments/assets/cdddde54-b0a0-4d87-a305-934f24a6777d" />

### After

<img width="1728" height="1040" alt="Image" src="https://github.com/user-attachments/assets/655eacd9-bac4-430a-bb28-20f42f3a6370" />

**Before:** the rollout's completions are sent back to the trainer-owned `reward_funcs`, along with the environment itself, which score them (and often only need the env state to do so).
**After:** the environment, which already holds the state, returns the reward directly. The trainer still turns rewards into advantages that part is unchanged.

### Example: wordle-like game

The reward (was the word guessed?) can be written as a function, but it needs the target word, which the environment already knows.

#### Before

The reward function lives in the trainer and must receive the target through an access to the environment (`environments.target`)

```python
def solved_reward(completions, environments, **kwargs):
    return [1.0 if env.target in completion else 0.0 for completion, env in zip(completions, environments)]

trainer = GRPOTrainer(
    model=my_model,
    reward_funcs=solved_reward,              # trainer owns the reward
    train_dataset=dataset, 
    environment_factory=WordleEnvironment,
)
```

#### After

The environment scores the episode it just ran:

```python
class WordleEnvironment:
    def reset(self) -> None:
        self._target = sample(list_of_words)
        self._solved = False

    def guess(self, word: str) -> str:
        """Submit a guess, get per-letter feedback."""
[...]

Issue #5903, opened 2026-06-01
RFC: Agent RL training with environment: environment-owned dataset, if any


qgallouedec:
**Scope:** `GRPOTrainer` + `AsyncGRPOTrainer`. Same pattern applies to any online trainers with environment support.
**Out of scope:** reward de-correlation (separate RFC).

## Summary

Let the environment own the dataset, if any.
Today the trainer owns the data. The trainer samples a prompt from the dataset, resets the environment with that prompt as argument.
This RFC shifts that ownership to the env, which can hold its own corpus, self-sample, or generate procedurally.

For some environments, a dataset does not make sense (e.g., procedurally generated environments). This new design should allow this, instead of the current dummy dataset workaround.

## Other data points

- verifiers: the **environment owns the dataset**. But an env **must** have a dataset (it raises if none is provided), which I don't think is a good design choice, see below.
- OpenEnv: most environments don't need any external dataset to sample from.

### Before

<img width="1729" height="1040" alt="Image" src="https://github.com/user-attachments/assets/a0309953-3368-4351-8be0-686b46fe2ee2" />

**Before:** the Trainer samples from the dataset it owns and pushes the data into the env via
`reset(prompt)`.

### After

<img width="1928" height="1054" alt="Image" src="https://github.com/user-attachments/assets/d8b4a719-665a-4adf-a35f-ec1fcb163d71" />

**After:** the Trainer only calls `env.reset()` and the env returns the prompt (sampled
from its own dataset, if any).

### Example 1, wordle-like, i.e. when a dataset would fit

For some environments, like Wordle, there is a natural notion of dataset (the list of target words).

#### Before

Two options, both awkward. Either the trainer owns the words and iterates over them, with each row
passed to `reset` as a kwarg:

```python
dataset = load_dataset("my_words")  # rows like {"prompt": "Guess the 5-letter word.", "target": "crane"}

class WordleEnvironment:
    def reset(self, target: str) -> None:    # receives the row's columns
        self._target = target

    def guess(self, word: str) -> str:
        """Submit a guess, get per-letter feedback."""
        ...

trainer = GRPOTrainer(
    model=my_model,
    reward_funcs=my_reward_func,
    train_dataset=dataset,                   # trainer owns and iterates the words
    environment_factory=WordleEnvironment,
)
```

…or the environment owns the words, but you still have to fabricate a dummy dataset whose only job is
to drive the loop length:

```python
[...]

PR #6068, opened 2026-06-15
async_grpo: inject the environment_factory reset() observation into the prompt

### LINKED ISSUE #6067: `AsyncGRPOTrainer`: the `environment_factory` `reset()` observation is dropped — the policy trains on the bare dataset prompt
### Summary

`AsyncGRPOTrainer`'s rollout worker calls `env.reset(**row)` but **discards its return value**. `GRPOTrainer` treats `reset()`'s return as the environment's initial observation and appends it to the prompt. So an `environment_factory` whose task is delivered via `reset()` (rather than baked into the dataset `prompt` column) is generated against the bare prompt — and if that prompt is intentionally empty, the policy gets **no task**.

### Where

`trl/experimental/async_grpo/async_rollout_worker.py`: the prompt is built from `row["prompt"]` and tokenized, then `self.environments[slot].reset(**row)` is called with its return ignored.

For contrast, `GRPOTrainer` (`grpo_trainer.py`):

```python
observation = environment.reset(**reset_kwargs)
...
prompt[-1]["content"] += observation
```

### Evidence

Same environment + config, Qwen3.5-4B on a Harbor agentic suite:

- `GRPOTrainer`: reward `0.25, 0, 0, 0.5, 0.5`.
- `AsyncGRPOTrainer`: reward `0, 0, 0, 0, 0` — the agent has no task ("the /workdir directory is empty … no task description"), burns its tool budget, submits nothing useful.
- After the fix below: `0.5, 1.0, 1.0, 1.0, 1.0`.

The same environment earns reward under `GRPOTrainer` and zero under `AsyncGRPOTrainer`, which isolates the gap to the async rollout path.

### Fix

Reset the env before building the prompt and fold the observation into the last prompt message, mirroring `GRPOTrainer` (injected once per group — it's task-level). **TITO-safe:** this only builds the turn-0 prompt (tokenized once via the training template); no sampled tokens are re-encoded. PR: #6068.

### Related

#6027 (env rewards never passed to `reward_funcs`), #6028 (VL checkpoints fail to load) — other async `environment_factory` gaps.

adithya-s-k:
Closes #6067.

`AsyncGRPOTrainer`'s rollout worker called `env.reset(**row)` but discarded its return value, so an `environment_factory` whose task is delivered via the initial observation (rather than the dataset `prompt` column) was generated against the bare prompt — if that prompt is empty, the policy sees no task and trains on irrelevant completions. `GRPOTrainer` already folds `reset()`'s observation into the prompt (`prompt[-1]["content"] += observation`); this mirrors that in the async rollout worker: reset the env before building the prompt, and fold the observation into the last prompt message once per group (it is task-level, identical across the group's generations).

**TITO-safe.** This only affects the turn-0 prompt, which is tokenized once via the training chat template. It does not decode-then-re-encode any sampled tokens — the running buffer (`prompt_ids = prompt_ids + turn_ids + suffix_ids`), `parse_response`-for-dispatch, and the tool-response delta are all unchanged. The policy is still trained on the exact tokens it sampled; it just samples them in response to the actual task instead of an empty prompt.

**Verified end-to-end.** An `environment_factory` (the Harbor data-agent suite) that earns reward under `GRPOTrainer` previously earned 0 under `AsyncGRPOTrainer` (the agent flailed with "no task description") and now earns reward with the same config after this change.

Happy to add a unit test if you have a preferred harness for the async rollout loop.

Related: #6027, #6028 (other async `environment_factory` gaps).

---

> [!NOTE]
> **Low Risk**
> Turn-0 prompt assembly only in the async rollout child; no auth or training-token path changes beyond fixing missing task text for env-based tasks.
> 
> **Overview**
> **Async GRPO rollout** now matches **GRPOTrainer** for `environment_factory` setups where the task text comes from `reset()` rather than the dataset `prompt` column.
> 
> In `_generate_loop`, the worker **resets the env before** building `prompt_ids`, **keeps** the `reset(**row)` return value, and **merges** that observation into the last chat message when a new rollout group is created (once per group, not per generation). String and multimodal-style list content are both handled.
> 
> Previously `reset()` ran after slot assignment but its return was ignored, so generation could run on an empty or incomplete prompt and earn no reward.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b158e15c69b0557ddc83c8ffa5223d5ebd9c3bb5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/async_grpo/async_rollout_worker.py
+++ b/trl/experimental/async_grpo/async_rollout_worker.py
@@ -313,8 +313,34 @@ class _AsyncRolloutLoop:
                 self._heartbeat_value.value = time.time()
                 while free_slots and not stop_event.is_set():
                     group_id, row = next(work_iter)
+                    slot = free_slots.pop()
+                    # Reset the env BEFORE building the prompt and capture its initial observation (e.g. a
+                    # task instruction). The reset() return was previously discarded, so an
+                    # environment_factory whose task lives in the observation (not the dataset prompt) was
+                    # generated against the bare prompt. Mirror GRPOTrainer: fold the observation into the
+                    # last prompt message. It is task-level (identical across the group's generations), so
+                    # inject it once, when the group is first created.
+                    observation = self.environments[slot].reset(**row) if self.environments is not None else None
+
                     if group_id not in pending_groups:
                         prompt = row["prompt"]
+                        if observation is not None:
+                            last = prompt[-1]
+                            if isinstance(last["content"], str) and isinstance(observation, str):
+                                content = last["content"] + observation
+                            else:
+                                obs = (
+                                    observation
+                                    if isinstance(observation, list)
+                                    else [{"type": "text", "text": observation}]
+                                )
+                                base = (
+                                    last["content"]
+                                    if isinstance(last["content"], list)
+                                    else [{"type": "text", "text": last["content"]}]
+                                )
+                                content = base + obs
+                            prompt = prompt[:-1] + [{**last, "content": content}]
                         prompt_ids = self.tokenizer.apply_chat_template(
                             prompt,
                             return_dict=False,
@@ -341,10 +367,6 @@ class _AsyncRolloutLoop:
                             model_version=self.model_version,
                         )
                         pending_completed[group_id] = 0
-
-                    slot = free_slots.pop()
-                    if self.environments is not None:
-                        self.environments[slot].reset(**row)
 
                     task = asyncio.create_task(
                         self._generate_one(pending_groups[group_id].prompt, tool_dict=self._sync_tool_dicts[slot])


trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -313,8 +313,34 @@ async def _generate_loop(self, stop_event: asyncio.Event) -> None:
...
+                    slot = free_slots.pop()
+                    # Reset the env BEFORE building the prompt and capture its initial observation (e.g. a
+                    # task instruction). The reset() return was previously discarded, so an
+                    # environment_factory whose task lives in the observation (not the dataset prompt) was
+                    # generated against the bare prompt. Mirror GRPOTrainer: fold the observation into the
+                    # last prompt message. It is task-level (identical across the group's generations), so
+                    # inject it once, when the group is first created.
+                    observation = self.environments[slot].reset(**row) if self.environments is not None else None
+
                     if group_id not in pending_groups:
```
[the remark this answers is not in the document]
adithya-s-k: Thanks. This is by design: a GRPO group is N samples of the **same** prompt (group-relative advantage requires it), so `reset(**row)` is expected to be deterministic per row — all N observations are identical, and folding one in is correct (and matches `GRPOTrainer` for that case). The async `RolloutGroup` is single-prompt by construction and can't represent per-rollout prompts anyway; a reset that returns rollout-specific text would already be outside GRPO's group semantics.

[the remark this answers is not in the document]
qgallouedec: yep that's the discussion we had some time ago...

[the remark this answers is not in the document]
qgallouedec: > This is by design: a GRPO group is N samples of the same prompt

That's true, but this is something we can't guaranty on the grpo side. The best is probably to just recommend that the environment returns the very same state when reset with the same argument.
I'll open a next pr to "fix" this.

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -313,8 +313,34 @@ async def _generate_loop(self, stop_event: asyncio.Event) -> None:
...
+                                    if isinstance(observation, list)
+                                    else [{"type": "text", "text": observation}]
+                                )
+                                base = (
+                                    last["content"]
+                                    if isinstance(last["content"], list)
+                                    else [{"type": "text", "text": last["content"]}]
+                                )
+                                content = base + obs
+                            prompt = prompt[:-1] + [{**last, "content": content}]
```
qgallouedec, line 343:
```suggestion
                            prompt = prompt[:-1] + [{**last, "content": last["content"] + observation}]
```

qgallouedec (commented):
async GRPO is text-only (no multimodal), so the list branches are dead here. Rebuilt the message instead of mutating in place since row is reused across generations/epochs

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: Is this observation based on `reset` of the env really a needed contract? It seems to me that we assume too much from the environment ?

I do understand feature parity with `GRPOTrainer` but  maybe we can relax this contract.

qgallouedec: the thing is that, for some cases, it's almost necessary to have an initial observation. Take sudoku. How to prompt the model with the initial board if the first observation only comes after the first action?

You can have a workaround with the dataset containing the initial state, but it's weird. Hopefully everything around environment will be easier once we've discussed #5903 (and #5912 for rewards)

AmineDiro: I see the point. For the Sudoku example, we can include the board's initial state in the dataset prompt column so the env doesn't have to generate it itself. But I see the global point of a partially/hidden state that we can't get from datasets for the environment. 

My main question is maybe design and ux. `reset` function should only exist to create _fresh_ envs and is overloaded here. If we need to seed something, this can be from the userside. The discussions https://github.com/huggingface/trl/issues/5903 (and https://github.com/huggingface/trl/issues/5912 for rewards) definitely will help, but still leaning towards giving the user a lot more freedom to control generation (with some bounds 👍🏼 ) to avoid weird constraints in the API

qgallouedec: Good points, but I'd push back on "`reset()` is overloaded." Under the gym-style contract `reset()` returns the initial observation, and in the MDP/POMDP view that observation *is* the first element of the trajectory ($o_0$). So consuming its return isn't repurposing `reset()`, it's using the value it's defined to produce.

Seeding can (should) comes from the `reset(**row)` argument, which is the gym pattern (`reset(seed=, options=)`), so I think this design is actually the standard one rather than a constraint to avoid.

On "put the initial state in the dataset prompt column": that's the part I find awkward (hence #5903)
tldr: the dataset would then own environment-internal state. And it doesn't generalize: eg, for a POMDP the states are hidden by design. Serializing it into the prompt would leak the latent state to the policy; another example, a static column can't carry a freshly-sampled hidden state anyway (think about PGE). So `reset() → o_0` is the only place that observation can and should come from for a whole class of envs.

Re: the "bounded hook to control generation": I don't think this PR touches generation at all? It's purely turn-0 prompt assembly (what task text the policy conditions on), not how tokens get sampled. And #5903 isn't really about generation control either; it's about data/observation ownership. This PR's `reset() → observation` folding goes in that direction, so I'd see it as a narrow parity fix consistent with where #5903 wants the API to head.

qgallouedec:
Thanks @adithya-s-k, closing this one as superseded, not rejected.

Your diagnosis was right and it shipped: #6072 built directly on this PR (you're a co-author on the merge commit) and closed #6067. The only thing that changed is where the observation is folded in; this PR injected it once per group and shared one `prompt_ids` across generations, which silently breaks TITO for every generation but the first when `reset()` is stochastic (which is not an edge case). So the merged version folds a per-generation observation and returns the `prompt_ids` that `_generate_one` already tokenized.

Nothing left to do here! the branch also conflicts with main now after the OpenEnv/pooling rework.

PR #6077, opened 2026-06-16
Fix per-chunk lm_head.weight all-gathers under FSDP2 + chunked_nll

### LINKED ISSUE #6076: chunked_nll under FSDP2 triggers per-chunk all-gathers of lm_head.weight during backward
When using loss_type='chunked_nll' (the default since #5846) under FSDP2 with reshard_after_forward=True (accelerate's default), training is significantly slower than necessary.

### Root cause

`_chunked_cross_entropy_loss` processes tokens in chunks, each wrapped with `torch.utils.checkpoint.checkpoint`. Gradient checkpointing discards chunk activations after the forward and recomputes each chunk's forward during the backward. If `lm_head.weight` is a DTensor at that point, FSDP2 must all-gather it before each chunk's recomputation. With `reshard_after_forward=True`, the parameter was resharded after the forward, so the gather happens N times (once per chunk) instead of once.

This affects all FSDP2 users of `chunked_nll`, regardless of PEFT.

### Current workaround

That is why currently an explicit warning is emitted recommending `--fsdp_reshard_after_forward false`, which avoids resharding and eliminates the redundant gathers. However, this is a global setting that increases peak memory for the entire model.

### Expected behaviour

TRL should fix this transparently.

albertvillanova:
Fix per-chunk lm_head.weight all-gathers under FSDP2 + chunked_nll.

Fix #6076.

This PR fixes a performance issue for chunked_nll under FSDP2, by calling full_tensor() on lm_head.weight once before the chunk loop, converting it to a plain tensor. All chunks then reference that plain tensor during backward recomputation: FSDP2 is never involved per-chunk.

### Motivation

Currently, for FSDP2 + chunked_nll, we raise a warning suggesting to use `--fsdp_reshard_after_forward false`.

However, `--fsdp_reshard_after_forward false` is a global setting that increases peak memory for the entire model. It keeps **all** model parameters resident in memory for the entire forward and backward.
- For a 7B model across 8 GPUs, that is the difference between parameters occupying `model_size / 8` vs. `model_size` on each device at peak.
- The `full_tensor()` fix gathers exactly one weight matrix (`lm_head.weight`) and only for the duration of the chunk loop. Every other parameter is unaffected.

Additionally, users should not pay for TRL's implementation detail: `reshard_after_forward=True` is the correct default for FSDP2: it minimises peak memory, which is the primary reason users choose FSDP2. Asking users to disable it globally because TRL's chunked-CE pattern accesses `lm_head.weight` outside FSDP's forward scope is making the user work around a TRL-internal concern. The fix belongs in TRL.

### Solution

- `_chunked_ce_forward` now calls `full_tensor()` on `lm_head.weight` (and bias) once before the gradient-checkpointed chunk loop, whenever the weight is a `DTensor`. The plain tensor result is passed to all chunks, so FSDP2 is never triggered per chunk during backward recomputation.
- Removes the `with_peft` flag from `_patch_chunked_ce_lm_head` — the fix applies to all FSDP2 paths (both `Shard(0)` under PEFT and `Replicate` under non-PEFT).
- Removes the warning that advised users to pass `--fsdp_reshard_after_forward false`; the issue is now fixed in code regardless of that setting.

#### Why `full_tensor()` is correct

`full_tensor()` is explicitly documented as differentiable. Its backward wraps the accumulated gradient back into a DTensor and reduce-scatters it to `lm_head.weight.grad`: the same communication that would happen anyway, now consolidated into a single op instead of N.

### Changes

**Refactoring and simplification:**

* Removed the `with_peft` parameter from `_patch_chunked_ce_lm_head` and all related logic, as PEFT-specific handling is no longer required in the chunked cross-entropy loss path (`trl/trainer/sft_trainer.py`) .
* Updated `_chunked_ce_forward` to always check for distributed tensors using `isinstance(lm_head_weight, torch.distributed.tensor.DTensor)`, removing the conditional on PEFT (`trl/trainer/sft_trainer.py`).

**Dependency and warning cleanup:**

* Removed the unused `import accelerate` statement (`trl/trainer/sft_trainer.py`).
* Deleted the runtime warning about FSDP2 with `reshard_after_forward=True`, since the code now properly avoids redundant all-gathers, making the warning unnecessary (`trl/trainer/sft_trainer.py`).

---

> [!NOTE]
> **Medium Risk**
> Touches distributed training loss/backward for the default SFT loss path; behavior should match prior math with better performance, but FSDP2 gradient sync on lm_head deserves regression checks.
> 
> **Overview**
> Fixes slow **FSDP2** training with the default **`loss_type="chunked_nll"`** by materializing **`lm_head.weight`** (and bias) once via **`DTensor.full_tensor()`** before the gradient-checkpointed chunk loop, so backward no longer triggers a separate FSDP2 gather on every chunk.
> 
> The **`with_peft`** flag on **`_patch_chunked_ce_lm_head`** and the init-time warning to set **`--fsdp_reshard_after_forward false`** are removed; docs and **`SFTConfig`** help text no longer mention that workaround.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63f31d0a7d1d9346f95570aee7e918af1bd3348e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (changes requested):
thanks! nice!, I'm still reviewing it carefully, but in the meantime: there is a mention of this to be removed in "Chunked cross-entropy for reducing peak memory usage" as well and in sft config

albertvillanova:
@qgallouedec, thanks for your review.

I have accordingly removed the suggestion to use `fsdp_reshard_after_forward false` from both SFTConfig and the docs about "Chunked cross-entropy".

In relation with the choice of using `full_tensor()` instead of `fsdp_reshard_after_forward false`, here are some arguments:

1. **Surgical fix vs. global memory regression**

    `--fsdp_reshard_after_forward false` keeps _all_ model parameters resident in memory for the entire forward and backward. For a 7B model across 8 GPUs, that is the difference between parameters occupying `model_size / 8` vs. `model_size` on each device at peak. The `full_tensor()` fix gathers exactly one weight matrix (`lm_head.weight`) and only for the duration of the chunk loop. Every other parameter is unaffected.

2. **Users should not pay for TRL's implementation detail**

    `reshard_after_forward=True` is the correct default for FSDP2: it minimises peak memory, which is the primary reason users choose FSDP2. Asking users to disable it globally because TRL's chunked-CE pattern accesses lm_head.weight outside FSDP's forward scope is making the user work around a TRL-internal concern. The fix belongs in TRL.

3. **full_tensor() is explicitly documented as differentiable**

    This is not a hack. PyTorch's own docstring states: "full_tensor is differentiable." Its backward is a well-defined reduce-scatter. Gradients flow correctly to `lm_head.weight` through a single communication, identical in volume to what FSDP2 would do anyway, just consolidated into one operation rather than N.

4. **The warning does not fix the PEFT case**

    Under FSDP2 + PEFT, `lm_head.weight` is `Shard(0)` (never gathered by the inner causal-LM FSDP unit) and causes a `RuntimeError: aten.mm.default got mixed torch.Tensor and DTensor`. Setting `reshard_after_forward=False` does not change this: the weight is still `Shard(0)` when `_chunked_ce_forward` runs. `full_tensor()` is the only fix for the PEFT path; extending it to the non-PEFT path is a zero-cost generalisation.

5. **"It just works" across any config**

    With `full_tensor()`, the behaviour is correct and efficient regardless of whether the user sets `reshard_after_forward=True` or `False`, regardless of whether they use PEFT, and regardless of the number of chunks. There is no config the user can set that produces the slow path. A workaround that requires users to know about an interaction between TRL's loss internals and an accelerate flag is a latent footgun: most users will never find the warning.

qgallouedec (approved):
Ran a correctness + peak-memory check on 4×H100, FSDP2, Qwen2.5-7B (untied + the separate-`lm_head.weight` DTensor case the fix targets), non-PEFT + PEFT.

Loss parity (20-step SFT): run-to-run noise floor (same config, two runs) is 0.067 max |Δloss|. Every PR-vs-reference diff is under it: PR vs main `chunked_nll` = 0.057, PR `chunked_nll` vs `nll` = 0.066, PR vs old `reshard=false` path = 0.051. Steps 1–2 bit-identical everywhere; later drift is just bf16/FSDP reduction-order noise.

Gradient (drift-free, step-1 `grad_norm`): non-PEFT main=84.11 / PR=83.93 / nll=84.13 (~0.3% spread); PEFT PR=main=1.619. No NaN/hang/crash.

Peak VRAM*(long-context: 4096 ctx, backbone checkpointing, ~16 chunks):

| config | peak GB |
|---|---|
| PR `chunked_nll`, reshard=true (new default) | 35.9 |
| main `chunked_nll`, reshard=true (old, buggy) | 35.9 |
| main `chunked_nll`, reshard=false (removed workaround) | 45.3 |
| PR `nll`, reshard=true | 38.0 |

`full_tensor` adds zero peak vs the old per-chunk path (35.9 == 35.9: peak is set by the backbone, and the per-chunk path already materializes the full weight transiently). The new default is 9.4 GB lower than the `reshard=false` workaround this PR removes, and chunked still beats `nll`, so the fix doesn't erode the chunked savings.

✅ Correctness equivalent within non-determinism (non-PEFT + PEFT), and no peak-memory regression -> removing the `reshard=false` advice is justified (it was actually worse for memory). LGTM.

PR #6079, opened 2026-06-16
fix(data): correct extract_prompt index for divergence at start and prefix completions

he-yufeng:
## What

`extract_prompt` finds the shared prefix between `chosen` and `rejected` by walking until they diverge. Two edge cases produce wrong (sometimes corrupt) output.

1. **Divergence at index 0 with a trailing space in `chosen`.** When the two completions differ at the very first element, `idx` is `0`, so the space check reads `example["chosen"][idx - 1]` which is `chosen[-1]`, the *last* element. If that happens to be a space, `idx` becomes `-1` and the slices wrap around. Example:

   ```python
   extract_prompt({"chosen": "Paris ", "rejected": "London"})
   # current:  {'prompt': 'Paris', 'chosen': ' ', 'rejected': 'n'}
   # expected: {'prompt': '',      'chosen': 'Paris ', 'rejected': 'London'}
   ```

2. **`chosen` is a prefix of `rejected` (or equal).** When no divergence is found within the common length, the loop never breaks and `idx` keeps the loop variable's last value, `min(len) - 1`, instead of `min(len)`. The last shared element leaks into both completions:

   ```python
   extract_prompt({
       "chosen":   [{"role": "user", "content": "Q"}, {"role": "assistant", "content": "A"}],
       "rejected": [{"role": "user", "content": "Q"}, {"role": "assistant", "content": "A"},
                    {"role": "user", "content": "more"}],
   })
   # current 'chosen' is [A] and 'rejected' still contains A; expected 'chosen' is [] and the prompt is the full [Q, A]
   ```

## Fix

Initialize `idx` to the common length, set it to the first mismatching position when one is found, and guard the space-trim with `idx > 0`. Behavior for all normally-diverging cases is unchanged (byte-identical).

## Verification

```
python -m pytest tests/test_data_utils.py::TestExtractPrompt -q   # 8 passed
ruff check trl/data_utils.py tests/test_data_utils.py             # All checks passed!
```

Added two tests to `TestExtractPrompt` (divergence at start, and chosen-is-prefix-of-rejected). Both fail on current `main` and pass with the fix; the existing cases keep passing.

使用 Claude Code 辅助

---

> [!NOTE]
> **Medium Risk**
> Incorrect prompt splitting can mislabel preference training data, but the change is localized to `extract_prompt` with targeted tests and unchanged behavior for normal cases.
> 
> **Overview**
> Fixes **`extract_prompt`** so implicit preference examples split into `prompt` / `chosen` / `rejected` correctly in two edge cases.
> 
> The divergence index now starts at the **full common length** and only moves to the first mismatch; when **`chosen` is a prefix of `rejected`**, the entire shared part becomes the prompt instead of leaving the last shared element in the completions. The trailing-space trim before the split runs only when **`idx > 0`**, so first-token divergence (e.g. `"Paris "` vs `"London"`) no longer reads `chosen[-1]` and corrupts slices.
> 
> **`TestExtractPrompt`** gains regression tests for both scenarios; typical diverging cases stay byte-identical.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ae449ca0f95e402377e1f4e158b7a12ea2cc093. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_data_utils.py
+++ b/tests/test_data_utils.py
@@ -1145,6 +1145,34 @@ class TestExtractPrompt(TrlTestCase):
         example_extracted_prompt = maybe_extract_prompt(self.example_explicit_prompt_standard)
         assert example_extracted_prompt == self.example_explicit_prompt_standard, "The prompt should remain unchanged."
 
+    def test_extract_prompt_diverge_at_start(self):
+        # Chosen and rejected diverge at the very first token: there is no shared prompt. A trailing space in chosen
+        # must not be mistaken for a space before the prompt (which previously wrapped the index to -1).
+        example = {"chosen": "Paris ", "rejected": "London"}
+        assert extract_prompt(example) == {"prompt": "", "chosen": "Paris ", "rejected": "London"}
+
+    def test_extract_prompt_chosen_is_prefix_of_rejected(self):
+        # When chosen is a prefix of rejected, the whole chosen is the shared prompt.
+        example = {
+            "chosen": [
+                {"role": "user", "content": "What color is the sky?"},
+                {"role": "assistant", "content": "It is blue."},
+            ],
+            "rejected": [
+                {"role": "user", "content": "What color is the sky?"},
+                {"role": "assistant", "content": "It is blue."},
+                {"role": "user", "content": "Are you sure?"},
+            ],
+        }
+        assert extract_prompt(example) == {
+            "prompt": [
+                {"role": "user", "content": "What color is the sky?"},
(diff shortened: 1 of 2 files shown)


qgallouedec:
Does it correspond to any real use case?

qgallouedec:
This profile belongs to a bot that opens a large number of pull requests. In line with our contribution guidelines, this is not the type of contribution we encourage.

PR #6085, opened 2026-06-17
Add `router_aux_loss_coef`

qgallouedec:
## What

Adds a first-class `router_aux_loss_coef` training argument (default `0.001`) to `SFTConfig`, `GRPOConfig`, `RLOOConfig`, and `AsyncGRPOConfig`. The MoE load-balancing auxiliary loss is now **on by default** for Mixture-of-Experts models and controlled directly from the config.

## Motivation

Previously, enabling the load-balancing aux loss required reaching into the model config — passing `output_router_logits=True` (and optionally `router_aux_loss_coef`) through `model_init_kwargs`. This was undocumented-by-feel, easy to forget, and meant most MoE runs silently trained *without* the load-balancing term, which is known to cause token/expert imbalance. Making it a config argument that's on by default aligns TRL with how MoEs are meant to be trained, and gives a single, discoverable knob to tune or disable it.

## Example

```python

# Aux loss is on by default for MoE models (no model_init_kwargs needed)
trainer = SFTTrainer(
    model="Qwen/Qwen3-30B-A3B",
    args=SFTConfig(router_aux_loss_coef=0.001),  # default; set 0.0 to disable
    train_dataset=dataset,
)
trainer.train()
```

For non-MoE models the argument has no effect.

## Migration

The aux loss is now **enabled by default** for MoE models. This changes training dynamics for MoE runs that previously did not opt in.

| Goal | Before | After |
|------|--------|-------|
| Enable aux loss | `model_init_kwargs={"output_router_logits": True}` | nothing — on by default |
| Tune the weight | `router_aux_loss_coef` in the model config | `SFTConfig(router_aux_loss_coef=...)` |
| Disable aux loss | (was off unless opted in) | `SFTConfig(router_aux_loss_coef=0.0)` |

If you want to reproduce pre-PR behavior exactly (aux loss off), set `router_aux_loss_coef=0.0`.

> [!NOTE]
> Applies to `SFTTrainer`, `GRPOTrainer`, `RLOOTrainer`, and `AsyncGRPOTrainer`. The DPO-family trainers (DPO/BCO/ORPO/KTO/CPO) are unchanged and still use `output_router_logits=True` in the model config; migrating them is a deferred follow-up.

---

> [!NOTE]
> **Medium Risk**
> Default-on MoE auxiliary loss (0.001) changes training loss and behavior for MoE models across several trainers; mis-detection or VLM forward kwargs could affect whether aux loss is applied.
> 
> **Overview**
> MoE load-balancing auxiliary loss is now controlled from trainer configs via **`router_aux_loss_coef`** (default **`0.001`**, set **`0.0`** to disable) on **SFT**, **GRPO**, **RLOO**, and **async GRPO**, instead of wiring **`output_router_logits`** / coef through **`model_init_kwargs`** or the model config.
> 
> Trainers treat a model as MoE when the text config exposes **`output_router_logits`**, enable aux loss only when the coef is non-zero, and pass **`output_router_logits=True`** on forward where needed (notably **VLMs**). **SFT** also keeps the text config’s router flags/coef in sync with the training arg so disabling via **`router_aux_loss_coef=0.0`** is reliable.
> 
> Tests assert **`aux_loss`** in generic train runs when **`aux_loss_enabled`**, and drop the dedicated SFT MoE-only test; **RewardTrainer** docs/plumbing no longer document MoE aux via **`model_init_kwargs`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 201bac8a38450e2aacef821558418ce104f39bad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -416,14 +416,10 @@ def __init__(
...
-        self.aux_loss_enabled = getattr(model.config, "output_router_logits", False)
-        self.router_aux_loss_coef = getattr(model.config, "router_aux_loss_coef", 0.0)
-        if self.aux_loss_enabled and self.router_aux_loss_coef == 0.0:
-            logger.warning(
-                "You set `output_router_logits=True` in the model config, but `router_aux_loss_coef` is `0.0`, so the "
-                "auxiliary loss has no effect. Set `router_aux_loss_coef > 0.0` to enable it."
-            )
+        # MoE load-balancing auxiliary loss, applied to Mixture-of-Experts models (no effect otherwise)
+        is_moe = getattr(model.config, "output_router_logits", None) is not None
+        self.aux_loss_enabled = is_moe and self.args.router_aux_loss_coef != 0.0
```
AmineDiro, line 421: So if I understand correctly it is enabled by default ?

qgallouedec: Do you think it's a good default?

AmineDiro: Yes I do think so, just checking that it is a decision.  token/experts imbalance seems to have clear grad amplification.

tests/test_sft_trainer.py
```diff
@@ -627,10 +627,10 @@ def test_train_nll_loss_vlm(self, model_id):
     def test_train_moe_model_with_aux_loss(self):
         dataset = load_dataset("trl-internal-testing/zen", "standard_language_modeling", split="train")
 
+        # MoE aux loss is on by default (router_aux_loss_coef defaults to 0.001)
         training_args = SFTConfig(
             output_dir=self.tmp_dir,
             report_to="none",
-            model_init_kwargs={"output_router_logits": True},
```
AmineDiro, line 633: Should we keep an explicit value or use the default (implicit)

AmineDiro (approved):

tests/test_sft_trainer.py
qgallouedec: I just removed this test, it didn't make sense anymore

PR #6116, opened 2026-06-19
Make `evaluate()` accept the same dataset types as the trainer

### LINKED ISSUE #6115: SFTTrainer.evaluate() to accept the same kinds of datasets as train()
### Feature request

`SFTTrainer.__init__()` preprocesses the train and eval datasets it receives, and so training works with dataset types described in the documentation (LM, prompt-completion, etc). However, `evaluate()` from the transformer `Trainer` is not overridden, and so it requires a processed dataset. My problem is that if I want to use this function, I have to replicate what happens in `__init__` exactly. Ideally, this shouldn't be needed; `evaluate()` could run the same preprocessing as `__init__()` and run the evaluation as it happens inside the training loop.

Alternatively, the `_preprocess_dataset` could be made public (or a public version could be added), with the arguments that take local values in `__init__` (e.g. `processor`) hidden. Users could then use this function to convert their datasets so that they are guaranteed to be compatible with the ones they added to `__init__`.

### Motivation

The motivation is to support the standard _train - valid - test_ set distinction. `SFTTrainer` only accepts a train and an eval dataset; there is no valid / test distinction. One can supply both valid and test datasets to `eval_dataset` as a `dict`, but then both will be used for evaluation during training, which goes against the idea of a test dataset: one that should only be used at the end.

Currently there is no easy way to use a test set as described above; not one that I know of, that is.

### Your contribution

I would gladly implement this feature if I could discuss it with a core contributor first (e.g. which of the above solutions they would prefer).

qgallouedec:
Fixes #6115.

`SFTTrainer`, `DPOTrainer`, and `RewardTrainer` preprocess datasets in `__init__`, but `evaluate()` (inherited from `transformers.Trainer`) does not. So a dataset passed directly to `evaluate(eval_dataset=...)` (e.g. a held-out test set) had to be manually preprocessed, while the exact same raw format works at init.

This adds an `evaluate()` override to each of these trainers that runs the same `_prepare_dataset` step as `__init__`. It's idempotent (already-tokenized datasets pass through untouched) and leaves `str` keys (datasets prepared at init) alone. 

Each override mirrors its own `__init__` block (SFT respects `skip_prepare_dataset`/packing/formatting_func; DPO skips for vision datasets; Reward always prepares).

Online trainers (GRPO, RLOO) don't need this: their eval datasets are prompt-only and processed during generation.

---

> [!NOTE]
> **Low Risk**
> Targeted API fix with mirrored init logic and new tests; Reward’s `pad_token_id` config sync is a small behavioral fix for eval/scoring correctness.
> 
> **Overview**
> **`evaluate(eval_dataset=...)` now accepts the same raw dataset formats as trainer construction** for `SFTTrainer`, `DPOTrainer`, and `RewardTrainer` (fixes #6115). Each trainer overrides `evaluate` to run the same `_prepare_dataset` path as `__init__` before delegating to `transformers.Trainer`, with idempotent handling for already-tokenized data and no re-processing for `str` dataset keys.
> 
> **DPO** skips preparation for vision datasets; when `precompute_ref_log_probs` is enabled it also precomputes reference log-probs on ad-hoc eval data. **SFT** stores `_formatting_func` and `_skip_prepare_dataset` on the instance so eval uses the same packing/formatting rules as training. **Reward** additionally sets `model.config.pad_token_id` from the tokenizer so sequence-classification scoring can find the last non-pad token during eval.
> 
> Regression tests cover raw (and dict) eval datasets for all three trainers, including DPO with/without precomputed ref log-probs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 06305ffdcd68f26c9929bb0dd58f597f61126e7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_dpo_trainer.py
+++ b/tests/test_dpo_trainer.py
@@ -217,6 +217,19 @@ class TestDPOTrainer(TrlTestCase):
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
 
+    def test_evaluate_with_raw_dataset(self):
+        # `evaluate` should accept the same (unprocessed) dataset types as the trainer, e.g. a held-out test set
+        # passed directly to `evaluate`. See https://github.com/huggingface/trl/issues/6115.
+        dataset = load_dataset("trl-internal-testing/zen", "standard_preference", split="train")
+
+        training_args = DPOConfig(output_dir=self.tmp_dir, report_to="none")
+        trainer = DPOTrainer(
+            model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5", args=training_args, train_dataset=dataset
+        )
+
+        metrics = trainer.evaluate(eval_dataset=dataset)
+        assert metrics["eval_loss"] is not None
+
     @pytest.mark.parametrize(
         "config_name",
         [
--- a/tests/test_dpo_trainer.py
+++ b/tests/test_dpo_trainer.py
@@ -217,12 +217,16 @@ class TestDPOTrainer(TrlTestCase):
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
 
-    def test_evaluate_with_raw_dataset(self):
+    @pytest.mark.parametrize("precompute_ref_log_probs", [False, True])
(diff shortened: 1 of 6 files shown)


albertvillanova (approved):
Thanks.

PR #6118, opened 2026-06-19
Handle missing vllm_ascend imports gracefully

burtenshaw:
## Summary
- guard the optional `vllm_ascend` imports in TRL's vLLM client and server paths
- fall back to the default NCCL communicator when the module metadata is present but the Python package is not importable
- preserve current behavior for environments where `vllm_ascend` is actually installed

## Reproduction
On Hopper, installing `trl[vllm]` with `vllm==0.11.0` caused both OpenEnv GRPO examples to fail during import:

```text
ModuleNotFoundError: No module named 'vllm_ascend'
```

The failing path was gated by `is_vllm_ascend_available()`, but the import itself was still not safe.

---

> [!NOTE]
> **Low Risk**
> Small defensive change at import/communicator setup only; behavior is unchanged when vllm_ascend is actually installed.
> 
> **Overview**
> Fixes **import-time crashes** when `is_vllm_ascend_available()` is true (package metadata present) but `vllm_ascend` cannot be imported—e.g. Hopper with `trl[vllm]` and `vllm==0.11.0`.
> 
> In **`trl/generation/vllm_client.py`** and **`trl/scripts/vllm_serve.py`** (`WeightSyncWorkerExtension.init_communicator`), the optional `PyHcclCommunicator` import is wrapped in `try`/`except ModuleNotFoundError`. On failure, TRL logs a warning and keeps the default **`PyNcclCommunicator`** from vLLM. Environments with a working `vllm_ascend` install are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec7942687ee8b2f1523ce88be88b56005305dfd9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


burtenshaw:
Found this whilst testing the examples. Not sure if maybe you want to move the gate into `is_vllm_ascend_available()`?

qgallouedec:
Thanks @burtenshaw, and sorry for the slow reply.

This should fix itself: #6765 removes `is_vllm_ascend_available()` and both `vllm_ascend` import blocks, so there is nothing left to guard.

You were right in your comment that the gate belonged in `is_vllm_ascend_available()`. It is a bare `find_spec`, so it reports the package as available even when it cannot be imported, which is what you hit.

Closing in favour of #6765.

PR #6120, opened 2026-06-19
feat(ppo): add save_value_model flag to PPOConfig

### LINKED ISSUE #3293: Keep PPOTrainer Value Model (Critic Model)

mukund1985:
Adds a new `save_value_model: bool = False` field to `PPOConfig`. When enabled, `PPOTrainer.save_model()` writes the value model (critic) to a `value_model/` sub-directory alongside the policy checkpoint, making critic checkpointing and faithful RLHF run resumption possible.

Fixes #3293

## Before submitting

## AI writing disclosure

---

> [!NOTE]
> **Low Risk**
> Behavior is off by default and only extends checkpoint I/O on an explicit flag; no training-loop or inference path changes.
> 
> **Overview**
> Adds an opt-in **`save_value_model`** flag on **`PPOConfig`** (default **`False`**) so PPO checkpoints can include the critic, not just the policy.
> 
> When enabled, **`PPOTrainer.save_model()`** still writes the policy as today, then saves **`backup_model.value_model`** under a **`value_model/`** subfolder in the same output/checkpoint directory, including the same DeepSpeed swap pattern used for the policy save.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 482546f32142449f3283256cc832c18d15437c55. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/ppo/ppo_config.py
+++ b/trl/experimental/ppo/ppo_config.py
@@ -111,6 +111,10 @@ class PPOConfig(_BaseConfig):
             This setting applies to DeepSpeed ZeRO-3. If enabled, the policy model weights are gathered for generation,
             improving generation speed. However, disabling this option allows training models that exceed the VRAM
             capacity of a single GPU, albeit at the cost of slower generation.
+        save_value_model (`bool`, *optional*, defaults to `False`):
+            Whether to save the value model (critic) alongside the policy when calling :meth:`save_model` or at each
+            checkpoint. When enabled, the value model is written to a ``value_model/`` sub-directory inside the
+            checkpoint directory. Useful for inference-time quality estimation, continued critic training, and research.
 
     > [!NOTE]
     > These parameters have default values different from [`~transformers.TrainingArguments`]:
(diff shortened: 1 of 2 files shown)


mukund1985:
Hi @qgallouedec @albertvillanova — tagging you both as the active maintainers in case this is useful.

This PR addresses #3293. The value model (critic) is currently dropped at every checkpoint — only the policy is saved — so any interrupted PPO run has to re-initialise the critic from scratch on resume.

The fix is a single new flag `save_value_model: bool = False` in `PPOConfig`. When enabled, `save_model()` writes the critic to a `value_model/` sub-directory using the existing `super().save_model()` path, so DeepSpeed and FSDP are handled transparently. Default behaviour is completely unchanged.

Happy to add tests or adjust to match whatever pattern you prefer. Thanks!

qgallouedec:
Thanks @mukund1985, we appreciate your contrib.
However PPO is now in maintenance mode. We keep it mostly for ref/legacy, but we won't extend it further. If its blocking for you, the best is probably to fork trl.

PR #6122, opened 2026-06-19
fix(online_dpo): add prediction_step to fix eval_strategy=steps crash

### LINKED ISSUE #2228: online DPO evaluation
### System Info

trl=0.11.2

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder
- [ ] My own task or dataset (give details below)

### Reproduction

for [online dpo code](https://github.com/huggingface/trl/blob/47d08a962679ba0f6ca1b94c0902479525235512/examples/scripts/dpo_online.py#L16-L15)

If I add `--eval_steps=5` and `--eval_strategy=steps`, it shows error:
```
  File "/mnt/task_runtime/trl/examples/scripts/dpo_online.py", line 120, in <module>
    trainer.train()
  File "/opt/miniconda/lib/python3.9/site-packages/transformers/trainer.py", line 1938, in train
    return inner_training_loop(
  File "/opt/miniconda/lib/python3.9/site-packages/transformers/trainer.py", line 2356, in _inner_training_loop
    self._maybe_log_save_evaluate(tr_loss, grad_norm, model, trial, epoch, ignore_keys_for_eval)
  File "/mnt/task_runtime/trl/trl/trainer/online_dpo_trainer.py", line 555, in _maybe_log_save_evaluate
    metrics = self._evaluate(trial, ignore_keys_for_eval)
  File "/opt/miniconda/lib/python3.9/site-packages/transformers/trainer.py", line 2761, in _evaluate
    metrics = self.evaluate(ignore_keys=ignore_keys_for_eval)
  File "/opt/miniconda/lib/python3.9/site-packages/transformers/trainer.py", line 3666, in evaluate
    output = eval_loop(
  File "/opt/miniconda/lib/python3.9/site-packages/transformers/trainer.py", line 3857, in evaluation_loop
    losses, logits, labels = self.prediction_step(model, inputs, prediction_loss_only, ignore_keys=ignore_keys)
  File "/opt/miniconda/lib/python3.9/site-packages/transformers/trainer.py", line 4085, in prediction_step
    outputs = model(**inputs)
  File "/opt/miniconda/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1553, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/opt/miniconda/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1562, in _call_impl
  File "/opt/miniconda/lib/python3.9/site-packages/torch/nn/parallel/data_parallel.py", line 186, in forward
    outputs = self.parallel_apply(replicas, inputs, module_kwargs)
  File "/opt/miniconda/lib/python3.9/site-packages/torch/nn/parallel/data_parallel.py", line 201, in parallel_apply
    return parallel_apply(replicas, inputs, kwargs, self.device_ids[:len(replicas)])
  File "/opt/miniconda/lib/python3.9/site-packages/torch/nn/parallel/parallel_apply.py", line 109, in parallel_apply
    output.reraise()
  File "/opt/miniconda/lib/python3.9/site-packages/torch/_utils.py", line 706, in reraise
    raise exception
TypeError: Caught TypeError in replica 0 on device 0.
Original Traceback (most recent call last):
  File "/opt/miniconda/lib/python3.9/site-packages/torch/nn/parallel/parallel_apply.py", line 84, in _worker
    output = module(*input, **kwargs)
  File "/opt/miniconda/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1553, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/opt/miniconda/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1562, in _call_impl
TypeError: forward() got an unexpected keyword argument 'prompt'
```

### Expected behavior

The eval is not using the correct input dataset.

mukund1985:
## Problem
`OnlineDPOTrainer` crashes with a `TypeError` when `eval_strategy="steps"` (or `"epoch"`) is set.

Root cause: `training_step` handles the raw `"prompt"` string column via its own generation loop. But `prediction_step` is not overridden, so evaluation falls through to `Trainer.prediction_step`, which calls `model(**inputs)` with **all** dataset columns — including the raw `"prompt"` string — causing `model.forward()` to raise:

```
TypeError: forward() got an unexpected keyword argument 'prompt'
```

Fixes #2228.

## Solution
1. Extract the loss-computation body of `training_step` into a shared `_compute_online_dpo_loss(model, inputs)` helper.
2. `training_step` calls the helper then `accelerator.backward()` as before.
3. New `prediction_step` override calls the helper under `torch.no_grad()` (no backward), returning `(loss, None, None)`.

No change to training behaviour.

## Before submitting

## AI writing disclosure

---

> [!NOTE]
> **Medium Risk**
> Touches training/eval paths and generation during eval (costly, nondeterministic) but training backward path is unchanged; subclass eval now fails fast instead of wrong metrics.
> 
> **Overview**
> Fixes **#2228**: `OnlineDPOTrainer` no longer crashes when `eval_strategy` is `"steps"` or `"epoch"`. Eval used to hit the default `Trainer.prediction_step`, which passed raw `"prompt"` strings into `model.forward()`.
> 
> The online DPO forward path (generate → reward → DPO loss) is moved into **`_compute_online_dpo_loss`**, shared by **`training_step`** (unchanged training behavior) and a new **`prediction_step`** that runs that logic under `torch.no_grad()` and returns eval loss. **`prediction_step`** snapshots and restores **`self.stats`** so eval batches do not skew training log metrics.
> 
> **`NashMDTrainer`** and **`XPOTrainer`** override **`prediction_step`** to raise **`NotImplementedError`** with guidance to set `eval_strategy='no'`, so they do not inherit Online DPO eval by mistake.
> 
> Regression tests cover step-based eval completing with **`eval_loss`** logged and **`prediction_step`** leaving **`self.stats`** unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 22952f45b39b33254c37763aa8e405fd584c8b2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mukund1985:
@qgallouedec @albertvillanova — would appreciate a review when you get a chance. Happy to add tests or docs if that's blocking merge.

albertvillanova (changes requested):
The CI to check code quality is failing. Please run: `make precommit`

albertvillanova (changes requested):
The core fix is correct and resolves a real, still-live bug. However, it is incomplete and ships a metric-correctness regression. All three Cursor Bugbot findings are valid (need to be addressed). Not mergeable as-is; needs a subclass fix, a stats fix, and the regression test that was promised on the issue.

One nuance worth noting: on current transformers the exception is no longer the 2024 
> TypeError: forward() got an unexpected keyword argument 'prompt' 

Modern models swallow **kwargs, so it nowsurfaces as 
> ValueError: You must specify exactly one of input_ids or inputs_embeds

Same root cause, different message.

The three Cursor Bugbot findings: all confirmed.

Additional gaps:
- No regression test
- Docstring style. _compute_online_dpo_loss uses bare model:/inputs: without the repo's (`type`) convention, and prediction_step has no Args: block. 

To merge, the PR should:
1. not break NashMD/XPO: either add a per-trainer prediction_step (preferred, matches the self-contained-trainer philosophy) or at minimum guard so the base method isn't silently inherited; 
2. stop eval from mutating self.stats; 
3. add the trainer.evaluate() regression test.

mukund1985:
Thanks for the thorough review @albertvillanova. Addressed all three gaps in the latest push:

**1. NashMD/XPO silently inheriting the wrong prediction_step**
Both `NashMDTrainer` and `XPOTrainer` now define their own `prediction_step` that raises `NotImplementedError` immediately, with a message pointing users to `eval_strategy='no'`. This prevents the base class eval logic (which calls `_compute_online_dpo_loss`, a function neither subclass uses) from running silently with wrong results.

**2. `self.stats` mutation during eval**
`prediction_step` now snapshots `self.stats` before calling `_compute_online_dpo_loss` and restores it after. Eval batches no longer pollute the training-stats accumulators that `_maybe_log_save_evaluate` reads and resets at each log step.

**3. Regression tests**
Added two tests to `tests/experimental/test_online_dpo_trainer.py`:
- `test_eval_strategy_steps_does_not_crash` - end-to-end: trains with `eval_strategy="steps"` and verifies `eval_loss` appears in log history
- - `test_eval_stats_not_polluted_during_eval` - unit-level: calls `prediction_step` directly and asserts `self.stats` lengths are unchanged afterwards
**Docstrings** updated to the repo's `(type)` convention in both `_compute_online_dpo_loss` and `prediction_step`.

mukund1985:
@albertvillanova — just checking in on the Jul 23 push (693083e). All three points from your review should be addressed: NashMD/XPO have their own `prediction_step` raising `NotImplementedError`, `self.stats` is snapshotted and restored around the eval call, and the two regression tests are in. Happy to address anything else you spot.

albertvillanova (changes requested):
Thanks for the follow-up @mukund1985, and sorry for the delay. I pushed a `make precommit` commit and merged `main` into the branch so CI is green now (code quality + experimental tests pass).

I re-reviewed the whole thing and also ran it locally. The good news: **the fix works.** The remaining news: it reinvents a pattern that already exists in TRL, and one of the three Bugbot findings is still open.

## 1. Blocking: this duplicates GRPO/RLOO instead of copying them

`GRPOTrainer` and `RLOOTrainer` already solved exactly this problem — a generation-based trainer that needs `prediction_step`. Both ship, verbatim:

```python
def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=None):
    if return_outputs:
        raise ValueError("The GRPOTrainer does not support returning outputs")
    return self._compute_loss(model, inputs)

def prediction_step(self, model, inputs, prediction_loss_only, ignore_keys: list[str] | None = None):
    with torch.no_grad():
        with self.compute_loss_context_manager():
            loss = self.compute_loss(model, inputs)
        loss = loss.mean().detach()
    return loss, None, None
```

(`trl/trainer/grpo_trainer.py:2926` / `:3283`, `trl/trainer/rloo_trainer.py:1666` / `:1742`.)

`AGENTS.md` is explicit that duplication is accepted but **consistency is mandatory** — same names, same control flow, same comments. Please mirror the above rather than introduce a parallel spelling:

- Rename `_compute_online_dpo_loss` → **`_compute_loss`** (`online_dpo` is redundant inside `OnlineDPOTrainer`, and `_compute_loss` is the established name in GRPO, RLOO and DPO), and add the thin `compute_loss` wrapper.
- Use `with self.compute_loss_context_manager():` inside the `no_grad` block, and `loss.mean().detach()`. Today neither path uses the context manager, so copying GRPO's version brings online DPO in line rather than changing behaviour.
- Drop `model.eval()` at line 1348 — `Trainer.evaluation_loop` already puts the model in eval mode, GRPO/RLOO don't do it, and here it's never undone (it only self-heals because `training_step` calls `model.train()`).
- `self._prepare_inputs(inputs)` at line 1347 is correct and matches GRPO — but note `training_step` doesn't call it, so the two paths are now asymmetric. Fine to leave as-is since you're copying GRPO, just flagging it.

This should make `prediction_step` about 6 lines and let you delete most of the new docstring.

## 2. Blocking: Bugbot finding #3 is still open

Your comment says all three were addressed, but the `torch_empty_cache_steps` one wasn't — the block is still inside the shared helper at `trl/experimental/online_dpo/online_dpo_trainer.py:1290-1294`. Because `self.state.global_step` is constant across an evaluation, if `global_step % torch_empty_cache_steps == 0` at the step where eval fires, `empty_cache()` now runs on **every eval batch** instead of once per training step. Move those five lines back into `training_step`, which is where transformers puts them (and where GRPO/RLOO get them from, via `super().training_step`).

## 3. Eval pays for generation and reports a single number

This is the one I'd most like your opinion on. In my run above, evaluation cost **12.7 s out of a 34.5 s** 4-step training run — it does a full generation + reward-model pass — and the snapshot/restore then discards all 14 online-DPO metrics, so the only thing surfaced is `eval_loss`.

GRPO/RLOO handle this with the `mode` pattern instead of snapshot/restore.

## 4. NashMD/XPO guard: right call, wrong shape

The guard is justified — I checked, and it's not just "different loss": `NashMDTrainer` sets `self.reward_funcs = self.reward_funcs[0]` (a single object, not a list) and its `self.stats` has no `objective/kl`, `objective/non_score_reward`, `val/contain_eos_token`, `objective/rlhf_reward`, `objective/scores*` keys, so the inherited path would `KeyError`/`TypeError`. XPO likewise. So thank you for adding it.

Two changes:

- **Trim it.** A 10-line Sphinx docstring wrapped around a one-line `raise` is a lot of surface for two experimental trainers; `AGENTS.md` asks for leanness. A short `raise NotImplementedError(...)` with the message you already wrote is enough — the message *is* the documentation.
- **Fail fast.** Raising from `prediction_step` means the user loses the run at step N. Please also reject it at construction, e.g. in `NashMDTrainer.__init__`/`XPOTrainer.__init__`:
  ```python
  if args.eval_strategy != "no":
      raise ValueError("Evaluation is not supported for NashMDTrainer. Set `eval_strategy='no'`.")
  ```
  Keep the `prediction_step` raise too, so a direct `trainer.evaluate()` call is still covered.

## 5. Docstring style

The repo format is types in backticks with the description on the **next** line, and `[`~transformers.Trainer.prediction_step`]` for cross-references — no Sphinx roles. Right now this PR introduces 7 of the 8 `:meth:` occurrences in the whole of `trl/`, so it's a new style, not an existing one. Please convert (and per §1 most of it can just be deleted — GRPO/RLOO's `prediction_step` has no docstring at all; a one-line comment with the issue link is plenty).

## 6. Tests

Close, but:

- `test_eval_stats_not_polluted_during_eval` passes `max_steps=2`, `eval_strategy="steps"`, `eval_steps=1` and then never calls `train()` — dead config that reads as if training happens. Drop those three args.
- Naming: the repo convention is `test_train_with_eval` (see `tests/test_grpo_trainer.py:430`). Prefer `test_train_with_eval_strategy_steps` over `..._does_not_crash`.
- The two tests each build a full trainer; only the first needs `train()`. Consider folding them so the experimental job (already ~12 min) doesn't grow more than necessary.
- **Missing:** no test for the new NashMD/XPO behaviour. Please add a `pytest.raises(NotImplementedError)` (or `ValueError`, per §4) around `trainer.evaluate()` in `tests/experimental/test_nash_md_trainer.py` and `test_xpo_trainer.py`. Those need no training, so they're cheap.
- Worth adding `eval_on_start=True` to one of them — GRPO has a dedicated regression test for it (`test_train_eval_on_start`) because eval-before-first-step has bitten us there. I confirmed it works here, so let's lock it in.

## 7. Minor

- Docs unchecked in the PR template. A two-line note in `docs/source/online_dpo_trainer.md` that evaluation generates completions (so it's slow and stochastic, and reports only `eval_loss`), plus a line in `nash_md_trainer.md` / `xpo_trainer.md` that evaluation is unsupported, would close the loop for users hitting #2228.
- `prediction_step` returns `None` for logits/labels, so a user-supplied `compute_metrics` is silently ignored. Same as GRPO/RLOO, so consistent — not asking for a change, just noting it.
- If you have a 2-GPU box handy, a quick `eval_strategy="steps"` sanity run would be reassuring, since the reporter in #2228 was multi-GPU and `_compute_loss` does `gather_for_metrics` collectives that now also fire on the eval dataloader.

qgallouedec:
Closing as no recent update, to keep the queue tidy, happy to re-open if needed

PR #6123, opened 2026-06-19
feat(dpo): implement Adaptive Beta-DPO (arXiv:2407.08639)

### LINKED ISSUE #5211: Adaptive Beta-DPO
### Feature request

Add **adaptive β strategies** to `DPOTrainer`, starting with **β-DPO (Dynamic β)**.

Paper: https://arxiv.org/pdf/2407.08639 

The proposal is to allow per-batch adaptive scaling of β instead of using a single global value throughout training.

Suggested addition to `DPOConfig`:

```python
adaptive_beta: Optional[str] = None                   # e.g. "beta-dpo"
beta_alpha: Optional[float] = None                     # scaling factor (α)
beta_reference_margin: Optional[float] = None  # optional M0 override
```

When enabled (e.g. `adaptive_beta="beta-dpo"`), β is dynamically adjusted at the batch level:

β_batch = [1 + α (mean(M_batch) − M₀)] β₀

where:
- `β₀` = base DPO β  
- `M_batch` = per-pair margins in the current batch (already computed in the DPO loss)  
- `M₀` = reference margin (running mean or fixed value)

**Why not implement it as a supported `loss_type`?**

As shown in the paper, **β-DPO (Dynamic β)** can be combined with other DPO variants such as SPPO and IPO, since it modifies the scaling parameter β rather than the loss formulation itself.  

It is therefore orthogonal to the specific DPO loss variant and should be implemented as a configurable training strategy (e.g. `adaptive_beta`) instead of a separate `loss_type`. This keeps the design modular and avoids duplicating loss implementations.

### Motivation

In standard DPO, β is fixed across all preference pairs. However, β-DPO shows that optimal β depends on the quality gap between preferred and dispreferred responses:
- Low-gap pairs (ambiguous preferences) benefit from smaller β (stronger updates).
- High-gap pairs (clear preferences) benefit from larger β (more conservative updates).

Real-world datasets contain heterogeneous pair quality and outliers. A single global β can therefore be suboptimal and often requires costly hyperparameter sweeps.

Adaptive β:
- Improves robustness to mixed-quality data
- Reduces sensitivity to β tuning
- Stabilizes training under noisy preference signals

Complexity & Cost
- Compute/Memory overhead: low, compute the average margin and scalar tensor opeartions to compute β_batch 
- No extra models

Fully compatible with current DPO training loop and, as showed in the paper, it can be combined with other losses such as IPO and SPPO. Implementation is localized to the DPO loss and does not affect the rest of the trainer stack.

### Your contribution

I’m willing to open a PR implementing this feature following the contribution guidelines.
The implementation would:
- Add the new config fields
- Modify the DPO loss to support per-batch β
- Include unit tests and documentation updates

Happy to align on design decisions before submitting.

mukund1985:
## Problem
A fixed `β` in DPO doesn't adapt to how well-separated chosen/rejected responses are in the current batch. The β-DPO paper (arXiv:2407.08639) shows that per-batch adaptive β improves alignment stability and final policy quality.

Closes #5211.

## Solution
Implements the β-DPO algorithm orthogonally to `loss_type` — every loss type (sigmoid, IPO, SPPO, robust, …) benefits automatically.

**Algorithm:**
```
M_batch = mean(chosen_logratios − rejected_logratios)   # current batch margin
M₀      = 0.9·M₀ + 0.1·M_batch                         # EMA reference (updated each step)
effective_β = max([1 + α(M_batch − M₀)] · β₀, 1e-6)    # clipped to stay positive
```

## New config fields

| Field | Type | Default | Description |
|---|---|---|---|
| `adaptive_beta` | `str \| None` | `None` | Set to `"beta-dpo"` to enable |
| `beta_alpha` | `float \| None` | `None` | Scaling factor α (required when enabled) |
| `beta_reference_margin` | `float \| None` | `None` | Fixed M₀; `None` = use EMA |

## Usage

```python

config = DPOConfig(
    adaptive_beta="beta-dpo",
    beta_alpha=0.5,
    beta=0.1,  # β₀ base value
)
```

A `beta/effective` metric is logged at each training step when `adaptive_beta` is set.

## Before submitting

## AI writing disclosure

---

> [!NOTE]
> **Medium Risk**
> Opt-in changes to the DPO loss and β used in training across every loss variant; default behavior is unchanged, but misconfiguration or distributed margin bugs could affect alignment stability.
> 
> **Overview**
> Adds optional **β-DPO** per-batch adaptive scaling to `DPOTrainer`, controlled by new `DPOConfig` fields: `adaptive_beta="beta-dpo"`, required `beta_alpha`, and optional fixed `beta_reference_margin` (otherwise an EMA of batch margins).
> 
> During **training**, the trainer computes a global batch margin from chosen vs rejected log-ratios (gathered across DDP ranks), updates the reference margin EMA once per optimizer step when not fixed, and sets `effective_beta = max([1 + α(M_batch − M₀)] × β₀, 1e-6)`. That value replaces static `beta` in **all** supported `loss_type` branches, reward metrics, and logs **`beta/effective`**. Init rejects unknown `adaptive_beta` values, missing `beta_alpha`, and **`use_liger_kernel=True`** (static fused β).
> 
> Checkpoint **resume** persists/restores the EMA via `adaptive_beta_state.json` when using the running margin. Eval keeps base `beta` (adaptive path is train-only). A small Liger path fix adds `.contiguous()` on `shift_completion_mask`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9c5b0b57b61de13fcee84e0680545dd203898e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mukund1985:
Hi @qgallouedec — I saw you were tagged on #5211, so pinging you here.

This PR implements Adaptive Beta-DPO from arXiv:2407.08639. The key idea: rather than a fixed `β`, the paper shows that scaling β per-batch based on how well-separated the chosen/rejected responses already are leads to better alignment stability.

The algorithm:
```
M_batch = mean(chosen_logratios − rejected_logratios)   # current batch margin
M₀      = 0.9·M₀ + 0.1·M_batch                         # EMA reference margin
effective_β = max([1 + α(M_batch − M₀)] × β₀, 1e-6)
```

Three new `DPOConfig` fields: `adaptive_beta` (set to `"beta-dpo"` to enable), `beta_alpha` (the α scaling factor), and `beta_reference_margin` (optional fixed M₀; `None` = use EMA). The `effective_beta` replaces `self.beta` in the loss computation so it works automatically with all `loss_type` variants (sigmoid, IPO, SPPO, robust, etc.). A `beta/effective` metric is logged when enabled.

Happy to add tests, docs, or adjust the implementation style to match the rest of the codebase. Thanks!

mukund1985:
Thanks for the Cursor Bugbot review — all 3 issues were valid, fixed in 647dd0c.

**Bug 1 (Medium) — Fixed reference margin overwritten**
When `beta_reference_margin` is set, `_running_margin` was still updated by the EMA every step, making the supposedly-fixed M₀ drift. Fixed by guarding the EMA update behind `if self.args.beta_reference_margin is None` — when the user provides an explicit M₀ it never changes.

**Bug 2 (High) — Liger path ignores adaptive beta**
`_compute_loss_liger` routes to `LigerFusedLinearDPOLoss` which has a static β baked into the kernel — there's no way to pass a per-step `effective_beta` through it. Rather than silently having no effect, we now raise a `ValueError` at `__init__` time when both `adaptive_beta` and `use_liger_kernel=True` are set, so the user gets a clear error immediately.

**Bug 3 (High) — Per-rank margin diverges under DDP**
`batch_margin` was computed from each GPU's local micro-batch only. Fixed by gathering across all ranks with `self.accelerator.gather(local_margin).mean().item()` before the EMA update, so every process uses the same global margin and applies an identical `effective_beta`.

mukund1985:
Thanks again — valid catch, fixed in 0ee75c2.

**Bug 4 (Medium) — EMA margin lost on resume**

`_running_margin` was updated every step but never written to or read from checkpoints. Resuming training re-bootstrapped M₀ from the first post-resume batch, so `effective_beta` no longer matched the pre-checkpoint trajectory.

Fix in 0ee75c2:
- `_save_checkpoint` now writes `adaptive_beta_state.json` into the checkpoint folder alongside the model/optimizer state. This only applies when `beta_reference_margin is None` (the EMA path) — users with a fixed M₀ don't need it since `_running_margin` is static.
- After `super().__init__()` we check whether `resume_from_checkpoint` points to a directory containing that file, and restore `_running_margin` from it if so.

mukund1985:
All 5 issues fixed in 06488cf.

**Bug 5 (High) — Removed evaluate preprocessing override**
This was a regression in our file — we accidentally dropped the `DPOTrainer.evaluate` override that tokenizes raw eval datasets and precomputes ref log-probs when `precompute_ref_log_probs=True`. Fixed by rebuilding the PR on top of the current upstream rather than our broken base, so the override is fully restored.

**Bug 6 (High) — Config trust_remote_code ignored**
Same root cause as Bug 5 — our file dropped `trust_remote_code` from `create_model_from_path`, `AutoProcessor.from_pretrained`, and ref model loading. Restored.

**Bug 7 (High) — Boolean resume path crashes init**
`os.path.isdir(True)` raises `TypeError` when `resume_from_checkpoint=True`. Added `isinstance(resume_from_checkpoint, str)` guard before the `os.path.isdir` call.

**Bug 8 (Medium) — EMA margin not restored on resume**
Moved EMA state loading from `__init__` into a `_load_from_checkpoint` override. That method is called in both the `args.resume_from_checkpoint` path and the `trainer.train(resume_from_checkpoint=...)` path, so EMA state is now restored in all resume scenarios.

**Bug 9 (Medium) — EMA updates each microbatch**
Gated the EMA update on `self.accelerator.sync_gradients`, which is `True` only on the last microbatch of each gradient-accumulation group. This ensures `_running_margin` advances exactly once per optimizer step, matching the β-DPO paper's per-batch formulation.

mukund1985:
@qgallouedec @albertvillanova — would appreciate a review when you get a chance. Happy to add tests or docs if that's blocking merge.

qgallouedec:
thanks a lot for the PR, we've decided not to support this method for now in DPOTrainer, see https://github.com/huggingface/trl/issues/5211#issuecomment-5145789856

PR #6126, opened 2026-06-20
Add X-Token cross-tokenizer KD to GOLD trainer

kashif:
Adds X-Token cross-tokenizer knowledge distillation to `GOLDTrainer`, as introduced in https://huggingface.co/papers/2605.21699.

Two new loss variants sit alongside the existing ULD path and are activated via `GOLDConfig.xtoken_loss_type`:

- `"p_kl"` (Projection KL): projects the full student distribution to the teacher vocab via a sparse matrix W, then computes forward KL. Recovers signal for tokens that land in the uncommon set (e.g. multi-digit numerals that split differently across tokenizer families).
- `"h_kl"` (Hybrid KL): retains the GOLD hybrid structure but builds the common set via the top-1 mapping under W (threshold >= 0.6) instead of strict string equality. Applies forward KL on common tokens and sorted-L1 on uncommon tokens.

Both paths apply T^2 gradient scaling and optionally a stop-gradient dynamic CE/KD balance (`xtoken_dynamic_scaling=True`).

New files/classes:
- `_Fp32SparseMM`: custom autograd function that keeps sparse matmul in FP32 under BF16 autocast (PyTorch has no BF16 sparse-mm kernel).
- `_load_sparse_projection_matrix` / `_load_exact_token_map`: lazy per-process caches for the precomputed projection matrix.
- `XTokenLoss(nn.Module)`: full loss implementation with `_compute_p_kl` and `_compute_h_kl`.
- `paper_index.md`: new file tracking paper implementations per CLAUDE.md requirement.

Token-span alignment reuses `ULDLoss._align_by_byte_offsets`.

All 25 existing GOLD trainer tests pass.

---

> [!NOTE]
> **Medium Risk**
> Changes the GOLD training and loss path for cross-tokenizer setups (dual models, projection matrices, new hyperparameters); scope is experimental GOLD with substantial test coverage but real training behavior is easy to misconfigure.
> 
> **Overview**
> Adds **X-Token** off-policy cross-tokenizer KD to `GOLDTrainer` / `GOLDConfig` (`xtoken_loss_type`: `p_kl` or `h_kl`), using a precomputed sparse projection matrix and separate teacher tokenizer—parallel to ULD, not combinable with ULD or the Liger JSD path.
> 
> Training wires dual forward passes, byte-offset alignment (same collator path as extended ULD), `XTokenLoss` (P-KL projection + global top-k KL, or H-KL common/uncommon split), FP32 sparse projection matmul under autocast, optional dynamic CE/KD scaling, and xtoken training metrics. Config enforces `lmbda=0.0` when X-Token is enabled and skips teacher vocab resize.
> 
> Ships **`examples/xtoken/`** scripts to build/trim projection matrices and run distillation, plus docs: new `xtoken` example row, **Cross-Tokenizer Knowledge Distillation** in `paper_index.md`, and an **IW-OPD** `DistillationConfig` snippet. Tests cover loss helpers, VLM cross-tokenizer setup, and slow end-to-end train steps for both loss types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce650fe117917d4332d9fb4228703c0982b79fbb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kashif:
Screenshot 2026-07-14 at 10 09 22
Sample run

trl/experimental/gold/gold_trainer.py
```diff
@@ -778,7 +778,435 @@ def _get_start_and_size_answers(self, answer_tensors):
...
+            valid_pairs_total += len(paired)
+            if self.loss_type == "p_kl":
+                kd_i, p_num, p_den = self._compute_p_kl(s_logits_i, t_logits_i, paired, device, T)
+                proj_acc_num += p_num
+                proj_acc_den += p_den
+            else:
+                kd_i = self._compute_h_kl(s_logits_i, t_logits_i, paired, device, T)
+            kd_terms.append(kd_i)
+
+        kd = torch.stack(kd_terms).mean()
```
albertvillanova, line 1080: This returns a plain per-microbatch mean and `XTokenLoss.forward` never takes `num_items_in_batch`. The JSD path (`generalized_jsd_loss`, via `compute_loss`) and the Liger path both normalize by `num_items_in_batch`. Because GOLD doesn't set `model_accepts_loss_kwargs=False`, transformers skips the `/gradient_accumulation_steps` division and trusts `compute_loss` to normalize globally; so under GA>1 (the example defaults to `gradient_accumulation_steps=4`) the X-Token gradient is ~GA× too large. Commit 17c74172 added this scaling and it was lost in a merge. Please thread `num_items_in_batch` through and normalize as the JSD path does.

trl/experimental/gold/gold_trainer.py
```diff
@@ -778,7 +778,435 @@ def _get_start_and_size_answers(self, answer_tensors):
...
+        self.last_ce_loss = ce.detach()
+        self.last_accuracy_num = acc_num
+        self.last_accuracy_den = acc_den
+        self.last_num_valid_pairs = float(valid_pairs_total) / batch_size
+        self.last_proj_accuracy_num = proj_acc_num
+        self.last_proj_accuracy_den = proj_acc_den
+
+        if self.dynamic_scaling:
+            scale = (ce.detach() / kd.detach().clamp(min=1e-8)).clamp(0.01, 100.0)
+            return scale * kd + self.ce_scale * ce
```
albertvillanova: `ce_scale` is applied here in the `dynamic_scaling=True` branch, but it's documented as "Only used when `xtoken_dynamic_scaling=False`" (gold_config.py:135, :459), and the dynamic-scaling formula is documented as paper Eq. 7 "…before adding CE" (coefficient 1.0). Since the dynamic scale makes `scale·kd ≈ ce`, the effective objective here is `≈ CE·(1 + ce_scale)`, so the defaults (`dynamic_scaling=True`, `ce_scale=0.1`) silently weight CE at 0.1, not 1.0. Should this be `+ ce` (matching the docs/paper), or should the three doc sites be corrected? This changes the default training objective.

trl/experimental/gold/gold_config.py
```diff
@@ -359,6 +381,86 @@ class GOLDConfig(SFTConfig):
...
+                "Weight for the X-Token KD loss relative to the CE loss. Only used when "
+                "`xtoken_dynamic_scaling=False`."
+            )
+        },
+    )
+    xtoken_ce_scale: float = field(
+        default=0.1,
+        metadata={
+            "help": (
+                "Scale factor for the CE term in the X-Token loss. Only used when `xtoken_dynamic_scaling=False`."
```
albertvillanova, line 458: As mentioned above, this is inaccurate: `ce_scale` is used in both branches of XTokenLoss.forward (gold_trainer.py:1091), including the default dynamic_scaling=True. Fix the code or this docstring (and the field help at :459, plus the class docstring at :135)

trl/experimental/gold/gold_config.py
```diff
@@ -529,3 +631,26 @@ def __post_init__(self):
                         raise ValueError("uld_hybrid_matched_weight must be non-negative.")
                     if self.uld_hybrid_unmatched_weight < 0.0:
                         raise ValueError("uld_hybrid_unmatched_weight must be non-negative.")
+
+        # Validate X-Token parameters
+        _valid_xtoken_types = ("none", "p_kl", "h_kl")
+        if self.xtoken_loss_type not in _valid_xtoken_types:
+            raise ValueError(f"xtoken_loss_type must be one of {_valid_xtoken_types}, got '{self.xtoken_loss_type}'.")
+        if self.xtoken_loss_type != "none":
```
albertvillanova, line 638: The PR/paper_index.md state X-Token is off-policy only (lmbda=0), but nothing enforces it and the default is lmbda=0.5, so half the batches would be on-policy student rollouts down an untested path. Consider adding a raise/warn here when `xtoken_loss_type != "none"` and `lmbda != 0` (kept minimal per the repo's anti-defensive-code stance).

albertvillanova (commented):
Thanks. 

Some concerns below: the problems are concentrated in behavioral defaults that contradict the PR's own documentation, thin test coverage of exactly those defaults, and a memory regression + scope creep in shared vLLM infra (I saw you reverted these changes while I was writing my review).

The loss module is thoughtfully written and the math checks out. The blockers are the two default-objective issues ((1) GA normalization, (2) ce_scale in the dynamic branch), both silently contradict the PR's own docs and are exercised by the shipped example; plus the untested default paths (4).

### 1. HIGH: X-Token loss ignores num_items_in_batch; gradients silently mis-scale by ~`gradient_accumulation_steps`×

XTokenLoss.forward returns a plain per-microbatch .mean() and never accepts num_items_in_batch (gold_trainer.py:1078-1092); the compute_loss X-Token block doesn't rescale it either (:3112-3119). The JSD path (generalized_jsd_loss, :3050-3056) and the Liger path (:3020-3024) in the same method both normalize by num_items_in_batch.

Why this matters: GOLD does not set self.model_accepts_loss_kwargs = False (unlike DPO/RLOO/KTO/GRPO, which do so precisely to re-enable the /GA division). With a standard student CausalLM, model_accepts_loss_kwargs=True, and transformers 5.14 then skips the /gradient_accumulation_steps division, trusting compute_loss to have normalized globally:
```python
# transformers/trainer.py training_step
if (not self.model_accepts_loss_kwargs or num_items_in_batch is None) and self.compute_loss_func is None:
    loss = loss / self.current_gradient_accumulation_steps
```
The JSD path honors that contract; the X-Token path doesn't → under GA the X-Token gradient is ~GA× too large, so effective LR silently depends on GA.

### 2. HIGH: xtoken_ce_scale is applied in the dynamic_scaling=True branch, contradicting its docs and the cited paper
```python
# gold_trainer.py:1089-1092
if self.dynamic_scaling:
    scale = (ce.detach() / kd.detach().clamp(min=1e-8)).clamp(0.01, 100.0)
    return scale * kd + self.ce_scale * ce      # ce_scale used here
return self.kl_weight * kd + self.ce_scale * ce
```
But xtoken_ce_scale is documented as "Only used when xtoken_dynamic_scaling=False" in both the class docstring (gold_config.py:135-136) and field help (:455-461), and the xtoken_dynamic_scaling help says the formula is paper Eq. 7 "…before adding CE" (:416-424) — CE coefficient 1.0, not ce_scale.

### 3. YOU REMOVED the changes on vLLM

### 4. MEDIUM: No trainer-level X-Token test; the default paths are entirely unexercised

xtoken_projection_matrix_path appears in the whole test suite exactly once (test_gold_trainer.py:1623), inside TestXTokenLoss.test_xtoken_loss, which:
- tests XTokenLoss in isolation and asserts only torch.isfinite(loss) — no numeric correctness;
- sets xtoken_dynamic_scaling=False (default is True → finding #2's path never runs);
- sets uld_skip_student_eos=False/uld_skip_teacher_eos=False (both default True).

There is no test that constructs GOLDTrainer with xtoken_loss_type != "none", so the _cross_tok teacher-input build, dataset prep with a real matrix, the off-policy byte_offsets checks, and the xtoken/* metrics — which use accelerator.gather on 0-dim tensors (:3123-3149), a pattern found nowhere else in the trainer and untested under multi-GPU where 0-dim gathers are a known footgun — are all unverified. JSD/ULD/Liger each have a train-step smoke test; X-Token has none. Recommended: a tiny-model p_kl/h_kl train-step smoke test with dynamic_scaling=True, and a numeric assertion (identical logits + identity projection ⇒ kd≈0, proj_accuracy==1.0).

### 5. MEDIUM:  "off-policy only (lmbda=0)" is documented but not enforced; default is lmbda=0.5

The PR body and paper_index.md state X-Token is off-policy (lmbda=0), but GOLDConfig.__post_init__ (gold_config.py:635-656) never validates it. A valid GOLDConfig(xtoken_loss_type="h_kl", ...) keeps lmbda=0.5, so half the batches would be on-policy student rollouts down an untested path (_maybe_add_completion_byte_offsets even builds on-policy offsets, :1937-1964). The example hard-codes lmbda=0.0, underlining that the default is wrong for this method. Either raise/warn when xtoken_loss_type != "none" and lmbda != 0, or drop the "only" claim. (I'd keep this lightweight given the repo's anti-defensive-code stance — a single validation in __post_init__.)

### 6. LOW: projection-script and matrix-pipeline nits

- Scale-trick shifts the h_kl partition. build_projection_matrix.py --enable-scale-trick (default True) sets top_likelihoods[:,-1]=0.2 and renormalizes; _load_exact_token_map thresholds on sorted_vals[:,0] >= 0.6 (gold_trainer.py:892) over those post-trick values, so the scale trick changes which tokens fall in the KL-common vs L1-uncommon set. Worth verifying against the reference and documenting.
- Dead/fragile strict branch. _load_exact_token_map's xtoken_loss=False path (:893-894, :921-922) is unreachable — _exact_map always passes xtoken_loss=(loss_type=="h_kl")==True (:995) — and it indexes top_indices[:,1], assuming ≥2 columns. Remove or guard.
- Trim inconsistency. build_projection_matrix.py:331 auto-runs sort_and_cut(...) without preserve_last=True, while the standalone script auto-detects it from metadata — the two entry points diverge with the scale trick on.
- Import fragility. build_projection_matrix.py:328 from sort_and_cut_projection_matrix import sort_and_cut only resolves when CWD is the script dir.
- xtoken_vocab_topk guidance is questionable. The docstring says it "should not exceed the --top-k used when building the matrix" (gold_config.py:129-132), but the top-k is selected over the full teacher distribution by teacher importance, independent of W's sparsity — arguably you want teacher-important tokens with no projected mass in a forward KL. Default 8192 vs. build/example defaults (32/4) means users following the example violate the stated rule by default.

### 7. Nits

- _load_exact_token_map's hardcoded 0.6 threshold (:892) isn't configurable (fine for experimental, but note it).
- In-function import os in both loaders (:820, :873) while the file imports at top elsewhere.
- Teacher forward in the _cross_tok branch omits use_cache=False (:2959) while the student passes it (:2953) — harmless under no_grad, inconsistent (pre-existing, shared with ULD).
- paper_index.md links the GOLD space as .../on-policy-distillation while the trainer citation uses .../general-on-policy-logit-distillation (gold_trainer.py:1221) — pick one.

trl/experimental/gold/gold_trainer.py
kashif: Good catch, this was indeed lost in the main merge. Restored in 1f50dd63, same normalization as the ULD path.

trl/experimental/gold/gold_trainer.py
kashif: Fixed in 1f50dd63: dynamic scaling now returns scale * kd + ce, matching the paper and the docs. kl_weight and ce_scale only apply when dynamic scaling is off, which is also what the NeMo-RL reference does (it ignores both knobs in the dynamic branch).

trl/experimental/gold/gold_config.py
kashif: Fixed the code side in 1f50dd63, the docstrings are accurate now.

trl/experimental/gold/gold_config.py
kashif: Added in 1f50dd63: GOLDConfig now raises when xtoken_loss_type != 'none' and lmbda != 0.

kashif:
Addressed the remaining review points:

- (4) added a numeric XTokenLoss test (identity projection + identical logits gives kd ~ 0 and proj_accuracy == 1.0, running the default dynamic_scaling=True and skip_eos paths) and a trainer-level p_kl train-step smoke test with grad accumulation (a76698d9)
- (6) removed the unreachable strict branch in _load_exact_token_map, made build_projection_matrix's --runtime-top-k path auto-enable preserve_last like the standalone script, fixed the CWD-dependent sibling import, documented the scale-trick effect on the H-KL partition (verified the >= 0.6 threshold over post-trick weights matches the reference), and reworded the vocab_topk guidance to describe the interaction instead of prescribing a bound (ad69b818, 6a4db70a)
- (7) moved the in-function os imports to the top, noted the hardcoded 0.6 threshold in the docstring, and aligned the GOLD space URL to huggingface.co/spaces/HuggingFaceH4/on-policy-distillation. The teacher-forward use_cache asymmetry is pre-existing and shared with the ULD branch, so I left it for a separate sweep.

paper_index.md now also shows the full pipeline commands (build_projection_matrix.py with --runtime-top-k, then examples/scripts/xtoken.py).

albertvillanova (commented):
Re-reviewed against `60005e0c`.

All seven findings from my previous review are properly addressed — I checked each one against the branch and the code matches your replies. Two small corrections on my side: the `num_items_in_batch` rescale follows the Liger block rather than the ULD path (the ULD path does no rescale at all), which is the right pattern; and my worry about 0-dim `accelerator.gather` under multi-GPU was unfounded — `accelerate._gpu_gather` promotes 0-dim tensors itself.

The PR isn't mergeable as-is, though, for a reason unrelated to those findings.

## Blocker: the branch is 242 commits behind and now duplicates `main`

The tip commit is `add iw-opd distillation (#6191)`. That work landed in `main` as `467c48a6` and was then moved out by the `[DistillationTrainer refactor]` series: IW-OPD now lives in `trl/experimental/iw_opd/`, and `DistillationTrainer` was promoted to the stable API in `trl/trainer/distillation_trainer.py` (#6640). So roughly 40% of the current diff is stale duplicate — `distillation_trainer.py` (+136), `distillation_config.py` (+47), `test_distillation_trainer.py` (+202), `distillation_trainer.md` (+18).

The good news is that the X-Token work itself rebases cheaply: `trl/experimental/gold/` has moved by exactly one 5-line docstring commit since your branch point, and a test-merge of `upstream/main` produces only 7 conflict hunks, all in the IW-OPD half (`iw_opd_config.py`, `test_iw_opd_trainer.py`, `distillation_trainer.md`); rename detection resolves `iw_opd_trainer.py` to `main`'s version verbatim.

Could you merge `main` and drop the IW-OPD content entirely, so the diff is X-Token only?

## New findings

**1. MEDIUM — `h_kl`'s kd term can go negative, and the dynamic scale then flips its sign.**

`_compute_h_kl` returns `(kl_common + l1_uncommon) * T²`, where `kl_common` is a forward KL over a *restricted, non-renormalized* index set (`gold_trainer.py:1155-1159`):

```python
kl_per_chunk = F.kl_div(
    s_chunks[:, common_s], t_chunks[:, common_t], reduction="none", log_target=True
).sum(dim=-1)
```

That's a partial sum of `exp(t_j) · (t_j - s_j)` over the common subset only, so it isn't bounded below by 0. In the default branch (`:1062`):

```python
scale = (ce.detach() / kd.detach().clamp(min=1e-8)).clamp(0.01, 100.0)
return scale * kd + ce
```

a negative `kd` clamps the *denominator* to `+1e-8`, so `scale` saturates at `+100` and `scale * kd` becomes a large negative term that the optimizer will drive further negative.

`p_kl` is safe here (both sides are renormalized over the shared top-k, so its KL is a proper one), and ULD's hybrid is safe too (JSD on matched tokens). Worth either clamping on magnitude or confirming the reference guarantees `kl_common >= 0` — and note this is the one path with no trainer-level test.

**2. LOW — `num_valid_local` is counted on a different basis than `num_items_in_batch`.**

`:3111` counts `(xtoken_student_labels != -100)` over *unshifted* labels with pad ids additionally masked, while transformers counts `labels[..., 1:]` and the sibling Liger block correctly uses the shifted `true_labels`. `skip_eos` drops one more token per sequence on top. The residual is a small constant factor rather than a GA-dependent one, so it's only a nit — but it's gratuitous divergence from the block right above it.

**3. LOW — the `xtoken/*` metrics are the only `accelerator.gather` calls in the trainer.**

Eight collectives per microbatch at `:3119-3141`, where every other metric in this trainer accumulates locally and does a single packed `dist.all_reduce` in `log()` (`:3335-3355`). The values come out right; it's a consistency and per-step-overhead point, and AGENTS.md's duplication rules make the established pattern the one to follow.

**4. LOW — dead code now that `lmbda=0` is enforced.**

`_maybe_add_completion_byte_offsets`'s `self.xtoken_loss_fn is not None` arm (`:1903`) is reachable only from the on-policy generation paths, which `__post_init__` now forbids. Same for the `xtoken_loss_fn = None` triplication — class attribute at `:1180`, pre-`super()` at `:1411`, post-`super()` at `:1478`. The class attribute exists only so `__new__`-built test stubs are safe, and those stubs already set it explicitly (`test_gold_trainer.py:487`, `:637`, `:703`), so it can go.

**5. LOW — `build_projection_matrix.py:249` re-imports `defaultdict as _dd` in-function** though it's already imported at module top (`:41`) — same class of nit as the `import os` one you fixed.

**6. LOW — the `paper_index.md` command won't resolve as written.**

It shows `--projection-matrix cross_tokenizer_data/projection_map_..._top_4_sorted.pt`, but the builder writes `<stem>_top{runtime_top_k}.pt` (no `_sorted`), and the stem embeds `--top-k` (default 32), not `--runtime-top-k`. The Python snippet just below uses a third, different filename.

## Test coverage and scope

The X-Token test gaps I raised are addressed, with two residuals: `test_xtoken_p_kl_train_step_smoke` is `@pytest.mark.slow` with `bf16=True`, so nothing exercises the trainer wiring in fast CI (consistent with the neighbouring VLM smoke tests, so I'm fine with it), and `h_kl` has no trainer-level test at all — which is the same path as finding 1 above.

Separately: the test-file reorganization (1626+/1138-, folding ~6 top-level tests into `test_init` / `test_truncation` / `test_chatml_collator`) does preserve the assertions it consolidated, as far as I could check. But it's unrelated churn that makes this diff considerably harder to review, and it would be better as its own PR.

qgallouedec:
@kashif @albertvillanova backing Albert's review here, I don't have much to add to it:
- the iw-opd duplication is the blocker, and the h_kl sign issue is the one I'd want settled before merge, especially since it's the path with no trainer-level test.
- two more for the rebase: `examples/scripts/xtoken.py` targets a directory that doesn't exist any more, it needs to become `examples/xtoken/xtoken.py` with a row in `example_overview.md`, and trl/experimental/gold/scripts/xtoken/ is 644 lines of offline tooling inside the package, where the only scripts/ we have under trl/ is the cli entry points. I'd rather that not ship in the wheel.

On the split: agree with Albert that the test reorg should be its own PR (test reorg -> rebase dropping iw-opd -> x-token)

examples/xtoken/sort_and_cut_projection_matrix.py
```diff
@@ -0,0 +1,192 @@
...
+            if sv_lik.numel() > 0:
+                sorted_l, order = torch.sort(sv_lik, descending=True)
+                sorted_i = sv_idx[order]
+                n = min(len(sorted_i), new_top_k - 1)
+                new_idx[row, :n] = sorted_i[:n]
+                new_lik[row, :n] = sorted_l[:n]
+
+            if last_i != -1 and last_l > 0:
+                new_idx[row, new_top_k - 1] = last_i
+                new_lik[row, new_top_k - 1] = last_l
```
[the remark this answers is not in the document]
kashif: Fixed in dc78a1aa. The preserved scale slot now retains its positive likelihood even when its sentinel teacher id is -1, including rows without another valid mapping. Added a regression test for both row shapes.

trl/experimental/gold/gold_trainer.py
```diff
@@ -765,7 +766,451 @@ def _get_start_and_size_answers(self, answer_tensors):
...
+        s_log = F.log_softmax(s_logits.float() / T, dim=-1)
+        t_log = F.log_softmax(t_logits.float() / T, dim=-1)
+        s_chunks = self._chunk_average(s_log, s_groups)
+        t_chunks = self._chunk_average(t_log, t_groups)
+
+        if common_s.numel() > 0:
+            kl_per_chunk = F.kl_div(
+                s_chunks[:, common_s], t_chunks[:, common_t], reduction="none", log_target=True
+            ).sum(dim=-1)
+            kl_common = kl_per_chunk.mean()
```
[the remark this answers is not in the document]
kashif: Fixed in dc78a1aa. The hybrid estimate is now floored at zero before dynamic scaling, preventing a negative partial-common/top-k approximation from canceling CE. Added a regression test covering the negative approximation.

qgallouedec (changes requested):
IW-OPD duplication is gone, diff is X-Token only now, thanks.

The h_kl sign isn't settled though. `clamp_min(0.0)` treats the negative value as a top-k artifact, but it isn't: `kl_common` sums `exp(t_j) * (t_j - s_j)` over a non-renormalized subset, so any token where the student is sharper than the teacher contributes a negative term. Check your test data, student `[0.8, 0.1, 0.1]` against teacher `[0.4, 0.3, 0.3]` on common set `{0}` gives `-0.277`.

And `clamp_min` has no gradient below the threshold, so in that regime h_kl contributes nothing and nothing says so. Renormalizing the restricted distributions before the KL makes it non-negative by construction and keeps the gradient.

PR #6138, opened 2026-06-22
[codex] Fix OpenEnv example remote compatibility

burtenshaw:
Fixes the OpenEnv example scripts so they install from the right Hugging Face Spaces and can run against the current async OpenEnv clients. It also adds small runtime knobs needed to exercise BrowserGym and multi-env examples on real remote infrastructure.

### DIFF
--- a/examples/scripts/openenv/browsergym.py
+++ b/examples/scripts/openenv/browsergym.py
@@ -17,7 +17,7 @@
 #     "trl[vllm,peft]",
 #     "trackio",
 #     "kernels",
-#     "openenv-browsergym @ git+https://huggingface.co/spaces/openenv/browsergym_env",
+#     "openenv-browsergym-env @ git+https://huggingface.co/spaces/openenv/browsergym_env",
 # ]
 # ///
 
@@ -30,7 +30,7 @@ to see the page visually after each action.
 
 Setup:
 ```sh
-pip install "openenv-browsergym @ git+https://huggingface.co/spaces/openenv/browsergym_env"
+pip install "openenv-browsergym-env @ git+https://huggingface.co/spaces/openenv/browsergym_env"
 ```
 
 Usage:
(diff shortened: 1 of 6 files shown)


sergiopaniego (commented):
Thanks for this, the async bridge is the right direction and it's built on the current `environment_factory` architecture, which is what we want. This effectively supersedes two older PRs of mine (#4949 async
  support, #5568 browsergym improvements), which I'm closing in favor of this one.

  A few things before merging:

  **1. Rebase on `main` (it's ~15 commits behind and conflicts).** Part of the BrowserGym changes here are already on `main` via #6117 (`openenv-browsergym-env` package name + `--task-name` /
  `reset(task_name=...)`). After rebasing, those hunks should resolve to what's already in `main`. The genuinely new parts (the `run_sync` shim, `--report-to`, the CARLA URL fix, the multi_env knobs) are what we
  want to keep.

  **2. CARLA URL fix is correct.** Confirmed `sergiopaniego/carla-env` is the live public Space (the underscore one is private now), so the `carla_env -> carla-env` rename is right.

  **3. `multi_env.py`: `--num-generations` default of `1` is a footgun for GRPO.** A group of 1 gives zero advantage (no learning signal). The previous code used the `GRPOConfig` default. Could we default to >= 2,
  or document it explicitly as a smoke-test value?

  **4. Minor: the per-instance event loops (`asyncio.new_event_loop()`) are never closed.** Fine for an example script, not blocking, just flagging.

  **5. Worth folding in (from #5568).** Since we're consolidating here, it would be nice to also bring in the three BrowserGym improvements from #5568: `_normalize_bid()` + `repr()` action formatting, the
  `reward_efficiency` reward func, and `_do_action` returning a string instead of raising on a done episode. I can push those onto this branch if you'd like.

  The `run_sync` helper is byte-identical across all four scripts, which keeps it consistent. One thing to keep in mind: `run_sync` uses `loop.run_until_complete()`, which won't work inside a live event loop (e.g. a
  notebook). For the script examples that's fine, just noting it in case we ever want the same in notebooks.

burtenshaw:
Thanks for the review @sergiopaniego 

I'm not a hundred percent sure I want to add the run sync examples on the TRL side. I think I'd prefer to fix this on the OpenEnv servers so that this shim isn't necessary.

PR #6140, opened 2026-06-22
Add entropy regularization to GRPO

### LINKED ISSUE #3320: Add Adaptive Entropy Control to GRPOTrainer

albertvillanova:
Add entropy regularization to GRPO.

Fix #3320.
Supersede and close #3628.

This PR adds support for entropy regularization to the GRPO trainer, including both static and adaptive entropy control. Entropy regularization helps encourage exploration and prevents the policy from collapsing to deterministic outputs. The changes include configuration options, loss computation, metrics logging, checkpointing, and documentation updates, as well as new tests to verify the feature.

### Changes

**GRPO Trainer: Entropy Regularization**

*Implementation and Loss Function*
- Added entropy regularization to the GRPO loss, supporting both static and adaptive entropy coefficients. The entropy bonus is applied to the loss, and the coefficient can be updated each optimizer step if adaptive entropy is enabled.
- The entropy coefficient and adaptive entropy settings are initialized from `GRPOConfig`, and an error is raised if entropy regularization is used with the Liger kernel.
- During training, entropy-related metrics (`entropy_loss`, `entropy_coef`) are logged, and the adaptive coefficient is updated based on the current entropy.

*Configuration and Checkpointing*
- Added new fields to `GRPOConfig` for entropy regularization, including `entropy_coef`, `use_adaptive_entropy`, `entropy_coef_min`, `entropy_coef_max`, `entropy_coef_delta`, and `entropy_target`, with detailed documentation.
- When using adaptive entropy, the current coefficient is saved and restored with checkpoints to ensure training is resumable.

*Documentation and Testing*
- Updated the documentation to describe entropy regularization, its configuration, and usage examples for both static and adaptive modes. Also documented new reward metrics related to entropy. 
- Added tests for both static and adaptive entropy regularization to verify training and logging of the new metrics.

---

> [!NOTE]
> **Medium Risk**
> Changes core GRPO loss and forward entropy computation in the trainer; scope is bounded by opt-in config and Liger is rejected, but incorrect scaling or adaptive gating could affect training dynamics when enabled.
> 
> **Overview**
> Adds **entropy regularization** to `GRPOTrainer`: the objective becomes policy loss minus `entropy_coef` times mean per-token entropy, with the bonus **not** rescaled per `loss_type`.
> 
> **`GRPOConfig`** exposes `entropy_coef`, `use_adaptive_entropy`, `entropy_target`, `entropy_coef_delta`, and min/max bounds. Adaptive mode adjusts the coefficient once per optimizer step from window-aggregated entropy (gradient accumulation–aware), applies the bonus only when entropy is at or below the target, and persists `entropy_coef` / `last_world_entropy` in `entropy_ctrl_state.json` on checkpoint resume. Entropy bonus is **incompatible with the Liger kernel**.
> 
> **Training behavior:** per-token entropies are computed **with gradients** when the bonus is active (fixing a prior `no_grad` no-op); otherwise they stay detached for memory. Logs **`policy_loss`** and **`entropy_coef`** when regularization is on.
> 
> Docs cover usage, metrics, and Skywork-OR1 in `paper_index.md`. New tests cover static/adaptive training, bonus scale across loss types, grad accumulation for adaptive control, and entropy gradients.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64932fdc1d9dfd306859d624450f8f38f5ccb461. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/trainer/grpo_trainer.py
```diff
@@ -2751,6 +2804,11 @@ def masked_batch_mean(x):
             else:
                 return (x * mask).sum() / completion_token_count
 
+        self._metrics[mode]["policy_loss"].append(self.accelerator.reduce(policy_loss, reduction="mean").item())
```
qgallouedec: policy_loss is captured after `loss = loss / normalizer`, where `normalizer = current_gradient_accumulation_steps`, so the logged `policy_loss` is the per-micro-batch contribution, not the step loss; it'll read ~accum× too small. I think we should capture before dividing.

qgallouedec: another remark, it can be misleading to have `policy_loss` logged when it's not used in the loss. Maybe we should gate its logging.

trl/trainer/grpo_trainer.py
```diff
@@ -2736,6 +2744,51 @@ def _compute_loss(self, model, inputs):
...
+        # Capture the pure policy loss for logging before entropy/aux modify it
+        policy_loss = loss.detach()
+
+        # Entropy bonus: add entropy regularization to encourage exploration.
+        # Gate: run whenever a non-zero static coef is set OR adaptive mode is enabled. Adaptive must always run even
+        # when self.entropy_coef has been decremented to entropy_coef_min (default 0) so it can recover once entropy
+        # drops below entropy_target again.
+        if self.entropy_coef != 0.0 or self.use_adaptive_entropy:
+            if self.loss_type in ["grpo", "sapo", "luspo"]:
+                entropy_loss = ((entropies * mask).sum(-1) / mask.sum(-1).clamp(min=1.0)).mean() / normalizer
```
qgallouedec: The entropy block lumps luspo with grpo/sapo, but luspo's actual loss is `(per_token_loss * mask.sum(1,keepdim=True)).mean()`. So for luspo the entropy bonus lives on a different scale than the policy term and `entropy_coef` means something different there. Either give luspo its own branch or note it. To be confirmed, but probably something like:

```python
entropy_loss = (entropies * mask).sum(-1).mean() / normalizer
```

trl/trainer/grpo_trainer.py
```diff
@@ -2751,6 +2804,11 @@ def masked_batch_mean(x):
             else:
                 return (x * mask).sum() / completion_token_count
 
+        self._metrics[mode]["policy_loss"].append(self.accelerator.reduce(policy_loss, reduction="mean").item())
+        if self.entropy_coef != 0.0 or self.use_adaptive_entropy:
+            self._metrics[mode]["entropy_loss"].append(world_entropy)
```
qgallouedec: We would have two near-duplicate entropy metrics. We could drop it, and just log `entropy`. The existing `entropy` already logs mean entropy; `entropy_loss` is just a slightly different (global vs gathered-local-mean) computation of the same quantity.

trl/trainer/grpo_trainer.py
qgallouedec: and just for matching home-style:

```python
self._metrics[mode]["policy_loss"].append(self.accelerator.gather(policy_loss).nanmean().item())
```

qgallouedec (commented):
thanks, a few remarks

trl/trainer/grpo_trainer.py
```diff
@@ -2726,26 +2734,94 @@ def _compute_loss(self, model, inputs):
...
             loss = loss / normalizer
         elif self.loss_type == "bnpo":
             loss = (per_token_loss * mask).sum() / mask.sum().clamp(min=1.0)
             normalizer = self.current_gradient_accumulation_steps if mode == "train" else 1.0  # no accum in eval
+            policy_loss = loss.detach()
             loss = loss / normalizer
         elif self.loss_type == "dr_grpo":
             loss = (per_token_loss * mask).sum() / (per_token_loss.size(0) * self.max_completion_length)
             normalizer = self.current_gradient_accumulation_steps if mode == "train" else 1.0  # no accum in eval
+            policy_loss = loss.detach()
```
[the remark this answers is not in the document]
albertvillanova: The observation is correct: the entropy term and the Dr. GRPO policy term are normalized differently. But this is intentional, and it's what makes the code match the documented objective rather than break it.

The documented objective is L = L_GRPO − α·H, where H is the mean per-token entropy. H is a single, well-defined quantity (nats per token); it does not depend on loss_type. The code computes exactly loss = L_GRPO − α · (mean per-token entropy), so entropy_coef weights H exactly as documented — for every loss type.

If we instead scaled the entropy term by 1/(B·max_len), the bonus would become α · H · (actual_len / max_len) — i.e. it would shrink the entropy regularization for shorter completions. There's no principled reason to regularize entropy less when responses are short, and that expression no longer equals the documented α·H. So matching the Dr. GRPO normalization is the change that would deviate from the documented objective, not the current code.

Two more reasons the current design is the right one:
- Consistency across loss types. As written, entropy_coef has one meaning everywhere: the weight on the mean per-token entropy in nats. If the entropy term tracked each loss type's policy normalization, entropy_coef would mean something different for every loss type (fixed-budget for dr_grpo, sequence-length-weighted for luspo, per-token for grpo, …).
- Adaptive control. The adaptive controller compares world_entropy (the global mean per-token entropy) to entropy_target. Because the regularized term is the mean per-token entropy, the quantity being gated/targeted and the quantity being optimized are on the same scale. Scaling the entropy term by 1/(B·max_len) would desync it from world_entropy.

Adding a sentence to the docs making explicit that H is always the mean per-token entropy regardless of loss_type, if that would prevent the confusion.

trl/trainer/grpo_trainer.py
```diff
@@ -2760,26 +2768,92 @@ def _compute_loss(self, model, inputs):
...
+            self._metrics[mode]["policy_loss"].append(self.accelerator.gather(policy_loss).nanmean().item())
+
+            # Adaptive update: once per optimizer step, measure the global token-weighted entropy and adjust
+            # the coefficient for the next step. Gated on train mode so evaluation cannot mutate the entropy
+            # controller state, and on sync_gradients so the all-reduce runs once per optimizer step rather
+            # than on every micro-batch of the accumulation window.
+            if self.use_adaptive_entropy and mode == "train" and self.accelerator.sync_gradients:
+                # Reduce sum and token count jointly for a true global mean (unbiased when ranks have
+                # different completion lengths).
+                entropy_stats = self.accelerator.reduce(
```
qgallouedec: `world_entropy` is measured from only the last micro-batch of the accumulation window, reduced across ranks but not across accumulation. So with` gradient_accumulation_steps > 1` it's a 1/GAS subsample, and "global token-weighted entropy" in the comment is a bit optimistic. The controller integrates slowly (Δ=0.005) so it's minor, but it's easy to make exact: accumulate `(entropies*effective_mask).sum()` and `effective_mask.sum()` (detached) into running buffers every micro-batch, then reduce the window totals at `sync_gradients` and reset.

trl/trainer/grpo_trainer.py
```diff
@@ -2929,3 +3003,18 @@ def _save_checkpoint(self, model, trial):
             model_name = self.args.hub_model_id.split("/")[-1]
         self.create_model_card(model_name=model_name)
         super()._save_checkpoint(model, trial)
+        if self.use_adaptive_entropy and self.args.should_save:
+            checkpoint_folder = f"checkpoint-{self.state.global_step}"
```
qgallouedec: prefer `PREFIX_CHECKPOINT_DIR` from `transformers.trainer_utils` so this stays correct if HF ever renames it.

trl/trainer/grpo_trainer.py
```diff
@@ -2929,3 +3003,18 @@ def _save_checkpoint(self, model, trial):
...
+
+    def _load_optimizer_and_scheduler(self, checkpoint):
+        super()._load_optimizer_and_scheduler(checkpoint)
+        if self.use_adaptive_entropy and checkpoint is not None:
+            path = os.path.join(checkpoint, "entropy_ctrl_state.json")
+            if os.path.exists(path):
+                with open(path) as f:
+                    state = json.load(f)
+                self.entropy_coef = state["entropy_coef"]
+                self._last_world_entropy = state.get("last_world_entropy", float("inf"))
```
qgallouedec: any reason to not use `state[...]`? This JSON is always written by the same code path with both keys, so the `.get(..., inf)` fallback is dead defensive code, no?

qgallouedec (approved):
ok lgtm, thanks

trl/trainer/grpo_trainer.py
```diff
@@ -2845,26 +2858,97 @@ def _compute_loss(self, model, inputs):
...
+
+            # Apply the coefficient and gating from the end of the previous optimizer step, so that every
+            # micro-batch in the current accumulation window applies the same entropy bonus. The adaptive
+            # update below only takes effect on the next step.
+            if self.use_adaptive_entropy:
+                apply_coef = self.entropy_coef if self._last_world_entropy <= self.args.entropy_target else 0.0
+            else:
+                apply_coef = self.entropy_coef
+
+            loss = loss - apply_coef * entropy_loss
```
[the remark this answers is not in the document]
albertvillanova: I think the review is valid and correct: a critical no-op bug. The no_grad was a valid memory optimization for entropy's original uses (the logged entropy metric and the top_entropy_quantile mask; both non-differentiable). The new bonus needs gradients.

I'm addressing it and adding a regression test.

PR #6161, opened 2026-06-24
Keep extra columns in unpair_preference_dataset

albertvillanova:
Keep extra column in `unpair_preference_dataset`.

This PR fixes unpair_preference_dataset to preserve extra columns instead of silently dropping them, e.g. `chat_template_kwargs`. 

Follow-up to:
- #6059

### Motivation

PR #6059 fixed a ValueError caused by a length mismatch: when a paired dataset had extra columns, those columns kept their original row count (N) after unpairing while "completion" and "label" grew to 2N rows. The fix at the time was to drop all columns with `remove_columns=column_names`, leaving only the function's output. However, this also silently discards valid user data such as `chat_template_kwargs`, breaking per-example tokenization in the KTO trainer.

### Solution

Preserve extra columns in `_unpair_row` by duplicating their values for the chosen and rejected rows, the same way "prompt" is already handled. This keeps all lengths consistent (2N) without discarding user data.

### Changes
- Update _unpair_row to duplicate extra columns alongside chosen and rejected rows
- Update unpair_preference_dataset docstring to reflect that extra columns are now preserved
- Update tests to assert extra columns are preserved rather than dropped

---

> [!NOTE]
> **Low Risk**
> Small, localized change to dataset shaping with updated tests; restores metadata needed by KTO without touching auth or training core logic.
> 
> **Overview**
> **`unpair_preference_dataset`** no longer drops non-core columns when converting paired preference data to unpaired rows. **`_unpair_row`** now copies every column except `chosen` and `rejected` by concatenating each value list with itself (same pattern as `prompt`), so row counts stay at 2N for fields like `chat_template_kwargs`.
> 
> The public docstring and tests were updated to expect preserved, duplicated extras on both `Dataset` and iterable paths instead of silent removal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc76c772abb819cf4f9b93feece825baffc2e821. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_data_utils.py
+++ b/tests/test_data_utils.py
@@ -995,7 +995,7 @@ class TestUnpairPreferenceDataset(TrlTestCase):
         )
 
     def test_unpair_preference_dataset_extra_columns(self):
-        # Test that extra columns are dropped (not causing a length mismatch error)
+        # Test that extra columns are preserved and duplicated for chosen and rejected rows
         paired_dataset = Dataset.from_dict(
             {
                 "prompt": ["The sky is", "The sun is"],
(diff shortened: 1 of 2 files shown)


qgallouedec (approved):
Ok lgtm!

PR #6175, opened 2026-06-25
Promote KTO to stable API

### LINKED ISSUE #4786: KTO refactoring

albertvillanova:
Promote KTO to stable API.

Close #4786.

This PR promotes the `KTOTrainer` and `KTOConfig` from the experimental API (`trl.experimental.kto`) to the stable API (`trl`). It updates all relevant documentation, scripts, and tests to use the new import paths, and deprecates the old experimental import with a warning. This change simplifies usage for end users and signals that the KTO API is now considered stable.

### Changes

**API Promotion and Deprecation:**

* The experimental `KTOConfig` and `KTOTrainer` classes now inherit from the stable API and emit a deprecation warning if imported from the experimental path, indicating users should switch to `from trl import ...`.

**Documentation Updates:**

* `docs/source/kto_trainer.md`, `docs/source/paper_index.md`, `docs/source/reducing_memory_usage.md`, `docs/source/speeding_up_training.md`: All references and code examples now use the stable import path for `KTOTrainer` and `KTOConfig`. 
* The warning about KTO being experimental has been removed.

**Code and Script Updates:**

* `examples/scripts/kto.py`, `trl/scripts/kto.py`: Updated imports to use `from trl import KTOConfig, KTOTrainer` and related types. 

**Test Updates:**

* `tests/test_kto_trainer.py` (renamed from `tests/experimental/test_kto_trainer.py`): Updated to import from the stable API and moved to the main test directory, reflecting the stable status.

---

> [!NOTE]
> **Low Risk**
> Mostly import-path and documentation changes with backward-compatible deprecation shims; training logic is relocated/re-wired via inheritance, not rewritten.
> 
> **Overview**
> **KTO** (`KTOTrainer` / `KTOConfig`) is promoted from `trl.experimental.kto` to the stable **`trl`** package, alongside DPO, GRPO, and SFT.
> 
> Documentation now lists KTO under **Trainers** (removed from Experimental), drops the experimental warning on the KTO guide, and updates examples to `from trl import KTOConfig, KTOTrainer`. The trainer implementation under `trl/trainer/` gets corrected relative imports; docstrings and collator examples point at `trl.trainer.kto_trainer` where relevant.
> 
> `trl.experimental.kto` remains as thin subclasses that emit **`FutureWarning`** (removal in v2.0.0) and delegate to the stable classes. Example script, CLI `kto` script, and **`tests/test_kto_trainer.py`** (moved from experimental) import from stable `trl`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e0d7510340d59d74bfe638720c52aec6caed9b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
I'm a bit worried that this would break the git history: 

Both `trl/trainer/kto_trainer.py` and `trl/experimental/kto/kto_trainer.py` already exist on `main`. This PR swaps their contents rather than renaming a file, so git sees ~1500 lines deleted from one and added to the other as independent edits -> no rename is detected, and `blame`/`log --follow` on the promoted implementation will collapse to this one commit.

We could make it a real rename instead: `git mv` the experimental implementation onto `trl/trainer/kto_trainer.py` (removing the old shim first), then add the new experimental shim as a fresh file. That lets git detect the rename and preserves blame history. Same applies to `kto_config.py`.

albertvillanova:
@qgallouedec, thanks for pointing this out. I agree with the concern in principle: this promotion creates a large history boundary for kto_trainer.py/kto_config.py.

That said, since this PR will be squash-merged, I don’t think using `git mv` inside the branch would materially change the final repository history. Git does not store renames explicitly; rename detection is inferred from the final diff. Because both files already exist on main and both still exist after the PR, the squashed commit would still look like a large rewrite/content swap rather than a clean rename. 

I tried the suggested approach anyway, but GitHub still shows additions/deletions rather than a rename.

albertvillanova:
Offline discussion:
> @qgallouedec: I think we should do it with two separate PRs btw: remove trl.trainer.kto_trainer/config.py, then move trl.experimental.kto.kto_trainer

I dug into this concretely and I don't think we should use two-PR split:
1. The two-PR split doesn't recover it either. I simulated PR-2 (start = stable shims removed, impl still in experimental; end = impl in trl/trainer, experimental reduced to a shim).
   - Git only detects a rename when the source path is deleted. But we deliberately keep a deprecation shim at trl.experimental.kto, so trl/experimental/kto/kto_trainer.py is never deleted — only shrunk. The only way to make git emit a rename would be to delete the experimental module outright, which would break from trl.experimental.kto import … for downstream users and defeat the whole point of the deprecation shim.
2. This history boundary already exists and predates this PR. When KTO was moved to experimental in #4575, git recorded the experimental files as new files, not renames.
   - So blame on the current implementation already collapses at #4575. This promotion isn't introducing the first boundary — it's the second half of a relocation whose first half was recorded the same way.
3. It can't be a pure rename anyway. Moving between package depths forces the relative imports to change (from ...trainer.base_config → from .base_config, from ...data_utils → from ..data_utils, …), which is why the move needs the follow-up 021039c4. A byte-identical rename would leave the file un-importable.
4. The two changes aren't independent. trl/__init__.py re-exports KTOConfig/KTOTrainer from trl.trainer. A standalone "remove trl.trainer.kto_*" PR would break from trl import KTOTrainer and leave main in an inconsistent state until the second PR lands. The promotion is only meaningful as a single atomic change: one PR would temporarily delete the stable API, the other would restore it, and neither leaves the repo consistent on its own.

trl/experimental/kto/kto_config.py
```diff
@@ -12,224 +12,20 @@
...
-        },
-    )
-    ref_model_sync_steps: int = field(
-        default=512,
-        metadata={
-            "help": "τ parameter from the TR-DPO paper, which determines how frequently the current policy is "
-            "synchronized with the reference policy. To use this parameter, you must set `sync_ref_model=True`."
-        },
-    )
+class KTOConfig(_KTOConfig):
```
qgallouedec: and what if we move this into `experimental/kto/__init__.py` instead?

qgallouedec:
for 1. I think it can be solved by https://github.com/huggingface/trl/pull/6175#discussion_r3520583188

for 2, yes, and this is unfortunate, that why I wanted to do differently this time.

for 3. I double checked, it seems like it doesn't have to be a pure rename, see https://github.com/qgallouedec/dummy-kto-trainer/commit/95e24b6383db962e5dc1e37ba9ce04cb0fb578d3 (detected as rename even with imports being modified)

for 4. I think it's really not a big deal, we can merge the second right after the first

---

your call, I think we can save the git history here, but if you feel it's not worth it, I'm ok with doing it like this

qgallouedec (approved):
Let's go!!

qgallouedec:
Please update toctree.yml as well

albertvillanova:
Finally, KTO files appear as moved and then conserve Git history. :sweat_smile: 

Screenshot from 2026-07-05 06-58-36

PR #6195, opened 2026-06-28
Align TPO with DPO: Add evaluate() override

DaoyuanLi2816:
Follow-up to #6116 (core trainers) and #6148 (KTO): `TPOTrainer` preprocesses `eval_dataset` in `__init__`, but doesn't override `evaluate()`, so passing a raw (unprocessed) dataset directly to `evaluate(eval_dataset=...)` — e.g. a held-out test set — fails:

```
ValueError: No columns in the dataset match the model's forward method signature:
(prompt_ids, chosen_ids, rejected_ids, reference_ids). The following columns have been
ignored: [prompt, rejected, chosen, reference].
```

This adds the same `evaluate()` override as DPO/KTO: when a dataset is passed directly to `evaluate`, it is preprocessed with the trainer's idempotent `_prepare_dataset` (a `str` selects a dataset already prepared at init time and is left untouched; dicts are handled per key). TPO has no reference model / `precompute_ref_log_probs` and no VLM path, so the override is the simplest form (no precompute or vision branch).

This is the TPO case of the same issue as #6115 / #6116.

### Verification

- Added `test_evaluate_with_raw_dataset` (mirrors KTO's): it fails on `main` with the `ValueError` above and passes with this change.
- Full `tests/experimental/test_tpo_trainer.py` passes locally; `ruff check` / `ruff format --check` clean.

---

> [!NOTE]
> **Low Risk**
> Small API parity change on the eval path; behavior for init-time eval datasets and str keys is unchanged.
> 
> **Overview**
> **`TPOTrainer.evaluate`** now runs **`_prepare_dataset`** on eval data passed at call time (single dataset or dict of datasets), matching DPO/KTO. String dataset keys that were already prepared in **`__init__`** are unchanged.
> 
> Adds **`test_evaluate_with_raw_dataset`** so a held-out raw triple-preference set can be evaluated without the prior column-mismatch **`ValueError`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39e5dcce770e0e576c15b0202821b8fb0d68d5ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/experimental/test_tpo_trainer.py
+++ b/tests/experimental/test_tpo_trainer.py
@@ -130,6 +130,20 @@ class TestTPOTrainer(TrlTestCase):
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
 
+    def test_evaluate_with_raw_dataset(self):
+        # `evaluate` should accept the same (unprocessed) dataset types as the trainer, e.g. a held-out test set
+        # passed directly to `evaluate`, mirroring DPO/KTO. See https://github.com/huggingface/trl/issues/6115.
+        dataset = load_dataset("trl-internal-testing/zen", "standard_preference", split="train")
+        dataset = dataset.map(_add_reference_column)
+
+        training_args = TPOConfig(output_dir=self.tmp_dir, report_to="none")
+        trainer = TPOTrainer(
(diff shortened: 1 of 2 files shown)


qgallouedec (approved):
thanks

PR #6216, opened 2026-06-30
Align BCO with DPO: Add evaluate() override and precompute_ref_batch_size

DaoyuanLi2816:
## Status

Closed to consolidate related trainer work instead of keeping several overlapping pull requests open.

## AI writing disclosure

### DIFF
--- a/tests/experimental/test_bco_trainer.py
+++ b/tests/experimental/test_bco_trainer.py
@@ -219,6 +219,34 @@ class TestBCOTrainer(TrlTestCase):
         assert metrics["eval_loss"] is not None
 
     @require_sklearn
+    def test_evaluate_with_raw_dataset_keeps_init_eval_dataset(self):
+        # evaluate(eval_dataset=<foreign>) must not overwrite the trainer's own eval_dataset, even with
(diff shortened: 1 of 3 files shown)


DaoyuanLi2816:
Good catch. Fixed: `get_eval_dataloader` now distinguishes the trainer's own eval dataset from one passed directly to `evaluate()`. A directly-passed (foreign) dataset has its reference log-probs precomputed fresh and is **not** cached back onto `self.eval_dataset` (nor does it flip `_precomputed_eval_ref_log_probs`), so a later no-arg `evaluate()` still runs on the original init eval set.

This also fixes the converse case the old logic had: once `_precomputed_eval_ref_log_probs` was set (e.g. after evaluating the init dataset), a subsequent `evaluate(eval_dataset=<foreign>)` would have skipped precompute entirely and failed because the foreign dataset lacked `reference_logps`.

Added `test_evaluate_with_raw_dataset_keeps_init_eval_dataset` to cover it (with `precompute_ref_log_probs=True`, asserts `self.eval_dataset` is unchanged after `evaluate(eval_dataset=<other>)`).

DaoyuanLi2816:
Closing this to consolidate the outstanding trainer-parity work instead of keeping several related PRs open in parallel.

PR #6218, opened 2026-06-30
Default AsyncGRPO token_budget to vLLM max_model_len

AmineDiro:
Follow-up to #6092. Fixes the default value of `AsyncGRPOConfig.token_budget`

At `per_device_train_batch_size=1` the budget collapses to exactly `max_completion_length`, so **every sample with a non-trivial prompt gets dropped**  and if all queued samples are oversize, the batcher never yields and the dataloader blocks. (Both flagged by Cursor Bugbot on #6092)

## Fix

Default `token_budget` to the **vLLM server's `max_model_len`** :  the cap on `prompt + completion`, so no rollout sample can ever exceed the budget.

## Verification

Against a live `vllm serve Qwen/Qwen3-0.6B --max-model-len 2048`:
- `GET /v1/models` → `"max_model_len": 2048`, and `_get_vllm_max_model_len` returns `2048`.
- `POST /v1/completions` with `max_tokens=2048` + a prompt is rejected with *"maximum context length is 2048 tokens … 2048 output tokens … 0 input tokens"* — confirming `max_model_len`
caps prompt **and** completion together.

Existing `TestPackingAwareBatching` tests pass.

---

> [!NOTE]
> **Medium Risk**
> Train-start HTTP to vLLM and a larger default budget change memory/throughput characteristics, but the change fixes sample dropping and blocking; custom rollout users must set token_budget explicitly.
> 
> **Overview**
> Fixes a follow-up to #6092 where the old default `token_budget = per_device_train_batch_size × max_completion_length` could equal only `max_completion_length` at batch size 1, so **prompt + completion** rollouts were dropped and the dataloader could stall.
> 
> When `token_budget` is unset, it is now resolved in **`get_train_dataloader`** by calling **`_get_vllm_max_model_len`** (`GET {vllm_server}/v1/models`) so the budget matches the server’s **`max_model_len`** (prompt + completion cap). Config **`__post_init__`** no longer sets this default. Custom **`rollout_worker`** (no built-in vLLM / `weight_transfer is None`) must pass **`token_budget` explicitly** or training raises **`ValueError`**.
> 
> Tests set **`token_budget=256`** for the stub worker and drop the test that asserted the old config default.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 72ac43a1e52f91dbf7e0c604f8011d2ce79a7e7b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/async_grpo/async_grpo_trainer.py
+++ b/trl/experimental/async_grpo/async_grpo_trainer.py
@@ -22,6 +22,7 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Any, Protocol
 
+import requests
 import torch
 from accelerate.logging import get_logger
 from datasets import Dataset, IterableDataset
@@ -174,6 +175,13 @@ class RolloutQueueDataset(torch.utils.data.IterableDataset):
             }
 
 
+def _get_vllm_max_model_len(server_url: str, timeout: float) -> int:
+    """Query the vLLM server for the served model's `max_model_len` (the cap on prompt + completion tokens)."""
+    response = requests.get(f"{server_url.rstrip('/')}/v1/models", timeout=timeout)
+    response.raise_for_status()
+    return response.json()["data"][0]["max_model_len"]
+
+
 def _balance_by_squared_length(examples: list[dict[str, Any]], num_groups: int) -> list[list[dict[str, Any]]]:
     """Greedily partition `examples` into `num_groups` rows (one per DP rank), balancing each row's Σ Lᵢ².
 
@@ -235,8 +243,8 @@ class TokenBudgetBatcher(torch.utils.data.IterableDataset):
 
     Every emitted micro-batch has all `num_processes` rows non-empty (a rank forwarding zero tokens would desync
     FSDP/EP collectives): a micro-batch is only closed once every row holds at least one sample. A sample longer than
-    `token_budget` fits in no row, so it is dropped with a warning; set `token_budget` ≥ the longest sample
-    (`max_completion_length` + the longest prompt) to avoid dropping samples.
+    `token_budget` fits in no row, so it is dropped with a warning; set `token_budget` ≥ the vLLM server's
+    `max_model_len` (the cap on prompt + completion) to avoid dropping samples.
 
     Args:
         dataset ([`RolloutQueueDataset`]):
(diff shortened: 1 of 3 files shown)


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -660,6 +668,19 @@ def get_train_dataloader(self) -> DataLoader:
...
             )
+            # Default the token budget to the vLLM server's max_model_len (the cap on prompt + completion), so no
+            # rollout sample can exceed it. Only the built-in worker manages a vLLM server (weight_transfer is set);
+            # with a custom rollout_worker there may be none to query, so require an explicit budget instead.
+            if self.args.token_budget is None:
+                if self.weight_transfer is None:
+                    raise ValueError(
+                        "Set `token_budget` explicitly when passing a custom `rollout_worker`: the default is the "
+                        "vLLM server's max_model_len, which is only queried for the built-in rollout worker."
+                    )
```
qgallouedec, line 679: we could have a required getter for this value. Idea for a follow-up pr

qgallouedec (approved):

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: sorry can you precise ? 

This branch is only when the user passes a custom rollout worker and we don't know we have a vllm server accessible

qgallouedec: Sorry, it wasn't clear. My point was that we could make the contract evolve to

```python
class RolloutWorkerProtocol(Protocol):
    rollout_buffer: queue.Queue
    def start(self) -> None:
    def stop(self) -> None:
    def update_model_version(self, version: int) -> None:
    def check_health(self, stale_after_s: float) -> None:
    def get_max_model_len(self) -> int:  # <- This
```

and here, simply:

```python
if self.args.token_budget is None:
    self.args.token_budget = self.rollout_worker.get_max_model_len()
```

AmineDiro: Aaahh I see, `get_max_model_len` doesn't seem to be related to the `RolloutWorker` interface. All other methods handle the rollout logic, so the `def get_max_model_len(self) -> int:  # <- This` feels a little bit out of field ? 

Actually looking at this, when a user passes a custom `rollout_worker`, we completely remove the `weight_transfer` which is problematic for custom workers 🤔 
```python
 if rollout_worker is not None:
      # Use the injected worker (e.g. a stub in tests). The queue is owned by the worker.
      # Weight transfer is also expected to be wired by the test fixture (or left as None
      # if the stub doesn't sync to a real vLLM).
      self.rollout_worker = rollout_worker
      self.weight_transfer = None
```

AmineDiro:
Thanks @qgallouedec  for the review 👍🏼 ! I agree that a getter is the right direction, but I don't think it belongs on `RolloutWorkerProtocol`.

Took a look over the design: three things independently talk to the same vLLM server today: 
- the rollout worker (generation)
- the WeightTransferClient (weight sync)
- the trainer (this max_model_len query).

So we duplicate the URL across all three, and the trainer uses `weight_transfer is None` as a stand-in for "is there a server I can query". That coupling is also why this query skips the `/health` readiness wait (the wait lives inside weight sync), and why passing a custom `rollout_worker` silently drops weight sync, which is really a bug. 

So what I'll fix in a separete PR is to instantiate  a single small concrete VLLMClient that owns the server (URL + readiness + max_model_len + world size). 

This PR's behavior is correct, so I'd keep it as-is and do the refactor in a follow-up.

PR #6227, opened 2026-06-30
Align KTO with DPO: Validate Liger kernel configuration before trainer initialization

albertvillanova:
This PR moves the validation of the `use_liger_kernel` configuration in the experimental `KTOTrainer` to before `super().__init__()`, aligning it with `DPOTrainer`.

### Motivation

The Liger kernel checks (kernel availability, unsupported loss type, precompute_ref_log_probs, and PEFT) depend only on the config and the input model, so there is no reason to run them after the expensive `super().__init__()` (model loading, dataset preparation, reference model creation, accelerator setup). `DPOTrainer` already validates these before `super().__init__()`, while `KTOTrainer` did it after, so an invalid configuration was rejected only at the end of a costly initialization.

### Solution

The validation checks are moved ahead of `super().__init__()` so an invalid Liger configuration fails fast. The construction of the fused Liger loss stays after `super().__init__()` because it depends on self.ref_model, which is only resolved during the reference model setup that runs after the parent initialization.

### Changes

- Move the Liger kernel validation checks (availability, loss type, `precompute_ref_log_probs`, PEFT) before `super().__init__()` in `KTOTrainer`
- Keep the `LigerFusedLinearKTOLoss` construction after `super().__init__()`, where `self.ref_model` is available
- Use the local model instead of self.model for the PEFT check, since self.model is not yet set before `super().__init__()`

---

> [!NOTE]
> **Low Risk**
> Init-order refactor only; validation rules are unchanged aside from using the local `model` for the PEFT check before `self.model` exists.
> 
> **Overview**
> **KTOTrainer** now runs all **`use_liger_kernel`** validation (package install, loss type, `compute_metrics`, `precompute_ref_log_probs`, PEFT) **before** `super().__init__()`, matching **DPOTrainer** so bad Liger settings fail immediately instead of after model load, dataset prep, and accelerator setup.
> 
> **`LigerFusedLinearKTOLoss`** is still constructed only after `super().__init__()` and reference-model setup, with a short comment that it depends on **`self.ref_model`**. The PEFT guard uses the local **`model`** argument instead of **`self.model`**, which is not assigned until the parent init runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4364bbaca6fa74c9ae36773c509994ddf798dd3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/kto/kto_trainer.py
```diff
@@ -675,6 +675,28 @@ def __init__(
...
+            if self.loss_type in ["apo_zero_unpaired"]:
+                raise ValueError(
+                    "You cannot set `loss_type='apo_zero_unpaired'` with liger-kernel."
+                    "Only KTO loss is supported with liger-kernel."
+                )
+            if self.precompute_ref_logps:
+                raise ValueError(
+                    "You cannot use `precompute_ref_log_probs=True` with liger kernel. Please set "
+                    "`precompute_ref_log_probs=False`."
+                )
```
qgallouedec, line 698: in dpo it's 

```python
"Liger DPO loss does not support precomputing reference log probabilities. Either disable "
"`precompute_ref_log_probs` or set `use_liger_kernel` to False."
```

albertvillanova: This PR is just a **move** of the Liger validation checks, to before `super().__init__()`; no error message modification.

Individual error message alignment is handled in other PR:
- https://github.com/huggingface/trl/pull/6270

qgallouedec (approved):

PR #5837, opened 2026-05-25
[WIP] Refactor CPOTrainer

qgallouedec:
Trying to one-shot CPOTrainer refactoring

I'll have to run the benchmark before it's ready

---

> [!NOTE]
> **High Risk**
> Breaking `CPOConfig` API and default truncation behavior can silently change training for existing scripts, and the trainer refactor touches core preference-optimization training paths now covered only by an expanded but still integration-heavy test suite.
> 
> **Overview**
> **Rebuilds `experimental.cpo.CPOTrainer` on the shared DPO-style trainer stack**, with a documented migration warning and a slimmer quick-start API (model id string, no manual tokenizer wiring in the main example).
> 
> **`CPOConfig` is reshaped to match DPO-style training options**: `loss_type` becomes a `list[str]` with optional `loss_weights`, plus new knobs like `truncation_mode`, `padding_free`, `pad_to_multiple_of`, `ld_alpha`, `f_divergence_type`, `use_weighting`, and `activation_offloading`. Removed fields (`label_smoothing`, `generate_during_eval`, `is_encoder_decoder`, `max_completion_length`) now error if passed; **`max_length` defaults to `1024`** instead of no truncation.
> 
> **Docs** expand CPO guidance (dataset formats, loss math, metrics, PEFT/VLM/tool-calling, Liger constraints) and lightly clarify AlphaPO in `paper_index.md`.
> 
> **Tests** replace the old Qwen/T5 matrix with broad integration coverage: preference collators (incl. VLM `mm_token_type_ids`), loss variants, multi-loss, WPO/LD/f-divergence, PEFT/MoE/quantization, Liger, padding-free, streaming, eval/metrics, and many VLM regression cases. **`conftest.py` sets `HF_HUB_DISABLE_TELEMETRY=1` by default during pytest** to avoid biasing Hub adoption stats.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b5351c6d62d3ddd3e8fb20fc7f591a25d52a415. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6233, opened 2026-07-01
Fix dataset fingerprinting in experimental KTO and TPO trainers

DaoyuanLi2816:
Follow-up to #6206, which fixed dataset fingerprinting in the core `DPOTrainer` / `SFTTrainer` / `RewardTrainer`. The two experimental trainers `KTOTrainer` and `TPOTrainer` still have the pattern #6206 removed: their tokenization map is a closure that calls `self._tokenize(...)`, so it captures `self`. A map function that closes over `self` (which holds the model and other unpicklable state) can't be hashed stably, so `datasets` falls back to a random fingerprint and silently disables caching of the tokenized dataset — re-tokenizing on every run.

This applies the same fix as #6206: make `_tokenize` a `@staticmethod` and bind it to a local (`tokenize = self._tokenize`) before defining the map function, so the closure no longer captures `self`.

- **KTO**: `_tokenize` uses the VLM flag, so `is_vlm` becomes an explicit parameter passed through `fn_kwargs` (exactly as in #6206's DPO).
- **TPO**: `_tokenize` doesn't reference `self` at all, so it simply becomes a `@staticmethod`.

No behavior change: the tokenized output is identical; only the map function's hashability (and therefore cache reuse across runs) changes.

`CPOTrainer` and `ORPOTrainer` share the same pattern (a `self`-bound `tokenize_row`) but are intentionally left out here to avoid conflicts: CPO's data pipeline is being refactored in #5837, and ORPO's `_prepare_dataset` is touched by #6230. They can be aligned once those land.

### Verification

- `tests/experimental/test_kto_trainer.py` and `tests/experimental/test_tpo_trainer.py` pass in full (tokenized output unchanged); `ruff check` / `ruff format --check` clean. Validated on a single GPU (RTX 4080).
- Like #6206, this doesn't add a fingerprint-specific test: the fallback to a random fingerprint only triggers for processing classes / models that `datasets` can't hash, which the tiny CI fixtures don't exercise. The existing suites cover tokenization correctness.

---

> [!NOTE]
> **Low Risk**
> Refactor limited to dataset preprocessing closures with no training-loss or tokenization logic changes; behavior is intended to be identical aside from caching.
> 
> **Overview**
> Aligns **experimental `KTOTrainer` and `TPOTrainer`** with the dataset-fingerprint fix from #6206 so `dataset.map` during preprocessing can be cached across runs instead of re-tokenizing every time.
> 
> In both trainers, **`_tokenize` is now a `@staticmethod`**, and **`tokenize_fn` calls a local `tokenize = self._tokenize`** so the map callback no longer closes over `self` (which held the model and broke stable hashing). **KTO** also threads **`is_vlm` as an explicit argument** via `fn_kwargs`, matching DPO’s VLM handling.
> 
> Tokenized outputs are unchanged; only **cache fingerprint stability** improves.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf7b8e486b9118f0c5ea8fdaff82601bdaedc2e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/kto/kto_trainer.py
+++ b/trl/experimental/kto/kto_trainer.py
@@ -828,10 +828,11 @@ class KTOTrainer(_BaseTrainer):
                         self.args.precompute_ref_batch_size or self.args.per_device_eval_batch_size,
                     )
 
+    @staticmethod
     def _tokenize(
-        self,
         processing_class: PreTrainedTokenizerBase | ProcessorMixin,
         input: str | list,
+        is_vlm: bool,
         **kwargs,
     ) -> dict[str, list]:
         """Tokenize a single example for dataset preprocessing.
@@ -844,6 +845,9 @@ class KTOTrainer(_BaseTrainer):
                 The tokenizer or processor to use.
             input (`str` or `list`):
                 A string for non-conversational input, or a list of message dicts for conversational input.
+            is_vlm (`bool`):
+                Whether the processing class is a VLM processor, requiring multimodal message preparation and batch
+                dimension normalization.
             **kwargs:
                 Forwarded to `apply_chat_template` (e.g. `add_generation_prompt`, `return_assistant_tokens_mask`).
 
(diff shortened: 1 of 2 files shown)


albertvillanova:
Duplicate of:
- #6212
- #6213

PR #6238, opened 2026-07-01
Environment-owned reward

qgallouedec:
Implements the RFC in #5912: lets the environment own the reward. If an `environment_factory` environment defines a reserved `get_reward()` method (no args → `float`), it is called once per completed rollout and added as a reward source. So `reward_funcs` becomes optional.

> [!IMPORTANT]
> This adds no new capability. It's an ergonomics change. The same reward was always expressible through `reward_funcs`; this just lets the environment own it directly, which is the natural formulation for stateful environments.

This matters for stateful / multi-turn environments, where the reward is a function of the environment's internal state (was the word guessed? did the game end in a win?). Today that state has to be leaked back out to a trainer-owned `reward_func`; now the environment scores itself.

## Before

The reward lives in the trainer and reaches into the environment to recompute what it already knows:

```python
class WordleEnv:
    # Reserved methods (not exposed as tools)
    def reset(self, **kwargs):
        self._target = sample(words)
        self._solved = False

    # Public methods (exposed as tools)
    def guess(self, word: str) -> str:   # exposed as a tool
        self._solved = word == self._target
        ...

def solved_reward(environments, **kwargs):
    return [1.0 if env._solved else 0.0 for env in environments]

trainer = GRPOTrainer(
    model=model,
    reward_funcs=solved_reward,          # trainer owns the reward
    train_dataset=dataset,
    environment_factory=WordleEnv,
)
```

## After

The environment scores the episode it just ran:

```python
class WordleEnv:
   # Reserved methods (not exposed as tools)
    def reset(self, **kwargs):  # required
        self._target = sample(words)
        self._solved = False

    def get_reward(self) -> float:  # optional
        return 1.0 if self._solved else 0.0

    # Public methods (exposed as tools)
    def guess(self, word: str) -> str:   # exposed as a tool
        self._solved = word == self._target
        ...

trainer = GRPOTrainer(
    model=model,
    train_dataset=dataset,
    environment_factory=WordleEnv,       # owns the reward; reward_funcs becomes optional
)
```

## Details

- `get_reward` is reserved (like `reset`): it is **not** exposed to the model as a tool.
- Logged under the environment's class name (`rewards/WordleEnv/mean`), always weight 1: the environment owns its scale.
- Combines with `reward_funcs`: all sources are summed; `reward_weights` applies to `reward_funcs` only.
- Raises at init if no reward source is provided (no `reward_funcs` and no `get_reward`).
- Implemented for `GRPOTrainer`, `DPPOTrainer`, `GRPOWithReplayBufferTrainer`, and `AsyncGRPOTrainer`. Docs + tests included.

## AI writing disclosure

---

> [!NOTE]
> **Medium Risk**
> Changes core RL reward plumbing (optional `reward_funcs`, new reward columns, async scoring timing); mistakes could zero out or mis-weight rewards, though tests cover main paths.
> 
> **Overview**
> Adds **environment-owned rewards**: environments from `environment_factory` can implement optional `get_reward()` (reserved like `reset`, not exposed as a tool). The trainer calls it once per finished rollout and treats the score as an extra reward column named after the env class, **weight 1**, so **`reward_funcs` is optional** when `get_reward` exists.
> 
> **GRPO** registers env rewards at init (deduped by env class for dict factories; mixed env types get `None`/NaN for non-matching rollouts), sums them with trainer `reward_funcs` (`reward_weights` only apply to `reward_funcs`), and **fails fast** if neither source is provided. **`environments`** is always passed into reward kwargs. **Async rollout worker** scores `get_reward` after each rollout (before returning env to pool) and merges env columns at score time.
> 
> Same optional `reward_funcs` + `get_reward` docs on **AsyncGRPO**, **DPPO**, and replay-buffer GRPO; tool exposure excludes `get_reward`. Docs rename per-source metrics to `rewards/{name}/…` and expand agent-training guidance. New GRPO tests cover env-only reward, coexistence with `reward_funcs`, mixed factories, and missing reward source.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e30545d897455d772e739a63cd0e3563608f134. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -242,6 +243,7 @@ def __init__(
         tools = tools or []
         self._standalone_tools = tools  # tools that are not bound to the environment
         self.environment_factory = environment_factory
+        self._env_owns_reward = False
```
AmineDiro: small nit: owns reward points to an env that handles all the reward computation, but we can still pass in separate reward functions?

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -431,6 +446,10 @@ async def _generate_loop(self, stop_event: asyncio.Event) -> None:
                     group.tool_mask.append(tool_mask)
                     group.tool_call_counts.append(tool_call_count)
                     group.tool_failure_counts.append(tool_failure_count)
+                    # The environment owns the reward: score it now, while this rollout's environment still holds its
+                    # final state (it is reset only when drawn again for the next rollout).
+                    if self._env_owns_reward:
+                        group.env_rewards.append(environment.get_reward())
```
AmineDiro: Looking at this, we might need to support async reward also ( like LLM as judge) 
so we probably need an async version of the get_reward like `get_reward_async`. 

We also are in the main generation loop. So if a user passes a sync reward function that take time, this will halt the infllght requests  loop

AmineDiro (commented):
that's a very niice PR that will cleanup the API interface a lot !

docs/source/grpo_trainer.md
```diff
@@ -647,8 +647,16 @@ RapidFire AI is an open-source experimentation engine that sits on top of TRL an
 
 ## Agent Training
 
-GRPO supports **agent training** through the `tools` argument in [`GRPOTrainer`].
-This parameter expects a list of Python functions (sync or async) that define the tools available to the agent:
+GRPO supports **agent training**: the model calls tools during generation and learns from the outcome.
+
+- A **tool** is a plain Python function (sync or async) exposed to the model. Use `tools` for stateless calls (a calculator, a web search).
+- An **environment** is the more general form: a stateful object built fresh per rollout, whose public methods are exposed as tools, plus a `reset` lifecycle hook and an optional `get_reward` that lets it own the reward. Use `environment_factory` when you need per-rollout state, a reset hook, or environment-owned reward.
```
sergiopaniego, line 695: not sure if these ideas here of `get_reward` could collide with how rewards are retrieved from OpenReward/Harbor @adithya-s-k

sergiopaniego (commented):
this is great, thanks!
some parts that need to be revisited:
- OpenEnv guide
- Example scripts/notebooks using OpenEnv

trl/experimental/async_grpo/async_rollout_worker.py
qgallouedec: Right, renamed to `_env_has_reward` in 9b810a76. You're right that "owns" over-claims: `get_reward` is just one reward source among several and sums in alongside `reward_funcs`

trl/experimental/async_grpo/async_rollout_worker.py
qgallouedec: Addressed in ab5ec489: `get_reward` can now be `async def`, so an LLM-judge reward is awaited instead of blocking the loop.

In the async worker it's awaited before the env returns to the pool (so a concurrent rollout can't reset it mid-await); in sync GRPO it routes onto the daemon event loop like any other async reward func. A slow sync `get_reward` still blocks (can't thread it off since the env gets reused) so I/O scoring should be `async def`.

AmineDiro: Ok great 👍🏼 

can't we run get_reward in a threadpool executor for the sync versions  ?

docs/source/grpo_trainer.md
qgallouedec: `get_reward` is purely additive and opt-in: when an env defines it, it's registered as one more reward source that sums in (weight 1) alongside` reward_funcs`. It doesn't intercept or replace how rewards are fetched elsewhere. So a reward coming from OpenReward/Harbor stays a normal `reward_funcs` entry and the two compose (rather than collide). I will assume that matches how Harbor surfaces rewards

trl/experimental/async_grpo/async_rollout_worker.py
qgallouedec: We can't safely offload sync `get_reward` to a threadpool here: the env instance is pooled and `reset()` between rollouts, and `get_reward` scores from its final state. If we returned control (thread or await) before scoring without holding the instance, a concurrent rollout could draw it and reset it mid-scoring. That's why the env is returned to the pool only after scoring, and why a slow sync `get_reward` blocks by design. I/O-bound scoring (LLM judge, etc.) should be `async def`, which we do await in place.

trl/trainer/grpo_trainer.py
```diff
@@ -611,9 +621,40 @@ def __init__(
...
+
+                    else:
+
+                        def get_reward(environments, _env_type=env_type, **kwargs):
+                            return [e.get_reward() if isinstance(e, _env_type) else None for e in environments]
+
+                    self.reward_funcs.append(get_reward)
+                    self.reward_func_names.append(env_type.__name__)
+                    self.reward_processing_classes.append(None)
+                    self.reward_weights = torch.cat([self.reward_weights, torch.ones(1)])
```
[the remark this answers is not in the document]
qgallouedec: Good catch, fixed in 7a25a49d: reward columns are now deduplicated by env class

trl/trainer/grpo_trainer.py
```diff
@@ -611,9 +623,36 @@ def __init__(
...
+            if has_reward:
+                if inspect.iscoroutinefunction(reward_env_type.get_reward):
+
+                    async def get_reward(environments, **kwargs):
+                        return [await environment.get_reward() for environment in environments]
+
+                else:
+
+                    def get_reward(environments, **kwargs):
+                        return [environment.get_reward() for environment in environments]
```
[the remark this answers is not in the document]
qgallouedec: Fixed in a1c0ed6a: instead of one global env-reward, each env class that defines `get_reward` gets its own reward column, `isinstance`-filtered so a rollout is scored only when its env is that class. A multi-env run that mixes a reward-owning env with a plain one now returns `None` (→ NaN, ignored) for the plain one instead of raising `AttributeError`.

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -462,6 +478,16 @@ async def _generate_loop(self, stop_event: asyncio.Event) -> None:
...
                     group.tool_call_counts.append(tool_call_count)
                     group.tool_failure_counts.append(tool_failure_count)
+                    # The environment owns the reward: score it now, while this rollout's environment still holds its
+                    # final state and before returning it to the pool. `get_reward` may be async awaiting yields to
+                    # inflight requests instead of halting them. The env is returned to the pool only after scoring, so
+                    # a concurrent rollout can't draw and reset it during the await.
+                    if self._env_has_reward:
+                        get_reward = environment.get_reward
+                        reward = await get_reward() if inspect.iscoroutinefunction(get_reward) else get_reward()
+                        group.env_rewards.append(reward)
```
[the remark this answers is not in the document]
qgallouedec: False positive: `inspect.iscoroutinefunction` returns `True` for a bound `async def` method (`inspect.iscoroutinefunction(instance.get_reward) == True`), so the coroutine is awaited. Verified.

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -431,6 +444,16 @@ async def _generate_loop(self, stop_event: asyncio.Event) -> None:
...
+                    if environment is not None:
+                        self._environment_pool.append(environment)
```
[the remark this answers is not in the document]
qgallouedec: Not a live leak: a failing rollout task re-raises out of _generate_loop (and a raising get_reward propagates the same way), which tears the whole rollout loop down via its finally. There's no continued operation left to starve the pool, so returning the instance on the error path would be dead cleanup.

trl/trainer/grpo_trainer.py
```diff
@@ -569,10 +579,37 @@ def __init__(
...
             self.tools = tools + methods
+
+            # Expose the environment's `get_reward` as an extra reward source (named after the env class, weight 1).
+            # `get_reward` may be async (e.g. an LLM judge); wrap it accordingly so `_calculate_rewards` runs it on the
+            # daemon event loop like any other async reward func.
+            if has_reward:
+                if inspect.iscoroutinefunction(instance.get_reward):
+
+                    async def get_reward(environments, **kwargs):
+                        return [await environment.get_reward() for environment in environments]
```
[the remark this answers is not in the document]
qgallouedec: False positive: environments is added to `reward_kwargs` unconditionally (`_calculate_rewards`), and that dict is passed via `**reward_kwargs` to both the sync call and the async _invoke_async path. The async wrapper receives environments fine.

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -462,6 +481,21 @@ async def _generate_loop(self, stop_event: asyncio.Event) -> None:
...
+                    # a concurrent rollout can't draw and reset it during the await. Record `(env class, reward)` so
+                    # `_score_group` can place it in the matching env's reward column; rollouts whose env owns no reward
+                    # record `None` (turned into NaN and ignored) to stay aligned with the group's other per-rollout lists.
+                    if self._env_reward_types:
+                        if type(environment) in self._env_reward_types:
+                            get_reward = environment.get_reward
+                            reward = await get_reward() if inspect.iscoroutinefunction(get_reward) else get_reward()
+                            group.env_rewards.append((type(environment), reward))
+                        else:
+                            group.env_rewards.append(None)
```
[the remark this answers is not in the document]
qgallouedec: this was an inconsistency between the two trainers. Fixed in 2b0b60aa

PR #6240, opened 2026-07-02
Fix BrowserGym example dependencies

burtenshaw:
Updates the BrowserGym OpenEnv example dependency headers to install the actual Space package and include jmespath for GRPO tool parsing.

### DIFF
--- a/examples/scripts/openenv/browsergym.py
+++ b/examples/scripts/openenv/browsergym.py
@@ -17,7 +17,8 @@
 #     "trl[vllm,peft]",
 #     "trackio",
 #     "kernels",
-#     "openenv-browsergym @ git+https://huggingface.co/spaces/openenv/browsergym_env",
+#     "jmespath",
+#     "openenv-browsergym-env @ git+https://huggingface.co/spaces/openenv/browsergym_env",
 # ]
 # ///
 
@@ -30,7 +31,7 @@ to see the page visually after each action.
 
 Setup:
 ```sh
-pip install "openenv-browsergym @ git+https://huggingface.co/spaces/openenv/browsergym_env"
+pip install "openenv-browsergym-env @ git+https://huggingface.co/spaces/openenv/browsergym_env"
 ```
 
 Usage:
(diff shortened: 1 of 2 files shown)


sergiopaniego (changes requested):
thanks! 
the branch is outdated and some of the changes are already in main. could you update your branch? 
the `jmespath` is the only change

albertvillanova (commented):
Thanks @burtenshaw! The rename half of this already landed in #6117 back in June, which is why the branch now conflicts — after a rebase only the `jmespath` line survives, as @sergiopaniego pointed out. That part is a genuine bug: GRPOTrainer raises ImportError when `tools`/`environment_factory` is set and jmespath is missing, and jmespath currently lives only in the `dev` extra, so a plain `uv run` on these scripts fails before training starts. But the same is true of all 10 openenv examples that pass `environment_factory=` (carla*, catch, echo, multi_env, sudoku, wordle), not just the two BrowserGym ones — so patching 2 of 10 leaves the rest broken. Rather than adding the line to 10 PEP 723 headers, ~~would you mind promoting `jmespath` to an actual dependency instead, so every example is covered at once~~? Happy to take it either way, but the per-header version will drift again the next time someone adds an env example. What do you think?

EDIT: not sure about my comment above!

albertvillanova:
Thanks again @burtenshaw, and thanks for your patience here. Looking into the `jmespath` half more closely, it turns out the examples were not missing a dependency after all: `jmespath` is only used by the legacy `response_schema` parser, which is all that transformers below 5.13.0 ships. From 5.13.0 on, TRL sets a `response_template` and the new-style parser is used, which never imports it (transformers dropped its own usage in huggingface/transformers#47320). The guard in `GRPOTrainer` was unconditional, so it was refusing to train over a library that plays no part in it on any current transformers.

So rather than declaring the dependency, #6597 narrows the guard to the legacy path. All 10 `openenv` examples that pass `environment_factory` then work with no header changes at all, and there is nothing for the next env example to remember. It also fixes the same over-broad condition in the `require_jmespath` test marker, which was silently skipping the tool tests that would have caught this.

Since the `openenv-browsergym-env` rename already landed in #6117, nothing is left of this branch, so I think we can close this in favour of #6597. Your report is what surfaced the real bug, so thanks a lot for raising it.

PR #6250, opened 2026-07-02
 implement message level rollout with linear trajectories

AmineDiro:
# AsyncGRPO: message-mode rollouts

Adds an opt-in way to build training rows from a multi-turn conversation.

Message mode keeps the conversation as messages and **re-tokenizes the whole thing each turn**, then checks whether the fresh tokens still start with the tokens held so far. If yes → append the new part (same as token mode). If no → a rewrite happened → close the row and open a new one that matches what the model actually read.

## config

```python
AsyncGRPOConfig(
    rollout_protocol="message",   # "token" (default) | "message"
    fork_threshold_tokens=1024,   # message mode only
)
```

## How rows are built

Heavily inspired by slime.
Each turn is a `TurnRecord(prompt_ids, output_ids, output_log_probs)`. At the end, `_chain_to_sequences` walks the turns and per turn classifies the drift vs. the tokens held so far:

- **CLEAN** — new prompt starts with held tokens → append (prompt/tool = context, generated = trained).
- **REALIGN** — only the last answer's tail wobbled and the new turn < `fork_threshold_tokens` → overwrite that tail as context, same row.
- **FORK** — a real rewrite → start a new row.

**Advantage:** one per conversation, stamped on every row it produced, no split. Under the token-mean loss a fork is invisible (each generated token trained once, same advantage, same denominator).

## Next work: Tree trajectories

The design already covers the two hard parts:

- **Scoring is branch-agnostic.** Every `TrainingSequence` carries a `rollout_id` and `_score_group` groups by it, not by list position. N rows per conversation already works today; a tree that yields several rows per conversation needs no scoring change.
- **The reconciler is tree-agnostic.** `_common_prefix_len` / `_SampleBuilder` / `_chain_to_sequences` reconcile a single linear chain of turns. A root→leaf path in a tree *is* such a chain.

So the only change a tree adds is on the rollout side: replace the flat `turns: list[TurnRecord]` (+ one `_chain_to_sequences` call) with a **tree of turns**, then run the reconciler once per root→leaf path, with all rows sharing the conversation's `rollout_id`. Recording would carry the turn's message context so each turn can be placed under its parent node (this is what lets shared prefixes branch and stay trained once). Nothing downstream (collator, loss, scoring) changes.

---

> [!NOTE]
> **Medium Risk**
> This changes how multi-turn and tool-calling trajectories become training tensors and which tokens keep loss signal when chat templates rewrite history—high impact on RL training correctness, though scoped to experimental async GRPO.
> 
> **Overview**
> **AsyncGRPO multi-turn rollouts** now keep the conversation as messages and **re-tokenize the full history each turn**, instead of incrementally appending generated/tool token suffixes. After rollout, turns are reconciled into one or more **`TrainingSequence`** training rows via **`_chain_to_sequences`** (CLEAN append, REALIGN small tail wobble, **FORK** on larger re-tokenization drift).
> 
> Adds **`fork_threshold_tokens`** (default 1024) on **`AsyncGRPOConfig`**, wired through the rollout worker and trainer. **`RolloutGroup`** now stores **`completions_sequences`** (per-conversation list of rows) rather than separate **`prompt_ids` / `tool_mask` / per-token logprob lists**; **`_score_group`** emits **one `RolloutSample` per row** with the **same conversation-level advantage** on every forked row. Tool turns are fed back as messages and picked up on the next full template pass—**`_get_tool_suffix_ids`** is removed.
> 
> Large **unit test** coverage for the reconciler, **`_generate_one`**, and forked scoring behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4b7fca382d4b9211a8b9cc1f3902960cf74b112. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_grpo_config.py
```diff
@@ -56,6 +56,15 @@ class AsyncGRPOConfig(_BaseConfig):
...
             `max_completion_length`.
+        rollout_protocol (`str`, *optional*, defaults to `"token"`):
+            How a multi-turn conversation is turned into training rows. `"token"` grows a token buffer, appending each
+            turn's generated tokens and tokenized tool results (fast; cannot represent a conversation rewrite).
+            `"message"` re-tokenizes the whole conversation every turn and reconciles the result against the tokens held
+            so far: a clean append stays one row, a rewrite (dropped reasoning, summarized history) forks a new row.
+        fork_threshold_tokens (`int`, *optional*, defaults to `1024`):
+            Message mode only. When a turn's re-tokenized prompt drifts inside the last generated answer, a drift with a
+            generated turn shorter than this many tokens is treated as a re-tokenization wobble (realigned as context)
+            rather than a rewrite (a new row). Ignored when `rollout_protocol="token"`.
```
qgallouedec: I'm going to advocate for simplicity here: because it might become impossible to maintain if we continue to support every possible configuration. What do you think about supporting only the `message` protocol? This trainer is still experimental; we don't need to be backward compatible, so it's a good time to make bold and radical decisions.

qgallouedec (commented):
cool, thanks! discussed internally. I'm sharing the figure I made 

Image

trl/experimental/async_grpo/async_grpo_config.py
AmineDiro: I was thinking the same thing. Only thing that bothers me is the performance penalty of supporting this `messages` mode :/  Need to measure this to be sure 🫡 .

But I agree with the idea 1000%

tests/experimental/test_async_grpo_message_loop.py
qgallouedec, line 1: to follow the repo structure, we should have only one `test_async_grpo_trainer.py`

trl/experimental/async_grpo/async_grpo_config.py
qgallouedec: yes I understand the concern. With `tokens`, you can only get `G` sequences for 1 prompt. For `messages`, you end up with, worst case scenario `G*max_num_turns`

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -761,6 +857,60 @@ async def _do_post():
...
+            turns.append(TurnRecord(prompt_ids, turn_ids, turn_logprobs))
+            tool_calls = assistant_message.get("tool_calls")
+            if tool_calls is None or (max_iterations is not None and iteration_num >= max_iterations):
+                break
+            tool_messages, n_calls, n_failures = self._execute_tool_calls(tool_calls, tool_dict)
+            tool_call_count += n_calls
+            tool_failure_count += n_failures
+            completion.extend(tool_messages)
+            messages.extend(tool_messages)  # tool result goes back as a MESSAGE, re-tokenized next turn
+            iteration_num += 1
```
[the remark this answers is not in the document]
qgallouedec: valid defensive concern, real hang path, but a genuine edge case.
Worth a one-line `if not tool_calls`: in both loops?

AmineDiro:
cc. @qgallouedec 

Run a little bit of experiments to answer the question: can we drop token mode altogether? 

### Setup

- Two ways the rollout worker turns a multi-turn conversation into training rows:
  - `token`: append-only buffer, so it always produces **1 row per conversation**.
  - `message`: re-tokenizes the whole conversation each turn and reconciles drift, producing **1 to T rows** (it
    forks when the history is rewritten).
- Deterministic, CPU / single-GPU, **no vLLM**: generation is **scripted** so both protocols see identical model
  outputs, and only the bookkeeping differs. The multi-turn case uses a reasoning model that drops prior `<think>`
  on re-render (verified for Qwen3).
- The real trainer step is measured with a **stub rollout worker** that feeds pre-shaped rows into the real
  trainer (real forward/backward/optimizer), 1×H100, Qwen3-0.6B.

### Results

**1. Re-tokenization overhead: negligible except at very large context**
- message re-tokenizes the whole conversation every turn, which is O(T²). It is about 4% of the generation time at 8
  turns, but about 56% at 128 turns / 69k-token context. Cheap to fix later with a prefix cache if it ever matters 🤔 

fig_timing

**2. Row multiplication: depends on turn length vs `fork_threshold`**
- long turns (at or above the threshold) **fork**, producing up to T rows. Trained tokens are **unchanged**
  (fork invariance), but forward (context) tokens blow up.
- short turns (below the threshold) **realign**: the conversation stays 1 row, but **intermediate turns' training
  signal is silently dropped**.

**3. Real trainer step (FORK regime)**
- both protocols are capped to roughly equal forward tokens per step, so this is a fair, matched-compute comparison.
- message rows are mostly masked context, so they spend about **10 forward tokens per trained (gradient) token**
  versus about 1.15 for token, at 16 turns.
- **trained tokens / sec:** token holds around 8k to 10k (flat in turns); message drops from 8.3k to 2.1k, so
  message needs **1.2x (2 turns) to 3.9x (16 turns)** more GPU wall-time to train.
- peak memory is comparable (about 12 to 13.7 GB), so the penalty is time, not memory.

fig_trainer_step


### Conclusion

- Re-tokenization overhead is not a reason to keep `token`.
- FORK cost is bounded: about 1.2x to 3.9x GPU time for deep rollouts, memory unchanged, trained-token count identical. So this can be really problematic if we don't need a setup where messages are being rewritten
- The knob to tune is **`fork_threshold`** (forking costs more compute, realigning drops signal), not the protocol.
- `token` mode doesn't even work for multi-turn reasoning + tools today.

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -66,15 +69,113 @@ async def _retry_on_http_error(coro_factory: Callable[[], Awaitable], *, label:
...
+        self.logprobs: list[float] = []
+        self.last_response_start_idx: int | None = None
+
+    def classify_token_drift(self, turn: TurnRecord) -> DriftKind:
+        matched = _common_prefix_len(self.tokens, turn.prompt_ids)
+        drift = len(self.tokens) - matched
+        if drift == 0:
+            return DriftKind.CLEAN
+        start = self.last_response_start_idx
+        if start is not None and matched >= start and len(turn.output_ids) < self._fork_threshold:
```
qgallouedec: the quantity that matters is the drift size, ie how much previously-trained content `REALIGN` is about to mask (`len(self.tokens) - matched`).

Failure case = the motivating Qwen3 one: a long reasoning answer dropped by the template, followed by a short next turn (< 1024), gets `REALIGN`ed → the long reasoning turn is silently masked to context and its training signal lost, when FORK is what preserves it. Suggest thresholding on drift size (or the previous response length), not `len(turn.output_ids)`.

```txt
held after turn 1:  [prompt] [<think> 800-tok reasoning </think> answer]   <- all mask=1
                            ^start                                    ^end

turn 2 re-tokenizes: template drops the <think>...</think> block
  prompt_ids = [prompt] [answer] [tool_result] [gen_prompt]
  matched = start (diverges right where reasoning began)
  drift   = ~800   ->  huge, we're invalidating 800 trained tokens

turn 2 output = "ok" (2 tokens) < 1024
  -> classify_token_drift returns REALIGN
  -> _align_to_prompt masks tokens[start:] : the 800-tok reasoning turn is overwritten as context and never trained.

want: FORK -> row1 keeps the reasoning answer trained,
              row2 = rewritten history (no reasoning) + turn-2 answer trained.
```

Gating on `drift = len(self.tokens) - matched` (or the last response's length) classifies this as `FORK`; `len(turn.output_ids)` can't see it because the incoming turn is tiny.

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -702,50 +767,44 @@ async def _score_group(self, group: RolloutGroup) -> list[RolloutSample]:
...
+                samples.append(
+                    RolloutSample(
+                        prompt=prompt,
+                        completion=completion,
+                        input_ids=seq.input_ids,
+                        completion_mask=seq.completion_mask,
+                        old_log_probs=seq.old_log_probs,
+                        advantage=float(advantage),
+                        model_version=group.model_version,
+                        metrics=dict(metrics),  # own copy per row; _compute_rollout_metrics mutates it
```
qgallouedec, line 870: Copying metrics onto every row means a conversation that forks into k rows contributes its `reward`/`reward_std`/`rewards/*` k times to per-sample averaging (and `print_prompt_completions_sample` prints it k times). Loss/advantage stay correct, but logged reward skews with fork rate. slime avoids this by de-duping on `rollout_id`; consider aggregating metrics per-conversation. It doesn't have to be fixed in this PR, but pls be aware of this

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -427,11 +529,10 @@ async def _generate_loop(self, stop_event: asyncio.Event) -> None:
                     group.prompt_ids.append(prompt_ids)
```
qgallouedec: `group.prompt_ids` and the returned `turns[0].prompt_ids` no longer feed `input_ids` (rows carry their own). Appears dead now.

qgallouedec (approved):
lgtm, there is only one major blocking concern imo, otherwise we're good to move forward. Massive effort, well done!!

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: Good catch 👍🏼

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: Ah, good catch, that was a bug in my reasoning 👍🏼  drif is the correct heuristic to comparte to the `fork_threshold`  I also fixed ` _align_to_prompt` to not override from start but from the matched because we check that `matched >= start`

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: right, will be fixed in next PR 👍🏼

adithya-s-k (approved):

adithya-s-k:
Looks really good ! great work! 
Will try it out with Harbor and OpenEnv and let you know how things go. Overall LGTM.

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -656,6 +656,7 @@ def __init__(
                     max_tool_calling_iterations=self.args.max_tool_calling_iterations,
                     log_completions=self.args.log_completions,
                     num_completions_to_print=self.args.num_completions_to_print,
+                    fork_threshold_tokens=self.args.fork_threshold_tokens,
```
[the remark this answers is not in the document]
AmineDiro: This is a real concern, and will be addressed in followup PR

PR #6269, opened 2026-07-03
Async grpo vllm client

AmineDiro:
## What does this PR do?

Follow-up to #6218. 

Adds a small `VLLMClient` single place that speaks HTTP to the vLLM server — and uses it to tidy up two things #6218 left rough.

### Changes
- **`VLLMClient`** (new): named methods for each sync call to the server (`wait_for_server_ready`, `get_max_model_len`, `get_world_size`, `pause`/`resume`, the weight-update endpoints). Stateless so it's picklable. Generation is deliberately *not* routed here, it stays on its own async `aiohttp` session in the rollout child process.
- **`token_budget` default** now goes through `VLLMClient`: it waits for `/health` before `get_max_model_len()`, so a still-loading vLLM waits instead of failing the run 
- **Weight sync decoupled from rollout production**: `weight_transfer` is now built from config independent of `rollout_worker`, so a custom rollout worker still gets weight sync (it was previously force-set to `None`). Made injectable via a new `WeightTransferProtocol`.

### Tests
- `TestPackingAwareBatching` pass; the stub-worker tests inject a no-op weight transfer so they run without a live vLLM server.

---

> [!NOTE]
> **Medium Risk**
> Touches the async GRPO training path (vLLM readiness, default token budget, and NCCL weight sync wiring); behavior changes for custom rollout workers that previously skipped weight sync, though defaults preserve NCCL sync when not overridden.
> 
> **Overview**
> Introduces **`VLLMClient`** as the shared synchronous HTTP layer for the configured vLLM server (readiness polling, `max_model_len`, pause/resume, and NCCL weight-update endpoints). **`WeightTransferClient`** now delegates to that client instead of scattering `requests` calls and duplicate “wait for health” logic.
> 
> **`AsyncGRPOTrainer`** always builds a `VLLMClient` on the main process from config. Weight sync is **decoupled from rollout production**: a custom `rollout_worker` no longer disables weight transfer; callers can inject a **`WeightTransferProtocol`** implementation (including a no-op) via the new `weight_transfer` argument. When `token_budget` is unset, the trainer **waits for `/health`** before defaulting to `max_model_len`, so a slow-starting vLLM is less likely to fail dataloader setup.
> 
> Tests add **`_StubWeightTransfer`** so stub-worker trainer runs exercise the weight-sync callbacks without a live server.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e91b1ed44ae05a49fd0436402e9430df306f7a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/vllm_client.py
```diff
@@ -0,0 +1,100 @@
...
+                    "Make sure the vLLM server is running and reachable. If the server needs more time to load "
+                    "the model, increase `vllm_server_timeout` in your AsyncGRPOConfig."
+                )
+            if int(elapsed) % 10 < poll_interval_s:
+                logger.info(f"Still waiting for vLLM server... ({elapsed:.0f}s)")
+            time.sleep(poll_interval_s)
+
+    def get_max_model_len(self) -> int:
+        """Return the served model's `max_model_len` (the cap on prompt + completion tokens)."""
+        response = requests.get(f"{self.server_url}/v1/models", timeout=self.server_timeout)
```
qgallouedec:
```suggestion
        response = requests.get(f"{self.server_url}/v1/models")
```

I think this timeout is only for setting up the server

qgallouedec (approved):
I think it looks good, just a nit

PR #6280, opened 2026-07-03
Clarify stream-only dedupe check

kashif:
Makes the storage dedupe check explicitly streams-only. Follow up from #6241

This matches the existing comment: `fwd_stash` gives streams mode a clear release point, while single-stream mode does not.

## Selective activation checkpointing (SAC) for SFT

Also adds selective activation checkpointing to SFT. When gradient checkpointing is on, the attention output is saved during the forward pass instead of being recomputed in the backward pass. At long context this recovers most of the checkpointing slowdown for one extra hidden-state sized tensor per layer, and it runs fully eager with no torch.compile needed.

Turn it on with `selective_activation_checkpointing=True` on `SFTConfig` (needs `gradient_checkpointing=True`, which SFT already defaults to). It works by wrapping the model's `gradient_checkpointing_enable` to inject a policy that saves attention and recomputes the rest, and it forces non-reentrant checkpointing which SAC relies on. This is the same eager SAC approach torchtitan uses under FSDP2. PEFT plus DeepSpeed ZeRO-3 is rejected up front since that combo needs reentrant checkpointing.

Checked that SAC gradients match full checkpointing exactly and that attention is no longer recomputed in backward. Added a test for both.

This is the SFT-only first step. DPO, KTO, reward and GRPO can follow the same two lines in a later PR.

---

> [!NOTE]
> **Medium Risk**
> SAC changes gradient-checkpointing behavior on the training path (non-reentrant only) and is incompatible with PEFT+ZeRO-3; the streams-only dedupe tweak could alter offload behavior in single-stream mode.
> 
> **Overview**
> This PR adds **selective activation checkpointing (SAC)** for SFT and tightens activation-offloading dedupe behavior.
> 
> **SAC:** New `enable_selective_activation_checkpointing()` wraps the model’s `gradient_checkpointing_enable` so checkpointing uses a dispatcher policy that **saves attention outputs** (SDPA / flash-attn markers) and **recomputes other ops**, with **non-reentrant** checkpointing forced. Enable via `SFTConfig.selective_activation_checkpointing=True` (requires `gradient_checkpointing=True`); `SFTTrainer` applies the wrapper before the base trainer init and **rejects** SAC with PEFT + DeepSpeed ZeRO-3 (reentrant vs non-reentrant conflict). Tests assert SAC gradients match full checkpointing and that SDPA runs fewer times in backward.
> 
> **Offloading:** Storage-pointer deduplication during offload now runs only when **`use_streams`** is on, matching streams-mode `fwd_stash` release semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba5c2a8368b67b1fadac7713d5678e6e5dd049b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
cc @winglian

qgallouedec:
Thanks @kashif. Can you split this one? The dedupe change and SAC are independent, and bundling them is part of why it has been sitting. You can use the new stacking pr feature.

On the dedupe check: making it streams-only changes behavior in single-stream mode, since storages that were deduped will now be offloaded again. I cc'd winglian back in July and we never heard back, so that half needs either his confirmation or a test that actually shows the memory behavior.

On SAC, two things before I can approve:

- The new test class is missing `@require_torch_accelerator`. The SDPA count is backend dependent, so it will be flaky on CPU runners.
- `assert sdpa_sac < sdpa_full` is too weak I think. SAC saves every layer's attention, so i'd assert the exact count instead. As written it passes even if a single layer gets saved.

Also, `enable_selective_activation_checkpointing` does not really belong in `activation_offloading.py`. Offloading moves activations to CPU, SAC changes what gets recomputed. They only share the word activation, no?

kashif:
Split into #7193 (dedupe only), #7194 (SAC), and #7195 (two stream-sync fixes found while reviewing this, related to what axolotl works around). Closing this one in favor of those.

PR #4509, opened 2025-11-11
`device_map` and `dtype` to `"auto"` by default

qgallouedec:
This PR does two things:

- It sets `device_map` to `"auto"` by default. It avoids loading the model in RAM then moving it to VRAM. It's basically free lunch (either as fast or faster); here is a small benchmark for GRPO initialization:

```
Qwen/Qwen3-VL-8B-Instruct 1 GPU 
Before: 11.1 seconds
After: 6.5 seconds

Qwen/Qwen3-VL-8B-Instruct 4 GPUs and FSDP2 
Before 15.4 seconds
After 14.4 seconds

Qwen/Qwen3-14B 1 GPU
Before 19.1 seconds
After 8.2 seconds

Qwen/Qwen3-14B 4 GPU and FSDP2
Before 21.4 seconds
After 21.8 seconds

Qwen/Qwen3-32B 4 GPU and FSDP2
Before 42.1 seconds
After 41.9 seconds
```

2. I sets `dtype` to `"auto"` by default.

I always felt it's strange to have a model in bfloat16, and when you load it, you get it in float32:

Screenshot 2025-11-11 at 3 53 05 PM

```python
>>> from transformers import AutoModelForCausalLM
>>> AutoModelForCausalLM.from_pretrained("Qwen/Qwen3-14B").lm_head.weight.dtype
torch.float32
>>> AutoModelForCausalLM.from_pretrained("Qwen/Qwen3-14B", torch_dtype="auto").lm_head.weight.dtype
torch.bfloat16
```

this PR sets `"auto"` as default for the dtype.

PR #6295, opened 2026-07-05
Respect use_cpu in create_model_from_path device_map default

Strongich:
# What does this PR do?

`use_cpu=True` is silently ignored when a trainer loads a model from a string on a machine with a visible GPU: the model lands on `cuda:0` while the `accelerator` (and anything placed through it) stays on CPU. For single-model trainers this just means `use_cpu` doesn't do what it says; for two-model trainers (GOLD, GKD-style distillation) it splits student and teacher across devices and crashes the loss.

## How I ran into it
After reinstalling my dev env, my CUDA device became visible to the tests during the #5969 implementation, and this started failing on the feature branch, so I checked `main` too:

```text
FAILED tests/experimental/test_gold_trainer.py::TestGOLDTrainerLoss::test_loss_normalizes_by_num_items_in_batch -
RuntimeError: Expected all tensors to be on the same device, but got index is on cuda:0,
different from other tensors on cpu
(when checking argument in method wrapper_CUDA__index_select)
```

The test builds a `GOLDTrainer` with `use_cpu=True`. On a CPU-only machine (like CI) it passes; on any machine with a GPU it crashes. Same on `main` — you can reproduce with:

```shell
pytest tests/experimental/test_gold_trainer.py::TestGOLDTrainerLoss::test_loss_normalizes_by_num_items_in_batch
```

on a GPU host, and it passes again with `CUDA_VISIBLE_DEVICES=""`.

## Root cause
`create_model_from_path` unconditionally defaults `device_map="auto"`:

```python
kwargs["device_map"] = kwargs.get("device_map", "auto")
```

The callers only override it to `None` for `MULTI_GPU`/`DEEPSPEED`. A `use_cpu=True` (or plain single-GPU) run is neither, so with a visible CUDA device the model is dispatched to `cuda:0` and stamped with `hf_device_map={'': 0}`. That makes `Trainer.__init__` treat it as device-mapped (`is_model_parallel=True`, `place_model_on_device=False`), so the Trainer deliberately never relocates it to the (CPU) accelerator device.

The chain for GOLD specifically:
- student: `create_model_from_path` → `device_map="auto"` → `cuda:0` with `hf_device_map={'': 0}`; transformers' `Trainer` treats any `hf_device_map` on a device ≠ `args.device` as model-parallel and refuses to move it
- teacher: also loaded on `cuda:0`, but GOLD passes it through `accelerator.prepare_model(...)`, and `accelerate` treats a single-entry device map as movable (`verify_device_map` only pins maps with >1 entry) → relocated to `cpu`, which is where `use_cpu=True` wanted everything in the first place
- `compute_loss` feeds the same batch to both → device mismatch → `RuntimeError`

## The fix
Make the `device_map` default CPU-aware in `create_model_from_path` (the shared loader all trainers use), instead of patching the ~20 duplicated caller sites:

```python
if "device_map" not in kwargs:
    kwargs["device_map"] = None if PartialState().device.type == "cpu" else "auto"
```

By the time trainers call this, `args.distributed_state` is already initialized, so `PartialState()` reflects `use_cpu`. Behavior is unchanged otherwise: non-CPU devices keep `"auto"`, and callers that explicitly pass `device_map` (e.g. `None` for `MULTI_GPU`/`DEEPSPEED`) bypass the default entirely. As a bonus, the teacher/ref-model paths no longer do a wasteful load-to-GPU-then-move-to-CPU round trip under `use_cpu=True`.

Verified on a GPU host:
- the failing GOLD test passes (student and teacher both on CPU)
- GPU training without `use_cpu` still resolves to `device_map="auto"` → `cuda:0`, as before

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can ignore the checklist below if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md), Pull Request section?

## AI writing disclosure
We welcome the use of AI tools to help with contributions. To help us improve our review process, please indicate the level of AI involvement in this PR.
- [ ] No AI usage: the PR was written entirely by a human
- [ ] AI-generated: the PR was mostly or fully generated

## Who can review?
Anyone in the community is free to review the PR once the tests pass. Feel free to tag members/contributors who may be interested in your PR.

@qgallouedec 

---

> [!NOTE]
> **Medium Risk**
> Touches the shared model loader used by all trainers; behavior change is narrow (CPU-only runs on GPU hosts) but affects where weights land at load time.
> 
> **Overview**
> **Fixes `use_cpu=True` being ignored** when loading models from a path on machines with a visible GPU. `create_model_from_path` no longer always defaults `device_map` to `"auto"`; it sets **`device_map=None` on CPU** (via `PartialState().device`) and keeps **`"auto"` otherwise**, only when the caller did not pass `device_map`.
> 
> Previously, `"auto"` could place weights on `cuda:0` while the Trainer/accelerator stayed on CPU, which broke **`Trainer` placement** (device-mapped models are not moved) and caused **student/teacher device splits** in distillation trainers (e.g. GOLD), leading to loss-time `RuntimeError`s. Explicit `device_map` overrides (e.g. `None` for multi-GPU/DeepSpeed) are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d32ff4dc7d56c01ac83c23c72b0b60403819cace. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
Do you train on cpu?

Strongich:
@qgallouedec I see ur point, fair :D Nevertheless, since this parameter exists in `TrainingArguments` and both accelerate and transformers "honor" it, this gap from the TRL perspective feels a bit awkward: accelerate places the trainer on CPU, but `create_model_from_path` dispatches the model to GPU. I think the real use case for this parameter is not about training on CPU, but rather smoke-testing something (like I did) without hitting that crash.

Strongich:
Following the case #6504 fixes, I think it's a good time to decide on this PR -- covering the reverse case, where a CUDA device is available but the user wants to, for example, smoke-test something on CPU, and hits an error instead. (This will matter more once the `DistillationTrainer `moves to a stable release, where other users could run into the same issue I did.)

So, your call -- if you'd rather keep the current default, you can close it. Either way, happy to have it decided.
@qgallouedec

albertvillanova (approved):
Thanks.  I'd merge it. Some notes:
- This isn't about CPU training: it's about scoping an optimization to the case it was written for. `device_map="auto"` was introduced in #4509 explicitly as a load-speed optimization: "It avoids loading the model in RAM then moving it to VRAM", with a benchmark table to back it. When `args.device` is CPU there is no VRAM to skip, so `"auto"` delivers exactly zero of that benefit and only costs us: the model lands on `cuda:0` with `hf_device_map={'': 0}`, `Trainer` then treats it as model-parallel and refuses to relocate it, and two-model trainers end up split across devices.
- Second argument: this is a test-suite fix for GPU dev machines. #6504 fixed the CPU-only-machine direction (autouse fixture forcing use_cpu=True when no accelerator is present). It is a no-op on a GPU host. The two PRs are complementary, and this one becomes more relevant once DistillationTrainer graduates out of experimental.

PR #6297, opened 2026-07-05
fix(grpo): handle NaN/None vLLM logprobs in importance-sampling correction

shadowmodder:
# What does this PR do?

Fixes a crash in `GRPOTrainer` when `vllm_importance_sampling_correction=True` and vLLM emits `NaN` logprobs for near-deterministic tokens. `extract_logprobs()` converts NaN to `None`, and `torch.tensor([..., None, ...])` raises `RuntimeError: Could not infer dtype of NoneType`.

**Root cause:** vLLM returns `NaN` log-probabilities for tokens whose probability is effectively 1.0 (log p ≈ 0 underflows to -0 or NaN depending on numerical precision). `extract_logprobs()` converts these NaN values to `None` as a sentinel. When the importance-sampling correction path in `GRPOTrainer` calls `torch.tensor()` on a list that contains `None`, PyTorch cannot infer the dtype and raises a `RuntimeError`.

**Fix:** convert `None → float("nan")` before tensor construction, then use `torch.where` to replace NaN positions with neutral IS weights (`exp(ref_logprob - logprob) = exp(0) = 1`), which is mathematically correct — a token with identical policy and reference log-probabilities contributes no IS correction.

Fixes # (no issue — the bug has a clear reproduction in the existing test suite)

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md)?

## Who can review?

@lewtun @kashif

---

> [!NOTE]
> **Medium Risk**
> Changes the GRPO loss path for vLLM rollouts (IS correction and off-policy mask inputs), but behavior is narrowly scoped to invalid logprob positions with a neutral-weight fallback.
> 
> **Overview**
> Fixes a **crash in `GRPOTrainer`** when vLLM sampling logprobs include `None` (from `extract_logprobs()` after vLLM emits NaN for near-deterministic tokens). **`_generate_and_score_completions`** now maps `None → float("nan")` before building tensors instead of calling `torch.tensor()` on mixed types.
> 
> For **vLLM importance-sampling correction**, NaN sampling logprobs are replaced with recomputed `old_per_token_logps` at those positions so the IS ratio is **1.0** (no correction where sampling logprobs are undefined). The same NaN fill-in runs **before off-policy masking** when `off_policy_mask_threshold` is set, so NaNs do not poison sequence KL / loss means when IS correction is off.
> 
> Adds **`TestVllmNanLogprobHandling`** regression tests for tensor construction, neutral IS ratios, and full NaN cleanup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbafaf8cfb141696dab4ec2f22128d153d46a421. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -238,6 +238,64 @@ class TestGRPORolloutDispatch:
             trainer._generate(["prompt"])
 
 
+class TestVllmNanLogprobHandling:
+    """Regression tests for NaN/None logprob handling in GRPO with vLLM importance sampling.
+
+    vLLM returns NaN logprobs for near-deterministic tokens (sampled-token probability ≈ 1).
+    extract_logprobs() converts those NaN values to None. Prior to the fix these None values
+    caused a crash in _generate_and_score_completions: torch.tensor() raised
+    "RuntimeError: Could not infer dtype of NoneType" (issue #6166).
+    """
+
+    def test_none_logprobs_do_not_crash_tensor_construction(self):
+        """None entries from extract_logprobs() must not raise RuntimeError."""
+        # Simulate sampling_per_token_logps_list with a None at position 1
+        # (token 1 was near-deterministic; vLLM gave it NaN, extract_logprobs set it to None).
+        sampling_per_token_logps_list = [
+            [-0.5, None, -0.3],  # sequence 0: token 1 is near-deterministic
+            [-0.2, -0.7],  # sequence 1: all tokens have valid logprobs
+        ]
+
+        # Replicate the fixed construction from grpo_trainer.py
+        result = [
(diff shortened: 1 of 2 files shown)


shadowmodder:
Good catch — the NaN guard was correctly placed for the IS correction path but missed the off-policy mask branch. Fixed in bbafaf8: added an identical `torch.where(nan_mask, old_per_token_logps, ...)` guard at the top of the `off_policy_mask_threshold` block in `_compute_loss`, before `get_off_policy_mask` is called. This ensures NaN sampling logprobs are replaced with recomputed logprobs regardless of whether `vllm_importance_sampling_correction` is enabled.

PR #6301, opened 2026-07-06
Simplify tokenization [2/N]: Make _tokenize a module-level function

albertvillanova:
Simplify tokenization [2/N]: Make _tokenize a module-level function.

This PR converts the private `_tokenize` static method into a module-level function in the DPO, KTO, reward, and SFT trainers.

Follow-up to (required):
- [x] #6298

### Motivation

`_tokenize` is a stateless static method: it uses neither self nor cls, so it does not need to live on the trainer class. Because it was a method, each _prepare_dataset had to bind it to a local (`tokenize = self._tokenize`) before defining the Dataset.map callback, so that the callback would not close over self. A closure over self makes the map function unhashable, which forces a random fingerprint and silently disables dataset caching. This workaround, plus an explanatory comment, was duplicated across all four trainers:
- #6206

### Solution

Moving `_tokenize` to module level lets the `tokenize_fn` map callback reference it directly, so the local-binding workaround and its comment can be removed. Behavior is unchanged: signatures and docstrings are preserved per file, and dataset caching still produces a stable content-based fingerprint. This also matches the existing precedent in the experimental BCO trainer, which already defines `_tokenize` at module level.

### Changes

- Move the `_tokenize` static method to a module-level function in the DPO, KTO, reward, and SFT trainers, keeping each trainer's own copy rather than sharing one, consistent with the self-contained trainer design.
- Add a short comment on each module-level function explaining why it is not a method, to prevent it from being moved back onto the class.
- Remove the `tokenize = self._tokenize` binding and its comment from each `_prepare_dataset`, and call `_tokenize` directly from the tokenization map callback.

---

> [!NOTE]
> **Low Risk**
> Refactor-only relocation of stateless tokenization helpers with no intended behavior change; risk is limited to preprocessing/caching edge cases if a trainer-specific `_tokenize` diverged.
> 
> **Overview**
> Moves **`_tokenize`** out of the DPO, KTO, Reward, and SFT trainer classes into a **module-level function** in each file (same logic and signatures as before; one copy per trainer, not shared).
> 
> **`_prepare_dataset`** no longer uses `tokenize = self._tokenize` or the related comment. The `Dataset.map` **`tokenize_fn`** calls **`_tokenize`** directly so the callback does not close over **`self`**, keeping the map function hashable and **Hugging Face `datasets` caching** fingerprints stable instead of silently disabling cache.
> 
> Each module-level **`_tokenize`** includes a short comment explaining why it is not a class method, aligned with the experimental BCO trainer pattern.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 300586db2b70acedc7a601bcabfee6e3a75a2ab8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -954,7 +954,6 @@ class DPOTrainer(_BaseTrainer):
     def _tokenize(
         processing_class: PreTrainedTokenizerBase | ProcessorMixin,
         input: str | list,
-        is_vlm: bool,
         **kwargs,
     ) -> dict[str, list]:
         """Tokenize a single example for dataset preprocessing.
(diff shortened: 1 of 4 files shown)


qgallouedec (approved):
lgtm

PR #6307, opened 2026-07-06
Fix precompute_ref_log_probs guard for iterable datasets nested in a dict

albertvillanova:
This PR makes DPOTrainer and KTOTrainer raise a clear error when `precompute_ref_log_probs=True` is used with an evaluation dataset that is a dict of `IterableDataset` or an `IterableDatasetDict`.

### Motivation

The existing guard was meant to reject iterable datasets, but it only matched a single IterableDataset. Dataset preparation rebuilds any dict into a plain dict before the guard runs, so an `IterableDatasetDict` was silently converted and a plain `dict` of `IterableDataset` was never inspected. Both cases slipped through and later failed with a confusing AttributeError about a missing _fingerprint attribute.

### Solution

Move the check into `_precompute_ref_logps`, which every dataset passes through (the train dataset, each eval dataset value, and datasets passed directly to evaluate). This raises a clear, consistent error for every iterable dataset shape and removes the separate up-front guard along with its per-type checks.

### Changes

- Move the iterable dataset rejection from the up-front guard into _precompute_ref_logps in DPOTrainer and KTOTrainer
- Remove the now unused `IterableDatasetDict` import in both trainers
- Add parametrized tests covering iterable train and eval datasets, dicts of iterable datasets, and `IterableDatasetDict`

---

> [!NOTE]
> **Low Risk**
> Validation-only change with clearer errors; map-style datasets and training without precompute are unchanged.
> 
> **Overview**
> **DPOTrainer** and **KTOTrainer** now reject `precompute_ref_log_probs=True` with streaming data in all shapes that reach precompute, not only a top-level train/eval `IterableDataset`.
> 
> The old init-time guard missed iterable eval data inside a plain `dict` or an `IterableDatasetDict` (after dataset prep), which could fail later with a missing `_fingerprint` **AttributeError**. The check moves into **`_precompute_ref_logps`**, which every train/eval split passes through, so each dataset gets the same clear **ValueError**. The redundant upfront guard and **`IterableDatasetDict`** trainer imports are removed.
> 
> Parametrized tests in **`test_dpo_trainer.py`** and **`test_kto_trainer.py`** cover iterable train, eval, `{"eval": iterable}`, and **`IterableDatasetDict`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 532f5426a45dfae56e7e2aa20547d7cb5790288b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_dpo_trainer.py
+++ b/tests/test_dpo_trainer.py
@@ -15,7 +15,7 @@
 import pytest
 import torch
 import transformers
-from datasets import load_dataset
+from datasets import IterableDatasetDict, load_dataset
 from packaging.version import Version
 from transformers import AutoModelForCausalLM, AutoTokenizer, BitsAndBytesConfig
 from transformers.testing_utils import torch_device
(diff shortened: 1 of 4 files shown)


trl/trainer/dpo_trainer.py
```diff
@@ -922,14 +922,6 @@ def __init__(
...
 
         if args.precompute_ref_log_probs:
-            if isinstance(self.train_dataset, IterableDataset) or isinstance(
-                self.eval_dataset, (IterableDataset, IterableDatasetDict)
-            ):
-                raise ValueError(
-                    "`precompute_ref_log_probs=True` is not supported with IterableDataset. Please use a map-style "
-                    "Dataset or set `precompute_ref_log_probs=False`."
-                )
-
```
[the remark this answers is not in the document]
albertvillanova: This is an intentional trade-off.

qgallouedec (approved):

tests/test_dpo_trainer.py
```diff
@@ -1379,6 +1379,35 @@ def test_precompute_ref_log_probs_raises_for_vision(self):
...
+        if iterable_as == "train":
+            train_dataset = iterable_dataset
+        elif iterable_as == "eval":
+            eval_dataset = iterable_dataset
+        elif iterable_as == "eval_dict":
+            eval_dataset = {"eval": iterable_dataset}
+        elif iterable_as == "eval_iterable_dataset_dict":
+            eval_dataset = IterableDatasetDict({"eval": iterable_dataset})
+
+        training_args = DPOConfig(output_dir=self.tmp_dir, report_to="none", precompute_ref_log_probs=True)
```
qgallouedec: the ci fails, I think you should have `max_steps=...` here

EDIT: max_length -> max_steps

albertvillanova: I passed `max_steps=1`

PR #6322, opened 2026-07-08
Support DatasetDict and IterableDatasetDict as eval_dataset in trainers

albertvillanova:
This PR adds explicit support for passing a `DatasetDict` or `IterableDatasetDict` as `eval_dataset` when initializing the stable trainers (SFT, DPO, KTO, Reward, GRPO, RLOO).

This PR adds and tests support for passing `DatasetDict` and `IterableDatasetDict` objects as evaluation datasets (`eval_dataset`) to all TRL trainers, not just plain dictionaries or datasets. This makes it easier to use the raw output of `load_dataset` for evaluation, including both map-style and streaming datasets where supported. The tests are updated to cover these cases, and the trainer docstrings and type annotations are updated accordingly. For trainers that do not support streaming (`IterableDatasetDict`), explicit error handling is added.

### Motivation

`load_dataset` returns a `DatasetDict` (or an `IterableDatasetDict` when streaming) whenever no split is specified, so users commonly hold these types. They already flowed through at runtime because both subclass dict, but the trainers neither declared nor documented them, so the support was implicit and untested.

### Solution

The `eval_dataset` parameter of each stable trainer's `__init__` now accepts `DatasetDict` and `IterableDatasetDict` explicitly, in its type annotation and docstring. Each split keeps being prepared independently at initialization. GRPO and RLOO still do not support streaming datasets, so an `IterableDatasetDict` raises the existing informative error.

### Changes

- Import DatasetDict and IterableDatasetDict from datasets in the SFT, DPO, KTO, Reward, GRPO and RLOO trainers
- Extend the eval_dataset type annotation and docstring in each trainer's __init__
- Add tests verifying a `DatasetDict` and an `IterableDatasetDict` eval dataset are accepted and prepared per split
- Add tests verifying GRPO and RLOO reject an `IterableDatasetDict` eval dataset with a clear error

#### Detailed changes

**Trainer support for `DatasetDict` and `IterableDatasetDict` as eval datasets:**

- All trainer classes (`DPOTrainer`, `KTOTrainer`, `RewardTrainer`, `SFTTrainer`, `GRPOTrainer`, `RLOOTrainer`) now accept `eval_dataset` as a `DatasetDict` or `IterableDatasetDict`, in addition to the previously supported types. The relevant type annotations and docstrings are updated for clarity. 

**Expanded and improved test coverage:**

[...]

Issue #3213, opened 2025-04-02
Critical Issue in GRPO with Iterable datasets


wilrop:
### Reproduction

In the GRPO Trainer class, a `RepeatRandomSampler`class is defined which takes care of sampling from the training dataset, and repeating the prompts for a number of generations, as necessary for GRPO to work correctly. This sampler is utilised in `_get_train_sampler()` which itself overrides the method from the `Trainer` class. However, when using an iterable dataset, this sampler is never actually used, see here https://github.com/huggingface/transformers/blob/3d133cc5576e08a37a7f8a3a39a5ef3316c98c99/src/transformers/trainer.py#L1026. Note also this comment from the `get_train_dataloader()` function:

> Will use no sampler if `train_dataset` does not implement `__len__`, a random sampler (adapted to distributed training if necessary) otherwise.

The problem for iterable datasets occurs when sampling a batch, in which case the trainer receives a batch of data **without repetitions**. Unfortunately, this is not clear to the `GRPOTrainer` which continues computing the advantages as if the batch does contain `self.num_generations` per prompt.

I have manually verified this by putting a breakpoint in `self._generate_and_score_completions()` and checking the inputs.

Since I actually need to use an iterable dataset and would prefer to use GRPO, I'm going to be working on a fix ASAP. I think there are essentially two possible fixes: Override the `get_train_dataloader` and `get_eval_dataloader`methods or use the original methods but pass them `per_device_train_batch_size / num_generations` and add the duplications after. I will open a pull request after it's fixed if one of these solutions sounds acceptable to the maintainers. I'd be happy to hear other suggestions as well.

TLDR: When executing GRPO on an iterable dataset, everything is wrong. Please don't use it for now!

### System Info

- Platform: Linux-6.12.12+bpo-amd64-x86_64-with-glibc2.36
- Python version: 3.11.2
- TRL version: 0.17.0.dev0
- PyTorch version: 2.6.0
- CUDA device(s): NVIDIA A40, NVIDIA A40
- Transformers version: 4.50.3
- Accelerate version: 1.5.2
- Accelerate config: 
  - compute_environment: LOCAL_MACHINE
  - distributed_type: DEEPSPEED
  - mixed_precision: bf16
  - use_cpu: False
  - debug: False
  - num_processes: 1
  - machine_rank: 0
  - num_machines: 1
  - rdzv_backend: static
  - same_network: True
  - main_training_function: main
  - enable_cpu_affinity: False
[...]

PR #6324, opened 2026-07-08
Fix error type for IterableDataset in GRPO and RLOO

albertvillanova:
This PR fixes the error type raised when an `IterableDataset` is passed to `GRPOTrainer` or `RLOOTrainer`.

### Motivation

`NotImplementedError` (a subclass of `RuntimeError`) is meant for abstract methods that subclasses must override, or as a development placeholder. Rejecting an invalid combination of constructor arguments is not that case: the user passed arguments of the right type whose combination is unsupported, which is idiomatically a `ValueError`.

This aligns with the fix already applied for a similar case in DPOTrainer in:
- #6225

Related change: tests are implemented (and should be updated) in:
- #6322

### Solution

Replace `NotImplementedError` with `ValueError` for the iterable dataset check in both trainers.

### Changes

- GRPOTrainer: raise `ValueError` instead of `NotImplementedError` when `train_dataset` or `eval_dataset` is an `IterableDataset`
- RLOOTrainer: same fix as above

---

> [!NOTE]
> **Low Risk**
> Constructor-only exception type change with no training logic or API surface changes beyond how callers catch the error.
> 
> **Overview**
> **GRPOTrainer** and **RLOOTrainer** now raise **`ValueError`** (instead of **`NotImplementedError`**) when `train_dataset` or `eval_dataset` is an **`IterableDataset`**, including iterable splits nested in an eval dict. The error message is unchanged.
> 
> Matching tests in **`test_grpo_trainer.py`** and **`test_rloo_trainer.py`** expect **`ValueError`**, consistent with the same pattern already used for **DPOTrainer**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53ce2f28f12ab21bd6d743c13312bbe130d927cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/trainer/grpo_trainer.py
```diff
@@ -767,7 +767,7 @@ def __init__(
             )
         ):
             # See https://github.com/huggingface/trl/issues/3213
-            raise NotImplementedError(
+            raise ValueError(
```
qgallouedec, line 820: shouldn't we use TypeError?

albertvillanova: TypeError in this codebase is reserved for when an argument doesn't match the type signature at all. 

Here that's not the case: both trainers' signatures explicitly declare `train_dataset: Dataset | IterableDataset | None`. Therefore, `IterableDataset` is a documented, accepted type. The rejection isn't "wrong type"; it's "this valid type isn't supported by this trainer's current implementation" (per the linked issue #3213). That's a value/capability constraint, not a type mismatch; the same reasoning #6225 used (use_liger_kernel=True is not supported with PEFT models → ValueError, even though PEFT-ness is itself a type-like property).

So ValueError is the correct match for the existing convention. No change needed.

qgallouedec: thanks, interesting

qgallouedec (approved):

PR #6335, opened 2026-07-09
Simplify KTO KL completion construction into a single batched map

albertvillanova:
This PR changes how KTO builds the mismatched completions used to estimate the KL divergence term, so that the same code path works efficiently for both map-style and iterable (streaming) datasets.

### Motivation

KTO estimates the KL term from mismatched (prompt, completion) pairs. These were built by creating a separate KL dataset and joining it back with a column-wise `concatenate_datasets`. On streaming datasets this forces the upstream tokenization to run several times per example, which is slow, and it makes iterable support harder to reason about.

### Solution

The `KL_completion_ids` column is now added directly with a single in-place batched map that rotates the completions by +1 within each per-device batch. This produces the exact same pairing as before for map-style datasets, runs the upstream tokenization only once per example, and behaves correctly on iterable datasets.

### Changes

- Add the `KL_completion_ids` column in `_prepare_dataset` with a single in-place batched map instead of building a separate dataset and concatenating it
- Remove the now-unused `_get_kl_dataset` helper method
- Remove the now-unused `get_dataset_column_names` import

---

> [!NOTE]
> **Medium Risk**
> Changes KTO dataset prep for the KL term; pairing logic is preserved but training behavior depends on sequential sampling and batch size, so regressions would affect alignment loss rather than auth or data security.
> 
> **Overview**
> KTO **KL divergence** mismatched pairs are now built in **`_prepare_dataset`** with one **batched `map`** that adds **`KL_completion_ids`** by rotating **`completion_ids`** by +1 within each **`per_device_train_batch_size`** batch—the same pairing as before for map-style data.
> 
> The old path built a separate KL dataset (extra maps, column rename/strip) and **`concatenate_datasets(..., axis=1)`**, which re-ran upstream steps on streaming **`IterableDataset`**s and was harder to reason about. **`_get_kl_dataset`** and the **`get_dataset_column_names`** import in **`kto_trainer.py`** are removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f36db9ae2d5c210905d4706146076aa5fee6ca17. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/trainer/kto_trainer.py
```diff
@@ -1109,12 +1058,24 @@ def tokenize_fn(example, processing_class):
...
-                dataset = concatenate_datasets([dataset, kl_dataset], axis=1)
+                if isinstance(dataset, Dataset):  # `IterableDataset.map` does not support `desc`
+                    map_kwargs["desc"] = f"Extracting KL {dataset_name} dataset"
+
+                def add_kl_completion_ids(examples):
+                    examples["KL_completion_ids"] = [examples["completion_ids"][-1]] + examples["completion_ids"][:-1]
+                    return examples
+
+                dataset = dataset.map(
+                    add_kl_completion_ids, batched=True, batch_size=args.per_device_train_batch_size, **map_kwargs
```
qgallouedec, line 1077: why do we use `args.per_device_train_batch_size` here? I know it's not new, but it feels arbitrary. the default batch_size=1000 would work as well no?

trl/trainer/kto_trainer.py
```diff
@@ -1109,12 +1058,24 @@ def tokenize_fn(example, processing_class):
...
             dataset = dataset.map(tokenize_fn, fn_kwargs={"processing_class": processing_class}, **map_kwargs)
 
-            # Get KL datasets if needed
+            # Add KL completions if needed. The KL term is estimated from mismatched (prompt, completion) pairs, built
+            # by rotating the completions by +1 within each batch of size `per_device_train_batch_size`:
+            # (x_1, y_1), ..., (x_n, y_n) --> (x_1, y_n), (x_2, y_1), ..., (x_n, y_{n-1}). For best results, the
+            # mismatched outputs y' used to estimate the KL term for a batch should be the same set as the matched
+            # outputs y used to estimate the rewards in that batch, just paired with different x. Adding the
+            # `KL_completion_ids` column with a single in-place batched map (rather than building a separate dataset and
+            # concatenating it) keeps this correct and efficient for both map-style and iterable datasets.
```
qgallouedec:
```suggestion
            # outputs y used to estimate the rewards in that batch, just paired with different x.
```

Coding agents tend to justify in a comment why the new approach is better than the old one. But reading it later, detached from the PR, it always makes me think "why would I even consider doing it the other way?". Unless the alternative is a natural thing to reach for and it's genuinely unclear why we didn't, I'd drop that kind of justification.

trl/trainer/kto_trainer.py
qgallouedec: Justification is here I guess, although I'm still not sure to deeply understand:

```
For best results, the
            # mismatched outputs y' used to estimate the KL term for a batch should be the same set as the matched
            # outputs y used to estimate the rewards in that batch, just paired with different x.
```

qgallouedec (approved):
thanks!

trl/trainer/kto_trainer.py
albertvillanova: I think this is intentional: the map batch_size is the window over which completions are rotated to build the mismatched KL pairs. The KL term is computed as a per-device-batch mean of `KL_logps - ref_KL_logps` and used as a reference point inside the loss, so the window has to be the per-device batch. With `batch_size=per_device_train_batch_size`, each batch's mismatched completions `{y'}` are a permutation of that same batch's own completions `{y}` (the "same set" property). With the default 1000, the rotation would span many batches and each batch's boundary example would be paired with a completion from the neighboring batch, so `{y'}` would no longer be the batch's own `{y}`. It would still train, but it changes the KL estimate.

Therefore, I think we should keep it.

trl/trainer/kto_trainer.py
albertvillanova: Agree: I'm dropping the "why this over the old approach".

trl/trainer/kto_trainer.py
albertvillanova: Yes, https://github.com/huggingface/trl/pull/6335#discussion_r3561108612 is the reason for this. 

kl is subtracted from the reward logratios inside the loss, so estimating it from the same completions the batch uses for rewards (just repermuted across prompts) keeps it on the same scale. Happy to reword if it's still unclear.

PR #6353, opened 2026-07-10
Destroy vLLM colocate mode process group after each CI test

albertvillanova:
This PR adds a test fixture that cleans up the `torch.distributed` process group left behind by vLLM colocate mode after each test.

### Motivation

vLLM colocate mode (`distributed_executor_backend="external_launcher"`) initializes `torch.distributed` internally, but the offline `LLM` class never tears it down: neither `LLM` nor `LLMEngine` calls vLLM's own `cleanup_dist_env_and_memory()` on exit. Left undestroyed, PyTorch warns at interpreter shutdown:

> [rank0]:[W...] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources.

This can be reproduced on `main` today by running 
```shell
RUN_SLOW=1 pytest -sv tests/test_grpo_trainer.py::TestGRPOTrainerSlow::test_vlm_processor_vllm_colocate_mode
```

### Solution

Add an autouse pytest fixture that, after each test, destroys the process group via vLLM's own `cleanup_dist_env_and_memory()` helper (rather than a bare `torch.distributed.destroy_process_group()`), so that vLLM's internal parallel-state globals (`_TP`, `_WORLD`, ...) are reset too, letting the next colocate test reinitialize cleanly. This mirrors the pattern vLLM's own test suite uses for the same purpose.

### Changes
- Add `cleanup_vllm_dist_env` autouse fixture to `tests/conftest.py`

---

> [!NOTE]
> **Low Risk**
> Test-only change in `conftest.py`; no production or training runtime behavior is modified.
> 
> **Overview**
> Adds an **autouse** pytest fixture `cleanup_vllm_dist_env` in `tests/conftest.py` that runs after every test.
> 
> When vLLM colocate mode has left `torch.distributed` initialized, the fixture calls vLLM's `cleanup_dist_env_and_memory()` instead of only `destroy_process_group()`, so internal parallel-state globals are reset and later colocate tests can start cleanly. This avoids PyTorch shutdown warnings (e.g. from `test_vlm_processor_vllm_colocate_mode`) and follows the same teardown pattern vLLM uses in its own tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df48ba287e2c3ceafab0b5a28fda27b87ded1c3b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -134,3 +134,22 @@ def cleanup_gpu():
     if torch.cuda.is_available():
         torch.cuda.empty_cache()
         torch.cuda.synchronize()
+
+
+@pytest.fixture(autouse=True)
+def cleanup_vllm_dist_env():
+    """
+    Destroy the torch.distributed process group left behind by vLLM colocate mode after each test.
+
+    vLLM colocate mode initializes `torch.distributed` internally (`distributed_executor_backend="external_launcher"`)
+    but the offline `LLM` class never tears it down: neither `LLM` nor `LLMEngine` calls vLLM's own
+    `cleanup_dist_env_and_memory()` on exit. Left undestroyed, PyTorch warns at interpreter shutdown (e.g.
+    `TestGRPOTrainerSlow::test_vlm_processor_vllm_colocate_mode`). We call vLLM's own cleanup helper rather than a bare
+    `torch.distributed.destroy_process_group()` so that vLLM's internal parallel-state globals (`_TP`, `_WORLD`, ...)
+    are reset too, letting the next colocate test reinitialize cleanly.
+    """
+    yield
+    if torch.distributed.is_initialized():
+        from vllm.distributed.parallel_state import cleanup_dist_env_and_memory
+
+        cleanup_dist_env_and_memory()


qgallouedec (approved):
ok, out of curiosity, why not in `VLLMGeneration.__del__` instead  for example?

albertvillanova:
> why not in `VLLMGeneration.__del__` instead  for example?

Good question, @qgallouedec! Indeed, I considered fixing this inside VLLMGeneration (e.g. __del__) instead of in the test suite, but ruled it out:
- `destroy_process_group()` is a collective call: every rank must call it together. Garbage collection is per-process, refcount-driven, with no cross-rank coordination. This is exactly why PyTorch's own docs recommend an explicit, synchronized teardown call rather than relying on a finalizer.
- There's no existing point in the trainer lifecycle that's both (1) a safe cross-rank sync barrier and (2) guaranteed to mean "the user is done with vLLM"
- In real (single-shot) training scripts this is cosmetic: the process exits right after anyway, and vLLM's own init_distributed_environment already skips re-init if a group exists, so nothing accumulates across a single run
  - The place this actually recurs predictably is a long-lived interpreter running many colocate instantiations back-to-back: the test suite

Additionally, I checked `vllm` codebase. `vllm` colocate mode makes vLLM call `torch.distributed.init_process_group()` internally, but the offline `LLM`/`LLMEngine` classes never tear it down: `LLMEngine.__del__` only ever destroys a data-parallel group, and skips even that for `external_launcher`.

Given all this, this PR mirrors the pattern vLLM's own test suite already uses for the identical problem: an autouse pytest fixture that destroys the process group after each test via vLLM's own cleanup_dist_env_and_memory() (not a bare torch.distributed.destroy_process_group(), so vLLM's internal parallel-state globals like _TP/_WORLD get reset too, letting the next colocate test reinitialize cleanly).

If a real user workflow later needs safe teardown between sequential trainer runs in one process (e.g. a hyperparameter sweep), the right fix would be an explicit opt-in method the caller invokes at a point they control (not automatic `__del__`-based cleanup) but that's out of scope here since no such need has been reported yet.

PR #6357, opened 2026-07-11
Fix EOS being masked when pad_token_id equals eos_token_id in GKD and GOLD

roycho96:
# What does this PR do?

When `pad_token_id == eos_token_id`, GKD and GOLD build their masks by matching the pad id. That id is also the terminating EOS, so the EOS is treated as padding:

- On-policy generation: the terminating EOS label is set to `-100` on every completion, so the student never learns to stop, and real prompt tokens equal to the pad id (chat markers like `<|im_end|>`) lose attention.
- GOLD ULD teacher inputs: the appended teacher EOS is stripped the same way.

`pad == eos` is native for SmolLM2-Instruct, and is the standard setup (`tokenizer.pad_token = tokenizer.eos_token`) for any model without a dedicated pad token, which TRL's own GKD tests use.

## Repro

```python
import inspect
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer, GenerationConfig
from trl.experimental.gkd.gkd_trainer import GKDTrainer

model_id = "HuggingFaceTB/SmolLM2-360M-Instruct"  # ships pad_token == eos_token == <|im_end|>
tok = AutoTokenizer.from_pretrained(model_id)
tok.padding_side = "left"
model = AutoModelForCausalLM.from_pretrained(model_id, dtype="float32")
eos = tok.eos_token_id
assert tok.pad_token_id == eos  # native, nothing set by hand

convs = [[{"role": "user", "content": "What is 2 + 2?"}], [{"role": "user", "content": "Name one color."}]]
texts = [tok.apply_chat_template(c, tokenize=False, add_generation_prompt=True) for c in convs]
enc = tok(texts, return_tensors="pt", padding=True)
inputs = {"prompts": enc["input_ids"], "prompt_attention_mask": enc["attention_mask"]}
gen_config = GenerationConfig(max_new_tokens=48, do_sample=False, pad_token_id=eos, eos_token_id=eos)

kwargs = {}
if "pad_token_id" in inspect.signature(GKDTrainer.generate_on_policy_outputs).parameters:
    kwargs["pad_token_id"] = tok.pad_token_id  # present on main, removed by this PR

with torch.no_grad():
    ids, attn, labels = GKDTrainer.generate_on_policy_outputs(model, inputs, gen_config, **kwargs)

pw = inputs["prompts"].shape[1]
for r in range(ids.shape[0]):
    prow, pmask = inputs["prompts"][r], inputs["prompt_attention_mask"][r]
    marker_pos = [i for i in range(pw) if prow[i].item() == eos and pmask[i].item() == 1]
    comp, comp_lab = ids[r, pw:], labels[r, pw:]
    eos_pos = (comp == eos).nonzero().flatten().tolist()
    print(f"row {r}: prompt markers (==eos) attention {[attn[r, i].item() for i in marker_pos]} (want all 1)")
    print(f"row {r}: completion eos positions {eos_pos} -> labels {[comp_lab[p].item() for p in eos_pos]} (want {eos})")
```

On `main`:

```
row 0: prompt markers (==eos) attention [0, 0] (want all 1)
row 0: completion eos positions [8] -> labels [-100] (want 2)
row 1: prompt markers (==eos) attention [0, 0] (want all 1)
row 1: completion eos positions [5, 6, 7, 8] -> labels [-100, -100, -100, -100] (want 2)
```

The terminating EOS is labeled `-100` (the student never learns to stop) and the `<|im_end|>` markers in the prompt lose attention.

With this PR:

```
row 0: prompt markers (==eos) attention [1, 1] (want all 1)
row 0: completion eos positions [8] -> labels [2] (want 2)
row 1: prompt markers (==eos) attention [1, 1] (want all 1)
row 1: completion eos positions [5, 6, 7, 8] -> labels [2, -100, -100, -100] (want 2)
```

The first (terminating) EOS keeps its label, only the padding after it is masked, and the prompt markers keep attention.

## Fix

Mask completions by the first EOS position, the idiom GRPO and RLOO already use, instead of by pad-id identity: keep the first stop token attended and labeled, mask only what comes after, and take the prompt half of the mask from `prompt_attention_mask`. `torch.isin` handles `eos_token_id` being a list of stop tokens (e.g. Llama 3). GKD and GOLD use the same block. On the GOLD ULD path the pad-id strips are removed since padding is already masked by construction (labels padded with `-100`, attention with `0`). Two now-dead branches (`_build_sequence_batch`'s no-mask branch, an unused `pad_token_id` param) are dropped.

New tests cover pad == eos, a missing eos id, and multiple stop tokens.

## Before submitting

## AI writing disclosure

---

> [!NOTE]
> **Medium Risk**
> Changes how distillation trainers label and attend generated sequences, which directly affects the training loss, but the scope is limited to experimental GKD/GOLD with strong regression tests.
> 
> **Overview**
> **GKD and GOLD** no longer build attention/labels by matching `pad_token_id`, which wrongly treated terminating EOS and in-prompt special tokens (e.g. chat markers) as padding when `pad_token_id == eos_token_id`.
> 
> `generate_on_policy_outputs` drops the `pad_token_id` argument and masks the completion through the **first EOS** (`torch.isin` when `eos_token_id` is a list; full completion when it is `None`). The prompt side of the mask comes from **`prompt_attention_mask`**, and labels are `-100` only where attention is zero.
> 
> **GOLD** applies the same on-policy logic, removes redundant pad-id stripping on the ULD teacher-input path, `_decode_completion_texts_from_labels`, and `_build_sequence_batch`, and always decodes prompts via the prompt mask.
> 
> New tests cover pad==eos, missing `eos_token_id`, and multiple stop tokens.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0bc85e9e1a04878a9b65b52761f5ff5815c358c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/gkd/gkd_trainer.py
```diff
@@ -457,19 +457,34 @@ def generate_on_policy_outputs(model, inputs, generation_config, pad_token_id=No
...
+            eos_idx = torch.full((is_eos.size(0),), is_eos.size(1), dtype=torch.long, device=device)
+            eos_idx[is_eos.any(dim=1)] = is_eos.int().argmax(dim=1)[is_eos.any(dim=1)]
+            sequence_indices = torch.arange(is_eos.size(1), device=device).expand(is_eos.size(0), -1)
+            completion_mask = (sequence_indices <= eos_idx.unsqueeze(1)).int()
+
+        # Pad ids can appear inside real prompt text, so use the prompt mask instead of matching ids
         new_attention_mask = torch.ones_like(generated_tokens)
-        new_labels = generated_tokens.clone()
+        new_attention_mask[:, prompt_length:] = completion_mask
+        if prompt_mask is not None:
```
cmpatino: Should we require the prompt mask now?

This condition leaves every prompt position attended when no mask is provided. In practice, that path is not valid because both built-in collators always emit `prompt_attention_mask`, and generating from padded inputs without it is already incorrect.

Maybe we could use `inputs["prompt_attention_mask"]` instead, so malformed custom-collator output fails loudly rather than silently attending padding?

The same applies to the duplicated block in GOLD. Once the mask is required there, the `elif pad_token_id is not None` decoding fallback becomes unreachable, so `pad_token_id` can also be removed from the signature and call sites, as this PR already does for GKD.

cmpatino (commented):
Thank you for raising this issue and fixing it :) 

I left a comment about how we should handle the prompt mask given the changes in the PR.

trl/experimental/gkd/gkd_trainer.py
roycho96: Required prompt_attention_mask in both GKD and GOLD, dropped the is not None guards, and removed the now-dead pad_token_id fallback plus the GOLD signature/call sites. Also fixed a GKD test that passed the mask under the wrong key, which had been silently taking the removed path.

cmpatino (approved):
LGTM! Thank you for the contribution.

PR #6378, opened 2026-07-13
Simplify AsyncGRPOTrainer: drop redundant metric guards, merge train-begin callbacks

qgallouedec:
Remove the valid_mask.any() guards in compute_loss (empty masked sums already return a 0 scalar on the right device) and merge _InitialWeightSyncCallback + `_StartRolloutWorkerCallback` into a single `_TrainBeginCallback` (ordering was only enforced by registration order before). No behavior change.

---

> [!NOTE]
> **Medium Risk**
> Clip min/max logging changes meaning when multiple completions share a rank row; training loss is unchanged but dashboards comparing to old runs may differ on those metrics.
> 
> **Overview**
> **Train begin:** `_InitialWeightSyncCallback` and `_StartRolloutWorkerCallback` are merged into one idempotent `_TrainBeginCallback` that runs NCCL init, cold weight sync, then rollout worker start in a fixed order (no reliance on separate callback registration order).
> 
> **`compute_loss` metrics:** Redundant `valid_mask.any()` branches are removed; masked sums on empty selections already yield a zero scalar on the correct device.
> 
> **Clip extrema:** `clip_ratio/low_min` and `clip_ratio/high_max` no longer take the min/max of each rank’s *token-level* low/high clip fraction. They now segment the packed padding-free row by completion (`position_ids`), compute each completion’s clip fraction, then `gather` + `nanmin`/`nanmax`—the same pattern as `GRPOTrainer`’s `masked_seq_mean`, which matters when a rank packs multiple completions in one row.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6dbdf28fd2a4b15302ec7e7ecc40d6ceb254f816. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -665,9 +654,8 @@ def __init__(
             self.rollout_worker = None
             self.weight_transfer = None
 
-        # Add callbacks. Registration order matters: weight sync first, then worker start.
-        self.add_callback(_InitialWeightSyncCallback(self))
-        self.add_callback(_StartRolloutWorkerCallback(self))
```
AmineDiro, line 1000: I am pretty sure this StartRolloutWorkerCallback was separated for a reason, need to dig into my notes.

qgallouedec: did you find the reason?

qgallouedec:
@AmineDiro gentle bump on the callback split.

I went back to #5749 and I think the answer is in its own description: "_StartRolloutWorkerCallback: rollout_worker.start() (registration order enforces 'sync before start')". So the separation looks organizational rather than functional -> the ordering constraint is the entire reason there were two callbacks, and `_TrainBeginCallback` now enforces it in code instead of relying on `add_callback` order.
The `_fired` guards are kept, so the `remove_callback(type(self))` footgun you flagged in #5749 doesn't come back either.

If you do turn up a different reason (some fork/spawn timing thing?), I'll drop that half of the PR; otherwise I'll merge in a couple of days.

AmineDiro:
I checked, I think the order was the whole point, and enforcing it in the method body is better. LGTM, let's merge it 👍🏼

PR #6379, opened 2026-07-13
Fix async GRPO clip-ratio extrema to global per-completion

qgallouedec:
Requires #6378

Same fix as the sync GRPO/RLOO metric-aggregation PR #6380, applied to the experimental async trainer.

`clip_ratio/low_min` and `clip_ratio/high_max` were the min/max of each rank's whole-row mean clip fraction — data-parallel-layout-dependent, not a true per-completion extremum. Async packs several completions into one row per rank, so the fix splits the row by `position_ids` resets, computes each completion's clip fraction over its own completion tokens, and takes the global min/max as the min/max of the per-rank extrema (min-of-min / max-of-max) — equivalent to gathering every completion but robust to ranks holding different numbers of completions.

---

> [!NOTE]
> **Low Risk**
> Metrics-only change in `compute_loss` logging; PPO clipping and gradients are untouched.
> 
> **Overview**
> Fixes **`clip_ratio/low_min`** and **`clip_ratio/high_max`** in the experimental async GRPO trainer so they reflect true per-completion clip saturation instead of each data-parallel rank’s whole packed-row average.
> 
> Async training concatenates multiple completions into one row per rank. The previous logic gathered each rank’s single row-mean low/high clip fraction and took the global min/max, which depended on how completions were packed across ranks. The change uses **`position_ids`** resets to segment the packed row, sums low/high clip flags and valid completion tokens per segment, and forms per-completion clip fractions. Each rank contributes **`nanmin` / `nanmax`** over its local completions; those values are gathered and reduced again for the logged extrema (aligned with the sync GRPO/RLOO fix). Per-rank **`clip_ratio/*_mean`** metrics are unchanged.
> 
> **`n_samples`** for rollout metric aggregation now reuses the same **`n_completions`** count instead of recomputing it from **`position_ids`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3c387e727d270b229448737a24101e577c516c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -997,12 +993,23 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
-            gathered_low_clip = self.accelerator.gather(local_low_clip_mean)
-            gathered_high_clip = self.accelerator.gather(local_high_clip_mean)
-            self._metrics["train"]["clip_ratio/low_min"].append(nanmin(gathered_low_clip).item())
-            self._metrics["train"]["clip_ratio/high_max"].append(nanmax(gathered_high_clip).item())
+            seq_ids = (position_ids[0] == 0).cumsum(0)[1:] - 1  # (T-1,) completion index per (shifted) token
+            n_completions = (position_ids == 0).sum()  # number of packed completions in this rank's row
+            num_seq = int(n_completions)
+            comp_mask = completion_mask[0].float()  # (T-1,) valid completion-token mask
+
+            def seg_sum(vals):  # per-completion segment sum over the packed row
```
AmineDiro, line 1001: nit: close capturing non local state makes me uncomfortable 😆  But its all right

AmineDiro (approved):

PR #6386, opened 2026-07-14
Batch vLLM weight sync into size-capped buckets

kashif:
Weight sync to vLLM currently does one HTTP request, one collective_rpc and one NCCL broadcast per parameter. For models with hundreds of tensors and per-step syncing this fixed overhead adds up.

This PR batches the sync path:

- new batched `/update_named_params/` endpoint and worker method that receives a list of tensors per request and loads them in a single `load_weights` call (the vLLM model API already takes an iterable of pairs)
- `VLLMGeneration` accumulates params into 256 MB buckets before pushing, so peak transient memory (client-side gathers, server receive buffers) stays bounded regardless of model size
- FSDP2 sync now materializes `full_tensor()` lazily inside the bucket loop instead of building the whole unsharded model up front
- unit tests for the bucketing and the batched update path

Same bucketing pattern as other RL stacks (slime and miles use a 512 MB buffer, NeMo-RL a configurable refit buffer). All trainers using `VLLMGeneration` pick this up through the unchanged `sync_weights()` interface.

Extracted from #6126, which no longer needs to carry it.

---

> [!NOTE]
> **Medium Risk**
> Changes the distributed weight-sync protocol (HTTP + NCCL/XCCL ordering); mistakes could desync trainer and vLLM weights, though behavior is covered by new unit tests and existing integration tests still call `update_model_params`.
> 
> **Overview**
> **Weight sync to vLLM is batched** instead of one HTTP request, `collective_rpc`, and broadcast per parameter.
> 
> The client adds **`update_named_params`** (single POST, ordered broadcasts, one barrier) and **`update_model_params`** now sends all tensors through that path. The serve worker and **`/update_named_params/`** endpoint load each batch in one **`load_weights`** call.
> 
> **`VLLMGeneration`** groups parameters into **256 MB buckets** via **`_bucket_push`**, uses batched **`_push_params_to_vllm`**, and routes FSDP/PEFT/non-PEFT **`sync_weights`** through bucketing. **FSDP2** materializes **`full_tensor()`** lazily per bucket rather than unsharding the full model up front.
> 
> Unit tests cover the client batch API and bucketing behavior; **`sync_weights()`**’s public interface is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c165c11f4042803a787baedc9086f53ac4a17fac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_vllm_client_server.py
+++ b/tests/test_vllm_client_server.py
@@ -14,15 +14,18 @@
 
 import os
 import subprocess
+from contextlib import contextmanager
 from types import SimpleNamespace
+from unittest.mock import MagicMock
 
 import pytest
+import torch
 from packaging.version import Version
 from transformers import AutoModelForCausalLM, AutoProcessor, AutoTokenizer
 from transformers.testing_utils import torch_device
 
 from trl.generation.vllm_client import VLLMClient
-from trl.generation.vllm_generation import extract_logprobs
+from trl.generation.vllm_generation import VLLMGeneration, extract_logprobs
 from trl.import_utils import is_vllm_available
 from trl.scripts.vllm_serve import chunk_list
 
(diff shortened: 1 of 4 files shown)


kashif:
Closing in favor of vLLM's native weight transfer API (NCCLWeightTransferEngine with packed=True, already used by AsyncGRPOTrainer), which does packed double-buffered bucketing internally. Migrating VLLMGeneration server mode to that API is the better path than optimizing the custom sync protocol.

PR #6388, opened 2026-07-14
[GOLD] Recommend LFM2.5 and Gemma 4 students

kashif:
Updates the GOLD example to recommend LFM2.5-1.2B and Gemma 4 E4B students with a Qwen3-4B teacher. It uses the same Formal Logic dataset as the NeMo RL x-token recipe and converts its raw text into prompt/completion pairs.

This also enables positional ULD for SentencePiece tokenizers and fixes an MPS logging issue found during local testing.

Smoke tested one training step on an M5 Max:

- LFM2.5-1.2B → loss 0.3359
- Gemma 4 E4B with LoRA → loss 0.4245

The focused GOLD tests pass (`4 passed`).

---

> [!NOTE]
> **Medium Risk**
> Changes GOLD dataset tokenization, teacher encoding, and ULD alignment paths; incorrect `use_extended_uld` settings could misalign student/teacher logits for SentencePiece students.
> 
> **Overview**
> **GOLD** now supports **positional ULD** (`use_extended_uld=False`) for SentencePiece and other tokenizers without a fast `backend_tokenizer`, enabling cross-tokenizer distillation for models like Gemma 4.
> 
> Teacher/student dataset prep and `build_teacher_inputs_from_texts` branch on `use_extended_uld`: extended mode still uses byte-offset alignment; positional mode tokenizes with `return_offsets_mapping` and zero byte offsets. EOS is appended during tokenization (replacing a separate dataset map), and prompt/completion boundaries use an end-offset rule so seam tokens are not dropped. VLM teacher input building follows the same flag. Config docs note setting `use_extended_uld=False` for non-ByteLevel tokenizers.
> 
> The **gold_chatbot_arena** example is retargeted to LFM2.5 / Gemma 4 students, Qwen3-4B teacher, and Nemotron Formal Logic, with `split_text_for_gold` to turn raw `text` fields into 75/25 prompt/completion pairs.
> 
> Distributed on/off-policy loss aggregation in **GOLD** and **IW-OPD** uses `float32` tensors instead of `float64` (MPS compatibility). Tests cover positional ULD, Gemma 4, extended-ULD seam behavior, and slow-tokenizer paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 972ca5c81ada477f96e39e806106280e26efe6c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/gold/gold_trainer.py
+++ b/trl/experimental/gold/gold_trainer.py
@@ -182,13 +182,15 @@ def build_teacher_inputs_from_texts(
     tokenizer: PreTrainedTokenizerBase,
     prompt_texts: list[str],
     completion_texts: list[str],
+    use_extended_uld: bool = True,
 ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
     """Tokenize teacher prompts/completions and produce tensors ready for GOLD loss.
 
     Returns ``(input_ids, labels, attention_mask, byte_offsets)``. ``byte_offsets`` is a ``[batch, seq, 2]`` tensor of
     UTF-8 byte ``(start, end)`` for each token: prompt and padding positions are filled with ``(0, 0)``; completion
     tokens carry offsets relative to the corresponding ``completion_text``; the appended EOS gets ``(content_len,
-    content_len)``. Byte offsets are derived from the fast tokenizer's char offsets via ``encode_with_byte_offsets``.
+    content_len)``. Byte offsets are derived from the fast tokenizer's char offsets via ``encode_with_byte_offsets``
+    when extended ULD is enabled. Positional ULD returns zero offsets because it does not consume them.
     """
 
     pad_token_id = tokenizer.pad_token_id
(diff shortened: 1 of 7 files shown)


trl/experimental/gold/gold_trainer.py
cmpatino, line 2133: Should we add a `completion_text = text` here?

Otherwise, I think the trainer crashes when the dataset only contains `text`, `use_extended_uld=True`, and EOS needs to be appended.

cmpatino: Should we also add the `use_extended_uld` gate we are using in `build_teacher_inputs_from_texts`? 

The docstring says we mirror that method, but we're missing here the condition that depends on whether we're using extended ULD.

cmpatino (approved):
LGTM!

qgallouedec:
@kashif what's the state here? it shows approved but i don't think it's landable as is.

the approval is from jul 15 and you pushed again aug 19, so it covers a different diff. @cmpatino also left two questions on gold_trainer.py that look unanswered, one of them is a crash when the dataset only has text.

It also patches examples/scripts/gold.py, which doesn't exist any more after the examples reorg, so this needs a rebase onto the per-example folder layout.

The title says recommend students but this adds a positional uld path to the trainer. I'd split: the trainer change on its own with the alignment question settled, then the example recommendation on top.

qgallouedec:
Both open threads are addressed in code, thanks!

@cmpatino's gate question: `build_teacher_inputs_from_texts` and the mirroring path now both do `(content_len, content_len) if use_extended_uld else (0, 0)` when appending EOS, so they agree. And the codex P2 about the fast backend is what c5da0513 fixes.

Clean and mergeable now. Approving on the current diff since Carlos' is from july and predates two pushes 👍👍

qgallouedec (approved):

PR #6389, opened 2026-07-15
Auto-load `processing_class` in PRMTrainer when omitted

DaoyuanLi2816:
## What does this PR do?

Auto-loads `processing_class` from the model when it is omitted, matching SFT, DPO, and Reward trainers. This applies with either the default or a custom data collator and for both raw and pre-tokenized datasets.

`PRMConfig` now exposes `trust_remote_code` for tokenizer loading.

## Tests

- processing-class autoload coverage for all three dataset/collator paths (3 passed)
- Ruff check and format check on the changed files

## AI writing disclosure

### DIFF
--- a/tests/experimental/test_prm_trainer.py
+++ b/tests/experimental/test_prm_trainer.py
@@ -347,6 +347,30 @@ class TestPRMTrainer(TrlTestCase):
             if param.sum() != 0:  # ignore 0 biases
                 assert not torch.equal(param, new_param)
 
+    def test_processing_class_autoloaded(self):
+        dataset = load_dataset("trl-internal-testing/zen", "standard_stepwise_supervision", split="train")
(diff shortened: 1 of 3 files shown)


qgallouedec:
thanks. I think to better align with the other trainer, a better path would be to automatically load the processor in all cases. see how it's done in SFT, DPO etc.

DaoyuanLi2816:
Updated this to auto-load the tokenizer whenever `processing_class` is omitted, including custom collators and pre-tokenized datasets. Added coverage for all three paths.

DaoyuanLi2816:
@qgallouedec this now loads the processor automatically in all cases, as you suggested, rather than only for the collator path — it covers custom collators and pre-tokenized datasets too, with coverage for all three. Ready for another look whenever you have time.

qgallouedec:
Closing in favour of #7133, which removes `trl.experimental.prm`. The autoload is right and matches what SFT, DPO and Reward do, there is just no trainer left to carry it. thks for the work on it.

PR #6390, opened 2026-07-15
Fix BEMACallback bias_power docstring, which described the opposite of the actual behavior

DaoyuanLi2816:
## What this fixes

`BEMACallback`'s docstring for `bias_power` says:

> Power for the BEMA scaling factor. Denoted \\( \eta \\) in the paper. **To disable BEMA, set this to `0.0`.**

This is backwards. The actual computation (`trl/trainer/callbacks.py`):

```python
def _bema_alpha(self, step: int) -> float:
    """Compute the BEMA scaling factor αₜ = (ρ + γ·t)⁻ᵉᵗᵃ."""
    return (self.lag + self.multiplier * step) ** (-self.bias_power)

def _update_bema_weights(self, step: int):
    ...
    run_param.copy_(ema + alpha * (thetat - theta0))  # BEMA update
```

At `bias_power=0.0`, `alpha_t = (lag + multiplier*step) ** 0 == 1` for **every** step, forever. So `run_param = ema + 1 * (theta_t - theta_0)` — the full, undecayed bias-correction term is applied permanently. That's the *maximum* possible bias correction, not "disabled".

A **large** `bias_power` (e.g. `8.0`) instead makes `alpha_t` decay toward `0` as training progresses, so `run_param -> ema` — a plain EMA with no bias-correction term left. That's what actually approximates "disabled".

(For comparison, the adjacent `ema_power` param's "`0.0` disables EMA" claim *is* correct: at `ema_power=0`, `beta_t == 1`, so the EMA update `ema = (1-beta)*ema + beta*theta_t` collapses to `ema = theta_t` — i.e. the average carries no history and just tracks the live weights, which is a reasonable reading of "disabled". `bias_power` doesn't have that property, since `alpha_t` scales a term that's *added on top of* `ema`, not blended with it.)

## Verification

I ran the real `BEMACallback` end-to-end (not a re-derivation of the algebra) against a tiny real `GPT2LMHeadModel`: `on_train_begin`, then 200 simulated training steps (`on_step_end` each step, with small random perturbations applied to the live model weights so `theta_t` actually diverges from `theta_0`), for several values of `bias_power`.

At the end, I compared two tensors the callback itself produces:
- `callback.running_model` — the actual output users get (`ema + alpha*(theta_t - theta_0)`)
- `callback.ema_params` — the callback's own internal pure-EMA buffer, updated only via `ema.mul_(1-beta).add_(thetat, alpha=beta)`, which has **zero dependence on `bias_power`/`alpha`** — i.e. it *is* the "no bias-correction" baseline, tracked internally by the class under test itself.

| `bias_power` | mean \|running_model − ema_params\| | relative to typical weight magnitude |
|---|---|---|
| 0.0 (docstring claims "disabled") | 0.11154473 | **109.84%** |
| 0.2 (current default) | 0.03828293 | 37.70% |
| 1.0 | 0.00053117 | 0.52% |
| 2.0 | 0.00000253 | 0.0025% |
| 4.0 | 0.00000000 | ~0% |
| 8.0 | 0.00000000 | ~0% |
| 16.0 | 0.00000000 | ~0% |

`bias_power=0.0` deviates from pure EMA by ~110% of typical weight magnitude (maximum bias-correction, as the formula predicts), while `bias_power>=4.0` collapses to the pure-EMA baseline (deviation exactly 0 in float32) — the opposite of what the current docstring claims.

Direct `alpha_t` values confirm the mechanism:

```
step=     0  alpha(bias_power=0.0)=1.000000   alpha(bias_power=16.0)=1.000e-16
step=     1  alpha(bias_power=0.0)=1.000000   alpha(bias_power=16.0)=2.176e-17
step=    10  alpha(bias_power=0.0)=1.000000   alpha(bias_power=16.0)=1.526e-21
step=   100  alpha(bias_power=0.0)=1.000000   alpha(bias_power=16.0)=2.176e-33
step=  1000  alpha(bias_power=0.0)=1.000000   alpha(bias_power=16.0)=8.528e-49
step= 10000  alpha(bias_power=0.0)=1.000000   alpha(bias_power=16.0)=9.841e-65
```

## Fix

Docs-only, no behavior/math change. Corrected the `bias_power` docstring in both `trl/trainer/callbacks.py` and its duplicate in `trl/experimental/bema_for_ref_model/callback.py` (kept consistent, per this repo's convention for duplicated docstrings) to describe the actual behavior: a large `bias_power` approximates disabling the bias-correction term; `0.0` instead pins it at maximum.

Note: #3917 previously touched this same `Args` block (fixed the `lag` entry, added `# docstyle-ignore`), but the `bias_power` line itself was untouched by it and has been wrong since the callback was introduced.

## Before submitting

---

> [!NOTE]
> **Low Risk**
> Documentation-only change to callback docstrings; training behavior is unchanged.
> 
> **Overview**
> Corrects the **`bias_power`** argument docs on **`BEMACallback`** in `trl/trainer/callbacks.py` and the mirrored class in `trl/experimental/bema_for_ref_model/callback.py`.
> 
> The old text said **`0.0` disables BEMA**; the implementation uses **`α_t = (ρ + γ·t)^{-η}`**, so **`bias_power=0`** keeps **`α_t = 1`** (strongest bias correction) and **large values** (e.g. **`8.0`**) drive **`α_t` toward 0** (output ≈ pure EMA). The docstring now states that behavior. **No runtime or math changes.**
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1d938c7464adcb0396d8d9fdc5dfbbf3d06368a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova (commented):
I think this CI error is unrelated: https://github.com/huggingface/trl/actions/runs/29396050577/job/88363627839?pr=6390
- Just need to merge main into this branch to incorporate the fix by:
  - #6413

albertvillanova (approved):
Thanks for the fix: the fix is correct.

Before merging, I would recommend first merging main into this PR and wait for the CI to be green, to clear a stale, un-related CI failure.

PR #6400, opened 2026-07-15
Refactor evaluate [1/N]: Fully fix DPOTrainer/KTOTrainer crash on dict eval_dataset with precompute_ref_log_probs

albertvillanova:
This PR is a follow-up to #6370, which applied a narrow fix (an idempotent guard in `_precompute_ref_logps`) for a crash when `evaluate(eval_dataset=<dict-like>)` was combined with `precompute_ref_log_probs=True`. This PR replaces that guard with a structural fix that addresses the root cause.

Follow-up to:
- #6370

### Motivation

`transformers.Trainer.evaluate()` handles a dict `eval_dataset` by recursing into `self.evaluate(...)` once per split. Since `self.evaluate` resolves polymorphically, this re-enters the trainer's own `evaluate()` override a second time on an already-processed dataset. The narrow fix in #6370 papered over this for `precompute_ref_log_probs`, but the underlying double-dispatch was still there, latent, waiting for the next stateful preprocessing step to trip over it.

### Solution

Move dataset preparation and ref-logps precomputation out of the `evaluate()` override and into `get_eval_dataloader()`, which is the extension point transformers documents for this purpose:
> Subclass and override this method if you want to inject some custom behavior

Unlike `evaluate()`, `get_eval_dataloader()` is called exactly once per split, so it's structurally immune to the double-dispatch: `Trainer.evaluate()` already resolves a dict eval_dataset into individual splits before `get_eval_dataloader()` is ever called. This also removes the now-unnecessary idempotent guard from #6370, since the scenario it protected against can no longer occur.

### Changes

- trl/trainer/dpo_trainer.py: replace the `evaluate()` override with a `get_eval_dataloader()` override; remove the idempotent guard in `_precompute_ref_logps`
- trl/trainer/kto_trainer.py: same change, mirroring DPO per the repo's duplicated-code convention

---

> [!NOTE]
> **Medium Risk**
> Changes the eval dataloader path for DPO/KTO when passing ad-hoc eval data or using precomputed reference log-probs; behavior should improve but eval preprocessing timing shifts relative to the old override.
> 
> **Overview**
> **DPOTrainer** and **KTOTrainer** no longer override `evaluate()` for eval-time dataset handling. That logic now lives in **`get_eval_dataloader()`**, which tokenizes/prepares a raw eval set and optionally runs **`_precompute_ref_logps`** when `precompute_ref_log_probs=True`, then delegates to the base `Trainer`.
> 
> This fixes crashes and redundant work when **`eval_dataset` is a dict of splits**: Hugging Face `Trainer.evaluate()` already iterates splits and calls `get_eval_dataloader()` once per split, whereas the old `evaluate()` override re-entered on already-processed data (double preprocessing). Dict branching in the override is removed because `eval_dataset` is never a dict at this hook.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8eb6ac52b47ea54c47b29cd8246949e98f1868c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/trainer/dpo_trainer.py
+++ b/trl/trainer/dpo_trainer.py
@@ -1622,43 +1622,15 @@ class DPOTrainer(_BaseTrainer):
 
         return (loss, outputs) if return_outputs else loss
 
-    def evaluate(
-        self,
-        eval_dataset: Dataset
-        | IterableDataset
-        | DatasetDict
-        | IterableDatasetDict
-        | dict[str, Dataset | IterableDataset]
-        | None = None,
-        ignore_keys: list[str] | None = None,
-        metric_key_prefix: str = "eval",
-    ) -> dict[str, float]:
-        # When a dataset is passed directly to `evaluate` (e.g. a held-out test set), preprocess it the same way
-        # `__init__` does, so that `evaluate` accepts the same dataset types as the trainer. `_prepare_dataset` is
-        # idempotent: it skips datasets that are already tokenized. A `str` selects a dataset that was already prepared
-        # at init time, so it's left untouched.
+    def get_eval_dataloader(self, eval_dataset: str | Dataset | IterableDataset | None = None) -> DataLoader:
+        # `Trainer.evaluate` already resolves a dict `eval_dataset` into individual splits before calling this, once
+        # per split, so `eval_dataset` here is never a dict. `_prepare_dataset` is idempotent.
         if not self._is_vision_dataset and eval_dataset is not None and not isinstance(eval_dataset, str):
-            if isinstance(eval_dataset, dict):
-                eval_dataset = {
-                    key: self._prepare_dataset(dataset, self.processing_class, self.args, key)
-                    for key, dataset in eval_dataset.items()
-                }
-            else:
-                eval_dataset = self._prepare_dataset(eval_dataset, self.processing_class, self.args, "eval")
-            # With `precompute_ref_log_probs`, `_compute_loss` reads the reference log-probs from the batch, so they
-            # must be precomputed here as well, mirroring `__init__`.
+            eval_dataset = self._prepare_dataset(eval_dataset, self.processing_class, self.args, "eval")
             if self.precompute_ref_logps:
                 batch_size = self.args.precompute_ref_batch_size or self.args.per_device_eval_batch_size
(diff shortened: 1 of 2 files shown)


qgallouedec:
I don't think the premise holds: `get_eval_dataloader()` isn't only reached once per split, it's reached on **every** eval with the already-prepared dataset.

The old `evaluate()` override got `None` here and skipped via `eval_dataset is not None`. So this doesn't relocate the logic: it moves it from a rare path (user passes a dataset) onto the hot path, re-running `_prepare_dataset` at every eval.

`_precompute_ref_logps` is fine, the idempotent guard absorbs the extra call. `_prepare_dataset` has no such guard, and it re-tokenizes the whole eval set on every eval.

Repro (passes on `main`, fails on this branch):

```python
import trl.trainer.dpo_trainer as M

ds = load_dataset("trl-internal-testing/zen", "standard_preference")
n, orig = [0], M._tokenize
M._tokenize = lambda *a, **k: (n.__setitem__(0, n[0] + 1), orig(*a, **k))[1]

trainer = DPOTrainer(
    model="trl-internal-testing/tiny-Qwen3ForCausalLM",
    args=DPOConfig(
        precompute_ref_log_probs=True,
        eval_strategy="steps",
        eval_steps=2,
        max_steps=6,
        per_device_eval_batch_size=2,
        report_to="none",
    ),
    train_dataset=ds["train"],
    eval_dataset=ds["test"],
)
n[0] = 0  # ignore tokenization at init
trainer.train()
assert n[0] == 0, f"eval set re-tokenized {n[0]}x during training"
```

```
main:   OK
branch: AssertionError: eval set re-tokenized 18x during training
```

albertvillanova:
Thanks for your insightful review, @qgallouedec! :hugs: 

The get_eval_dataloader migration turned out to be the wrong fix: it only works if Trainer.evaluate() passes through None when no override is given, but it actually substitutes self.eval_dataset before ever calling get_eval_dataloader. That broke two things flagged in review: reference log-probs being recomputed on an already-precomputed eval set, and the eval set being fully re-tokenized on every scheduled evaluation during train(), not just once.

Closing this PR in favor of a simpler approach.

PR #6405, opened 2026-07-15
Promote `DistillationTrainer` to the stable API

qgallouedec:
Every major recent release ships on-policy logit distillation (Qwen3.x, DeepSeek-V4, GLM-5, Nemotron 3). We had 5 experimental trainers for it, 0 stable.

`DistillationTrainer` is GRPO with the policy-gradient loss swapped for a teacher-KL loss. Only the teacher setup and the loss differ, so GRPO fixes propagate — it already picks one up: both GKD and the experimental Distillation skipped FSDP when preparing the teacher.

The loss is a chunked generalized JSD, modeled on SFT's `chunked_nll`, so the `(B, T, V)` distributions are never materialized. `beta` picks the divergence: `0` forward KL (Qwen3), `1` reverse KL (DeepSeek-V4, GLM-5; the default). Verified equal to GKD's `generalized_jsd_loss`, so the objective is unchanged.

Single-teacher, same-vocabulary only. `trl.experimental.distillation` becomes a FutureWarning shim (KTO precedent, #6175); the teacher server, sparse top-1 and IW-OPD are dropped.

```python
from trl import DistillationTrainer

trainer = DistillationTrainer(
    teacher_model="Qwen/Qwen2.5-7B-Instruct",
    train_dataset=load_dataset("trl-lib/DeepMath-103K", split="train"),
)
trainer.train()
```


qgallouedec:
superseded by https://github.com/huggingface/trl/pull/6640

PR #6416, opened 2026-07-16
Fix BEMA callbacks for PEFT models

DaoyuanLi2816:
## Status

Closed because this approach still has correctness risks for QLoRA weights and reference-model updates. It needs a narrower design before being revisited.

## AI writing disclosure

### DIFF
--- a/tests/experimental/test_bema_for_ref_model.py
+++ b/tests/experimental/test_bema_for_ref_model.py
@@ -1,0 +1,77 @@
+# Copyright 2020-2026 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+from datasets import load_dataset
+from transformers import TrainerControl, TrainerState
+from transformers.utils import is_peft_available
+
+from trl import DPOConfig
+from trl.experimental.bema_for_ref_model import BEMACallback, DPOTrainer
+
+from ..testing_utils import TrlTestCase, require_peft
+
+
+if is_peft_available():
+    from peft import LoraConfig
+
+
+class TestBEMACallback(TrlTestCase):
+    def setup_method(self):
+        self.dataset = load_dataset("trl-internal-testing/zen", "standard_preference", split="train")
+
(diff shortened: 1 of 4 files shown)


DaoyuanLi2816:
Both findings are real, not false positives — thanks, fixed in 01231ba.

**PEFT BEMA checkpoint drops adapters (High)**

Confirmed by inspection and repro: `get_base_model()` returns the base model with LoRA layers still injected in place (e.g. `q_proj` becomes a `peft.tuners.lora.layer.Linear` wrapping `base_layer`/`lora_A`/`lora_B`), so `running_model.state_dict()` keys look like `...q_proj.base_layer.weight` / `...q_proj.lora_A.default.weight` instead of the plain `...q_proj.weight`. Reloading that checkpoint with `AutoModelForCausalLM.from_pretrained` builds a vanilla architecture expecting `...q_proj.weight`, so none of those keys match — the LoRA delta is dropped entirely, and `transformers`' own load report shows the base weight for every LoRA-targeted module coming back "missing" (randomly reinitialized) rather than reloaded. So it's actually worse than "the untouched pretrained model" — the targeted layers come back as fresh random weights.

Fix: `on_train_begin` now keeps the (deep-copied) `PeftModel` wrapper around as `running_peft_model`, and `on_train_end` calls `running_peft_model.merge_and_unload()` before saving, so the checkpoint is a properly flattened model whose keys match the plain architecture. Added `test_peft_model_checkpoint_has_merged_adapter_weights`, which snapshots `running_model`'s own forward output right before the merge and asserts the saved-and-reloaded checkpoint reproduces it. Against the old code it fails with ~94% of logits mismatched (max abs diff ~0.05); against the fix it passes within float-rounding tolerance (`atol=rtol=1e-3`).

**Deepcopy spikes GPU memory (Medium)**

Also confirmed, and measured directly: `copy.deepcopy(model)` preserves the source tensor's device, so for a PEFT model resident on GPU, the deepcopy briefly holds a full second copy on the training device before `.to(self.device)` moves it off. On a 404 MB synthetic model this added exactly 404 MB (100%) of extra peak GPU memory during `on_train_begin`.

Fix: build `running_peft_model` via a pre-populated `copy.deepcopy` memo dict, mapping each source parameter/buffer's `id()` to its own already-`.to(self.device)`'d clone, so deepcopy substitutes the pre-moved tensor directly instead of cloning-then-moving the whole tree — parameters/buffers move one at a time and peak usage on the training device never exceeds steady state. On the same 404 MB model, extra peak GPU memory dropped from 404 MB to 0 MB. Added `test_peft_model_running_model_construction_avoids_device_memory_spike` (gated on `@require_torch_accelerator`), asserting the extra peak stays under 50% of the base model's size.

`tests/test_callbacks.py` (9 passed, 2 skipped — unrelated wandb/comet) and `tests/experimental/test_bema_for_ref_model.py` (2 passed) both pass in full after the change.

DaoyuanLi2816:
Closing this version because the current approach still has correctness risks for QLoRA weights and reference-model updates. It needs a narrower design before it is safe to continue.

PR #6439, opened 2026-07-18
Fix GRPO truncated-completion loss normalization

### LINKED ISSUE #6369: GRPO: dapo/cispo/vespo loss denominator ignores `mask_truncated_completions` (the numerator doesn't)
### Reproduction

`mask_truncated_completions=True` is supposed to fully exclude truncated completions (ones that ran out of room before hitting EOS) from the loss: the docstring says "truncated completions are excluded from the loss calculation.". The reporting for it was fixed in about a year old issue+PR, but calculations never were ( #3476  and #3482 )

For `loss_type` in `["dapo", "cispo", "vespo"]`, the loss is `sum(per_token_loss) / num_items_in_batch`. Right now, only the **numerator** actually excludes truncated tokens. The **denominator** (`num_items_in_batch`) does not, it's computed earlier in `_generate()`, before `mask_truncated_completions` zeroes out those tokens' mask. So a truncated completion's tokens get counted in the denominator (diluting the average) while contributing nothing to the numerator.
 
Issue was discovered while working on the issue (#6367 ) and PR (#6368 ).
### Concrete example

Say a batch has 30 completion tokens total, and 5 of them belong to a truncated completion that `mask_truncated_completions` is supposed to drop entirely.

- Numerator: only sums over the 25 non-truncated tokens (correct).
- Denominator: still 30 (the pre-masking count) instead of 25.

The loss ends up `sum / 30` instead of `sum / 25` at about 17% smaller than it should be, and this shrinks further as more of a batch gets truncated. It's a silent scale error, not a crash, so it's easy to miss.

### Reproduction (real trainer, `loss_type="dapo"`)

With `mask_truncated_completions=True` and a short `max_completion_length` (so the model reliably produces truncated, non-EOS-terminated completions), spying on `_compute_loss`:

```
completions/clipped_ratio: 1        (every row truncated)
num_items_in_batch (denominator):   24
true post-truncation-mask count:    0
per-row completion_mask sums:       [0, 0, 0, 0, 0, 0]
```

The denominator the code actually divides by (24) is nonzero, while the true active-token count after truncation-masking is 0.

A hand-built mixed batch (half the rows truncated) quantifies the exact effect:

```
num_items_in_batch used by the real code path: 24.0
honest post-truncation-masking token count:    12.0
loss with the code's actual denominator:  -0.5
loss with an honest denominator:          -1.0
ratio: 0.5, exactly matching honest_count / raw_count (12/24)
```

The loss magnitude is silently scaled by `honest_token_count / raw_token_count` i.e. truncated tokens are excluded from the sum but still counted in "how many tokens we're averaging over."

### Why I think this is a real, unintended bug (not "working as designed")

#3476 hit the exact same root cause (computing something from `completion_mask` before `mask_truncated_completions` zeroes it) for the **logging** metrics, and got fixed there, the fix explicitly kept a copy of the mask before truncation-masking so logging stays accurate. That fix never touched the **loss denominator** path, which has the identical timing problem.

(Not the same issue as #5619 / PR #6024, which is about `num_items_in_batch` not matching the gradient-accumulation window when `steps_per_generation != gradient_accumulation_steps` a different root cause but same variable. Worth being aware of both if touching this code.)

### Where the code is

- Numerator uses the truncation-masked `completion_mask` (correctly): `mask = completion_mask if "tool_mask" not in inputs else completion_mask * inputs["tool_mask"]` in `_compute_loss`.
- Denominator (`num_items_in_batch`) is computed from `completion_lengths`/`agg_completion_lengths` inside `_generate()`, **before** `mask_truncated_completions` zeroes `completion_mask` for truncated rows (that zeroing happens later, in `_generate_and_score_completions`).

Checked for compensation elsewhere in the pipeline: none found. `GRPOTrainer.compute_loss` accepts the `num_items_in_batch` argument the HF `Trainer` base loop computes automatically, but ignores it and then reads its own internal, pre-truncation-masking value from `inputs["num_items_in_batch"]` instead.

### Suggested fix

Not as simple as "just compute `num_items_in_batch` after masking" since that variable is dual-use: the same value also feeds `state.num_input_tokens_seen` (the training-throughput counter), which *should* reflect tokens actually generated, truncated or not. Changing `num_items_in_batch` itself would silently start undercounting throughput too.

The shape of the fix that avoids that: keep a separate, masking-aware token count specifically for the `dapo`/`cispo`/`vespo` loss denominator, distinct from the one used for `num_input_tokens_seen` which mirrors how #3476 kept a separate copy of the mask for logging rather than changing the shared one. Happy to send a patch either way, just didn't want to hand you a fix that trades one silent-divergence bug for another.

### Scope note

This affects `dapo`, `cispo`, and `vespo` whenever `mask_truncated_completions=True` and at least one completion in the batch gets truncated.

### AI disclosure
same as attached PR (#6368 )

### System Info

- Platform: macOS-26.5.1-arm64-arm-64bit
- Python version: 3.12.6
- TRL version: 1.9.0.dev0+69f5907
- PyTorch version: 2.12.0
- accelerator(s): MPS
- Transformers version: 5.11.0
- Accelerate version: 1.14.0
- Datasets version: 5.0.0
- HF Hub version: 1.19.0
- PEFT version: 0.19.1
- vLLM version: not installed

### Checklist

Functionhx:
# What does this PR do?

Fixes #6369 

This PR ensures that token-normalized GRPO losses use the same effective token mask for both the numerator and denominator.

  Previously, when `mask_truncated_completions=True`, truncated tokens were excluded from the loss numerator but were still counted in
  `num_items_in_batch`. This diluted DAPO, CISPO, and VESPO losses.

  The loss denominator is now computed after applying truncation and tool masks, while the raw generated-token count remains unchanged
  for throughput metrics. Fully masked batches are also handled without division by zero.

  The corresponding duplicated logic is updated consistently in GRPO, DPPO, replay-buffer GRPO, and GSPO-token.
## Before submitting

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **Medium Risk**
> Changes core RL loss scaling across GRPO, DPPO, replay-buffer, and GSPO-token trainers; incorrect masking would directly skew gradients, though behavior is narrowed to truncation/tool masking and guarded with clamp(min=1.0).
> 
> **Overview**
> Fixes **token-normalized GRPO-family losses** so the denominator matches the tokens actually included in the loss when `mask_truncated_completions` and/or tool masking apply.
> 
> **`num_items_in_batch`** is no longer taken from raw completion token counts in `_generate`. It is computed after truncation and tool masks as the global sum of `loss_mask` (`completion_mask`, or `completion_mask * tool_mask`), and passed into DAPO/CISPO/VESPO-style losses. Denominators use **`clamp(min=1.0)`** to avoid division by zero when every token is masked.
> 
> The same masking and denominator logic is applied in **DPPO**, **GRPO with replay buffer** (including a **recompute** of `num_items_in_batch` after replay-buffer replacement), and **GSPO-token** (loss, entropy, and sequence-level importance weights all use the unified mask).
> 
> **`_generate` return tuples** drop the completion-token count used for loss; throughput/token logging still uses gathered completion lengths. Tool handling keys off **`tool_mask_list`** rather than `self.tools` where relevant.
> 
> Tests for masked truncated completions now use **`loss_type="dapo"`** and assert **`train_loss == 0.0`** when all completions are truncated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2ddc3e158910dda76e658c41b2890672f63f688. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -2200,7 +2200,7 @@ class TestGRPOTrainer(TrlTestCase):
             num_generations=3,  # reduce the number of generations to reduce memory usage
             max_completion_length=8,  # reduce the completion length to reduce memory usage
             mask_truncated_completions=True,  # Enable masking of truncated completions
-            loss_type="dapo",
+            loss_type="dapo",  # we test specifically dapo because it normalizes by num_items_in_batch
             report_to="none",
         )
         trainer = GRPOTrainer(
@@ -2212,18 +2212,8 @@ class TestGRPOTrainer(TrlTestCase):
 
         previous_trainable_params = {n: param.clone() for n, param in trainer.model.named_parameters()}
 
-        captured = {}
-        original_compute_loss = trainer._compute_loss
+        trainer.train()
 
-        def capture_compute_loss(model, inputs):
-            captured["completion_tokens"] = inputs["completion_mask"].sum().item()
-            captured["num_items_in_batch"] = inputs["num_items_in_batch"].item()
-            return original_compute_loss(model, inputs)
-
-        with patch.object(trainer, "_compute_loss", side_effect=capture_compute_loss):
-            trainer.train()
-
-        assert captured["num_items_in_batch"] == captured["completion_tokens"]
         assert trainer.state.log_history[-1]["train_loss"] is not None
 
         # Check that the params have changed
@@ -2249,7 +2239,7 @@ class TestGRPOTrainer(TrlTestCase):
             num_generations=3,  # reduce the number of generations to reduce memory usage
             max_completion_length=8,  # reduce the completion length to reduce memory usage
             mask_truncated_completions=True,  # Enable masking of truncated completions
-            loss_type="dapo",
+            loss_type="dapo",  # we test specifically dapo because it normalizes by num_items_in_batch
             report_to="none",
         )
         trainer = GRPOTrainer(
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -2170,93 +2170,178 @@ class TestGRPOTrainer(TrlTestCase):
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
 
-    @patch("transformers.generation.utils.GenerationMixin.generate")
-    def test_train_with_mask_truncated_completions(self, mock_generate):
-        """Test that training works with mask_truncated_completions=True parameter."""
+    @pytest.mark.parametrize("loss_type", ["dapo", "cispo", "vespo"])
+    def test_train_with_mask_truncated_completions(self, loss_type):
+        """Test that token-normalized losses use the final completion mask for normalization."""
 
-        # We mock the generate method because the model's random weights make it extremely unlikely to produce a
-        # sequence containing the EOS token within the allowed max_completion_length. As a result, all tokens are
-        # masked in the loss, the model doesn't update, and the final check (which verifies the update) fails.
+        dataset = Dataset.from_dict({"prompt": ["What is 2 + 2?"]})
+
+        def reward_func(completions, **kwargs):
+            return list(range(len(completions)))
+
+        training_args = GRPOConfig(
+            output_dir=self.tmp_dir,
+            per_device_train_batch_size=3,
+            num_generations=3,
+            max_completion_length=8,
+            mask_truncated_completions=True,
+            loss_type=loss_type,
+            max_steps=1,
+            bf16=False,
+            report_to="none",
+        )
+        trainer = GRPOTrainer(
+            model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
+            reward_funcs=reward_func,
+            args=training_args,
+            train_dataset=dataset,
+        )
+
+        eos_token_id = trainer._tokenizer.eos_token_id
+        pad_token_id = trainer._tokenizer.pad_token_id
+
         def fake_generate(input_ids, **kwargs):
-            # pad_token_id = 151643; eos_token_id = 151645
             completion_ids = torch.tensor(
                 [
-                    [1, 2, 3, 4, 5, 6, 7, 8],  # this one is truncated
-                    [9, 10, 11, 151645, 151643, 151643, 151643, 151643],  # this one contains eos
-                    [12, 13, 14, 15, 16, 17, 18, 151645],  # particular case, eos is generated just within the limit
+                    [1, 2, 3, 4, 5, 6, 7, 8],  # truncated: excluded from the loss
+                    [9, 10, 11, eos_token_id, pad_token_id, pad_token_id, pad_token_id, pad_token_id],
+                    [12, 13, 14, 15, 16, 17, 18, eos_token_id],
                 ],
                 device=input_ids.device,
             )
             return torch.cat([input_ids, completion_ids], dim=1)
 
-        mock_generate.side_effect = fake_generate
+        captured = {}
+        original_compute_loss = trainer._compute_loss
 
-        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+        def capture_compute_loss(model, inputs):
+            captured["completion_tokens"] = inputs["completion_mask"].sum().item()
+            captured["num_items_in_batch"] = inputs["num_items_in_batch"].item()
+            return original_compute_loss(model, inputs)
 
-        training_args = GRPOConfig(
-            output_dir=self.tmp_dir,
-            learning_rate=0.1,  # use higher lr because gradients are tiny and default lr can stall updates
-            per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
-            num_generations=3,  # reduce the number of generations to reduce memory usage
-            max_completion_length=8,  # reduce the completion length to reduce memory usage
-            mask_truncated_completions=True,  # Enable masking of truncated completions
-            report_to="none",
-        )
-        trainer = GRPOTrainer(
-            model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
-            reward_funcs="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
-            args=training_args,
-            train_dataset=dataset,
-        )
+        with (
+            patch.object(trainer.model, "generate", side_effect=fake_generate),
+            patch.object(trainer, "_compute_loss", side_effect=capture_compute_loss),
+        ):
+            trainer.train()
 
-        previous_trainable_params = {n: param.clone() for n, param in trainer.model.named_parameters()}
+        # The raw generated total is 20 tokens, but the truncated 8-token completion is excluded from the loss.
+        assert captured["completion_tokens"] == 12
+        assert captured["num_items_in_batch"] == 12
+        assert np.isfinite(trainer.state.log_history[-1]["train_loss"])
 
-        trainer.train()
-
-        assert trainer.state.log_history[-1]["train_loss"] is not None
-
-        # Check that the params have changed
-        for n, param in previous_trainable_params.items():
-            new_param = trainer.model.get_parameter(n)
-            assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
-
-    def test_train_with_mask_truncated_completions_all_masked(self):
+    @pytest.mark.parametrize("loss_type", ["dapo", "cispo", "vespo"])
+    def test_train_with_mask_truncated_completions_all_masked(self, loss_type):
         """
         Test that when all generated completions are truncated (i.e., none contain an EOS token), and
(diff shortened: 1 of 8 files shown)


trl/experimental/grpo_with_replay_buffer/grpo_with_replay_buffer_trainer.py
```diff
@@ -475,7 +473,15 @@ def _generate_and_score_completions(
...
                 output["tool_mask"] = tool_mask
-            return output
+        else:
+            output = outputs_after_sampling_buffer
+
+        # Recompute the loss denominator after replay-buffer replacement from the final masks used by the loss.
+        loss_mask = (
+            output["completion_mask"] if "tool_mask" not in output else output["completion_mask"] * output["tool_mask"]
+        )
+        output["num_items_in_batch"] = self.accelerator.gather(loss_mask.sum()).sum()
```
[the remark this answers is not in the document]
Functionhx: yeah，you're right.

qgallouedec:
thanks, I was able to reproduce the issue;
```python
from datasets import Dataset


def msg(p):
    return [{"role": "user", "content": p}]


# alternate short (terminating) and long (truncated) so each device batch is mixed
prompts = [
    msg("Reply with only the word: hi"),
    msg("Write a detailed 500-word essay about the ocean."),
    msg("Name one color, one word only."),
    msg("Count slowly from 1 to 100 with commentary."),
]
dataset = Dataset.from_dict({"prompt": prompts * 8})


    return [float(len(c[0]["content"])) for c in completions]


config = GRPOConfig(
    loss_type="dapo",
    mask_truncated_completions=True,
    max_completion_length=64,
    num_generations=2,
    max_steps=2,
    report_to=[],
)
trainer = GRPOTrainer(model="Qwen/Qwen2.5-0.5B-Instruct", reward_funcs=reward_len, args=config, train_dataset=dataset)
orig = trainer._compute_loss.__func__


def patched(self, model, inputs):
    cm = inputs["completion_mask"]
    tm = inputs.get("tool_mask")
    honest = cm if tm is None else cm * tm
    honest_count = self.accelerator.gather(honest.sum()).sum().item()
    used = inputs["num_items_in_batch"].item()
    print(
        f"denom used (raw): {used:.0f} | honest (post-mask): {honest_count:.0f} | per-row sums: {cm.sum(1).tolist()}"
    )
    return orig(self, model, inputs)


trainer._compute_loss = patched.__get__(trainer, type(trainer))
trainer.train()
```

```
denom used (raw): 160 | honest (post-mask): 32 | per-row sums: [4, 11, 0, 0, 2, 10, 2, 3]
denom used (raw): 276 | honest (post-mask): 20 | per-row sums: [3, 2, 0, 0, 0, 11, 4, 0]
```

Functionhx:
Thanks for confirming. This matches the issue, and the PR separates the raw throughput token count from the post-mask loss denominator.

tests/test_grpo_trainer.py
qgallouedec, line 1: I think you can minimize the changes in this file, there seems to be a few unrelated changes

eos_token_id = trainer._tokenizer.eos_token_id instead of the initial hard coded one, comment remove etc

tests/test_grpo_trainer.py
```diff
@@ -2170,94 +2170,179 @@ def test_train_scale_rewards(self, scale_rewards):
...
+        assert captured["num_items_in_batch"] == 0
+        assert captured["loss"] == pytest.approx(0.0)
+        assert trainer.state.log_history[-1]["train_loss"] == pytest.approx(0.0)
+
+        # Check that the params have not changed
         for n, param in previous_trainable_params.items():
             new_param = trainer.model.get_parameter(n)
             assert torch.equal(param, new_param), f"Parameter {n} has changed."
 
+    def test_train_with_tool_mask_and_truncated_completions(self):
```
qgallouedec: I don't think this test is really useful, it uses `rollout_func`, which is not the suggested path for generation

trl/trainer/grpo_trainer.py
```diff
@@ -2200,7 +2200,7 @@ def _generate(self, prompts: list):
         agg_prompt_lengths = self.accelerator.gather(prompt_lengths)
         agg_completion_lengths = self.accelerator.gather(completion_lengths)
         total_prompt_tokens = agg_prompt_lengths.sum()
-        total_completion_tokens = agg_completion_lengths.sum()  # = num_items_in_batch, required for the DAPO loss
+        total_completion_tokens = agg_completion_lengths.sum()
```
qgallouedec: same here, please don't remove the comments

trl/trainer/grpo_trainer.py
```diff
@@ -2345,7 +2345,7 @@ def _generate_and_score_completions(
             completion_ids_list,
             tool_mask_list,
             completions,
-            num_items_in_batch,
+            _total_completion_tokens,
```
qgallouedec: let's just remove this returned value instead no?

trl/trainer/grpo_trainer.py
```diff
@@ -2405,6 +2405,10 @@ def _generate_and_score_completions(
             if tool_mask is not None:
                 tool_mask = tool_mask * (~is_truncated).unsqueeze(1).int()
 
+        # Keep loss normalization separate from `_generate`'s raw completion count, which feeds throughput metrics.
```
qgallouedec: this comment doesn't really make sense outside the context of this pr, please remove it

Functionhx: Agreed, I’ll remove it.

trl/trainer/grpo_trainer.py
Functionhx: Good point,that’s cleaner.

trl/trainer/grpo_trainer.py
Functionhx: Sorry,I’ll restore the comments.

tests/test_grpo_trainer.py
Functionhx: Makes sense.I'll remove it.

tests/test_grpo_trainer.py
Functionhx: Agreed.I’ll restore the original test structure.

Functionhx:
Thanks for the thorough and detailed review!

qgallouedec:
Thanks @Functionhx, the fix is right. I pushed a few changes on top:

**1. GRPO: folded `total_completion_tokens` into the `train` branch.** It's only used for `num_input_tokens_seen`, so no reason to bind a name or compute it outside the branch that consumes it. Also drops it from `_generate`'s return tuple instead of keeping an unused `_total_completion_tokens`.

**2. Aligned RLOO and DPPO, even where unrelated.** Trainers here are self-contained, so duplicated blocks must stay identical, that's what makes a fix in one portable to the others. RLOO had a stale `# = num_items_in_batch, required for the DAPO loss` comment (RLOO has no DAPO loss, nothing to fix there), DPPO was missing a `# Log the metrics` comment. Unrelated to this bug, but leaving them divergent is how these blocks drift.

**3. Reverted most of the replay-buffer changes.** The `tool_mask` plumbing fixes a real but separate pre-existing bug: `update_with_replay_buffer` never carried `tool_mask`, so a buffer swap drops tool masking and tool-result tokens get trained on. Orthogonal to this fix: without the plumbing, the recompute sees `"tool_mask" not in output` and falls back to `completion_mask`, and `_compute_loss` does the identical check, so numerator and denominator still agree and #6369 is still fixed.

Our philosophy on `experimental`: we propagate a fix there when it's straightforward, but experimental code should never block a fix on a stable trainer. That diff was ~60% unrelated plumbing. Reverted to the ~20 lines this fix needs; the tool-mask bug deserves its own PR.

**4. Dropped the `patch.object` capture in the tests.** Asserting on `inputs["num_items_in_batch"]` by wrapping `_compute_loss` tests an internal, brittle, breaks on unrelated refactors, and the failure is never obvious. Coverage is slightly lower; right trade. Kept: `loss_type="dapo"` (the only family normalized by `num_items_in_batch`) and `train_loss == pytest.approx(0.0)` on the all-masked case, which guards the new `.clamp(min=1.0)` against `0/0 → NaN`.

**5. Dropped the GSPO-token test built on `object.__new__(GSPOTokenTrainer)`.** Constructing a trainer without `__init__` and hand-setting attributes is (for some reason) a common coding-agent pattern that we actively avoid; not Pythonic, couples the test to private state, rots on any constructor change. Better less coverage than maintaining that. A better test probably exists, but it's not worth delaying a fix to the stable GRPO trainer over an experimental trainer's coverage.

Functionhx:
Thanks for taking the time to refine the PR and for the detailed explanation. I agree with the changes, especially keeping the stable fix focused and handling the replay-buffer tool_mask issue separately. I’ve reviewed the updates and learned a lot from the cleanup.

qgallouedec (approved):

Issue #4973, opened 2026-02-05
[GRPO] AssertionError in vLLM when using QLoRA (4-bit) due to weight shape mismatch during merge_adapter


ChengHu98:
### Reproduction

I am trying to train a model using GRPOTrainer with QLoRA (4-bit quantization) and vLLM enabled (use_vllm=True). However, the training fails during the generation step when TRL attempts to synchronize weights with vLLM.

GRPOTrainer calls self.model.merge_adapter() to merge LoRA weights before generation.

Because the base model is loaded in 4-bit (BitsAndBytesConfig), merging the LoRA adapter forces the model to dequantize (upcast) the weights to bf16/fp16 to perform the addition.

TRL then sends these dequantized (unpacked) weights to vLLM via self.vllm_client.update_named_param.

However, the vLLM engine was initialized with 4-bit quantization (to match the training setup). It expects packed 4-bit weights (e.g., shape [out_features, in_features // 2]), but receives the dequantized full-shape weights (e.g., [out_features, in_features]).

This triggers an AssertionError in vLLM's weight_loader.

My commands:

for the server:
```
CUDA_VISIBLE_DEVICES=0 trl vllm-serve \
    --model llm/models/Qwen2.5-3B-Instruct
``` 

for the training:
```
accelerate launch  trl_test.py
``` 

My error：
```
File "/.../trl/trainer/grpo_trainer.py", line 1306, in _generate_single_turn
    self._move_model_to_vllm()
  File "/.../trl/trainer/grpo_trainer.py", line 1137, in _move_model_to_vllm
    llm_model.load_weights([(name, param.data)])
  ...
  File "/.../vllm/model_executor/layers/linear.py", line 1237, in weight_loader
    assert param_data.shape == loaded_weight.shape
AssertionError
``` 

### System Info

- Python version: 3.10.19
- TRL version: 0.27.1
- PyTorch version: 2.9.0
- accelerator(s): NVIDIA GeForce RTX 4090, NVIDIA GeForce RTX 4090
- Transformers version: 4.57.6
- Accelerate version: 1.12.0
- Accelerate config: not found
- Datasets version: 4.5.0
- HF Hub version: 0.36.1
- bitsandbytes version: 0.49.1
- DeepSpeed version: 0.18.5
- Liger-Kernel version: not installed
- LLM-Blender version: not installed
- OpenAI version: 1.109.1
- PEFT version: 0.18.1
- vLLM version: 0.12.0

### Checklist

[...]

PR #6440, opened 2026-07-18
Raise a clear error for QLoRA + `vllm_mode="server"` instead of a cryptic vLLM crash

DaoyuanLi2816:
## What does this PR do?

Rejects QLoRA with `vllm_mode="server"` before weight synchronization. Merging a 4-bit PEFT adapter dequantizes its weights, which cannot be loaded by a server expecting packed 4-bit tensors.

The same guard is applied to `VLLMGeneration` and the self-contained `OnlineDPOTrainer`. The error directs users to an unquantized server or colocate mode. This is a fail-fast fix for #4973, not full quantized-server support.

## Tests

- server-mode guard tests for both implementations (3 passed, 1 skipped without local 4-bit support)
- Ruff check and format check on the changed files

## AI writing disclosure

### DIFF
--- a/tests/experimental/test_online_dpo_trainer.py
+++ b/tests/experimental/test_online_dpo_trainer.py
@@ -12,14 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from unittest.mock import MagicMock, patch
+
 import pytest
+import torch
 from datasets import Dataset, DatasetDict, features, load_dataset
-from transformers import AutoModelForCausalLM, AutoModelForSequenceClassification, AutoTokenizer
+from transformers import AutoModelForCausalLM, AutoModelForSequenceClassification, AutoTokenizer, BitsAndBytesConfig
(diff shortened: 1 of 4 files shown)


qgallouedec:
Thanks @DaoyuanLi2816!
The problem is the guard's condition: it only inspects the *training* model, so it fires on every `vllm_mode="server"` run with a 4-bit base, including QLoRA + an unquantized bf16 server, which is the path we documented as working in #4973.

Since `trl vllm-serve` exposes no `--quantization` flag, every TRL-launched server is unquantized, so this would turn the one combination that currently works into a hard error, and the message's own advice ("use an unquantized server") would be unreachable.

or maybe I'm missing something here?

DaoyuanLi2816:
You're right—the guard cannot infer the server's quantization state from the training model, and it would reject the supported QLoRA + unquantized server path. Closing this rather than blocking that configuration.

Issue #6449, opened 2026-07-20
Refactor `DistillationTrainer` and promote it to the stable API


qgallouedec:
Every major recent release ships on-policy logit distillation (Qwen3.x, DeepSeek-V4, GLM-5, Nemotron 3). We have 5 experimental trainers for it and 0 stable.

`experimental.DistillationTrainer` is the best candidate: small core, but a lot of bolt-on axes. The plan is to refactor it in place until it is GRPO-shaped — distillation is GRPO with the policy-gradient loss swapped for a teacher-KL loss — then promote it. Doing it gradually rather than as one big rewrite, so each step stays reviewable.

**End state:** single-teacher, always on-policy, prompt-only datasets, one chunked generalized-JSD loss (`beta` selects forward/reverse/JSD), living in `trl/trainer/` with a deprecation shim left in `trl/experimental/`.

**Ground rules:** one concern per PR; replacements land as add-unwired → switch → delete; the objective stays pinned by a test throughout. Each item below is exactly one PR, stacked on the previous. Only three PRs change behaviour values (14, 35, 38) and they say so in the title. Alignment with GRPO is the north star across code, config, tests and docs: new tests are written GRPO-shaped (no bespoke helpers), and the full test-file rewrite to match `test_grpo_trainer.py` is item 47.

**Re-derived (2026-07-21):** items 14+ were re-ordered after finding that prompt-only training NaNs until `num_items_in_batch` is fixed (a prompt-only dataset has 0 dataset-completion tokens → `num_items = 0` → loss `sum/0`). So the `num_items` fix moves to **item 14** (front of Phase 3), and the collator→`_tokenize_prompts` swap moves into Phase 6 (generation). Full rationale in `distillation_prs/PLAN_REMAINING.md`.

**Re-derived again (2026-07-23):** Phases 5+ were re-ordered strictly **bottom-up (state before consumer)** after finding a forward-reference hazard — GRPO's generation methods *consume* config fields (`min_p`, `generation_kwargs`, …) and ctor attrs (`_dist`, `_tokenizer`, `model_kwarg_keys`, `_last_loaded_step`) that the old order added *later*. Those are now added add-only in **Phase 5** *before* the generation adoption in **Phase 6**. Two features are **dropped from scope entirely**: `cast_lm_head_to_fp32` and continuous batching — they never enter the config, ctor, or `_generate_single_turn` (which also removes the biggest source of forward-references). Net item count 60 → 57.

## Phase 0 — Safety net

[...]

PR #6450, opened 2026-07-20
[DistillationTrainer refactor] Pin the distillation objective before refactoring

qgallouedec:
*Plan step 0.1 — first PR of the distillation refactor sequence. Part of #6449.*

`DistillationTrainer` is about to go through a long refactor: top-k support removal, then a switch to a chunked loss. The implementation is expected to change; the value it computes is not. This adds the tests that hold that line.

Three pins, all seeded, over `beta` ∈ {0, 0.5, 1}, with and without a label mask:

- **vs. a naive reference implementation** written straight from the definition. It forms probabilities explicitly and builds the mixture in probability space, rather than reusing `F.kl_div`'s inverted argument order and the `logsumexp` trick — so it can actually catch an argument-order or mixture-weight regression rather than reproducing one.
- **vs. `GKDTrainer.generalized_jsd_loss`**, which implements the same objective. This is the cross-trainer contract: if it breaks, either the refactor changed the objective or GKD drifted.
- **the temperature path**, which is scheduled to become sampling-only later in the sequence. Pinning it now makes that a visible diff instead of a silent drift.

No production code touched.

### Note

The existing `TestGeneralizedJSDLoss` fixture builds its logits with an unseeded `torch.randn`, so it cannot pin values — it only checks relations (`loss_beta_0 != loss_beta_1` and similar). Left as is; the new class is seeded and additive.

### Risk

Low — tests only.

---

> [!NOTE]
> **Low Risk**
> Tests-only PR with no runtime or training behavior changes; risk is limited to CI coverage of the distillation objective.
> 
> **Overview**
> Adds **regression tests only** (no production changes) to lock `DistillationTrainer.generalized_jsd_loss` before upcoming refactors (top-k removal, chunked loss).
> 
> Introduces a **naive `_reference_generalized_jsd`** helper (explicit probabilities and mixture in probability space, independent of `F.kl_div` tricks) and a seeded **`TestGeneralizedJSDLossIsPinned`** class that asserts the trainer loss matches that reference for `beta` ∈ {0, 0.25, 1.0}, with and without label masks.
> 
> Also pins a **cross-trainer contract** against `GKDTrainer.generalized_jsd_loss` on the same fixtures, and **explicit temperature=2.0** checks so a future “temperature is sampling-only” change must fail tests visibly. Imports `GKDTrainer` for that comparison.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 911b0e5c7d81da00c9d599d32d9f485117678361. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/experimental/test_distillation_trainer.py
```diff
@@ -367,6 +368,81 @@ def test_mask_keeps_forward_and_backward_finite(self):
...
+    computes is not. Any diff that moves these numbers is changing the objective and must say so.
+    """
+
+    def setup_method(self):
+        generator = torch.Generator().manual_seed(42)  # seeded: an unseeded fixture cannot pin anything
+        self.student_logits = torch.randn(2, 3, 5, generator=generator)
+        self.teacher_logits = torch.randn(2, 3, 5, generator=generator)
+        self.labels = torch.tensor([[-100, 1, 2], [-100, -100, 3]])
+
+    @pytest.mark.parametrize("beta", [0.0, 0.5, 1.0])
```
cmpatino: I think we should use a value different from 0.5 since that would help us test the asymmetric case of the generalized JSD loss. 

The function is symmetric when `beta=0.5`, so adding a non-symmetric value like 0.25 will help us uncover issues when we swap student and teacher mixture weights.

cmpatino (commented):
I included a comment about an additional `beta` value to test the asymmetric case of the JSD loss.

tests/experimental/test_distillation_trainer.py
qgallouedec: Thanks, addressed in https://github.com/huggingface/trl/pull/6450/commits/c978997a6f1a2152f3919063166334d4f060bdda

PR #6451, opened 2026-07-20
[DistillationTrainer refactor] Pin on-policy and off-policy training end to end

qgallouedec:
*Plan step 0.2 — stacked on #6450. Part of #6449.*

`lmbda` selects between training on the dataset's own completions (`lmbda=0.0`) and on completions the student generates (`lmbda=1.0`). The trainer is scheduled to become always-on-policy, so both modes are pinned now: the later removal of the off-policy path should be a deliberate deletion with a visibly failing test, not a silent behaviour change.

Asserts the loss is recorded and that every parameter actually moved, with `torch.equal` rather than a tolerance so a stalled update cannot pass.

No production code touched.

### Note

Uses a higher learning rate than the default. Gradients are tiny on the test model and the default rate can stall the update entirely, which would make the "params changed" assertion vacuous.

### Risk

Low — tests only. The `lmbda=1.0` case exercises the generation path on CPU, so it is slower than the rest of the file.

---

> [!NOTE]
> **Low Risk**
> Test-only change. The lmbda=1.0 case runs student generation on CPU and is slower than typical tests in this file.
> 
> **Overview**
> Adds **`test_train_updates_params_on_and_off_policy`**, parametrized over **`lmbda=0.0`** (dataset completions) and **`lmbda=1.0`** (student-generated completions), so both policy modes stay covered until the trainer goes always-on-policy.
> 
> The test runs a short local-teacher train with **higher learning rate**, **teacher weights perturbed** away from the student (non-zero distillation signal), then checks **train loss is logged** and **every student parameter changed** via strict **`torch.equal`**—so a stalled optimizer cannot pass on numerical noise.
> 
> **Tests only**; no production code changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0d532808db206625c53c707b0988213acbf192e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
I'm going to merge this one, it's just adding a test, cc @cmpatino, we can modify later if needed

PR #6456, opened 2026-07-20
[DistillationTrainer refactor] Remove top-k support from the base generalized JSD loss

qgallouedec:
*Plan step 1.6 — stacked on PR #6455. Part of #6449.*

The base trainer always has full teacher logits, so it never needs the top-k approximation — only the extracted server path (which sees just the teacher's top-k logprobs) does. This strips top-k from the base loss and moves its knobs to the server subclass.

- `generalized_jsd_loss` drops the `top_k` / `add_tail` parameters and the entire top-k support-selection block. It is now the pure full-vocabulary path.
- config `loss_top_k` / `loss_add_tail` move from `DistillationConfig` to `ServerDistillationConfig`; the `self.loss_top_k` / `self.loss_add_tail` assignments move from the base `__init__` to the subclass `__init__`; the base `compute_loss` call drops the two kwargs.
- `_add_tail_bucket` was only reachable from the top-k block and the server methods, so it moves to the server trainer module. `_jsd_divergence` stays shared in the base module (base uses its dense branch, server its masked branch).
- `test_loss_normalizes_by_num_items_in_batch` drops its `loss_top_k` parametrization (base is full-vocab only); the server test now imports `_add_tail_bucket` from the server module.

After this, the base `generalized_jsd_loss` is exactly the objective the chunked loss will later replace, and all sparse/top-k machinery lives in `ServerDistillationTrainer`.

### Verification done without pytest

- **The objective is provably unchanged**: `generalized_jsd_loss` matches a naive full-vocab reference for `beta ∈ {0, 0.5, 1}`, with and without labels (this is exactly what the #6450 pin guards). Signature no longer exposes `top_k`/`add_tail`.
- Base module no longer defines `_add_tail_bucket`; the server module does; base config no longer has `loss_top_k`/`loss_add_tail`; server config does.
- The server sparse loss is numerically identical (`0.2492382…`) using its now-local `_add_tail_bucket`.

### Risk

Low–medium. The surviving full-vocab path is guarded by the #6450 numeric pin and re-verified here, so the objective cannot have drifted silently. The risk is confined to the config/test coordination (a stray reference to the removed `loss_top_k` on the base would surface at import/collection time). Run the two experimental test modules to confirm collection.

### Amendment (post-pytest)

Relaxed `test_distillation_trainer_train_runs_with_local_teacher`'s `train_loss >= 0.0` to `>= -1e-4`. That test is self-distillation (teacher == student), so the divergence is approx 0; with this PR making full-vocab reverse-KL the default path (was sparse top-1), the mean loss rounds to a tiny negative (~-2e-12) and tripped the strict bound. The relaxed bound still catches a genuinely negative loss.

Full experimental suite green after both amendments: 49 passed, 1 xfailed.

---

> [!NOTE]
> **Low Risk**
> The surviving full-vocab loss is pinned by existing numeric tests; risk is mainly config/API breakage if callers still set `loss_top_k` on base `DistillationConfig`.
> 
> **Overview**
> **Local `DistillationTrainer` now always uses full-vocabulary generalized JSD/KL.** `generalized_jsd_loss` drops `top_k` / `add_tail` and the support-selection logic; `compute_loss` no longer passes those settings. **`loss_top_k` and `loss_add_tail` move from `DistillationConfig` to `ServerDistillationConfig`**, with the server subclass reading them in its own `__init__` instead of the base trainer.
> 
> **Sparse top-k machinery is scoped to `ServerDistillationTrainer`:** `_add_tail_bucket` relocates from the base trainer module to the server trainer module (tests import it there). Shared **`_jsd_divergence` stays in the base module** for dense and masked server paths.
> 
> Docs drop `loss_top_k` from the base config “key parameters” list. **`test_loss_normalizes_by_num_items_in_batch`** is no longer parametrized over `loss_top_k` because the base path is full-vocab only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93f6f5552a0f6cd94791cd0b8b178864c7dbc0bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/distillation/distillation_trainer.py
```diff
@@ -1041,12 +1026,10 @@ def generalized_jsd_loss(
         beta=0.5,
         temperature=1.0,
         reduction="batchmean",
-        top_k=0,
-        add_tail=True,
         num_items_in_batch=None,
     ):
         """
-        Compute the generalized Jensen-Shannon Divergence loss for knowledge distillation.
+        Compute the generalized Jensen-Shannon Divergence loss for knowledge distillation over the full vocabulary.
```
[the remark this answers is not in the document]
qgallouedec: Docs for this trainer are synced to the refactored API in one dedicated pass at promotion (tracking issue #6449, item 56), rather than per-PR, the page needs restructuring around the new DistillationConfig/ServerDistillationConfig split, not incremental trims. Deferring.

cmpatino:
I'm centralizing the discussion about dropping top-k and external teacher support here. I commented on these topics in https://github.com/huggingface/trl/pull/6454#pullrequestreview-4734504282 and https://github.com/huggingface/trl/pull/6455#pullrequestreview-4735415379.

I think there's a nuance in the effectiveness of the full vocab compared to top-k distillation. On the pro side, full vocabulary distillation should, in principle, be an exact objective for the KL loss we use in distillation. On the con side, multiple works have shown that using full-vocabulary distillation (or higher k for top-k) starts giving diminishing returns on distillation performance (see [this paper](https://arxiv.org/abs/2604.13016v2)). It also restricts our trainer to teachers that fit on the training GPUs, so it effectively rules out teachers with over 30B parameters on most setups.

Given these nuances, I agree we should drop support for top-k and external teachers because they introduce complex logic to include those features. Also, the capabilities are not lost since we're keeping them in experimental trainers, so the community still has access to them.

Hopefully, we can include external teachers in the future with approaches like the one from DeepSeek v4, but I think it's better to drop that for now.

trl/experimental/distillation/distillation_config.py
```diff
@@ -220,22 +211,6 @@ class DistillationConfig(_BaseConfig):
...
-            "union of both for JSD) and re-normalized before computing divergence. "
-            "If 0, the full vocabulary is used (slower but exact)."
-        },
-    )
-    loss_add_tail: bool = field(
-        default=True,
-        metadata={
-            "help": "Whether to append a tail bucket representing the remaining probability mass outside the selected top-k support."
-        },
-    )
```
cmpatino, line 238: These should also be removed from the distillation_trainer.md docs.

qgallouedec: Good catch, removed the loss_top_k / loss_add_tail bullet from distillation_trainer.md's "Usage tips" (and fixed the parameter count).

The remaining top-k references are in the external teacher server example, which is now ServerDistillationConfig territory and gets rewritten in the promotion docs pass (#6449, item 56). Flagging so it's tracked rather than lost.

PR #6458, opened 2026-07-20
[DistillationTrainer refactor] Deprecate `lmbda` on/off-policy mixing

qgallouedec:
*Plan step 2.1 — stacked on PR #6457. First PR of Phase 2. Part of #6449.*

The distillation trainer is converging on always-on-policy training (the GRPO-shaped stable target generates every batch). `lmbda` — which mixes on-policy (student-generated) and off-policy (dataset) completions per gradient-accumulation slice — is being removed.

This PR only deprecates it: a `FutureWarning` fires when `lmbda != 1.0`, pointing at `GKDTrainer` (which exposes the same knob). The default `lmbda=1.0` (fully on-policy) does not warn.

Deprecate-before-remove; the removal lands in #6460 (plan 2.3). Both policy modes are already pinned end to end (#6451), so that removal will be visible and guarded.

### Verification (pytest)

`test_lmbda_deprecation_warns` (warns on `0.5`) and `test_lmbda_default_does_not_warn` (silent on default) added. Full suite: **52 passed, 1 xfailed**. The 10 warnings are the deprecation firing on the existing `lmbda=0.0` tests — expected and harmless (no error-mode filterwarnings).

### Risk

Low — additive warning + two tests. No behaviour change.

### Note

`_make_args` in the base test still defaults to `lmbda=0.0`, so several tests now emit the deprecation. That default is dropped in #6460 when `lmbda` is removed (the trainer becomes always-on-policy and those tests generate instead).

---

> [!NOTE]
> **Low Risk**
> Additive deprecation warning and tests only; no change to distillation training logic.
> 
> **Overview**
> **Deprecates** the `lmbda` on/off-policy mixing knob on `DistillationConfig` ahead of making `DistillationTrainer` always on-policy. Instantiating config with **`lmbda` other than `1.0`** now emits a **`FutureWarning`** that the option will be removed and directs users who need mixing to **`GKDTrainer`**.
> 
> The default **`lmbda=1.0`** path is unchanged and does not warn. Training behavior is unchanged in this PR—only the warning and two pytest cases (`test_lmbda_deprecation_warns`, `test_lmbda_default_does_not_warn`) are added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 94a185d2c0de3de47c99a80c03dd48b4cc359520. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmpatino (approved):
LGTM

PR #6465, opened 2026-07-20
Filter reworded torch.jit.script_method warning on Python 3.14

albertvillanova:
This PR extends the existing PyTorch JIT `filterwarnings` entry so it also silences the Python 3.14+ variant of the `torch.jit.script_method` deprecation warning.

### Motivation

`pyproject.toml` already filters the `torch.jit.script_method` `DeprecationWarning`, but only its original wording (`... is deprecated`). On Python 3.14 PyTorch reworded the same deprecation:
> DeprecationWarning: `torch.jit.script_method` is not supported in Python 3.14+ and may break. Please switch to `torch.compile` or `torch.export`.

```
torch/jit/_script.py:359: DeprecationWarning: `torch.jit.script_method` is not supported in Python 3.14+ and may break. Please switch to `torch.compile` or `torch.export`.
```

Since the message no longer starts with "is deprecated", the existing filter does not match and the warning resurfaces in CI on Python 3.14.

### Solution

Add a targeted `filterwarnings` entry matching the reworded message. Matching stops at `` `torch.jit.script_method` is not supported `` so it stays specific to this deprecation (not broad enough to catch unrelated future `torch.jit.script_method` warnings) while being robust to the Python version number in the text.

### Changes

- Add an `ignore` entry in `pyproject.toml` for the `torch.jit.script_method` "is not supported in Python 3.14+" `DeprecationWarning`.

---

> [!NOTE]
> **Low Risk**
> Test configuration only; no runtime or library behavior changes.
> 
> **Overview**
> Extends pytest `filterwarnings` in `pyproject.toml` so CI on **Python 3.14+** stays quiet for the same upstream PyTorch JIT deprecation that was already ignored when the message said `` `torch.jit.script_method` is deprecated ``.
> 
> Adds a second ignore rule matching `` `torch.jit.script_method` is not supported `` (plus a short comment), since PyTorch reworded that warning on 3.14 and the old pattern no longer matched.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 07b65bb16b8d687362c3412ed76b9387bc85377d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (commented):
right!

PR #6472, opened 2026-07-20
Fix extract_prompt dropping the last shared turn when one completion is a prefix

Osamaali313:
## Problem

`extract_prompt` splits the shared prefix of `chosen`/`rejected` into `prompt` by scanning to the first divergence:

```python
for idx in range(min(len(example["chosen"]), len(example["rejected"]))):
    if example["chosen"][idx] != example["rejected"][idx]:
        if example["chosen"][idx - 1] == " ":  # remove space before the prompt
            idx -= 1
        break
return {
    "prompt": example["chosen"][:idx],
    "chosen": example["chosen"][idx:],
    "rejected": example["rejected"][idx:],
}
```

Two edge cases are handled wrong:

1. **One completion is a prefix of the other.** If `chosen` and `rejected` are equal up to the shorter length (e.g. `chosen = "The sky is blue"`, `rejected = "The sky is blue and vast"`, or conversational where one has an extra turn), the loop finishes **without** `break`, so `idx` is left at `min_len - 1`. The last shared turn/token is then excluded from `prompt` and leaks into `chosen`/`rejected`.

2. **First elements differ (`idx == 0`).** `example["chosen"][idx - 1]` becomes `example["chosen"][-1]` (the last element). For string data that can be a trailing space, which wrongly decrements `idx` to `-1`.

## Reproduction

```
chosen   = "The sky is blue"
rejected = "The sky is blue and vast"   # chosen is a prefix of rejected
```

| | prompt | chosen | rejected |
|---|---|---|---|
| before | `"The sky is blu"` ❌ | `"e"` | `"e and vast"` |
| after | `"The sky is blue"` ✅ | `""` | `" and vast"` |

## Fix

- Add a `for/else` so that when the completions never diverge within the shorter length, `idx = min(len(chosen), len(rejected))` — the full shared prefix becomes the prompt.
- Guard the space check with `idx > 0` so `idx == 0` doesn't index `[-1]`.

Normal cases (a genuine shared prefix followed by divergence) are unchanged.

---

> [!NOTE]
> **Medium Risk**
> Changes preference dataset preprocessing logic; incorrect splits would corrupt training labels, but normal diverging cases are unchanged and the fix is narrowly scoped.
> 
> **Overview**
> Fixes **`extract_prompt`** so implicit preference examples split correctly when `chosen` and `rejected` share a prefix.
> 
> When one completion is a prefix of the other (no divergence within the shorter length), a **`for`/`else`** now sets the split index to the full shared length so the entire common prefix stays in **`prompt`** instead of leaking the last shared turn into **`chosen`**/**`rejected`**.
> 
> The optional space-backtrack before the split is guarded with **`idx > 0`**, avoiding a mistaken read of **`chosen[-1]`** when the first elements already differ.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4bbe5a0fb672bacdeaec6b53f0275e7d6658b56. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/data_utils.py
+++ b/trl/data_utils.py
@@ -629,9 +629,13 @@ def extract_prompt(example: dict[str, Sequence]) -> dict[str, Sequence]:
     """
     for idx in range(min(len(example["chosen"]), len(example["rejected"]))):
         if example["chosen"][idx] != example["rejected"][idx]:
-            if example["chosen"][idx - 1] == " ":  # remove space before the prompt
+            if idx > 0 and example["chosen"][idx - 1] == " ":  # remove space before the prompt
                 idx -= 1
             break
+    else:
+        # chosen and rejected never diverge within the shorter length (one is a prefix of the
+        # other), so the entire shared prefix is the prompt.
+        idx = min(len(example["chosen"]), len(example["rejected"]))
     return {
         "prompt": example["chosen"][:idx],
         "chosen": example["chosen"][idx:],


qgallouedec:
Thanks for the contribution! However, besides the fact that this appears to be largely AI-generated, I don't think it solves a real problem.

`extract_prompt` is intended as a best-effort fallback when a user chooses the implicit-prompt preference dataset format instead of the explicit-prompt format, which is the recommended approach. If the extraction isn't satisfactory, users should switch to the recommended explicit prompt format, as described in the documentation.

Thanks!

Osamaali313:
Thanks, and point taken — fair on both counts.

To be transparent: this came from an automated code-maintenance/bug-fixing pipeline I've been testing for long-running sessions, not a person hand-picking trl. You've highlighted two real gaps in it: (1) it flagged something that isn't actually a problem — `extract_prompt` is a best-effort fallback for the implicit-prompt format, and the right move when extraction isn't ideal is the explicit-prompt format, exactly as your docs say; and (2) it didn't follow the PR template. Both are going straight into the pipeline's guardrails: a much higher bar for what counts as a real, worth-fixing issue, and respecting each repo's contribution template.

Apologies for the noise, and thanks for taking the time to explain the intended design — that's genuinely the most useful signal I could get. Leaving this closed.

qgallouedec:
> this came from an automated code-maintenance/bug-fixing pipeline I've been testing for long-running sessions

Thanks for your transparency.
Could you please exclude `trl` and `transformers` from the list of repositories targeted by this pipeline? We already have our own bots for those repositories, and we'd prefer to avoid having maintainers spend time interacting with external agents instead of human contributors.
Thanks!

Osamaali313:
Done, and sorry for the noise. Both trl and transformers are on the pipeline's exclusion list now, so it won't open PRs or issues on either again.

Honestly, on how this slipped through: the pipeline was screening each repo's AGENTS.md for agent restrictions, but trl's AI policy lives in CONTRIBUTING and the PR template, so it didn't get flagged before the PR went out. That's on us. I've since hardened the guardrails to read CONTRIBUTING and the PR template too, and to back off entirely from any repo with a strict AI/agent policy rather than open anything there. (It also can't force-push or run destructive git commands, for what it's worth.)

One small thing, only if it's useful: trl's restriction isn't in AGENTS.md yet, just in CONTRIBUTING and the template. Since AGENTS.md tends to be the first file agents read, mirroring the "autonomous agents shouldn't open PRs" line there (like transformers already does) would let well-behaved pipelines self-exclude automatically. Thanks for flagging it, and for the disclosure checklist.

PR #6479, opened 2026-07-21
[DistillationTrainer refactor]  Switch tests, docs, and example to prompt-only datasets

qgallouedec:
On-policy distillation only needs prompts (the student generates completions), so the training surface should use prompt-only datasets. Now unblocked by the `num_items` fix (#6478), which stops prompt-only batches from NaN-ing.

- **Tests:** `conversational_language_modeling` → `conversational_prompt_only` in the end-to-end training/init tests. Kept on messages: the deprecation test and `_make_local_trainer` (feeds `test_loss_normalizes`, which reads completion tokens straight from the collator — switches to prompt-only once generation replaces the collator).
- **Docs + example:** quick-start and example script now use the `prompt` column (`trl-lib/ultrafeedback-prompt`, a conversational prompt-only dataset). Also removed the dead `lmbda` / off-policy references (the param was already removed from the config).

Verified: `pytest tests/experimental/test_distillation_trainer.py` — 40 passed.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to documentation, example CLI docstrings, and test dataset fixtures; no production trainer logic is modified in this diff.
> 
> **Overview**
> **Documentation and examples** now describe **fully on-policy** `DistillationTrainer` usage with a **`prompt`** column (prompt-only conversational data) instead of **`messages`** / language-modeling completions. The quick start, dataset format section, and CLI snippets use `trl-lib/ultrafeedback-prompt` (and GSM8K mapped to `prompt`). Stale **`lmbda`** / mixed off-policy guidance and the **external teacher server** section are removed so the page matches the current config surface.
> 
> **Tests** that run end-to-end training, eval init, Liger, and error paths switch from `conversational_language_modeling` to **`conversational_prompt_only`**. `_make_local_trainer` still uses the messages split where tests need completion tokens from the collator without generation; a comment documents that follow-up.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fe497cd4c7b19e3a7e9b176b29fcada57b8ffa11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/source/distillation_trainer.md
```diff
@@ -93,50 +90,43 @@ When using the teacher server:
 
 ### Expected dataset type
 
-The dataset should be formatted as a [conversational](dataset_formats#conversational) [language modeling](dataset_formats#language-modeling) dataset:
+The dataset should be formatted as a [prompt-only](dataset_formats#prompt-only) dataset. The student generates its own completions on-policy, so only the prompt is needed:
```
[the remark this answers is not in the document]
qgallouedec: Good catch, the collator is conversational-only for now, so the standard string form isn't supported yet.

docs/source/distillation_trainer.md
```diff
@@ -74,7 +72,6 @@ config = DistillationConfig(
     teacher_model_server_url="http://teacher-host:8000",
     loss_top_k=1,       # required with teacher server when beta > 0
```
cmpatino: These args don't exist anymore.

tests/experimental/test_distillation_trainer.py
cmpatino, line 259: The eval loss with this new eval dataset will always be zero because we don't generate the rollouts for the eval prompts. The eval dataset consists only of masked prompts, so the loss is 0. This is a shortcoming of our initial implementation of the trainer.

We can perhaps follow the approach from the GRPO trainer, where we also generate the completion when in eval mode.

I think the change is out of scope for this PR, so maybe we can insert an extra step between 14 and 15 that addresses this.

docs/source/distillation_trainer.md
qgallouedec: Good catch, this snippet is stale, thanks, fixing

tests/experimental/test_distillation_trainer.py
```diff
@@ -221,6 +221,8 @@ def _make_args(self, **kwargs):
         return DistillationConfig(**args)
 
     def _make_local_trainer(self, **kwargs):
+        # Messages-format: `_make_batch` reads completion tokens straight from the collator (no generation), which the
+        # `prompt`-only format cannot provide until generation replaces the collator. Switched to prompt-only then.
```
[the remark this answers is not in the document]
qgallouedec: The comment means "will switch to prompt-only then" (once generation replaces the collator, item 26); the helper deliberately keeps messages here. Item 16 removes both the messages usage and this comment.

tests/experimental/test_distillation_trainer.py
qgallouedec: Agreed, eval loss is 0 because we don't generate rollouts in eval (masked prompts only).

It's already resolved later in the stack rather than needing a step between 14–15. At item 26 we adopt GRPO's `_prepare_inputs`, which generates completions in eval too, so eval starts producing real rollouts there. Keeping it out of scope for this PR as you suggest.

cmpatino (approved):

PR #6488, opened 2026-07-21
Add regression tests for evaluating init-time eval datasets after training with precomputed reference log-probs

albertvillanova:
This PR adds regression tests to DPO and KTO verifying that an eval_dataset set at initialization keeps producing correct evaluation metrics after training when precompute_ref_log_probs=True.

### Motivation

Follow-up to the eval-time reference log-prob guard added for DPO and KTO:
- #6443

That guard raises when a new dataset is passed to evaluate() after training in full fine-tuning setups, where the reference model is no longer available. A concern was raised that the guard might also break evaluation of datasets set at initialization. It does not: those datasets have their reference log-probs precomputed once, against the untrained reference, and stored on the dataset, so evaluation reuses the stored values instead of recomputing them against the trained policy. These tests lock in that behavior so it cannot silently regress.

### Solution

Add a parametrized test to each trainer that sets an eval_dataset at initialization, covering a plain Dataset, a DatasetDict, and a dict of datasets, then trains and calls evaluate() with no arguments, asserting it returns per-split metrics without raising.

### Changes

- Add a parametrized regression test to the DPO trainer test suite
- Add the same parametrized regression test to the KTO trainer test suite, kept consistent with the DPO version

---

> [!NOTE]
> **Low Risk**
> Test-only changes with no production code or runtime behavior modifications.
> 
> **Overview**
> Adds **parametrized regression tests** to `tests/test_dpo_trainer.py` and `tests/test_kto_trainer.py` so init-time `eval_dataset` still works after training when `precompute_ref_log_probs=True`.
> 
> Each test builds a trainer with `eval_dataset` at construction (plain `Dataset`, `DatasetDict`, or a dict of datasets), runs a short `train()`, then calls no-arg `evaluate()` and checks per-split loss metrics are returned **without** raising. This complements the existing test that **does** raise when a **new** dataset is passed to `evaluate()` post-training under the same settings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 76fccf295c8ee90819b665742e5951adb783010f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (approved):
thanks!

tests/test_kto_trainer.py
```diff
@@ -425,6 +425,40 @@ def test_evaluate_precompute_ref_log_probs_after_training_raises(self):
...
+        if eval_dataset_type == "dataset":
+            eval_dataset = eval_split
+        elif eval_dataset_type == "dataset_dict":
+            eval_dataset = DatasetDict({"data1": eval_split, "data2": eval_split})
+        else:  # "dict_of_dataset"
+            eval_dataset = {"data1": eval_split, "data2": eval_split}
+
+        training_args = KTOConfig(
+            output_dir=self.tmp_dir,
+            loss_type="apo_zero_unpaired",
```
qgallouedec: I think `loss_type="apo_zero_unpaired"` isn't necessary here

PR #6420, opened 2026-07-16
Async grpo OpenEnv harness rollout

AmineDiro:
## Loop-owning agent training (AsyncGRPO harness) + opencode example

Adds an experimental **loop-owning** path to `AsyncGRPOTrainer` for training external agents that run their own tool loop (e.g. [`opencode`](https://opencode.ai)), plus a self-contained example. Unlike the white-box `environment_factory` path where TRL drives each turn, here the agent runs to completion and TRL trains on its captured proxy trace.

### How it works
- The agent runs in an OpenEnv session in `transparent_proxy` mode; an in-sandbox proxy captures each turn's token ids and logprobs.
- On completion, TRL reads the trace, rebuilds per-turn training rows, and scores the workspace with the session's `verify()`; GRPO trains on those rows.
- One rollout = one isolated session (sandbox + proxy + agent process); `max_inflight_tasks` controls concurrency.

### Caller-supplied policy hooks (typed inputs, sensible defaults)
- `rollout_reward_fn(HarnessRolloutOutcome) -> float | None`: maps the rollout outcome (env reward, tool counts, `timed_out`, trace) to the training reward, or `None` to drop it from the group baseline. Distinct from `verify()`, which only measures task correctness and can't see trajectory signals.
- `train_turn_fn(HarnessTurn) -> bool`: which turns to reinforce (ships `has_tool_call` for action-only agents; default trains all turns).
- `agent_turn_fn(list[TraceEntry]) -> list[TraceEntry]`: which trace entries are real agent turns, so framework aux calls (title generator, context summarizer) are never   trained or scored.

### Also included
- `group_id`-seeded rollouts (stable per-group task selection), graceful session cleanup on stop (no leaked sandboxes), and typed `HarnessRolloutOutcome` / `HarnessTurn` / `TraceEntry` / `LoopOwningSession`.
- Example `examples/scripts/openenv/opencode.py`: self-contained, local subprocess sandbox, `agentica-org/DeepCoder-Preview-Dataset` with a held-out stdin/stdout  verifier. Validated end-to-end on Qwen3-4B.
- Docs: new "Training on harnesses" section in `docs/source/openenv.md` and a row in `docs/source/example_overview.md`.

---

> [!NOTE]
> **Medium Risk**
> Touches experimental async GRPO rollout/scoring paths and spawns external agent processes; failures are mostly dropped as unscorable rollouts, but misconfigured rewards or trace filtering could silently weaken training signal.
> 
> **Overview**
[...]

PR #6491, opened 2026-07-21
Record rollout traces to trackio 

AmineDiro:
# What does this PR do?

- Adds log_rollout_traces: logs rollout samples (full prompt/completion transcript + reward, reward_std, advantage, model_version, completion_tokens) to trackio's Traces
- Wired into `RolloutQueueDataset`: buffers consumed samples and flushes a batch (up to 16) every 8, stepped by policy version so the UI groups by policy.
- Best-effort: any trackio error is caught and logged, never breaks training.

---

> [!NOTE]
> **Low Risk**
> Observability-only path gated on trackio; failures are caught and trace drops are intentional to avoid blocking training.
> 
> **Overview**
> Adds **inspectable rollout traces** in trackio when `"trackio"` is in `report_to`: consumed `RolloutSample`s are logged as prompt+completion messages with reward metrics, advantage, group/model version, and token counts, grouped by **policy version** (`step`).
> 
> **`RolloutQueueDataset`** buffers samples and flushes every 8 to a **daemon drain thread** via a small bounded queue; **`contextvars.copy_context()`** replays the main-thread trackio run on that thread. Full queues drop trace batches so logging never blocks training. **`log_rollout_traces`** is best-effort (warnings only on failure).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 25b446da97c265b8078c44e6228a5db67dc9f884. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/async_grpo/async_grpo_trainer.py
+++ b/trl/experimental/async_grpo/async_grpo_trainer.py
@@ -16,6 +16,7 @@
 import math
 import queue
 import textwrap
+import threading
 import time
 from collections import defaultdict
 from collections.abc import Callable, Iterator
@@ -199,6 +200,15 @@ class RolloutQueueDataset(torch.utils.data.IterableDataset):
         self.report_to = report_to
         self._trace_buf: list = []
         self._trace_log_interval = 8
+        # Log traces off the training path: __iter__ enqueues, a daemon thread drains. Bounded + drop-on-full so a
+        # slow trackio backend never blocks sample delivery.
+        self._trace_queue: queue.Queue = queue.Queue(maxsize=4)
+        threading.Thread(target=self._drain_traces, name="rollout-trace-logger", daemon=True).start()
+
+    def _drain_traces(self):
+        while True:
+            samples, step = self._trace_queue.get()
+            log_rollout_traces(samples, step=step, report_to=self.report_to)
 
     def __iter__(self):
         while True:
@@ -222,7 +232,10 @@ class RolloutQueueDataset(torch.utils.data.IterableDataset):
 
             self._trace_buf.append(sample)
             if len(self._trace_buf) >= self._trace_log_interval:
-                log_rollout_traces(self._trace_buf, step=self.model_version_fn(), report_to=self.report_to)
+                try:
+                    self._trace_queue.put_nowait((self._trace_buf, self.model_version_fn()))
+                except queue.Full:
+                    pass
                 self._trace_buf = []
 
             yield {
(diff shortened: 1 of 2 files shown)


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -188,13 +232,26 @@ def __init__(
         stale_after_s,
         max_staleness=3,
         poll_interval_s=5.0,
+        report_to=(),
```
qgallouedec: Nit, we usually use a default to None for mutables args

qgallouedec (approved):
Thanks, i didn't run myself but it looks good

adithya-s-k (approved):

sergiopaniego:
Tried this on the loop-owning harness path (#6420) with two coding agents (OpenCode and Pi) in remote HF sandboxes, `report_to="trackio"`. The rollout-trace logging is skipped on every flush with:
`[RANK 0] rollout trace logging skipped: RuntimeError: Call trackio.init() before trackio.log().`
The best-effort catch keeps training alive, but no traces are ever logged. It looks like `log_rollout_traces` runs in a context (the daemon background flush thread) where trackio's run is not initialized, while the HF trackio callback inits on the main thread/process. Main-loop metrics logged fine to `.../huggingface/trackio`, only the rollout-trace flush hits this. Repros (Pi and OpenCode harness, HF sandboxes): https://gist.github.com/sergiopaniego/ed71514ff1ae53cf522b23c0b890d939 and https://gist.github.com/sergiopaniego/f33459e2c497ade1422c37c314133de6

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: fixed in [6e4643d](https://github.com/huggingface/trl/pull/6491/commits/6e4643d69b5893bd709ce9135d8bf0a9e0a6f727)

AmineDiro:
@sergiopaniego  Thanks for catching it 👍🏼  

This is a regression I introduced in https://github.com/huggingface/trl/pull/6491/changes/3b650da2a4572058433a0ae25c61fd423b0a5cfd by moving the logging traces from the critical path from training, looks like trackio doesnt have the context in the new thread . 

Fixing 👍🏼

AmineDiro:
@sergiopaniego  fixed in  25b446d. 

Example traces in : https://huggingface.co/spaces/aminediroHF/async-grpo-gsm8k

PR #6492, opened 2026-07-21
Test train_dataset=None raises for core trainers

albertvillanova:
This PR adds a dedicated test to DPOTrainer, KTOTrainer, RewardTrainer, RLOOTrainer and SFTTrainer asserting that constructing the trainer without a `train_dataset` raises `ValueError` with the message `` `train_dataset` is required ``.

### Motivation

We're auditing `train_dataset` handling across core trainers to bring it in line with the test coverage `eval_dataset` already has (e.g. #6336, "Test all supported eval_dataset input types in core trainer init tests"). Each of these five trainers already raises `` ValueError("`train_dataset` is required") `` when `train_dataset` is omitted, but nothing in the test suite exercised that path.

While doing this audit we noticed `GRPOTrainer` already covers this exact case, just incidentally: `test_dataset_required_without_environment` happens to hit the same `` `train_dataset` is required `` error, but its name and framing are about GRPO's environment-factory feature (a dataset is optional only when an environment owns the data), not about the plain "no dataset at all" requirement the other trainers share. A follow-up PR could rename that test (or add a same-named counterpart) so GRPO's coverage reads consistently with this one.

### Solution

Add one `test_train_dataset_required` per trainer (one commit each), each constructing the trainer with a minimal valid config and no `train_dataset`, and asserting the expected `ValueError`.

This is one PR in a small series aligning `train_dataset` handling and test coverage across core trainers with the existing `eval_dataset` work.

### Changes

- `tests/test_dpo_trainer.py`: add `test_train_dataset_required`
- `tests/test_kto_trainer.py`: add `test_train_dataset_required`
- `tests/test_reward_trainer.py`: add `test_train_dataset_required`
- `tests/test_rloo_trainer.py`: add `test_train_dataset_required`
- `tests/test_sft_trainer.py`: add `test_train_dataset_required`

---

> [!NOTE]
> **Low Risk**
> Test-only additions with no production code changes; lowest risk.
> 
> **Overview**
> Adds **`test_train_dataset_required`** to the DPO, KTO, Reward, RLOO, and SFT trainer test modules so init without `train_dataset` is covered the same way as existing **`eval_dataset`** init tests.
> 
> Each test builds the trainer with a minimal config (and **`reward_funcs`** for RLOO) and expects **`ValueError`** with message `` `train_dataset` is required ``. No trainer implementation changes—only regression tests for behavior that was already present but untested.
> 
[...]

PR #6493, opened 2026-07-21
Reject unsupported train_dataset types in core trainers

albertvillanova:
This PR raises a clear `TypeError` in DPOTrainer, GRPOTrainer, KTOTrainer, RewardTrainer, RLOOTrainer and SFTTrainer when `train_dataset` is not a `Dataset` or `IterableDataset`.

### Motivation

This is part of a small series aligning `train_dataset` handling and test coverage across core trainers with the `eval_dataset` work already done (#6492 added the `test_train_dataset_required` coverage that's an earlier step in the same series). Unlike `eval_dataset`, which was extended to support `DatasetDict`, `IterableDatasetDict` and plain `dict` for multi-split evaluation, `train_dataset` intentionally does not support those container types, since training always runs over a single dataset.

Today, passing an unsupported container type as `train_dataset` doesn't fail cleanly. For example, passing a `DatasetDict` to `SFTTrainer` silently iterates over the dict's string keys as if they were examples, and crashes deep in internal preprocessing with a confusing `AttributeError: 'str' object has no attribute 'get'` instead of a clear, actionable error at construction time.

### Solution

Add one explicit, consistently-worded check per trainer, right next to the existing `` `train_dataset` is required `` check: if `train_dataset` isn't a `Dataset` or `IterableDataset`, raise `` TypeError("`train_dataset` must be a `Dataset` or `IterableDataset`, got `{type(train_dataset).__name__}`.") ``.

A follow-up PR will add init tests covering supported and unsupported `train_dataset` container types across all 6 core trainers.

### Changes

- `trl/trainer/dpo_trainer.py`: reject unsupported `train_dataset` types
- `trl/trainer/grpo_trainer.py`: reject unsupported `train_dataset` types
- `trl/trainer/kto_trainer.py`: reject unsupported `train_dataset` types
- `trl/trainer/reward_trainer.py`: reject unsupported `train_dataset` types
- `trl/trainer/rloo_trainer.py`: reject unsupported `train_dataset` types
- `trl/trainer/sft_trainer.py`: reject unsupported `train_dataset` types

---

> [!NOTE]
> **Low Risk**
> Constructor-only input validation with no training logic changes; only affects callers passing invalid dataset types.
> 
> **Overview**
> Adds an early **`TypeError`** in **DPO**, **GRPO**, **KTO**, **Reward**, **RLOO**, and **SFT** trainers when `train_dataset` is not a Hugging Face **`Dataset`** or **`IterableDataset`**.
> 
> The check sits next to the existing required-dataset / `IterableDataset` handling so unsupported containers (e.g. **`DatasetDict`**) fail at construction with a clear message instead of breaking later in preprocessing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0595fa2b322ebf78a76d6d2b45c768936cd2377. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
Thanks! TBH, I'm not fully convinced this is worth it as written. A couple of concerns:

1. **It's not really a "type check", and the benefit is thin.** For genuinely wrong input the error is already clear and immediate, e.g. passing a str gives `AttributeError: 'str' object has no attribute 'column_names'` at init.
    ```python
  
    trainer = SFTTrainer(model="trl-internal-testing/tiny-Qwen3ForCausalLM", train_dataset="trl-lib/Capybara")
    ```
    ```
        trainer = SFTTrainer(model="trl-internal-testing/tiny-Qwen3ForCausalLM", train_dataset="trl-lib/Capybara")
      File "/fsx/qgallouedec/trl/trl/trainer/sft_trainer.py", line 1271, in __init__
        train_dataset = self._prepare_dataset(
            train_dataset, processing_class, args, args.packing, formatting_func, "train"
        )
      File "/fsx/qgallouedec/trl/trl/trainer/sft_trainer.py", line 1394, in _prepare_dataset
        column_names = get_dataset_column_names(dataset)
      File "/fsx/qgallouedec/trl/trl/data_utils.py", line 1026, in get_dataset_column_names
        return list(next(iter(dataset)).keys()) if dataset.column_names is None else dataset.column_names
                                                   ^^^^^^^^^^^^^^^^^^^^
    AttributeError: 'str' object has no attribute 'column_names'
    ```
    which is, imo, quite easy to investigate from.
2. The only case with a confusing error is the dict family (`DatasetDict`/`IterableDatasetDict`), which is iterable and yields string keys → `'str' object has no attribute 'get'`. So the whitelist buys us a nicer message for one specific family, not a broad class of broken input.
    ```python
    from trl import SFTTrainer
    trainer = SFTTrainer(model="trl-internal-testing/tiny-Qwen3ForCausalLM", train_dataset=load_dataset("trl-lib/Capybara"))
    ```
    ```
        trainer = SFTTrainer(model="trl-internal-testing/tiny-Qwen3ForCausalLM", train_dataset=load_dataset("trl-lib/Capybara"))
      File "/fsx/qgallouedec/trl/trl/trainer/sft_trainer.py", line 1271, in __init__
        train_dataset = self._prepare_dataset(
            train_dataset, processing_class, args, args.packing, formatting_func, "train"
        )
      File "/fsx/qgallouedec/trl/trl/trainer/sft_trainer.py", line 1423, in _prepare_dataset
        if is_conversational_from_value(first_example):
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
      File "/fsx/qgallouedec/trl/trl/data_utils.py", line 964, in is_conversational_from_value
        maybe_messages = example.get("conversations")
                         ^^^^^^^^^^^
    AttributeError: 'str' object has no attribute 'get'
   ```

3. **Slippery slope.** If we start validating `train_dataset` types, the natural question is why not the processing class, the peft config, the quantization config and every config field?

I think this kind of type check only makes sense when the wrong type is silently accepted and produces plausible-but-wrong behavior instead of failing on its own. Example

```python
def truncate_at_stops(text, stop_sequences):
    """Cut `text` at the earliest stop sequence."""
    cut = len(text)
    for seq in stop_sequences:
        i = text.find(seq)
        if i != -1:
            cut = min(cut, i)
    return text[:cut]

text = "a < b, then foo<end> bar"

truncate_at_stops(text, ["<end>"])   # -> 'a < b, then foo'   (correct)
truncate_at_stops(text, "<end>")     # -> 'a '                (wrong!)
```

If you really want to fix the actual footgun, I'd rather reject *just* the known-bad types (`DatasetDict`/`IterableDatasetDict`/`dict`) with a clear message, rather than whitelist `isinstance(Dataset)`.

albertvillanova:
Thanks for the detailed review, @qgallouedec.

Done, updated the check to `isinstance(train_dataset, dict)` (covers `DatasetDict`/`IterableDatasetDict`/plain `dict` in one go, since both dict subclasses already extend `dict`), matching the pattern already used everywhere for eval_dataset.

On the "the error is already clear" point: that holds for SFTTrainer, which eagerly preprocesses train_dataset in `__init__`. It doesn't hold uniformly across trainers. GRPOTrainer and RLOOTrainer don't call `_prepare_dataset` on train_dataset at all: it's passed straight through to the base `Trainer` and only touched once the dataloader starts fetching batches. I checked this concretely: constructing `GRPOTrainer` with a `DatasetDict` as `train_dataset` succeeds without error. The failure only surfaces during `trainer.train()`, inside PyTorch's `DataLoader` fetch path:

> KeyError: "Invalid key: 0. Please first select a split. For example: `my_dataset_dictionary['a'][0]`. Available splits: ['a', 'b']"

That's strictly worse than the SFT case: it fails well after construction (after model/optimizer setup), from inside `torch.utils.data`, with a message that actively points at the wrong problem: it reads like a dataset-splits bug, not "you passed the wrong container type for train_dataset." Someone hitting this has to work backward from a PyTorch internals traceback to realize the actual issue is the argument type. I'd say that's exactly the "silently accepted, fails somewhere unrelated" case worth guarding against, even if you don't count the SFT case as compelling on its own.

See errors raised by #6494 before merging this PR:

```python
FAILED tests/test_dpo_trainer.py::TestDPOTrainer::test_init_with_train_dataset[unsupported_dataset_dict] - AttributeError: 'str' object has no attribute 'keys'
FAILED tests/test_reward_trainer.py::TestRewardTrainer::test_init_with_train_dataset[unsupported_dataset_dict] - AttributeError: 'str' object has no attribute 'keys'
FAILED tests/test_kto_trainer.py::TestKTOTrainer::test_init_with_train_dataset[unsupported_dataset_dict] - KeyError: 'chosen'
FAILED tests/test_rloo_trainer.py::TestRLOOTrainer::test_init_with_train_dataset[unsupported_dataset_dict] - Failed: DID NOT RAISE TypeError
FAILED tests/test_grpo_trainer.py::TestGRPOTrainer::test_init_with_train_dataset[unsupported_dataset_dict] - Failed: DID NOT RAISE TypeError
FAILED tests/test_sft_trainer.py::TestSFTTrainer::test_init_with_train_dataset[unsupported_dataset_dict] - AttributeError: 'str' object has no attribute 'get'
= 6 failed
```

qgallouedec:
Right, I see:

```python
>>> from trl import GRPOTrainer
>>> from trl.rewards import accuracy_reward
>>> 
>>> trainer = GRPOTrainer(
...     model="Qwen/Qwen2.5-0.5B-Instruct",
...     reward_funcs=accuracy_reward,
...     train_dataset="trl-lib/DeepMath-103K",
... )
Loading weights: 100%|██████████████████████████████████████████████████████████████████████████████████████████| 290/290 [00:00<00:00, 655.49it/s]
>>> trainer.train()
[transformers] The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config. The model config and generation config were aligned accordingly, being updated with the tokenizer's values. Updated tokens: {'bos_token_id': None, 'pad_token_id': 151643}.
  0%|                                                                                                                       | 0/63 [00:00<?, ?it/s]Traceback (most recent call last):
    trainer.train()
    ~~~~~~~~~~~~~^^
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.13/site-packages/transformers/trainer.py", line 1437, in train
    return inner_training_loop(
        args=args,
    ...<2 lines>...
        ignore_keys_for_eval=ignore_keys_for_eval,
    )
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.13/site-packages/transformers/trainer.py", line 1519, in _inner_training_loop
    self._run_epoch(
    ~~~~~~~~~~~~~~~^
        model=model,
        ^^^^^^^^^^^^
    ...<9 lines>...
        steps_trained_in_current_epoch=steps_trained_in_current_epoch,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.13/site-packages/transformers/trainer.py", line 1747, in _run_epoch
  File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 1545, in training_step
    output = super().training_step(model, inputs, num_items_in_batch)
  File "/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.13/site-packages/transformers/trainer.py", line 1913, in training_step
  File "/fsx/qgallouedec/trl/trl/extras/profiling.py", line 211, in wrapper
  File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 1574, in _prepare_inputs
  File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 2274, in _generate_and_score_completions
               ~^^^^^^^^^^
TypeError: string indices must be integers, not 'str'
```

this weakens/invalidates my first 2 points.

Well, then I think the initial type check actually makes sense for GRPO/RLOO. And since we want something consistent across trainers, it would make sense to have the same type check everywhere, including SFTTrainer.

So, your call, either revert to the initial proposal, or keep targeting dict only.

qgallouedec (approved):

albertvillanova:
Reverting to the whitelist (isinstance(train_dataset, (Dataset, IterableDataset))) is the right call. The str case is exactly the generalization qgallouedec's own reproduction just demonstrated: the dict-only blacklist protects against DatasetDict/dict, but GRPO/RLOO have no validation net at all, so any wrong type (not just dict-family) silently passes construction and detonates deep in training. A whitelist is the only form that closes that gap for every input type, and it also gives genuine cross-trainer consistency rather than "dict is special-cased." Implementing now.

qgallouedec (approved):

Issue #4967, opened 2026-02-04
Support Unbiased "K1-in-Reward" KL Estimator for GRPO


ktolnos:
### Feature request

I propose implementing the K1 KL estimator added to the reward for the GRPO trainer.
Currently, the GRPO implementation calculates the KL penalty using the [K3 estimator](http://joschu.net/blog/kl-approx.html) and adds it directly to the loss.
The recent paper [A Comedy of Estimators: On KL Regularization in RL Training of LLMs](https://www.google.com/url?sa=E&q=https%3A%2F%2Farxiv.org%2Fabs%2F2512.21852) demonstrates that:
- K3-in-Loss (current implementation) results in biased gradients.
- K1-in-Reward (proposed implementation) provides unbiased gradients for the Reverse KL objective.

## Proposed Implementation Plan

There are two option: replacing the current implementation or adding options to the config to switch between the current and the proposed implementations.
We could add a config like kl_estimator or split it into:
- kl_estimator: "k1" (naive log-ratio) or "k3" (Schulman approximation).
- use_kl_in_reward: True/False.

*Current Behavior*: kl_estimator="k3", use_kl_in_reward=False (Calculated in loss).
*Proposed New Defaults*: kl_estimator="k1", use_kl_in_reward=True (Calculated in reward).

Replacing the current implementation would reduce code complexity, adding options will allow to reproduce the old behavior available. Some combinations of options lead to unstable training (e.g. k1 in the loss)

### Motivation

The cited paper shows that while the current K3-in-loss approach is stable, it underperforms unbiased estimators. Switching to K1-in-reward was shown to significantly improve generalization on Out-of-Distribution (OOD) tasks (e.g., Biology/Chemistry/Physics when training on Math) while maintaining or improving In-Domain performance.
Other libraries are already moving in this direction. For example, verl has adopted K1-in-reward as the only option for reverse KL.

### Your contribution

I am ready to implement this change.
I am looking for feedback on the API design: do you prefer a breaking change (replacement) or preserving the legacy K3-in-loss behavior via configuration flags?

LeonEricsson:
[...]

PR #6503, opened 2026-07-22
Default `use_bias_correction_kl=True` in GRPOConfig

gowtham-sai-yadav:
# What does this PR do?

Sets `use_bias_correction_kl=True` as the default in `GRPOConfig`, following @albertvillanova's suggestion in #4967 to open a focused PR for just this default.

GRPO's KL penalty uses Schulman's `k3` estimator added to the loss. Multiplying that term by the importance sampling ratio makes the KL gradient the unbiased reverse-KL gradient, which is the estimator adopted by DeepSeek-V3.2 ([paper](https://huggingface.co/papers/2512.02556)) and the direction verl has moved in. Since `docs/source/grpo_trainer.md` documents `kl` as the divergence between the model and the reference model, that is the gradient the default should follow.

Note (corrected from an earlier version of this description): this is **not** a no-op on-policy. `coef_1` is differentiable, so multiplying the KL term by it changes the gradient even where the ratio's value is exactly 1. The flip therefore changes training dynamics for every run with `beta != 0`, not only off-policy ones. Thanks @albertvillanova for catching that, the reproduction in the review makes it clear.

The unbiased reverse-KL property applies to `importance_sampling_level="token"`. With `"sequence"`, `coef_1` is the sequence-mean ratio of shape `(B, 1)`, so a sequence-level weight is broadcast onto the per-token KL. That is pre-existing behaviour, and the docstring is now qualified accordingly. Happy to open a follow-up issue for it.

Changes:
- `use_bias_correction_kl` now defaults to `True`.
- Docstring and field help reworded: dropped the value-vs-gradient framing and the misleading vLLM example, stated that the KL gradient changes whenever `beta != 0` including on-policy, and qualified the unbiased claim by importance sampling level.
- `test_train_bias_correction_kl` is parametrized over both values instead of being flipped to `False`, so the `True` path stays covered on the Liger side.
- `docs/source/paper_index.md` no longer tells users to enable the flag.

## Before submitting

- [x] Was this discussed/approved via a GitHub issue? Please add a link to it if that's the case. -> #4967
- [x] Did you make sure to update the documentation with your changes? (Config docstring, field help, and `paper_index.md`.)
- [x] Did you write any new necessary tests? (`test_train_bias_correction_kl` is parametrized over both values, covering the new default on the Liger and non-Liger paths.)

## AI writing disclosure

## Who can review?

cc @albertvillanova @qgallouedec


qgallouedec:
thanks! you'll also need to change the test `test_train_with_bias_correction_kl` -> `test_train_without_bias_correction_kl`

gowtham-sai-yadav:
Done, renamed it to `test_train_without_bias_correction_kl` and set `use_bias_correction_kl=False`, so it now covers the opt-out path with the new default. Thanks for the review!

albertvillanova (changes requested):
@gowtham-sai-yadav, thanks for picking this up so quickly, and sorry for the delay in getting back to you!

The direction is right and matches what @LeonEricsson and I converged on in #4967, and the diff itself is minimal, which is exactly what I hoped for. A few things I think may need fixing before we merge, though: mostly because I think the justification is not quite right, and one of them is a coverage regression.

1. The "no-op for on-policy steps" claim is incorrect

This is the main one, because the claim is now shipped in the config docstring and the field help:
> It is a no-op for on-policy steps (where the ratio is 1)

If I understand correctly, `coef_1 = exp(per_token_logps - old_per_token_logps.detach())` has value 1 on-policy, but it is differentiable, so multiplying the KL term by it changes the gradient even when the ratio is exactly 1:

```python
import torch

torch.manual_seed(0)
logits = torch.randn(3, 5, 7, requires_grad=True)
logp = lambda: torch.log_softmax(logits, -1)[..., 0]
ref = logp().detach() - 0.4  # strictly on-policy: old == current

def grad(fn):
    return torch.autograd.grad(fn().sum(), logits, retain_graph=True)[0]

def plain():  # k3 in the loss
    d = ref - logp()
    return torch.exp(d) - d - 1

def corrected():  # k3 * coef_1, coef_1 == 1.0 exactly, but differentiable
    lp = logp()
    d = ref - lp
    return (torch.exp(d) - d - 1) * torch.exp(lp - lp.detach())

def reverse_kl():  # standard unbiased estimator of grad KL(pi || pi_ref)
    lp = logp()
    return (lp - ref).detach() * lp

g_plain, g_corr, g_rev = grad(plain), grad(corrected), grad(reverse_kl)
print("plain vs corrected      max|diff|:", (g_plain - g_corr).abs().max().item())  # 6.9e-02
print("plain vs corrected      rel. norm:", ((g_plain - g_corr).norm() / g_plain.norm()).item())  # 0.21
print("corrected vs reverse-KL max|diff|:", (g_corr - g_rev).abs().max().item())  # 6.0e-08
```

(numbers from torch 2.8.0; they'll shift a little across versions, the point is that the first two are not ~0 and the third is)

And that difference is the entire point of the correction. Analytically, `∇[(π/π_old)·(r − log r − 1)]` collapses to `(π/π_old)·∇log π·log(π/π_ref)`, whose expectation is exactly `∇KL(π‖π_ref)`, which is what the third check above confirms, including at ratio 1. Without the correction the gradient is `(1 − r)·∇log π`, which differs from the forward-KL IS gradient `−r·∇log π` only by the zero-mean score term `∇log π`, i.e. plain k3-in-loss follows the forward KL in expectation.

So the consequence is the opposite of what the PR description says: this flip changes training dynamics for every run with beta != 0, not only off-policy ones. That doesn't make the change wrong: arguably it makes it more of a fix, since docs/source/grpo_trainer.md documents kl as the divergence between the model and the reference model, i.e. reverse KL; but the docstring can't claim it's a no-op. 

Could you reword both the docstring and the help? I'd drop the value-vs-gradient framing entirely rather than patch it, because the other half is off too: k3's unbiasedness as a value also only holds when sampling from π_θ — off-policy it is value-biased as well, and the importance ratio fixes both. Something like: the correction multiplies the KL term by the importance sampling ratio so that the KL gradient becomes the unbiased reverse-KL gradient; it changes the KL gradient whenever beta != 0, including on-policy.

One caveat for the wording: with importance_sampling_level="sequence", coef_1 is the sequence-mean ratio of shape (B, 1) (grpo_trainer.py:3055-3057), so per_token_kl * coef_1 broadcasts a sequence weight onto per-token KL and the gradient picks up an extra (Σ_t k3_t)·∇(mean_t log π_t) term — it is neither the paper's per-token correction nor the unbiased reverse-KL gradient there. That's pre-existing, not something you introduced, but since this PR makes the correction the default and states the property in the docstring, it'd be good to either keep the claim level-qualified or open a follow-up issue.

2. The vLLM generation example is misleading

> in off-policy updates (e.g. num_iterations > 1 or vLLM generation) the plain k3-in-loss gradient is biased

vLLM generation on its own does not make coef_1 != 1: old_per_token_logps is a forward pass of the training model (grpo_trainer.py:2565), so with aligned generation/optimization steps the ratio is still exactly 1. The vLLM↔training mismatch is a separate ratio, and we deliberately apply it to per_token_loss only, never to the KL term (grpo_trainer.py:3117-3122). I'd just drop the vLLM mention and keep num_iterations > 1 / misaligned gradient accumulation.

3. The rename lost coverage of the new default

@qgallouedec's rename request was reasonable, but flipping the value has a side effect worth catching: beta= appears in only five places in tests/test_grpo_trainer.py, and this test is the only one that also exercises Liger. So after this PR nothing tests use_bias_correction_kl=True on the Liger path: and Liger implements the correction in a separate codebase (liger_kernel/chunked_loss/grpo_loss.py), so that's real coverage we'd be dropping.

Could you parametrize over both values instead of flipping, e.g. something like:
```python
@pytest.mark.parametrize("use_bias_correction_kl", [True, False])
@pytest.mark.parametrize("use_liger_kernel", [False, pytest.param(True, marks=require_liger_kernel)])
def test_train_bias_correction_kl(self, use_liger_kernel, use_bias_correction_kl):
```

4. Stale docs

docs/source/paper_index.md:560-568 still says "To enable this feature, set the `use_bias_correction_kl` parameter to `True` in the [`GRPOConfig+ ]" with an example doing so. Needs updating to reflect that it's now the default.

albertvillanova:
For CI errors, please merge main into this PR branch.

gowtham-sai-yadav:
Thanks for the detailed review, and for the reproduction. You're right that the no-op claim was wrong: coef_1 is differentiable, so it changes the KL gradient even where the ratio is exactly 1. I reproduced your script locally and see the same thing (about 21% relative difference on-policy, and the corrected gradient matching reverse-KL to ~1e-8), so I've dropped that framing from the docstring and the help, and from the PR description.

Also fixed:
1. Dropped the vLLM mention, since old_per_token_logps is a forward pass of the training model and the vLLM mismatch ratio isn't applied to the KL term.
2. Qualified the unbiased reverse-KL claim to importance_sampling_level="token", given coef_1 is the (B, 1) sequence-mean ratio for "sequence". Happy to open a follow up issue for that if you'd like it tracked separately.
3. Parametrized test_train_bias_correction_kl over both values instead of flipping it, so the True path stays covered on the Liger side. Good catch, that was a real coverage loss.
4. Updated paper_index.md, which still said to enable the flag.

Merged main into the branch for CI.

albertvillanova (approved):
Thanks, this addresses everything I raised.

On the follow-up issue, yes please, that's worth tracking separately.

PR #6032, opened 2026-06-12
fix: load image-text policy for async grpo

he-yufeng:
# What does this PR do?

Fixes #6028.

`AsyncGRPOTrainer` always loaded the policy with `AutoModelForCausalLM`. For image-text / conditional-generation checkpoints this gives the trainer `model.*` parameter names, while vLLM serves the same checkpoint with `language_model.model.*` and vision-tower keys. The first weight sync can then fail because the two sides do not agree on parameter names.

This PR keeps text models on the existing causal-LM path, but detects image-text / conditional-generation architectures from `AutoConfig.architectures` and loads them with `AutoModelForImageTextToText`. The vision tower parameters are frozen for the text-only RL use case described in the issue, so the trainer namespace matches the vLLM server while avoiding accidental vision-tower updates.

I also updated the trainer argument docs and added focused tests for both loader paths.

## Before submitting

- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request), Pull Request section?
- [x] Was this discussed/approved via a GitHub issue? Please add a link to it if that's the case: #6028.
- [x] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

## AI writing disclosure

We welcome the use of AI tools to help with contributions. For transparency and to help us improve our review process, please indicate the level of AI involvement in this PR.

- [ ] No AI usage: the PR was written entirely by a human.
- [x] AI-assisted: some parts were suggested or improved by AI, but the PR was written and reviewed by a human.
- [ ] AI-generated: the PR was mostly or fully generated by an AI tool.

## Who can review?

Anyone in the community is free to review the PR once the tests have passed.

## To verify

- `python -m py_compile trl\experimental\async_grpo\async_grpo_trainer.py tests\experimental\test_async_grpo_trainer.py`
- `python -m ruff check trl\experimental\async_grpo\async_grpo_trainer.py tests\experimental\test_async_grpo_trainer.py`
- `python -m ruff format --check trl\experimental\async_grpo\async_grpo_trainer.py tests\experimental\test_async_grpo_trainer.py`
[...]

PR #6515, opened 2026-07-23
Async GRPO: vision-language model (image) support

### LINKED ISSUE #6028: AsyncGRPOTrainer: vision-language (*ForConditionalGeneration) checkpoints can't be trained (weight-sync key mismatch)

adithya-s-k:
## What

Enables `AsyncGRPOTrainer` (`trl.experimental.async_grpo`) to train **vision-language models** end-to-end. Closes #6028.

## Why

`AsyncGRPOTrainer` was text-only: it loaded the model with `AutoModelForCausalLM` (text-tower keys `model.*`), which both mismatched the vLLM server's `*ForConditionalGeneration` namespace during NCCL weight sync and left no vision tower for the training forward. VLM checkpoints therefore failed at the first weight sync (#6028).

## Changes (all gated on `self._is_vlm`; the text-only path is byte-for-byte unchanged)

- **Model + processor load**: use `create_model_from_path` (auto-detects `AutoModelForImageTextToText` for `*ForConditionalGeneration`/`*ImageTextToText`) + `AutoProcessor`. Loading the full VLM makes `named_parameters()` match the vLLM server (fixes the weight-sync namespace mismatch) and provides the vision tower.
- **Rollout generation**: for VLMs, render the prompt through the processor (placeholder-expanded ids + pixel tensors) and generate via `/v1/chat/completions` with OpenAI `image_url` parts + `return_token_ids`/`logprobs` — stock vLLM `/v1/completions` has no `multi_modal_data`. Text-only continues to use `/v1/completions`.
- **Plumbing**: `RolloutSample`/`RolloutGroup` carry per-rollout CPU pixel tensors across the mp.Queue; `DataCollatorForRollout` concatenates each row's samples' `pixel_values`/`image_grid_thw`; `compute_loss` passes multimodal `forward_kwargs`.
- **Dispatch**: pixel tensors are kept in a FIFO on the collator (not the returned batch), because `DataLoaderDispatcher` slices batch tensors on dim 0 and prefetches one batch ahead, which corrupts ragged `pixel_values`.
- **Response parsing**: `add_response_schema`/`parse_response` are only needed for tool-call parsing; they're skipped (completion decoded as plain text) when there are no tools/environments, since some VLM chat templates aren't recognized by `add_response_schema`.

## Validation

Trained end-to-end on a multimodal (image → text) RL task, using a stock `vllm serve` with NCCL weight transfer and a single trainer GPU:

| Model | Result |
|---|---|
| `Qwen/Qwen3.5-2B` | ✅ full run, reward increasing, KL ~0.07 |
| `Qwen/Qwen3-VL-2B-Instruct` | ✅ trains, KL ~0.006 |
| `Qwen/Qwen2.5-VL-3B-Instruct` | ✅ trains (cap `max_inflight_tasks` — heavier vision encoder OOMs the server at high concurrency) |

Weight sync (trainer→server, NCCL) confirmed working for all.

## Consistency with the sync trainer

The multimodal handling mirrors the sync `GRPOTrainer` (`create_model_from_path`, `AutoProcessor`, `_is_vlm`, `prepare_multimodal_messages`, `split_pixel_values_by_grid`). Weight-name handling also matches: like sync's `_fix_param_name_to_vllm`, the async streaming now strips `module.` **and** `_checkpoint_wrapped_module.`; neither trainer does any VLM-specific name remap (both send HF names and rely on vLLM's `load_weights`).

## Known limitations / follow-ups

- **Gemma-3 / PaliGemma (vLLM-side, not this PR)**: blocked both ways in vLLM 0.25.1 — (a) native impl: HF names the Siglip vision tower `vision_tower.embeddings.*` while vLLM expects `vision_tower.vision_model.embeddings.*` (weight-sync load fails; would affect sync too); (b) `--model-impl transformers`: vLLM's transformers-backend Gemma3 multimodal forward raises `You must specify exactly one of input_ids or inputs_embeds` at engine init. Both are upstream vLLM issues.
- Validated at `--num_processes 1`; multi-rank needs the per-row pixel scatter (FSDP2) worked through.
- `print_prompt_completions_sample` (`log_completions`) doesn't yet render multimodal prompts.

---

> [!NOTE]
> **High Risk**
> Touches core training forward (packed VLM M-RoPE, multimodal batching), async rollout/vLLM generation, and NCCL weight-name alignment—errors would skew log-probs or break sync without obvious crashes.
> 
> **Overview**
> Adds **end-to-end image RL** to `AsyncGRPOTrainer`, fixing #6028 where text-only loading broke vLLM weight sync and training forwards for VLM checkpoints.
> 
> **Trainer:** Replaces `AutoModelForCausalLM` with `create_model_from_path` and default `AutoProcessor`, sets `_is_vlm` and per-architecture image-mask metadata, and extends `DataCollatorForRollout` to pack `pixel_values` (and related keys) per DP row with `multimodal_counts` for dispatcher scatter. `compute_loss` strips padding, rebuilds per-token image masks on packed rows, and for image rows runs `_packed_vlm_forward_inputs` so Qwen-style **3D M-RoPE** and FlashAttention varlen match vLLM generation. Weight streaming also strips `_checkpoint_wrapped_module.` prefixes like sync GRPO.
> 
> **Rollout worker:** VLMs tokenize via the processor, carry `multimodal_inputs` on samples/groups, and generate through **`/v1/chat/completions`** with base64 `image_url` parts (matching sampling params to the text path). `add_response_schema` is skipped when no tools are exposed (`_environment_exposes_tools`); otherwise completions fall back to plain decode. OpenEnv harness gets JSON tool-argument decoding and aligned chat templates for trace re-tokenization.
> 
> **Tests:** VLM load/metadata checks, collator multimodal merge/pad, and tool-exposure unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca682ec3bf89aa821cf94418475c59ada73867c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AmineDiro:
Thanks @adithya-s-k for taking this one 🙏🏼  ! 

I actually have already reviewed a PR that was closed (AI generated) : https://github.com/huggingface/trl/pull/6032

So maybe there is something you can check it out

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -433,6 +450,22 @@ def torch_call(self, examples: list[Any]) -> dict[str, Any]:
...
+        # `pixel_values`; `compute_loss` reads `self.multimodal_rows[rank]`. Rows with no images contribute `None`.
+        if self.is_vlm:
+            multimodal_rows = []
+            for group in groups:
+                row_mm = [ex["multimodal_inputs"] for ex in group if ex.get("multimodal_inputs")]
+                if row_mm:
+                    multimodal_rows.append({k: torch.cat([m[k] for m in row_mm], dim=0) for k in row_mm[0]})
+                else:
+                    multimodal_rows.append(None)
+            self.multimodal_queue.append(multimodal_rows)
```
AmineDiro: This `deque` doesn't actually work multi-rank, does it ? 
If I understand correctly these pixels from the collator to `compute_loss` past  the dispatcher. If that's a problem with how the dispatcher splits on rank ( like it would split an image at the middle). We should rethink this design entirely. 

I just dislike having 2 communication channels for a very tightly coupled data (pixels, text). 

My first instinct is to do something like pad the pixels: 
```python
  if self.is_vlm:
      per_rank_pv, per_rank_thw = [], []
      for group in groups:                         # one row per rank, in rank order
          row = [ex["multimodal_inputs"] for ex in group if ex.get("multimodal_inputs")]
          per_rank_pv.append(torch.cat([m["pixel_values"]   for m in row]) if row else torch.zeros(0, FEAT))
          per_rank_thw.append(torch.cat([m["image_grid_thw"] for m in row]) if row else torch.zeros(0, 3, dtype=torch.long))

      n_patches = torch.tensor([t.shape[0] for t in per_rank_pv])   # (num_processes,)
      n_grids   = torch.tensor([t.shape[0] for t in per_rank_thw])
      P_max, G_max = int(n_patches.max()), int(n_grids.max())

      # dim 0 == num_processes  ->  dispatcher slices it exactly like input_ids
      batch["pixel_values"]   = torch.stack([F.pad(t, (0, 0, 0, P_max - t.shape[0])) for t in per_rank_pv])
      batch["image_grid_thw"] = torch.stack([F.pad(t, (0, 0, 0, G_max - t.shape[0])) for t in per_rank_thw])
      batch["num_patches"]    = n_patches           # so each rank can strip its padding
      batch["num_grids"]      = n_grids
```

AmineDiro (changes requested):
Thx @adithya-s-k  for the effort 

Core approach is right: loading the full VLM via `create_model_from_path` so param names match vLLM fixes #6028 and matches sync GRPO. Text path is untouched. 

I do see three things that needs to be addressed before merge:

**1. Deque batching**
- Only works at `num_processes=1`: with `dispatch_batches`, only rank 0 runs the collator, so other ranks pop an empty deque.
- It is a fragile second queue that relies on collate and loss staying 1:1 in order. See (my comment on having two channels of correlated data.). Pixels are already on the rollout queue we just need to address how the Dispatcher works.

For me the fix: pad pixels to a rank major `(num_processes, P_max, feat)` tensor like `input_ids`, carry a per rank count, let the dispatcher scatter it. 

**2. Freeze the vision tower**
cc @qgallouedec 
- I suggest maybe defaulting to frozen for this first version to shrink the trainable surface and land a clean baseline especially that we won't need to worry about pixel tokens of VLM training stability and the **packing** problem (see 3. ⏬ )
- We can then add a vision freeze option for the whole model trains. This should be an opt-in user choice

**🔴  3. Packing and position ids (main correctness risk)**
This is a real issue today
- Padding free mode passes 1D `position_ids` and derives `cu_seqlens`. Qwen VLMs needs **3D MRoPE** from `image_grid_thw`, which 1D cannot represent. Image tokens lose their 2D layout, text after an image is misplaced, and training positions do not match vLLM generation at all. So its weird that it trained @adithya-s-k  🤔 ? 
- I checked the API and dug into it we can pass boundaries via `FlashAttentionKwargs` (`cu_seq_lens_q/k`, `max_length_q/k`), which transformers already supports for Qwen VL, freeing `position_ids`.

adithya-s-k:
Thanks for the detailed review, this was really useful.

**1. Deque batching. Done.**

Multimodal fields now ride in the batch as rank major `(num_processes, N_max, ...)` tensors padded on dim 0, with a per rank count so each rank strips its own padding, and the dispatcher scatters them like `input_ids`. The deque is gone. One deviation from your sketch: the counts are per key rather than one global count, because a key's dim 0 is not always the image count (Qwen `pixel_values` is `(sum patches, dim)` while `image_grid_thw` is `(n images, 3)`). Added a multi rank collator test that covers the empty row case too.

While doing this I also found the whitelist of processor fields was dropping `image_position_ids` and `mm_token_type_ids`. That breaks Qwen3-VL on transformers 5.14 with `mm_token_type_ids is missing ... M-RoPE`, and Gemma 4 with an `AttributeError` from its vision tower. It now keeps every non text field, like `GRPOTrainer` does for `forward_kwargs`.

**2. Vision tower freeze.**

Happy to default to frozen with an opt in for training it, if that is the call. Let me know and I will add it.

**3. Packing and position ids. You are right, and I have a repro.**

`Qwen3VLModel.forward` only computes MRoPE when position ids are absent:

```
modeling_qwen3_vl.py:1234
if position_ids is None:
    position_ids = self.compute_3d_position_ids(...)
```

Async GRPO always passes explicit 1D position ids, so that branch never runs. I hooked the text model on a packed two sequence batch and `get_rope_index` was never called, and the position ids arriving at the text model were `(1, 160)`, 2D, not the 3D MRoPE needs. So image tokens get flat sequential positions and training disagrees with what vLLM used to generate.

To your "so it's weird that it trained": packing itself is fine. I checked `_is_packed_sequence` on the ids reaching the text model and the resets survive, so attention is still blocked per sequence. Only the positional encoding is wrong, which degrades quality rather than crashing. It also explains two things I could not account for before: GLM-OCR sits at `ratio` 0.39 and `kl` 2.6 (vs 0.99 and 0.01 for the Qwens), and long runs collapse at a consistent ~900 steps regardless of LR, schedule, clip range or number of generations.

I tried the `FlashAttentionKwargs` route and hit two things worth flagging before I pick a shape for the fix.

First, the kwargs cannot reach the language model through the top level VLM forward. `get_image_features` does `self.visual(pixel_values, grid_thw=..., **kwargs)`, and the vision attention already computes its own `cu_seqlens` for image patches, so passing ours raises `TypeError: flash_attention_forward() got multiple values for keyword argument 'cu_seq_lens_q'`.

Second, the packing helpers are 2D only. `_is_packed_sequence` indexes `position_ids.shape[1]` and `prepare_fa_kwargs_from_position_ids` does `position_ids.reshape(-1)` and looks for zeros. Neither is meaningful for a 3D `(3, B, N)` MRoPE tensor, so handing the text model 3D positions and letting FA infer boundaries does not work either.

So the three obvious options each fail somewhere:

| approach | outcome |
|---|---|
| 1D position ids (today) | packing works, MRoPE silently skipped |
| top level `cu_seq_lens` kwargs | TypeError, collides with the vision tower |
| 3D MRoPE position ids | packing detection breaks, garbage cu_seqlens |

What I think works is splitting the forward for VLMs in `compute_loss`: call `get_image_features` directly, scatter the embeds into `inputs_embeds`, build 3D positions per sequence with `get_rope_index` and concatenate them, then call the text model with `inputs_embeds` plus the `cu_seq_lens` kwargs. The vision tower never sees the FA kwargs and both MRoPE and packing end up correct.

Before I build that, two questions. Does that belong in TRL, or should the packing helpers learn about MRoPE upstream so this is not re-implemented per trainer? And since this is really "padding free packing is incompatible with MRoPE models" rather than anything VLM specific, would you rather it landed as its own PR and kept this one to the plumbing?

albertvillanova (commented):
Thanks @adithya-s-k, this is a clean, well-documented port — reusing the shared multimodal utils and mirroring the sync GRPOTrainer's _is_vlm path is exactly the right approach, and patching the weight-name stripping in both the sync-list build and _streaming_iter keeps the two in step. 

A few blockers before merge: 
1. making multimodal_inputs a required field on RolloutGroup/RolloutSample breaks 4 existing tests (TypeError: missing 1 required positional argument at test_async_grpo_trainer.py:657) — please update the fixtures (or default the field) so the suite is green; 
2. code-quality is red on two unused imports (split_pixel_values_by_grid / unsplit_pixel_values_by_grid) plus a long line — a pre-commit run clears it; 
3. the branch now conflicts with main, so a merge of main is needed. 

Could you also add a small unit test for the new VLM path (e.g. the collator's multimodal_queue ordering) so it doesn't silently regress? Scope-wise this is fine for experimental since it just brings async to parity with sync.

adithya-s-k:
Thanks both. Pushed an update that clears the blockers and lands the M-RoPE fix.

**@albertvillanova**, all three addressed:

1. **Tests.** The fixtures now pass `multimodal_inputs`, so the suite is green again (31 passed / 11 GPU-skipped locally). I kept the field required rather than defaulting it, since every real `RolloutSample`/`RolloutGroup` carries it (`None` for text), and a silent default tends to hide the missing-field case.
2. **Code quality.** Unused `*_pixel_values_by_grid` imports and the long line are gone; `ruff` is clean.
3. **Main.** Merged, no conflicts, PR shows mergeable again.

On the extra test: the multi-rank collator test already covers the new path. It packs the multimodal fields rank-major with per-key counts and includes the empty-row (text-only rank) case, which is exactly where the old deque broke.

**@AmineDiro**, built the forward-split fix from my last comment, and it works. In `compute_loss` for VLMs: `get_image_features`, scatter into `inputs_embeds`, 3D M-RoPE built per packed sub-sequence via `get_rope_index`, then the text model with explicit `cu_seq_lens`. The vision tower never sees the FA kwargs, so no `cu_seq_lens_q` collision, and packing plus M-RoPE are both correct.

Verified against the model's own 3D-M-RoPE forward (each sequence run alone = gold):

| | before (1D positions) | after |
|---|---|---|
| completion-token logprob vs gold (2nd packed seq) | 2.82 off | **0.03** |
| train/gen `ratio`, 5k-step run | drifts | **~1.0, stable, no ~900 collapse** |

One wrinkle worth flagging: **Qwen3.5** is a hybrid model (gated-DeltaNet plus short conv layers interleaved with full attention). Its linear-attention layers segment packed sequences via `cu_seq_lens` (delta rule) and `seq_idx` (the causal conv), and both need their fast kernels (`flash-linear-attention` and `causal-conv1d`); the pure-torch fallbacks silently don't segment, so state leaks across the packed boundary. So the fix passes `seq_idx` too, and those kernels are a runtime dep for hybrid VLMs. Full-attention VLMs (Qwen3-VL) need only the M-RoPE part.

Your two questions:

- **TRL vs upstream helpers.** For now it lives in the trainer's `compute_loss` (self-contained, gated on `_is_vlm`, mirrors the existing path). But you're right that this is really "padding-free packing is incompatible with M-RoPE", not VLM-specific. Teaching `prepare_fa_kwargs_from_position_ids` / `_is_packed_sequence` about 3D positions upstream would stop every trainer re-implementing it. Happy to open that as a follow-up so this doesn't rot.
- **Own PR vs plumbing-only.** I folded it in here because without it the plumbing trains on wrong positions (the ~900-step collapse), so the PR isn't actually correct without it. But if you'd rather keep this one to plumbing and land the packing/M-RoPE fix separately, I'm fine splitting it, just say the word.

albertvillanova (changes requested):
Nice progress — CI is green, the branch is mergeable, and the two substantive items got fixed properly rather than worked around. The deque is gone and the multimodal fields now ride in the batch rank-major with per-key counts (plus three collator tests including the empty-row case), and the M-RoPE verification (2.82 → 0.03 logprob delta vs the gold single-sequence forward) is convincing. Sending `tools`, carrying `tool_calls`/`name`/`tool_call_id` through, and the sampling params are all good fixes.

Three things before merge.

### 1. 🔴 The M-RoPE forward is Qwen-only, but runs for every VLM

`compute_loss` (L1137) gates on `self._is_vlm and "pixel_values" in forward_kwargs`, then `_packed_vlm_forward_inputs` hard-depends on Qwen internals: `forward_kwargs["image_grid_thw"]` (L1070), `vlm.visual.spatial_merge_size` (L1085), `vlm.get_rope_index` (L1097).

Checked against `transformers` main rather than inferred:

```
tiny-Gemma3ForConditionalGeneration     -> ['attention_mask', 'input_ids', 'pixel_values', 'token_type_ids']
tiny-Qwen2_5_VLForConditionalGeneration -> ['attention_mask', 'image_grid_thw', 'input_ids', 'mm_token_type_ids', 'pixel_values']
```

Gemma 3/4 emit `pixel_values` (so the gate opens) but no `image_grid_thw` → `KeyError` at L1070 on the first step carrying an image. `get_rope_index` doesn't exist on Gemma either, and it names its tower `vision_tower`, not `visual`.

What makes this sharper: the PR builds infrastructure specifically for Gemma — `_PER_TOKEN_MM_KEYS` exists so Gemma 3's `token_type_ids` and Gemma 4's `mm_token_type_ids` resolve — and `TestAsyncGRPOTrainerVLM` parametrizes over both tiny Gemma checkpoints. Those tests only assert loading + metadata, so they pass while the forward would crash. A green suite that reads as "Gemma supported" is worse than no test here.

Either fall back to the plain `**forward_kwargs` path for models without the Qwen API, or raise a clear `NotImplementedError` — but the parametrized Gemma tests shouldn't stay as they are either way.

### 2. 🔴 Qwen3-VL deepstack features are silently dropped

`Qwen3VLModel.forward` builds `deepstack_visual_embeds` from `image_outputs.deepstack_features` and injects it into the text stack per layer (`modeling_qwen3_vl.py` L1197-1214, consumed at L840). `_packed_vlm_forward_inputs` takes only `.pooler_output` and then calls the model with `inputs_embeds` and `pixel_values=None`, so that branch never runs and deepstack injection is skipped.

No crash — just a training forward that differs from what vLLM generated with, which is the exact failure class this fix exists to eliminate. `Qwen3-VL-2B-Instruct` is listed as validated, so worth re-running the same logprob-vs-gold check that validated Qwen3.5.

(Two things I suspected here are *not* problems, for the record: `pooler_output` is already a per-image tuple from `torch.split`, so `torch.cat(..., dim=0)` is right; and `flash_attention_forward` takes `**kwargs`, so the extra `seq_idx` is harmless on full-attention layers.)

### 3. The chat template still isn't sent to the server

The payload now sends `tools` with the comment *"the server re-renders the whole conversation, so it needs the same tools"* — the same argument applies to `self.chat_template`, which is exactly the prefix-preserving training template that gets swapped in when tools are in play. vLLM's `ChatCompletionRequest` has a `chat_template` field (`chat_completion/protocol.py` L349), so it's a one-liner:

```python
if self.chat_template:
    payload["chat_template"] = self.chat_template
```

You fixed precisely this class of skew in `openenv_harness.py` this round ("training on a prompt rendered by a different template than the one the rollout was generated with is silent skew, not an error") — the VLM path deserves the same.

### Smaller / process

- `_packed_vlm_forward_inputs` is ~100 lines reimplementing model internals and is the only thing standing between this PR and wrong positional encodings, but no test touches it. A CPU test on a tiny Qwen2.5-VL checkpoint asserting the packed forward's logprobs match a per-sequence reference forward — your own 0.03 check, as a regression test — would be worth more than the four loading tests combined.
- `_pil_to_base64` still duplicates `trl.generation.vllm_client.pil_to_base64` (its docstring says so). Just import it.
- `processor.apply_chat_template` (full image preprocessing) and `_pil_to_base64` are synchronous, inside `async def`, and run once per turn — they stall every other in-flight rollout. `asyncio.to_thread`, and hoist the base64 encode out of the turn loop.
- Docs and typing are unchanged: `processing_class` is still annotated `PreTrainedTokenizerBase | None` with a docstring that says `AutoTokenizer.from_pretrained` and never mentions `ProcessorMixin`, and `docs/source/async_grpo_trainer.md` has no VLM section. Given the sharp edges now known (Qwen-only forward, hybrid models needing `causal-conv1d` + `flash-linear-attention` at runtime, `log_completions` not rendering multimodal prompts), a short "Vision-language models" section is needed before merge.
- Scope: the `openenv_harness.py` changes (`_decode_tool_call_arguments` + chat-template threading) look right and the JSON-decode fix is genuinely valuable, but they change text-only tool-calling behaviour and have nothing to do with VLMs. They'd be better as their own PR — reviewable on their own merits and revertable independently.

### On splitting the M-RoPE fix

You asked whether it should be its own PR. My read: yes, and more so now that it's Qwen-specific and lives in `compute_loss`. It's the same bug in sync `GRPOTrainer`, and per `AGENTS.md` the fix will have to be duplicated there verbatim — much easier to review and keep aligned as a standalone change than folded into a VLM-plumbing PR. The plumbing here is close to landable once the Gemma crash is handled.

@qgallouedec @AmineDiro — that's a maintainer call, along with the still-open vision-tower freeze default.

qgallouedec:
Thanks, two calls asked of me:

**Split the M-RoPE fix.** Yes, agreed with Albert. It's Qwen-only but the gate opens for every VLM. Much easier standalone.

>  It's the same bug in sync GRPOTrainer, and per AGENTS.md the fix will have to be duplicated there verbatim

I don't think that's true though Albert

Drop the parametrized Gemma tests with it, they only assert loading so they stay green while the forward would crash. Raise `NotImplementedError` for non-Qwen, or fall back to plain `**forward_kwargs`.

**Don't freeze the vision tower by default.** We have no freeze handle anywhere and sync `GRPOTrainer` trains the whole VLM, so a frozen default diverges the two for no reason a user asked for.
The motivation was the packing risk, and that's fixed iiuc. If it's needed later we add it to both trainers as an opt-in.

Also take the `openenv_harness.py` changes out, they change text-only tool calling and have nothing to do with VLMs.

Rest of Albert's list still applies.

AmineDiro:
@adithya-s-k

Adding to  @qgallouedec 's comment. We now have support for VLMs for `AsyncGRPOTrainer` : 
https://github.com/huggingface/trl/pull/6839 so this PR scope can be reduces further 👍🏼

qgallouedec:
Closing this to keep the queue tidy. VLM support for `AsyncGRPOTrainer` landed via #6839 at the end of august, so most of what this branch carries is now on main, and the branch itself is conflicting and hasn't moved since the 3rd.

The part that didn't land is the M-RoPE fix, and main still has nothing for it: no `get_rope_index`, no packed VLM forward. That's the piece worth bringing back, standalone as we discussed, with the non-Qwen path raising rather than crashing and without the parametrized Gemma tests that pass while the forward would fail.

Thanks for the work here!, the deque removal and the 2.82 to 0.03 logprob check are both solid.

PR #6530, opened 2026-07-24
[DistillationTrainer refactor] Wire compute_loss to the chunked JSD path; delete the full-logit loss

qgallouedec:
*Item 33 (Group E) — stacked on PR 32. Part of #6449. The loss handover.*

Switches `compute_loss` from the full-logit `generalized_jsd_loss` to the chunked path added at item 32, then deletes the old loss. The teacher's dense distribution is now matched without ever materializing `(B, C, V)` logits.

New `compute_loss` (non-Liger branch):
- Student backbone via `_get_last_hidden_state` routed through `_forward_redirection` (so DDP.forward / `prepare_for_backward` fire); teacher backbone via `_get_last_hidden_state` under `no_grad`.
- `_chunked_divergence_loss(...)` over the two hidden states + each model's `lm_head`, with per-model `logit_scale`/`final_logit_softcapping` sourced from `config` (`getattr`, as SFT does).
- `_forward_redirection` moved to unconditional ctor state (was Liger-only; the chunked path needs it too).

Deleted (orphaned): `generalized_jsd_loss`, `_jsd_divergence`, `_reduce_divergence_loss`, `_get_teacher_logits`. Siblings (GKD/GOLD/IW-OPD/server) keep their own copies. Liger path untouched (removed at item 34).

**Handover proof:** `test_compute_loss_matches_full_logit_reference` asserts the chunked `compute_loss` equals a from-scratch full-logit JSD over the completion positions (`assert_close`; exact at the tiny model's fp32). The item-32 unit tests continue to pin the chunk math.

Tests: deleted `TestGeneralizedJSDLoss`, `TestGeneralizedJSDLossIsPinned`, `_reference_generalized_jsd`, and `test_num_items_in_batch_counts_the_tokens_trained_on` (its `_reduce_divergence_loss` recording closure is gone). `test_loss_normalizes_by_num_items_in_batch` kept (routes through the chunked path now). Suite green.

**Subsumes item 35 (temperature).** The chunked loss has no temperature term (SFT parity), so `compute_loss` no longer divides logits by `self.temperature` — temperature is now sampling-only, which was item 35's goal. At the default `temperature=1.0` this is exactly equivalent to the old loss (the handover proof holds); it changes loss values only for `temperature != 1.0`. Item 35 reduces to a docstring/cleanup (or drops).

---

> [!NOTE]
> **Medium Risk**
> This rewires the core distillation training loss and distributed forward wrapping; regressions would affect loss values and multi-GPU/FSDP behavior, though the new integration test targets parity with the old full-logit objective at default temperature.
> 
> **Overview**
> **`DistillationTrainer.compute_loss`** no longer runs full forward passes to build `(B, C, V)` logits and **`generalized_jsd_loss`**. The default (non-Liger) path delegates to **`_compute_loss`**, which pulls student/teacher **last hidden states**, projects through each **`lm_head`** in **`_chunked_divergence_loss`**, and matches the teacher distribution without materializing full vocab logits.
> 
> **Distributed training:** **`_forward_redirection`** is always set in the constructor (not only for Liger) so student and teacher backbone + chunked projection still trigger DDP **`prepare_for_backward`** and FSDP/DeepSpeed parameter materialization.
> 
> **Removed** from this trainer: **`generalized_jsd_loss`**, **`_jsd_divergence`**, **`_reduce_divergence_loss`**, **`_get_teacher_logits`**. Chunked **`_chunk`** / **`_chunked_divergence_loss`** gain an optional **`temperature`** (logits divided after per-model scale/softcap) and **`compute_loss`** passes **`self.temperature`**.
> 
> **Tests:** Dropped full-logit unit suites and **`_reference_generalized_jsd`**; added **`test_applies_temperature`** on the chunked helper and **`test_compute_loss_matches_full_logit_reference`** as a handover check against an explicit full-logit JSD reference.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1bceaf9335e876b9d8305465e9f121d3d1ca9c66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/distillation/distillation_trainer.py
```diff
@@ -1176,30 +1057,56 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
         )
-        loss = self._reduce_divergence_loss(
-            jsd, completion_mask=completion_mask, num_items_in_batch=num_items_in_batch
+
+        self.teacher_model.eval()
+        unwrapped_teacher = self.accelerator.unwrap_model(self.teacher_model)
+        with torch.no_grad():
+            teacher_hidden_states = self._get_last_hidden_state(
+                unwrapped_teacher, input_ids, attention_mask, logits_to_keep
+            )
```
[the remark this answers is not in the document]
qgallouedec: Fixed in 29cd0ac3. The teacher backbone now goes through its own wrapper via `_forward_redirection`, so FSDP/DeepSpeed materialize its sharded params before the forward, same as the student side.

trl/experimental/distillation/distillation_trainer.py
```diff
@@ -1176,30 +1057,56 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+
+        student_lm_head = unwrapped_student.get_output_embeddings()
+        teacher_lm_head = unwrapped_teacher.get_output_embeddings()
+        student_config, teacher_config = unwrapped_student.config, unwrapped_teacher.config
+
+        loss, _, _ = _chunked_divergence_loss(
+            student_hidden_states,
+            teacher_hidden_states,
+            student_lm_head.weight,
+            teacher_lm_head.weight,
```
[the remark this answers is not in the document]
qgallouedec: Good catch. The student backbone is already redirected, but the chunked projection runs outside the wrapper, so under FSDP the head can be resharded. I'm folding the fix into the next stacked PR, which restructures `compute_loss` to run the whole student loss through `_forward_redirection` (the shape GRPO's `compute_liger_loss` uses), so both the backbone and the `lm_head` projection stay inside the wrapper. Tracking it there rather than expanding this PR.

trl/experimental/distillation/distillation_trainer.py
```diff
@@ -1176,30 +1057,56 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+            completion_mask,
+            self.beta,
+            _CHUNKED_LM_HEAD_CHUNK_SIZE,
```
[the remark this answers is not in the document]
qgallouedec: Applied in dfa8e99e. `temperature` is now threaded through the chunked loss and divides both distributions before the softmax, after any scale/softcapping, so the documented config contract holds and it matches the Liger path (`LigerFusedLinearJSDLoss(temperature=...)`).

trl/experimental/distillation/distillation_trainer.py
```diff
@@ -1086,84 +1054,6 @@ def _prepare_inputs(self, generation_batch: dict[str, torch.Tensor | Any]) -> di
...
-        elif reduction == "mean":
-            return jsd.mean()
-        else:
-            return jsd
-
-    @staticmethod
-    def generalized_jsd_loss(
-        student_logits,
-        teacher_logits,
-        labels=None,
```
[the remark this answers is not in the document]
qgallouedec: Fixed in 3f572a5c. compute_loss now runs the entire chunked loss ( student backbone, lm_head projection, and JSD) through _forward_redirection, mirroring GRPO's compute_liger_loss, so the lm_head matmul stays inside the wrapped forward under FSDP/ZeRO-3 (not just the backbone). The teacher backbone keeps its own redirection. maybe_gather_lm_head_ctx still covers the ZeRO-3 gather inside each chunk.

trl/experimental/distillation/distillation_trainer.py
```diff
@@ -1176,30 +1066,65 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+            )
+
+        student_lm_head = unwrapped_student.get_output_embeddings()
+        teacher_lm_head = unwrapped_teacher.get_output_embeddings()
+        student_config, teacher_config = unwrapped_student.config, unwrapped_teacher.config
+
+        loss, _, _ = _chunked_divergence_loss(
+            student_hidden_states,
+            teacher_hidden_states,
+            student_lm_head.weight,
```
cmpatino, line 1127: Not calling `student_lm_head.full_tensor()` causes a crash when using FSDP2. At this point, the head may be sharded, which can cause a crash during loss computation.

We can do something like the snippet below before calling `_chunked_divergence_loss`.

```python
if isinstance(student_lm_head_weight, torch.distributed.tensor.DTensor):
    student_lm_head_weight = student_lm_head_weight.full_tensor()
    if student_lm_head_bias is not None:
        student_lm_head_bias = student_lm_head_bias.full_tensor()
```

trl/experimental/distillation/distillation_trainer.py
```diff
@@ -1176,30 +1066,65 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+            teacher_lm_head.weight,
```
cmpatino, line 1128: When the `student_lm_head` is a PEFT `BaseTunerLayer`, passing .weight directly uses only the frozen base weight and omits the adapter delta. The head adapter therefore receives no gradient.

We could follow SFT/GRPO trainers that reject PEFT adapters for the `lm_head`, but then we wouldn't have a setup that supports PEFT for all layers.

This technically also applies to `teacher_lm_head`, but I don't think we would support a PEFT-wrapped teacher at all.

WDYT?

trl/experimental/distillation/distillation_trainer.py
```diff
@@ -1176,30 +1066,65 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+    def _compute_loss(self, unwrapped_student, inputs, num_items_in_batch):
+        # Chunked JSD path: project the teacher/student hidden states to vocab logits one chunk at a time (never
+        # materializing the full `(B, C, V)` logits), so the teacher's dense distribution can be matched without
+        # buffering it. Runs inside the student wrapper's forward (see `compute_loss`).
+        input_ids = torch.cat([inputs["prompt_ids"], inputs["completion_ids"]], dim=1)
+        attention_mask = torch.cat([inputs["prompt_mask"], inputs["completion_mask"]], dim=1)
+        completion_mask = inputs["completion_mask"]
+        logits_to_keep = inputs["completion_ids"].size(1)  # only the completion tokens are trained on
+
+        student_hidden_states = self._get_last_hidden_state(
```
cmpatino, line 1101: This call silently breaks some PEFT methods that add virtual tokens through `PeftModel.forward()`. Specifically, `_get_last_hidden_state` skips `PeftModel.forward()`, so the virtual prompt is present during generation but missing during the loss calculation.

The solution here would be to reject any PEFT methods related to prompt learning.

Combined with https://github.com/huggingface/trl/pull/6530/changes#r3666909181, I think we should consider adding an option for non-chunked loss to keep support for PEFT methods. That way non-PEFT users get the efficient loss, but we still keep the functionality for others interested in user the trainer with PEFT methods. This discussion is probably out of scope for this PR, but maybe we can consider it as a step in the refactor plan in https://github.com/huggingface/trl/issues/6449.

tests/experimental/test_distillation_trainer.py
```diff
@@ -442,44 +299,6 @@ def test_train_runs_with_prompt_only_dataset(self):
 
         assert all(torch.isfinite(param).all() for param in trainer.model.parameters())
 
-    def test_num_items_in_batch_counts_the_tokens_trained_on(self, monkeypatch):
```
cmpatino, line 445: I think we should keep a version of this test that checks whether we keep the correct value for `num_items_in_batch`.

This trainer has the following steps.

```
generate completions
    ↓
count valid completion tokens
    ↓
shuffle_sequence_dict
    ↓
split_tensor_dict into accumulation microbatches
    ↓
_prepare_inputs returns each slice
    ↓
compute_loss reads inputs["num_items_in_batch"]
    ↓
_chunked_divergence_loss divides by it
```

We calculate `num_items_in_batch` in the first step and need it to contain the correct value in the last step.

Having a full generation-to-loss test would help us catch an error in the loss scaling.

tests/experimental/test_distillation_trainer.py
```diff
@@ -567,6 +386,48 @@ def test_loss_normalizes_by_num_items_in_batch(self):
         # Doubling the global count exactly halves the loss (sum / num_items is linear in 1/num_items).
         torch.testing.assert_close(loss_double, loss_mean / 2, rtol=1e-4, atol=1e-6)
 
+    def test_compute_loss_matches_full_logit_reference(self):
```
cmpatino, line 389: This test uses `beta=0.5`. That value is symmetric and doesn't detect whether we swapped the teacher and the student. I suggest we parametrize `beta in {0.0, 0.5, 1.0}` or use a non-symmetric value like `beta=0.25`.

trl/experimental/distillation/distillation_trainer.py
qgallouedec: Fixed. Mirrored SFT's chunked path: `full_tensor()` the DTensor `lm_head` weight/bias once before the loop, for both student and teacher. Chunked path in a6621c64, and the Liger path gets the same treatment in 9ac8ce19. Guarded by `isinstance(..., DTensor)`, so it's a no-op off FSDP2.

trl/experimental/distillation/distillation_trainer.py
qgallouedec: Went with rejecting it (like SFT/GRPO/DPO): 38ca60f6 raises if the `lm_head` is a BaseTunerLayer. Unlike GRPO I didn't gate on `use_liger_kernel`, since distillation's chunked path also reads `lm_head.weight` directly.

trl/experimental/distillation/distillation_trainer.py
qgallouedec: You're right. I think that wrapping the lm_head with is quite uncommon. Same for prompt learning, we added it not so long ago to sft. 
I'd suggest to have this as a follow-up after the refactor lands. Because supporting it means, as you say, having another implementation for the loss, which is something I'd avoid if it's something that is not used in practice

tests/experimental/test_distillation_trainer.py
qgallouedec: Agreed. I'll add a grad-accum-invariance test ((bs=8, ga=1) vs (bs=4, ga=2)) in the test-rewrite pass (item 47 of #6449): it guards the `num_items_in_batc`h scaling end-to-end, which is the guarantee the deleted recording-closure test gave.
Tell me if you'd prefer to add this test now instead

tests/experimental/test_distillation_trainer.py
qgallouedec: Fixed in 8802bf05 switched the handover test to a non-symmetric beta=0.25, so a student/teacher swap now changes the loss and fails the test.

Issue #6483, opened 2026-07-21
SFTTrainer crashes on Qwen3.5 because forward is functools.partial


Eraly-ml:
# SFTTrainer crashes on Qwen3.5 multimodal model because `forward` is a `functools.partial`

## Description

`trl.SFTTrainer` crashes during initialization when used with the official `Qwen/Qwen3.5-0.8B` checkpoint from Transformers.

The failure occurs before training starts, while TRL tries to apply its chunked cross-entropy LM-head patch. The patch assumes that `model.forward` is a bound method and accesses `original_forward.__func__`, but Qwen3.5's decorated multimodal forward is a `functools.partial` and does not expose `__func__`.

This prevents using `SFTTrainer` with `Qwen3_5ForConditionalGeneration`, even for text-only training with the vision tower frozen.

## Minimal reproduction

```python
import torch
from datasets import load_dataset
from transformers import AutoTokenizer, Qwen3_5ForConditionalGeneration
from peft import LoraConfig
from trl import SFTConfig, SFTTrainer

model_id = "Qwen/Qwen3.5-0.8B"

model = Qwen3_5ForConditionalGeneration.from_pretrained(
    model_id,
    torch_dtype=torch.float16,
    device_map="auto",
)

# Text-only experiment: freeze vision explicitly.
for parameter in model.model.visual.parameters():
    parameter.requires_grad = False

model.config.use_cache = False

tokenizer = AutoTokenizer.from_pretrained(model_id)
dataset = load_dataset(
    "Eraly-ml/qwen35-kazakh-instruction-pilot",
    split="train",
)

# The dataset contains a `messages` column. Convert it to a text column.
def format_example(example):
    return {
        "text": tokenizer.apply_chat_template(
            example["messages"],
            tokenize=False,
            add_generation_prompt=False,
            enable_thinking=False,
        )
    }

dataset = dataset.map(format_example)

peft_config = LoraConfig(
    r=16,
    lora_alpha=32,
    lora_dropout=0.05,
    bias="none",
    task_type="CAUSAL_LM",
    target_modules=[
        "q_proj", "k_proj", "v_proj", "o_proj",
        "out_proj",
        "gate_proj", "up_proj", "down_proj",
    ],
)

args = SFTConfig(
    output_dir="./qwen35-test",
    num_train_epochs=1,
    per_device_train_batch_size=1,
    max_length=512,
    report_to="none",
)

trainer = SFTTrainer(
    model=model,
    args=args,
[...]

PR #6539, opened 2026-07-25
Fix SFTTrainer chunked-CE patch crash when forward is not a bound method

verma8076:
# What does this PR do?

Fixes the crash in `_patch_chunked_ce_lm_head` reported in #6483. The issue title says it's a Qwen3.5 problem, but it's actually broader: the patch assumes `model.forward` is always a bound method and reads `original_forward.__func__` to build the patched forward's signature. That breaks for *any* model whose forward has been wrapped by accelerate's device-dispatch hooks — which is exactly what `device_map="auto"` does when accelerate decides to dispatch/offload the model.

`accelerate.hooks.add_hook_to_module` replaces `model.forward` with:

```python
module.forward = functools.update_wrapper(functools.partial(new_forward, module), old_forward)
```

A `functools.partial` has no `__func__`, hence the `AttributeError` in the issue.

**Fix:** use `inspect.unwrap(original_forward)` instead of `original_forward.__func__`. `functools.update_wrapper` always sets `__wrapped__` on the wrapper, pointing back at `old_forward`, so `inspect.unwrap` recovers the original callable regardless of whether it went through an accelerate hook. It also happens to correctly handle another layer of wrapping that's unrelated to this bug: some `transformers` forwards are themselves decorated with `@functools.wraps` (e.g. via `can_return_tuple`), which also sets `__wrapped__`. I checked that `inspect.unwrap` produces a byte-identical signature to the old `.__func__` access on a plain (non-hooked) model, so this isn't a behavior change for the common case — it just also works for the hooked one.

I reproduced the crash locally against a small model wrapped with `accelerate.hooks.add_hook_to_module` (no GPU or the actual Qwen3.5 checkpoint needed — the bug is in signature introspection, not model computation), confirmed the fix resolves it, and confirmed a real forward/backward pass through the patched model still produces correct results afterward.

## Before submitting

This wasn't pre-discussed with a maintainer — it's a small, self-contained bug fix (no public API or checkpoint format change) with a repro and regression test, so I went ahead and opened the PR directly rather than waiting on a design discussion. Happy to adjust if there's a reason to handle it differently.

## AI writing disclosure

To be upfront about this: the investigation, fix, and test were done by Claude Code under my direction — I reviewed the reasoning and the diff, ran the tests myself, and I'm able to discuss/defend the change, but I want to flag the AI involvement honestly rather than undersell it.

## Who can review?

@qgallouedec — you commented on the original issue and might have context on why the patch does the `.__func__` dance in the first place.

---

> [!NOTE]
> **Low Risk**
> Small, localized fix to signature copying during SFT chunked-CE patching, with a targeted regression test and no training API changes.
> 
> **Overview**
> Fixes **`AttributeError`** when **`_patch_chunked_ce_lm_head`** runs on models whose **`forward`** is not a plain bound method (e.g. **`device_map="auto"`** via Accelerate **`functools.partial`**, or Transformers **`@wraps`** layers).
> 
> **Signature introspection** now uses **`inspect.signature(inspect.unwrap(original_forward))`** instead of **`original_forward.__func__`**, so the patched forward keeps the real parameter list for **`generate`** / VLMs without changing behavior on unwrapped models.
> 
> Adds a **CPU regression test** that wraps **`forward`** with **`AlignDevicesHook`**, asserts patching succeeds, and checks loss matches the reference model.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 571c9cbb57b85d76b7392891427be2c000803391. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -23,6 +23,7 @@ import pytest
 import torch
 import torch.nn.functional as F
 import transformers
+from accelerate.hooks import AlignDevicesHook, add_hook_to_module
 from accelerate.utils.memory import release_memory
 from datasets import Dataset, DatasetDict, IterableDatasetDict, load_dataset
 from packaging.version import Version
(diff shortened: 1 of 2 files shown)


qgallouedec:
closing, see https://github.com/huggingface/trl/issues/6483#issuecomment-5086471355

verma8076:
Makes sense for this specific report — that config avoids ever hitting the partial-wrapped forward, so closing the issue on that basis is reasonable.

Just want to flag (not to relitigate the close, your call either way): the underlying bug this PR fixes is independent of Qwen3.5 specifically. Any model whose `forward` gets wrapped by `accelerate.hooks.add_hook_to_module` — e.g. real multi-GPU `device_map="auto"` dispatch, not just this reporter's setup — will still hit the same `AttributeError: '...' object has no attribute '__func__'` inside `_patch_chunked_ce_lm_head`. The fix is small, self-contained, and has a regression test that reproduces the crash on CPU without needing Qwen3.5 at all.

Happy to reopen if that's useful, but no worries if you'd rather leave it closed for now — just wanted the general case on record in case someone else hits it from a different angle.

PR #3862, opened 2025-08-07
🔮 Native VLM support for `SFTTrainer`

qgallouedec:
```python
from trl import SFTConfig, SFTTrainer
from datasets import load_dataset

trainer = SFTTrainer(
    model="Qwen/Qwen2.5-VL-3B-Instruct",
    args=SFTConfig(max_length=None),
    train_dataset=load_dataset("trl-lib/llava-instruct-mix", split="train"),
)
trainer.train()
```

---

all CI failures unrelated

PR #6547, opened 2026-07-26
Allow packing and padding-free on VLMs with text-only data

### LINKED ISSUE #6545: Packing is refused for VLMs even on text-only datasets
`SFTTrainer` gates packing on the *model* being a VLM, not on the *data* containing images. So a text-only conversational dataset can't be packed on a VLM checkpoint, even though nothing image-related is involved.

## Repro

```python

ds = load_dataset("trl-internal-testing/zen", "conversational_language_modeling", split="train")

SFTTrainer(
    model="trl-internal-testing/tiny-Qwen2VLForConditionalGeneration",
    train_dataset=ds,
    args=SFTConfig(packing=True),
)
```

```
ValueError: Packing is not supported for vision-language models. Please set `packing=False` in the SFTConfig.
```

Same for `padding_free`, `assistant_only_loss` and `truncation_mode="keep_end"`.

packing isn't allowed when we process data on-the-fly, which is needed for image-dataset (prohibitively epxensive to pre-process upfront), but not for VLM. We should support this case.

cris96spa:
# What does this PR do?

`SFTTrainer` refuses `packing`, `padding_free`, `assistant_only_loss` and `truncation_mode="keep_end"` when the
*model* is a VLM, even if the *dataset* has no images:

```python
ds = load_dataset("trl-internal-testing/zen", "conversational_language_modeling", split="train")

SFTTrainer(
    model="trl-internal-testing/tiny-Qwen2VLForConditionalGeneration",
    train_dataset=ds,
    args=SFTConfig(packing=True),
)
# ValueError: Packing is not supported for vision-language models. Please set `packing=False` in the SFTConfig.
```

That dataset's only column is `messages`. Nothing image-related happens, the text pipeline handles it fine, and TRL
already tests that path (`test_train_vlm_text_only_data`). Losing packing here is not free either: it is the main
throughput lever on short-sequence SFT.

This PR gates the four options on `_is_vision_dataset` instead of `_is_vlm`.

Fixes #6545

### Why the dataset is the right condition

The four options are incompatible with on-the-fly processing, and on-the-fly processing is chosen by the data. The
rest of the trainer already knows this: the collator choice (`sft_trainer.py:1190`), `_skip_prepare_dataset` (`1256`),
the padding-free `max_length` guard (`1248`), `_set_signature_columns_if_needed` (`1637`) and
`_reject_skip_prepare_without_labels` (`1651`) all read `_is_vision_dataset`. Only the four gates read `_is_vlm`.

Those gates are simply older than the flag. #3862 added them when `_is_vlm` was the only signal. #4080 introduced
`_is_vision_dataset` and moved the collator and `skip_prepare_dataset` onto it, but left the gates behind. This
finishes #4080.

DPO and KTO already have the shape proposed here: probe the dataset, run the vision-dataset-on-a-text-model check,
then gate the options on `_is_vision_dataset` (`dpo_trainer.py:718-738`, `kto_trainer.py:758-777`). `AGENTS.md`
requires that duplicated trainer logic stays consistent, and SFT is the one core trainer still gating on the model.

### What changed

`_is_vision_dataset` was computed 160 lines below the gates, so the probe and the mismatch check that depends on it
move up to just above them. `completion_only_loss` reads the same `dataset_sample`, so the dataset is still read once.
The four messages now say "vision datasets", the wording DPO and KTO already use.
`_patch_chunked_ce_lm_head(..., is_vlm=self._is_vlm)` is a question about model structure and keeps the model flag.
Two docstrings in `sft_trainer.py` and `sft_config.py` described the same model-versus-data confusion and are
corrected.

Image datasets still refuse all four options, same exception, same point in `__init__`. No API, config or default
changes.

### Tests

`test_train_vlm_text_only_data_uses_text_pipeline` trains on the five configurations that used to raise, then asserts
every `model.visual` parameter is bit-identical afterwards. That assertion is what shows the text path ran: a
text-only batch gives the vision tower no gradient.

`test_vision_dataset_unsupported_options_raise` covers the four gates on `zen-image`. On `main`,
`grep -rn "Packing is not supported" tests/` returns nothing, so the refusal had no coverage at all.

Nine parametrised cases in total. All nine fail on `main` and pass with the change, and the existing
`test_train_vlm_text_only_data` cases still pass alongside them. The wider sweep,
`-k "packing or padding_free or assistant_only or vlm or vision or completion_only"`, gives 54 passed, 20 skipped, 0
failed. The skips are `@require_kernels` and GPU-only cases, since I ran on CPU. `pre-commit run --all-files` is
clean.

### Open questions

1. `keep_end` is deprecated and goes in v2.0.0. Tell me if you would rather that gate stay on `_is_vlm`, and I drop it
   from the change and from the tests.
2. Rewording that message makes it diverge from `dpo_trainer.py:727`, which gates on `_is_vision_dataset` but still
   says "vision-language models". Two lines, no behaviour change. Here or in its own PR?
3. `gold_trainer.py:886` has the same bug. I left it out because the experimental tree runs on a separate workflow.
   Happy to fold it in if you want it.

One pre-existing hole worth naming: `_is_vision_dataset` is read from the first training example only, and the eval
dataset is never probed. This PR widens the second by one option. Probing every dataset is a bigger change and looked
like its own issue.

## Before submitting

- [x] Was this discussed/approved via a GitHub issue? Please add a link to it if that's the case. -> #6545
- [x] Did you make sure to update the documentation with your changes? (two docstrings corrected)

## AI writing disclosure

## Who can review?

@qgallouedec (issue author), @sergiopaniego (reviewed #4080, which introduced `_is_vision_dataset`)

---

> [!NOTE]
> **Low Risk**
> Behavioral fix in trainer init and docs with a focused regression test; image datasets keep the same restrictions, with no auth or data-path changes.
> 
> **Overview**
> **`SFTTrainer` no longer blocks packing, padding-free, assistant-only loss, or `truncation_mode='keep_end'` solely because the checkpoint is a VLM.** Those checks now key off whether the train set actually contains `image` / `images`, matching DPO/KTO and the rest of SFT (collator, `skip_prepare_dataset`, etc.).
> 
> Vision **datasets** still raise the same errors; only text-only data on a VLM processor is affected. Dataset probing and the text-vs-VLM mismatch check move up in `__init__` so gates run on `_is_vision_dataset`. Docstrings in `SFTConfig` / `SFTTrainer` describe dataset-based behavior instead of model-based.
> 
> Adds **`test_train_vlm_text_only_data_packing`** (regression for #6545): trains a tiny Qwen2.5-VL on conversational text with `packing=True` and asserts non-visual weights update while `model.visual` stays frozen.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c7128cc543fe5dd8500f9220fc3cebc57f1864a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -2062,6 +2062,77 @@ class TestSFTTrainer(TrlTestCase):
             else:
                 assert not torch.equal(param, new_param), f"Param {n} is not updated"
 
+    @pytest.mark.parametrize(
+        "config_kwargs",
+        [
+            {"packing": True, "max_length": 32},
+            {"packing": True, "packing_strategy": "wrapped", "max_length": 32},
+            {"padding_free": True, "max_length": None},
+            {"assistant_only_loss": True},
+            {"truncation_mode": "keep_end", "max_length": 32},
+        ],
+    )
+    @ignore_warnings(message="You are using packing, but the attention implementation is not.*", category=UserWarning)
+    @ignore_warnings(message="Padding-free training is enabled, but the attention.*", category=UserWarning)
+    @ignore_warnings(message="The `'keep_end'` truncation mode is deprecated.*", category=FutureWarning)
+    @require_vision
+    def test_train_vlm_text_only_data_uses_text_pipeline(self, config_kwargs):
+        # These options are incompatible with on-the-fly image processing, not with VLMs. A text-only dataset goes
+        # through the regular text pipeline, so they must be available on a VLM checkpoint too. Regression test
+        # for #6545.
+        dataset = load_dataset("trl-internal-testing/zen", "conversational_language_modeling", split="train")
+
+        training_args = SFTConfig(
+            output_dir=self.tmp_dir,
+            learning_rate=0.1,  # use higher lr because gradients are tiny and default lr can stall updates
+            report_to="none",
+            **config_kwargs,
+        )
+        trainer = SFTTrainer(
+            model="trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration",
+            args=training_args,
+            train_dataset=dataset,
+        )
+
+        previous_trainable_params = {n: param.clone() for n, param in trainer.model.named_parameters()}
+
+        trainer.train()
+
+        assert trainer.state.log_history[-1]["train_loss"] is not None
+
+        # Check that the params have changed
+        for n, param in previous_trainable_params.items():
+            new_param = trainer.model.get_parameter(n)
+            if n.startswith("model.visual"):
+                torch.testing.assert_close(param, new_param, rtol=1e-12, atol=1e-12, msg=f"Param {n} is updated")
+            else:
+                assert not torch.equal(param, new_param), f"Param {n} is not updated"
+
+    @pytest.mark.parametrize(
+        ("config_kwargs", "error_match"),
+        [
+            ({"packing": True}, "Packing is not supported for vision datasets"),
+            ({"padding_free": True}, "Padding-free training is yet not supported for vision datasets"),
+            ({"assistant_only_loss": True}, "Assistant-only loss is not yet supported for vision datasets"),
+            (
+                {"truncation_mode": "keep_end", "max_length": 32},
+                "truncation_mode='keep_end' is not supported for vision datasets",
+            ),
+        ],
+    )
+    @ignore_warnings(message="The `'keep_end'` truncation mode is deprecated.*", category=FutureWarning)
+    @require_vision
+    def test_vision_dataset_unsupported_options_raise(self, config_kwargs, error_match):
+        dataset = load_dataset("trl-internal-testing/zen-image", "conversational_language_modeling", split="train")
+
+        training_args = SFTConfig(output_dir=self.tmp_dir, report_to="none", **config_kwargs)
+        with pytest.raises(ValueError, match=error_match):
+            SFTTrainer(
+                model="trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration",
+                args=training_args,
+                train_dataset=dataset,
+            )
+
     @require_vision
     def test_vision_dataset_with_text_model_raises(self):
         dataset = load_dataset("trl-internal-testing/zen-image", "conversational_language_modeling", split="train")
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -2073,7 +2073,6 @@ class TestSFTTrainer(TrlTestCase):
             output_dir=self.tmp_dir,
             learning_rate=0.1,  # use higher lr because gradients are tiny and default lr can stall updates
             packing=True,
-            max_length=32,
             report_to="none",
         )
         trainer = SFTTrainer(
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -2062,31 +2062,19 @@ class TestSFTTrainer(TrlTestCase):
             else:
                 assert not torch.equal(param, new_param), f"Param {n} is not updated"
 
-    @pytest.mark.parametrize(
-        "config_kwargs",
-        [
-            {"packing": True, "max_length": 32},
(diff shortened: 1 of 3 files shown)


tests/test_sft_trainer.py
```diff
@@ -2062,6 +2062,77 @@ def test_train_vlm_text_only_data(self, model_id, dataset_config):
...
+    @pytest.mark.parametrize(
+        "config_kwargs",
+        [
+            {"packing": True, "max_length": 32},
+            {"packing": True, "packing_strategy": "wrapped", "max_length": 32},
+            {"padding_free": True, "max_length": None},
+            {"assistant_only_loss": True},
+            {"truncation_mode": "keep_end", "max_length": 32},
+        ],
+    )
```
qgallouedec: no we don't need this

tests/test_sft_trainer.py
```diff
@@ -2062,6 +2062,77 @@ def test_train_vlm_text_only_data(self, model_id, dataset_config):
...
+    @ignore_warnings(message="You are using packing, but the attention implementation is not.*", category=UserWarning)
+    @ignore_warnings(message="Padding-free training is enabled, but the attention.*", category=UserWarning)
+    @ignore_warnings(message="The `'keep_end'` truncation mode is deprecated.*", category=FutureWarning)
+    @require_vision
+    def test_train_vlm_text_only_data_uses_text_pipeline(self, config_kwargs):
```
qgallouedec: no, just `test_train_vlm_text_only_data_packing`

tests/test_sft_trainer.py
```diff
@@ -2062,6 +2062,77 @@ def test_train_vlm_text_only_data(self, model_id, dataset_config):
...
+    def test_vision_dataset_unsupported_options_raise(self, config_kwargs, error_match):
+        dataset = load_dataset("trl-internal-testing/zen-image", "conversational_language_modeling", split="train")
+
+        training_args = SFTConfig(output_dir=self.tmp_dir, report_to="none", **config_kwargs)
+        with pytest.raises(ValueError, match=error_match):
+            SFTTrainer(
+                model="trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration",
+                args=training_args,
+                train_dataset=dataset,
+            )
```
qgallouedec: no you can remove, it's out of the scope here

qgallouedec (changes requested):
thanks, a few simplifications

tests/test_sft_trainer.py
cris96spa: Done

tests/test_sft_trainer.py
cris96spa: Done

tests/test_sft_trainer.py
cris96spa: Updated with focus on packing only

cris96spa:
Thank you. I addressed the changes, glad to contribute!

tests/test_sft_trainer.py
```diff
@@ -2062,6 +2062,40 @@ def test_train_vlm_text_only_data(self, model_id, dataset_config):
...
+    def test_train_vlm_text_only_data_packing(self):
+        # Packing is incompatible with on-the-fly image processing, not with VLMs. A text-only dataset goes through
+        # the regular text pipeline, so packing must be available on a VLM checkpoint too. Regression test for #6545.
+        dataset = load_dataset("trl-internal-testing/zen", "conversational_language_modeling", split="train")
+
+        training_args = SFTConfig(
+            output_dir=self.tmp_dir,
+            learning_rate=0.1,  # use higher lr because gradients are tiny and default lr can stall updates
+            packing=True,
+            max_length=32,
```
```suggestion
```

qgallouedec (approved):

tests/test_sft_trainer.py
cris96spa: Applied

qgallouedec:
thanks!

PR #6564, opened 2026-07-28
Close in-flight harness sessions concurrently on shutdown

sergiopaniego:
## What does this PR do?

Fixes an orphan-sandbox leak on shutdown in the loop-owning AsyncGRPO path. In-flight OpenEnv sessions were closed one by one with blocking `close()` calls in the event loop, and for a remote sandbox backend each `close()` is a network teardown, so closing several in series can exceed the 15s shutdown window and orphan the rest. This closes them concurrently instead.

Validated on a real `HFSandboxBackend` run: 0 orphan sandboxes after teardown with the fix, 1 to 8 without.

## AI writing disclosure

- [x] AI-assisted

cc @AmineDiro

---

> [!NOTE]
> **Medium Risk**
> Shutdown/teardown behavior changes for live harness sessions; incorrect concurrent close handling could still leak resources, but the logic mirrors the prior per-session try/except close.
> 
> **Overview**
> **Fixes orphan sandbox leaks** when the loop-owning AsyncGRPO harness shuts down with multiple in-flight OpenEnv sessions.
> 
> On stop, `_close_live_sessions_on_stop` no longer calls blocking `session.close()` serially on the event loop. Each close runs in a worker thread via `asyncio.to_thread`, and **`asyncio.gather`** waits for all of them together so remote sandbox teardown can overlap instead of stacking past the shutdown window.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65148339cac46fc07dd89351d742fc82c987667d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/openenv_harness.py
```diff
@@ -158,12 +158,17 @@ async def _generate_one(self, prompt, tool_dict, tools, group_id=0):
...
+
+            def _close(session):
                 try:
                     session.close()
                 except Exception:
                     logger.warning("closing in-flight harness session on stop failed", exc_info=True)
 
+            # Close concurrently: for a remote backend each close() is a network teardown, so closing them in
+            # series can exceed the shutdown window and orphan the rest.
+            await asyncio.gather(*(asyncio.to_thread(_close, session) for session in list(self._live_sessions)))
```
AmineDiro, line 169: if closing is sync, we are just spawning multiple threads, maybe we can do it in the same threadpool before closing the threadpool. 
Anyway I am not sure this cleanup is problematic. What does `session.close()` do concretely for example ?

sergiopaniego: `session.close()` is a remote sandbox teardown for the HF backend (network call), or a local process kill for the subprocess one. On an early stop the in-flight sessions still hold live remote sandboxes, so without closing them they leak: a small HF Jobs run showed orphans without the fix and `ORPHAN_SANDBOXES_AFTER=0` with it

Can't reuse `self._session_pool`: at stop its workers are all blocked in `wait_for_completion` on the exact sessions we're closing, so closes submitted there would just queue behind them (the close is what frees the worker). `to_thread` runs them outside the pool, and concurrently because each teardown is a network call that in series can exceed the shutdown window

AmineDiro: > to_thread runs them outside the pool, and concurrently because each teardown is a network call that in series can exceed the shutdown window


yeah that's exactly my nit, to_thread will spawn a thread for each session, if we have 10k session we'd cause a massive os scheduling issue :/ 

I think we can a todo here or a fixme, because IMO the session should be async objects so we can schedule canceling in the event loop 👍🏼

sergiopaniego: ok makes sense (hope we have to worry about 10k parallel sessions :))! added a `TODO(@openenv)` for the async version. In practice `_live_sessions` is bounded by `max_inflight_tasks` (the `_session_pool` size), so it is a handful of closes, not a thread per rollout. Agreed the clean fix is awaitable sessions cancelled on the event loop.

trl/experimental/async_grpo/openenv_harness.py
```diff
@@ -158,12 +158,18 @@ async def _generate_one(self, prompt, tool_dict, tools, group_id=0):
...
-            for session in list(self._live_sessions):
+
+            def _close(session):
                 try:
                     session.close()
                 except Exception:
                     logger.warning("closing in-flight harness session on stop failed", exc_info=True)
 
+            # Close concurrently: for a remote backend each close() is a network teardown, so closing them in
+            # series can exceed the shutdown window and orphan the rest.
```
AmineDiro:
```suggestion
```

AmineDiro (approved):

PR #5997, opened 2026-06-11
Announce upcoming SFT `loss_type` default change from `'nll'` to `'chunked_nll'`

qgallouedec:
The default `SFTConfig.loss_type` will change from `"nll"` to `"chunked_nll"` in TRL 1.7 (https://github.com/huggingface/trl/pull/5846)

This PR adds a `FutureWarning` (shipping in 1.6) when `loss_type` is left unset. No behavior change yet, the default still resolves to `"nll"`.
The warning tells users no action is needed (they'll get the new default automatically on upgrade) unless they want to pin the current behavior with `loss_type="nll"` (which can be necessary for custom models)

---

> [!NOTE]
> **Low Risk**
> Documentation and deprecation warning only; effective default remains `nll` until TRL 1.7.
> 
> **Overview**
> Prepares for TRL 1.7 by **announcing** that `SFTConfig.loss_type` will default to `"chunked_nll"` instead of `"nll"`.
> 
> `loss_type` now defaults to `None` in the dataclass; `__post_init__` still sets it to `"nll"` when unset, but emits a **`FutureWarning`** explaining the upcoming change and when to pin `loss_type="nll"` (e.g. custom models). Docs and field help text add a **Deprecated** note (1.6.0) and mention the future default.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c874456fe3ef34cc26e2877667789ffd4c87fa4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6571, opened 2026-07-28
Entropy and mean accuracy skip feature in `SFTTrainer`

michaelbenayoun:
# What does this PR do?

Computing the entropy and the mean token accuracy is quite expensive because it requires materializing the logits on device and collective ops. Using `chunked_nll` can solve this but provides poorer performance compared to `nll`. 

This PR allows the user to disable the computation of this metrics to speed things up.

---

> [!NOTE]
> **Low Risk**
> Default remains enabled; changes only optional metric computation and logging on the existing SFT training path.
> 
> **Overview**
> Adds **`compute_token_metrics`** on [`SFTConfig`] (default `True`) so training can skip expensive **`entropy`** and **`mean_token_accuracy`** logging when those metrics are not needed.
> 
> When set to **`False`**, `SFTTrainer` no longer aggregates or logs those metrics: the chunked cross-entropy path skips per-chunk **`argmax`** and entropy math, the Liger path sets **`return_token_accuracy`** from the flag, and post-loss entropy/accuracy blocks in **`compute_loss`** are gated. Loss training behavior stays the same; only the extra vocab-wide metric work is omitted.
> 
> Documentation notes the overhead and the new flag; tests cover **`nll`** and **`chunked_nll`** with **`compute_token_metrics=False`** (loss still logged, no entropy/accuracy in history).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2588a3c5dd931f9ab1e9e6476c46115b7724b5da. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
Thanks @michaelbenayoun 

> Using chunked_nll can solve this but provides poorer performance compared to nll.

Quite curious to know what you mean by this? The general direction for SFT is to drop `nll` in favour of `chunked_nll` (that's why we made it default in https://github.com/huggingface/trl/pull/5997 and https://github.com/huggingface/trl/pull/5997)

michaelbenayoun:
@qgallouedec when I profile both losses on the branch by disabling the token metrics computation, I get the following profiles:

Capture d’écran 2026-07-29 à 16 02 13

Top is `chunked_nll`, bottom is `nll`. 
There are roughly similar but we can see we have a slighly worse hardware utilization in the case of the `chunked_nll`. 
Maybe `chunked_nll` can be improved because I suspect some ops are done on the host right now. 

Now in terms of metrics computation, the difference is much worse.

Capture d’écran 2026-07-29 à 16 13 27

Here we compile two runs:

- Top is `chunked_nll` when we compute the metrics
- Bottom is `chunked_nll` when we do not compute them

We are not running on a distributed setting here so we cannot see how the 3 all gather calls would make things worse but we can already see a huge difference.

Of course it is probably possible to make things a bit better but I do not think they will ever match the case when we do not compute the metrics, hence this PR.

qgallouedec:
ok I see. The chunked loss is expected to have a lower device utilization, because instead of processing the full input at once, we're chunking it into small pieces. But what we win is massive vram reduction. 

I'll need to profile more deeply to see if things can be improved on the chunked side. And put this PR on hold in the meantime. Let me know on Slack if you've any constraint / deadline

qgallouedec:
@michaelbenayoun there were two host syncs in `_chunked_cross_entropy_loss`, which is why chunked was not beating `nll`. I would rather fix that than gate the metrics, so #6842 removes one. The second stays on purpose: it costs ~13 ms/step but enables the masked-chunk skip, worth ~190 ms/step with `completion_only_loss=True`.

For your case #6678 is the bigger win anyway, and it needs no config option.

Closing in favour of #6842 and #6678. Could you run your Trainium profile against #6842? We have no Neuron here.

PR #6587, opened 2026-07-30
fix(vllm): support bracketed IPv6 communicator hosts

### LINKED ISSUE #3164: grpo init_communicator ipv6 can not reach
Hello, when I use GRPO provided by trl for training, I found that when only IPv6 is available, there is the following code in `trl/extras/vllm_client.py` and `trl/scripts/vllm_serve.py`:

```python
pg = StatelessProcessGroup.create(host=self.host, port=self.group_port, rank=self.rank, world_size=world_size)
```

There is a timeout issue, it seems that communication cannot be established through IPv6. When I change the above code to the following, it seems to work, but I am not sure if it is correct:

```python
temp_host = self.host.replace("[", "").replace("]", "")
pg = StatelessProcessGroup.create(host=temp_host, port=self.group_port, rank=self.rank, world_size=world_size)
```

Have you tested it in an IPv6 environment?

tandixit95:
# What does this PR do?

Fixes #3164.

The vLLM client/server path needs two representations of an IPv6 literal:

- HTTP URLs require the bracketed form, such as `[2001:db8::1]`.
- `StatelessProcessGroup.create` and `TCPStore` require the bare address, such as `2001:db8::1`.

This change follows the approach suggested by the maintainer in #3164: it strips surrounding brackets only at the communicator call sites on both the client and server. `self.host` and `base_url` remain unchanged, so URL construction keeps the correct bracketed form.

## Regression coverage

Focused client and server tests cover both bracketed IPv6 and ordinary IPv4 hosts. They verify that:

- communicator creation receives the bare IPv6 address;
- IPv4 behavior is unchanged;
- the client's stored host and HTTP base URL remain bracketed.

```text
uv run --with pytest pytest tests/test_vllm_client_server.py -k 'communicator_host' -q
....                                                                     [100%]
4 passed, 51 deselected
```

No documentation change is needed because this is an internal host-normalization fix with no public API change.

## AI writing disclosure

### DIFF
--- a/tests/test_vllm_client_server.py
+++ b/tests/test_vllm_client_server.py
@@ -14,17 +14,20 @@
 
 import os
 import subprocess
-from types import SimpleNamespace
+import sys
+from types import ModuleType, SimpleNamespace
+from unittest.mock import MagicMock, patch
 
 import pytest
 from packaging.version import Version
 from transformers import AutoModelForCausalLM, AutoProcessor, AutoTokenizer
 from transformers.testing_utils import torch_device
 
+from trl.generation import vllm_client as vllm_client_module
 from trl.generation.vllm_client import VLLMClient
 from trl.generation.vllm_generation import extract_logprobs
 from trl.import_utils import is_vllm_available
-from trl.scripts.vllm_serve import chunk_list
+from trl.scripts.vllm_serve import WeightSyncWorkerExtension, chunk_list
 
 from .testing_utils import (
     TrlTestCase,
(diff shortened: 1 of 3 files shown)


tandixit95:
@qgallouedec this implements the bracket-strip-at-the-communicator-call-site approach you suggested in #3164, while keeping the bracketed HTTP URL unchanged. It includes client and server regression coverage for bracketed IPv6 and ordinary IPv4 hosts. The fork workflows are currently awaiting maintainer approval; when you have a moment, could you please take a look?

Issue #5222, opened 2026-03-05
DPOTrainer ref adapter crashes with PEFT `target_parameters` (MoE models on Transformers 5.x)


pritmish:
### Summary

DPOTrainer's ref adapter creation (`model.add_adapter("ref", ...)`) crashes when the LoRA config uses `target_parameters` — a PEFT feature required for training MoE models on Transformers 5.x, where expert modules are fused `nn.Parameter` tensors instead of individual `nn.Linear` layers.

PEFT currently restricts multiple adapters with `target_parameters` to one per model ([peft#2710](https://github.com/huggingface/peft/pull/2710)), so TRL's attempt to create a second "ref" adapter always fails.

### Environment

- `transformers==5.3.0`
- `trl==0.29.0`
- `peft==0.18.1`
- Model: Qwen3-30B-A3B

### Root Cause Chain

**1. Transformers 5.x changed MoE architecture**

In Transformers 4.x, each MoE expert was a separate `nn.Module` with individual `nn.Linear` layers:
```
model.layers.0.mlp.experts.0.gate_proj  → nn.Linear
model.layers.0.mlp.experts.0.up_proj    → nn.Linear
model.layers.0.mlp.experts.0.down_proj  → nn.Linear
```

In Transformers 5.x, experts are fused into a single module with stacked `nn.Parameter` tensors:
```
model.layers.0.mlp.experts              → Qwen3MoeExperts (single module)
model.layers.0.mlp.experts.gate_up_proj → nn.Parameter, shape [128, 1536, 2048]
model.layers.0.mlp.experts.down_proj    → nn.Parameter, shape [128, 2048, 768]
```

This means PEFT's `target_modules` (which matches `nn.Module` names) can no longer target expert layers. Users must use `target_parameters` instead (introduced in [peft#2498](https://github.com/huggingface/peft/pull/2498)), which targets `nn.Parameter` objects directly.

**2. PEFT restricts multi-adapter with `target_parameters`**

PEFT intentionally blocks creating multiple adapters that use `target_parameters` ([peft#2710](https://github.com/huggingface/peft/pull/2710)) due to unresolved issues with nested `.base_layer`, state_dict corruption, and load-order dependencies.

**3. TRL DPOTrainer creates a "ref" adapter**

DPOTrainer creates a frozen copy of the "default" adapter for computing reference logprobs:

```python
# trl/trainer/dpo_trainer.py
if is_peft_available() and is_peft_model(model) and ref_model is None:
[...]

PR #6591, opened 2026-07-30
Use a reference adapter with `target_parameters` on `peft>=0.20.0`

albertvillanova:
This PR lets GRPO, DPO, KTO and RLOO create the frozen `"ref"` adapter when the pretrained LoRA adapter uses `target_parameters` and PEFT is recent enough to support it, instead of always falling back to the base model as reference.

Follow-up to:
- #5222

### Motivation

When the model passed to the trainer is already wrapped with a LoRA adapter that uses `target_parameters` (fused-MoE experts), the trainers skip the `"ref"` adapter and compute the reference log probs with adapters disabled, i.e. from the base model. That fallback exists because PEFT only supported a single adapter per model in that case (peft#3340).

For a pretrained, non-freshly-initialized adapter, the base model is the wrong reference, and the trainer only warns about it, so the run silently trains against the wrong reference.

PEFT 0.20.0 lifts the restriction (peft#3350), so the fallback is no longer needed there. The fix only allows adapters targeting the same parameter set, which is satisfied here since the `"ref"` adapter reuses the `"default"` config.

### Solution

The guard becomes version-dependent instead of unconditional: below PEFT 0.20.0 the existing warn-and-fall-back behavior is kept, from 0.20.0 on the `"ref"` adapter is created and initialized as a copy of the `"default"` adapter, as it already is for adapters that don't use `target_parameters`.

The minimum PEFT version is intentionally left untouched. This repository already gates PEFT features at runtime rather than raising the floor, and raising it to a very recent release for this case would affect every PEFT user. The gate can be dropped once the minimum version reaches 0.20.0 for other reasons.

### Changes

- Gate the `target_parameters` reference-adapter fallback on `peft<0.20.0` in the GRPO, DPO, KTO and RLOO trainers, and point the warning at the upgrade.
- Update the corresponding comments to refer to peft#3350 and to record why its same-parameters restriction is satisfied.
- Make `test_train_moe_peft_model` assert the presence or absence of the `"ref"` adapter depending on the installed PEFT version, and exclude the frozen `"ref"` parameters from the parameter-change checks.

---

> [!NOTE]
> **Medium Risk**
> Changes which reference distribution is used for preference/RL training when MoE LoRA uses target_parameters on newer PEFT, which can alter training behavior versus the previous silent base-model fallback.
> 
> **Overview**
> On **PEFT ≥ 0.20.0**, DPO, GRPO, KTO, and RLOO now add a frozen **`ref`** adapter (copy of **`default`**) when the pretrained LoRA uses **`target_parameters`**, instead of always using the base model for reference log-probs.
> 
> The old skip-and-warn path is kept only for **`peft<0.20.0`**, with warnings updated to recommend upgrading. **`test_train_moe_peft_model`** in the four trainer test files asserts **`ref`** presence by PEFT version and skips frozen **`ref`** params in change checks; RLOO tests gain a **`peft`** import for version checks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d57ae8ae20b3d84311ddc25cf2474cc7f2eb182c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_dpo_trainer.py
+++ b/tests/test_dpo_trainer.py
@@ -783,9 +783,9 @@ class TestDPOTrainer(TrlTestCase):
 
     @require_peft
     def test_train_moe_peft_model(self):
-        # Regression test for https://github.com/huggingface/trl/issues/5222. PEFT only supports one adapter per model
-        # when the LoRA config uses `target_parameters` (see peft#3340), so no "ref" adapter can be created and the
-        # reference log probs are computed with adapters disabled instead.
+        # Regression test for https://github.com/huggingface/trl/issues/5222. Before PEFT 0.20.0, only one adapter per
+        # model was supported when the LoRA config uses `target_parameters` (see peft#3340, fixed in peft#3350), so no
+        # "ref" adapter could be created and the reference log probs were computed with adapters disabled instead.
         model_id = "trl-internal-testing/tiny-GptOssForCausalLM"
         model = AutoModelForCausalLM.from_pretrained(model_id, dtype="float32")
         base_param_names = [f"base_model.model.{n}" for n, _ in model.named_parameters()]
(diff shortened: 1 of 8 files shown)


qgallouedec (approved):
perfect!

PR #6594, opened 2026-07-30
fix(grpo): use per-token IS ratio for KL bias correction regardless of importance_sampling_level

### LINKED ISSUE #6586: KL bias correction with importance_sampling_level="sequence" broadcasts a sequence-level ratio onto per-token KL
### Description

Follow-up to #6503, splitting out a caveat @albertvillanova raised in review there.

`use_bias_correction_kl` multiplies the per-token KL by `coef_1`, but `coef_1`'s meaning depends on `importance_sampling_level`:

```python
# trl/trainer/grpo_trainer.py
log_ratio = per_token_logps - old_per_token_logps
if self.importance_sampling_level == "token":
    log_importance_weights = log_ratio                                    # (B, T)
elif self.importance_sampling_level == "sequence":
    log_importance_weights = (log_ratio * mask).sum(-1) / mask.sum(-1).clamp(min=1.0)
    log_importance_weights = log_importance_weights.unsqueeze(-1)         # (B, 1)

coef_1 = torch.exp(log_importance_weights)
...
if self.args.use_bias_correction_kl:
    per_token_kl = per_token_kl * coef_1
```

With `importance_sampling_level="token"` this is the per-token correction from the [DeepSeek-V3.2 paper](https://huggingface.co/papers/2512.02556), and the KL gradient is the unbiased reverse-KL gradient.

With `"sequence"`, `coef_1` is the sequence-mean ratio of shape `(B, 1)`, so a single sequence-level weight is broadcast across the per-token KL. The gradient then picks up an extra `(Σ_t k3_t) · ∇(mean_t log π_t)` term, so it is neither the paper's per-token correction nor the unbiased reverse-KL gradient in that configuration.

This is pre-existing behaviour and not a crash. It matters a bit more now that #6503 made the correction the default, so a user combining it with `importance_sampling_level="sequence"` (for example the GSPO setup) silently gets a different KL gradient than the docstring describes. The docstring added in #6503 currently qualifies the claim to `"token"`, which is why this is a follow-up rather than part of that PR.

### Possible directions

1. Use the per-token ratio for the KL term regardless of `importance_sampling_level`, so the correction always matches the paper.
2. Keep the current behaviour and document it in more detail as a known limitation of the sequence level.
3. Warn when `use_bias_correction_kl=True` is combined with `importance_sampling_level="sequence"`.

I don't have a strong view on which is right, since option 1 changes behaviour for existing sequence-level runs. Happy to open a PR once you decide the direction.

priyanka25aug:
## What does this PR do?

Fixes a gradient-correctness bug in GRPO's KL bias correction when `importance_sampling_level="sequence"`.

Closes #6586

### Bug

When `use_bias_correction_kl=True` (the default since PR #6503), the KL correction in `_compute_loss` multiplies per-token KL by `coef_1`:

```python
if self.args.use_bias_correction_kl:
    per_token_kl = per_token_kl * coef_1
```

`coef_1 = torch.exp(log_importance_weights)`, and when `importance_sampling_level="sequence"`, `log_importance_weights` is the **sequence-mean** log-ratio of shape `(B, 1)`:

```python
log_importance_weights = (log_ratio * mask).sum(-1) / mask.sum(-1).clamp(min=1.0)
log_importance_weights = log_importance_weights.unsqueeze(-1)  # (B, 1)
```

Broadcasting a `(B, 1)` weight onto per-token KL introduces a spurious gradient term. The per-token gradient of the total KL becomes:

```
∇_{θ} [exp(mean_t log_ratio_t) · Σ_t k3_t]
= exp(mean_t log r_t) · ∇_{θ}(Σ_t k3_t)
  + (Σ_t k3_t) · ∇_{θ}[exp(mean_t log r_t)]   ← spurious term
```

This second term — `(Σ_t k3_t) · ∇(mean_t log π_t)` — is neither the DeepSeek-V3.2 paper's per-token correction nor the unbiased reverse-KL gradient.

For `level="token"`, `coef_1` is already per-token so there is no bug in that case.

### Fix

Always multiply per-token KL by the **per-token** IS ratio `torch.exp(log_ratio)` (shape `(B, T)`), regardless of `importance_sampling_level`:

```python
# Before (buggy when importance_sampling_level="sequence"):
if self.args.use_bias_correction_kl:
    per_token_kl = per_token_kl * coef_1

# After (correct in all cases):
if self.args.use_bias_correction_kl:
    per_token_kl = per_token_kl * torch.exp(log_ratio)
```

`log_ratio = per_token_logps - old_per_token_logps` is always per-token (shape `(B, T)`), so this fix is correct regardless of `importance_sampling_level`. For `level="token"` the behavior is **identical** to before (both expressions equal `exp(log_ratio)`); only `level="sequence"` is affected.

Also updates the `use_bias_correction_kl` docstring in `GRPOConfig` to document this invariant, and updates the default to `True` to match the merged PR #6503.

### References

- DeepSeek-V3.2 paper (§ training details): https://huggingface.co/papers/2512.02556
- Issue #6586 (this bug)
- PR #6503 (which changed the default to `True`, making this bug active by default)

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md)?

> [!NOTE]
> This contribution was developed with AI assistance (Claude). The mathematical analysis, gradient derivation, and fix were verified against the DeepSeek-V3.2 paper and TRL source.

cc @qgallouedec @albertvillanova

---

> [!NOTE]
> **Medium Risk**
> Changes training gradients for sequence-level IS with `beta != 0` and default `use_bias_correction_kl=True`, but aligns KL updates with DeepSeek-V3.2 and fixes a correctness bug rather than adding new behavior paths.
> 
> **Overview**
> Fixes **incorrect KL bias correction** in `GRPOTrainer._compute_loss` when `use_bias_correction_kl=True` and `importance_sampling_level="sequence"`. The KL term was scaled by `coef_1` (the same importance weights as the policy loss), which at sequence level is a `(B, 1)` sequence-mean ratio broadcast onto per-token KL and adds a spurious gradient term.
> 
> KL correction now always multiplies per-token KL by **`torch.exp(log_ratio)`** (per-token π_θ/π_old), independent of `importance_sampling_level`. Policy-loss IS still follows `importance_sampling_level`; behavior at `"token"` is unchanged.
> 
> **`GRPOConfig`** docstrings for `use_bias_correction_kl` are updated to state that KL correction always uses the per-token IS ratio.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6060a865ef3e4ba4193dfb16b732e66e9d17890e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -357,12 +357,13 @@ class GRPOConfig(_BaseConfig):
             parameter corresponds to the `delta` threshold in Equation 9 of the [DeepSeek-V3.2
             paper](https://huggingface.co/papers/2512.02556). It expects a positive value (e.g., 0.5).
         use_bias_correction_kl (`bool`, *optional*, defaults to `True`):
-            Whether to multiply the KL term by the importance sampling ratio, so that the KL gradient becomes the
-            unbiased reverse-KL gradient, as described in the
-            [DeepSeek-V3.2 paper](https://huggingface.co/papers/2512.02556). This changes the KL gradient whenever
-            `beta != 0`, including on-policy: the ratio is differentiable, so it affects the gradient even where its
-            value is exactly 1. The unbiased reverse-KL property holds for `importance_sampling_level="token"`; with
-            `"sequence"` a sequence-level weight is broadcast onto the per-token KL.
+            Whether to multiply the KL term by the per-token importance sampling ratio
+            ``π_θ(y_t|x) / π_old(y_t|x)`` at each position, so that the KL gradient becomes the unbiased
+            reverse-KL gradient, as described in the
+            [DeepSeek-V3.2 paper](https://huggingface.co/papers/2512.02556). This always uses the per-token IS
+            ratio regardless of ``importance_sampling_level`` (which governs only the policy-loss IS ratio).
+            This changes the KL gradient whenever `beta != 0`, including on-policy: the ratio is differentiable,
+            so it affects the gradient even where its value is exactly 1.
 
         > Parameters that control the logging
 
@@ -965,12 +966,13 @@ class GRPOConfig(_BaseConfig):
     use_bias_correction_kl: bool = field(
         default=True,
         metadata={
-            "help": "Whether to multiply the KL term by the importance sampling ratio, so that the KL gradient "
-            "becomes the unbiased reverse-KL gradient, as described in the [DeepSeek-V3.2 "
-            "paper](https://huggingface.co/papers/2512.02556). This changes the KL gradient whenever `beta != 0`, "
-            "including on-policy: the ratio is differentiable, so it affects the gradient even where its value is "
-            "exactly 1. The unbiased reverse-KL property holds for `importance_sampling_level='token'`; with "
-            "'sequence' a sequence-level weight is broadcast onto the per-token KL."
+            "help": "Whether to multiply the KL term by the per-token importance sampling ratio "
+            "(pi_theta(y_t|x) / pi_old(y_t|x)) at each position, so that the KL gradient becomes the "
+            "unbiased reverse-KL gradient, as described in the DeepSeek-V3.2 paper "
+            "(https://huggingface.co/papers/2512.02556). This always uses the per-token IS ratio regardless "
+            "of `importance_sampling_level` (which governs only the policy-loss IS ratio). "
+            "This changes the KL gradient whenever `beta != 0`, including on-policy: the ratio is "
+            "differentiable, so it affects the gradient even where its value is exactly 1."
         },
     )
 
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -3087,9 +3087,15 @@ class GRPOTrainer(_BaseTrainer):
             per_token_kl = (
                 torch.exp(ref_per_token_logps - per_token_logps) - (ref_per_token_logps - per_token_logps) - 1
             )
-            # Importance sampling correction for the KL divergence
+            # Importance sampling correction for the KL divergence.
+            # Always use the per-token IS ratio (exp(log_ratio), shape (B, T)), regardless of
+            # `importance_sampling_level`. When level="sequence", `coef_1` is exp(mean(log_ratio))
+            # of shape (B, 1), which broadcasts a sequence-mean weight onto per-token KL and
+            # introduces a spurious term `(Σ_t k3_t)·∇(mean_t log π_t)` in the gradient —
+            # neither the paper's per-token correction nor the unbiased reverse-KL gradient.
+            # Per-token correction here matches the DeepSeek-V3.2 paper exactly (see #6586).
             if self.args.use_bias_correction_kl:
-                per_token_kl = per_token_kl * coef_1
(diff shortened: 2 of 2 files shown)


trl/trainer/grpo_config.py
```diff
@@ -357,12 +357,13 @@ class GRPOConfig(_BaseConfig):
...
-            `beta != 0`, including on-policy: the ratio is differentiable, so it affects the gradient even where its
-            value is exactly 1. The unbiased reverse-KL property holds for `importance_sampling_level="token"`; with
-            `"sequence"` a sequence-level weight is broadcast onto the per-token KL.
+            Whether to multiply the KL term by the per-token importance sampling ratio
+            ``π_θ(y_t|x) / π_old(y_t|x)`` at each position, so that the KL gradient becomes the unbiased
+            reverse-KL gradient, as described in the
+            [DeepSeek-V3.2 paper](https://huggingface.co/papers/2512.02556). This always uses the per-token IS
+            ratio regardless of ``importance_sampling_level`` (which governs only the policy-loss IS ratio).
+            This changes the KL gradient whenever `beta != 0`, including on-policy: the ratio is differentiable,
+            so it affects the gradient even where its value is exactly 1.
```
LeonEricsson, line 366:
```suggestion
            Whether to multiply the KL term by the per-token (regardless of `importance_sampling_level`) importance sampling ratio (regardless  so that the KL gradient becomes the unbiased
            reverse-KL gradient, as described in the
            [DeepSeek-V3.2 paper](https://huggingface.co/papers/2512.02556).
            This changes the KL gradient whenever `beta != 0`, including on-policy: the ratio is differentiable,
            so it affects the gradient even where its value is exactly 1.
```

remember to change metadata help as well

trl/trainer/grpo_trainer.py
```diff
@@ -3087,9 +3087,15 @@ def _compute_loss(self, model, inputs):
...
                 torch.exp(ref_per_token_logps - per_token_logps) - (ref_per_token_logps - per_token_logps) - 1
             )
-            # Importance sampling correction for the KL divergence
+            # Importance sampling correction for the KL divergence.
+            # Always use the per-token IS ratio (exp(log_ratio), shape (B, T)), regardless of
+            # `importance_sampling_level`. When level="sequence", `coef_1` is exp(mean(log_ratio))
+            # of shape (B, 1), which broadcasts a sequence-mean weight onto per-token KL and
+            # introduces a spurious term `(Σ_t k3_t)·∇(mean_t log π_t)` in the gradient —
+            # neither the paper's per-token correction nor the unbiased reverse-KL gradient.
+            # Per-token correction here matches the DeepSeek-V3.2 paper exactly (see #6586).
```
LeonEricsson, line 3096:
```suggestion
            # Importance sampling correction for the KL divergence.
```

we're calculating token level kl,  so a token level importance weight is naturally warranted. doesn't need further motivation imo.

LeonEricsson (commented):
minor nits.

i believe this to be the correct approach in theory. @albertvillanova thoughts?

PR #6590, opened 2026-07-30
Remove bitsandbytes `_check_is_size` warning filter

yashb98:
# What does this PR do?

Removes the `_check_is_size` `filterwarnings` entry from `pyproject.toml`, which is the action item in #6447.

Fixes #6447

The entry was added because bitsandbytes called the deprecated `torch._check_is_size` in its CUDA quant ops. The upstream fix (bitsandbytes-foundation/bitsandbytes#1940) was merged but unreleased at the time, so the issue asked to remove the filter once a release containing it reached CI. `bitsandbytes 0.50.0` is that release, and CI is already on it (#6546).

## Verification

Ran the three tests named in the issue with the filter removed, on an aarch64 GB10, torch 2.13.0+cu130, capturing all warnings:

| bitsandbytes | result | `_check_is_size` warnings |
|---|---|---|
| 0.50.0 | 3 passed | **0** |
| 0.49.2 | 3 passed | 14, from `bitsandbytes/backends/cuda/ops.py:213` |

So the filter is genuinely dead on 0.50.0, and the 0.49.2 row confirms the check is meaningful rather than vacuous: it does still catch the warning on the older release. Independently, `_check_is_size` appears 0 times anywhere in the installed 0.50.0 package and 6 times in 0.49.2.

Worth noting that `filterwarnings` does not start with `error`, so this cannot fail CI on an older bitsandbytes either. It would only restore log noise.

## Before submitting

- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request), Pull Request section?
- [x] Was this discussed/approved via a GitHub issue? Please add a link to it if that's the case. Yes, #6447, which states this removal as its action item.
- [ ] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests? No new tests, the three existing `test_peft_with_quantization` tests already cover it; results above.

## AI writing disclosure

- [ ] No AI usage: the PR was written entirely by a human.
- [x] AI-assisted: some parts were suggested or improved by AI, but the PR was written and reviewed by a human.
- [ ] AI-generated: the PR was mostly or fully generated by an AI tool.

## Who can review?

Anyone once CI is green.

---

> [!NOTE]
> **Low Risk**
> Test-configuration-only change with no runtime or dependency logic; worst case on an older bitsandbytes is extra warning noise in logs, not CI failure.
> 
[...]

Issue #6555, opened 2026-07-27
Remove kernels-community/mamba-ssm autocast warning filter once the Hub kernel build stops calling `torch.get_autocast_gpu_dtype()`


albertvillanova:
### What

`tests_dev` (CI job installing `accelerate`/`transformers`/`datasets`/`peft` from git main) emits repeated `DeprecationWarning`s from the NemotronH (Nemotron 3) parametrized case:

```
tests/test_dpo_trainer.py: 18 warnings
tests/test_sft_trainer.py: 18 warnings
  /github/home/.cache/huggingface/hub/kernels--kernels-community--mamba-ssm/snapshots/.../build/torch213-cxx11-cu130-x86_64-linux/ops/triton/ssd_combined.py:1511: DeprecationWarning: torch.get_autocast_gpu_dtype() is deprecated. Please use torch.get_autocast_dtype('cuda') instead. (Triggered internally at /__w/pytorch/pytorch/torch/csrc/autograd/init.cpp:1007.)
    dtype = torch.get_autocast_gpu_dtype()
```

Affected tests: `tests/test_dpo_trainer.py::TestDPOTrainer::test_train[trl-internal-testing/tiny-NemotronHForCausalLM-nano]` and `tests/test_sft_trainer.py::TestSFTTrainer::test_train[trl-internal-testing/tiny-NemotronHForCausalLM-nano]` (see run https://github.com/huggingface/trl/actions/runs/30240017065/job/89895088560).

### Root cause

The warning originates in the Hub-hosted kernel `kernels-community/mamba-ssm` — an auto-fetched build of `state-spaces/mamba`'s `mamba_ssm/ops/triton/ssd_combined.py`, used by transformers' NemotronH Mamba2 mixer fast path (`transformers.models.nemotron_h.modeling_nemotron_h`, via `lazy_load_kernel("mamba-ssm")`). That kernel still calls the deprecated `torch.get_autocast_gpu_dtype()` instead of `torch.get_autocast_dtype('cuda')`. This is not TRL code, not transformers code, not even the `kernels` integration glue — it's inside the kernel build pulled from the HF Hub.

[...]

Issue #6466, opened 2026-07-20
Remove triton `AnnAssign` warning filter once no supported torch resolves a triton without the fix (>= 3.8.0)


albertvillanova:
### What

Many trainer test modules emit a `DeprecationWarning` in CI on Python 3.14 during Triton kernel compilation:

```
triton/compiler/code_generator.py:629: DeprecationWarning: AnnAssign.__init__ missing 1 required positional argument: 'simple'. This will become an error in Python 3.15.
  init_node = ast.AnnAssign(target=st_target, value=default_value, annotation=annotation)
```

It shows up across `tests/test_grpo_trainer.py`, `tests/test_kto_trainer.py`, `tests/test_dpo_trainer.py`, `tests/test_rloo_trainer.py`, `tests/test_sft_trainer.py`, and `tests/test_utils.py`.

### Root cause

The warning originates in Triton, not TRL. In `visit_FunctionDef`, Triton builds an `ast.AnnAssign` node without the required `simple` field when a `@triton.jit` kernel argument has both an annotation and a default value. TRL only triggers it indirectly, by compiling Triton kernels during tests; there is nothing actionable on the TRL side.

### Upstream status

Reported upstream: https://github.com/triton-lang/triton/issues/10981

**Not fixed yet** — Triton `main` (3.8.0 dev, latest release v3.7.1) still constructs the node without `simple`. Note this is currently only a `DeprecationWarning`, but it **will become a hard error on Python 3.15**, which the warning filter will not suppress.

### Workaround in place

We suppress the warning via `filterwarnings` in `pyproject.toml`:

```toml
"ignore:AnnAssign.__init__ missing 1 required positional argument:DeprecationWarning",
```

### Action item

- [ ] Remove the `filterwarnings` entry once every supported `torch` resolves a triton containing #10986, confirmed by a warning-free Python 3.14 CI run.
- [ ] **Separate, harder gate: do not conflate with the item above:** before adding Python 3.15 to the CI matrix, triton must actually be upgraded past #10986. On 3.15 the `DeprecationWarning` becomes a hard error, which no `filterwarnings` entry can suppress.

The second item is the reason this issue should not be closed by a filter-removal PR alone: the filter and the 3.15 blocker have independent triggers.

Issue #6447, opened 2026-07-19
Remove bitsandbytes `_check_is_size` warning filter once the minimum required bitsandbytes is >= 0.50.0


albertvillanova:
### What

`test_peft_with_quantization` in the DPO, KTO, and SFT trainer tests emits a benign `FutureWarning` in CI when loading a bitsandbytes-quantized model:

```
bitsandbytes/backends/cuda/ops.py:213: FutureWarning: _check_is_size will be removed in a future PyTorch release along with guard_size_oblivious. Use _check(i >= 0) instead.
  torch._check_is_size(blocksize)
```

Affected tests:

- `tests/test_dpo_trainer.py::TestDPOTrainer::test_peft_with_quantization`
- `tests/test_kto_trainer.py::TestKTOTrainer::test_peft_with_quantization`
- `tests/test_sft_trainer.py::TestSFTTrainer::test_peft_with_quantization`

### Root cause

The warning originates in `bitsandbytes`, not TRL. Its CUDA quantization ops call the deprecated `torch._check_is_size(blocksize)`. PyTorch has deprecated `torch._check_is_size` (along with `guard_size_oblivious`) in favor of `torch._check(i >= 0)`, and emits a `FutureWarning`. TRL only triggers it indirectly by loading a 4-bit/8-bit quantized model in these PEFT + quantization tests; there is nothing actionable on the TRL side.

### Upstream status

Already fixed upstream by replacing the deprecated call: https://github.com/bitsandbytes-foundation/bitsandbytes/pull/1940 (merged 2026-05-07).

The fix is **merged but unreleased** — the latest numbered release is `0.49.2` (which predates the fix); it is currently only in the `continuous-release_main` rolling build. TRL requires `bitsandbytes` with no version pin.

### Workaround in place

We suppress the warning via `filterwarnings` in `pyproject.toml`:

```toml
"ignore:_check_is_size will be removed in a future PyTorch release:FutureWarning",
```

### Action item

The fix is released in `bitsandbytes` 0.50.0. Since `pyproject.toml` declares `bitsandbytes` with no lower bound, older releases remain supported configurations and can still emit the warning, so the filter must stay.

Remove the `filterwarnings` entry once TRL's minimum required `bitsandbytes` is `>= 0.50.0`. Do not bump the floor solely to drop this filter: wait for an independent reason to require 0.50.0.

PR #6595, opened 2026-07-31
Document removal criteria for pytest warning filters

albertvillanova:
This PR documents, next to every pytest `filterwarnings` entry in `pyproject.toml`, the condition under which that entry can be removed, and refreshes the upstream status notes that had become stale.

### Motivation

The tracking issue of each filter states its action item as removing the entry "once upstream ships the fix". For a dependency we require without a floor, that criterion is wrong: an upstream release does not stop older releases from being supported configurations, and `filterwarnings` is part of the test configuration we declare for all of them, not a snapshot of what CI happens to resolve today. Removing an entry when the fix is merely released brings the warning noise back for every supported older version, including for contributors running the suite in an existing environment.

The ambiguity is not theoretical:

- #6590 proposed dropping the `_check_is_size` filter as soon as `bitsandbytes` 0.50.0 reached CI, faithfully implementing the criterion as worded in #6447.
- We still carry the `torch.jit.script` filters although their `deepspeed` fix has been released since 0.18.6, precisely because our floor has not moved yet.

So both readings currently coexist in the same file, with nothing on the entries themselves to tell them apart. Recording the criterion at each entry puts that information where the removal decision is actually taken, rather than in an issue title.

### Solution

Add one `Remove once:` line per entry. The criterion takes one of two shapes, depending on whether TRL can constrain the component that emits the warning:

- for dependencies we declare (`bitsandbytes`, `liger-kernel`, `deepspeed`), removal is gated on our minimum required version reaching the first release that contains the fix;
- for dependencies we do not declare, either transitive (`triton`, pinned by `torch`) or resolved at runtime (the `kernels-community/mamba-ssm` Hub kernel build), there is no floor to bump, so removal is gated on no supported configuration emitting the warning any more.

The criterion lines deliberately do not restate the version we currently require, which would be one more thing to keep in sync with the requirements.

Reviewing every entry also surfaced three upstream status notes that are no longer accurate, updated here. The wording of the tracking issues (#6435, #6447, #6466, #6555) is aligned separately.

### Changes

- Add a `Remove once:` line to every `filterwarnings` entry, giving the target version for the entries gated on a requirement bump and the observable condition for the others
- Update the `liger-kernel` SAPO note: the fix is released in 0.8.1, which we exclude for an unrelated crash (GH-6517), so the criterion is 0.8.2
- Update the `bitsandbytes` `_check_is_size` note: the fix is released in 0.50.0
- Add the upstream fix PR to the `triton` `AnnAssign` entry, merged and unreleased as of 3.7.1

---

> [!NOTE]
> **Low Risk**
> Comment-only changes to test configuration documentation; no runtime or dependency requirement changes.
> 
> **Overview**
> Adds a **`Remove once:`** comment beside each pytest `filterwarnings` entry in `pyproject.toml`, so maintainers know when to drop a filter without relying on tracking-issue wording alone.
> 
> Criteria differ by dependency: filters tied to declared extras (e.g. **deepspeed**, **liger-kernel**, **bitsandbytes**) say removal waits until TRL’s minimum required version includes the upstream fix; others (sentencepiece, torch, triton, Hub mamba kernel) say removal waits until no supported configuration still emits the warning.
> 
> Also refreshes stale upstream notes: **liger-kernel** SAPO fix is in 0.8.1 with removal gated on **>= 0.8.2** (0.8.1 excluded per GH-6517), **bitsandbytes** `_check_is_size` fix is in **0.50.0**, and **triton** `AnnAssign` documents merged fix PR **#10986** (unreleased as of 3.7.1).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83fcf571cc437878d03a770115a2522b78186bab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -186,6 +186,7 @@ norecursedirs = [
 filterwarnings = [
     # SWIG deprecations from SWIG-generated C/C++ extensions: sentencepiece
     # Upstream issue: https://github.com/google/sentencepiece/issues/1150
+    # Remove once: no supported sentencepiece version emits it
     "ignore:builtin type SwigPyPacked has no __module__ attribute:DeprecationWarning",
     "ignore:builtin type SwigPyObject has no __module__ attribute:DeprecationWarning",
     "ignore:builtin type swigvarlink has no __module__ attribute:DeprecationWarning",
@@ -194,6 +195,7 @@ filterwarnings = [
     # Upstream issue: https://github.com/deepspeedai/DeepSpeed/issues/7835
     # Upstream PR: https://github.com/deepspeedai/DeepSpeed/pull/7840
     # Upstream fix released in deepspeed v0.18.6: https://github.com/deepspeedai/DeepSpeed/releases/tag/v0.18.6
+    # Remove once: deepspeed >= 0.18.6 is required
     "ignore:`torch.jit.script_method` is deprecated:DeprecationWarning",
     "ignore:`torch.jit.script` is deprecated:DeprecationWarning",
     # On Python 3.14+ the same deprecation is reworded to "is not supported in Python 3.14+ and may break"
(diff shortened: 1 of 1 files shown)


qgallouedec (approved):
useful, thanks

PR #6597, opened 2026-07-31
Require jmespath only on transformers below 5.13.0

albertvillanova:
This PR fixes `GRPOTrainer` refusing to train with tools or environments when `jmespath` is not installed, even on transformers versions that never use it.

Supersedes #6240.

### Motivation

`jmespath` is only used by the legacy `response_schema` parser, which is all that transformers below 5.13.0 ships. From transformers 5.13.0 on, TRL sets a `response_template` and the new-style parser is used instead, so `jmespath` is never imported (transformers removed its own usage in huggingface/transformers#47320).

The guard was unconditional, so it blocked tool and environment training on current transformers for a library that plays no part in it. This broke every `openenv` example that passes `environment_factory`, since none of them installs `jmespath`, as reported in #6240.

### Solution

Fire the guard only on the legacy parsing path, reusing the `_SUPPORTS_RESPONSE_TEMPLATE` check that already decides which parser is used, and offer upgrading transformers as an alternative to installing `jmespath`. The examples need no dependency changes: a plain `uv run` resolves the latest transformers, where `jmespath` is not needed, so all of them work as is.

### Changes

- Require `jmespath` in `GRPOTrainer` only when the installed transformers ships the legacy response parser
- Reword the error to name the affected transformers versions and to offer upgrading as an alternative
- Add a test asserting that tools work without `jmespath` on transformers 5.13.0 and higher
- Rename `require_jmespath` to `require_response_parsing` and align its condition with the trainer, so tool and parsing tests are no longer skipped on newer transformers when `jmespath` is missing

---

> [!NOTE]
> **Low Risk**
> Narrower dependency guard aligned with existing `_SUPPORTS_RESPONSE_TEMPLATE` behavior; no change to training logic on supported transformers versions.
> 
> **Overview**
> **`GRPOTrainer` no longer blocks tool or environment training when `jmespath` is missing on transformers ≥ 5.13**, where response parsing uses the new `response_template` path instead of the legacy `response_schema` parser that depends on `jmespath`.
> 
> The import guard now mirrors `_SUPPORTS_RESPONSE_TEMPLATE`: `jmespath` is required only on older transformers, and the error message suggests upgrading transformers as an alternative to installing `jmespath`.
> 
> Test helpers rename `require_jmespath` to **`require_response_parsing`** with the same version-aware skip logic so parsing and tool tests still run on newer transformers without `jmespath`. A new **`test_tools_without_jmespath`** asserts `GRPOTrainer` constructs with tools when `is_jmespath_available` is patched to `False` on transformers ≥ 5.13.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1aac5cef4c4049368c8cec6142f9112c0ce33231. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:
The "Tests without optional dependencies" failures here are not caused by this PR's logic, but they are unmasked by it. They need #6602 to be merged first.

The 37 failures are all in `tests/test_chat_template_utils.py`, and all on the `*ForConditionalGeneration` parametrizations of `TestParseResponse` / `TestAddResponseSchema`:

```
ImportError: Qwen2VLImageProcessor requires the PIL library but it was not found in your environment.
             Qwen2VLImageProcessor requires the Torchvision library ...
```

Those parametrizations load an `AutoProcessor`, so they need the `vlm` extra, but they carry no `require_vision` mark. Until now that never mattered: both classes were decorated with `require_jmespath`, and this job installs only `.[test]`, so no `jmespath` meant the whole class was skipped, incidentally hiding the missing vision guard.

Once `require_response_parsing` correctly stops skipping on transformers 5.13.0 and higher (this job runs 5.14.1), those parametrizations run for the first time in an environment without `Pillow` and `torchvision`, and fail.

So the defect is a pre-existing missing `require_vision`, reproducible on `main` in any environment that has `jmespath` but not the `vlm` extra. It is independent of this PR, hence the separate #6602, which adds the mark and aligns these parametrizations with all the other VLM ones in the same file. Note that `test_train_with_tools_multimodal_response` in `test_grpo_trainer.py` made the same marker switch here and did not fail, precisely because its class already carries `@require_vision`.

I will rebase this PR once #6602 is in.

PR #6600, opened 2026-07-31
Add Pi coding-agent loop-owning GRPO examples

sergiopaniego:
## What does this PR do?

Adds a loop-owning GRPO training example for the real Pi coding agent, mirroring the opencode example. Two scripts under `examples/scripts/openenv/`:

- `pi.py`: local subprocess sandbox (one node, no tunnel)
- `pi_hf_sandbox.py`: one remote HF sandbox per rollout

Pi speaks OpenAI natively, so the in-sandbox interception proxy captures each turn's token ids and logprobs directly (no shim). `AsyncGRPOTrainer` + `HarnessRolloutWorker` (`harness_adapter=None`) train on the trace, with a held-out DeepCoder verifier as the reward.

Depends on the `pi_env` environment (huggingface/OpenEnv#999). Draft until that merges, then the uv headers point at the released package.

## Before submitting

- [x] This PR improves the examples.

## AI writing disclosure

- [x] AI-assisted

### DIFF
--- a/examples/scripts/openenv/pi.py
+++ b/examples/scripts/openenv/pi.py
@@ -1,0 +1,634 @@
+# Copyright 2020-2026 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


sergiopaniego:
Parking this as draft. The per-agent env approach here is superseded by the Harbor integration (huggingface/OpenEnv#1036), which covers these harnesses generically. Keeping it open rather than closing until Harbor actually lands, then it can be closed.

sergiopaniego:
Closing in favor of the generic harness infrastructure coming with Harbor (huggingface/OpenEnv#1036), which supersedes per-harness example wiring. Happy to revisit as a Harbor-based example.

PR #6609, opened 2026-08-01
[DistillationTrainer refactor] Constructor cleanup (40b): drop eval-metric params, require `model`

qgallouedec:
*Item 40b (Group F) — split from item 40, stacked on 39b. Part of #6449. API change.*

The signature-changing half of item 40, coupled to `ServerDistillationTrainer` (its ctor forwards these params to the base `super().__init__`):

- **Drop `compute_metrics` / `preprocess_logits_for_metrics`** ctor params (GRPO parity — GRPO's ctor exposes neither). Distillation has no eval-metric path that consumes them; they were accepted and forwarded but never used.
- **Make `model` required** (`model: … | None = None` → `model: PreTrainedModel | nn.Module | str`, GRPO parity). Item 25's `inspect.signature(model.forward)` already assumes a non-`None` model, so `model=None` produced an obscure `'NoneType' has no attribute 'forward'` instead of Trainer's clear "requires either a model or a model_init". The now-dead `if model is not None else None` guard on `model_name_or_path` is dropped.
- **`ServerDistillationTrainer.__init__`** updated in the same PR: drop the two forwarded params so the delegation to the base `super().__init__` stays valid. Its own `model=None` default is left as-is (removed post-promotion at item 55).
- Dropped the now-unused `Callable` / `EvalPrediction` imports.

Verified: ruff clean; full distillation + server suites 52 green.

---

> [!NOTE]
> **Medium Risk**
> Breaking constructor API for anyone passing the removed kwargs or omitting `model`; behavior change is intentional and low impact because the dropped parameters were unused.
> 
> **Overview**
> **Constructor API cleanup** for `DistillationTrainer` and `ServerDistillationTrainer`, aligned with GRPO.
> 
> `compute_metrics` and `preprocess_logits_for_metrics` are removed from both constructors and are no longer forwarded to the base `Trainer`. Distillation never used them for eval metrics.
> 
> `model` is now **required** (`PreTrainedModel | nn.Module | str` with no `None` default). Passing `model=None` used to fail late on `inspect.signature(model.forward)`; callers must supply a model or path string like other TRL trainers.
> 
> `ServerDistillationTrainer` drops the same two kwargs so its `super().__init__` stays valid. Unused `Callable` and `EvalPrediction` imports are removed from the distillation trainer module.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8fdeec634fdb56dd01882d8a32c54b2a55f13c81. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kashif (approved):
LGTM — dropping unused `compute_metrics` / `preprocess_logits_for_metrics` and requiring `model` matches GRPO. Minor follow-up: `ServerDistillationTrainer` still defaults `model=None` in its signature; callers omitting `model` already failed later, so not a regression, but tightening that default would make the new contract consistent.

PR #6618, opened 2026-08-01
Fix unused VLM parameters for text-only training

### LINKED ISSUE #6617: Training a VLM on a text-only dataset leaves the vision tower trainable — DDP crash, and wasted optimizer state even on one GPU
Passing a vision-language model id to `SFTTrainer` with a **text-only** dataset (the documented happy path) loads the full model including the vision tower. The vision tower never participates in the loss, so:

- **On >1 GPU** training dies with DDP's generic reducer error, which does not name the cause.
- **On 1 GPU** it trains fine but the vision tower sits in the optimizer, consuming memory and slowing every step, silently.

### Reproduction

```python
import torch
from datasets import Dataset
from trl import SFTTrainer

ds = Dataset.from_list([{"prompt": "What is 2+2?", "completion": " 4."}] * 8)

trainer = SFTTrainer(
    model= "trl-internal-testing/tiny-Qwen2VLForConditionalGeneration",
    train_dataset=ds,
)
m = trainer.model
vis = [(n, p) for n, p in m.named_parameters() if "visual" in n or "vision" in n]
print(f"vision params: {len(vis)}, requires_grad: {sum(p.requires_grad for _, p in vis)}")

batch = {k: (v.to(m.device) if torch.is_tensor(v) else v) for k, v in batch.items()}
trainer.compute_loss(m, batch).backward()
print("no gradient after backward:", sum(p.requires_grad and p.grad is None for _, p in vis))
```

Output:

```
vision params: 31, requires_grad: 31
no gradient after backward: 31
```

Under `accelerate launch --num_processes 2` the same setup fails with:

```
RuntimeError: Expected to have finished reduction in the prior iteration before starting a new
one. This error indicates that your module has parameters that were not used in producing loss.
```

### Why it's worth fixing

Measured on `Qwen/Qwen3.5-2B-Base` (`Qwen3_5ForConditionalGeneration`, 1.88B text tower + 331M vision tower), text-only SFT:

- ~5.3 GB per rank of fp32 weights + Adam state for a tower that receives no gradient.
- Freezing the vision tower gave a **3× step-time improvement!** (6.9 s/it → 2.1 s/it) in our run, because it also drops out of the optimizer.
- Reaching that fix took **three failed launches and ~11 minutes**, including turning on `TORCH_DISTRIBUTED_DEBUG=DETAIL` and reading `modeling_*.py`, because the DDP error names no parameters by default.

### Suggested fixes, roughly in order of preference

**Freeze the vision tower when the dataset is text-only.** TRL knows both facts at construction time. This is almost certainly what the user wants (they are not training vision) and it fixes the DDP crash and the wasted optimizer state together.

Note that this can be applied to all trainers.

### Environment

trl 1.9.2
transformers 5.14.1
torch 2.11.0+cu130
2× H100

cc @AmineDiro because you mentioned this issue for async grpo as well

DaoyuanLi2816:
Fixes #6617

Text-only VLM datasets now freeze parameters outside the language model in SFT, DPO, KTO, GRPO, and RLOO. This keeps the unused vision path out of DDP reduction and optimizer state while preserving existing language-model and output-head trainability, including PEFT adapters.

Image datasets remain fully trainable. GRPO also keeps non-language parameters trainable when tools can introduce multimodal results during a rollout.

Validation:

- `pytest tests/test_model_utils.py -q` (4 passed, 3 skipped)
- text-only VLM regressions for SFT, DPO, and KTO (8 passed)
- text-only GRPO/RLOO initialization regressions, including iterable datasets (4 passed)
- image-dataset GRPO/RLOO training regressions with tiny Qwen2.5-VL (2 passed)
- tiny Qwen2.5-VL prompt-tuning integration check
- Ruff lint and format checks on all changed files

## Before submitting

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **Medium Risk**
> Changes which parameters are trainable at trainer init across six algorithms; behavior is gated on VLM + text-only data but affects distributed training and optimizer state.
> 
> **Overview**
> Adds **`freeze_non_language_model_parameters`** so vision-language trainers automatically freeze everything outside the language stack (and **`lm_head`**) when the train set has **no `image`/`images` columns**, avoiding unused vision weights in DDP and the optimizer while language (and existing PEFT) parameters stay trainable.
> 
> **SFT**, **DPO**, **KTO**, **Distillation**, **GRPO**, and **RLOO** call this at init when `_is_vlm` and the data is text-only. **GRPO** and **Distillation** skip freezing if **tools** are configured, since rollouts can still introduce images. **RLOO** now sets **`_is_vlm`** from whether the processing class is a processor.
> 
> Tests add **`get_vision_parameter_names`** (vision tower via `vision_config`, not hard-coded `model.visual` prefixes), unit tests for the freeze helper (including PEFT prompt encoders), and broader VLM text-only coverage (e.g. LLaVA).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1185dc9f489a70939bd17fe6b4ae78b46c6a9f23. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_kto_trainer.py
```diff
@@ -1545,6 +1545,17 @@ def test_train_vlm_text_only_data(self, model_id, dataset_config):
             train_dataset=dataset,
         )
 
+        vision_parameters = [
+            parameter for name, parameter in trainer.model.named_parameters() if name.startswith("model.visual")
```
qgallouedec: does it really work with any vlm?

DaoyuanLi2816:
Yes for the implementation, no for the test I wrote — you were right to ask.

`freeze_non_language_model_parameters` never looks at module names. It finds the text model by identity against `base_model.config.get_text_config()`, freezes everything, then re-enables only the parameters that were trainable before and live under the text model, its output embeddings, and the prompt encoder for prompt-learning PEFT. Nothing in that is Qwen-specific, and it degrades safely: a model whose config exposes no separate text config returns early and is left untouched.

The test was the Qwen-specific part. `name.startswith("model.visual")` only matches Qwen-VL — LLaVA and Gemma 3 use `vision_tower`, SmolVLM uses `vision_model` — so on any other architecture the `assert vision_parameters` would have failed for the wrong reason, and the post-training branch would have classified every vision parameter as trainable.

So I removed the name matching. `tests/testing_utils.py` gains `get_vision_parameter_names`, which locates the tower through `config.vision_config` the same way the fix locates the text model, and the tests now derive the frozen/trainable split from `requires_grad` rather than from prefixes. That also fixes a subtler problem with the old branch: the freeze covers everything outside the language path, including the multimodal projector, so keying the post-training check on "is it vision" was wrong for LLaVA-style models even in principle.

To make "any VLM" a claim the suite actually checks rather than one I assert, I added `tiny-LlavaForConditionalGeneration` to the KTO test you commented on. Confirmed the helper resolves against four families:

| model | tower module | params found |
| --- | --- | --- |
| `tiny-Qwen2_5_VLForConditionalGeneration` | `model.visual` | 30 |
| `tiny-LlavaForConditionalGeneration` | `model.vision_tower` | 39 |
| `tiny-SmolVLMForConditionalGeneration` | `model.vision_model` | 37 |
| `tiny-Gemma3ForConditionalGeneration` | `model.vision_tower` | 37 |

Happy to add a second architecture to the SFT, DPO, GRPO and RLOO tests too — I kept it to KTO for now to avoid multiplying CI time across their larger `dataset_config` matrices. Say the word if you would rather have it everywhere.

I left the two pre-existing `model.visual` branches in `test_train_vlm_text_only_data_packing` alone, since they are outside this PR.

Validation:

- `pytest tests/test_kto_trainer.py -k test_train_vlm_text_only_data`: 4 passed (both architectures)
- `pytest tests/test_dpo_trainer.py tests/test_sft_trainer.py -k test_train_vlm_text_only_data`: 7 passed
- `pytest tests/test_grpo_trainer.py tests/test_rloo_trainer.py -k "freezes_non_language_parameters"`: 4 passed
- `pytest tests/test_grpo_trainer.py::TestGRPOTrainerVLM::test_train_vlm tests/test_rloo_trainer.py::TestRLOOTrainerVLM::test_train_vlm`: 14 passed
- `ruff check` and `ruff format --check` on the changed files

tests/test_kto_trainer.py
DaoyuanLi2816: Answered this in a top-level comment on 5 Aug rather than here, which left the thread looking unaddressed — sorry about that. Short version, with the detail in that comment:

Yes for the implementation, no for the test as it stood. `freeze_non_language_model_parameters` never matches on module names — it resolves the text model by identity against `base_model.config.get_text_config()` — so it is genuinely architecture-agnostic. The test was not: it asserted on a `vision_tower` attribute that only some families expose.

I checked the helper against Qwen2-VL, Qwen2.5-VL, LLaVA, Idefics2, PaliGemma and SmolVLM, and rewrote the test to asserton the frozen/trainable partition instead of on any attribute name. Happy to narrow the claim in the docstring if you would rather it only promise the families that are covered by tests.

DaoyuanLi2816: Correction to my previous message, before it misleads: I listed the wrong architectures. I wrote them from memory instead of from the PR, and three of the six were never tested.

What the suite actually covers is four families, all of them tiny checkpoints:

| model | tower module | params |
| --- | --- | --- |
| `tiny-Qwen2_5_VLForConditionalGeneration` | `model.visual` | 30 |
| `tiny-LlavaForConditionalGeneration` | `model.vision_tower` | 39 |
| `tiny-SmolVLMForConditionalGeneration` | `model.vision_model` | 37 |
| `tiny-Gemma3ForConditionalGeneration` | `model.vision_tower` | 37 |

Qwen2-VL, Idefics2 and PaliGemma are not tested — I should not have named them. Gemma 3 is, and I left it out.

The substance of the answer stands: the helper resolves the text model by identity through `config.get_text_config()` and the tower through `config.vision_config`, so it is not keyed on module names, and the four above are what actually demonstrates that. The offer also stands — I can narrow the docstring to promise only what is covered, or extend coverage to more families, whichever you prefer.

DaoyuanLi2816: Second correction on this thread, and I owe you an explanation for why there are two. Both earlier lists were written from memory instead of read off the diff. This one is line-by-line from the branch, so it can be checked against the files directly.

The freeze assertions in this PR hang off six tests. What each one actually runs:

| test | models |
| --- | --- |
| `test_dpo_trainer.py::test_train_vlm_text_only_data` (L1595) | Qwen2.5-VL |
| `test_kto_trainer.py::test_train_vlm_text_only_data` (L1547) | Qwen2.5-VL, LLaVA |
| `test_sft_trainer.py::test_train_vlm_text_only_data` (L2042) | Qwen2.5-VL |
| `test_grpo_trainer.py::test_text_only_dataset_freezes_non_language_parameters` (L3584) | Qwen2.5-VL |
| `test_rloo_trainer.py::test_text_only_dataset_freezes_non_language_parameters` (L1806) | Qwen2.5-VL |
| `test_model_utils.py` (2 tests) | synthetic `PreTrainedModel` subclasses, no checkpoint |

So two real families, not four and not six. Gemma 3 and SmolVLM do appear in `test_dpo_trainer.py` / `test_kto_trainer.py`, but in the *other* VLM parametrization in those classes, not in the tests carrying these assertions — and SmolVLM is commented out there in any case ("seems not to support bf16 properly").

On the actual question: the generality rests on the mechanism rather than the matrix. `freeze_non_language_model_parameters` never matches module names — it resolves the text model by identity against `base_model.config.get_text_config()`, and `get_vision_parameter_names` resolves the tower by identity against `config.vision_config`. That is why `visual`, `vision_tower` and `vision_model` all work without being enumerated anywhere. The synthetic tests in `test_model_utils.py` pin that path plus the PEFT `prompt_encoder` case Bugbot raised.

But the executed evidence is narrower than I twice implied, and two families is what you should weigh it as. If you would rather the matrix matched the claim, I am happy to widen the parametrization on those five tests — LLaVA-Next and Gemma 3 are already used in the same files, so it is a one-line change each, and I would rather add the runs than keep arguing the mechanism.

qgallouedec:
Thanks for staying on top of the review, @DaoyuanLi2816. One more spot to cover: DistillationTrainer graduated to the stable trainer surface with VLM support after you opened this PR, and it sets `_is_vlm` the same way SFT/DPO/KTO do, so it has the same unused-vision-tower issue.

Could you add the freeze there too? And since you already offered it on the test-coverage thread, widening the parametrization to match the generality claim would close that thread out as well.

DaoyuanLi2816:
Added the freeze to the stable DistillationTrainer and covered it with Qwen2.5-VL and LLaVA text-only regression cases. I also widened the SFT, DPO, GRPO, and RLOO text-only VLM coverage to the same two architecture families.\n\nValidation: 2 Distillation cases passed; 24 focused SFT/DPO/KTO/GRPO/RLOO cases passed; focused Ruff check and format passed.

PR #6625, opened 2026-08-02
Fix missing entropy gradient in _ChunkedLogProbFunction backward

verma8076:
## What does this PR do?

`_ChunkedLogProbFunction` (`trl/trainer/utils.py`) is a custom `torch.autograd.Function` used by `patch_chunked_lm_head` to compute `(logprobs, entropy)` from the lm_head without materializing the full `[N, V]` logits tensor. Its `backward` signature accepts both upstream gradients:

```python
def backward(ctx, grad_logprobs: torch.Tensor, grad_entropy: torch.Tensor):
```

but `grad_entropy` was never used anywhere in the method body — only `grad_logprobs` fed into `grad_hidden`/`grad_weight`. Any loss term that depends on the `entropy` output (e.g. an entropy bonus) would silently get zero gradient contribution from it, with no error or warning.

This PR implements the missing gradient. For softmax entropy `H(p) = -sum_k p_k log p_k` over logits `z`, `dH/dz_j = -p_j * (log p_j + H)`. This is added into the existing chunked backward accumulation (same chain-rule structure already used for the `logprobs` gradient, so it correctly composes with temperature scaling / logit softcapping / logit scale), and `entropy` is now saved via `ctx.save_for_backward` (it was computed in `forward` but previously discarded before `backward` needed it).

## Why this matters despite no live crash

The only current caller (`AsyncGRPOTrainer`) only reads `entropy` inside a `torch.no_grad()` block for metric logging, so there's no user-facing regression today. But this is exported, non-experimental code (`trl.trainer.utils`), and any future feature that adds an entropy-bonus loss term (a common technique in policy-gradient RL) would hit this silently — the loss would appear to include an entropy term, training would run without error, and the entropy gradient would simply be missing throughout.

## Testing

Added `test_backward_entropy` (backprop through `entropy` alone) and `test_backward_combined` (backprop through a weighted sum of `logprobs` and `entropy` together, to catch the two gradient contributions overwriting rather than accumulating) to `TestChunkedLogProbFunction` in `tests/test_utils.py`, following the exact structure/tolerances of the existing `test_backward`/`test_backward_bfloat16` tests in that class.

Verified both new tests fail against the pre-fix code (95%+ mismatched elements, confirming they exercise the bug) and pass against the fix. Also verified the fix against a from-scratch reference implementation across several independent loss combinations (entropy-only, logprobs-only, combined with mixed coefficients, per-row-weighted entropy) and a coarse-tolerance `torch.autograd.gradcheck` (the function intentionally accumulates in `float32` regardless of input dtype per the existing `NOTE(@aminediro)` comment, so gradcheck needs a step size/tolerance compatible with that precision floor rather than the default `eps=1e-6`).

Full existing `TestChunkedLogProbFunction` suite (`test_forward`, `test_backward`, `test_backward_bfloat16`) still passes unmodified — no regression to the `logprobs` gradient path.

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr)?
- [x] Was this discussed/approved via a GitHub issue? N/A — small, self-contained bugfix.
- [x] Did you make sure to update the documentation with your changes? N/A — no user-facing API/behavior change beyond the bugfix.
- [x] Did you write any new necessary tests? Yes, see Testing above.

---

> [!NOTE]
> **Medium Risk**
> Changes custom autograd for chunked LM-head logprobs/entropy; incorrect entropy grads would silently break RL-style entropy bonuses, though current async GRPO usage only reads entropy under `no_grad`.
> 
> **Overview**
> Fixes a bug in **`_ChunkedLogProbFunction`** where `backward` accepted `grad_entropy` but never applied it, so any loss that backprops through the **entropy** output (e.g. an entropy bonus) would get **zero** gradient from that term with no error.
> 
> The chunked backward now adds the softmax-entropy derivative \(-p_j (\log p_j + H)\) into the per-chunk `grad_logits` accumulation (alongside the existing logprob path), persists **`entropy`** in `ctx.save_for_backward`, and treats **`grad_logprobs` / `grad_entropy` as optional** so each branch can be skipped. **`ctx.set_materialize_grads(False)`** avoids materializing zero entropy grads when entropy is only used for logging.
> 
> **Tests:** `test_backward_entropy` and `test_backward_combined` in `TestChunkedLogProbFunction` compare chunked grads to a reference implementation (entropy-only and mixed logprob+entropy losses).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f4af95cc1c3e4d0b779ed4826d0f634a6470beb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verma8076:
Bumping this one — it's been about three weeks with no review yet. @qgallouedec, would you be able to take a look, or point me to who owns GRPO trainer changes? Happy to adjust if anything's off.

qgallouedec:
Sorry for the slow turnaround here. We get a huge amount of automated contributions these days and it honestly makes us less willing to spend time on external PRs, which is unfair to the ones like this that are actually well done.

So on the substance: the math is right. And the bug is real: `grad_entropy` was accepted and dropped, so any loss backpropagating through entropy would silently get nothing.

One thing important though. nothing consumes that gradient today (`AsyncGRPOTrainer` reads entropy under `no_grad` for logging only), but autograd materializes `grad_entropy` as zeros, so every backward now computes an entropy term that gets multiplied by 0. So merging as is would add compute time + VRAM for everyone, for nothing.

Can you add `ctx.set_materialize_grads(False)` in `forward` and skip the entropy block when `grad_entropy is None`? Note `grad_logprobs` can be `None` too then, so both need the guard. Your two new tests keep covering that path.

Thks for the clean writeup and the tests, that made this easy to check.

verma8076:
Pushed. Added `ctx.set_materialize_grads(False)` to `forward`, and guarded both `grad_logprobs` and `grad_entropy` independently in `backward` (early-return if both are `None`, and each contribution to `grad_logits` is now skipped when its corresponding grad is `None`).

Verified against the reference implementation with a standalone script hooking `backward` to confirm the guard is actually exercised (not just dead code): with `logprobs.sum().backward()`, `grad_entropy` does come back `None`; with `entropy.sum().backward()`, `grad_logprobs` does come back `None`; gradients match the reference to 1e-4 in both cases and in combined backward. The two new tests (`test_backward_entropy`, `test_backward_combined`) still pass under this.

qgallouedec:
Thks, that's it. re-benched on H100, N=8192, H=4096, V=128k, chunk=8192, logprobs-only backward:

```
main:    413.7 ms, peak 4.93 GiB
before:  429.5 ms, peak 5.09 GiB
now:     418.9 ms, peak 5.09 GiB
```

so most of it is gone. two lines in the loop still run when `g_entropy is None` though:

- `log_p_chunk = logits_chunk - log_z.unsqueeze(-1)` is computed unconditionally but only read by the entropy branch, that's the extra `[N, C]` fp32 buffer and the +160 MiB
- `grad_logits = torch.zeros_like(probs)` then `grad_logits -= g.unsqueeze(-1) * probs` costs an alloc and a pass more than the old fused `(-g).unsqueeze(-1) * probs`

moving `log_p_chunk` into the entropy branch and keeping the fused form when `g is not None` gets us to 412.1 ms / 4.93 GiB, i.e. the baseline, tests still pass. Last thing and then this is good to go.

Nit: the `if grad_logprobs is None and grad_entropy is None` early return is unreachable. with `set_materialize_grads(False)`, if neither output is used the node isn't on the backward path and `backward` never gets called. can you drop it, we try to keep this file free of defensive branches.

verma8076:
Pushed. Moved \`log_p_chunk\` into the entropy branch (only computed when \`g_entropy is not None\`), kept the fused \`(-g).unsqueeze(-1) * probs\` form for the logprobs term instead of the \`zeros_like\` + subtract, and dropped the unreachable both-\`None\` early return.

Re-verified gradient correctness against the reference for logprobs-only, entropy-only, and combined backward (all match to 1e-4) — don't have H100 access to reproduce your exact benchmark numbers, but the code path for \`g_entropy is None\` now does the identical work main does (no \`log_p_chunk\` alloc, no extra buffer, same fused subtract), so it should land at the same ~412ms/4.93GiB baseline.

qgallouedec (approved):
that's it, thks. re-benched the merged head on H100, same setup as above:

```
main:     410.9 / 412.5 ms, peak 4.93 GiB
this PR:  411.9 / 412.6 ms, peak 4.93 GiB
```

parity on both time and memory, so the entropy path is now free when nobody asks for it. chunked tests pass on the merge.

~one thing for later, not blocking: the new tests only parametrize temperature, so the entropy gradient isn't covered for the softcapping and logit_scale paths. same gap the existing logprobs tests have, so I'm not asking you to fix it here.~

nice work, and sorry again for the wait.

verma8076:
Thanks for the review and the re-bench — good to know it lands at parity. Noted on the softcapping/logit_scale gap in the entropy tests; agreed it's the same gap the existing logprobs tests have, so leaving it for whatever follow-up eventually closes that for both rather than doing it piecemeal here.

PR #6629, opened 2026-08-02
[DistillationTrainer refactor] Reject `use_liger_kernel` with `logit_scale` / `final_logit_softcapping`

qgallouedec:
Part of #6449. Correctness fix surfaced by a deep loss audit.*

The chunked JSD loss applies each model's Cohere `logit_scale` and Gemma-2 `final_logit_softcapping` inside `_chunk`, matching the model's real forward. `LigerFusedLinearJSDLoss` has **no** such parameters and the trainer never applies them — so for a Gemma-2 or Cohere student/teacher with `use_liger_kernel=True`, the Liger path **silently optimizes a different objective** than the default chunked path (and one that no longer matches the model's forward). `test_liger_loss_agrees_with_chunked` doesn't catch it (tiny-Qwen2 has neither).

- `__init__` now raises when `use_liger_kernel=True` and either the student or teacher config sets `logit_scale != 1.0` or `final_logit_softcapping is not None`, pointing users to `use_liger_kernel=False` (the chunked path applies both).
- Test `test_liger_incompatible_with_logit_softcapping_raises`.

Verified: ruff clean; guard raises for softcap+Liger, allows softcap+chunked and plain+Liger.

---

> [!NOTE]
> **Medium Risk**
> Changes distillation training correctness and init-time validation for Liger users on Cohere/Gemma-style models; chunked-path behavior for `logit_scale=None` is a small semantic fix.
> 
> **Overview**
> **`DistillationTrainer` now fails fast** when `use_liger_kernel=True` and the student or teacher config uses Cohere-style `logit_scale` (anything other than `None` or `1.0`) or Gemma-style `final_logit_softcapping`, because the fused Liger JSD path cannot mirror those transforms and would otherwise train on the wrong objective. The error tells users to turn Liger off and use the chunked loss.
> 
> On the **chunked loss path**, `logit_scale=None` is treated as unscaled (`1.0`) while a real `0.0` is still passed through; that matches how MPT-style configs differ from a degenerate zero scale.
> 
> **Tests** cover softcapping + Liger (raises), `logit_scale=None` + Liger (allowed), and `logit_scale=0.0` + Liger (raises).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 408250e8c45cb883a3631f0fd5d5cbd03463b221. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/distillation/distillation_trainer.py
+++ b/trl/experimental/distillation/distillation_trainer.py
@@ -597,7 +597,7 @@ class DistillationTrainer(_BaseTrainer):
             # objective than the model's real forward.
             if self.use_liger_loss:
                 for name, config in [("student", self.model.config), ("teacher", teacher_model.config)]:
-                    scaled = getattr(config, "logit_scale", 1.0) != 1.0
+                    scaled = getattr(config, "logit_scale", 1.0) not in (None, 1.0)
                     softcapped = getattr(config, "final_logit_softcapping", None) is not None
                     if scaled or softcapped:
                         raise ValueError(
@@ -1231,8 +1231,8 @@ class DistillationTrainer(_BaseTrainer):
             num_items_in_batch=num_items_in_batch,
             student_lm_head_bias=student_lm_head.bias,
             teacher_lm_head_bias=teacher_lm_head.bias,
-            student_logit_scale=getattr(student_config, "logit_scale", 1.0),
-            teacher_logit_scale=getattr(teacher_config, "logit_scale", 1.0),
+            student_logit_scale=getattr(student_config, "logit_scale", 1.0) or 1.0,
+            teacher_logit_scale=getattr(teacher_config, "logit_scale", 1.0) or 1.0,
             student_final_logit_softcapping=getattr(student_config, "final_logit_softcapping", None),
             teacher_final_logit_softcapping=getattr(teacher_config, "final_logit_softcapping", None),
             temperature=self.temperature,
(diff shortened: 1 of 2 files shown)


trl/experimental/distillation/distillation_trainer.py
```diff
@@ -1215,8 +1231,8 @@ def _compute_loss(self, unwrapped_student, inputs, num_items_in_batch):
             num_items_in_batch=num_items_in_batch,
             student_lm_head_bias=student_lm_head.bias,
             teacher_lm_head_bias=teacher_lm_head.bias,
-            student_logit_scale=getattr(student_config, "logit_scale", 1.0),
-            teacher_logit_scale=getattr(teacher_config, "logit_scale", 1.0),
+            student_logit_scale=getattr(student_config, "logit_scale", 1.0) or 1.0,
+            teacher_logit_scale=getattr(teacher_config, "logit_scale", 1.0) or 1.0,
```
cmpatino: Turbo Nit: Should we align the behavior of `logit_scale=0` with what we do for the Liger version (https://github.com/huggingface/trl/pull/6629/changes#diff-03c25c1f305d4a0687da7ce7207f80fe582927a92dd37ae8efd5b80341b4bc1cR600)?

We reject `logit_scale=0` for Liger but we instead change it to 1 for the chunked loss.

cmpatino (approved):
I left a non-blocking comment about how we handle the `logit_scale`.

trl/experimental/distillation/distillation_trainer.py
qgallouedec: thanks, addressed in https://github.com/huggingface/trl/pull/6629/commits/af39fc3bdc652603ee5fedb8ef443cd939a5904c

qgallouedec: actually, disallowing 0.0 is better: https://github.com/huggingface/trl/pull/6629/commits/408250e8c45cb883a3631f0fd5d5cbd03463b221

PR #5189, opened 2026-02-26
35% faster packing + rename `bfd-requeue` to `bfd_split`

mariosasko:
# What does this PR do?

This PR improves the packing logic to make it faster, less error-prone, and easier to read.

The main changes are:

* Replacing the AI-generated BFD splitting (a.k.a. "requeuing") logic from https://github.com/huggingface/trl/pull/4632
 with a vectorized implementation that is significantly shorter and 30% faster.

* Updating `pack_examples` to restore the input dataset’s format and perform proper input validation.

* Applying a minor optimization to the `wrapped` packing implementation by reusing the `offsets` across all packed columns.

* Aligning the naming with the literature (e.g., `requeue` → `split`) while preserving backward compatibility.

P.S. The recent [Qwen3-Coder-Next technical report](https://github.com/QwenLM/Qwen3-Coder/blob/main/qwen3_coder_next_tech_report.pdf) includes a nice comparison of packing techniques, which would be a nice addition to the docs. ~~However, the report is not yet available on arXiv, so it cannot be cited as an HF paper.~~ Done 😄 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

---

> [!NOTE]
> **Medium Risk**
> Touches core SFT data preprocessing (packing/truncation) and changes strategy semantics/validation, which can alter token ordering/retention and training results if misconfigured; backward compatibility is handled via a deprecation warning.
> 
> **Overview**
> **Improves SFT dataset packing and clarifies packing strategies.** The BFD packer is rewritten with a more vectorized PyArrow implementation, adds shared column validation, filters empty sequences, and introduces an explicit overflow behavior switch (*truncate* vs *split*) used by the new `"bfd_split"` strategy.
> 
[...]

PR #6670, opened 2026-08-06
Fix wrapped packing duplicating data when the input table is sliced

### LINKED ISSUE #6669: `pack_dataset` with `strategy="wrapped"` silently duplicates data across `map` batches
### Reproduction

```python
from datasets import Dataset
from trl import pack_dataset

ds = Dataset.from_dict({"input_ids": [[1, 2], [3, 4], [5, 6], [7, 8], [9, 10], [11, 12]]})
packed = pack_dataset(ds, seq_length=4, strategy="wrapped", map_kwargs={"batch_size": 3})
print(packed.to_dict()["input_ids"])
```

outputs:

```
[[1, 2, 3, 4], [5, 6], [1, 2, 3, 4], [5, 6]]
```

expected:

```
[[1, 2, 3, 4], [5, 6], [7, 8, 9, 10], [11, 12]]
```

Rows 3 to 5 come back as a second copy of rows 0 to 2. Nothing raises, and the row and token counts are both unchanged, so the corruption is silent.

The `batch_size=3` above is only to keep the snippet short. The same thing happens at the default `map` batch size once the dataset is larger than one batch:

```python
from datasets import Dataset
from trl import pack_dataset

n, per_row = 2500, 4
ids = [[r * per_row + k for k in range(per_row)] for r in range(n)]
ds = Dataset.from_dict({"input_ids": ids})

packed = pack_dataset(ds, seq_length=8, strategy="wrapped")
expected = {t for row in ids for t in row}
got = {t for row in packed.to_dict()["input_ids"] for t in row}
print(f"tokens lost: {len(expected - got)} / {len(expected)}")
```

outputs:

```
tokens lost: 6000 / 10000
```

Reproduces identically on `datasets` 4.7.0 (the minimum TRL requires) and 5.0.1, with `pyarrow` 21.0.0 and 25.0.0.

## Identified Cause

`_pack_wrapped` computes the number of values from the array's sliced view, then builds the new array from the raw child buffer:

https://github.com/huggingface/trl/blob/main/trl/data_utils.py#L829-L842

```python
offsets, values = columns[0].offsets, columns[0].values
values = values[offsets[0].as_py() : offsets[-1].as_py()]   # slice-corrected
num_elements = len(values)                                  # uses the correction
offsets = np.arange(0, num_elements, seq_length, dtype=...)
offsets = np.concatenate((offsets, [num_elements]))
columns = [
    type(column).from_arrays(offsets.astype(...), column.values)   # ignores it
    for column in columns
]
```

`column.values` is the whole child buffer of the underlying table, so it still contains the rows before the array's own offset. The new offsets are then applied from position 0 instead of from the slice start. The `values` local is corrected and then never used.

`datasets.map(batched=True)` hands each batch to the packing function as a slice of the underlying table, so batches after the first can start at a non-zero offset.

This looks like a regression from #5189, which rewrote the loop as a list comprehension over all columns and switched to `column.values`. The version before it passed the corrected `values` and was right.

## Scope

Only `wrapped` is affected. `bfd` and `bfd_split` reach the same `column.values` pattern but are safe, because `pc.filter` and `pc.take` materialize fresh arrays with offset 0 first. So switching strategy is a workaround.

Triggering needs the table to actually be sliced, which is why it is not constant. When chunk boundaries line up with the `map` batch size, `combine_chunks` concatenates into a fresh array and masks it, and an indices mapping from `shuffle`, `filter`, or `select` makes `datasets` build batches with `take`, which masks it too.

On `SFTTrainer` this means the default path is not affected, since `shuffle_dataset=True` shuffles right before packing. The combination that does corrupt is `packing_strategy="wrapped"` with `shuffle_dataset=False` and `dataset_num_proc` set, which loses 1000 of 10000 tokens in my test.

### System Info

- Platform: macOS-26.5.2-arm64-arm-64bit
- Python version: 3.12.4
- TRL version: main @ d242e584
- PyTorch version: 2.13.0
- Transformers version: 5.14.1
- Datasets version: 5.0.1
- PyArrow version: 25.0.0
- Accelerator: none, CPU-only reproduction

### Checklist

SuryanshSS1011:
## What does this PR do?

Fixes #6669

`_pack_wrapped` sizes the output from the array's sliced view but reads from the raw child buffer, so when the input table is a slice it re-reads from the start of the buffer instead of from the slice. `datasets.map(batched=True)` passes each
batch as a slice, so every batch after the first can come back holding data from the beginning of the dataset. Row and token counts are unchanged, so nothing errors.

```python
ds = Dataset.from_dict({"input_ids": [[1, 2], [3, 4], [5, 6], [7, 8], [9, 10], [11, 12]]})
pack_dataset(ds, seq_length=4, strategy="wrapped", map_kwargs={"batch_size": 3})["input_ids"]
# before: [[1, 2, 3, 4], [5, 6], [1, 2, 3, 4], [5, 6]]
# after:  [[1, 2, 3, 4], [5, 6], [7, 8, 9, 10], [11, 12]]
```

The correction was already computed into a local, but #5189 rewrote the loop as a list comprehension over all columns and passed `column.values`, which left that local unused. This switches to `column.flatten()`, which respects the list array's own offset, per column so each keeps its own offsets.

Only `wrapped` is affected. `bfd` and `bfd_split` hit the same pattern but are protected by the `pc.filter` and `pc.take` that run before it.

Output is byte-identical to current behavior when the table is not sliced, so no existing correct result changes. Also checked on `large_list` with int64 offsets.

## Tests

Added `TestPackDatasetWrapped::test_with_multiple_map_batches`, which packs across two `map` batches. It fails on main with the duplication above and passes here. The two existing `wrapped` tests use 3-row datasets, so they only ever exercise a single batch, which is why this was not caught.

```
pytest tests/test_data_utils.py -k "Pack"
10 passed, 548 deselected
```

Beyond the added test, I checked `_pack_wrapped` against an independent reference (concatenate each column's rows in order, then re-chunk) over 600 randomized shapes plus targeted edge cases: slices at every offset, `seq_length` of 1 and
equal to and greater than the total, ragged and empty rows, zero-row tables, multi-chunk tables, 1 to 3 columns, `list` and `large_list`, int32 and int64 offsets. All pass with the fix and 127 of them fail without it. `pack_dataset` also round-trips end to end across batch sizes 1 to 777, `num_proc` 2 and 4, and all three strategies, with `input_ids` / `attention_mask` / `labels` staying aligned.

Since #5189 was a performance PR, I benchmarked this too. `flatten()` is metadata work like `values`, so throughput is unchanged (at 500k rows x 16 tokens the median of 7 runs is 0.08 ms before and 0.09 ms after, which is within noise).

## Before submitting

## AI writing disclosure

## Who can review?

@qgallouedec @albertvillanova

---

> [!NOTE]
> **Medium Risk**
> The change alters training preprocessing for wrapped packing on multi-batch maps (silent wrong tokens before); unsliced inputs should stay byte-identical per the PR.
> 
> **Overview**
> Fixes **wrapped** sequence packing when `pack_dataset` runs through batched `Dataset.map`: batches after the first arrive as PyArrow table slices, and `_pack_wrapped` was sizing output from the slice but reading token data from `column.values` (the full child buffer), so later batches could repeat tokens from the start of the dataset instead of the current batch.
> 
> The implementation now uses **`column.flatten()`** per column so list data honors each array’s offset, then rebuilds packed list columns from those flattened values. **`bfd`** / **`bfd_split`** are unchanged in this diff.
> 
> Adds **`test_with_multiple_map_batches`** to pack with `map_kwargs={"batch_size": 3}` across two map batches and assert `input_ids` / `attention_mask` stay correct.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34ad2adbb06f92367de1d8e0721adc35d4afa715. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_data_utils.py
+++ b/tests/test_data_utils.py
@@ -1218,6 +1218,21 @@ class TestPackDatasetWrapped(TrlTestCase):
         assert next(iter(dataset.with_format(None).batch(batch_size=num_examples))) == expected_output
         assert formatting == dataset._formatting
 
+    def test_with_multiple_map_batches(self):
+        # `map` passes each batch as a slice, so every batch but the first starts at a non-zero offset
+        examples = {
+            "input_ids": [[1, 2], [3, 4], [5, 6], [7, 8], [9, 10], [11, 12]],
+            "attention_mask": [[0, 1], [1, 0], [0, 1], [1, 0], [0, 1], [1, 0]],
+        }
+        dataset = Dataset.from_dict(examples)
+        seq_length = 4
+        expected_output = {
+            "input_ids": [[1, 2, 3, 4], [5, 6], [7, 8, 9, 10], [11, 12]],
+            "attention_mask": [[0, 1, 1, 0], [0, 1], [1, 0, 0, 1], [1, 0]],
+        }
+        dataset = pack_dataset(dataset, seq_length, strategy="wrapped", map_kwargs={"batch_size": 3})
+        assert dataset.to_dict() == expected_output
+
 
 class TestPackDatasetBfd(TrlTestCase):
     def test_with_dataset(self):
--- a/trl/data_utils.py
+++ b/trl/data_utils.py
@@ -830,14 +830,15 @@ def _pack_wrapped(examples: pa.Table, seq_length: int) -> pa.Table:
     """Pack sequences in a pyarrow Table using a wrapped strategy."""
     columns = [column.chunks[0] for column in examples.combine_chunks().columns]
     _check_if_columns_can_be_packed(columns)
-    offsets, values = columns[0].offsets, columns[0].values
-    values = values[offsets[0].as_py() : offsets[-1].as_py()]
-    num_elements = len(values)
+    # Use `flatten()` and not `values`, which returns the whole child buffer and ignores the array's own offset. That
+    # offset is non-zero whenever the table is a slice, as it is for every batch but the first in a batched `map`.
+    values = [column.flatten() for column in columns]
+    num_elements = len(values[0])
     offsets = np.arange(0, num_elements, seq_length, dtype=columns[0].offsets.type.to_pandas_dtype())
     offsets = np.concatenate((offsets, [num_elements]))
     columns = [
-        type(column).from_arrays(offsets.astype(column.offsets.type.to_pandas_dtype()), column.values)
-        for column in columns
+        type(column).from_arrays(offsets.astype(column.offsets.type.to_pandas_dtype()), column_values)
+        for column, column_values in zip(columns, values, strict=True)
     ]
     return pa.Table.from_arrays(columns, names=examples.column_names)
 


qgallouedec (approved):
Thanks, verified independently

One thing to correct. The perf claim is not noise:

| | main | this PR |
|---|---|---|
| `_pack_wrapped`, 500k rows | 0.114 / 0.115 ms | 0.138 / 0.135 ms |
| `_pack_wrapped`, 2M rows | 0.296 / 0.305 ms | 0.346 / 0.342 ms |
| `pack_dataset`, 200k rows, bs=1000 | 36.6 / 37.0 ms | 41.3 / 40.9 ms |
| `pack_dataset`, 200k rows, bs=10000 | 15.7 / 15.7 ms | 16.8 / 16.9 ms |

That is +7% to +19%, and your own 0.08 vs 0.09 ms is the same +12%. It works out to a roughly constant 25 us per call, so 4 ms on a 200k-row dataset, once, and cached afterwards. Not a reason to hold this though.

PR #6671, opened 2026-08-06
fix(sft): warn when all labels are masked instead of silent 0.0 loss

### LINKED ISSUE #6668: SFTTrainer reports loss 0.0 when every label in the batch is masked
### Reproduction

When every label in a batch is masked, `SFTTrainer` reports `loss: 0.0` and finishes normally, even though the underlying forward loss is `NaN`. Nothing warns, and the run looks successful.

The easiest way to reach that state is truncation: with prompt-completion data and `completion_only_loss=True`, only completion tokens carry labels, so a `max_length` that cuts the sequence before any completion token survives leaves every label at `-100`.

```python
import torch
from datasets import Dataset

MODEL_ID = "EleutherAI/pythia-160m"

# Long prompt, short completion: as max_length shrinks, the completion is truncated away first.
data = Dataset.from_list(
    [
        {
            "prompt": f"You are a helpful assistant. Please answer this arithmetic question number {i}. "
                      f"Think about it carefully before responding. What is {i} plus {i}?",
            "completion": f" The answer is {2 * i}.",
        }
        for i in range(64)
    ]
)

for max_length in (128, 48, 32, 24, 16):
    args = SFTConfig(
        output_dir=f"/tmp/repro_{max_length}", max_length=max_length, completion_only_loss=True,
        packing=False, num_train_epochs=1, per_device_train_batch_size=8, learning_rate=5e-5,
        logging_steps=1, report_to="none", use_cpu=True, seed=42, save_strategy="no",
    )
    trainer = SFTTrainer(model=MODEL_ID, args=args, train_dataset=data)

    unmasked = int((batch["labels"] != -100).sum())
    trainer.model.eval()
    with torch.no_grad():
        raw = trainer.model(**{k: v for k, v in batch.items()
                               if k in ("input_ids", "attention_mask", "labels")}).loss.item()

    reported = trainer.train().training_loss
    print(f"max_length={max_length:>4}  unmasked_labels={unmasked:>3}  raw_forward_loss={raw:>10.6f}  reported={reported:>10.6f}")
```

Output:

```
max_length  unmasked_labels   raw_forward_loss   REPORTED_loss
       128               48           2.833474        1.888258
        48               48           2.833474        1.888258
        32               40           2.371209        1.422494
        24                0                nan        0.000000
        16                0                nan        0.000000
```

### Expected behavior

When a batch contains no unmasked labels there is nothing to learn from, so the run should fail loudly (or at least warn) rather than report `loss: 0.0` and exit successfully. Reporting `0.0` is actively misleading: the true value is `NaN`, and from the outside it is indistinguishable from a model that simply is not improving.

This matters beyond the truncation case, because any misconfiguration that masks every label (a chat template that never marks the assistant turn, an over-aggressive `max_length`, a prompt-only dataset) produces the same silent `0.0`. Several "my loss will not move" reports could plausibly land here, and there is currently no signal pointing at the cause. I ran into it while investigating #3910.

#5826 fixes the same class (`NaN` when completions are fully truncated) for `DPOTrainer` and the experimental `TPOTrainer`, but it does not touch `trl/trainer/sft_trainer.py`, so `SFTTrainer` still has this behaviour. Thanks @matdou for that work.

I am happy to send a PR, but the shape is your call, so I would rather ask first:

1. **Raise** on a batch with zero unmasked labels, which is a hard failure but unambiguous.
2. **Warn once** and keep going, which preserves current behaviour for anyone relying on it.
3. Handle it further upstream, at dataset preparation, by erroring when truncation empties every completion.

Also worth deciding whether the guard should live only in `SFTTrainer` or be shared, given #5826 is addressing the same thing for DPO and TPO.

### System Info

```
trl 1.0.0 (main @ d242e584)
transformers 5.11.0
torch 2.11.0
datasets 4.8.4
Python 3.11, CPU
```

cc @qgallouedec @albertvillanova

navaneethkrishnansuresh:
# What does this PR do?

Fixes #6668

When every label in a batch is masked (`-100`) — e.g. `completion_only_loss=True` and a small `max_length` cuts off all completion tokens — the real loss can be NaN while training reports `0.0` and looks successful.

This PR:
- Warns **once** when a batch has no unmasked labels (with guidance to increase `max_length` / fix data)
- Keeps a graph-safe `0.0` so backward/DDP still work
- Replaces NaN from the default CE path when all labels are masked
- Guards DFT loss against 0/0
- Adds unit tests

Chose **warn** (not hard fail) so runs continue but misconfiguration is visible.

## Before submitting

      https://github.com/huggingface/trl/issues/6668

## AI writing disclosure

## Who can review?

@qgallouedec @albertvillanova

---

> [!NOTE]
> **Medium Risk**
> Touches core SFT loss paths (`compute_loss`, `dft_loss`, chunked CE) and changes reported loss when all labels are masked; behavior is intentional but affects training metrics and distributed runs.
> 
> **Overview**
> Fixes **#6668**: batches where every label is `-100` could show **0.0** (or **NaN** on the default CE path) while training looked fine—common with `completion_only_loss` and a too-small `max_length`.
> 
> Adds **`_warn_all_labels_masked()`** (once per process) with actionable guidance, plus **`_labels_are_all_masked()`** for detection (including `shift_labels` under CP/SP). Warnings fire from chunked CE, **`dft_loss`**, and **`compute_loss`** when a batch is fully masked.
> 
> **Loss handling** stays graph-safe for backward/DDP: chunked CE unchanged; **`dft_loss`** avoids `0/0` NaN; **`compute_loss`** swaps NaN from default HF CE for a logits-connected **0.0**. Training continues (warn, not fail).
> 
> Regression tests cover the helper, DFT all-masked path, and chunked CE warning behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4d1d5609baca64f226544329211e0dae4ef326ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -40,6 +40,7 @@ from trl import SFTConfig, SFTTrainer
 from trl.trainer.sft_trainer import (
     DataCollatorForLanguageModeling,
     _chunked_cross_entropy_loss,
+    _labels_are_all_masked,
     _patch_chunked_ce_lm_head,
     dft_loss,
 )
@@ -69,6 +70,30 @@ if is_peft_available():
         TaskType,
         get_peft_model,
     )
+
+
+class TestAllLabelsMaskedHelpers(TrlTestCase):
+    """Regression for https://github.com/huggingface/trl/issues/6668."""
+
+    def test_labels_are_all_masked(self):
+        assert _labels_are_all_masked(None) is False
+        assert _labels_are_all_masked(torch.tensor([[-100, -100], [-100, -100]])) is True
+        assert _labels_are_all_masked(torch.tensor([[-100, 1], [-100, -100]])) is False
+
+    def test_dft_loss_all_masked_returns_zero_not_nan(self):
+        import trl.trainer.sft_trainer as sft_mod
+
+        batch_size, seq_len, vocab_size = 2, 3, 5
+        logits = torch.randn(batch_size, seq_len, vocab_size, requires_grad=True)
+        labels = torch.full((batch_size, seq_len), -100)
+        outputs = type("O", (), {"logits": logits})()
+        sft_mod._ALL_LABELS_MASKED_WARNED = False
+        with pytest.warns(UserWarning, match="no unmasked labels"):
+            loss = dft_loss(outputs, labels)
+        assert loss.item() == 0.0
+        assert not torch.isnan(loss)
+        loss.backward()
+        assert logits.grad is not None
 
 
 class TestDFTLoss(TrlTestCase):
@@ -2703,13 +2728,18 @@ class TestChunkedCrossEntropyLoss:
 
         Every trainable parameter of the chunked path (hidden_states, lm_head_weight, and lm_head_bias when present)
         must receive a gradient — otherwise DDP / FSDP synchronization hangs or errors at the all-reduce step.
+        Also warns once so a silent 0.0 is not mistaken for a healthy run (#6668).
         """
         hidden, weight, labels = self._inputs(requires_grad=True)
         bias = torch.zeros(self.V, dtype=torch.float32, requires_grad=True)
         labels[:] = -100
-        loss, correct, ent_sum, n_valid = _chunked_cross_entropy_loss(
-            hidden, weight, self.CHUNK_SIZE, labels, lm_head_bias=bias
-        )
+        import trl.trainer.sft_trainer as sft_mod
+
+        sft_mod._ALL_LABELS_MASKED_WARNED = False
+        with pytest.warns(UserWarning, match="no unmasked labels"):
+            loss, correct, ent_sum, n_valid = _chunked_cross_entropy_loss(
+                hidden, weight, self.CHUNK_SIZE, labels, lm_head_bias=bias
+            )
         assert loss.item() == 0.0
         assert correct.item() == 0.0
         assert ent_sum.item() == 0.0
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -190,7 +190,9 @@ def _chunked_cross_entropy_loss(
     if n_valid_tensor == 0:
         # Whole micro-batch masked (e.g. completion-only loss + truncation). Keep the loss connected
         # to the autograd graph through every trainable parameter so `.backward()` succeeds and DDP /
-        # FSDP gradient sync doesn't hang on a missing param.
+        # FSDP gradient sync doesn't hang on a missing param. Warn once so a reported 0.0 loss is
+        # not mistaken for a healthy run (#6668).
+        _warn_all_labels_masked()
         with maybe_gather_lm_head_ctx(lm_head_weight, lm_head_bias):
             loss = (hidden_states.float().sum() + lm_head_weight.float().sum()) * 0.0
             if lm_head_bias is not None:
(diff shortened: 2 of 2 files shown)


behroozazarkhalili (commented):
Thanks for picking this up. I filed #6668, so I ran your branch against the original reproduction rather than only reading the diff. Two results worth having on the PR.

**1. The reproduction in #6668 no longer reproduces on main, and that is not your doing.** #6320 added a `dataset.filter(any(label != -100))` at `trl/trainer/sft_trainer.py:1651`, so with `packing=False` the fully-masked rows are dropped during preprocessing. My original script now dies earlier with:

```
ValueError: num_samples should be a positive integer value, but got num_samples=0
```

Loud rather than silent, which is an improvement, but it never says why the dataset ended up empty.

**2. The bug is still reachable with `packing=True`, and your warning does fire there.** That filter sits inside `if args.max_length is not None and not packing:`, so packing skips it entirely. Long prompts plus a small `max_length` then produce packed chunks that are pure prompt:

```python
data = Dataset.from_list([
    {"prompt": f"You are a helpful assistant. Please answer this arithmetic question number {i}. "
               f"Think about it carefully and show your reasoning before responding. What is {i} plus {i}?",
     "completion": f" {2 * i}."}
    for i in range(32)
])
args = SFTConfig(output_dir=..., completion_only_loss=True, max_length=16, packing=True,
                 per_device_train_batch_size=4, use_cpu=True, report_to="none")
```

Same script, same data, on `main` and on this branch (the control's `sft_trainer.py` is byte-identical to `origin/main`):

| branch | packing | rows | all-masked batches | reported loss | warned |
|---|---|---|---|---|---|
| main | False | 0 (emptied by #6320) | n/a | n/a | n/a |
| main | True | 32 | 8 / 8 | 0.000000 | **False** |
| this PR | False | 0 (emptied by #6320) | n/a | n/a | n/a |
| this PR | True | 32 | 8 / 8 | 0.000000 | **True** |

So the PR does what it claims on the one path that still gets there: a silent `0.0` becomes a warned `0.0`.

**Two suggestions, both optional:**

- The new tests call `dft_loss` and `_chunked_cross_entropy_loss` directly, so nothing exercises `compute_loss` through a real trainer. The `packing=True` config above reaches your guard end-to-end and would make a good integration test, since it is currently the only configuration that does.
- The `packing=False` route is outside your warning's reach, because the rows are gone before training starts. A short warning where the filter drops rows (and especially where it empties the dataset) would cover the case that `num_samples=0` currently leaves unexplained. Happy to send that as a separate PR if you would rather keep this one focused.

One thing I checked and decided not to raise: the added `.item()` calls look like a per-step device sync, but `sft_trainer.py` already does that a dozen times per step in the metrics block (lines 1828 to 1897), so this is consistent with existing practice rather than a new cost.

cc @qgallouedec @albertvillanova

navaneethkrishnansuresh:
Thanks for the feedback , im happy to add the suggested end-to-end test if maintainers would prefer it in this PR.

behroozazarkhalili:
@qgallouedec @albertvillanova flagging a CI gate on this one. The `Tests` and `Build PR Documentation` workflow runs on head `4d1d5609` are both sitting at `conclusion=action_required`, dated 2026-08-10. That is the first-time-contributor approval gate rather than a failure, so the test suite has never run against this PR.

The reason it is easy to miss: the only green check here is Cursor Bugbot, which runs outside GitHub Actions and reports `success` regardless of the gate. So the PR reads as tested at a glance when nothing has executed.

#5790 is in the same state, `action_required` since 2026-07-24, also a first-time contributor. Whoever releases one can release both.

PR #6674, opened 2026-08-07
ci: exclude the buggy TruffleHog lob detector

behroozazarkhalili:
## What this fixes

The `Secret Leaks` workflow fails on branches whose push range contains a pytest function name of exactly 35 characters after `test_`. TruffleHog's Lob detector matches `\b((live|test)_[a-zA-Z0-9_]{35})\b`, which such a name satisfies exactly.

Eight names currently on `main` match, all of them ordinary tests:

```
test_applies_logit_scale_and_softcapping
test_history_rewrite_forks_into_two_rows
test_iw_opd_uses_cached_rollout_logprobs
test_length_exceeding_max_len_is_clamped
test_reward_func_wrong_number_of_rewards
test_train_runs_with_prompt_only_dataset
test_train_with_environment_owned_reward
test_vlm_uld_cross_arch_train_step_smoke
```

## Why they are reported as "verified"

The Lob detector treats two HTTP statuses as proof a key is live:

```go
switch res.StatusCode {
case http.StatusForbidden, http.StatusUnprocessableEntity:
    // 403 indicates key is active but no billing method on file
    // 422 indicates key is active but request body is invalid
    return true, nil
case http.StatusUnauthorized:
    return false, nil
```

The 403 is the one that fires here. Credit to @albertvillanova for pinning the actual trigger: Lob sits behind Cloudflare, which blocks TruffleHog's hardcoded `User-Agent: TruffleHog` and answers 403, while the same request from an ordinary client gets the correct 401:

```
UA='curl/8.0'           -> 401
UA='Go-http-client/2.0' -> 401
UA='TruffleHog'         -> 403   # Cloudflare "Attention Required!" HTML page
```

So the detector reads a WAF block as a verified credential. This also explains the timing, which nothing on either side accounts for otherwise: the detector is byte-identical in v3.95.9 and v3.96.0, and the symptom appeared on 2026-08-06 with no code change. It further explains why only new-branch pushes fail: with no base commit the action scans full history, while incremental pushes cover too few chunks to hit a match.

An earlier revision of this description attributed the 403/422 to the detector's empty request body producing a 422. That was wrong: Lob returns 401 for a bogus key from a normal client. The correction is @albertvillanova's, tested directly.

Upstream: trufflesecurity/trufflehog#5184. Related: trufflesecurity/trufflehog#5187 and trufflesecurity/trufflehog#5188, which tightens `keyPat` to exclude underscores and would fix this symptom but not the underlying "403 means verified" behavior.

## The change

Adds `lob` to the existing exclusion list, mirroring what the file already does for `postgres` and for the same reason. TRL has no Lob dependency; Lob is a direct-mail API, so the detector can only ever produce false positives here and there is no need to revert once upstream is fixed.

No credential was found, and nothing in the repository needs rotating.

cc @qgallouedec @kashif

### DIFF
--- a/.github/workflows/trufflehog.yml
+++ b/.github/workflows/trufflehog.yml
@@ -14,5 +14,7 @@ jobs:
     - name: Secret Scanning
       uses: trufflesecurity/trufflehog@27b0417c16317ca9a472a9a8092acce143b49c55  # v3.95.9
       with:
-        # exclude buggy postgres detector that is causing false positives and not relevant to our codebase
-        extra_args: --results=verified,unknown --exclude-detectors=postgres
+        # exclude buggy detectors that cause false positives and are not relevant to our codebase.
+        # lob matches `(live|test)_[a-zA-Z0-9_]{35}`, which any pytest name of exactly 35 characters after
+        # `test_` satisfies (e.g. `test_reward_func_wrong_number_of_rewards`).
+        extra_args: --results=verified,unknown --exclude-detectors=postgres,lob


albertvillanova (approved):
Approving — the exclusion is right, and the full-history scan on this branch confirms it (33,282 chunks, `verified_secrets: 0`).

One correction to the description, since it changes what we should expect from upstream: the empty-body/422 explanation doesn't hold — Lob returns 401 for a bogus key from a normal client. The actual trigger is Lob's Cloudflare WAF, which 403s on TruffleHog's hardcoded `User-Agent: TruffleHog`:

```
UA='curl/8.0'           -> 401
UA='Go-http-client/2.0' -> 401
UA='TruffleHog'         -> 403   # Cloudflare "Attention Required!" HTML page
```

and the detector maps 403 -> verified. That also explains why this appeared on 2026-08-06 with no change on either side (the detector is identical in v3.95.9 and v3.96.0), and why only new-branch pushes fail: no base commit means a full-history scan, while incremental pushes scan too few chunks to hit a match.

Details posted in trufflesecurity/trufflehog#5184.

No need to revert this once upstream is fixed — we have no Lob dependency, so the detector can only ever produce false positives for us. Same rationale as `postgres`.

albertvillanova:
Direct link to the upstream analysis, for tracking: trufflesecurity/trufflehog#5184 (comment) — https://github.com/trufflesecurity/trufflehog/issues/5184#issuecomment-5215426371

Related: trufflesecurity/trufflehog#5187 and PR trufflesecurity/trufflehog#5188 (tightens `keyPat` to exclude underscores — would fix our specific symptom, but not the 403-means-verified bug behind it).

behroozazarkhalili:
Thanks @albertvillanova, and you are right that my mechanism was wrong. I have corrected the PR description and credited the finding.

For the record on how I got it wrong, since the distinction matters: I read `verify()` and saw both 403 and 422 hardcoded to `return true, nil`, then reasoned from the empty request body that 422 was the status being hit. I never actually sent a request to `api.lob.com`, so I verified what the code *would* do and then narrated the runtime from it. One `curl` would have shown 401 and killed that story before it reached a description. Your User-Agent test is the real evidence.

Your version also explains two things mine did not. The timing: the detector is byte-identical in v3.95.9 and v3.96.0, so a code change on either side cannot account for the symptom appearing on 2026-08-06, whereas a WAF rule change can. And the shape: full-history scans on new branches hit a match while incremental pushes cover too few chunks, which is a cleaner account than the scan-range framing I had.

Worth flagging as well: the `// 403 indicates key is active but no billing method on file` comment sitting on that branch is the thing that makes it look settled. That is a true statement about a genuine Lob 403 and completely irrelevant to a Cloudflare 403, and it discourages exactly the check you ran.

Agreed on not reverting once upstream lands, and agreed that trufflesecurity/trufflehog#5188 addresses our symptom rather than the "403 means verified" behavior underneath it. Happy to leave this exclusion permanent on the same footing as `postgres`.

PR #6676, opened 2026-08-07
RFC: pluggable training client for AsyncGRPOTrainer

kashif:
Opening this as an RFC to get feedback before going further.

`AsyncGRPOTrainer` already takes `rollout_worker` and `weight_transfer` as protocols passed to the constructor, so rollout and weight sync are both pluggable. The model compute is the one leg that is not. This adds the third one:

```python
rollout_worker:  RolloutWorkerProtocol  | None = None
weight_transfer: WeightTransferProtocol | None = None
training_client: TrainingClientProtocol | None = None  # new
```

Default is `LocalTrainingClient`, which runs the model in process, so nothing changes unless you pass a client.

### How it works

The backend runs the forward and returns per token log probs as a leaf tensor. The trainer builds its loss on that tensor exactly as before, and a hook hands `d(loss)/d(log_probs)` back to the backend, which backpropagates `sum(grad * log_probs)` into the model. That sum is a first order surrogate whose parameter gradients equal those of the real loss, so the loss never crosses the boundary and a backend does not need to know what GRPO is. TRL keeps its loss variants, a backend only has to move tensors.

The seam is one line in `compute_loss`:

```python
log_probs.register_hook(self.training_client.backward)
```

I used a hook rather than calling `backward` explicitly in `training_step` because the hook fires inside `accelerator.backward`, so the model backward stays within accelerate's `no_sync` context. Calling it afterwards would trigger a DDP all reduce on every accumulation micro step.

### Verification

`tests/experimental/test_api_training_client.py` runs the real GRPO loss through `patch_chunked_lm_head` and checks that parameter gradients match the in process path at `rtol=0, atol=0`. It covers both the single forward case (a co located backend keeps the graph) and the two forward case (a remote backend replays the forward). CPU only, runs in a few seconds.

### Not in this PR

* `optim_step`. `Trainer` already accepts a user supplied optimizer through `optimizers=`, so a backend that owns the parameters supplies its own optimizer and TRL adds nothing.
* The skipped step signal. `accelerator.optimizer_step_was_skipped` reads a local overflow flag, so a skip on the backend side is invisible and the LR schedule could drift. It is inline in `_inner_training_loop` with no override point.
* Optimizer and scheduler checkpoint state.
* MoE `aux_loss` is handled in code (the coefficient is passed to `forward` and the backend adds it to its own backward) but has no test, since that needs a tiny MoE model.
* A fully remote client. HF and accelerate still want a local model to prepare, zero_grad and stream from in `_sync_weight`, so something has to stand in for it. That may belong in a loop owning trainer instead of here.

### Questions

1. Is the backward hook acceptable, or would you rather see it called explicitly even at the cost of the `no_sync` issue?
2. Should the protocol live in `trl/experimental/api/` so other trainers can reuse it, or next to the other two protocols inside `async_grpo_trainer.py`?
3. Is there interest in a wire contract and a reference server on top of this, so backends can plug in over HTTP instead of in Python? The shape I have in mind is one `HTTPTrainingClient` per protocol, not one client per provider, so vendor adapters stay in vendor repos.

---

> [!NOTE]
> **Medium Risk**
> Refactors `AsyncGRPOTrainer.compute_loss` on the hot training path; default client preserves behavior but custom backends must correctly handle aux loss, DDP scaling, and rank topology.
> 
> **Overview**
> Adds a third extension point for [`AsyncGRPOTrainer`]: optional `training_client` implementing [`TrainingClientProtocol`], alongside existing `rollout_worker` and `weight_transfer`. Default [`LocalTrainingClient`] keeps in-process forward/backward unchanged.
> 
> The trainer now builds GRPO as a `loss_fn` closure over shifted masks/advantages/old log probs and delegates model compute to `training_client.forward_backward(...)`, which returns [`ForwardBackwardOutput`] (scalar loss, detached log probs/entropy, optional MoE `aux_loss`). Off-process backends can evaluate `loss_fn` locally and backprop a log-prob surrogate (or hook-based remote backward) while TRL keeps the algorithm and metrics.
> 
> New module `trl/experimental/api/` exports the protocol, output dataclass, and local client. Docs add an **External training compute** section and autodoc entries. Tests verify bit-identical gradients vs inline loss (local + remote-style surrogate) and that per-sequence clip metrics still use the shifted completion mask.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 91c50f21483ea10aa1ba62dc5a0ec8cb08153047. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/api/training_client.py
```diff
@@ -0,0 +1,194 @@
...
+    from computing the loss inline: gradients are bit-identical to the pre-interface behavior.
+
+    Examples:
+
+    ```python
+    >>> client = LocalTrainingClient()
+    >>> output = client.forward(model, input_ids, position_ids, completion_mask)
+    >>> loss = my_loss(output.log_probs)
+    >>> loss.backward()  # populates output.log_probs.grad
+    >>> client.backward(output.log_probs.grad)
```
sfc-gh-truwase: @kashif, thanks for this snippet. It provides greater clarity for design discussions. 

One difference that I see from ArcticPlatform design is that fwd/bwd is folded into a single call which specifies the desired loss function via string key. The loss key will be used to access ArcticPlatform's  [registry](https://github.com/Snowflake-AI-Research/Arctic-Platform/blob/aed5916cdff6b56815f3c6fb187e234d31e6e324/arctic_platform/rl/processors/grpo.py#L300) of loss function variants.  So, the example snippet would modified to below (after adding "grpo_loss" to the registry). 

```python
    >>> output = client.forward_backward(..., "grpo_loss") 
```

What do you think?

sfc-gh-truwase: @sfc-gh-sbekman, @sfc-gh-kganesan, FYI

sfc-gh-truwase: @kashif, is your approach same as Tinker custom loss (i.e., `forward_backward_custom`)?

kashif: Yes, same as Tinker's `forward_backward_custom`. TRL computes the loss, the backend gets `d(loss)/d(logprobs)` and backprops `sum(grad * logprobs)`. Same parameter gradients, verified bit exact against the in process path.

On folding it into one call, I think your API already fits as is. An adapter would call `fwd_no_grad` from `forward` and `fwd_bwd` from `backward`, so the fused call still happens and the split is only on the TRL side. Two round trips are unavoidable either way, since the weights depend on the logprobs.

What I would rather avoid is naming the loss on the wire. `integrations/verl/grpo_loss.py` is 523 lines mirroring verl's loss internals and still only covers `vanilla` out of the 12 in `POLICY_LOSS_REGISTRY`. TRL has 8 loss types and adds them fairly often. One generic entry covers all of them:

```python
loss = (logprobs * weights).sum()
```

Reading `processors/pipeline.py`, it looks like you may not need even that: `_resolve_fn` falls back to a dotted path import, and any `[B, S]` batch tensor gets unpadded automatically, so the weights can just ride along. Is that fallback something I can rely on, or an implementation detail? And for managed deployments, would a generic registry entry be ok? Weighted cross entropy is the same surrogate since your CE is `sum(-logprobs * weights)`, which is how Tinker gets away without adding a loss.

sfc-gh-truwase: @kashif thanks for the clarification and for laying out the design considerations on the ArcticPlatform PR. I am fully aligned.

trl/experimental/api/training_client.py
```diff
@@ -0,0 +1,162 @@
...
+    interface existed. Implement this protocol to run the model somewhere else (another process, another set of GPUs,
+    or a remote service) while the trainer keeps owning the loss.
+
+    Gradient clipping and the optimizer step are deliberately absent. [`~transformers.Trainer`] already accepts an
+    optimizer through its `optimizers` argument, so a backend that owns the parameters supplies one whose `step`
+    reaches them and clips them on the way. That matches how existing backends are built: Tinker carries
+    `grad_clip_norm` in its optimizer params, and Arctic Platform clips inside its `step` call.
+    """
+
+    def forward_backward(
```
AmineDiro, line 83: I think that fwd_bwd could be fully async. I think this mean we can do grad acc a little bit faster. we'd also need to define some __barrier__ operations  like `optimizer_step` where we actively do one step. This means that this function should return a future `Future[ForwardBackwardOutput]`

trl/experimental/api/training_client.py
```diff
@@ -0,0 +1,162 @@
...
+
+        aux_loss = outputs["aux_loss"] if aux_loss_coef else None
+        if aux_loss is not None:
+            loss = loss + aux_loss_coef * aux_loss
+
+        return ForwardBackwardOutput(
+            loss=loss,
+            log_probs=log_probs.detach(),
+            entropy=outputs["entropy"].detach(),
+            aux_loss=aux_loss.detach() if aux_loss is not None else None,
```
AmineDiro, line 170: should we consider  `aux_loss` a part of the loss computed by `loss_fn`  🤔

tests/experimental/test_api_training_client.py
```diff
@@ -0,0 +1,170 @@
...
+    def _client_grads(self, client, loss_scale):
+        self.model.zero_grad(set_to_none=True)
+        outputs = client.forward_backward(
+            self.model,
+            input_ids=self.input_ids,
+            position_ids=self.position_ids,
+            completion_mask=self.completion_mask,
+            loss_fn=self.loss_fn,
+        )
+        (outputs.loss * loss_scale).backward()
```
AmineDiro, line 129: this means we have a copy of the model locally loaded so we can resolve computational graph. Is this something we need for the remote training. In my mind, the goal of the remote backend is being able to train for example large model on remote infra without keeping a local copy of the weight ? I could misunderstand so please dont hesitate to clarify 👍🏼

sfc-gh-truwase: Yes, no local model, TRL is cpu-only. 
https://github.com/Snowflake-AI-Research/Arctic-Platform/pull/84#issuecomment-5397631924

PR #6680, opened 2026-08-08
Document the GRPO reward-function batch layout and group-relative rewards

### LINKED ISSUE #4707: GRPOTrainer: Relative LLM-as-a-judge rankings
### Feature request

Is it currently possible to use LLM-as-a-judge with groupwise relative rankings as in OpenPipe's ART RULER?
https://art.openpipe.ai/fundamentals/ruler

If so would like to request an example, and if not would like to request this feature!

### Motivation

Space of problems the GRPOTrainer becomes useful for expands greatly if LLM-as-a-judge ranking works well, but it seems LLM-as-a-judge performs a lot better providing relative rather than absolute rankings

### Your contribution

Not sure — possibly if guided!

behroozazarkhalili:
## What this does

Documents two things about custom reward functions in GRPO that the docs did not state, and adds an example for the use case in #4707.

`docs/source/grpo_trainer.md` specified the reward-function contract (arguments, return value) but never the shape of the call. Reward functions are invoked **once per batch with every completion at once**, grouped by prompt as `num_generations` consecutive entries, and `num_generations` itself is not among the arguments.

Without that stated, a judge that ranks a prompt's completions against each other (the ART RULER style asked about in #4707) reads as unsupported. It works today: all six existing examples are pointwise, so nothing showed the group structure is already available.

## Verification

Ran a probe against current `main` rather than reasoning from the source:

| Question | Observed |
|---|---|
| Whole batch in one call | yes, 8 completions for `B=2, G=4` |
| Layout G-consecutive per prompt | yes, prompt run lengths `[4, 4]` over 2 distinct prompts |
| `num_generations` reachable from the signature | no, kwargs are `{log_extra, log_metric, trainer_state}` and `TrainerState` has no such field |
| Group-relative reward trains | yes, `reward_std 0.3984` and `frac_reward_zero_std 0`, against `0` and `1` for a constant reward |

The layout is an invariant in three places, not an accident of one run:

1. `RepeatSampler`'s docstring, where `mini_repeat_count=3` gives `[0, 0, 0, 1, 1, 1, 2, 2, 2, ...]`
2. `grpo_trainer.py:1102`, which passes `mini_repeat_count=self.num_generations`
3. `grpo_trainer.py:2417`, where `rewards.view(-1, num_generations)` depends on it to compute the advantage

The added snippet was executed against a stub judge for `num_generations` 1 through 4, and each group's best completion scores `1.0` independently of the other groups.

## Note for reviewers

Neither automated gate covers this file: the ruff hooks are `types_or: [python, pyi]`, and `doc-builder style trl tests docs/source` filters to `.mdx` and `.py` (`style_doc.py:537`) while `docs/source` holds 63 `.md` files and no `.mdx`. I confirmed that with a negative control, so this change was checked by parsing and executing the snippet instead.

Closes #4707

cc @qgallouedec, whose reply on the issue is the basis for the example, and @albertvillanova

---

> [!NOTE]
> **Low Risk**
> Markdown-only documentation; no trainer or reward API behavior changes.
> 
> **Overview**
> **Documentation-only** update to `docs/source/grpo_trainer.md` for custom GRPO reward functions.
> 
> Adds a **Batch layout** contract item: the trainer calls each reward function **once per batch** with completions ordered as **`num_generations` consecutive entries per prompt**, and **`num_generations` is not passed** into the function (callers must close over it, e.g. from `training_args.num_generations`).
> 
> Adds **Example 7** for **group-relative / LLM-as-judge rankings**: split the batch into prompt groups, rank within each group, and map ranks to spread rewards so group advantage is non-zero—contrasted with pointwise grading that can drive **`frac_reward_zero_std`** when every completion in a group gets the same score.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e934f46bd4a925c76ee79007ce41f521c75b25cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/grpo_trainer.md
+++ b/docs/source/grpo_trainer.md
@@ -431,6 +431,8 @@ Reward functions can be either synchronous Python callables or asynchronous `asy
 
 2. **Return value**: The function must return a list of floats. Each float represents the reward corresponding to a single completion.
 
+3. **Batch layout**: The function is called once per batch with every completion at once, not once per completion. The completions are grouped by prompt: the first `num_generations` entries are the generations of the first prompt, the next `num_generations` entries those of the second, and so on. A reward function can therefore score a completion relative to the other generations of the same prompt. See Example 7. Note that `num_generations` is not passed as an argument, so a group-aware reward function has to capture it.
+
 #### Example 1: Reward longer completions
 
 Below is an example of a reward function for a standard format that rewards longer completions:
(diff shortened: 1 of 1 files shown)


qgallouedec:
closing, see https://github.com/huggingface/trl/issues/4707#issuecomment-5344266897

behroozazarkhalili:
Understood, and agreed on the technique: it works with the existing API and needs no core change.

Fair on the size too. Looking at it again, the PR bundled two separate things, a worked ranking example and a one-line statement of the calling convention, and the example is the part that did not earn 26 lines of docs.

The one piece I would still offer: the reward-function contract in `grpo_trainer.md` has two numbered items today, "Input arguments" and "Return value". Item 1 lists which kwargs arrive, but nothing on the page says the function is called once per batch with every completion, or that completions come grouped by prompt in blocks of `num_generations`. A reader can reasonably assume per-completion calls, which is what made this a question in the first place, and the answer now lives in a closed issue.

That would be one bullet added to the existing list. No example, no new section. Happy to send it if useful, and equally happy to leave it as is.

PR #6705, opened 2026-08-11
Add AsyncDistillationTrainer

kashif:
Async on-policy distillation, architected like `AsyncGRPOTrainer`: a background rollout worker generates the student's own completions and scores them against a teacher served over HTTP, so generation and training overlap instead of alternating. The teacher is never loaded locally, just a vLLM server URL.

Also supports MOPD (multi-teacher on-policy distillation) — pass more than one entry in `teacher_server_urls` and route each sample to a teacher via a `teacher_id` column (e.g. a math teacher and a code teacher, each served independently).

This went through several rounds of GPU validation in a staging fork before landing here, and the multi-teacher routing design was cross-checked against how NeMo RL, verl, and miles do the same kind of per-sample teacher routing.

Adds:
- `trl/experimental/async_distillation/` (config, trainer, rollout worker, vLLM client, weight transfer)
- `tests/experimental/test_async_distillation_trainer.py`
- `examples/scripts/async_distillation.py` (single-teacher) and `async_distillation_mopd.py` (math+code MOPD demo)
- `docs/source/async_distillation_trainer.md` + toctree entry, one-line cross-ref from `distillation_trainer.md`

`make precommit` passes. Test suite passes (26 passed / 8 GPU-gated skipped without a GPU); also ran the GPU-gated tests via `srun --gres=gpu:1` — one pre-existing failure reproduces there, but it's unrelated to this PR (it also reproduces on unmodified `async_grpo` tests on the same node, looks like a tiny-test-fixture/environment issue rather than anything this PR changes).

Happy to open a tracking issue first if that's the preferred process for a new experimental trainer — let me know.

---

> [!NOTE]
> **High Risk**
> Large new distributed training path (async queues, off-policy staleness, NCCL weight sync, sparse teacher logprobs) with strict vLLM/transformers/FSDP2 assumptions; misconfiguration (tokenizer mismatch, teacher flags) can silently skew training.
> 
> **Overview**
> Introduces **`AsyncDistillationTrainer`**, an experimental trainer that overlaps student rollout generation with gradient updates (same async pattern as `AsyncGRPOTrainer`). A background worker samples completions from the student’s vLLM server, teacher-forces them through one or more **HTTP vLLM teacher** endpoints (`prompt_logprobs`, sparse `teacher_top_k`), and enqueues scored samples; the main loop minimizes **generalized JSD** against that sparse teacher signal, with **NCCL weight sync** back to the student server on a schedule and **staleness** / queue backpressure controls.
> 
> **Multi-teacher (MOPD)** is supported via `teacher_server_urls` plus per-row `teacher_id` routing (no ensembling); docs and `async_distillation_mopd.py` demonstrate math/code experts. Loss handling mirrors server distillation: full top-k support at `beta=0`, narrowed top-1 + realized token support for `beta>0`, optional tail bucket, and **chunked `lm_head` + padding-free packing** with Σ L² row balancing and optional `token_budget` batching.
> 
> Also adds **`AsyncDistillationConfig`**, rollout worker / vLLM client / weight-transfer modules, extensive **metrics** (rollout, queue, batch, perf/MFU), **~780 lines of tests** (collator, MOPD routing, JSD math, stubbed train), GSM8K and MOPD **example scripts**, a long **user guide**, toctree entry, paper-index MOPD blurb, and a cross-link from the sync distillation doc.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 361edd3645012c41f5636fa2a785b5c8e3052f33. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_distillation/async_rollout_worker.py
```diff
@@ -0,0 +1,632 @@
...
+        A single configured teacher scores every row regardless of any `teacher_id` column (plain on-policy
+        distillation). With multiple teachers, `row["teacher_id"]` selects among them; a missing or unmapped
+        `teacher_id` is a configuration error (raises), not a silent misroute to the wrong teacher. Returns
+        `(teacher_id, server_url)`; the resolved id is carried on the `RolloutSample` so `compute_loss` can break down
+        `jsd`/`entropy`/`teacher_entropy` per teacher.
+        """
+        if len(self.teacher_server_urls) == 1:
+            ((teacher_id, url),) = self.teacher_server_urls.items()
+            return teacher_id, url
+        teacher_id = row.get("teacher_id")
```
AmineDiro, line 489: If I understand this couples the mapping teacher_id->vllm_server_url to the dataset having a specific column `teacher_id` in the format ?
I wonder if we can check/enforce this somehow at runtime like checking that the dataset has this columns and the unique values map correctly to the teacher 🤔  ? wdyt ? 

Not a strong requirement

trl/experimental/async_distillation/async_rollout_worker.py
```diff
@@ -0,0 +1,632 @@
...
+        """
+        payload = {
+            "sequences": [prompt_ids + completion_ids],
+            "prompt_lengths": [len(prompt_ids)],
+            "top_logprobs": self.teacher_top_k,
+            "temperature": self.teacher_temperature,
+            "response_format": "json",
+        }
+        output = await _retry_on_http_error(
+            lambda: self._post(teacher_server_url, "/get_sequence_logprobs/", payload, self.request_timeout),
```
AmineDiro: Q: I was unaware of `/get_sequence_logprobs` endpoint. Is this in stock vllm ? :o

trl/experimental/async_distillation/async_distillation_trainer.py
```diff
@@ -0,0 +1,1257 @@
...
+        if n_valid == 0:
+            # Whole micro-batch has no usable teacher signal. Keep the loss connected to the autograd graph through
+            # every trainable parameter (backbone via `hidden_states`, `lm_head` via `lm_head_weight`) so
+            # `.backward()` succeeds and DDP/FSDP gradient sync doesn't hang on a "missing" parameter's grad.
+            loss = (hidden_states.sum() + lm_head_weight.sum()) * 0.0
+            if lm_head_bias is not None:
+                loss = loss + lm_head_bias.sum() * 0.0
+        else:
+            chunk_size = _CHUNKED_LM_HEAD_CHUNK_SIZE
+            for start in range(0, n_valid, chunk_size):
```
AmineDiro: maybe move to a separate chunking function ?

AmineDiro (commented):
Implementation looks good to me.  Well done @kashif  👏🏼 

The only design question is about separating the RolloutWorker in a separate process. We've done this in `AsyncGRPOTrainer`  in [PR](https://github.com/huggingface/trl/pull/5749) mainly because recursive_parse / accuracy_reward were locking the GIL. In this AsyncDistillationTrainer I dont see anything similar to that so maybe a simple threaded Worker is the simpler + remove serialization/deserialization cost ?

trl/experimental/async_distillation/async_rollout_worker.py
kashif: Yep, with more than one teacher each row needs a `teacher_id` matching a key in `teacher_server_urls`. I kept it the same as AsyncGRPO, which routes multi-env runs off an `environment` column and checks it per row in the worker rather than validating the dataset upfront. A missing or unknown id raises instead of silently picking a teacher. I did align the error with the GRPO one (ValueError, same wording), so it reads the same in both.

trl/experimental/async_distillation/async_rollout_worker.py
kashif: Nope, that one is ours, not stock vllm. It only exists on `trl vllm-serve` (`trl/scripts/vllm_serve.py`), wrapping vllm `prompt_logprobs` for teacher forcing. Same endpoint sdft/sdpo/server_distillation already use.

AmineDiro: Oh ok I see, should we update the `examples/scripts/async_distillation.py` and `examples/scripts/async_distillation_mopd.py` docs to use `trl vllm-serve`. Also this command might be depereciated in the future (cc. @qgallouedec ) so I wonder if there is a stock vllm way to get the same info ?

qgallouedec: Yes, `trl vllm-serve` is going away. In #6765 it becomes a thin deprecated wrapper that just translates its flags into a plain `vllm serve` command, and `/get_sequence_logprobs` is replaced by vLLM's stock `prompt_logprobs`. So no need to update the examples here: this PR already does it the stock way, `/v1/completions` with `prompt_logprobs=teacher_top_k` against a plain `vllm serve`, which is what the sync trainers will move to.

`get_sequence_logprobs` only exists today for the sync server trainers and disappears with #6765.

qgallouedec (approved):
lgtm! let's try to keep this as aligned as possible with asynchronous grpo from now. Great work @kashif

AmineDiro (approved):

PR #6710, opened 2026-08-11
⚡️ Guard the RLOO completions table against prompts with no image

vineethsaivs:
# What does this PR do?

`RLOOTrainer._generate_and_score_completions` builds the per-sample image entry as

```python
images = [[example.get("image")] if example.get("image") is not None else None for example in inputs]
```

so a batch that mixes multimodal and text-only prompts puts `None` into `_logs["images"]`. `RLOOTrainer.log` then iterates it unguarded while building the completions table:

```python
for image_list in self._logs["images"]:
    images.append([logging_backend.Image(image) for image in image_list])
```

```
TypeError: 'NoneType' object is not iterable
```

`GRPOTrainer.log` already handles it and appends an empty list. The two blocks are otherwise identical, so this is the same fix propagated; they are byte identical after this change.

Reached whenever `log_completions=True` and `wandb` or `trackio` is in `report_to`, on any RLOO run whose dataset does not give every prompt an image.

`AGENTS.md` names this exact class: "A fix in GRPO often implies the same fix in RLOO, and vice versa. Not propagating a change is a bug."

Fixes # (issue)

## Before submitting

## AI writing disclosure

## Tests

`test_log_completions_with_some_prompts_missing_images` in `tests/test_rloo_trainer.py` builds a trainer the same way the neighbouring VLM logging tests do, sets `_logs["images"]` to one real image and one `None`, stands in a fake backend for wandb, and asserts the table comes out with image counts `[1, 0]`.

Being straight about what I could and could not run: I do not have a machine that can run TRL's trainer tests, so the new test has not been executed and CI is the check. What I did verify locally is the failure itself, by running both blocks over `[["img"], None, ["img", "img"], None]`: RLOO's raises `TypeError: 'NoneType' object is not iterable`, GRPO's returns lengths `[1, 0, 2, 0]`. I also confirmed by AST-comparison that `grpo_trainer.py` and `rloo_trainer.py` are the only two files carrying this block, that only RLOO's is unguarded, and that after this change the two are identical. All of that was against `main` at 24db2a8.

If you would rather have the test exercise a real mixed dataset end to end than stand in a backend, say so and I will rewrite it that way.

---

> [!NOTE]
> **Low Risk**
> Narrow logging-path fix aligned with existing GRPO behavior; no training or reward logic changes.
> 
> **Overview**
> Fixes a crash in **`RLOOTrainer.log`** when **`log_completions=True`** and wandb/trackio logging is enabled on batches that mix image and text-only prompts.
> 
> Per-sample entries in **`_logs["images"]`** can be **`None`** for prompts without an image; the completions table builder now skips those like **`GRPOTrainer`** already does, appending an empty image list instead of iterating **`None`**.
> 
> Adds **`test_log_completions_with_some_prompts_missing_images`** to assert the logged table gets image counts **`[1, 0]`** for one image and one **`None`** entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18edfa4dcbd1fdb46df3f45a043ca1f8c86c9339. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_rloo_trainer.py
+++ b/tests/test_rloo_trainer.py
@@ -17,9 +17,9 @@ from unittest.mock import patch
 import pytest
 import torch
 import transformers
-from PIL import Image
 from datasets import DatasetDict, IterableDatasetDict, load_dataset
 from packaging.version import Version
+from PIL import Image
 from transformers import (
     AutoModelForCausalLM,
     AutoModelForImageTextToText,
@@ -2141,8 +2141,9 @@ class TestRLOOTrainerVLM(TrlTestCase):
         trainer._logs["images"] = [[Image.new("RGB", (4, 4))], None]  # second prompt has no image
 
         with (
-            patch.object(type(trainer.args), "report_to", ["wandb"]),
-            patch("trl.trainer.rloo_trainer.wandb", _Backend),
+            patch.object(trainer.args, "report_to", ["wandb"]),
+            # wandb is imported only when it is installed, so the name may not exist
+            patch("trl.trainer.rloo_trainer.wandb", _Backend, create=True),
             patch("trl.trainer.rloo_trainer.is_rich_available", return_value=False),
         ):
             _Backend.run = object()
--- a/tests/test_rloo_trainer.py
+++ b/tests/test_rloo_trainer.py
@@ -17,6 +17,7 @@ from unittest.mock import patch
 import pytest
 import torch
 import transformers
+from PIL import Image
 from datasets import DatasetDict, IterableDatasetDict, load_dataset
 from packaging.version import Version
 from transformers import (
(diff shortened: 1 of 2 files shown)


vineethsaivs:
Bugbot is right, and the test was not covering the fix at all. Corrected in `18edfa4`.

`report_to` is a dataclass instance field that `__post_init__` has already rewritten from `"none"` to `[]` by the time the trainer is built, so patching the class left the instance value winning:

```python
with patch.object(type(trainer.args), "report_to", ["wandb"]):
    trainer.args.report_to   # []   <- falsy, the wandb branch never runs

with patch.object(trainer.args, "report_to", ["wandb"]):
    trainer.args.report_to   # ['wandb']
```

Running it turned up a second problem in the same block that has nothing to do with the first. `wandb` is imported only under `if is_wandb_available()`, so on an environment without wandb installed `patch("trl.trainer.rloo_trainer.wandb", _Backend)` raises `AttributeError: ... does not have the attribute 'wandb'` before the `report_to` issue can even show itself. That one hides in CI, where wandb is present. `create=True` makes the test independent of whether it is installed, which seems right here given the test is about building the table rather than about wandb.

Both fixed, and I have now actually run it against the tree at both revisions rather than assuming:

```
                                        unpatched rloo_trainer.py   with the fix
test_log_completions_..._missing_images  FAILED                      passed
                                         TypeError: 'NoneType' object
                                         is not iterable
                                         rloo_trainer.py:1811
```

`python -m pytest tests/test_rloo_trainer.py -k test_log_completions_with_some_prompts_missing_images` on CPU, transformers 5.14.1, torch 2.11.0. `ruff check` and `ruff format --check` are clean on both changed files, and clean on the unmodified tree as a control.

Worth flagging that the only checks that ran on this PR are `check-pr-template` and Bugbot itself, so nothing in CI would have caught a test that cannot reach its own assertion.

qgallouedec:

PR #6713, opened 2026-08-12
Add loss_type support to AsyncGRPOTrainer

kashif:
Ports the GRPO loss types (grpo, bnpo, dr_grpo, dapo, cispo, sapo, luspo, vespo) and importance_sampling_level to AsyncGRPOTrainer, matching GRPOTrainer's formulas. Default stays dapo-equivalent, so existing configs keep the same behavior unless loss_type is set explicitly.

AsyncGRPOTrainer packs all samples for a rank into a single padding-free row instead of GRPOTrainer's (B, T) batch, so the per-sequence pieces (grpo/sapo/luspo normalizers, VESPO gamma weights, sequence-level importance sampling) needed a segment-sum over a derived per-token sequence id instead of a plain dim=-1 reduction. Verified the segment-based formulas against GRPOTrainer's literal batched ones on a synthetic packed example, numerically identical across all loss types and both importance_sampling_level settings.

Tests are added (test_train_loss_types, test_train_with_multiple_samples_per_row) but I haven't been able to run them on this machine yet, my local vllm/kernels setup has some unrelated issues blocking the async_grpo test class here. Will verify and follow up.

---

> [!NOTE]
> **Medium Risk**
> Changes the core RL loss path and distributed normalization for AsyncGRPO, though default `dapo` preserves prior behavior; incorrect segment/global counts could skew gradients for non-default loss types.
> 
> **Overview**
> **AsyncGRPO** now supports the same **`loss_type`** options as **`GRPOTrainer`** (`grpo`, `bnpo`, `dr_grpo`, `dapo`, `cispo`, `sapo`, `luspo`, `vespo`) plus **`importance_sampling_level`** (`token` vs `sequence`). **`AsyncGRPOConfig`** gains the matching knobs (`delta`, SAPO/VESPO hyperparameters, docs). Default **`loss_type`** remains **`dapo`**, so existing runs should behave the same unless you opt in.
> 
> Because async training packs many sequences into one padding-free row per rank, per-sequence math uses **`_segment_sum`** over a **`sequence_id`** derived from **`position_ids`**, and the collator adds **`global_n_seqs`** so sequence-normalized losses stay correct under DDP when ranks hold different sequence counts. **`compute_loss`** branches on loss type (CISPO clip-on-ratio, SAPO soft gate, VESPO **`get_gamma_weights`**, LUSPO segment aggregation, etc.) and logs loss-specific metrics like **`GRPOTrainer`**.
> 
> Tests **`test_train`** are parametrized over all loss types (with sequence IS for LUSPO) and add **`test_train_with_multiple_samples_per_row`** (`fork_k=2`) to exercise multi-sequence rows; training loss must be finite and parameters must update.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f0fa007e491e704d945d0ab88e20a98f45e38038. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -478,6 +478,13 @@ def torch_call(self, examples: list[Any]) -> dict[str, Any]:
         global_n_tokens = sum(sum(example["completion_mask"]) for example in all_examples)
         global_n_tokens = torch.full((self.num_processes,), float(global_n_tokens), dtype=torch.float32)
 
+        # Total packed sequences across all rows in the full batch. `TokenBudgetBatcher`/`FixedCountBatcher` balance
+        # rows by Σ Lᵢ² (compute cost), not by sequence count, so ranks can end up with different numbers of
+        # sequences per row; the "grpo"/"sapo"/"dr_grpo"/"luspo" loss types need this global count (see
+        # `compute_loss`) instead of each rank's local count to stay correctly normalized under DDP/FSDP gradient
+        # averaging.
```
AmineDiro:
```suggestion

```

qgallouedec:
Thks @kashif, i'm going to park this one for now

The port is good, I checked it against `GRPOTrainer` line by line and it's faithful. That's not the issue. The issue is nobody's asking for it, and AsyncGRPO is deliberately the minimal trainer for now.

Because we pack into a single padding-free row, this can't reuse GRPO's formulas: every per-sequence reduction is re-derived as a segment sum. So every future change to a GRPO loss has to be re-derived here too, with no user to tell us when we get it wrong. Permanent cost, zero demand.

Let's revive it when an internal experiment needs a specific loss type, or someone asks for one. And then just that one, not all eight.

Sorry for the delay, you shouldn't have waited three weeks for this

Issue #6711, opened 2026-08-12
CI fails for slow tests: RuntimeError: min(): Expected reduction dim to be specified for input.numel() == 0


albertvillanova:
CI fails for slow tests: https://github.com/huggingface/trl/actions/runs/31582787840/job/94069521759
> RuntimeError: min(): Expected reduction dim to be specified for input.numel() == 0. Specify the reduction dim with the 'dim' argument.

```python
FAILED tests/test_grpo_trainer.py::TestGRPOTrainerSlow::test_train_with_transformers_continuous_batching[trl-internal-testing/tiny-LlamaForCausalLM-3.2] - RuntimeError: min(): Expected reduction dim to be specified for input.numel() == 0. Specify the reduction dim with the 'dim' argument.
FAILED tests/test_grpo_trainer.py::TestGRPOTrainerSlow::test_train_with_transformers_continuous_batching[trl-internal-testing/tiny-MistralForCausalLM-0.2] - RuntimeError: min(): Expected reduction dim to be specified for input.numel() == 0. Specify the reduction dim with the 'dim' argument.
= 2 failed
```

Traceback:
```python
    def test_train_with_transformers_continuous_batching(self, model_name):
        ...
>       trainer.train()

tests/test_grpo_trainer.py:4474:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.venv/lib/python3.11/site-packages/transformers/trainer.py:1449: in train
    return inner_training_loop(
.venv/lib/python3.11/site-packages/transformers/trainer.py:1531: in _inner_training_loop
    self._run_epoch(
.venv/lib/python3.11/site-packages/transformers/trainer.py:1759: in _run_epoch
    tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
trl/trainer/grpo_trainer.py:1558: in training_step
    output = super().training_step(model, inputs, num_items_in_batch)
.venv/lib/python3.11/site-packages/transformers/trainer.py:1925: in training_step
    inputs = self._prepare_inputs(inputs)
trl/extras/profiling.py:211: in wrapper
    return func(self, *args, **kwargs)
trl/trainer/grpo_trainer.py:1587: in _prepare_inputs
    generation_batch = self._generate_and_score_completions(generation_batch)
trl/trainer/grpo_trainer.py:2441: in _generate_and_score_completions
[...]

PR #6714, opened 2026-08-12
Raise a clear error when generation returns no completions

albertvillanova:
### What does this PR do?

When a generation backend returns no completions at all, the completion-length metric logging crashes with a cryptic error:

> RuntimeError: min(): Expected reduction dim to be specified for input.numel() == 0. Specify the reduction dim with the 'dim' argument.

(`mean()` tolerates an empty tensor and returns `nan`, but `min()`/`max()` raise.) This hides the real problem, which is that generation produced nothing.

This PR adds a guard right after the completion lengths are gathered: if there are no completions, it raises a clear, actionable `RuntimeError` instead. The same block is duplicated across trainers, so the guard is applied consistently to all of them.

Context: surfaced by #6711 (transformers 5.15.0 continuous batching crashing on a pre-Ampere CI GPU, see huggingface/transformers#47926), where the generation thread died and returned no completions.

### Changes

Add an empty-completions guard to the metric logging of `GRPOTrainer`, `RLOOTrainer`, and the experimental `DistillationTrainer`, `SDFTTrainer`, `SDPOTrainer`, and `SSDTrainer`.

---

> [!NOTE]
> **Low Risk**
> Diagnostic-only change on the metrics path; no changes to loss, generation, or distributed training behavior beyond earlier failure when no completions exist.
> 
> **Overview**
> When the generation backend returns **zero completions** across ranks, completion-length metric logging used to fail with a cryptic PyTorch `min()` error on an empty tensor. This PR adds the same **empty-tensor guard** immediately after gathered completion lengths in every affected trainer.
> 
> **`GRPOTrainer`**, **`RLOOTrainer`**, and **`DistillationTrainer`** now raise an explicit `RuntimeError` pointing users to generation logs before computing min/max length metrics. The experimental **`SDFTTrainer`**, **`SDPOTrainer`**, and **`SSDTrainer`** get the same check in their completion-metric paths (SSD uses `agg_lengths` from the completion mask).
> 
> No training logic or generation behavior changes—only **fail-fast diagnostics** when rollouts are missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 72a01a868e7c30d90c7752c705ed14e0746b8a43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (approved):
thanks, this is better indeed

PR #6715, opened 2026-08-12
Rework AsyncGRPO metrics metrics and docs

AmineDiro:
# What does this PR do?
Async GRPO has more places to be slow than sync GRPO: generation, scoring, the queue between the worker and the trainer, weight sync. 

The current metric set can't tell you which one is the problem. This PR restructures the metrics around *where* a number is measured, fixes the reductions, and adds a "Logged metrics" section to the docs explaining what each one means and how to read them together. A rollout passes through several stages before it becomes a gradient: 

```
Dataset row = PROMPT (message list + reward kwargs)
 └─ GROUP                  (1 prompt, `num_generations` rollouts, 1 advantage baseline)
     └─ ROLLOUT            (1 conversation, keyed by `rollout_id`)
         ├─ TURN           (1 vLLM /v1/completions call + tool messages fed back)
         │   └─ TurnRecord (prompt_ids, output_ids, output_log_probs)
         └─ reconcile      (`_chain_to_sequences` classifies drift per turn: CLEAN/REALIGN/FORK)
             └─ SEQUENCE   (≥1 per rollout; new one per fork; dropped if no trained token)
                 └─ SAMPLE (Sequence + group advantage + reward + metrics)

════════════ Trainer process boundary: `rollout_buffer` (mp.Queue) ════════════

                 └─ SAMPLE (Pulled 1 at a time; dropped if staleness > `max_staleness`)
                     └─ ROW            (Planner assigns it to one of `dp` rows, Σ Lᵢ²-balanced)
                         └─ MICRO-BATCH (`dp` rows, one per rank)
                             └─ PACKED ROW (1 concat sequence, `position_ids` reset per sample)
                                 └─ FORWARD (`compute_loss`, bs=1, inter-rank padding stripped)
                     └─ OPTIMIZER STEP (from `grad_accum` micro-batches)
```

So I added more metrics based on these entities: 

| namespace      | entity it counts                                                                       |
| -------------- | -------------------------------------------------------------------------------------- |
| `rollout/`     | represents one _full_ conversation: its turns, its forks, how long it took to generate |
| `completions/` | what the model (vLLM) generated for one prompt                                         |
| `tools/`       | tool calls metrics                                                                     |
| `sample/`      | represents _one_ training sample, as it arrives in the queue                           |
| `batch/`       | one micro-batch or one optimizer step (built from one or multiple samples)             |
| `perf/`        | measured seconds and FLOPs                                                             |

### Renames

Anyone with a dashboard on this trainer will need to update it:

| before | after |
| --- | --- |
| `training_tok/s` | `perf/forwarded_tok_s_fwd_bwd`, `perf/forwarded_tok_s_wall_clock`, `perf/trained_tok_s_wall_clock` |
| `step_time` | `perf/fwd_bwd_s` |
| `iteration_time_s` | `perf/step_s` |
| `forward_time_s` | `perf/fwd_s` |
| `weight_sync_time_s` | `perf/weight_sync_s` (+ `_pause_s`, `_barrier_s`, `_transfer_s`) |
| `queue_wait_time_s` | split into `perf/rollout_wait_s` and `sample/time_in_queue_s` |
| `generation_tok_per_s` | `rollout/generated_tok_s` (windowed, not cumulative) |
| `scoring_time_ms`, `wait_scoring_ms` | `rollout/score_s`, `rollout/score_wait_s` |
| `buffer_qsize` | `sample/rollout_queue_size` |
| `train_seq_len` | dropped — superseded by `sample/forwarded_tokens_mean` and `batch/row_tokens_mean` |

`ratio`, `kl`, `entropy`, `clip_ratio/*`, `aux_loss`, `completions/mean_length`, `reward`, `reward_std`, `rewards/<func>`, `tools/call_frequency` and `tools/failure_frequency` keep their names.

`queue_wait_time_s` is the one worth calling out: it was used for two opposite things — the trainer blocked because the queue was empty, and a sample's own time sitting in it. Those diagnose opposite bottlenecks, so they are now two metrics.

---

> [!NOTE]
> **Medium Risk**
> Changes affect training-loop logging and the rollout worker protocol (`metrics_queue`), so custom workers and existing dashboards must adapt; training math is largely unchanged aside from observability and collator batch fields.
> 
> **Overview**
> **Reworks Async GRPO logging** around pipeline stages (`rollout/`, `completions/`, `tools/`, `sample/`, `batch/`, `perf/`) so bottlenecks (generation, scoring, queue, trainer, weight sync) are diagnosable from dashboards.
> 
> The rollout worker now pushes metrics through a new **`metrics_queue`** on `RolloutWorkerProtocol` (rollout timing, fork/drift tallies, vLLM retries, tool latency, backpressure, windowed `rollout/generated_tok_s`). The trainer drains that queue in `log()`, aggregates sample/batch stats in the collator and `RolloutQueueDataset` (no more per-sample metrics tensors broadcast across ranks), and flushes per-optimizer-step **`perf/*`** accounting via `_log_step_metrics` and an optimizer-timing callback.
> 
> **`_reduce_metric`** fixes broken throughput: rates use Σnum/Σden tuples instead of averaging ratios. Several legacy keys are renamed or split (e.g. `training_tok/s` → `perf/forwarded_tok_s_*` / `perf/trained_tok_s_wall_clock`, `queue_wait_time_s` → `perf/rollout_wait_s` vs `sample/time_in_queue_s`). **`docs/source/async_grpo_trainer.md`** adds a large “Logged metrics” section (entity diagram, queue interpretation, metric tables).
> 
> Tests cover protocol compliance, reduction rules, worker metric push, collator metric keys, and drift tally behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17adf18e20a32f72350707aff86009aab1fc3045. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kashif:
Looked into the last open bugbot item (harness timeout metric never recorded). It's real but pretty narrow.

self.tools only gets populated from the tools= / environment_factory= kwargs in AsyncRolloutWorker.__init__, and HarnessRolloutWorker doesn't override __init__, it just swaps _loop_cls. So a harness run that relies purely on the OpenEnv session for tool calls (no local tools/environment_factory passed) leaves self.tools empty, even though the harness loop can genuinely go multi turn and hit a timeout. The if self.tools: gate at async_rollout_worker.py:804 was written for the built in env path where empty tools really does mean "loop always ends turn one", that assumption doesn't hold for harness sessions since their tool calling is tracked OpenEnv side, not through self.tools.

Net effect is just that tools/loop_exhausted_frac silently never shows up for pure harness setups, no crash, no impact on training. Given everything else got fixed already I'd be fine merging this and following up on that one separately, but wanted to flag it explicitly rather than let it quietly disappear.

kashif (approved):
one comment

AmineDiro:
hey thanks @kashif 
> self.tools only gets populated from the tools= / environment_factory= kwargs in AsyncRolloutWorker.**init**, and HarnessRolloutWorker doesn't override **init**, it just swaps _loop_cls. So a harness run that relies purely on the OpenEnv session for tool calls (no local tools/environment_factory passed) leaves self.tools empty, even though the harness loop can genuinely go multi turn and hit a timeout. The if self.tools: gate at async_rollout_worker.py:804 was written for the built in env path where empty tools really does mean "loop always ends turn one", that assumption doesn't hold for harness sessions since their tool calling is tracked OpenEnv side, not through self.tools.

Agreed on all of it, and thanks for the review ! 
I'd rather not spot patch this here. All`tools/*` metrics with the HarnessWorker should probably be handled separatly in a followup PR once the interface with OpenEnv is decided.
  
I'll merge this as-is, and have the harness tool tracking + `loop_exhausted` get their own PR.

PR #6754, opened 2026-08-14
Report slow tests results to Slack based on the job status

albertvillanova:
Stacked on top of #6753.

This PR makes the slow tests workflow report to Slack based on the job status, using the shared `post-slack` action already used by the other test workflows, and reduces `scripts/log_reports.py` to writing the job summary.

### Motivation

The slow tests workflow posts to Slack from `scripts/log_reports.py`, which only sends a message when it finds failed tests in the pytest report log. Any failure that prevents the log from being written, a failed installation, a container killed by the runner, a job timeout, or a pytest startup error, leaves the report empty, so the script sends nothing and the failure goes unnoticed.

Reporting the job status instead makes the notification independent of what pytest managed to write, and it is what the other test workflows already do, so the workflows now share a single Slack mechanism.

### Solution

Both jobs end with the shared `post-slack` action, passing `${{ job.status }}`, as in `tests.yml`. The report step keeps generating the job summary, and `log_reports.py` no longer builds Slack payloads, so it no longer needs `slack_sdk` nor the `SLACK_API_TOKEN` variable. The script also now reports missing and empty log files in the job summary, which were previously only mentioned in the message it failed to send.

### Changes

- Post the slow tests results to Slack with the shared `post-slack` action, based on the job status
- Remove the Slack payload handling from `scripts/log_reports.py`, keeping the failed tests table and adding the missing and empty log file warnings to the job summary
- Report in the job summary when all tests passed
- Replace the `SLACK_API_TOKEN` environment variable with `CI_SLACK_CHANNEL`, as in the other test workflows
- Stop installing `slack_sdk` in the report steps

### Note for reviewers

The Slack channel now comes from the `CI_PUSH_MAIN_CHANNEL` secret, for consistency with the other test workflows, whereas `scripts/log_reports.py` had `trl-push-ci` hardcoded as a default. Please confirm both point to the same channel.

---

> [!NOTE]
> **Low Risk**
> CI-only changes: notification path and report formatting; no runtime library or test logic changes.
> 
> **Overview**
> Slow tests on push now notify Slack from **job status** (same `post-slack` action as `tests.yml`), so infra/setup/timeout failures still alert even when pytest never writes a log.
> 
> **Workflow:** Both single- and multi-GPU jobs get an `if: always()` **Post to Slack** step; `CI_SLACK_CHANNEL` replaces workflow-level `SLACK_API_TOKEN`. Report steps only install `tabulate` (no `slack_sdk`).
> 
> **`log_reports.py`:** Slack payloads and `slack_sdk` are removed; output is markdown for `$GITHUB_STEP_SUMMARY` only—failed-test tables, ✅ when a log has passes and no failures, and ⚠️ for missing or empty logs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c4502a7d44cfece5b3b05e1f12179c9776dad80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/log_reports.py
+++ b/scripts/log_reports.py
@@ -12,20 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import argparse
 import json
 import logging
 import os
-from datetime import date
 from pathlib import Path
 
 from tabulate import tabulate
 
-
-MAX_LEN_MESSAGE = 2900  # Slack endpoint has a limit of 3001 characters
-
-parser = argparse.ArgumentParser()
-parser.add_argument("--slack_channel_name", default="trl-push-ci")
 
 # Set up logging
 logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
@@ -34,7 +27,6 @@ logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(
 def process_log_file(log):
     failed_tests = []
     passed_tests = []
-    section_num_failed = 0
 
     try:
         with open(log) as f:
@@ -47,7 +39,6 @@ def process_log_file(log):
 
                     if test_name:
                         if outcome == "failed":
-                            section_num_failed += 1
                             failed_tests.append([test_name, duration, log.stem.split("_")[0]])
                         else:
                             passed_tests.append([test_name, duration, log.stem.split("_")[0]])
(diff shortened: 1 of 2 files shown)


scripts/log_reports.py
```diff
@@ -59,111 +50,36 @@ def process_log_file(log):
...
-    return failed_tests, passed_tests, section_num_failed
+    return failed_tests, passed_tests
 
 
-def main(slack_channel_name):
-    group_info = []
-    total_num_failed = 0
-    total_empty_files = []
+def main():
+    print(f"## 🤗 Results of the {os.environ.get('TEST_TYPE', '')} TRL tests.")
```
qgallouedec:
```suggestion
    print(f"## 🤗 Results of the {os.environ["TEST_TYPE"]} TRL tests.")
```

it should always be defined

scripts/log_reports.py
```diff
@@ -59,111 +50,36 @@ def process_log_file(log):
...
+
+        if not failed and not passed:
+            print(f"⚠️ Empty log file `{log}`! Check the GitHub action job.")
+        elif failed:
+            print(f"### ❌ {len(failed)} failed test(s) in `{log}`")
+            failed_table = [test[0].split("::")[:2] + [test[0].split("::")[-1][:30] + ".."] for test in failed]
+            table = tabulate(failed_table, headers=["File", "Class", "Test Name"], tablefmt="grid")
+            print(f"\n```\n{table}\n```\n")
+        else:
+            print(f"### ✅ No failures, all {len(passed)} test(s) passed in `{log}`")
```
qgallouedec: nit: `if failed / elif passed / else empty` reads more naturally

qgallouedec (approved):
lgtm, only two nits

qgallouedec:
Ah, for the record, because it's about slow tests: I’ve never felt that the slow tests add much value. Most are redundant with the default tests, and some aren’t actually slow, they’re just labeled as such for some reason. Cleaning this up and seeing whether we can simply drop them has always been a low-priority item in the back of my mind.

PR #6758, opened 2026-08-14
Fail the secret scanning job when the scan errors out

albertvillanova:
Stacked on top of #6757.

This PR makes the secret scanning job fail when the scan itself errors out, instead of reporting success.

### Motivation

The scanner exits 0 when it cannot scan. If it fails to resolve the commit range it was handed, it logs an error, reports `"chunks": 0, "bytes": 0`, and the job goes green:

```
error  trufflehog  encountered errors during scan  {"errors": ["error chunking dir ...: unable to resolve ref: no base refs succeeded for base: \"main\""]}
info-0 trufflehog  finished scanning  {"chunks": 0, "bytes": 0, "verified_secrets": 0, "unverified_secrets": 0}
```

That is not hypothetical. The same workflow in `huggingface/transformers` has been in exactly this state since huggingface/transformers#47945, where every pull request branch and every merge queue push scans zero bytes and passes:

| run | branch | chunks |
| --- | --- | --- |
| 31792591798 | `fix_qwen2_5_vl_batch_wo_image` | 0 |
| 31793095807 | `serge/fix/itf-...` | 0 |
| 31795946540 | `gh-readonly-queue/main/pr-47960-...` | 0 |

A secret scan that covers nothing is worse than one that fails, because a green check invites nobody to look.

### Solution

Pass `--fail-on-scan-errors`, so the job fails closed on scan errors. Verified against the pinned scanner image:

| scenario | current | with the flag |
| --- | --- | --- |
| unresolvable commit range, 0 chunks scanned | exit 0 | exit 1 |
| verification error on a found secret | exit 183 | exit 183 |

The flag covers source and chunking errors only, so results whose verification errored are still reported through `--results=verified,unknown` exactly as before.

### Changes

- Add `--fail-on-scan-errors` to the scanner arguments, with a comment recording why

---

> [!NOTE]
> **Low Risk**
> Single workflow flag change that tightens CI behavior; verification-error handling via `--results=verified,unknown` is unchanged.
> 
> **Overview**
> The TruffleHog workflow now passes **`--fail-on-scan-errors`** in `extra_args`, so the job fails when the scanner hits source/chunking errors (e.g. unresolvable refs) instead of exiting 0 with zero bytes scanned.
> 
> A short comment documents that behavior so a “green” check cannot mask a broken scan.
> 
[...]

PR #6756, opened 2026-08-14
Declare the test report dependencies in the test extra

albertvillanova:
Stacked on top of #6754.

This PR declares `pytest-reportlog` and `tabulate` in the `test` extra, instead of installing them from the slow tests workflow.

### Motivation

Both jobs of the slow tests workflow install these two dependencies with their own `uv pip install` commands, while every other test dependency is declared in the `test` extra. As a consequence, they are missing from a local `.[dev]` install, so running `make slow_tests` with the report log flag locally fails on an unrecognized option, and the workflow has to know which dependencies the report needs.

### Changes

- Declare `pytest-reportlog` and `tabulate` in the `test` and `dev` extras
- Stop installing them from the slow tests jobs

---

> [!NOTE]
> **Low Risk**
> Packaging and CI-only dependency wiring with no runtime or security impact.
> 
> **Overview**
> Moves **`pytest-reportlog`** and **`tabulate`** from ad-hoc `uv pip install` steps in the slow-tests workflow into the **`test`** and **`dev`** optional dependencies in `pyproject.toml`.
> 
> Both single- and multi-GPU slow test jobs now rely on `uv pip install ".[dev]"` alone for report generation (`scripts/log_reports.py`), so local `make slow_tests` with report logging matches CI without extra packages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2deee62ce6e2afb247e2175b4098d0519662402f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6759, opened 2026-08-14
Scan the pushed commits in the secret scanning workflow

albertvillanova:
Stacked on top of #6758.

This PR makes the secret scanning workflow scan the commits under review instead of the whole repository history, and adds a weekly full-history sweep.

### Motivation

The `Secret Leaks` job fails intermittently on new branches, most recently on #6756: https://github.com/huggingface/trl/actions/runs/31793772381/job/94746235557

The action derives its base from `github.event.before`, which is all zeros on the push that creates a branch. Without a base commit the scanner walks every commit ever made, 33.5k chunks and 22.5 MB, instead of the pushed changes.

That history holds exactly one detectable secret: the public hub-ci dummy token added to `tests/testing_constants.py` in 6e80e09 (January 2023) and removed from the working tree in #1852. It is the `__DUMMY_TRANSFORMERS_USER__` token shared across HF test suites, valid on hub-ci.huggingface.co and rejected with 401 by huggingface.co, which is the endpoint the detector verifies against. No credential needs rotating.

Every new branch push therefore re-verifies that one token, and the job's outcome depends on what huggingface.co answers:

| run | verification | outcome |
| --- | --- | --- |
| 31792044718 | 66 ms | 401, unverified, filtered out |
| 31792208869 | 102 ms | 401, unverified, filtered out |
| 31793772381 | 2000 ms | no answer, so `unknown`, reported, exit 183 |

The detector treats anything that is neither 2xx nor 401 as a verification error, and a verification error lands in the `unknown` bucket that `--results=verified,unknown` reports. One slow or blocked request to huggingface.co is thus enough to fail the job, on a finding that lives in immutable history and cannot be acted on. Reproduced locally with the pinned scanner image and networking disabled: same warning on `tests/testing_constants.py`, same exit 183.

### Solution

Scan the commits under review. On `pull_request` the action passes `base.sha..head.sha`, and on a push to main it passes `before..after`, so the scan covers the new commits and never re-reads history.

Both of those paths pass commit shas, which is why no explicit `base` input is set here. `huggingface/transformers` does set one, falling back to the default branch name on new branches, and a ref name does not resolve in the clone the scanner makes: those runs report `unable to resolve ref: no base refs succeeded for base: "main"`, scan zero bytes, and pass green (huggingface/transformers#47945).

Narrowing the trigger gives up one case, commits pushed to a branch with no pull request open. History stays in scope through the weekly sweep, which runs with no base commit and therefore scans everything, including with detectors added after a commit landed. That sweep is expected to trip over the same hub-ci token from time to time, without blocking anyone.

Detection is otherwise unchanged: same detectors, same `--results=verified,unknown`, same digest pin from #6677 and lob exclusion from #6674.

### Changes

- Trigger the workflow on `pull_request` and on pushes to main, so the scan covers the pushed commits
- Add a weekly scheduled run covering the full history
- Record why an explicit base, if ever set, must be a sha and not a ref name

---

> [!NOTE]
> **Low Risk**
> CI workflow trigger and documentation only; no application code or secret-handling logic changes.
> 
> **Overview**
> **Secret Leaks** no longer runs on every branch `push`. It now runs on **`pull_request`**, on **`push` to `main` only**, and on a **weekly schedule** (Mondays 03:00 UTC) for a full-repo history pass.
> 
> Comments document that PR and main-push runs let TruffleHog derive a base commit so only new commits are scanned (avoiding full-history scans on new branches), and that any future explicit `base` must be a **SHA**, not a ref name.
> 
> Scanner steps, image pin, and `extra_args` are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 682a6b8aa23e36fbb814c56cf5a2f27b8fa195ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:
Reported the transformers side of this upstream: 
- huggingface/transformers#47973

Their workflow sets an explicit `base` that falls back to the default branch name on new branches, which does not resolve in the clone the scanner makes, so those runs scan zero bytes and still pass. That is the trap this PR avoids by letting the action derive base and head as commit shas.

albertvillanova:
For provenance: the diff-scanning approach here comes from the equivalent change in transformers, huggingface/transformers#47945 ([workflow at 0cdd8a1](https://github.com/huggingface/transformers/blob/0cdd8a1908949037cf7718769ffc03dbdf9d9fd6/.github/workflows/trufflehog.yml)). This PR keeps their intent, scanning the pushed commits rather than the whole history, and differs only in how the range is expressed: the action derives base and head as commit shas from the event, instead of an explicit `base` input that falls back to a ref name.

The run that started the investigation: https://github.com/huggingface/trl/actions/runs/31793772381/job/94746235557?pr=6756

qgallouedec (approved):
Thanks, that looks good, but to be honest, it's a little outside my area of expertise. Feel free to merge it, or wait for @McPatate's opinion if you think it's necessary.

Issue #6516, opened 2026-07-23
CI fails: RuntimeError: gemm input type at::BFloat16 and output type float is only supported for CUDA devices with compute capability 8.0 or higher


albertvillanova:
CI fails: https://github.com/huggingface/trl/actions/runs/29988653865/job/89146151900

> FAILED tests/test_sft_trainer.py::TestSFTTrainer::test_train_with_liger - RuntimeError: gemm input type at::BFloat16 and output type float is only supported for CUDA devices with compute capability 8.0 or higher

Traceback:
```python
>       trainer.train()

tests/test_sft_trainer.py:849: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.12/site-packages/transformers/trainer.py:1443: in train
    return inner_training_loop(
.venv/lib/python3.12/site-packages/transformers/trainer.py:1525: in _inner_training_loop
    self._run_epoch(
.venv/lib/python3.12/site-packages/transformers/trainer.py:1753: in _run_epoch
    tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
trl/trainer/sft_trainer.py:1842: in training_step
    return super().training_step(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/transformers/trainer.py:1925: in training_step
    loss = self.compute_loss(model, inputs, num_items_in_batch=num_items_in_batch)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
trl/trainer/sft_trainer.py:1740: in compute_loss
    (loss, outputs) = super().compute_loss(
.venv/lib/python3.12/site-packages/transformers/trainer.py:1999: in compute_loss
    outputs = model(**inputs)
              ^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/torch/nn/modules/module.py:1778: in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
[...]

Issue #1274, opened 2024-01-24
DPO and Packing?


kostum123:
Is there an incompatibility issue between dpo and the 'packing true' setting? My dataset mostly single turn.

younesbelkada:
I don't think it is possible to implement packing for DPO as you need a paired dataset for preferred and rejected responses; not sure how we can implement this in practice. cc @kashif to confirm but we can leave this as an open question

kostum123:
> I don't think it is possible to implement packing for DPO as you need a paired dataset for preferred and rejected responses; not sure how we can implement this in practice. cc @kashif to confirm but we can leave this as an open question

I tried DPO with packing in a llama factory, but I don't know if it automatically falls back to false. The results were not good for a large number of samples, but they were good for 50 samples.

PR #6518, opened 2026-07-23
Fix crash when training with Liger kernel on pre-Ampere GPUs

albertvillanova:
This PR replaces the temporary <0.8.1 liger-kernel hotfix with a narrower !=0.8.1 constraint, so future liger-kernel releases are allowed again while the broken 0.8.1 stays excluded.

Fix #6516.

### Motivation

liger-kernel 0.8.1 crashes on pre-Ampere GPUs (CUDA compute capability < 8.0) in the fused linear cross-entropy backward pass, which broke CI. The immediate hotfix capped the dependency at <0.8.1, but that also blocks every future release. 

The root cause has since been addressed upstream in:
- linkedin/Liger-Kernel#1316

### Solution

Narrow the constraint from <0.8.1 to !=0.8.1: keep excluding the single broken release, but allow the next (fixed) release once it ships.

### Changes

- Change the liger-kernel constraint from >=0.8.0,<0.8.1 to >=0.8.0,!=0.8.1 in both the liger extra and the dev extra.
- Update the accompanying comment to point at the upstream fix (linkedin/Liger-Kernel#1316).

---

> [!NOTE]
> **Low Risk**
> Dependency-only change with no runtime code; still excludes the known-bad 0.8.1 release.
> 
> **Overview**
> Relaxes the **liger-kernel** pin in `pyproject.toml` from `>=0.8.0,<0.8.1` to `>=0.8.0,!=0.8.1` in both the **`liger`** and **`dev`** optional extras. Installers still skip **0.8.1** (pre-Ampere crash in fused linear cross-entropy backward), but can pick up newer fixed releases instead of being capped below 0.8.1.
> 
> Comments now reference the upstream fix (linkedin/Liger-Kernel#1316) instead of the temporary #6516 hotfix note.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1572c8a6b1e8ad3208feda72370fce22524558fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Issue #6435, opened 2026-07-17
Remove SAPO `IndexPutBackward0` warning filter once `liger-kernel >= 0.8.2` is required


albertvillanova:
### What

`test_train_loss_types[sapo-True]` (SAPO loss with `use_liger_kernel=True`) emits a benign `UserWarning` in CI:

```
torch/autograd/graph.py: UserWarning: Error detected in IndexPutBackward0. Traceback of forward call that caused the error:
  File ".../liger_kernel/chunked_loss/grpo_loss.py", line 171, in torch_dynamo_resume_in_ppo_loss_fn_at_172
    per_token_loss[positive_advantages_mask] = sapo_loss_fn(...)
```

Example CI run: https://github.com/huggingface/trl/actions/runs/29507534675/job/87652100190

### Root cause

The warning originates in `liger-kernel`, not TRL. Its SAPO branch assigns the per-token loss with boolean-mask indexing (`per_token_loss[mask] = sapo_loss_fn(coef_1[mask], ...)`). That indexing lowers to `aten.nonzero` (data-dependent output shape), which Inductor cannot trace, so `torch.compile` inserts a graph break. AOTAutograd runs the forward metadata collection under `torch.autograd.detect_anomaly(check_nan=False)`, and the in-place `IndexPutBackward` gets flagged, producing the warning. Training itself is correct — the test passes and parameters update.

### Upstream status

Already fixed upstream by replacing the boolean-mask assignment with `torch.where`: https://github.com/linkedin/Liger-Kernel/pull/1274

The fix is **merged but unreleased** — the latest liger-kernel release is `0.8.0` (which predates the fix), and TRL currently requires `liger-kernel>=0.8.0`.

### Workaround in place

We suppress the warning via `filterwarnings` in `pyproject.toml`:

```toml
"ignore:Error detected in IndexPutBackward0:UserWarning",
```

### Action item

Remove the `filterwarnings` entry once `pyproject.toml` requires `liger-kernel >= 0.8.2`, i.e. the first non-excluded release containing #1274. That bump will happen naturally when 0.8.2 ships with #1316 and the `!=0.8.1` exclusion is dropped; the filter removal belongs in that same PR. Do not bump the floor solely to drop this filter.

PR #6761, opened 2026-08-14
Unpin liger-kernel 0.8.1

qgallouedec:
Requires #6741 (GPU CI on L40S).

0.8.1 was excluded because it needs bf16 gemm (compute capability >= 8.0) and our T4 runner is sm_75 see #6516. The L40S runner is sm_89, so the constraint is gone.

- `liger-kernel>=0.8.0,!=0.8.1` → `liger-kernel>=0.8.0` (`liger` + `dev` extras)
- Drop the SAPO `IndexPutBackward0` warning filter, fixed upstream in 0.8.1 (linkedin/Liger-Kernel#1274)

EDIT: after second thought, we shouldn't merge this now: `trl[liger]` is published, so unpinning breaks users on pre-Ampere GPUs (#6516) even though our CI moved to an L40S. PEP 508 markers can't scope a dependency by GPU arch. The upstream guard (linkedin/Liger-Kernel#1316) merged ~24h *after* 0.8.1 shipped, so it lands in 0.8.2 (not to be confused with #1274 (SAPO fix), which is in 0.8.1.)
When 0.8.2 ships: bump both extras to `liger-kernel>=0.8.2`, drop the SAPO `IndexPutBackward0` filterwarnings entry.

---

> [!NOTE]
> **Low Risk**
> Packaging and test-warning cleanup only; no runtime training logic changes.
> 
> **Overview**
> **Allows `liger-kernel` 0.8.1** again by changing `liger-kernel>=0.8.0,!=0.8.1` to `liger-kernel>=0.8.0` in the **`liger`** and **`dev`** optional dependencies. That exclusion existed because 0.8.1 needs bf16 GEMM (compute capability ≥ 8.0), which the old T4 CI runner could not satisfy; GPU CI on L40S removes that constraint.
> 
> **Removes** the pytest `filterwarnings` entry that ignored Liger SAPO `IndexPutBackward0` under `torch.compile`, since that behavior was fixed upstream in 0.8.1.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 545bf8f0defb7ceb6c7449656e6937340f4230a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ kernels = [  # transformers renamed the "hub-kernels" extra to "kernels" in 5.1.
     "transformers[hub-kernels]",  # transformers < 5.1.0
 ]
 liger = [
-    "liger-kernel>=0.8.0,!=0.8.1"  # upstream PR https://github.com/linkedin/Liger-Kernel/pull/1316
+    "liger-kernel>=0.8.0"
 ]
 peft = [
     "peft>=0.8.0"
@@ -111,7 +111,7 @@ dev = [
     "transformers[kernels]",  # transformers >= 5.1.0
     "transformers[hub-kernels]",  # transformers < 5.1.0
     # liger
-    "liger-kernel>=0.8.0,!=0.8.1",  # upstream PR https://github.com/linkedin/Liger-Kernel/pull/1316
+    "liger-kernel>=0.8.0",
     # openreward (requires Python 3.11+)
     "openreward>=0.1.109; python_version >= '3.11'",
     # peft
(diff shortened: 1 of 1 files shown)


albertvillanova (changes requested):
Thanks, @qgallouedec.

I agree with your EDIT: this shouldn't be merged, and I'd go further and close it.

Some context on why the constraint has its current shape: the `!=0.8.1` in #6518 was chosen over `<0.8.1` precisely for this situation. It excludes exactly the one broken release while leaving every future release allowed, so **no TRL-side change is needed for 0.8.2 to be picked up**; the current pin already permits it. That makes this PR unnecessary rather than merely early.

Three additional reasons beyond the GPU-arch point:
1. **The end state in this diff is `>=0.8.0`, which permanently re-permits 0.8.1.** Even once 0.8.2 exists, a bare `>=0.8.0` lets a resolver land on the broken release (old lockfiles, `--no-deps` installs, constrained resolutions). The right target is `>=0.8.2`, as your EDIT says.
2. **Timing makes it worse than it looks.** Because 0.8.1 is currently the newest release, unpinning today doesn't just *allow* the broken version: every fresh `pip install trl[liger]` would resolve to exactly it on pre-Ampere hardware (#6516).
3. **The `filterwarnings` removal is coupled to the same bump.** With `!=0.8.1` in force, CI resolves 0.8.0, which predates #1274, so dropping the ignore entry brings the `IndexPutBackward0` noise back. The existing comment already records this (`Remove once: liger-kernel >= 0.8.2 is required`).

And as you note, PEP 508 markers can't express "0.8.1 is fine iff compute capability >= 8.0", so there's no way to keep CI on L40S free of the constraint while protecting published `trl[liger]` users. A runtime guard in TRL would just duplicate the upstream fix, which I don't think is worth it.

So in conclusion, I would close this PR. Thanks for catching this before it went in: the GPU CI move to L40S (#6741) is genuinely the right change, but it just doesn't unblock this one.

albertvillanova:
For the record, the follow-up work is tracked in two issues, so nothing is lost when this PR is closed:

- #6767: bump `liger-kernel>=0.8.0,!=0.8.1` → `liger-kernel>=0.8.2` in both the `liger` and `dev` extras, once 0.8.2 ships with linkedin/Liger-Kernel#1316.
- #6435: remove the `"ignore:Error detected in IndexPutBackward0:UserWarning"` filter. Its upstream fix (linkedin/Liger-Kernel#1274) is in 0.8.1, which we exclude, so 0.8.2 is the first non-excluded release that contains it.

The two are coupled and should land in the same PR; #6767 carries the full checklist.

PR #6770, opened 2026-08-17
Ensure device agnostic for examples/docs/templates

kaixuanliu:
@qgallouedec @albertvillanova @kashif pls help review, thx!

---

> [!NOTE]
> **Low Risk**
> Mostly documentation and generated model-card snippets; the online DPO change is a small, localized swap to a device-aware cache helper with limited blast radius.
> 
> **Overview**
> Makes user-facing snippets and one training path less CUDA-specific so they work across accelerators and typical Hugging Face loading patterns.
> 
> **Docs** (`use_model.md`): the load-and-generate example now picks the device with `torch.accelerator.current_accelerator()` when available, otherwise CPU, instead of a fixed `"cpu"` / `"cuda"` string.
> 
> **Model cards** (`lm_model_card.md`, `rm_model_card.md`): quick-start `pipeline` examples use `device_map="auto"` instead of `device="cuda"`. Matching assertions in `tests/test_utils.py` were updated.
> 
> **Online DPO**: before waking colocated vLLM in sleep mode, cache clearing uses the shared `empty_cache()` helper rather than `torch.cuda.empty_cache()`, consistent with other experimental trainers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3bfae5d66c32b2a44d76a2b64aa774794e0f2ca0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kashif (approved):

kashif (changes requested):
So can you confirm that the vLLM example is device agnostic, as far as I know the ` VLLMClient.init_communicator` is CUDA only?

docs/source/use_model.md
```diff
@@ -7,10 +7,11 @@ Once you have trained a model using either the SFTTrainer, PPOTrainer, or DPOTra
 If you have fine-tuned a model fully, meaning without the use of PEFT you can simply load it like any other language model in transformers. E.g. the value head that was trained during the PPO training is no longer needed and if you load the model with the original transformer class it will be ignored:
 
 ```python
+import torch
```
kashif, line 10:
```suggestion
import torch

import accelerate
```

kashif:
Could you revert the changes to `trl/generation/vllm_client.py`? The weight-transfer communicator is explicitly CUDA/NCCL-only and rejects non-CUDA devices, so using `torch.accelerator.current_accelerator()` here is misleading. The other device-agnostic changes look good.

docs/source/use_model.md
kaixuanliu: Hi, thx for review! However, I don't think the extra `import accelerate` is required here. The snippet doesn't reference the `accelerate` module directly; `torch.accelerator` comes from PyTorch, and if `device_map="auto"` is used, Transformers handles the Accelerate dependency internally. So I'd prefer not to add an unused import unless we explicitly call an Accelerate API.

kaixuanliu:
@kashif Oops, my fault. Have reverted the changes in vllm_client.py. Pls help review again.

kaixuanliu:
@kashif Hi, Could you pls help review again? Thx!

kashif (approved):

PR #6772, opened 2026-08-17
fix: surface NaN GRPO/RLOO steps in log_history

### LINKED ISSUE #6702: GRPO: a NaN training step is invisible in log_history, so a corrupted run reports a clean loss curve
### Summary

A training step whose loss is NaN is not visible in `log_history`. The reported loss curve stays finite and plausible while the step that produced it was NaN, so a run that is silently corrupting its weights looks healthy in the logs and in any tracker fed from them.

This came out of the regression test for #6166 (PR #6693). @albertvillanova asked for it to be tracked separately, since it is orthogonal to that fix.

### Mechanism

`transformers.Trainer` filters NaN and Inf out of the logged running loss by design:

```python
# transformers/trainer.py:1746
if (
    self.args.logging_nan_inf_filter
    and not is_torch_xla_available()
    and (torch.isnan(tr_loss_step) or torch.isinf(tr_loss_step))
):
    # if loss is nan or inf simply add the average of previous logged losses
    self._tr_loss += self._tr_loss / (1 + self.state.global_step - self._globalstep_last_logged)
else:
    ...
```

`logging_nan_inf_filter` defaults to `True` (`transformers/training_args.py:977`), and its own help text says it "Does not affect gradient computation, only logging". So the NaN step still runs `backward()` and still takes an optimizer step. Only the *report* is repaired, by substituting the mean of the previously logged losses.

### Why this matters more for GRPO than for plain SFT

The GRPO loss has several paths that can produce a NaN from data rather than from divergence: an unscorable token logprob coming back from vLLM (#6166), a degenerate reward group, an overflowing `exp` of a KL log-ratio (#3015). In those cases the NaN is not a symptom of a bad learning rate that a user would go looking for. It is a single poisoned token or group, and the resulting NaN gradients propagate into the weights while the loss curve shows nothing at all.

### Reproduction

Observed while building the #6166 test. With the fix removed, an individual step's loss is NaN at the point `_compute_loss` returns it, and `state.log_history["train_loss"]` is still finite for that same run. That is why the test in PR #6693 asserts on the losses returned by `_compute_loss` rather than on `log_history`: asserting on `log_history` passes even with the fix removed, which makes it a vacuous test.

### Possible directions

Listing these as options rather than a proposal, since the right layer is a maintainer call:

1. Count NaN and Inf steps and surface the count as a metric, so the filtering stays but stops being silent.
2. Emit a warning the first time a step is filtered, naming the step index.
3. Leave `transformers` alone and check for a non-finite loss inside the GRPO loss path, where the cause is known and can be named.

Option 3 keeps the change inside TRL and can point at the actual source, which the generic filter cannot do.

I am happy to implement whichever direction you prefer.

cc @qgallouedec @kashif

tonycoder-hub:
# What does this PR do?

Trainer logging_nan_inf_filter replaces a NaN loss with the previous average, so a corrupted GRPO step looks clean in log_history. This logs frac_non_finite_loss on GRPO and RLOO so a non-finite step is visible.

Fixes #6702

## Before submitting

## AI writing disclosure

## Who can review?

GRPO tests: 157 passed. RLOO tests: 70 passed.

---

> [!NOTE]
> **Low Risk**
> Small observability-only change in compute_loss with no change to loss computation or optimization logic; low risk aside from minor distributed logging overhead.
> 
> **Overview**
> **GRPO** and **RLOO** now record **`frac_non_finite_loss`** in training/eval metrics so corrupted steps are visible in `log_history` even when Hugging Face **`Trainer`**’s default **`logging_nan_inf_filter`** replaces NaN/Inf losses with a smoothed finite value in the reported **`loss`**.
> 
> Each **`compute_loss`** call flags non-finite losses (gathered across processes), appends to the existing metrics pipeline, and still returns the raw loss for backward. GRPO covers both the standard and Liger-kernel paths.
> 
> Docs for GRPO and RLOO logged metrics describe the new field; regression tests poison the first step with NaN and assert **`frac_non_finite_loss > 0`** on that step only while **`loss`** stays finite (fixes #6702).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 497266229cdb75eac6e6e60ea17e01034f40e786. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/grpo_trainer.md
+++ b/docs/source/grpo_trainer.md
@@ -205,6 +205,7 @@ While training and evaluating, we record the following metrics:
 - `cispo_clip_ratio`: The ratio of token (or sequence, if `importance_sampling_level="sequence"`) importance sampling weights that were clipped at `epsilon_high` while having a positive advantage:  \\(r_{i,t}(\theta) > \epsilon_\mathrm{high}\\). Logged only when `loss_type="cispo"`.
 - `vespo/phi_seq_mean`: The average value of the VESPO Gamma weighting  \\(\varphi(w)\\)  applied to the sequence-level importance weights. Values below `1.0` mean sequences are being down-weighted. Logged only when `loss_type="vespo"`.
 - `clip_ratio`: The ratio of clipped tokens reported by the fused Liger GRPO loss. Logged only when `use_liger_kernel=True`, in which case it replaces the `clip_ratio/*` metrics above.
+- `frac_non_finite_loss`: The fraction of forward passes in the logging window whose loss was NaN or infinite. Such a loss is still backpropagated, but [`~transformers.Trainer`] hides it from the reported loss by replacing it with the average of the previous losses (`logging_nan_inf_filter`, enabled by default), so this metric is the only trace it leaves. Any nonzero value means the update was computed from a non-finite loss.
 
 ## Customization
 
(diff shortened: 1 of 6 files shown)


tonycoder-hub:
Closing as stale — opened on or before 2026-08-17 and still unmerged.

PR #6779, opened 2026-08-17
Run `TestUseAdapter` on the accelerator

### LINKED ISSUE #6778: CI fails: ValueError: Pointer argument cannot be accessed from Triton (cpu tensor?)
CI fails (transiently) with minimum versions: https://github.com/huggingface/trl/actions/runs/32046322535/job/95434970188

```
FAILED tests/test_utils.py::TestUseAdapter::test_disables_on_none - ValueError: Pointer argument cannot be accessed from Triton (cpu tensor?)
FAILED tests/test_utils.py::TestUseAdapter::test_restores_previous_adapter - ValueError: Pointer argument cannot be accessed from Triton (cpu tensor?)
FAILED tests/test_utils.py::TestUseAdapter::test_with_multiple_adapters - ValueError: Pointer argument cannot be accessed from Triton (cpu tensor?)
```

Stacktrace:

```python
tests/test_utils.py:73: in test_disables_on_none
    expected = model(input_ids).logits
...
transformers/models/qwen3/modeling_qwen3.py:258: in forward
    hidden_states = self.input_layernorm(hidden_states)
liger_kernel/transformers/rms_norm.py:38: in forward
    return LigerRMSNormFunction.apply(
liger_kernel/ops/rms_norm.py:442: in rms_norm_forward
    _rms_norm_forward_kernel[(n_rows,)](
E   ValueError: Pointer argument cannot be accessed from Triton (cpu tensor?)
```

Cause: Liger patches the modeling module globally, not just the instance:

```python
# liger_kernel/transformers/monkey_patch.py:1689, in apply_liger_kernel_to_qwen3
modeling_qwen3.Qwen3RMSNorm = LigerRMSNorm
```

`Trainer` calls this via `_apply_liger_kernel_to_instance` whenever `use_liger_kernel=True`, so every Qwen3 model instantiated afterwards **in the same process** gets the Triton RMSNorm. `TestUseAdapter` then loads `trl-internal-testing/tiny-PeftModel` (base: `tiny-Qwen3ForCausalLM`) on CPU and runs the Triton kernel on CPU tensors.

The polluter is most likely `tests/test_distillation_trainer.py::TestDistillationTrainer::test_liger_allows_none_logit_scale` (and siblings), which builds a `DistillationTrainer` with a Qwen3 student and `use_liger_kernel=True`. These tests moved into `tests/` in #6647 (11250f0a), so they now share a process with `tests/test_utils.py`. It's transient because `pytest -n auto` schedules dynamically: whether a Qwen3 Liger test lands in the same worker before `TestUseAdapter` varies per run.

Not the same as #4816 (unprepared CPU ref model), despite the identical error message.

qgallouedec:
- `TestUseAdapter` left the model and inputs on CPU, unlike the rest of the test suite.
- When a Liger test runs first in the same worker, it globally patches `modeling_qwen3.Qwen3RMSNorm = LigerRMSNorm`, so the CPU forward launches a Triton kernel on CPU tensors: `ValueError: Pointer argument cannot be accessed from Triton (cpu tensor?)`.
- Move the model and `input_ids` to `torch_device`.

Fixes the transient CI failures in https://github.com/huggingface/trl/actions/runs/32046322535/job/95434970188.

Closes #6778.

---

> [!NOTE]
> **Low Risk**
> Test-only device placement in `tests/test_utils.py`; no production or library API changes.
> 
> **Overview**
> **`TestUseAdapter`** now places the PEFT model and **`input_ids`** on **`torch_device`**, matching the rest of the TRL test suite instead of leaving them on CPU.
> 
> That avoids flaky CI when a Liger test runs earlier in the same worker and globally swaps in a Triton-backed norm: a CPU forward then errors with *Pointer argument cannot be accessed from Triton (cpu tensor?)*. Behavior under test is unchanged; only device placement is updated in all three adapter tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cdfc47464d5268a8a44a6356b3c8160926f09524. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -67,8 +67,8 @@ class TestUseAdapter(TrlTestCase):
     def test_disables_on_none(self):
         model = AutoPeftModelForCausalLM.from_pretrained(
             "trl-internal-testing/tiny-PeftModel", adapter_name="my_adapter"
-        )
-        input_ids = torch.tensor([[1, 2, 3], [4, 5, 6]])
+        ).to(torch_device)
+        input_ids = torch.tensor([[1, 2, 3], [4, 5, 6]], device=torch_device)
         with model.disable_adapter():
             expected = model(input_ids).logits
 
@@ -80,8 +80,8 @@ class TestUseAdapter(TrlTestCase):
     def test_restores_previous_adapter(self):
         model = AutoPeftModelForCausalLM.from_pretrained(
             "trl-internal-testing/tiny-PeftModel", adapter_name="my_adapter"
-        )
-        input_ids = torch.tensor([[1, 2, 3], [4, 5, 6]])
+        ).to(torch_device)
+        input_ids = torch.tensor([[1, 2, 3], [4, 5, 6]], device=torch_device)
         expected = model(input_ids).logits
         with use_adapter(model, "my_adapter"):
             pass
@@ -96,9 +96,9 @@ class TestUseAdapter(TrlTestCase):
     def test_with_multiple_adapters(self):
         model = AutoPeftModelForCausalLM.from_pretrained(
             "trl-internal-testing/tiny-PeftModel", adapter_name="my_adapter_1"
-        )
+        ).to(torch_device)
         model.load_adapter("trl-internal-testing/tiny-PeftModel-2", "my_adapter_2")
-        input_ids = torch.tensor([[1, 2, 3], [4, 5, 6]])
+        input_ids = torch.tensor([[1, 2, 3], [4, 5, 6]], device=torch_device)
 
         model.set_adapter("my_adapter_1")  # should be a no-op, but let's keep it for clarity
         expected_1 = model(input_ids).logits


albertvillanova:
Thanks for the quick fix. After digging into the root cause, I would rather not merge this one and I suggest we close it in favour of #6783.

The failure is not really about device placement. `use_liger_kernel=True` makes transformers call `_apply_liger_kernel_to_instance`, which rebinds module-level names in `transformers.models.<arch>.modeling_<arch>` (`Qwen3RMSNorm` -> `LigerRMSNorm`, `Qwen3MLP` -> `LigerSwiGLUMLP`, `apply_rotary_pos_emb` -> `liger_rotary_pos_emb`) and never restores them. Moving `TestUseAdapter` to the accelerator hides the symptom, but the patched globals stay in place for the rest of the worker process.

Concretely, my concerns:

- Any later test that builds a model of a Liger-patched architecture on CPU fails the same way, for any of the architectures Liger supports, so we would be back here on the next PR that shuffles the xdist scheduling. Several tests load models on CPU, `TestPatchChunkedLMHead` among them.
- On a machine without an accelerator, `torch_device` is `cpu`, so the flake remains for contributors running the suite locally and for any CPU-only job.
- Once the model runs on the accelerator, the assertions become scheduling dependent: `TestUseAdapter` would compare logits computed either with Liger kernels or with the reference implementation, depending on whether a Liger test happened to run earlier in the same worker. These tests assert exact equality on adapter switching, so I would rather keep the kernels out of the picture entirely.
- CPU placement is a deliberate and cheap choice for what these tests cover, so moving them also drops the CPU coverage of the adapter logic.

#6783 reverts the Liger patches after each test with an autouse fixture in `tests/conftest.py`. It fixes the whole class of failures and keeps `TestUseAdapter` on CPU.

Worth noting for the record that the leak is only triggered differently across versions, not fixed upstream: transformers applies the kernels in `Trainer.__init__` before 5.3 and in `Trainer.train()` from 5.3 on. That is why only the minimum-versions job fails today, since the Qwen3 Liger tests build a trainer but never train. After a real `train()` on transformers >= 5.3, the modules stay patched for the rest of the process just the same.

PR #6794, opened 2026-08-18
Rewrite the TRL skill around the Python API

qgallouedec:
Step 3 of #6743: rewrite the skill around the Python API — trainer selection, dataset format, the load-bearing config fields, the GRPO reward-function contract — with the CLI as a footnote.

This is a minimal starting point; we will refine it based on the output of the internal trl-autoresearch project.

---

> [!NOTE]
> **Low Risk**
> Documentation-only change to an agent skill file; no runtime or training code paths are modified.
> 
> **Overview**
> **Rewrites** `skills/trl-training/SKILL.md` from a long **CLI-first** reference into a **compact Python API** guide (step 3 of #6743).
> 
> The skill frontmatter is **trimmed** (drops version/commands/categories/tags) and the description now targets **writing or debugging TRL training code** via `*Trainer` / `*Config`, with the CLI as secondary. Body content **replaces** per-command `trl sft|dpo|grpo|rloo|reward` bash recipes, YAML config walkthroughs, distributed Accelerate/FSDP/DeepSpeed sections, troubleshooting, and best practices with: a **trainer ↔ dataset** table (including `DistillationTrainer` and a pointer to `trl.experimental`), a canonical **`SFTTrainer` example**, **dataset format** rules, **load-bearing `SFTConfig` fields**, and a **GRPO** section covering reward-function kwargs, generation batch sizing, vLLM modes, and `AsyncGRPOTrainer`. **RLOO/KTO/DPO** lose dedicated CLI sections; the **CLI** is a short note that flags mirror config fields plus `--config` / `--accelerate_config`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 916af83c10262a172b80c08f912d63e9d7353378. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


skills/trl-training/SKILL.md
```diff
@@ -1,333 +1,77 @@
...
-  --eval_steps 50 \
-  --output_dir Qwen2-0.5B-DPO \
-  --no_remove_unused_columns
+```python
+SFTConfig(
+    max_length=1024,        # truncation length; None disables truncation
+    packing=True,           # pack sequences into max_length blocks: fewer pad tokens, higher throughput
+    padding_free=True,      # flatten batch, no padding; requires FlashAttention; implied by packing
+    use_liger_kernel=True,  # fused Liger kernels, reduces peak memory
+    assistant_only_loss=True,  # loss only on assistant turns (conversational datasets)
```
lewtun, line 51: Do we need to spell this out or will it work for most models thanks to your chat template fixes?
```suggestion
    assistant_only_loss=True,  # loss only on assistant turns (conversational datasets); requires compatible chat template. 
```

lewtun (approved):
Nice clean up - LGTM with a small question about compatible chat templates for the assistant loss

skills/trl-training/SKILL.md
qgallouedec: It works for most models now: if the template already has `{% generation %}` markers it's used as is, otherwise trl swaps in a patched training template automatically for the common families (Qwen 2.5 through 3.8, Llama 3, Gemma 1-3, DeepSeek-V3 and R1-Distill, GPT-OSS, Cohere, Phi, LFM2, and a few VLMs). Anything else fails with a clear error asking to fix the template. So I dropped the caveat entirely in the current revision.

trl/.agents/skills/trl/SKILL.md
```diff
@@ -1,333 +1,148 @@
...
+trainer.train()
 ```
 
-## Configuration Files
-
-TRL supports YAML configuration files for reproducible training. All CLI arguments can be specified in a config file.
+**Reward function contract**: called with keyword arguments `prompts`, `completions`, `completion_ids`, `trainer_state`, plus every extra dataset column (e.g. `ground_truth`) — accept `**kwargs` for the ones you ignore. Returns `list[float]` (one reward per completion; `None` entries opt a function out for those samples). With conversational data, `completions` is a list of message lists (`completions[0][0]["content"]`), not strings. Functions may be `async def`. Built-ins live in `trl.rewards` (`accuracy_reward`, `think_format_reward`, …). `reward_funcs` also accepts a reward-model ID string.
 
-**Example config (sft_config.yaml):**
+**Generation batching**: each optimization step generates `num_generations` (default 8) completions per prompt. The generation batch is `per_device_train_batch_size × num_processes × steps_per_generation`, or set `generation_batch_size` directly (mutually exclusive); it must be divisible by `num_generations`.
```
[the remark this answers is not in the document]
qgallouedec: The section only covers GRPO in the current revision, RLOO isn't mentioned in the skill, and GRPOConfig.num_generations defaults to 8, so the statement is right as written.

PR #6799, opened 2026-08-19
Use the doc-builder pre-commit hook

albertvillanova:
This PR replaces the local `doc-builder-style` hook with the hook that `doc-builder` ships, so that its pinned revision can be kept up to date automatically.

### Motivation

The hook is declared locally and installs `doc-builder` from a git URL pinned inside `additional_dependencies`:

```yaml
- repo: local
  hooks:
    - id: doc-builder-style
      entry: doc-builder style trl tests docs/source --max_len 119
      additional_dependencies: ["git+https://github.com/huggingface/doc-builder@af49b8d"]
```

A full commit SHA is a strong pin, but this one is invisible to the tooling that would keep it current. `pre-commit autoupdate` only rewrites the `rev` of remote hook repos. Dependabot's `pre-commit` ecosystem does parse `additional_dependencies`, but it resolves them against package registries, so a `git+https://...@<sha>` spec gives it nothing to look up. The revision can therefore only be moved by hand, and silently goes stale.

That is what happened here: the pin sat untouched, and the upstream fix it was eventually bumped for (huggingface/doc-builder#785) turned out to be a no-op that nobody noticed until #6790.

Now that `doc-builder` ships a `.pre-commit-hooks.yaml` (huggingface/doc-builder#818) and declares `ruff` as a dependency (huggingface/doc-builder#815), the standard form can be used instead, where `rev` is a first-class field.

### Solution

```yaml
- repo: https://github.com/huggingface/doc-builder
  rev: 0ab9ea03baf111ed8dd83e88233430b663127368
  hooks:
    - id: doc-builder-style
      args: [--max_len, "119"]
      files: ^(trl|tests|docs/source)/
```

Three details matter, because the shipped hook passes filenames while the local one did not:

- `types_or` is dropped. Ours was `[python, markdown, rst]`, which only acted as a trigger filter, harmless while `pass_filenames` was `false`. Keeping it would override the manifest's `[python, mdx]` and start handing `.md` and `.rst` files to a tool that only warns `Ignoring <file>` for them. The manifest already restricts the hook to the file types `doc-builder style` handles.
- `files` keeps the current scope. Without it the hook would cover every Python file in the repo, and `doc-builder style <files> --check_only` reports **58 tracked files outside `trl`, `tests` and `docs/source` that would be restyled**, mostly under `examples/`. Widening the scope is worth doing, but as its own PR.
- `--max_len 119` stays explicit. It matches `doc-builder`'s default, but it also matches the `line-length` we set for ruff in `pyproject.toml`, so it is better stated than inherited.

The hook is now incremental: locally it styles only the files being committed, instead of walking `trl`, `tests` and `docs/source` on every commit. CI is unaffected, since it runs `pre-commit run --all-files`.

Verified with `ruff` removed from `PATH`, so the hook environment has to be self-sufficient. A run over the tree reports no change, but on its own that would not distinguish a working hook from one whose `files` pattern matches nothing, so the check was made against a file that has to be restyled: a docstring with a 160 character `Args:` line and a `probe( 1 )` code example, placed under `trl/` and, identically, under `examples/`.

The hook rewrapped the `Args:` line, bringing the longest line to 117, and reformatted the example to `probe(1)`, which also shows `ruff` is resolved inside the hook environment rather than from the shell. The copy under `examples/` was left untouched, so `files` scopes as intended, and a second run passes, so the hook is idempotent. Running the same probe with `--max_len 60` wraps at 60 columns, which confirms `args` reaches the tool rather than the default of 119 making the setting look effective either way.

On the current tree the hook then passes and modifies nothing, so the migration changes no content.

Two accompanying changes make this maintainable:

- `.pre-commit-config.yaml` is added to the `paths` filter of the `Tests` workflow. It was missing, so `check_code_quality` did not run on pull requests that only touch that file, which is why #6790 could be merged green while breaking the hook. Without this, neither this PR nor any future hook bump would be tested.
- A `pre-commit` ecosystem entry is added to `.github/dependabot.yml`, mirroring the `github-actions` entry that already keeps our pinned workflow SHAs current. It is deliberately not grouped, so that a `ruff` bump, which reformats code, arrives as its own reviewable pull request rather than combined with a `doc-builder` bump.

### Changes

- Replace the local `doc-builder-style` hook with the hook shipped by `doc-builder`, keeping the current scope and line length
- Run the code quality job when `.pre-commit-config.yaml` changes
- Let Dependabot update the pre-commit hooks weekly

### Notes

The hook styles the docstrings of our Python files, which is where essentially all of our doc styling happens: text rewrapping, `Args:` blocks, embedded code examples and Markdown callouts are all handled there. This PR does not change that coverage in any way.

Standalone Markdown pages are a different matter: `doc-builder style` handles `.py` and `.mdx` files, and `docs/source` holds 63 `.md` files and no `.py` or `.mdx`, so it contributes nothing to the hook, after this PR as well as before it. It is kept in `files` to preserve the current intent, rather than narrowing the scope based on what the folder happens to contain today.

Styling `.md` pages as well would be an upstream change to `doc-builder style`. It would not be a no-op for us: 7 of our 63 doc pages currently have ```python fences that ruff cannot parse, so they would need to be fixed or retagged first.

Dependabot's `pre-commit` support covers version updates only, not security updates, and it is unclear whether it moves a `rev` pinned to a commit SHA. So the entry may in practice only act on `ruff-pre-commit`, and this `rev` may still need to be bumped by hand. That is the cost of pinning a SHA, and it is the same trade-off we already accept for the action SHAs in our workflows.

### TODO

- [x] Merge huggingface/doc-builder#818
- [x] Repin `rev` to its merge commit, `0ab9ea03`

---

> [!NOTE]
> **Low Risk**
> Changes only pre-commit and CI/Dependabot configuration; no application runtime or auth/data paths are touched.
> 
> **Overview**
> Replaces the **local** `doc-builder-style` pre-commit hook (git pin in `additional_dependencies`) with the **official** hook from `huggingface/doc-builder`, pinned via `rev`, with `--max_len 119` and a `files` pattern so scope stays `trl`, `tests`, and `docs/source`. The hook becomes **incremental** on commit (filenames passed through) instead of restyling whole trees each time; CI still runs `pre-commit run --all-files`.
> 
> Adds **`.pre-commit-config.yaml`** to the Tests workflow `pull_request` path filter so hook-only PRs run `check_code_quality`. Adds a **Dependabot `pre-commit`** ecosystem entry (weekly, ungrouped from other bumps) so hook revisions can be updated automatically.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b0ce45a3bb938e64eb54831be2e25e0b954810e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (approved):
lgtm!

One note on the motivation: doc-builder's latest tag is v0.5.0 from march 2024, which predates `.pre-commit-hooks.yaml`. Autoupdate resolves to tags, so it'll refuse to move this rev (hook id missing at v0.5.0) I think, and dependabot's pre-commit ecosystem tracks the same field.

So we're still bumping the sha by hand for now. Worth asking upstream to cut a tag? then repinning to it. Not blocking though.

albertvillanova:

Confirmed about the tag issue with pre-commit autoupdate. However, we do not use `autoupdate` anywhere in the repo, so this only matters if someone runs it by hand and mistakes the exit code for a broken config.

The most important point: on Dependabot I think we are actually better off than "bumping by hand". I think `rev` is parsed as a git ref, and for a SHA it proposes the head of the branch containing the pin, i.e. `main`. Caveat: this is from reading the source, not from an observed run, so the first weekly cycle after merge will settle it either way.

I have added `# main` after the SHA to document that. It also matches the convention we already use for the pinned reusable workflows.

On cutting a tag, I don't think we really need that. We should not repin to it, as we pin SHAs deliberately because of security reasons: tags are mutable.

albertvillanova:
> The most important point: on Dependabot I think we are actually better off than "bumping by hand". I think rev is parsed as a git ref, and for a SHA it proposes the head of the branch containing the pin, i.e. main. Caveat: this is from reading the source, not from an observed run, so the first weekly cycle after merge will settle it either way.

Confirmed that Dependabot works as expected. See:
- https://github.com/huggingface/trl/pull/6969
- It now pins SHA to https://github.com/huggingface/doc-builder/commit/1b16dac5e33043af565fdf4c1b5b0fe81d0891c8

PR #6814, opened 2026-08-19
Test liger vs non-liger GRPO loss parity

kashif:
# What does this PR do?

Follow-up to #5890.

`test_compute_liger_loss_dapo_normalizer` patched two internals to capture the raw Liger output and the final loss, then asserted the same expression `compute_liger_loss` computes. It could only break on a refactor, not on a real regression, so I dropped it.

Replaced it with `test_liger_loss_matches_non_liger_loss`: generate and score one batch, then compute the loss on that exact batch through both paths (`use_liger_kernel` True and False) and assert they match. Parametrized over all loss types, so the dapo/cispo/vespo per-window rescale is covered along with the rest.

I checked it catches the regression it is meant to guard: removing the `/ steps_per_generation` from the Liger normalizer fails exactly the dapo, cispo and vespo cases, and passes for the other four.

Note that a full train-and-compare would not work here, since `use_liger_kernel` also makes Transformers patch the model, which changes generation and gives different completions on the two runs.

## Before submitting

## AI writing disclosure

## Who can review?

@qgallouedec

---

> [!NOTE]
> **Low Risk**
> Test-only change; no production GRPO or Liger training logic is modified.
> 
> **Overview**
> **Replaces** the GRPO Liger regression test that only asserted the trainer’s own rescale math (via patched `liger_loss.forward` and `compute_liger_loss`) with **`test_liger_loss_matches_non_liger_loss`**, which prepares one real training batch and checks that `compute_loss` matches between `use_liger_kernel=True` and `False`.
> 
> The new test is **parametrized** over `grpo`, `bnpo`, `dr_grpo`, `dapo`, `cispo`, `sapo`, `luspo`, and `vespo` (plus `dapo` with non-zero `beta` for KL), and uses `steps_per_generation=4` vs `gradient_accumulation_steps=2` so DAPO/CISPO/VESPO per-window rescaling is exercised. For `dapo`, it additionally asserts the Liger path applies the expected `steps_per_generation / gradient_accumulation_steps` factor on the raw kernel loss.
> 
> **Removes** `test_compute_liger_loss_dapo_normalizer` (~60 lines).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6289545fd15f440557014719885c08c2797bae47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


behroozazarkhalili:
The parity assertion is a strong detector of per-path divergence, and mutation testing confirms it. A mutant that changed only the dapo/cispo/vespo rescale was caught by exactly `{dapo, cispo, vespo}`; a mutant that changed the shared token-mean path was caught by exactly `{grpo, bnpo, dr_grpo, sapo, luspo}`. The partition was clean, with no false positives.

What a relative oracle cannot see is an error in a quantity both sides read. `torch.testing.assert_close(liger_loss, loss)` compares two values that both consume `trainer.current_gradient_accumulation_steps`, which the test sets by hand two lines earlier, so a wrong value there moves both sides together.

Setting `current_gradient_accumulation_steps = 99` puts the loss at 0.0202 times its correct value, which is 2/99, an absolute error of 49.5x. The liger and non-liger gap stays at about 3e-09, or 3e-05 of the `rtol` you pass, and the test passes on all eight `loss_type` values.

The deleted `test_compute_liger_loss_dapo_normalizer` was the only assertion pinning the rescale to a known number rather than to the other path. Re-adding one absolute check on a single `loss_type` would close this without giving up the parametrization.

albertvillanova (commented):
Nice replacement. The relative oracle is the right shape for this, and covering all eight loss types is strictly more than the deleted test gave us. Three things before I approve.

1. +1 to @behroozazarkhalili's point above, and I think it should be addressed before merge. Both paths divide by `current_gradient_accumulation_steps` (`grpo_trainer.py:3016` and `:3260`), which the test sets by hand, so an error in that factor cancels in the comparison. Their `= 99` mutant confirms it. Deleting `test_compute_liger_loss_dapo_normalizer` removed the only assertion tying the rescale to a known number. As they suggest, one absolute check on a single loss type is enough: for `dapo`, capture the raw `liger_loss.forward` output and assert the returned loss is `raw * steps_per_generation / gradient_accumulation_steps`. That keeps the parametrization and doesn't reintroduce the old double-patch structure.

2. `beta` defaults to `0.0`, so the Liger loss runs with `use_ref_model=False` and the `beta * per_token_kl` term in `_compute_loss` is never compared. A single `beta > 0` case would cover the KL branch.

3. `assert_close(0., 0.)` passes. Worth a non-triviality guard (`assert liger_loss.abs() > 0`) so the parity assertion can't succeed vacuously.

Rest looks good: `use_liger_kernel` flipped post-init does switch the branch at `:3027`, `current_gradient_accumulation_steps` is set before `_prepare_inputs` needs it at `:1569`, and the test mirrors `test_train_loss_types` in style and placement.

lewtun:
+1 to @albertvillanova's comments. Otherwise looks quite good to me

kashif:
ok sure on it!

qgallouedec (approved):
thanks!

PR #6774, opened 2026-08-17
AsyncGRPOConfig : dtype asyncgrpo training

AmineDiro:
# What does this PR do?

Added dtype to `AsyncGRPOTrainer` with inference dtype detection to raise a warning if mismatch is detected 

See:
https://huggingface.co/papers/2510.26788
https://huggingface.co/spaces/aminediroHF/trainer-generator-bf16-mismatch

---

> [!NOTE]
> **Medium Risk**
> Changes default model loading path and training-inference precision alignment, which can affect GRPO importance ratios and numerical behavior if users change `dtype` or ignore the new mismatch warning.
> 
> **Overview**
> Adds a first-class **`dtype`** on **`AsyncGRPOConfig`** (default **`float32`**, overridable via **`model_init_kwargs`**) and wires **`AsyncGRPOTrainer`** to load the policy through that setting instead of a hardcoded **`torch.float32`**.
> 
> On weight-transfer setup, the trainer compares the dominant parameter dtype sent to vLLM with the server’s served dtype (new **`VLLMClient.get_dtype()`** via **`/server_info`**) and **logs a warning** when they differ, pointing users to align **`AsyncGRPOConfig.dtype`** and **`vllm serve --dtype`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f6034d427bd596b06df595f231b5ca0ab3a340e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6823, opened 2026-08-20
Catch AsyncDistillationTrainer up with recent AsyncGRPO changes

kashif:
# What does this PR do?

Catches `AsyncDistillationTrainer` up with the AsyncGRPO changes that landed around and after #6705, since the two trainers are meant to stay in sync.

* #6378: merges the two train-begin callbacks into one `_TrainBeginCallback`, and drops the redundant `.any()` guard in the MOPD per-teacher stats loop (an empty masked selection already sums to a 0 scalar on the right device).
* #6774: adds `AsyncDistillationConfig.dtype` (default `"float32"`, which is what the student was hardcoded to) and warns when the student's vLLM server serves a different dtype than the weights we send it. Only the student is checked, teacher servers are never updated so their dtype is their own business.
* #6385: adds `_EpochStopCallback`. `max_steps` is derived assuming a fixed number of samples per step, which only holds under `FixedCountBatcher`. `TokenBudgetBatcher` is the default and packs whatever fits the budget, so `num_train_epochs` did not match the actual number of passes over the prompts. Samples now carry a `prompt_id`, which is AsyncGRPO's `group_id` with `num_generations` fixed at 1, and the callback stops once the requested number of prompts have been trained on.
* #5911: saves the index of the first prompt not yet trained on to `rollout_state.json` next to each checkpoint and restores it on resume, so a resumed run does not regenerate prompts it already trained on. `ignore_data_skip` now defaults to `True` here too.

One deliberate difference from AsyncGRPO: the `max_steps` ceiling is left uninflated. `AsyncGRPOConfig` defaults `lr_scheduler_type` to `constant` so its exact value does not matter there, but this config inherits `linear`, so `max_steps` also sets the decay horizon.

Also restores the `> [!NOTE]` list formatting in `AsyncDistillationConfig`'s docstring, which got collapsed into one paragraph when #6785 removed the `docstyle-ignore` tag.

## Before submitting

## AI writing disclosure

## Who can review?

@qgallouedec @aminediro

---

> [!NOTE]
> **Medium Risk**
> Changes training length semantics, checkpoint format, and vLLM weight-sync parameter sets; incorrect prompt_index or epoch counting could skip or repeat data, but behavior mirrors proven AsyncGRPO paths and is heavily tested.
> 
> **Overview**
> Brings **`AsyncDistillationTrainer`** in line with recent **`AsyncGRPOTrainer`** behavior: checkpoint/resume, epoch stopping, config defaults, and a few training-loop fixes.
> 
> **Checkpoint and resume** — Each checkpoint now writes **`rollout_state.json`** with the global **`prompt_index`** (first dataset row not yet trained), written before the parent upload so Hub copies include it. On resume, the rollout worker gets **`dataset_start_index`** from that file; **`ignore_data_skip`** defaults to **`True`** and is forced on because HF’s data-skip replay does not match a live rollout queue. **`IterableDataset`** resumes from prompt 0 with a warning.
> 
> **Epoch semantics** — Rollouts carry **`prompt_id`**; the collator accumulates distinct ids into **`_trained_prompts`**. **`_EpochStopCallback`** stops once **`num_train_epochs × |dataset|`** prompts have been trained (including **`_prompts_before_resume`** on resume), instead of relying only on **`max_steps`** under variable **`TokenBudgetBatcher`** packing.
> 
> **Config and loading** — **`AsyncDistillationConfig`** adds **`dtype`** (default **`float32`**) for student load; weight sync skips frozen parameters and normalizes gradient-checkpoint parameter names for vLLM.
> 
> **Callbacks** — Train-begin weight sync and rollout worker start are merged into **`_TrainBeginCallback`**.
> 
> Docs add a **Checkpoint and resume** section; tests cover epoch stop, **`rollout_state`** save/restore, iterator skip, and end-to-end resume smoke.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f45f0aeea84f3f2fbf07900d7972e7e3f32e83d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_distillation/weight_transfer.py
```diff
@@ -53,6 +54,17 @@ def __init__(
 
     def init_weight_transfer(self) -> None:
         self.vllm.wait_for_server_ready()
+        # Trainer and server precisions should agree: the precision gap between the two makes the sampled completions
```
AmineDiro: I actually wonder if this is really needed for distillation. does reverse KL introduce the same issue ?

AmineDiro (approved):

qgallouedec:
thanks, the catch-up itself is solid! I diffed against async_grpo on main and it looks right.

2 things before it goes in.
- @AmineDiro's question on `weight_transfer.py` is still open
- bugbot's resume finding is real, `_inner_training_loop` resets `dataset_start_index` and `_prompts_before_resume`  but never clears `_trained_prompts`

and the codex one about the epoch target is a both-trainers fix, so I'd recommend for a next pr

trl/experimental/async_distillation/vllm_client.py
```diff
@@ -70,6 +70,18 @@ def get_max_model_len(self) -> int:
         response.raise_for_status()
         return response.json()["data"][0]["max_model_len"]
 
+    def get_dtype(self) -> str:
```
AmineDiro: Is this called somewhere or just to keep clients in sync?

AmineDiro (approved):
Great adding robust checkpointing :+1 

Do we have a full end-to-end checkpoint example run  to check the rollout continues correctly ?

trl/experimental/async_distillation/weight_transfer.py
kashif: Good point, dropped it. There's no importance ratio here, and the student's logprobs always come from the trainer rather than vLLM (generation only returns token ids), so the two engines never meet in the loss. Same for forward and reverse KL, it's about where the samples come from, not the direction. And max_staleness defaults to 4, so we already accept far more off-policyness than a bf16 vs fp32 gap. Kept the dtype field, the student was hardcoded to float32 before.

trl/experimental/async_distillation/vllm_client.py
kashif: Only the dtype warning used it, so it's gone too.

Issue #6836, opened 2026-08-20
CI fails for slow tests on multiple GPUs: Fatal Python error: Segmentation fault


albertvillanova:
CI fails for slow tests on multiple GPUs: https://github.com/huggingface/trl/actions/runs/32366421358/job/96416805091
> Fatal Python error: Segmentation fault

pytest dies before printing the summary, so there is no `short test summary info` and no per-test traceback in the log. What the log does show:

```
collected 2662 items / 2564 deselected / 98 selected

tests/test_dpo_trainer.py s                                              [  1%]
tests/test_grpo_trainer.py FFFFFFFxss                                    [ 11%]
tests/test_kto_trainer.py s                                              [ 12%]
tests/test_sft_trainer.py s.F.F......FFFFFF......FFFFFF................. [ 59%]
Fatal Python error: Segmentation fault
```

Faulthandler stack of the crashing thread:
```
Current thread 0x00007f8766541740 (most recent call first):
  File ".../accelerate/utils/modeling.py", line 817 in get_max_memory
  File ".../transformers/integrations/accelerate.py", line 206 in get_max_memory
  File ".../transformers/integrations/accelerate.py", line 272 in get_balanced_memory
  File ".../transformers/integrations/accelerate.py", line 349 in _get_device_map
  File ".../transformers/modeling_utils.py", line 4328 in from_pretrained
  File "/__w/trl/trl/trl/trainer/utils.py", line 1174 in create_model_from_path
  File "/__w/trl/trl/trl/trainer/sft_trainer.py", line 977 in __init__
  File "/__w/trl/trl/tests/test_sft_trainer.py", line 2544 in test_train_offloading
```

```
Segmentation fault (core dumped)
make: *** [Makefile:15: slow_tests] Error 139
```

The job runs without xdist, so the progress characters map onto the collection order deterministically. The failures reported before the crash are:

```python
tests/test_grpo_trainer.py::TestGRPOTrainerSlow::test_train_with_liger_grpo_kernel[trl-internal-testing/tiny-LlamaForCausalLM-3.2]
tests/test_grpo_trainer.py::TestGRPOTrainerSlow::test_train_with_liger_grpo_kernel[trl-internal-testing/tiny-MistralForCausalLM-0.2]
tests/test_grpo_trainer.py::TestGRPOTrainerSlow::test_train_with_liger_grpo_kernel_and_peft[trl-internal-testing/tiny-LlamaForCausalLM-3.2]
[...]

PR #6840, opened 2026-08-20
Hotfix CI: skip the offloading test when several accelerators are visible

albertvillanova:
This PR stops the multi-GPU slow-tests job from being killed by a segmentation fault, so it reports its results instead of dying halfway.

Related to #6836.

### Motivation

`test_train_offloading` segfaults on the multi-GPU runner, taking the whole pytest process down with exit code 139:

```
Current thread (most recent call first):
  File ".../accelerate/utils/modeling.py", line 817 in get_max_memory
  File ".../transformers/integrations/accelerate.py", line 272 in get_balanced_memory
  File ".../transformers/modeling_utils.py", line 4328 in from_pretrained
  File "trl/trainer/utils.py", line 1174 in create_model_from_path
  File "trl/trainer/sft_trainer.py", line 977 in __init__
  File "tests/test_sft_trainer.py", line 2544 in test_train_offloading
```

Because the process dies, pytest never prints `short test summary info`, the `--report-log` file is truncated, and the other failures in the run cannot be identified at all. For #6836 they had to be recovered by hand, by mapping the progress characters onto the collection order.

The crash only happens when more than one accelerator is visible; the test passes in the single-GPU job. The cause is not yet understood, and no marker can catch a segfault, so the test has to not run there.

### Changes

- Skip `TestSFTTrainerSlow::test_train_offloading` when more than one accelerator is visible

---

> [!NOTE]
> **Low Risk**
> Test-only skip; no production or training-path changes. Coverage of activation offloading is reduced on multi-GPU CI until the underlying Accelerate crash is fixed.
> 
> **Overview**
> Skips `test_train_offloading` when more than one accelerator is visible so the multi-GPU slow-test job is not killed by a segfault in Accelerate’s `get_max_memory`.
> 
> The test still runs on single-accelerator jobs. The crash (issue #6836) takes down the whole pytest process, which truncates reports and hides other failures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 229eaebecb5dd52d0a8c535e1326f6d1d972a009. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -33,7 +33,7 @@ from transformers import (
     BitsAndBytesConfig,
     TrainingArguments,
 )
-from transformers.testing_utils import backend_empty_cache, torch_device
+from transformers.testing_utils import backend_device_count, backend_empty_cache, torch_device
 from transformers.utils import is_peft_available
 
 from trl import SFTConfig, SFTTrainer
@@ -2530,6 +2530,11 @@ class TestSFTTrainerSlow(TrlTestCase):
         ],
     )
     @require_torch_accelerator
+    @pytest.mark.skipif(
+        backend_device_count(torch_device) > 1,
+        reason="segfaults in accelerate's get_max_memory when more than one accelerator is visible, taking the whole "
+        "pytest process down; cause not yet diagnosed (https://github.com/huggingface/trl/issues/6836)",
+    )
     def test_train_offloading(self, model_name, packing):
         """Test that activation offloading works with SFTTrainer."""
         training_args = SFTConfig(


tests/test_sft_trainer.py
```diff
@@ -2530,6 +2530,11 @@ def cleanup_liger_patches(trainer):
         ],
     )
     @require_torch_accelerator
+    @pytest.mark.skipif(
+        backend_device_count(torch_device) > 1,
+        reason="segfaults in accelerate's get_max_memory when more than one accelerator is visible, taking the whole "
+        "pytest process down; cause not yet diagnosed (https://github.com/huggingface/trl/issues/6836)",
+    )
     def test_train_offloading(self, model_name, packing):
```
qgallouedec, line 2538: I don't know why this one is tested in the slow tester. But it's out of the scope of this pr, just for the record

qgallouedec (approved):

PR #6858, opened 2026-08-21
Add `loss_type="cce"`, backed by cut-cross-entropy

qgallouedec:
`chunked_nll` (TRL's default) keeps the `lm_head` projection and the softmax as separate steps, so peak memory scales with `chunk_size * vocab_size` and the logits are written to HBM and read back. [cut-cross-entropy](https://github.com/apple/ml-cross-entropy) fuses the projection into the cross-entropy kernel, so the logits never exist. This adds it as a `loss_type`, next to `nll`, `dft` and `chunked_nll`.

Modern vocabularies make this matter a lot: Qwen3.5/3.6 is 248,320 tokens against a 2048 hidden size, so the `lm_head` alone is 508M parameters and the logits for one 16k-token micro-batch are 16 GB in bf16.

## Numbers

`bench/loss_bench.py`, 16384 tokens x 248320 vocab x 2048 hidden, 1xH100, bf16, fwd+bwd:

fig_loss

| | time | peak memory | loss vs reference |
|---|---|---|---|
| `chunked_nll` (TRL default today) | **OOM** | — | — |
| `use_liger_kernel=True` | 321.16 ms | 6.05 GB | rel 0 |
| `chunked_nll` + PR 7 | 224.39 ms | 19.67 GB | rel 2.3e-07 |
| one big bf16 GEMM (reference) | 137.00 ms | 47.55 GB | rel 0 |
| **`loss_type="cce"`** | **92.55 ms** | **3.09 GB** | rel 3.8e-07 |

At 4096 tokens, where the current default does fit: **318.95 ms / 38.02 GB → 23.18 ms / 2.95 GB**, i.e. **13.8x faster and 12.9x lighter**.

End to end, 3.86B MoE (248k vocab, 64 experts), seq 4096, per-device batch 2, 1xH100: **9857 → 18402 tokens/s/GPU (1.87x)**.

Reproduce:

```bash
pip install cut-cross-entropy
python bench/loss_bench.py --tokens 4096 16384
```

<details>
<summary>End-to-end check that the two losses agree, on a released model</summary>

`learning_rate=0.0` matters here: with a live LR, four steps on random labels at a 248k vocabulary diverge chaotically between any two numerically-different implementations, so a multi-step comparison says nothing.

```python
import torch, numpy as np
from datasets import Dataset

ds = Dataset.from_dict({"input_ids": np.random.default_rng(0).integers(0, 200000, size=(24, 512)).tolist()})
for loss_type in ("chunked_nll", "cce"):
    cfg = SFTConfig(
        output_dir="/tmp/cce", per_device_train_batch_size=2, max_steps=1, learning_rate=0.0,
        save_strategy="no", report_to=[], bf16=True, max_length=512, loss_type=loss_type, seed=0,
        dataset_kwargs={"skip_prepare_dataset": True}, remove_unused_columns=False,
        model_init_kwargs=dict(dtype=torch.bfloat16, attn_implementation="sdpa"),
    )
    trainer = SFTTrainer(model="Qwen/Qwen3.5-2B-Base", args=cfg, train_dataset=ds)
    print(loss_type, trainer.train().training_loss)
```

```
chunked_nll 14.087625
cce         14.087476        # rel 1.1e-05
```

And the two loss functions called directly on the same tensors, with and without `num_items_in_batch` (2x512 tokens, 248320 vocab, a masked prompt prefix):

```
num_items_in_batch=None: chunked 12.444077  cce 12.444086  rel 7.7e-07  n_valid 1014 vs 1014
num_items_in_batch=1012: chunked 12.468671  cce 12.468681  rel 8.4e-07  n_valid 1014 vs 1014
```
</details>

<details>
<summary>Gradient accuracy, and why <code>filter_eps</code> is left at its default</summary>

CCE's `filter_eps` drops negligible terms from the backward. Against an fp32-softmax reference
(4096 tokens, 248320 vocab), it makes no measurable difference and turning it off costs 3.2x:

```
filter_eps=high   d_hidden rel 1.032e-02   d_lm_head rel 4.695e-03    23.19 ms
filter_eps=None   d_hidden rel 1.003e-02   d_lm_head rel 4.695e-03    74.66 ms
```

Both sit at the bf16 floor, so the patch keeps the default.
</details>

Under **real FSDP2 on 2xH100** (same model, seq 4096, per-device batch 2), so the option is not just a single-GPU effect:

fig_options_fsdp2

| config | throughput | peak memory |
|---|---|---|
| TRL defaults | 6413 tokens/s/GPU | 32.3 GB |
| `compile_layers=True` | 7396 (1.15x) | 31.3 GB |
| `loss_type="cce"` | 15453 (2.41x) | 32.3 GB |
| both | **22939 (3.58x)** | **31.3 GB** |

## The trade

CCE returns the loss and nothing else, there are no logits to take an `argmax` or a softmax over, so `mean_token_accuracy` and `entropy` are not logged under `loss_type="cce"`. That is the same trade `use_liger_kernel=True` already makes in TRL, and the metric branches are guarded the same way.

`softcap` and `shift` map onto CCE's own arguments, so `final_logit_softcapping` and pre-shifted `shift_labels` (CP/SP) keep working. `logit_scale` does **not** have a CCE equivalent, so the patch raises rather than silently training on a different loss than the model defines:

```
ValueError: `loss_type='cce'` does not support models with a logit scale (got 0.0625).
Use `loss_type='chunked_nll'` for this model.
```

## Relationship to the chunked-CE projection fix

A companion PR fixes the *default* path (`chunked_nll` currently runs its `lm_head` GEMM in fp32 on bf16 operands) and is worth 1.31–1.54x on its own. Link coming.

---

> [!NOTE]
> **Medium Risk**
> Touches the SFT loss/forward path (including FSDP2 weight gathering and metric logging) but is opt-in, gated on an extra package, and fails closed on unsupported models.
> 
> **Overview**
> Adds `loss_type="cce"` to `SFTTrainer`: NLL with the `lm_head` fused into Apple’s `cut_cross_entropy` kernel so logits are never materialized and peak memory no longer scales with `vocab_size`.
> 
> It reuses the existing chunked-CE forward patch (`_patch_chunked_ce_lm_head(..., use_cce=True)`). Softcap and pre-shifted `shift_labels` still work; models with `logit_scale` or an `lm_head` bias, Liger, and fp32 hidden states are rejected. Because there are no logits, `mean_token_accuracy` and `entropy` are not logged (same trade as Liger). The model must be loaded in fp16/bf16 — `bf16=True` autocast is not enough.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b48570574d741b7d1533b6f0e6f664d67765fb5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -49,6 +49,7 @@ from .testing_utils import (
     ignore_warnings,
     is_ampere_or_newer,
     require_bitsandbytes,
+    require_cut_cross_entropy,
     require_kernels,
     require_liger_kernel,
     require_peft,
@@ -409,6 +410,40 @@ class TestSFTTrainer(TrlTestCase):
         trainer.train()
 
         assert trainer.state.log_history[-1]["train_loss"] is not None
+
+        # Check that the params have changed
+        for n, param in previous_trainable_params.items():
+            new_param = trainer.model.get_parameter(n)
+            assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
+
+    @require_torch_accelerator
+    @require_cut_cross_entropy
+    def test_train_cce_loss(self):
+        dataset = load_dataset("trl-internal-testing/zen", "standard_language_modeling")
+
+        # CCE's backward kernel is half-precision only, so the model has to be loaded in bf16 -- `bf16=True`
+        # alone would not do it, since under autocast the norm feeding the `lm_head` still emits fp32.
+        training_args = SFTConfig(
+            output_dir=self.tmp_dir,
+            loss_type="cce",
+            model_init_kwargs={"dtype": "bfloat16"},
+            learning_rate=0.1,  # use higher lr because bf16 updates at the default lr round to zero
+            report_to="none",
+        )
+        trainer = SFTTrainer(
+            model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5",
+            args=training_args,
+            train_dataset=dataset["train"],
+        )
+
+        previous_trainable_params = {n: param.clone() for n, param in trainer.model.named_parameters()}
+
+        trainer.train()
+
+        assert trainer.state.log_history[-1]["train_loss"] is not None
+        # `cce` fuses the projection into the CE kernel, so there are no logits to read these off.
+        assert "mean_token_accuracy" not in trainer.state.log_history[-1]
+        assert "entropy" not in trainer.state.log_history[-1]
 
         # Check that the params have changed
         for n, param in previous_trainable_params.items():
--- a/tests/testing_utils.py
+++ b/tests/testing_utils.py
@@ -34,6 +34,7 @@ from transformers.utils import (
 
 from trl.chat_template_utils import _SUPPORTS_RESPONSE_TEMPLATE
 from trl.import_utils import (
+    is_cut_cross_entropy_available,
     is_harbor_available,
     is_jmespath_available,
     is_joblib_available,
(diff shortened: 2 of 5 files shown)


qgallouedec:
let's do https://github.com/huggingface/trl/pull/6859 directly instead

Issue #6773, opened 2026-08-17
Nemotron 3 SFT example uses target_modules that may not exist in hybrid architecture


davidnichols-ops:
## Problem

TRL's official Nemotron 3 SFT example (`examples/scripts/sft_nemotron_3.py`) uses LoRA target modules that may not exist in NemotronH's hybrid architecture:

```python
target_modules=["q_proj", "k_proj", "v_proj", "o_proj", 
                "gate_proj", "up_proj", "down_proj"]
```

The `gate_proj`, `up_proj`, `down_proj` modules are standard MLP projection names from Llama-style architectures. NemotronH uses Mamba-2 layers (with `in_proj`, `out_proj`) and MoE layers (with different naming). These MLP module names may silently not match anything in the model.

## Impact

Users following the official example may get LoRA adapters that only target a small fraction of the model's layers (the attention projections), with the MLP/Mamba modules silently skipped. This produces the same "loss not improving" failure mode we experienced:

- DPO LoRA training: 25 steps, loss stuck at 0.6931 (ln(2))
- Zero behavioral change in the model
- No warning that target modules didn't match

## Reproduction

We ran DPO LoRA on `nvidia/NVIDIA-Nemotron-Nano-9B-v2` with attention-only target modules (same as the example minus the MLP modules). The LoRA adapter only touched 4 out of 56 layers (7.1%). Loss never moved from ln(2). Full writeup: https://github.com/davidnichols-ops/claude-yolo-vibes-v5/blob/master/EVALUATION.md

## Proposed Fix

1. Update the example to use target modules that actually exist in NemotronH:

```python
# Inspect the model to find valid target modules
import re
target_modules = list(set(
    name.split(".")[-1] for name, module in model.named_modules()
    if isinstance(module, torch.nn.Linear) and 
    re.match(r".*\d+\..*", name)  # only layer-level modules
))
```

Or explicitly list NemotronH-appropriate modules:

```python
target_modules=[
    # Attention
    "q_proj", "k_proj", "v_proj", "o_proj",
    # Mamba (if PEFT allows — see PEFT issue about mamba_model_types ban)
    "in_proj", "out_proj", 
    # MoE/MLP — check actual module names in the model
    # "gate", "up", "down"  # verify these exist before including
]
```

2. Add a comment in the example explaining that NemotronH is a hybrid architecture and users should verify target modules with `model.named_modules()`.

3. Consider adding a runtime check in TRL that warns when configured `target_modules` don't match any modules in the model.

[...]

PR #6820, opened 2026-08-20
Reorganize examples into self-contained per-example folders

qgallouedec:
## Why

`examples/` was split by format (`scripts/` vs `notebooks/`), which scatters related files: `sft_nemotron_3` existed as both a script and a notebook in different folders, the OpenEnv notebooks lived apart from the OpenEnv scripts, and `sudoku_prompt.txt` sat loose next to unrelated scripts.

I don't think this makes a lot of sense.

I suggest a new approach:

- **One folder per example, and every folder is a "story"**: the name says the method and the task (`grpo_wordle`, `sft_gpt_oss`, `ppo_tldr`), and the folder holds everything the example needs (scripts, notebooks, prompts, chat templates, eval code). Exception: shared `accelerate_configs/` and the `datasets/` generation scripts stay at the root.
- **An example is not a bare training script**: thin single-trainer demo scripts are removed. Those live in `trl/scripts/` (CLI) and every trainer doc page has a runnable snippet.

### Before

```
examples/
├── accelerate_configs/
│   └── …
├── cli_configs/
│   └── example_config.yaml
├── datasets/
│   └── …
├── notebooks/
│   ├── README.md
│   ├── grpo_agent.ipynb
│   ├── grpo_ministral3_vl.ipynb
│   ├── grpo_qwen3_vl.ipynb
│   ├── grpo_rnj_1_instruct.ipynb
│   ├── grpo_trl_lora_qlora.ipynb
│   ├── openenv_sudoku_grpo.ipynb
│   ├── openenv_wordle_grpo.ipynb
│   ├── sft_ministral3_vl.ipynb
│   ├── sft_nemotron_3.ipynb
│   ├── sft_qwen_vl.ipynb
│   ├── sft_tool_calling.ipynb
│   └── sft_trl_lora_qlora.ipynb
├── scripts/
│   ├── harbor/
│   │   ├── harnesses/
│   │   │   ├── bash/
│   │   │   │   ├── README.md
│   │   │   │   └── __init__.py
│   │   │   ├── jupyter/
│   │   │   │   ├── README.md
│   │   │   │   ├── __init__.py
│   │   │   │   ├── env.py
│   │   │   │   ├── kernel_server.py
│   │   │   │   └── run_cell.py
│   │   │   ├── terminal_notes/
│   │   │   │   ├── README.md
│   │   │   │   ├── __init__.py
│   │   │   │   └── env.py
│   │   │   └── __init__.py
│   │   └── data_agent.py
│   ├── openenv/
[...]

PR #3289, opened 2025-04-14
[GRPO] Add metrics for low and high clipped token probabilities

lewtun:
# What does this PR do?

This PR adds metrics to track low and high clipped token probabilities independently. It is useful for finding a suitable empirical value of `ε_high` which cannot be done from `clip_ratio` alone.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #6866, opened 2026-08-22
Drop the dead gate_proj LoRA target from the Nemotron 3 SFT example

behroozazarkhalili:
The Nemotron example listed `gate_proj` as a LoRA target even though the model has no matching module. I removed it from the script and notebook, leaving the modules that PEFT already adapts unchanged. I checked the checkpoint and instantiated model's module names, and a PEFT probe confirmed that mixed target lists silently ignore the missing name.
Refs #6773.

---

> [!NOTE]
> **Low Risk**
> Example-only LoRA target list cleanup with no runtime or training logic changes.
> 
> **Overview**
> Removes **`gate_proj`** from the LoRA `target_modules` in the Nemotron 3 SFT **notebook** and **script** example so the config matches **NemotronH** (no `gate_proj` linear layers). Attention and shared-expert targets (`q_proj`, `k_proj`, `v_proj`, `o_proj`, `up_proj`, `down_proj`) are unchanged.
> 
> **Training behavior is unchanged**—PEFT already ignored unmatched names when other targets matched. The update only drops misleading config from the example and from saved adapter metadata. The notebook JSON also gets a trailing newline fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5d137eb6faab08f1cd1a9ce445cad8267354d4d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


davidnichols-ops:
**Correct:** `gate_proj` is dead — the evidence table is solid. But this is the small half of issue #6773.

The bigger problem is what’s **missing** from the target list, not what’s dead in it.

NemotronH is **46 Mamba layers + 6 attention layers** out of 52.  
The current target list (`q_proj`, `k_proj`, `v_proj`, `o_proj`, `up_proj`, `down_proj`) only covers attention + MoE MLP.  
The Mamba SSM layers (`in_proj`) are not targeted, so LoRA adapts ~11% of the model and the other 89% stays frozen.

We hit this in practice:  
DPO LoRA on Nemotron-Nano-9B-v2 with attention-only targets, 25 steps → loss never moved from \(\ln(2)\). Zero behavioral change. The model wasn’t learning because most of it was frozen.

**Full writeup:** https://github.com/davidnichols-ops/claude-yolo-vibes-v5/blob/master/EVALUATION.md

`in_proj` is a standard `nn.Linear` in the Mamba forward path and is **not** blocked by PEFT’s Mamba compatibility check:

```python
incompatible_modules = {"out_proj", "conv1d"}  # in_proj is not in this set 
``` 

PEFT PR #3289 added nemotron_h defaults but only included attention projections, not in_proj.

Suggestion: Add in_proj to the target list in this PR (or a follow-up).
That’s the fix that actually makes LoRA training work on this architecture.

behroozazarkhalili:
Thanks, this is a good catch on the dense hybrids, and the numbers are worth pinning down because they split by architecture.

I measured both checkpoints directly from their safetensors headers rather than from layer counts.

`nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16`, the model this script targets, 31.58B parameters, `hybrid_override_pattern` = 23 `M` + 23 `E` + 6 `*`:

| modules | count | params | share |
|---|---|---|---|
| `up_proj` + `down_proj` (MoE experts) | 2967 each | 29.83 B | 94.46% |
| `in_proj` (Mamba) | 23 | 0.64 B | 2.02% |
| `out_proj` (Mamba) | 23 | 0.25 B | 0.80% |
| `q/k/v/o_proj` (attention) | 6 each | 0.14 B | 0.44% |

The current target list covers 94.92% of this model, not 11%. The 23 `E` layers are MoE MLP, not Mamba, so `up_proj`/`down_proj` already reach them; that is where essentially all the mass is.

`nvidia/NVIDIA-Nemotron-Nano-9B-v2`, the model you trained, 8.89B, pattern = 27 `M` + 25 `-` + 4 `*`: the current list covers 41.99% and `in_proj` alone is 30.83%, the largest single block. Your result follows directly from that. It is dense, so there are no expert matrices to dominate the total, and skipping `in_proj` leaves the biggest block frozen.

On the PEFT point, you are right that `in_proj` is not blocked, and by a wider margin than the snippet suggests: `_check_lora_target_modules_mamba` only runs when `model_type` is one of `falcon_h1`, `mamba`, `mamba2`, `falcon_mamba`. `nemotron_h` is not in that set, so the guard never fires for either model. I also see no `nemotron` entry anywhere in PEFT 0.18.1, so nothing is being auto-populated; the script's explicit list is what is used.

Adding `in_proj` is still defensible for this script: those 23 Mamba mixers are genuinely untrained today. The justification is 2% of this model, though, not a stalled run, so I would rather land it as its own change with its own reasoning than fold it into a one-line removal. Happy to open that, and I think #6773 should stay open for the dense-hybrid case you hit, which is the larger problem.

behroozazarkhalili:
Opened #6870 with the `in_proj` addition, stacked on this branch so the diff is just the one token. It carries the per-model numbers and the PEFT matching check.

I kept it separate because the justification differs: on the 30B MoE model it is a 2% improvement, while on the dense hybrids you hit it is the largest block. #6773 is still worth keeping open for that case.

davidnichols-ops:
Thanks for the thorough follow-up — measuring the checkpoint directly and splitting this into two PRs with per-model evidence was the right call. The 94.92% vs 41.99% split between the 30B MoE and 9B dense hybrid corrects my earlier framing. Both look good to me.

behroozazarkhalili:
Refreshed against `main` and re-targeted, because the file this PR edits was moved.

`#6820` reorganized examples into per-example folders, so `examples/scripts/sft_nemotron_3.py` is now `examples/sft_nemotron_3/sft_nemotron_3.py`. Merging `main` produced no conflict: git's rename detection carried the change to the new path on its own. The branch is now 0 commits behind `main`.

The finding is unchanged at the new location. Line 55 on `main` still reads:

```
--lora_target_modules q_proj k_proj v_proj o_proj gate_proj up_proj down_proj
```

and `gate_proj` still matches nothing in Nemotron 3, while `up_proj` and `down_proj` each match 2967 modules.

One note so it is not mistaken for a regression from this change: `examples/sft_nemotron_3/sft_nemotron_3.py` does not currently satisfy the pinned doc-builder style check. I verified that it fails identically on unmodified `main` at `e17a8ae6`, so it is pre-existing and outside this PR's scope. `ruff check` and `ruff format --check` at the CI-pinned 0.13.3 both pass.

PR #6875, opened 2026-08-23
Fix CLI handling of --accelerate_config=<name>

### LINKED ISSUE #6874: Bug: normalize --accelerate_config=<name> CLI syntax
## Reproduction

`--accelerate_config` accepts the documented two-token form, but the equivalent standard equals form is not normalized before TRL forwards launch arguments to Accelerate.

```python
from trl.cli.accelerate_config import resolve_accelerate_config_argument

launch_args = resolve_accelerate_config_argument(
    ["--accelerate_config=single_gpu", "--num_processes", "2"]
)
assert launch_args[0] == "--config_file"
```

On `main` at `6d484baed4db18c2799b9d5b32c2520cba6b4df2`, the assertion fails because the function leaves `--accelerate_config=single_gpu` unchanged. The two-token form (`--accelerate_config single_gpu`) is normalized correctly.

## Expected behavior

Both spellings should resolve the packaged configuration and preserve remaining Accelerate launch arguments:

```text
--config_file .../single_gpu.yaml --num_processes 2
```

## System Info

The defect was reproduced from the main commit above. Local regression validation used:

- Platform: macOS-26.5.2-arm64-arm-64bit
- Python version: 3.11.15
- PyTorch version: 2.13.0
- accelerator(s): cpu
- Transformers version: 5.15.1
- Accelerate version: 1.14.0
- Datasets version: 5.0.1

## Checklist

- [x] I checked that this issue is not already filed.
- [x] I included system information.
- [x] The code sample is minimal, complete, and reproducible.
- [x] The code sample is formatted in a code block.

YZJF:
# What does this PR do?

Fixes #6874.

This makes the standard `--accelerate_config=<name>` spelling behave like the existing
`--accelerate_config <name>` form. Both forms now resolve the packaged Accelerate configuration before TRL forwards launch arguments to Accelerate, while preserving remaining launch arguments.

The regression test exercises the `trl sft` CLI path with a mocked launcher and covers both spellings.

## Validation

- `pytest -p no:rerunfailures tests/test_cli.py -k accelerate_config_argument -vv` — 2 passed.
- `ruff check trl/cli/accelerate_config.py tests/test_cli.py` — passed.
- `ruff format --check trl/cli/accelerate_config.py tests/test_cli.py` — passed.

The full `tests/test_cli.py` module was also run on a CPU-only host: 11 tests passed and 9 trainer-launch tests failed because their spawned subprocesses require `use_cpu` for the default bf16 configuration. The focused regression test does not launch training and passes.

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request-pr), Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? The related bug report is #6874.

## AI writing disclosure

---

> [!NOTE]
> **Low Risk**
> Narrow CLI argument parsing change with targeted tests; no auth, data, or training logic changes.
> 
> **Overview**
> Fixes handling of the `--accelerate_config=<name>` form so it matches `--accelerate_config <name>` when TRL resolves packaged Accelerate YAMLs and forwards `--config_file` plus remaining launch flags to Accelerate.
> 
> `resolve_accelerate_config_argument` now parses the equals form, strips one token instead of two when appropriate, and raises the same error when the value after `=` is empty. Regression tests cover both spellings on the `trl sft` path (mocked launcher) and the empty `--accelerate_config=` case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1428d80324b81b6d58252264e053a29bcd6ebb1a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


behroozazarkhalili:
I went through this against #6874 and checked it by running it rather than only reading it.

The root cause is what you describe. On unmodified main, `resolve_accelerate_config_argument` tests membership with `"--accelerate_config" not in launch_args` and then calls `launch_args.index(...)`. Both are exact-string comparisons, so the `=` spelling can never match:

```
two-token form : ['--config_file', '.../accelerate_configs/single_gpu.yaml', '--num_processes', '2']
equals form    : ['--accelerate_config=single_gpu', '--num_processes', '2']
```

With your patch both forms produce identical output. I also ran the cases around the edges:

| input | result |
|---|---|
| `--accelerate_config single_gpu` | resolved |
| `--accelerate_config=single_gpu` | resolved |
| `--accelerate_config=single_gpu` as the last argument | resolved |
| `--accelerate_config` with no value | `ValueError` |
| `--accelerate_config=` | `ValueError` |
| `--accelerate_config=nope` | `ValueError`, unknown config |
| neither form present | returned unchanged |
| `--accelerate_config_x=v` | left alone, not matched |

Red/green on your test: reverting only `trl/cli/accelerate_config.py` to main while keeping the new test leaves `accelerate_config_args1` (the equals case) failing and `accelerate_config_args0` passing. That is the shape you want, since the two-token case acts as a working negative control. Ruff 0.13.3 check and format and the pinned doc-builder are clean on both files.

Driving the test through `main()` with a patched `sys.argv` is the right level, and it covers something a unit test of the function alone would miss. I confirmed that argparse keeps `--accelerate_config=single_gpu` as a single token in its remainder list, and `trl/cli/commands/training.py:64` hands those remainders straight to the resolver. So the bug is reachable from a real command line, not only from a hand-built list.

One gap in the test. The `if not config_name: raise` branch is uncovered: I deleted those two lines and both of your cases still passed. Dropping the guard is not unsafe, because `--accelerate_config=` still raises, but it falls through to the other error path and the message degrades to ``Accelerate config  is neither a file nor a valid config in the `trl` package.`` A third case would pin the message you intended:

```python
def test_accelerate_config_requires_value():
    with pytest.raises(ValueError, match="Expected a value"):
        resolve_accelerate_config_argument(["--accelerate_config="])
```

Separately, and outside this PR's scope: `trl/scripts/utils.py:321` has the same shape for `--config`, and there the outcome is worse because nothing raises. Calling `TrlParser.parse_args_and_config` the way the CLI calls it:

```
['--config', 'probe_cfg.yaml']   lr=0.5     launch_args=[]
['--config=probe_cfg.yaml']      lr=1e-05   launch_args=['--config=probe_cfg.yaml']
```

The YAML is silently ignored, the dataclass keeps its default, and the token is handed on to `accelerate launch` as an unrecognized flag. I can file that separately if you would rather keep this PR to the one file.

The branch is 45 commits behind main, so merging main in would let CI report against current code.

PR #6882, opened 2026-08-24
Quote python-version value in the tests workflow

albertvillanova:
This PR quotes the `python-version` value in the tests workflow, making it consistent with every other `setup-python` usage in the repository.

Follow-up to #6881.

### Motivation

In YAML, an unquoted `3.12` is parsed as a float rather than a string. It happens to work here, but it is the classic pitfall behind versions such as `3.10` silently becoming `3.1`. Every other `setup-python` usage in the repository already quotes the value.

### Changes

- Quote the `python-version` value in the code quality job of the tests workflow, the only remaining unquoted occurrence

---

> [!NOTE]
> **Low Risk**
> One-line CI YAML quoting change with no runtime or security impact.
> 
> **Overview**
> Quotes `python-version` as `'3.12'` in the `check_code_quality` job of `tests.yml`.
> 
> This matches other `setup-python` usages and avoids YAML parsing the version as a float.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05fec6cab1112ff2011a618ef6ef45798f9e5d36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6881, opened 2026-08-24
Align Python version in the experimental tests workflow

albertvillanova:
This PR aligns the Python version used by the experimental tests workflow with the one used by all other test workflows.

Follow-up to #6763.

### Motivation

#6763 established a single Python version for PR CI: PRs test 3.12 only, and the other supported versions (3.10, 3.11, 3.13, 3.14) moved to the nightly `tests_python_versions.yml` matrix on `main`.

The experimental tests workflow was not updated to follow that policy and still runs on Python 3.13, making it the only PR workflow that does not use 3.12. It has been on 3.13 since it was introduced in #4330, without a stated reason.

No coverage is lost: Python 3.13 remains covered by the nightly matrix.

### Changes

- Set the Python version to 3.12 in the code quality and tests jobs of the experimental tests workflow
- Quote the version value, consistent with the other workflows and avoiding the YAML float pitfall

---

> [!NOTE]
> **Low Risk**
> CI-only Python version pin with no application, security, or data-handling changes.
> 
> **Overview**
> Switches `tests-experimental.yml` code quality and test jobs from Python **3.13** to **3.12**, matching other PR CI workflows.
> 
> The version is quoted as `'3.12'` to avoid YAML treating it as a float. Python 3.13 coverage stays on the nightly matrix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 874d0b8703df44c3e7ce927028ec162041f982b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6763, opened 2026-08-14
Test one Python version on PRs, matrix nightly

qgallouedec:
Every PR ran the test suite on 3.10, 3.11, 3.12, 3.13 and 3.14  and historically this caught almost no version-specific failures. TRL is pure Python; the version-sensitive work happens in torch/transformers.

- PRs now test 3.12 only (most used, per telemetry), like every other job in `tests.yml`.
- New nightly `tests_python_versions.yml` runs 3.10/3.11/3.13/3.14 on `main`, with Slack reporting and `workflow_dispatch` (handy before cutting a release).

Version-specific *syntax* regressions stay covered on PRs by ruff's `target-version = "py310"`.

Screenshot 2026-08-14 at 4 02 11 PM

---

> [!NOTE]
> **Low Risk**
> CI-only change. Version-specific bugs on non-3.12 Pythons will no longer be caught on PRs, only nightly.
> 
> **Overview**
> PR CI no longer matrixes Python 3.10–3.14. The main `tests` job now runs only **3.12**, matching the other jobs in `tests.yml`.
> 
> A new scheduled workflow (`tests_python_versions.yml`) covers **3.10, 3.11, 3.13, and 3.14** daily (plus `workflow_dispatch`), with Slack reporting. Syntax compatibility for 3.10 remains on PRs via ruff `target-version`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c3381b7ce99a2f837bcc9b81a5e37b2bdb8dfb9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6883, opened 2026-08-24
Pin Python version in the publish workflow

albertvillanova:
This PR pins the Python version used to build and publish the package, instead of tracking whatever the runner image provides.

Follow-up to #6881 and #6882.

### Motivation

The publish workflow requested `3.x`, which resolves to the newest Python available on the GitHub runner image. That version changes without any action on our side, so a runner image update can silently change the interpreter that builds a release, along with the `pip`, `build` and `setuptools` versions that come with it. A release build is the last place where we want an environment we did not choose.

Pinning to 3.12 also makes the workflow consistent with the rest of CI, where 3.12 is the single version used since #6763.

TRL is a pure Python package built with setuptools, so the resulting wheel is `py3-none-any` and is not affected by the interpreter used to build it. This change only makes the build environment predictable.

### Changes

- Pin the Python version to 3.12 in the publish workflow
- Name the setup step after the version it installs, consistent with the other workflows

---

> [!NOTE]
> **Low Risk**
> CI-only pin of the Python version used to build a pure-Python wheel; no application, auth, or package-content changes.
> 
> **Overview**
> Pins the PyPI publish workflow to **Python 3.12** instead of `3.x`, so the interpreter used to build and upload packages no longer tracks whatever the runner image currently ships.
> 
> The setup step is renamed to `Set up Python 3.12` to match other CI workflows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec6271af74426eed4b71258c2e83c1bd34cf46f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,10 +21,10 @@ jobs:
       - name: Debug - Show version.txt content
         run: echo "Version is ${{ steps.get_version.outputs.version }}"
 
-      - name: Set up Python
+      - name: Set up Python 3.12
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
         with:
-          python-version: "3.x"
+          python-version: '3.12'
 
       - name: Install dependencies
         run: |


qgallouedec (approved):
thanks, looks good

PR #6885, opened 2026-08-24
Pin the post-slack action to a commit SHA

albertvillanova:
This PR pins the `post-slack` action to a commit SHA everywhere it is used, so that no workflow resolves a third party action through a mutable ref.

Follow-up to #5435.

### Motivation

`#5435` pinned GitHub Actions to commit SHAs, but it only covered 8 workflow files and left the test workflows out of scope. As a result `huggingface/hf-workflows/.github/actions/post-slack@main` was still resolved through a moving branch in 11 places, and it was the only unpinned action reference left in the repository. Every other reference is already a 40 character SHA.

This matters because the action runs in workflows that hold `SLACK_CIFEEDBACK_BOT_TOKEN` and `HF_TOKEN`. Any commit landing on the `hf-workflows` default branch changes what those jobs execute, with no review or signal on our side.

### Solution

Pin the 11 remaining references to `a88e7fa2eaee28de5a4d6142381b1fb792349b67`, the SHA already used by the docker build workflow since #5435. All 13 references now share a single value, and the docker build workflow is left untouched.

This is a no-op in behaviour: the 8 commits between this SHA and the current tip of `hf-workflows` `main` do not touch `.github/actions/post-slack`, so the action code is identical either way.

The action pins its own `slackapi/slack-github-action` dependency by SHA, so the whole chain becomes immutable.

### Changes

- Pin the `post-slack` action to a commit SHA in the tests, slow tests, latest, Python versions and transformers branch workflows

---

> [!NOTE]
> **Low Risk**
> CI-only pin of an existing Slack notification action; no application or test logic changes. Residual risk is that the chosen SHA must remain the intended `post-slack` revision.
> 
> **Overview**
> Pins `huggingface/hf-workflows` `post-slack` from the mutable `@main` ref to commit `a88e7fa2eaee28de5a4d6142381b1fb792349b67` in the remaining test workflows (slow tests, tests, latest, Python versions, transformers branch).
> 
> This is a supply-chain hardening follow-up: those jobs pass Slack tokens and should not pick up unreviewed changes from `hf-workflows` `main`. Behavior is intended to be a no-op if that SHA is the current tip of `main`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e04dcc0e758b41451e8c6e6dbd6bf195e825e1f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -54,7 +54,7 @@ jobs:
 
       - name: Post to Slack
         if: always()
-        uses: huggingface/hf-workflows/.github/actions/post-slack@50acccfe9afb8927b9b74ed7e750f2b0d3991b62  # main
+        uses: huggingface/hf-workflows/.github/actions/post-slack@a88e7fa2eaee28de5a4d6142381b1fb792349b67  # main
         with:
           slack_channel: ${{ secrets.CI_DOCKER_CHANNEL }}
           title: 🤗 Results of the TRL Dev Docker Image build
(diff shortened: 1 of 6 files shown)


qgallouedec (approved):
Superseded by #6891 and #6892, right? If I'm missing something, the diff itself looks good, so I'm still approving.

Issue #6890, opened 2026-08-24
CI warning: Node.js 20 is deprecated (slackapi/slack-github-action in hf-workflows post-slack)


albertvillanova:
### What

CI emits a new `Node.js 20 is deprecated` annotation in every job that posts results to Slack:

```
Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: slackapi/slack-github-action@6c661ce58804a1a20f6dc5fbee7f0381b469e001. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

CI run: https://github.com/huggingface/trl/actions/runs/32630307657

The annotation comes from the `Post to Slack` step, which uses `huggingface/hf-workflows/.github/actions/post-slack`. That composite action calls `slackapi/slack-github-action@6c661ce58804a1a20f6dc5fbee7f0381b469e001` (= `v1.25.0`), whose `action.yml` declares `runs.using: 'node20'`.

Affected workflows (13 `Post to Slack` steps in total):

- `.github/workflows/tests.yml` — `tests`, `tests_dev`, `tests_wo_optional_deps`, `tests_min_versions`, `distributed_smoke`
- `.github/workflows/slow-tests.yml` — `run_all_tests_single_gpu`, `run_all_tests_multi_gpu`
- `.github/workflows/tests_transformers_branch.yml` — `tests_transformers_branch`, `distributed_smoke`
- `.github/workflows/docker-build.yml` — `trl`, `trl-dev`
- `.github/workflows/tests_latest.yml` — `tests`
- `.github/workflows/tests_python_versions.yml` — `tests`

The jobs themselves pass and the Slack messages are still posted; for now this is warning noise only.

### When it started

This is not caused by one of our commits nor by a dependency release: it comes from a change on the GitHub Actions runners.

Per [the GitHub changelog](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/):

- Node 20 reached end-of-life in April 2026.
- Since **2026-06-16**, runners force actions that declare `node20` to run on Node 24, emitting this annotation.
- In **fall 2026**, GitHub plans to upgrade the runner and remove Node 20 entirely. The `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true` opt-out only works until then.

[...]

PR #6891, opened 2026-08-24
Pin the mutable post-slack action reference to a commit SHA

albertvillanova:
This PR replaces the mutable `@main` reference to the `huggingface/hf-workflows/.github/actions/post-slack` action with a commit SHA, so that all thirteen call sites are pinned to the same SHA already used by the existing pinned ones.

Related to #6890, which it does not resolve: this change is behavior neutral, so the `Node.js 20 is deprecated` annotation still appears.

### Motivation

`post-slack` was the only action in our CI still referenced by a mutable ref. Eleven of the thirteen call sites used `@main`, which is resolved at run time, while every other action in our workflows (`actions/checkout`, `actions/setup-python`, `astral-sh/setup-uv`, `docker/*`, `trufflesecurity/trufflehog`, `github/codeql-action`, `huggingface/doc-builder`) is pinned to a SHA. Since these steps receive `secrets.SLACK_CIFEEDBACK_BOT_TOKEN`, any push to the `hf-workflows` default branch takes effect immediately in our jobs.

### Solution

Pin the eleven `@main` references to `a88e7fa2eaee28de5a4d6142381b1fb792349b67`, the SHA the two other call sites already use, following the `# main` comment convention also used for `huggingface/doc-builder`.

Keeping the same SHA as the existing pinned references makes this PR purely about removing the mutable ref, with no change to the action code being run. Bumping that SHA is left out on purpose, so that the two concerns can be reviewed separately.

Note that pinning is not a guarantee that the reference stays frozen: Dependabot does bump SHA references that track a branch, including in repositories without matching tags. It has been doing so for our `huggingface/doc-builder` references, most recently in #6727. It has never bumped `post-slack`, most likely because Dependabot does not track composite actions living in a repository subdirectory, but that is an implementation detail rather than something to rely on.

### Changes

- Pin the eleven `post-slack@main` references in `slow-tests.yml`, `tests.yml`, `tests_latest.yml`, `tests_python_versions.yml` and `tests_transformers_branch.yml` to `a88e7fa2eaee28de5a4d6142381b1fb792349b67`

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -54,7 +54,7 @@ jobs:
 
       - name: Post to Slack
         if: always()
-        uses: huggingface/hf-workflows/.github/actions/post-slack@50acccfe9afb8927b9b74ed7e750f2b0d3991b62  # main
+        uses: huggingface/hf-workflows/.github/actions/post-slack@a88e7fa2eaee28de5a4d6142381b1fb792349b67  # main
         with:
           slack_channel: ${{ secrets.CI_DOCKER_CHANNEL }}
           title: 🤗 Results of the TRL Dev Docker Image build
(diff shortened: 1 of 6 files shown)


albertvillanova:
Merging this PR as a security fix.

PR #5178, opened 2026-02-25
Handle mm_token_type_ids in SFT/GRPO/RLOO to fix IndexError

albertvillanova:
Handle mm_token_type_ids in SFT/GRPO/RLOO to fix IndexError: concatenate and truncate
- Handle `mm_token_type_ids` in SFT `_collate_prompt_completion`
- Handle `mm_token_type_ids`  in GRPO `_generate_and_score_completions` and `_compute_loss`
- Handle `mm_token_type_ids`  in RLOO `_generate_and_score_completions` and `_compute_loss`

This PR enhances the `_collate_prompt_completion` method in SFT trainer to support models with multimodal token type IDs, such as ERNIE-VL. The changes ensure that `mm_token_type_ids` are properly handled during collation, padding, and output construction.

Fix #5177.

### Changes

Support for multimodal token type IDs:

* Added logic to concatenate `mm_token_type_ids` from prompts and completions, and to flush left multimodal token type IDs along with other tensors when present.
* Ensured `mm_token_type_ids` are truncated to the maximum sequence length during collation.
* Included `mm_token_type_ids` in the output dictionary when present in the processed prompts.

PR #6899, opened 2026-08-24
Remove the workflow testing the latest release with dev dependencies

albertvillanova:
This PR removes the `Tests latest TRL release with dev dependencies` workflow.

### Motivation

The workflow ran the test suite of the latest release branch against the development versions of `accelerate`, `datasets` and `transformers`, daily. In practice it has never reported anything:

- Its `Post to Slack` step has no `if:`, so it inherits `if: success()`. It could only ever post `🟢 Tests are passing!`
- 29 of its last 100 runs failed, including six consecutive days from 2026-07-23 to 2026-07-28, and none of them produced a Slack notification

So the signal it produced was discarded, and the workflow spent a GPU runner every night to report only its successes.

The failures it did produce were also not the kind it exists to catch. The 2026-07-24 one is representative:

```
FAILED tests/test_sft_trainer.py::TestSFTTrainer::test_train_with_liger
  RuntimeError: gemm input type at::BFloat16 and output type float is only
  supported for CUDA devices with compute capability 8.0 or higher
```

That is the T4 of `aws-g4dn-2xlarge`, which has compute capability 7.5. The workflow is on that runner group rather than the L40S used elsewhere precisely because it checks out an old release branch, so its results diverge from the rest of our CI and it reports that divergence as failures.

The remaining coverage overlap is partial but acceptable: `tests.yml` already runs the same three dependencies from git in its `Tests with dev dependencies` job. What we lose is the daily cadence, since that job only runs on pushes, and coverage of the released branch rather than `main`. Both can come back later in a form that actually notifies someone, if we find we miss them.

### Changes

- Remove `.github/workflows/tests_latest.yml`
- Drop the corresponding step from the release process in `RELEASE.md`, which asked to bump the pinned release branch on every release

---

> [!NOTE]
> **Low Risk**
> CI and docs only: no product, auth, or data-path changes. Slightly less scheduled coverage of released branches against upstream git deps.
> 
> **Overview**
> Removes the daily `tests_latest.yml` workflow that ran the latest TRL *release* branch against git versions of `accelerate`, `datasets`, and `transformers`.
> 
> That job used a GPU runner every night, Slack-notified only on success, and often failed for runner/hardware reasons rather than real compatibility issues. Overlapping coverage remains in `tests.yml` on `main` (dev deps, push-triggered).
> 
> `RELEASE.md` no longer asks maintainers to bump the workflow’s pinned `v{major}.{minor}-release` ref.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23bce1b26b26b58fdf2e512e2e88775b1b21953a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:
Opened #6900 as the alternative: it fixes the missing `if: always()` on the Slack step instead of removing the workflow, so the signal reaches someone and can be judged with data. The two PRs are mutually exclusive.

For whoever picks: #6899 if the nightly GPU runner is not worth a signal that has never been read, #6900 if it is worth a few weeks of real notifications before deciding.

Note that #6895 depends on this one being merged: it leaves `tests_latest.yml` on the shared `post-slack` action, since that job checks out `v1.10-release`, which predates the local action. If #6900 is chosen over #6899 instead, #6895 needs the extra checkout described in https://github.com/huggingface/trl/pull/6895#discussion_r3844151402 restored for that one call site.

qgallouedec:
Thanks for the detailed analysis, but I'd like to keep this workflow.

It was created in reaction to a real incident (it seems that I didn't document it at this time unfortunately). huggingface/transformers#34507 (merged 2024-11-19) added a `start_time` arg to `Trainer.log()`, and when it shipped in transformers 4.47.0 on 2024-12-05 it broke the released TRL for everyone who upgraded (massive blast radius). A user found out before we did (#2445, filed 2024-12-06), and we had to ship an emergency pin the same day (#2447, released as v0.12.2 on 2024-12-06). This workflow was added four days later (#2457, 2024-12-10) precisely so that users are not our canary.
So emergency on 2024-12-06, but we could have catch it as soon as 18 days earlier!

It has done its job since, and it catches a class of breakage that the `Tests with dev dependencies job on main` cannot catch by design. That job tests main, and main gets fixed within days, so its CI goes back to green while the released version stays broken. It basically catch what user **will** face when transformers release. So it's a good proxy to know what's needed in the next patch release.

In Feb 2026, huggingface/transformers#43972 (mm_token_type_ids, merged 2026-02-24) broke us. Main was fixed two days later (#5178, merged 2026-02-26) and main-dev went back to green, but users on v0.29.0 hit the `IndexError` for real once transformers 5.3.0 shipped on 2026-03-04. For the 16 days until v0.29.1 delivered the fix on 2026-03-20, this nightly was the only CI reporting the breakage.

There are probably other occurences but it's hard to identify them, it's mostly on top of my head.

albertvillanova:
Superseded by:
- #6900

PR #6903, opened 2026-08-25
Post the Docker image build results to the shared CI Slack channel

albertvillanova:
This PR sends the Docker image build results to the shared CI Slack channel, like every other workflow.

### Motivation

`docker-build.yml` was the only workflow pointing to its own channel, as noted in revision https://github.com/huggingface/trl/pull/6895#discussion_r3844347430. Every other workflow declares `CI_SLACK_CHANNEL` once at the workflow level and passes it at each call site, so the thirteen call sites did not all resolve the channel the same way.

Note that this changes where those notifications are delivered.

### To do on merge

- [ ] Delete the now unused `CI_DOCKER_CHANNEL` repository secret. It is no longer referenced anywhere once this PR is merged, and deleting it is a repository setting rather than something this PR can do.

### Changes

- Declare `CI_SLACK_CHANNEL` at the workflow level in `docker-build.yml`, from the same secret as the other workflows
- Use it in both `Post to Slack` steps, in place of the Docker specific channel

---

> [!NOTE]
> **Low Risk**
> Only changes which Slack channel receives CI notifications; no runtime or build logic changes.
> 
> **Overview**
> Aligns **Docker image build** Slack notifications with the rest of the repo’s CI workflows.
> 
> **`docker-build.yml`** now sets workflow-level `CI_SLACK_CHANNEL` from `CI_PUSH_MAIN_CHANNEL` (same pattern as workflows like `slow-tests.yml`). Both `Post to Slack` steps (`trl` and `trl-dev`) pass `env.CI_SLACK_CHANNEL` instead of `CI_DOCKER_CHANNEL`, so build results land in the **shared CI channel** rather than the Docker-only channel. After merge, `CI_DOCKER_CHANNEL` can be removed from repo secrets.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80e42c37f50aa0347a276227bd630309be5b705b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6904, opened 2026-08-25
Read the Slack channel from the environment in the local Slack action

albertvillanova:
This PR moves the Slack channel from the call sites into the local Slack action.

Stacked on top of #6903. Please review and merge that one first: this PR targets its branch and the base will switch to `main` once it lands.

### Motivation

Once #6903 lands, all thirteen call sites pass the very same `slack_channel: ${{ env.CI_SLACK_CHANNEL }}`. Repeating an identical value thirteen times is what the action exists to avoid, as suggested while reviewing #6895.

### Solution

Drop the `slack_channel` input and read `CI_SLACK_CHANNEL` from the environment inside the action, which every workflow using it already declares at the workflow level. A comment in the action records that expectation.

The action already read a job scoped environment variable this way, for the message built by its first step, so the channel resolves through the same mechanism.

Only the channel is moved. The rest of the action is left exactly as it was copied from the shared one in #6895.

### Changes

- Drop the `slack_channel` input from the action and read the channel from `CI_SLACK_CHANNEL`
- Remove the input from the thirteen call sites in `docker-build.yml`, `slow-tests.yml`, `tests.yml`, `tests_latest.yml`, `tests_python_versions.yml` and `tests_transformers_branch.yml`

---

> [!NOTE]
> **Low Risk**
> CI notification wiring only; behavior is unchanged as long as workflows keep declaring `CI_SLACK_CHANNEL`, with slightly weaker explicit coupling at call sites.
> 
> **Overview**
> Centralizes Slack destination configuration in the composite **`post-slack`** action instead of repeating it at every workflow step.
> 
> The action **drops** the `slack_channel` input and sets the Block Kit payload’s `channel` from **`${{ env.CI_SLACK_CHANNEL }}`**, with comments noting that callers must define that workflow-level env (already mapped from `secrets.CI_PUSH_MAIN_CHANNEL`). **Thirteen** `Post to Slack` steps across `docker-build.yml`, `slow-tests.yml`, `tests.yml`, `tests_latest.yml`, `tests_python_versions.yml`, and `tests_transformers_branch.yml` no longer pass `slack_channel`; they still pass `title`, `status`, and `slack_token` only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec6f85107933944d38ac419edc3ed6f32ed64ec8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/actions/post-slack/action.yml
+++ b/.github/actions/post-slack/action.yml
@@ -4,29 +4,42 @@ description: 'Send results to slack'
 author: 'Hugging Face'
 inputs:
   slack_channel:
-    description: 'Slack channel id, channel name, or user id to post the message to'
     required: true
+    type: string
   title:
-    description: 'Title of the message'
     required: true
+    type: string
   status:
-    description: 'Status of the job, as reported by `job.status`'
     required: true
+    type: string
   slack_token:
-    description: 'Slack bot token'
     required: true
+    type: string
 
 runs:
   using: "composite"
   steps:
-    - name: Post results to Slack
+    - name: Create content to post
+      id: create-message
+      run: |
+        if [ "${{ inputs.status }}" == "success" ]; then
+          echo STATUS_MESSAGE='🟢 Tests are passing!' >> $GITHUB_ENV
+        else
+          echo STATUS_MESSAGE='🔴 Tests failed! Please check the GitHub action link below' >> $GITHUB_ENV
+        fi
+      shell: bash
+
+    - name: Post Canceled results Slack channel
+      id: post-slack
       uses: slackapi/slack-github-action@dcb1066f776dd043e64d0e8ba94ca15cc7e1875d  # v4.0.0
       with:
         method: chat.postMessage
         token: ${{ inputs.slack_token }}
         # Fail the step when the Slack API call fails, as v1 did by default
         errors: true
-        # For posting a rich message using Block Kit
+        # For posting a rich message using Block Kit. The channel is a Slack
+        # channel id, channel name, or user id to post message to.
+        # See also: https://api.slack.com/methods/chat.postMessage#channels
         payload: |
           {
             "channel": "${{ inputs.slack_channel }}",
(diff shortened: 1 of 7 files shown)


albertvillanova:
Validated on a dispatch of \`tests_latest.yml\` against this branch, rather than waiting for the merge: https://github.com/huggingface/trl/actions/runs/32828542104 (conclusion \`success\`).

\`\`\`
10. Check out the Slack action      : success
11. Post to Slack                   : success
18. Post Check out the Slack action : success
    Test with pytest                : success
\`\`\`

The interesting part is not that the step passed, but that the log shows the rendered payload as:

\`\`\`
  "channel": "***",
\`\`\`

Secret values are masked in logs, and \`CI_SLACK_CHANNEL\` comes from \`secrets.CI_PUSH_MAIN_CHANNEL\`, so \`***\` means a non-empty secret value was substituted. Had the \`env\` context not resolved inside the composite action, that line would read \`"channel": ""\` in the clear. Together with \`method: chat.postMessage\` and \`errors: true\` completing without an error, the message really was posted.

The log also confirms the action now receives only \`title\`, \`status\` and \`slack_token\`, and that everything else is unchanged: the two step form and the \`Post Canceled results Slack channel\` name are still there, so the revert did its job and only the channel moved.

This also happens to be the first run of the \`./slack-action/...\` checkout path added in #6895, which had not been exercised yet: the nightly last ran before that PR merged.

One caveat: the test suite passed, so this exercised the 🟢 branch of the status message and not the 🔴 one. That code is untouched here.

qgallouedec:
No strong opinion, but I'd rather keep the channel at the call site. It trades a declared input for an undeclared env contract: a workflow that forgets `CI_SLACK_CHANNEL` posts `"channel": ""` and we find out on the red run we wanted the notification for.

And `slack_token` is identical at the same thirteen call sites but stays an input. It also comes from a secret. Either both move to env or neither, no?

Mechanism is fine btw, composite steps do get the env context. My objection is mostly the interface.

qgallouedec (approved):
I'll approve though, mostly your call

albertvillanova:
Closing this one, keeping the channel at the call site as you preferred.

Your correction is right and I will fix the #6908 description accordingly: the secrets context does not force `slack_token` to be an input. Keeping it there is a scope decision, since workflow level env would put a credential in the environment of every step of every job, `make test` included, while today it only exists in the inputs of one step. That argument does not apply to a channel id, so with this closed both stay inputs and the interface stays uniform.

Rebasing #6908 on main.

PR #6908, opened 2026-08-25
Read the job status from the job context in the local Slack action

albertvillanova:
This PR moves the job status from the call sites into the local Slack action.

### Motivation

All thirteen call sites pass the very same `status: ${{ job.status }}`, which is the only value the action needs that is a property of where it runs rather than something to configure. It can be read inside the action instead of being handed in thirteen times.

The other inputs stay at the call site: `title` differs every time, `slack_channel` is configuration, and `slack_token` is a credential that is better kept in the inputs of a single step than in the environment of every step of every job.

### Solution

Drop the `status` input and read `job.status` in the action.

A composite action runs inline in the job that calls it, so the job context it sees is the caller's, evaluated when the step runs. That is exactly what the call sites were passing, so the message is unchanged.

Note that this also drops the ability for a call site to report something other than its own job status. Nothing does that today, and the input can come back if we ever need to notify on a combination of jobs.

### Changes

- Drop the `status` input from the action and read `job.status` instead
- Remove the input from the thirteen call sites in `docker-build.yml`, `slow-tests.yml`, `tests.yml`, `tests_latest.yml`, `tests_python_versions.yml` and `tests_transformers_branch.yml`

---

> [!NOTE]
> **Low Risk**
> CI-only Slack notification wiring with equivalent behavior; no application or security logic changes.
> 
> **Overview**
> The **`post-slack`** composite action no longer takes a **`status`** input. It reads **`job.status`** in the “Create content to post” step when choosing the green vs red Slack message, instead of relying on callers to pass the same value every time.
> 
> All **13** “Post to Slack” steps across **`docker-build.yml`**, **`slow-tests.yml`**, **`tests.yml`**, **`tests_latest.yml`**, **`tests_python_versions.yml`**, and **`tests_transformers_branch.yml`** drop **`status: ${{ job.status }}`**. **`title`**, **`slack_channel`**, and **`slack_token`** are unchanged at the call sites.
> 
> Because composite steps run in the calling job’s context, notifications should match what they did before; callers can no longer override status with a different value (nothing did that today).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c37939c2aa27473d8adcdf14ee4157703a981ff5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/actions/post-slack/action.yml
+++ b/.github/actions/post-slack/action.yml
@@ -9,9 +9,6 @@ inputs:
   title:
     required: true
     type: string
-  status:
-    required: true
-    type: string
   slack_token:
     required: true
     type: string
@@ -22,7 +19,7 @@ runs:
     - name: Create content to post
       id: create-message
       run: |
-        if [ "${{ inputs.status }}" == "success" ]; then
+        if [ "${{ job.status }}" == "success" ]; then
           echo STATUS_MESSAGE='🟢 Tests are passing!' >> $GITHUB_ENV
         else
           echo STATUS_MESSAGE='🔴 Tests failed! Please check the GitHub action link below' >> $GITHUB_ENV
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -59,7 +59,6 @@ jobs:
         with:
           slack_channel: ${{ env.CI_SLACK_CHANNEL }}
           title: 🤗 Results of the TRL Docker Image build
-          status: ${{ job.status }}
           slack_token: ${{ secrets.SLACK_CIFEEDBACK_BOT_TOKEN }}
 
   trl-dev:
@@ -93,5 +92,4 @@ jobs:
         with:
           slack_channel: ${{ env.CI_SLACK_CHANNEL }}
           title: 🤗 Results of the TRL Dev Docker Image build
-          status: ${{ job.status }}
           slack_token: ${{ secrets.SLACK_CIFEEDBACK_BOT_TOKEN }}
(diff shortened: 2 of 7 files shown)


albertvillanova:
Validated with a dispatch of \`tests_latest.yml\` against this branch: https://github.com/huggingface/trl/actions/runs/32837942027 (conclusion \`success\`).

\`\`\`
10. Check out the Slack action      : success
11. Post to Slack                   : success   (554 ms)
18. Post Check out the Slack action : success
    Test with pytest                : success
\`\`\`

The `job` context is available inside a composite action, which the metadata syntax reference does not state either way. The log shows the shell step taking the success branch:

```
  STATUS_MESSAGE: 🟢 Tests are passing!
```

and the rendered payload carrying it:

```
        "text": "🟢 Tests are passing!"
```

A green step on its own would not have proved this. Had `job.status` resolved to an empty string, `[ "" == "success" ]` would have been false and the message would have been `🔴 Tests failed!`. The test suite passed in this run, so `🟢` is the correct value rather than a coincidence, and there is no `Unrecognized named-value` error, which is what an unavailable context would have produced.

Same caveat as #6904: this exercised the `🟢` branch. The `🔴` branch is the same comparison with the opposite outcome, so the context resolution it depends on is what was just verified.

qgallouedec (approved):
lgtm (more than #6904). `job.status` has only one correct value at the call site, so deriving it in the action beats passing it thirteen times.

Small correction: `slack_token` doesn't have to stay an input because of the secrets context. #6904 moves the channel to env and that's a secret too

If I don't take #6904 this needs a rebase on main.

PR #6889, opened 2026-08-24
Build the GRPO VLM test images as PIL images to avoid a datasets downcast warning

albertvillanova:
This PR removes the `datasets` downcast `UserWarning` emitted by the slow GRPO VLM training test.

Fix #6888.

### Motivation

Since the slow tests moved from T4 to L40S in #6741, `test_vlm_training` is no longer skipped by its Ampere-or-newer gate, and CI emits:

```
tests/test_grpo_trainer.py::TestGRPOTrainerSlow::test_vlm_training[HuggingFaceTB/SmolVLM-Instruct]
  /__w/trl/trl/.venv/lib/python3.12/site-packages/datasets/features/image.py:397: UserWarning: Downcasting array dtype int64 to uint8 to be compatible with 'Pillow'
    warnings.warn(f"Downcasting array dtype {dtype} to {dest_dtype} to be compatible with 'Pillow'")
```

The array the test builds is already `uint8`. Since #5550 the dataset is built with `Dataset.from_list` and no feature spec, so the images are stored as nested lists and `cast_column("image", Image())` reads them back as Python integers, which `numpy` turns into `int64`, prompting `datasets` to warn about a downcast the data never needed.

The misleading message is reported upstream in https://github.com/huggingface/datasets/issues/8511, but the round-trip is avoidable on our side.

### Solution

Build the test images as PIL images. `datasets` then encodes them when the dataset is created, so the cast to the `Image` feature takes the struct path and no array dtype is ever inferred. This works with every supported `datasets` version, down to the `>=4.7.0` floor, and matches the way other tests in this file build images.

### Changes

- Build the images of the GRPO VLM slow test with `PIL.Image.fromarray` instead of passing raw arrays, with a comment explaining why

---

> [!NOTE]
> **Low Risk**
> Test-only fixture change with no production, security, or training-logic impact.
> 
> **Overview**
> Stops a `datasets` downcast `UserWarning` in the slow GRPO VLM training test by wrapping synthetic images with `PIL.Image.fromarray` before `Dataset.from_list` / `cast_column("image", Image())`.
> 
> Raw uint8 arrays were stored as nested lists and read back as int64, so the Image feature warned about a downcast that was never needed. Encoding as PIL at construction time avoids that round-trip.
> 
[...]

PR #6909, opened 2026-08-25
Add a QLoRA test to the GRPO and RLOO VLM test suites

albertvillanova:
This PR adds a QLoRA test to the GRPO and RLOO VLM test suites.

Follow-up of the review comment in #6889.

### Motivation

QLoRA training is currently only covered by the slow `test_vlm_training`, which is `xfail` because of the Flash Attention 2 kernel issues, and which tests quantization, kernels, bf16 and LoRA all at once. Removing it (#6912) needs a focused replacement first.

Two code paths are involved, and are not covered by any other test:

- the detection of a quantized model, which decides whether `autocast_adapter_dtype=False` is passed to `get_peft_model`
- the cast of the adapter weights to bf16 for quantized models, following the QLoRA paper recommendation

Both are duplicated identically in GRPO and RLOO, so the test is added to both.

The `quantization_config` trainer argument is also untested, in every trainer that exposes it: all existing quantization tests pass the config to `from_pretrained` instead. Since it is the documented way to do QLoRA training ("Combine with `peft_config` for QLoRA training"), the new test uses it.

### Solution

Add a small QLoRA test next to `test_train_vlm_peft` in each trainer, with the same tiny model, dataset and training config, so that quantization is the only difference between the two. The GRPO and RLOO copies are identical modulo the trainer name.

The test does not check that the base model params are unchanged, unlike the equivalent tests in SFT, DPO and KTO: bitsandbytes casts the biases of a `Linear4bit` in-place during the forward pass, and those tests keep a hardcoded list of the affected biases to work around it. That list depends on which layers ran a forward pass, which is not predictable in GRPO and RLOO, since generation runs before training.

### Changes

- Add a QLoRA test to the GRPO VLM test suite, using the `quantization_config` trainer argument together with `peft_config`
- Add the same test to the RLOO VLM test suite

---

> [!NOTE]
> **Low Risk**
> Test-only changes with no production trainer or runtime logic modified.
> 
> **Overview**
> Adds focused **QLoRA** integration tests for vision-language training on **GRPO** and **RLOO**, mirroring the existing `test_train_vlm_peft` setup but loading the model with **`quantization_config`** (4-bit BitsAndBytes) plus **`peft_config`** (LoRA on `q_proj` / `v_proj`).
> 
> Each new test runs a short train step on the tiny Qwen2.5-VL fixture, asserts the model is loaded in 4-bit, checks that **`train_loss`** is logged, and verifies **LoRA weights update** and stay in **bfloat16** as expected for QLoRA. Base quantized weights are intentionally not compared to pre-training snapshots because bitsandbytes can mutate some biases during forward passes.
> 
> **RLOO** test file also wires in `BitsAndBytesConfig` and the `@require_bitsandbytes` marker for the new case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a55d9e369b48aca918e646079dfae14849e2fc21. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -4098,6 +4098,60 @@ class TestGRPOTrainerVLM(TrlTestCase):
             "trl-internal-testing/tiny-Qwen2_5_VLForConditionalGeneration",
         ],
     )
+    @require_peft
+    @require_bitsandbytes
+    def test_train_vlm_peft_and_quantization(self, model_id):
+        dataset = load_dataset("trl-internal-testing/zen-image", "conversational_prompt_only", split="train")
+
+        def reward_func(completions, **kwargs):
+            """Reward function that rewards longer completions."""
+            return [float(len(completion[0]["content"])) for completion in completions]
+
+        training_args = GRPOConfig(
+            output_dir=self.tmp_dir,
+            learning_rate=0.1,  # use higher lr because gradients are tiny and default lr can stall updates
+            per_device_train_batch_size=2,  # VLM training is memory intensive, reduce batch size to avoid OOM
+            num_generations=2,  # VLM training is memory intensive, reduce num_generations to avoid OOM
+            # note: num_generations=2 is only suitable for CI testing; production training should use more generations
+            max_completion_length=8,  # reduce the completion length to reduce memory usage
+            report_to="none",
+        )
+        quantization_config = BitsAndBytesConfig(
+            load_in_4bit=True,
+            bnb_4bit_use_double_quant=True,
+            bnb_4bit_quant_type="nf4",
+            bnb_4bit_compute_dtype=torch.bfloat16,
+        )
+        trainer = GRPOTrainer(
+            model=model_id,  # pass the model as an identifier, so that the trainer applies the quantization config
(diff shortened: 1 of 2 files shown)


tests/test_grpo_trainer.py
qgallouedec, line 1: we should have this for rloo as well no?

albertvillanova: Yes, added.

I had scoped it to GRPO because I wrote this test in the context of #6912: it is the replacement for the quantization coverage that the removed slow test provided, so the whole series stayed inside `test_grpo_trainer.py`. Widening it to RLOO is the right call.

RLOO is the only other trainer with a `test_train_vlm_peft` to sit next to, and its VLM tests are verbatim GRPO copies, so the new test ports unchanged. Both copies are now byte-identical modulo the trainer name.

I also added an assertion that the model is really loaded in 4-bit, so the test checks that the `quantization_config` argument did its job, and not only that training runs.

albertvillanova:
Merging this PR as I already addressed reviewer's request.

PR #6910, opened 2026-08-25
Add a QLoRA test to the trainers that support quantization_config

albertvillanova:
This PR adds a QLoRA test to the trainers that support `quantization_config` and did not have one.

### Motivation

`SFTTrainer`, `DPOTrainer` and `KTOTrainer` have a `test_peft_with_quantization` test, but `GRPOTrainer`, `RLOOTrainer`, `RewardTrainer` and `DistillationTrainer` have none, although all seven expose `quantization_config` and share the same handling of quantized models. As a result, these code paths are untested in four trainers:

- the detection of a quantized model, which decides whether `autocast_adapter_dtype=False` is passed to `get_peft_model`
- the cast of the adapter weights to bf16 for quantized models, following the QLoRA paper recommendation

The `quantization_config` trainer argument itself is untested everywhere: all existing quantization tests pass the config to `from_pretrained` instead, although the argument is the documented way to do QLoRA training ("Combine with `peft_config` for QLoRA training").

This is the language model counterpart of #6909, which covers the same paths for VLMs.

### Solution

Add a QLoRA test to GRPO, RLOO, reward and distillation, next to their existing PEFT test, with the same tiny model, dataset and training config, so that quantization is the only difference between the two. The four copies share the same body, the trainer call and its config aside.

The test does not check that the base model params are unchanged, unlike the equivalent tests in SFT, DPO and KTO: bitsandbytes casts the biases of a `Linear4bit` in-place during the forward pass, and those tests keep a hardcoded list of the affected biases to work around it. That list depends on which layers ran a forward pass, which is not predictable in GRPO and RLOO, since generation runs before training. Aligning SFT, DPO and KTO on this shape, and dropping the bias lists, is left to a follow-up.

### Changes

- Add a QLoRA test to the GRPO, RLOO, reward and distillation test suites, using the `quantization_config` trainer argument together with `peft_config`

---

> [!NOTE]
> **Low Risk**
> Test-only changes with no production trainer logic modified; CI may need bitsandbytes/GPU where these tests run.
> 
> **Overview**
> Adds **`test_train_peft_and_quantization`** to the fast test suites for **GRPO**, **RLOO**, **Reward**, and **Distillation** trainers, mirroring the existing QLoRA tests on SFT/DPO/KTO.
> 
> Each test loads a tiny model with **`BitsAndBytesConfig`** (4-bit NF4) plus **`LoraConfig`**, asserts **`is_loaded_in_4bit`**, runs a short **`train()`**, and checks that **LoRA weights update** and stay in **bfloat16** per QLoRA. Imports gain **`BitsAndBytesConfig`** and **`@require_bitsandbytes`**.
> 
> Unlike the non-quantized PEFT tests, these **do not assert frozen base weights**, because bitsandbytes can mutate **`Linear4bit`** biases during forward (especially when generation runs before training).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d6e8448bead648dc31ac5c045055909bfd40bb59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_distillation_trainer.py
+++ b/tests/test_distillation_trainer.py
@@ -21,7 +21,7 @@ import transformers
 from accelerate.utils.memory import release_memory
 from datasets import DatasetDict, IterableDatasetDict, load_dataset
 from packaging.version import Version
-from transformers import AutoModelForCausalLM
+from transformers import AutoModelForCausalLM, BitsAndBytesConfig
 from transformers.utils import is_peft_available
 
 from trl import DistillationConfig, DistillationTrainer
@@ -30,6 +30,7 @@ from trl.trainer.distillation_trainer import _chunked_divergence_loss
 
 from .testing_utils import (
     TrlTestCase,
+    require_bitsandbytes,
     require_liger_kernel,
     require_peft,
     require_response_parsing,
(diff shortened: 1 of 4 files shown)


qgallouedec (commented):
same question, we should have the same test in all the trainers that support `quantization_config`?

albertvillanova:
You are right, and same reason as in #6909: I wrote this in the context of #6912, where it was the language model counterpart of the removed test, so it stayed GRPO-only.

All seven trainers that expose `quantization_config` are now covered: sft, dpo and kto already had `test_peft_with_quantization`, and this PR adds one to grpo, rloo, reward and distillation.

There is one thing to decide, because there are now two shapes for the same test:

- the existing sft/dpo/kto one: model quantized through `from_pretrained`, plus a hardcoded `bnb_bias_params_that_change` list gated on bitsandbytes 0.50.0
- the one added here: `quantization_config` passed to the trainer, which is the documented QLoRA entry point and was untested in every trainer until now, checking that the model is loaded in 4-bit and that the adapter weights are trained and cast to bf16

I did not copy the first shape, because its bias list depends on which layers ran a forward pass, which is not predictable in GRPO and RLOO where generation runs before training. What that list protects, the base model staying frozen, is already asserted by each trainer's non-quantized PEFT tests.

If you want a single shape everywhere, I would converge on the second one and drop the bias lists from sft/dpo/kto in a follow-up. #6911 already adds the trainer argument test for SFT.

albertvillanova:
Merging this PR as I already addressed reviewer's request.

PR #6913, opened 2026-08-25
Fix:  vLLM weight sync fails instead of hanging

### LINKED ISSUE #6796: AsyncGRPO: NCCL weight transfer can hang forever with no error
When the NCCL weight sync between `AsyncGRPOTrainer` and the vLLM server can't proceed, the trainer hangs instead of raising. No trainer-side log, no timeout, GPUs stay busy until the job time limit. I hit two variants of this at 32B. Both required ssh-ing into the node and reading `nvidia-smi` and the server log to diagnose.

## Reproduction

`train.py`:

```python
from datasets import Dataset

    return [-abs(20 - len(c[-1]["content"])) for c in completions]

dataset = Dataset.from_list(
    [{"prompt": [{"role": "user", "content": f"Write a short story about item {i}."}]} for i in range(4096)]
)
config = AsyncGRPOConfig(
    output_dir="out", per_device_train_batch_size=1, gradient_accumulation_steps=128,
    max_completion_length=1024, max_steps=12, save_strategy="no",
)
trainer = AsyncGRPOTrainer(model="Qwen/Qwen3-32B", args=config, train_dataset=dataset, reward_funcs=reward_len)
trainer.train()
```

`fsdp2_offload.yaml`:

```yaml
compute_environment: LOCAL_MACHINE
distributed_type: FSDP
mixed_precision: bf16
num_machines: 1
num_processes: 4
fsdp_config:
  fsdp_version: 2
  fsdp_offload_params: true   # variant A trigger
  fsdp_reshard_after_forward: true
  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
  fsdp_state_dict_type: SHARDED_STATE_DICT
```

On a single 8xH100 node:

```bash
# terminal 1: server on 4 GPUs
CUDA_VISIBLE_DEVICES=0,1,2,3 VLLM_SERVER_DEV_MODE=1 vllm serve Qwen/Qwen3-32B \
    --tensor-parallel-size 4 --max-model-len 2048 \
    --logprobs-mode processed_logprobs --weight-transfer-config '{"backend":"nccl"}'

# terminal 2: trainer on the other 4
CUDA_VISIBLE_DEVICES=4,5,6,7 accelerate launch --config_file fsdp2_offload.yaml train.py
```

## Outcome

Variant A, with `fsdp_offload_params: true`: the initial weight sync never completes. Trainer rank 0 pins one GPU at 100% inside `vllm/distributed/weight_transfer/packed_tensor.py::pack_tensors` for 50+ minutes, the other trainer ranks sit at 0%, the server is idle, and neither side prints anything.

Variant B, with `fsdp_offload_params: false` and the server at its default `--gpu-memory-utilization 0.9`: the transfer progresses but the server worker OOMs while receiving the chunks. This is only visible in the server log:

```
Exception: Call to collective_rpc method failed: Worker failed with error 'CUDA out of memory. Tried to allocate 372.00 MiB. ...'
```

The trainer then waits on the sync forever. Starting the server with `--gpu-memory-utilization 0.75` avoids it at this scale.

## Expected behavior

A failed transfer should raise on the trainer side, via a timeout or a propagated server error. If CPU-offloaded FSDP shards are unsupported in the transfer path, detect the config and fail fast with a clear message. A progress line every N tensors during the initial sync (~64 GB at 32B) and a documented headroom rule for the server `--gpu-memory-utilization` would also help.

## System info

trl 1.10.0.dev (main), vllm 0.27.1, torch 2.13.0+cu130, transformers 5.x, accelerate FSDP2, 8xH100 80GB. Variant A probably also reproduces at smaller scale with any `fsdp_offload_params: true` config, not verified yet.

AmineDiro:
# What does this PR do?
Closes #6796

The async GRPO weight-sync path discarded every HTTP response. Meanwhile the trainer sat in a raw ncclBroadcast, which has no timeout and cannot be interrupted, so the run hung until the job was killed, with the real error only ever visible in the server's own logs.

Check the status of every weight-sync request and raise with the server's error text. 

Swap which side of the transfer runs in the background: the NCCL collective now runs on a daemon thread while the HTTP request — which does time out, and raises when the server fails — stays on the main thread. That is what turns a failed transfer into an error instead of a hang; the abandoned thread dies with the process.

A failed transfer also leaves the server mid-update, and start_weight_update refuses while one is active, which would fail the next run's cold sync too. Clear it best-effort on the way out, so restarting the trainer is enough to recover: the cold sync re-sends every parameter.

weight_sync_timeout is now configurable and covers the whole sync path, replacing init_weight_transfer_timeout, which nothing could set.

Mirrored in async_distillation, a verbatim port of both files.

---

> [!NOTE]
> **Medium Risk**
> Changes the critical trainer-to-vLLM weight sync path and threading model; failures should surface sooner but behavior on partial failure and recovery depends on the new cleanup and timeout split.
> 
> **Overview**
> Async GRPO and async distillation no longer hang indefinitely when vLLM weight sync fails. **vLLM HTTP clients** now validate responses on pause/resume and all weight-update endpoints and raise with the server error body instead of ignoring failures.
> 
> **Weight transfer** flips the concurrency model: the blocking **NCCL** broadcast/init runs on a **daemon thread** while the **HTTP** call that drives the server stays on the main thread, so `requests` timeouts and non-200 responses surface as errors instead of leaving the trainer stuck in NCCL. Failed transfers trigger a best-effort `finish_weight_update` so a stuck server update does not block the next cold sync.
> 
> Training configs gain **`weight_sync_timeout`** (default 1800s), wired through the trainers into `WeightTransferClient` and used for handshake/transfer/finalize; pause, resume, and reload setup use a separate 300s control timeout. The old `init_weight_transfer_timeout` constructor knob is removed in favor of this config field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 377a14202e551156d118f5221406543228f0de87. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verma8076:
This is a solid fix for the failure mode where the server responds (with an error, or a timeout on an unresponsive HTTP layer) — that path now surfaces cleanly instead of hanging, and the `finish_weight_update` best-effort cleanup on failure is exactly right for making a restart actually recoverable.

One thing I want to flag rather than assert, since I can't see the vLLM-side handler code from here: in both `init_weight_transfer` and `send_weights`, the sequence is still `HTTP call (has a timeout) -> thread.join() (no timeout)`. If the corresponding vLLM HTTP endpoint returns/acks before its own NCCL op is actually done — e.g. `/init_weight_transfer_engine` just kicks off the server's rendezvous and returns, rather than blocking until the NCCL handshake completes — then a stall purely at the NCCL bootstrap layer (its own TCP store port blocked by a firewall that still lets the HTTP port through, for instance) would leave the daemon thread parked in `NCCLWeightTransferEngine.trainer_init()` forever, and `thread.join()` would reproduce the original hang for that specific case, just one level down.

If the server's handlers are synchronous and block on their own NCCL call before responding (which I'd guess is true for `/update_weights` at least, since the docstring says it "returns once the server has received every chunk" — matching your fix for the pack_tensors-stall variant of #6796/#6797, since a stalled send would leave the server's recv incomplete too, so the HTTP read times out correctly), this concern doesn't apply and the fix is complete. But if `/init_weight_transfer_engine` is closer to fire-and-forget on the server side, it might be worth confirming, or adding a bounded timeout to `thread.join()` too so every path has a way out. Might already be a non-issue depending on how that endpoint is implemented — figured it's worth asking given the whole point of this PR is closing exactly this class of hang.

behroozazarkhalili:
Reviewed this against `origin/main`. The diagnosis holds and the scope is right. Checks below, with what I could and could not verify.

**The dead parameter is genuinely dead.** `init_weight_transfer_timeout` appears only inside `weight_transfer.py` itself: declared at `:38`, assigned at `:45`, passed to the thread at `:73`. Both `WeightTransferClient(...)` call sites, `async_grpo_trainer.py:958` and `async_distillation_trainer.py:1058`, contain zero occurrences of it, and no config field ever surfaced it. Replacing it with a configurable `weight_sync_timeout` removes a knob that could not be reached rather than changing a setting anyone was using.

**The mirroring is complete.** `weight_transfer.py` and the experimental `vllm_client.py` exist in exactly two copies each, async_grpo and async_distillation, and this PR updates both. Nothing is left behind.

**One thing worth stating explicitly, because it reads like a gap and is not.** There is a third `vllm_client.py`, the stable `trl/generation/vllm_client.py`, which this PR does not touch. That is correct: it already routes every call through wrapper methods at `vllm_client.py:246-254` that raise on any non-200. So this change is not adding a new convention, it is bringing the two experimental copies up to what the stable client has been doing all along. Might be worth a line in the description, since a reviewer checking the consistency rule will land on that third file and wonder.

**No conflict with the other async config change in flight.** #6944 also edits `async_grpo_config.py`, adding a `__post_init__` warning near line 384, while this PR adds `weight_sync_timeout` near 311 and a docstring entry near 94. `git merge-tree --write-tree` on the two heads returns clean, so they merge in either order without intervention.

**What I did not verify.** The daemon-thread change is the part I cannot speak to. Whether an abandoned NCCL thread can leave the communicator in a state that blocks a later collective needs a live two-GPU run, and my async job is still queued. The reasoning in the description is sound, and moving the interruptible side to the main thread is the right shape for turning a hang into an error. I am flagging it as untested from my side rather than endorsing it.

The `finish_weight_update` best-effort cleanup is a good call. A server stuck mid-update failing the next run's cold sync is exactly the kind of second-order failure that turns one bad run into a confusing morning.

trl/experimental/async_distillation/async_distillation_config.py
```diff
@@ -83,6 +83,9 @@ class AsyncDistillationConfig(_BaseConfig):
             `teacher_id` of `"math"` or `"code"` per row.
         request_timeout (`int`, *optional*, defaults to `600`):
             Timeout in seconds for individual HTTP requests to any vLLM server.
+        weight_sync_timeout (`int`, *optional*, defaults to `1800`):
+            Timeout in seconds for a weight transfer to the student's vLLM server. A transfer that does not complete
+            within this time raises instead of hanging the run.
```
qgallouedec, line 88: It's saying, that `weight_sync_timeout` covers the whole sync path, but `start_weight_update()` and both `finish_weight_update()` calls in `send_weights` are unchanged and pass no timeout, so they keep the hardcoded 1800. set `weight_sync_timeout=120` and two of the five requests ignore it.

qgallouedec (approved):
thks, the shape is right and the `finish_weight_update` cleanup is a good call!

@verma8076 on the unbounded thread.join(): I don't think it can bite. `/init_weight_transfer_engine` awaits `collective_rpc`, so the 200 only lands after every worker returned from `ncclCommInitRank`, which can't happen unless the trainer joined the bootstrap as rank 0. If the trainer is stuck there the workers are too, the endpoint never answers, and the HTTP timeout fires before we reach the join.

trl/experimental/async_distillation/async_distillation_config.py
AmineDiro: Agreed I went with sane defaults for init and finish but I should specify this better in the docs 👍

behroozazarkhalili: Checked both points against the PR head (`bb72f1c8`), and the count is exact.

All six `VLLMClient` methods on the sync path are declared `timeout: int = 1800` and pass that straight into `requests.post(..., timeout=timeout)` (`vllm_client.py:90-130`), so an omitted argument really does fall back to 1800 rather than to something the server controls.

One weight sync issues five requests, in this order:

| # | call | call site | honors `weight_sync_timeout` |
|---|---|---|---|
| 1 | `pause` | `weight_transfer.py:159` | yes |
| 2 | `start_weight_update` | `weight_transfer.py:124` | no |
| 3 | `update_weights` | `weight_transfer.py:140` | yes |
| 4 | `finish_weight_update` | `weight_transfer.py:154` | no |
| 5 | `resume` | `weight_transfer.py:164` | yes |

So "two of the five" is right on the nose. `async_distillation/weight_transfer.py` has the same shape at `:116`, `:132`, `:146`, `:151`, `:156`.

One consequence worth weighing before settling on a documentation change. On the failure path, `update_weights` raises after `weight_sync_timeout`, and the cleanup call inside the `except` block is `finish_weight_update()` with no timeout, wrapped in `contextlib.suppress` (`:146-147`). Against an unresponsive server that adds up to another 1800s before the `RuntimeError` reaches the user. With `weight_sync_timeout=120` the run can sit for 120 + 1800 = 1920s, roughly 16 times the configured value, and because the exception is suppressed nothing is logged while it waits.

Documenting that init and finish keep their own defaults would be accurate, but it does not remove that overshoot, and the failure path is where a short timeout is set in the first place. Passing `timeout=self.weight_sync_timeout` at the three unparameterized call sites in each file looks like a smaller change than the paragraph that would be needed to explain why they differ.

If the intent is that init and finish should stay long deliberately, that is a reasonable call, and in that case two separate settings would read more clearly than one whose name covers three of the five requests.

AmineDiro: Fixed in [d30cac9](https://github.com/huggingface/trl/pull/6913/commits/d30cac9b7155373ad71e3ee0e021bf13645b4fac)

PR #6923, opened 2026-08-26
Run the training invariant tests in CI

qgallouedec:
Moves the daily invariant run off my local SLURM cron into GitHub Actions, same shape as the other scheduled test workflows. Daily at 2am UTC (midnight and 1am are taken by tests_latest and tests_python_versions), plus workflow_dispatch. Also adds a `make invariant_tests` target so the workflow looks like its siblings.

Runner is aws-g5-12xlarge-cache, not g6e-4xlarge, because sft_ddp2 and dpo_ddp2 need 2 GPUs.

Caveat: the references in tests/invariant/references/ are recorded on H100 80GB and there is no H100 runner group, so this runs on A10G. If the tolerances don't hold, we re-record on the CI GPU and lose the H100 pin.

---

> [!NOTE]
> **Low Risk**
> CI and Makefile wiring only; no application or training code changes, though GPU mismatch vs reference snapshots could cause initial CI failures.
> 
> **Overview**
> Adds a **scheduled GitHub Actions workflow** that runs the training invariant test suite daily at 2am UTC (with `workflow_dispatch`), replacing a local SLURM cron setup.
> 
> The job uses the **2-GPU** `aws-g5-12xlarge-cache` runner and the same PyTorch CUDA container / `uv` install pattern as sibling workflows, then runs **`make invariant_tests`**. Results are posted to Slack on every run.
> 
> The **Makefile** gains an `invariant_tests` target that runs `pytest -m "invariant"` against `tests/invariant`.
> 
> Note: push is temporarily limited to `invariant-ci` until merge; references were recorded on H100 while CI uses A10G, so tolerances may need re-recording if the job flakes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 74af71748f6b73b67b10a240fc6eb4f3170e7e9f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/tests_invariant.yml
+++ b/.github/workflows/tests_invariant.yml
@@ -1,0 +1,65 @@
+name: Training invariant tests
+
+on:
+  schedule:
+    - cron: '0 2 * * *'  # Runs daily at 2am UTC
+
+  workflow_dispatch:
+
+env:
+  TQDM_DISABLE: 1
+  CI_SLACK_CHANNEL: ${{ secrets.CI_PUSH_MAIN_CHANNEL }}
+  HF_TOKEN: ${{ secrets.HF_TOKEN }}
+  PYTORCH_CUDA_ALLOC_CONF: "expandable_segments:True"
+  PYTORCH_ALLOC_CONF: "expandable_segments:True"
+
+jobs:
+  tests:
+    name: Training invariant tests
+    runs-on:
+      group: aws-g5-12xlarge-cache  # the ddp2 configs need 2 GPUs
+    container:
+      image: pytorch/pytorch:2.8.0-cuda12.8-cudnn9-devel
+      options: --gpus all --shm-size "16gb"
+    defaults:
+      run:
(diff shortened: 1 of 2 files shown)


qgallouedec:
Closing this. GH Actions can't run this suite.

Full run on aws-g5-12xlarge-cache (4x A10G 24GB): 10/10 failed, none on tolerances. dpo_default and dpo_no_grad_ckpt OOM (pdb=8 fp32 + ref model doesn't fit in 24GB), sft_fa2 has no sm86 build variant in kernels-community/flash-attn2. Every GPU group in the org is 24GB per GPU, so fitting it means shrinking DPO batch sizes and dropping the FA2 class, which changes what the suite asserts.

Going with HF Jobs on h200x2 instead: Hopper keeps the H100 references valid, 141GB clears the OOM, sm90 has the FA2 builds.

PR #6933, opened 2026-08-26
Remove the unused TRL dev Docker image

albertvillanova:
This PR removes the TRL dev Docker image, which has had no consumer for almost a year.

### Motivation

The Docker images were introduced in #1215 as pre-baked GPU environments for the slow tests workflow, which was created one day later in #1223 and consumed them as a `container:` matrix. The dev image is the descendant of `huggingface/trl-source-gpu`, named "Latest TRL + HF ecosystem from source", whose only purpose was to give the slow tests an environment built from the main branch.

#3931 replaced the two old images with `huggingface/trl` and `huggingface/trl:dev`. Three days later, #4085 switched the slow tests to `pytorch/pytorch:2.8.0-cuda12.8-cudnn9-devel` with `uv pip install ".[dev]"`, dropping both images.

Since then, `huggingface/trl:dev` has had no consumer:

- No workflow references it.
- It is not mentioned anywhere in the documentation, unlike `huggingface/trl`, which the Jobs guide documents as `--image huggingface/trl`.

It is still rebuilt and pushed on every single push to main, which is about 150 builds per month of a 9.6 GB image that nothing pulls.

Removing it is also a prerequisite for building the released image on release rather than on every push to main: the dev image tracks the main branch, so it is the only reason the workflow needs to run per commit.

### Changes

- Remove the dev image Dockerfile
- Remove the dev image build job from the Docker build workflow

### Before merging

Does anyone rely on `huggingface/trl:dev`? It has never been documented, so I assume not, but I would rather confirm than break an undocumented workflow. Keeping this as a draft until then.

CC: @huggingface/trl 

---

> [!NOTE]
> **Low Risk**
> CI-only cleanup with no application or runtime code changes; only removes an undocumented image that nothing in the repo references.
> 
> **Overview**
> Stops building and publishing **`huggingface/trl:dev`** by deleting the **`trl-dev`** workflow job and removing **`docker/trl-dev/Dockerfile`**.
> 
> The main **`docker-build`** workflow now only builds the release **`huggingface/trl`** image (version-tagged from PyPI plus **`latest`**), with Slack notification unchanged for that job.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26b5a8935fda2cdf0d43b0bffb85a0a61366328d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -63,37 +63,3 @@ jobs:
           title: 🤗 Results of the TRL Dev Docker Image build
           status: ${{ job.status }}
           slack_token: ${{ secrets.SLACK_CIFEEDBACK_BOT_TOKEN }}
-
-  trl-dev:
-    name: "Build and push TRL Dev Docker image"
-    runs-on:
-      group: aws-general-8-plus
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e  # v4.3.0
-
-      - name: Login to DockerHub
-        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f  # v4.6.0
-        with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_PASSWORD }}
-
-      - name: Build and Push
-        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a  # v7.3.0
-        with:
-          context: docker/trl-dev
-          push: true
-          tags: |
-            huggingface/trl:dev
-
-      - name: Post to Slack
-        if: always()
-        uses: ./.github/actions/post-slack
-        with:
-          slack_channel: ${{ env.CI_SLACK_CHANNEL }}
-          title: 🤗 Results of the TRL Dev Docker Image build
-          status: ${{ job.status }}
-          slack_token: ${{ secrets.SLACK_CIFEEDBACK_BOT_TOKEN }}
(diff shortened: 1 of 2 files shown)


qgallouedec:
We can't really know. it's a public image, someone could be pulling it without ever referencing it in the repo, and docker hub pull counts are per-repo not per-tag (AFAIK) so they won't answer it either.

fwiw I've used it a few times: working on trl-jobs, and just running hf jobs when I wanted main rather than the last release. So at least one consumer 😅

It's not really costing us anything (same tag gets overwritten, layers are shared), so i'd rather keep it. if the goal is to stop the workflow running per commit (which again I think is cheap enough), build it nightly on a cron instead of on every push to main.

albertvillanova:
You're right, that settles it. And the pull count point is fair: Docker Hub reports pulls per repo, not per tag, so there was no way to answer this with the data I had. I should have asked before opening the PR rather than asserting there was no consumer. Closing.

For what it's worth, nightly cron is where this started: #1215 built the images on `schedule: cron "0 1 * * *"`, and #3931 moved them to push on main without saying why. So your suggestion is a restoration rather than a change.

I'm leaving `:dev` exactly as it is, because the thing I actually care about turns out to be independent of it. The released image has a real problem: `huggingface/trl:1.9.2` was rebuilt on all 65 pushes to main between its release and 1.10.0, ending up at digest `d57701ac...` pushed 12 seconds before 1.10.0 landed on PyPI. On top of that the Dockerfile does an unpinned `uv pip install trl[...]`, so a version tag does not identify what is inside it. That matters more now that trl-jobs hardcodes `image="huggingface/trl"`.

So instead of removing anything, I'll split the workflow: the released image builds on `release: published`, `:dev` keeps building on push to main exactly as today. Nightly cron for `:dev` then becomes a one line follow-up whenever you want it.

One question while you're here, since you use `:dev` to get main rather than the last release: does `hf jobs uv run --image` actually use the packages baked into the image? The Jobs guide advertises `--image huggingface/trl` two paragraphs after stating that every TRL example carries a `# /// script` header, and I could not tell whether uv builds an isolated env from that header and ignores the preinstalled trl. Your usage suggests it does not, but I would rather hear it from you than guess.

PR #5731, opened 2026-05-08
[GKD] Fix Liger fused JSD path computing wrong loss

roycho96:
## What

When `use_liger_kernel=True`, `GKDTrainer.compute_loss` was silently training against a different objective than the default path:

1. `LigerFusedLinearJSDLoss` defaults to `weight_hard_loss=0.5, weight_soft_loss=0.5`, so the loss was `0.5 * CE(student, true_labels) + 0.5 * JSD(student, teacher)` instead of the pure JSD that `generalized_jsd_loss` computes.
2. The kernel was constructed with `temperature=args.temperature`. `args.temperature` is the generation-sampling temperature (default `0.9`); the non-Liger path calls `generalized_jsd_loss(...)` without a `temperature` argument and therefore uses `1.0`. The two paths were scaling logits differently before computing JSD.

Both came in with #3946 ([GKD] add liger loss). `distillation_trainer.py` and `gold_trainer.py` already pass `weight_hard_loss=0.0, weight_soft_loss=1.0`; only GKD was missed.

## Reproducer

Same student + same teacher, batch from `trl-internal-testing/zen`, tiny Qwen2.5 model:

| path | loss |
|---|---|
| non-Liger `generalized_jsd_loss` | 0.297 |
| Liger (defaults: `0.5 * CE + 0.5 * JSD`, `temperature=args.temperature=0.9`) | 0.342 |
| Liger after fix (`pure JSD`, `temperature=1.0`) | 0.297 |

The 15% gap is the silent objective swap. With student == teacher (so JSD ≈ 0) the gap was `~6.0`, dominated by the unwanted `0.5 * CE` term.

## Fix

```python
self.liger_jsd_loss = LigerFusedLinearJSDLoss(
    beta=args.beta,
    ignore_index=-100,
    temperature=1.0,        # was args.temperature
    compiled=False,
    weight_hard_loss=0.0,   # added
    weight_soft_loss=1.0,   # added
)
```

## Tests

- New `test_liger_loss_matches_non_liger_loss` builds two trainers with `use_liger_kernel=False/True` on the same student/teacher and asserts `compute_loss` agrees within fp tolerance.
- Existing `test_compute_loss_return_outputs_with_liger` still passes.
- `test_gkd_trainer_with_liger` OOMs on a 16 GB card both before and after this PR (independent issue).

---

> [!NOTE]
> **Medium Risk**
[...]

PR #3946, opened 2025-08-24
[GKD] add liger loss

kashif:
# What does this PR do?

Added Liger GKD loss and test

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

Issue #6849, opened 2026-08-21
CI emits torch inductor UserWarning: TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled


albertvillanova:
### What

CI emits a new `UserWarning` from `torch._inductor` in the Liger tests:

```
tests/test_dpo_trainer.py::TestDPOTrainerVLM::test_train_vlm_liger
tests/test_kto_trainer.py::TestKTOTrainer::test_train_with_liger
tests/test_dpo_trainer.py::TestDPOTrainer::test_train_with_liger
tests/test_kto_trainer.py::TestKTOTrainerVLM::test_train_vlm_liger
  /__w/trl/trl/.venv/lib/python3.12/site-packages/torch/_inductor/compile_fx.py:321: UserWarning: TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled. Consider setting `torch.set_float32_matmul_precision('high')` for better performance.
    warnings.warn(
```

CI run: https://github.com/huggingface/trl/actions/runs/32414561958/job/96572632322

Affected tests:

- `tests/test_dpo_trainer.py::TestDPOTrainer::test_train_with_liger`
- `tests/test_dpo_trainer.py::TestDPOTrainerVLM::test_train_vlm_liger`
- `tests/test_kto_trainer.py::TestKTOTrainer::test_train_with_liger`
- `tests/test_kto_trainer.py::TestKTOTrainerVLM::test_train_vlm_liger`

The tests themselves pass; for now this is warning noise only.

### When it started

First seen in the run for #6741 (`1701d3e`), which moved the GPU CI from T4 to L40S: https://github.com/huggingface/trl/actions/runs/32366392208/job/96416712686 (4 warnings, all of them this one). No warning was reported in the run for `06c89977`, on the previous runner: https://github.com/huggingface/trl/actions/runs/32348545011/job/96362435182.

### Next steps

Investigate the root cause, then decide whether anything is actionable on the TRL side and/or should be reported upstream.

albertvillanova:
## Root cause pinned down

### Why the runner change surfaced it

`torch._inductor` emits an advisory on every compile when TF32 tensor cores exist but the float32 matmul precision is left at its default — `torch/_inductor/compile_fx.py:315` ([v2.13.0](https://github.com/pytorch/pytorch/blob/v2.13.0/torch/_inductor/compile_fx.py#L315), the version CI installs; the `warnings.warn` is at line 321, matching the reported location):

```python
[...]

PR #6942, opened 2026-08-26
enable compiled Liger JSD losses

mdeff:
# What does this PR do?

This PR enables `torch.compile` for `LigerFusedLinearJSDLoss` by removing TRL's explicit `compiled=False` override.

It updates all six trainers using Liger JSD:
- `DistillationTrainer`
- `GKDTrainer`
- `GOLDTrainer`
- `IWOPDTrainer`
- `SDFTTrainer`
- `SDPOTrainer`

The corresponding DistillationTrainer parity test now also exercises Liger's compiled default.

## Motivation

`LigerFusedLinearJSDLoss` defaults to `compiled=True`, and Liger's correctness tests exercise that default. Other Liger chunked losses used by TRL, including DPO and KTO, already use compilation.

TRL has explicitly disabled compilation for JSD since its initial integration in #3946. I found no documented correctness or compatibility reason for this override in the original PR or subsequent JSD fixes and refactors, including #5731 and #6537.

## Benchmark

The benchmark used the same single H100 environment for all configurations:
- Llama 3.2 1B student and teacher
- bfloat16
- sequence length 1024
- batch size 32
- 1,000 optimizer steps
- fused AdamW
- no gradient checkpointing
- no evaluation or checkpoint saving

Elapsed time was measured from the start of training through completion, so the initial compilation cost is included.

| Chunk size | Uncompiled | Compiled | Runtime reduction | Throughput gain |
|---:|---:|---:|---:|---:|
| 512 | 1.224 s/step, 26.8k tok/s | 0.874 s/step, 37.5k tok/s | 28.6% | 40.1% |
| 1024 | 1.202 s/step, 27.3k tok/s | 0.832 s/step, 39.4k tok/s | 30.8% | 44.6% |
| 2048 | 1.165 s/step, 28.1k tok/s | 0.829 s/step, 39.5k tok/s | 28.8% | 40.5% |
| 4096 | 1.157 s/step, 28.3k tok/s | 0.811 s/step, 40.4k tok/s | 29.9% | 42.7% |

Across the four tested chunk sizes, compilation reduced elapsed training time by approximately 29.5% and increased throughput by approximately 42%.

## Tests

The affected Liger tests were run with TRL's CI configuration:
- PyTorch 2.8.0+cu128
- Liger Kernel 0.8.2
- fresh TorchInductor cache

Result:
```text
14 passed, 1 skipped, 257 deselected
```

The suite covers the compiled JSD paths in Distillation, GKD, GOLD, IW-OPD, SDFT, and SDPO, including loss-parity and end-to-end training tests.

## Before submitting

## AI writing disclosure

---

> [!NOTE]
> **Medium Risk**
> Training loss now goes through `torch.compile` on the fused JSD path, which can affect numerics, warmup/compile behavior, and PyTorch/Liger version compatibility despite existing parity tests.
> 
> **Overview**
> Stops forcing `compiled=False` on `LigerFusedLinearJSDLoss` so TRL’s distillation trainers use Liger’s default **torch.compile** path for fused JSD.
> 
> The change is applied everywhere TRL wires up that loss: **`DistillationTrainer`**, **`GKDTrainer`**, **`GOLDTrainer`**, **`IWOPDTrainer`**, **`SDFTTrainer`**, and **`SDPOTrainer`**. **`test_liger_loss_agrees_with_chunked`** in `test_distillation_trainer.py` is updated the same way, so parity checks run against the compiled kernel instead of an explicitly uncompiled instance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 720bc4634f8ebd3f13bee691e45af359312fbeb0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_distillation_trainer.py
+++ b/tests/test_distillation_trainer.py
@@ -1228,7 +1228,6 @@ class TestDistillationTrainer(TrlTestCase):
                 beta=trainer.beta,
                 ignore_index=-100,
                 temperature=trainer.temperature,
-                compiled=False,
                 weight_hard_loss=0.0,
                 weight_soft_loss=1.0,
             )
--- a/trl/experimental/gkd/gkd_trainer.py
+++ b/trl/experimental/gkd/gkd_trainer.py
@@ -153,7 +153,6 @@ class GKDTrainer(SFTTrainer):
             self.liger_loss = LigerFusedLinearJSDLoss(
                 beta=args.beta,
                 ignore_index=-100,
-                compiled=False,
                 weight_hard_loss=0.0,
                 weight_soft_loss=1.0,
             )
--- a/trl/experimental/gold/gold_trainer.py
+++ b/trl/experimental/gold/gold_trainer.py
@@ -927,7 +927,6 @@ class GOLDTrainer(SFTTrainer):
                 beta=args.beta,
                 ignore_index=-100,
                 temperature=args.temperature,
-                compiled=False,
                 weight_hard_loss=0.0,
                 weight_soft_loss=1.0,
             )
--- a/trl/experimental/iw_opd/iw_opd_trainer.py
+++ b/trl/experimental/iw_opd/iw_opd_trainer.py
@@ -493,7 +493,6 @@ class IWOPDTrainer(_BaseTrainer):
                 beta=args.beta,
                 ignore_index=-100,
                 temperature=args.temperature,
-                compiled=False,
                 weight_hard_loss=0.0,
                 weight_soft_loss=1.0,
             )
--- a/trl/experimental/sdft/sdft_trainer.py
+++ b/trl/experimental/sdft/sdft_trainer.py
@@ -397,7 +397,6 @@ class SDFTTrainer(_BaseTrainer):
                 beta=args.distillation_alpha,
                 ignore_index=-100,
                 temperature=args.temperature,
-                compiled=False,
                 weight_hard_loss=0.0,
                 weight_soft_loss=1.0,
             )
--- a/trl/experimental/sdpo/sdpo_trainer.py
+++ b/trl/experimental/sdpo/sdpo_trainer.py
@@ -536,7 +536,6 @@ class SDPOTrainer(_BaseTrainer):
                 beta=args.distillation_alpha,
                 ignore_index=-100,
                 temperature=args.temperature,
-                compiled=False,
                 weight_hard_loss=0.0,
                 weight_soft_loss=1.0,
             )
--- a/trl/trainer/distillation_trainer.py
+++ b/trl/trainer/distillation_trainer.py
@@ -652,7 +652,6 @@ class DistillationTrainer(_BaseTrainer):
                 beta=args.beta,
                 ignore_index=-100,
                 temperature=args.temperature,
-                compiled=False,
                 weight_hard_loss=0.0,
                 weight_soft_loss=1.0,
             )


behroozazarkhalili:
I checked the four factual claims behind this change. All four hold, and the consistency argument is stronger than the PR body makes it.

**The diff does what it says.** Constructing the loss with exactly the kwargs the trainers now use, on liger-kernel 0.8.2:

```
compiled default = True
constructed OK, compiled = True
```

So removing the override really does hand the path to Liger's default rather than relying on anything implicit.

**Completeness.** There are six `LigerFusedLinearJSDLoss` construction sites under `trl/` plus the one in `tests/test_distillation_trainer.py`, and the PR touches all seven. After it, no Liger loss construction anywhere in `trl/` passes `compiled` at all: the DPO, KTO, and GRPO sites never did. That makes JSD uniform with every other fused loss in the repo, which is the argument I would lead with, since TRL treats consistency across duplicated call sites as a requirement rather than a preference.

**Origin of the override.** `39cc9a82` (`[GKD] add liger loss`, #3946) introduced `compiled=False` with no comment and no rationale in the diff, and nothing since added one. Your reading of the history matches mine.

**Gates.** ruff 0.13.3 `check` and `format --check` and the pinned doc-builder are all clean on all seven files.

Two things to flag, neither of them a blocker.

**`pyproject.toml:253` becomes inaccurate.** That filter carries the comment:

```
# Triggered by the DPO and KTO Liger tests, the only ones that compile, since the GPU CI moved from T4 to L40S
```

After this PR the JSD tests compile too, so "the only ones that compile" is no longer true. Nothing breaks: the entry is an `ignore` and the `filterwarnings` list has zero `error` entries, so CI cannot fail on that advisory either way. The cost is a comment that now misdescribes the situation, on a filter linked to tracking issue #6849. A one-line edit alongside this change would keep it honest.

**What the green parity test does and does not establish.** `test_liger_loss_agrees_with_chunked` carries its own caveat: it scales both LM heads by 50 because, at the float32 noise floor with a near-uniform distribution over the ~150k vocab, the chunked and fused paths' matmul reduction orders diverge by about 2x. So the assertion at `rtol=1e-3, atol=1e-4` is parity on a deliberately well-conditioned batch. That is a reasonable test design, and it is also narrower than "compilation leaves the numerics alone", which is where Bugbot's medium-risk note points. Worth saying out loud in the PR body so a later reader does not over-read the green check.

One limitation on my side, stated plainly: I could not reproduce your benchmark or run the Liger tests. This machine has no GPU (`torch.cuda.is_available()` is `False`, device count 0) and the cluster queue is long enough that a run is not practical right now. Everything above is source-level and CPU-level verification. The 29.5% runtime and 42% throughput figures and the compiled-path test results are yours, not independently confirmed by me.

On the substance I think this should go in. A 29% runtime reduction for deleting seven lines is a good trade, the override had no recorded reason, and the change moves JSD toward the rest of the repo instead of away from it.

kashif:
Heads up: #7301 removes the fused Liger JSD loss from these five trainers entirely (matching #7064 for `DistillationTrainer`), as part of #7063's plan to drop the Liger fused-loss paths in favor of the chunked log-prob path everywhere. Once that lands, this PR's `compiled=False` removals won't have anything left to apply to. Sorry for the wasted effort — thanks for digging into the compile flag.

behroozazarkhalili:
Thanks for the heads-up. Agreed, once #7301 lands there is nothing left here for the compile flag to apply to, and dropping the fused path is the cleaner outcome anyway. It also settles the `pyproject.toml` filterwarnings note I raised above: with the JSD path gone, the DPO and KTO Liger tests really are the only ones that compile, so that comment stays accurate as written.

Issue #6789, opened 2026-08-18
GRPO: vLLM importance-sampling ratio is biased when top_p/top_k/min_p truncate sampling


qgallouedec:
When `top_p < 1`, `top_k > 0` or `min_p` is set, GRPO's vLLM importance-sampling correction compares two log-probabilities computed over **different distributions**, so the ratio is biased. vLLM returns log-probs renormalized over the truncated (nucleus) distribution, while the trainer recomputes them over the full vocabulary.

Two other RL frameworks have now shipped a training-side fix for exactly this, and the native vLLM fix is merged but not yet in a release TRL supports.

### Where the mismatch is

Generation side asks vLLM for post-processor log-probs, in both colocate and server mode:

https://github.com/huggingface/trl/blob/main/trl/generation/vllm_generation.py#L365

Training side divides by temperature and takes a full-vocab log-softmax:

https://github.com/huggingface/trl/blob/main/trl/trainer/grpo_trainer.py#L1530-L1534

`processed_logprobs` means "after every logits processor", which includes top-p/top-k truncation followed by renormalization over the surviving tokens. Temperature is handled symmetrically (line 1532 mirrors it, and that symmetry was the fix for #4159); truncation is not. For a sampled token that survives the nucleus, vLLM's log-prob is therefore *higher* than the trainer's by `-log(sum of surviving probability mass)`, which is a per-token, systematically negative bias in `log π_θ - log π_behaviour`.

`importance_sampling_ratio` multiplies the per-token loss directly:

https://github.com/huggingface/trl/blob/main/trl/trainer/grpo_trainer.py#L3208-L3209

so the effect is a silent, token-dependent down-weighting of the policy gradient rather than a crash.

### Scope

- Defaults are safe: `top_p=1.0`, `top_k=0`, `min_p=None`, so nothing is truncated and the two distributions agree.
- `AsyncGRPOTrainer` is affected more severely. Server launch passes `--logprobs-mode processed_logprobs`, and those generator log-probs are not an optional correction there: they are the denominator of the policy ratio itself, with no flag guarding them.

  ```python
  log_ratio = log_probs - old_log_probs   # async_grpo_trainer.py#L945
  coef_1 = torch.exp(log_ratio)
  ```

  `AsyncGRPOConfig` exposes `top_p`/`top_k`/`min_p` like GRPO, and it logs no `sampling_logp_difference` equivalent, so there is nothing to notice the drift with.
[...]

PR #6944, opened 2026-08-27
Warn when reshaped sampling biases the AsyncGRPO importance ratio

behroozazarkhalili:
Async GRPO uses processed sampling log probabilities that can bias its importance ratio when sampling reshapes the distribution. I added configuration warnings for those settings and for positive temperatures below vLLM's minimum. The configuration tests pass, including default settings that stay silent, but I have not verified the ratio bias in a live async training run.
Related #6789, #6880.

---

> [!NOTE]
> **Low Risk**
> Config-time warnings and documentation only; no change to training math or rollout behavior beyond alerting users.
> 
> **Overview**
> **`AsyncGRPOConfig`** now emits **`UserWarning`s at construction** when generation settings would skew the PPO-style ratio `exp(log_probs - old_log_probs)`. Async GRPO uses vLLM’s per-token logprobs (with TRL’s prescribed `processed_logprobs`) as `old_log_probs` while the trainer scores the full vocabulary, and unlike sync GRPO there is no importance-sampling correction to disable.
> 
> One warning fires if **`top_p`**, **`top_k`**, **`min_p`**, or **`repetition_penalty`** leave their defaults (any combination is listed in the message). A separate warning covers **`0 < temperature < 0.01`**, where vLLM clamps to `0.01` but the trainer still uses the requested value; ordinary temperatures (e.g. `0.7`) stay silent.
> 
> Docstrings for those fields are expanded to explain the mismatch. **`TestAsyncGRPOConfig`** adds parametrized CPU tests so defaults stay quiet and the two warning paths are pinned without the Flash-Attention GPU gate on **`TestAsyncGRPOTrainer`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5213c0d402927e1bc8571425f99abc18edd9f416. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


verma8076:
Glad this got picked up — thanks for building the actual fix.

Independently re-verified the two claims the derivation leans on, against current main:

- \`--logprobs-mode processed_logprobs\` is unconditionally appended in \`vllm_serve.build_command\` (no override point before it in the constructed command), so every AsyncGRPO server really is affected, not just a subset of configs.
- \`grep -n "importance_sampling"\` across \`async_grpo_config.py\` and \`async_grpo_trainer.py\` returns nothing — confirms there's genuinely no correction toggle to point users at, so a warning is the right shape of fix here (matches the reasoning in #6880 for why this isn't the same fix).

The \`async_grpo_trainer.py:1113-1114\` citation for \`coef_1 = exp(log_probs - old_log_probs)\` is exact on current main too.

Honest framing in "What is not verified here" is appreciated — the derivation is sound and the guard is what's testable without a live two-GPU job.

PR #6946, opened 2026-08-27
Remove the unreachable prepare_multimodal_messages_vllm

### LINKED ISSUE #6928: prepare_multimodal_messages_vllm fails on content-less tool-call messages
## Problem

`prepare_multimodal_messages` accepts an assistant message that has `tool_calls` but no `content`. Its vLLM companion, `prepare_multimodal_messages_vllm`, unconditionally indexes `message["content"]`, so it raises `KeyError` for that valid message representation.

## Minimal reproduction

```python
from trl.data_utils import prepare_multimodal_messages, prepare_multimodal_messages_vllm

messages = prepare_multimodal_messages(
    [
        {"role": "user", "content": "What is the weather in New York?"},
        {
            "role": "assistant",
            "tool_calls": [
                {
                    "type": "tool",
                    "function": {
                        "name": "get_current_weather",
                        "arguments": {"location": "New York"},
                    },
                }
            ],
        },
    ]
)
prepare_multimodal_messages_vllm(messages)
```

On current `main` (`0261c7c8c78ada470856709b6fc7760cf2a13f31`), this raises `KeyError: 'content'`.

## Expected behavior

The vLLM conversion should preserve a content-less tool-call message, just as the base multimodal-message helper does, while continuing to convert image blocks when `content` is present.

I have added a focused regression test. AI assistance was used to investigate and draft this report; the human reporter reviewed it and reproduced the failure.

behroozazarkhalili:
Assistant tool-call messages without `content` caused a key error in the live vLLM message conversion path. I made that path preserve these messages and removed the unused `prepare_multimodal_messages_vllm` helper and its exports. The new tool-call test fails on main, the data utility tests pass, and import checks confirm the removed name is gone.
Resolves #6928.

---

> [!NOTE]
> **Low Risk**
> Small API removal plus a defensive fix on the active vLLM chat path; behavior change is limited to conversations that previously crashed on content-less assistant tool-call turns.
> 
> **Overview**
> Removes the unused public helper **`prepare_multimodal_messages_vllm`** (implementation, **`trl`** exports, and its test class) after vLLM generation switched to **`to_openai_messages`**.
> 
> Fixes **`to_openai_messages`** so assistant turns with only **`tool_calls`** (no **`content`**) are left unchanged instead of raising **`KeyError`**, which affected **`VLLMClient.chat`** and init probing (#6928). Adds **`TestToOpenAIMessages`** for that tool-call path and for PIL image → data URL conversion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f8f9d768bcb0d56d64a4e73cebb48f839abeb3d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


behroozazarkhalili:
`prepare_multimodal_messages_vllm` is listed in `trl/__init__.py`, so `from trl import prepare_multimodal_messages_vllm` resolves. I checked every tag from v1.5.0 through v1.12.0 and the export is present in all twelve, so a user on any released version can import it from the package root.

That does not contradict the dead-code reading. In TRL's own call graph it is dead: on `main` the only files that mention it are `trl/__init__.py`, `trl/data_utils.py`, and its tests. Internal unreachability and external usage are separate questions, and the export answers the second one differently. @YZJF's repro imports it directly, which is the case you asked about, and there has been no reply either way.

TRL has removed public symbols before, in two steps. XPO, PPO, ORPO, PRM, and Judges were each deprecated first with a `warnings.warn` naming v2.0.0 as the removal release, then deleted in a follow-up. This PR deletes in one step.

I am not proposing a change here. You asked for the removal and the diff does that. If v2.0.0 is where breaks are being collected, this fits and there is nothing to do. If the import should keep resolving through 1.x, the shim is small and I can add it to this PR.

Issue #1036, opened 2023-11-27
Is it possible to use AutoModelForCausalLMWithValueHead without merging adapters first?


Reza-esfandiarpoor:
Hi,

In the new version of `transformers` and `peft`, it is possible to directly load a model with `AutoModelForCausalLM` instead of `AutoPeftModelForCausalLM`.

[This DPO script](https://github.com/huggingface/trl/blob/main/examples/research_projects/stack_llama_2/scripts/dpo_llama2.py) loads the adapters directly and attaches new adapters without using `merge_and_unload`.

Is it possible to do the same for the PPO script as well? Basically, is it correct to do the following?

```python
sft_model = AutoModelForCausalLM.from_pretrained(peft_model_id) # peft_model_id is the path to sft adapters

model = AutoModelForCausalLMWithValueHead.from_pretrained(
    sft_model,
    load_in_8bit=True,
    device_map={"": current_device},
    peft_config=args.lora_config,
)
```

Reza-esfandiarpoor:
Thank you! That sounds reasonable!

PR #6947, opened 2026-08-27
AsyncGRPO: train loop-owning agents on the engine's own token ids, with a Harbor/OpenEnv example

adithya-s-k:
An installed coding agent can own its tool loop while AsyncGRPO trains on the exact tokens its inference engine sampled. This adds a Harbor/OpenEnv example and captured-trace ingestion.

The loop-owning path validates engine prompt/completion IDs, sampled log probabilities, masks and sampling policy. If a harness rewrites history, lossless capture starts a new row instead of overwriting sampled tokens. White-box reconciliation keeps its existing behavior.

```python
HarborSessionFactory(
    server,
    split="<Harbor dataset>",
    harness="mini-swe-agent",
    sandbox="e2b",
    llm_url=vllm_url,
    model=model,
    sampling={"temperature": 0.8, "top_p": 1.0, "top_k": -1},
)
```

`examples/async_grpo_harbor/` includes local commands and a Hugging Face Jobs launcher. The example keeps verifier scores separate from its correctness-gated efficiency reward. Weight updates drain inference before replacing weights; empty successful cache-reset responses and tool-only messages are handled. Training requires engine token IDs and processed logprobs.

Validation: current main merged; all CI passes, including distributed GPU smoke. Local checks include 245 CPU tests, three HTTP control tests and pre-commit. TiTO regressions cover rewritten histories, masks, token retention, invalid logprobs and policy mismatch. Pinned-runtime Harbor/OpenCode, native OpenCode and SETA GPU smokes each passed four optimizer steps and native checkpoint save/remote restore; receipts are in adithya-s-k/HuggingEnvs#7.

Merge after [OpenEnv #1036](https://github.com/huggingface/OpenEnv/pull/1036), which provides the capture API. Lossless forks can increase row/token budgets; rollout-normalized weighting and equal harness contributions are not implemented here.


sergiopaniego (commented):
quick review, first pass, we need to add the example to `docs/source/example_overview.md`

qgallouedec:
Nice!
Not blocking but just two suggestion the CLI
- t's 30 flags against opencode's 15, and seven of yours just re-expose `TrainingArguments`: `--optim`, `--no-bf16`, etc. An example shouldn't re-export the trainer's config surface, it's meant to be edited, so the user can change those inline.
- And the three `os.environ.get` constants should be plain module constants like opencode's, one channel for tunables.

Also the docstring says "a 15-harness sweep" then "three of the twelve harnesses measured". which is it, and is the sweep written up anywhere we can link?

Feel free to merge

qgallouedec (approved):

qgallouedec:
Holding off on merging this, one thing I'd like to settle first.

`openenv_harness.py` imports `TraceEntry` and `LoopOwningSession` from `openenv.core.harness`, and neither exists there today. The RFC 005 harness PRs look like where they'd come from and they're still open. Which OpenEnv PR lands these, and can we pin a version once it does?

Also, since the producer isn't written yet, merging this first sets that contract "by accident" (rather than by agreement). I'd wait until what defines `loss_mask` on the OpenEnv side? Also, can a single turn carry a mix of 0s and 1s over its completion?

No rush from my side

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -64,6 +65,12 @@ class TurnRecord:
     prompt_ids: list[int]
     output_ids: list[int]
     output_log_probs: list[float] = field(default_factory=list)
+    # Per-token trainability over `output_ids`. None means "train all of them", which is what a
+    # locally-driven turn always wants because the loop just sampled every one of those tokens.
+    # A CAPTURED turn is different: the producer may mask a turn out while keeping its tokens as
+    # context -- e.g. its logprobs were rejected on ingest -- and that is not inferable here.
+    # Training those positions anyway means training against a logprob of 0.0, i.e. p = 1.0.
+    output_mask: list[int] | None = None
```
AmineDiro, line 73: Is this really needed outside harness support ? 

https://github.com/huggingface/trl/blob/b7df6a8f0562a26d25ac4324848a07f1bc40617d/trl/experimental/async_grpo/openenv_harness.py#L128-L129

these functions are basically made to expose the same mechanism. If OpenEnv changed to produce output_mask we might need to completely change harnessWorker

trl/experimental/async_grpo/openenv_harness.py
```diff
@@ -245,6 +266,15 @@ def _run_session(self, prompt, group_id=0):
                 timed_out=timed_out,
             )
             reward = self._rollout_reward_fn(outcome) if self._rollout_reward_fn else env_reward
+            if self._rollout_reward_fn is not None and env_reward is not None:
+                # Keep the VERIFIER's score visible on its own. `rollout_reward_fn` replaces `env_reward` wholesale,
```
AmineDiro, line 270: I'd prefer we cleanup these LLM generated comments in general. The provide very low signal and make it hard to focus on code IMHO

examples/async_grpo_harbor/async_grpo_harbor.py
```diff
@@ -0,0 +1,327 @@
...
+        train_turn_fn=has_tool_call,
+        # No `agent_turn_fn`: the capture layer already dropped auxiliary calls and de-duplicated forked
+        # paths structurally, which a flat trace cannot do.
+        model_name=args.model,
+        dataset=dataset,
+        reward_funcs=[],  # the reward is the task's own verifier, via `rollout_reward_fn`
+        processing_class=tokenizer,
+        # Must match how the engine was served, or every prompt is re-rendered under a different template
+        # than the rollout was generated with — silent skew, not an error.
+        chat_template_kwargs={"enable_thinking": False},
```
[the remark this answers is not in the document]
adithya-s-k: Fixed. We now use the captured engine prompt IDs directly, so there is no local chat-template re-rendering.

examples/async_grpo_harbor/async_grpo_harbor.py
```diff
@@ -0,0 +1,327 @@
...
+        num_generations=args.num_generations,
+        max_inflight_tasks=args.max_inflight,
+        vllm_server_url=args.vllm_url,
+        max_tokens=args.max_completion_length,
+        temperature=args.temperature,
```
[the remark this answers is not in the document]
adithya-s-k: Fixed. The example passes the same temperature to HarborSessionFactory and the trainer. Captures with mismatched sampling settings are rejected.

examples/async_grpo_harbor/async_grpo_harbor.py
```diff
@@ -0,0 +1,327 @@
...
+    p.add_argument("--seed", type=int, default=0)
+    return p.parse_args()
+
+
+def main() -> None:
+    args = parse_args()
+    # Trackio keys a run by name inside a project, so two relaunches of the same config land on top of
+    # each other and the earlier metrics read as part of the later run's history — worst exactly when
+    # relaunching after a crash. Stamping the name keeps them apart.
+    stamp = os.environ.get("SLURM_JOB_ID", "local")
```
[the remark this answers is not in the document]
adithya-s-k: Fixed. Default run names now include a unique suffix, including local runs. Explicit run names still work.

examples/async_grpo_harbor/launcher.py
```diff
@@ -0,0 +1,422 @@
...
+    )
+    if n_gpu < 2:
+        raise RuntimeError(
+            f"needs 2 GPUs (engine + trainer on one host for NCCL) but sees {n_gpu}; try --flavor h200x2"
+        )
+
+    # 1. the Harbor dataset + the capture proxy, published for the sandboxed agent
+    server_log = logs / "openenv-server.log"
+    public_proxy = start_harbor_server(args, server_log)
+    threading.Thread(target=supervise_tunnel, args=(args, server_log, args.tunnel_check_s), daemon=True).start()
```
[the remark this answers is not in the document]
adithya-s-k: Fixed. Setting the interval to 0 no longer starts the supervisor. Added a regression test.

trl/experimental/async_grpo/openenv_harness.py
```diff
@@ -245,6 +266,15 @@ def _run_session(self, prompt, group_id=0):
...
+                # and the column it feeds is already named `rewards/harness_reward`, so with a shaping term in play
+                # `reward` and `rewards/harness_reward` are the SAME shaped number and raw correctness is logged
+                # nowhere. Every shaping term then looks exactly like progress: a reward that climbs because the
+                # policy got faster is indistinguishable from one that climbs because it solved more, which is the
+                # one question the run exists to answer.
+                self._rates["rollout/correctness_mean"][0] += float(env_reward)
+                self._rates["rollout/correctness_mean"][1] += 1.0
```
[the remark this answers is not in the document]
adithya-s-k: Fixed. Session threads return correctness with the other metrics; the event loop updates the counters. Added a concurrency test.

trl/experimental/async_grpo/async_rollout_worker.py
adithya-s-k: The hook selects whole turns, but cannot express a partial mask like [1, 0, 1]. I kept the optional completion mask in TurnRecord and kept the OpenEnv full-mask conversion inside the adapter. Tests cover both masked and default behavior.

trl/experimental/async_grpo/openenv_harness.py
adithya-s-k: Agreed, trimmed the comments and docstrings and removed the repeated explanations. Kept the setup details in the README.

examples/async_grpo_harbor/launcher.py
```diff
@@ -0,0 +1,412 @@
...
+            os.killpg(proc.pid, signal.SIGKILL)
+        except ProcessLookupError:
+            pass
+        proc.wait(timeout=5)
+
+
+def _cleanup(*_: object) -> None:
+    _stopping.set()
+    for proc in list(_children):
+        _stop_process(proc)
```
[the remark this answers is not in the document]
adithya-s-k: Fixed in 5367603b. Trainer and warm-up are tracked children now. Signals exit the wait before cleanup stops them. Both SIGTERM and SIGINT tests pass.

examples/async_grpo_harbor/launcher.py
```diff
@@ -0,0 +1,412 @@
...
+            found = published.search(ansi.sub("", log.read_text(errors="replace")))
+            if found:
+                url = found.group(1).rstrip(".,")
+                print(f"[launcher] server published {url}; verifying it reaches the proxy", flush=True)
+        if url and _http_json_has(f"{url}/health", "status"):
+            print(f"[launcher] capture proxy reachable at {url}", flush=True)
+            return url
+        time.sleep(5)
+    tail = log.read_text(errors="replace")[-1500:] if log.exists() else "(no log)"
+    raise RuntimeError(f"the capture proxy never became publicly reachable in {deadline_s:.0f}s\n{tail}")
```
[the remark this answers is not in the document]
adithya-s-k: Fixed in 5367603b. The waiter checks the latest published URL on every attempt. Added a test where the first URL fails and a replacement succeeds.

trl/experimental/async_grpo/openenv_harness.py
```diff
@@ -256,7 +262,11 @@ def _run_session(self, prompt, group_id=0):
                 loop_exhausted=timed_out,
                 duration_s=time.monotonic() - t_dispatch,
             )
+            if self._rollout_reward_fn is not None and env_reward is not None:
+                metrics["correctness"] = env_reward
             return (completion, completion_ids, sequences, tool_call_count, tool_failure_count, reward), metrics
+        except CaptureContractError:
+            raise
```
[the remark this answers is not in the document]
adithya-s-k: Fixed in 5367603b. Reconciliation errors now propagate as CaptureContractError too. Added a test for worker failure and session cleanup.

trl/experimental/async_grpo/async_rollout_worker.py
AmineDiro: I see, but then we have two ways to express the same thing right ?

trl/experimental/async_grpo/vllm_client.py
```diff
@@ -123,8 +123,14 @@ def get_world_size(self) -> int:
...
     def pause(self, timeout: int = 1800) -> None:
-        """Pause generation while keeping the KV cache warm (`mode=keep`), so weights can be swapped in."""
-        response = requests.post(f"{self.server_url}/pause", params={"mode": "keep"}, timeout=timeout)
+        """Drain active requests and clear their caches before replacing the model weights.
+
+        New requests wait until resume. Keeping active requests would continue their decoding with new weights and
+        cached state produced by the previous weights.
+        """
+        response = requests.post(
+            f"{self.server_url}/pause", params={"mode": "wait", "clear_cache": True}, timeout=timeout
```
AmineDiro, line 132: Just catched that, why are we changing pausing mode ?

trl/trainer/utils.py
```diff
@@ -791,7 +791,8 @@ def format_entry(entry) -> Text:
                         t.append(reasoning, style="italic dim white")
                         t.append("\n")
                     if "content" in msg:
-                        t.append(msg["content"])
+                        # Tool-only assistant messages may have null content.
+                        t.append(msg["content"] or "")
```
AmineDiro, line 795: Nice catch

trl/experimental/async_grpo/async_rollout_worker.py
```diff
@@ -134,10 +141,26 @@ def _align_to_prompt(self, prompt_ids: list[int]) -> None:
...
+        if any(masks) and (logprobs is None or len(logprobs) != len(ids)):
+            raise ValueError("trainable tokens require one sampled logprob per token")
+        if logprobs is not None and (
+            len(logprobs) != len(ids)
+            or any(
+                isinstance(lp, bool) or not isinstance(lp, (int, float)) or not math.isfinite(lp) or lp > 0
+                for lp in logprobs
+            )
+        ):
+            raise ValueError("logprobs must be finite, nonpositive, and aligned with tokens")
```
AmineDiro: Are these checks necessary or just defensive programming ? I am not sure in what case we will have this ?

trl/experimental/async_distillation/vllm_client.py
```diff
@@ -76,8 +76,14 @@ def get_world_size(self) -> int:
...
     def pause(self, timeout: int = 1800) -> None:
-        """Pause generation while keeping the KV cache warm (`mode=keep`), so weights can be swapped in."""
-        response = requests.post(f"{self.server_url}/pause", params={"mode": "keep"}, timeout=timeout)
+        """Drain active requests and clear their caches before replacing the model weights.
+
+        New requests wait until resume. Keeping active requests would continue their decoding with new weights and
+        cached state produced by the previous weights.
+        """
+        response = requests.post(
+            f"{self.server_url}/pause", params={"mode": "wait", "clear_cache": True}, timeout=timeout
```
AmineDiro, line 85: Same question

trl/experimental/async_grpo/openenv_harness.py
```diff
@@ -125,15 +118,37 @@ def __init__(
...
+                        for key, value in {
+                            "top_p": self.top_p,
+                            "top_k": self.top_k,
+                            "min_p": self.min_p,
+                            "repetition_penalty": self.repetition_penalty,
+                        }.items()
+                        if value is not None
+                    },
+                }
+            )
```
AmineDiro:
```suggestion
if harness_adapter is None:
    sampling_kwargs = {
        "temperature": self.temperature,
        "top_p": self.top_p,
        "top_k": self.top_k,
        "min_p": self.min_p,
        "repetition_penalty": self.repetition_penalty,
    }
    sampling_kwargs = {k: v for k, v in sampling_kwargs.items() if v is not None}
    self._sampling_policy = training_sampling(sampling_kwargs)
```

trl/experimental/async_grpo/openenv_harness.py
```diff
@@ -331,37 +339,46 @@ def _entry_to_turn(entry: TraceEntry) -> HarnessTurn:
...
+            if any(captured.get(key) != value for key, value in sampling.items()):
+                raise ValueError(
+                    "capture sampling does not match the trainer policy; pass the trainer's temperature "
+                    "as sampling to HarborSessionFactory and use full-vocabulary sampling"
+                )
+        prompt_ids = entry.get("prompt_token_ids")
+        if not prompt_ids:
+            raise ValueError(
+                "a captured turn carried no `prompt_token_ids`. Serve the engine with "
+                "`--return-tokens-as-token-ids --logprobs-mode processed_logprobs`."
```
AmineDiro, line 361: is it just defensive, or would that really occur ? If we are passing sample kwargs, we would like to rely on the API of the session to use them without rechecking the trace afterward

AmineDiro (changes requested):

trl/experimental/async_grpo/vllm_client.py
adithya-s-k: This lets active requests finish before we replace the weights. With `keep`, a request can resume across a weight update. `wait` finishes it first and clears the old cache. It can increase sync latency; happy to split this into a separate fix if you prefer.

trl/experimental/async_distillation/vllm_client.py
adithya-s-k: Same reason as above. Both clients use the same weight-update flow, so I kept them consistent. If we split the fix out, I'll move both together.

trl/experimental/async_grpo/async_rollout_worker.py
adithya-s-k: Agreed, the value checks were duplicated here. Removed them from the builder and kept validation at the OpenEnv boundary. The builder only checks alignment and missing logprobs now, to avoid mismatched arrays or zero-filled probabilities for trained tokens.

trl/experimental/async_grpo/openenv_harness.py
adithya-s-k: Yep, simplified it as suggested.

trl/experimental/async_grpo/openenv_harness.py
adithya-s-k: I agree once the session receives the trainer's policy directly. Currently `create()` only takes the task, seed and episode ID; factory sampling is configured separately. A factory at temperature 1.0 with a trainer at 0.8 passes each side's checks. Added a test for that case. I'd keep this check until the contract connects the two.

trl/trainer/utils.py
adithya-s-k: Thanks!

adithya-s-k:
@AmineDiro updated with main in a02f8128. Simplified the sampling setup and removed duplicate value checks from the builder; 72 local tests pass.

The pause-mode and sampling-policy explanations are in the review threads, including a test for factory/trainer temperature mismatch. Kept those threads open for your feedback. Could you take another look? CI is rerunning on the updated branch.

examples/async_grpo_harbor/async_grpo_harbor.py
```diff
@@ -0,0 +1,263 @@
...
+    print(f"rollouts  {args.harness} on {args.sandbox}, {args.num_generations}x{args.max_inflight}")
+    print(f"tasks     {len(dataset)} from {args.split}")
+    print(f"output    {output_dir}")
+
+    tokenizer = AutoTokenizer.from_pretrained(args.model)
+
+    config = AsyncGRPOConfig(
+        output_dir=output_dir,
+        save_strategy="no",
+        save_total_limit=3,
```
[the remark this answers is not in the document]
adithya-s-k: Fixed in 1109d4f8. The example now checkpoints every 50 steps and saves the final model and tokenizer after training, including runs shorter than 50 steps. Added regression coverage; 15 example tests pass.

adithya-s-k:
Fixed the dev-dependency CI failure in d60b4073. Transformers #48793 fixed processor assistant masks, so the two strict Llava xfails now apply only before 5.18.0.dev0. The assertions are unchanged.

Verified against CI's exact Transformers revision: 4 passed. Released 5.14.1: 2 passed, 2 expected failures. Also merged the latest main; CI is rerunning.

trl/experimental/async_grpo/vllm_client.py
AmineDiro: yes this keep was intentionnal, as this will slow down Asyncgrpo. the clear_cache option is needed for kvcache clearing but I think keep is the mode we want to keep

trl/experimental/async_grpo/openenv_harness.py
AmineDiro: Ok isee, in that case I would prefer to redesign the API than to add defensive code, this clearly show design mismatch I think

trl/experimental/async_grpo/vllm_client.py
adithya-s-k: Switched back to `keep` with `clear_cache=True` in 64cb31de. Kept AsyncGRPO and AsyncDistillation aligned and updated the HTTP tests.

trl/experimental/async_grpo/openenv_harness.py
adithya-s-k: Changed the API in 64cb31de: the worker now calls the factory constructor with its own `sampling` policy. The Harbor example uses `partial(HarborSessionFactory, ...)`, so it no longer configures sampling separately. Removed the post-trace comparison and tested that the policy reaches the session constructor.

examples/async_grpo_harbor/async_grpo_harbor.py
```diff
@@ -0,0 +1,267 @@
...
+        sandbox=args.sandbox,
+        llm_url=args.vllm_url,
+        model=args.model,
+        agent_timeout_sec=args.agent_timeout,
+        agent_step_limit=args.agent_step_limit,
+        reward_key=args.reward_key,
+        num_tasks=args.n_tasks,
+        indices=task_indices(args.task_indices),
+    )
+    dataset = Dataset.from_list(factory().prompt_rows())
```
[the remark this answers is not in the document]
adithya-s-k: Fixed in f0f96b9b. The parent metadata client is now closed in `finally` before the worker starts, including when dataset loading fails. Added regression tests against the pinned Harbor factory. Also merged latest main; all 71 contract tests pass locally.

PR #6955, opened 2026-08-28
Validate reward model tokenizer matches policy in XPOTrainer/NashMDTrainer

### LINKED ISSUE #6951: XPOTrainer accepts reward_processing_classes but never uses it, and NashMDTrainer does not accept it at all
### Summary

`XPOTrainer` accepts and documents a `reward_processing_classes` argument that its reward path never uses, so a reward tokenizer passed by a user is silently ignored. `NashMDTrainer` does not expose the argument at all and hardwires it to `processing_class`. The two sibling trainers disagree about the same parameter, and one of the two behaviors is misleading.

### What the code does

`XPOTrainer.__init__` declares the parameter (`trl/experimental/xpo/xpo_trainer.py:127`), documents it (`:77`), and forwards it to `super().__init__` (`:143`).

`NashMDTrainer.__init__` does not declare it, and passes the policy's own processing class to the parent instead (`trl/experimental/nash_md/nash_md_trainer.py:201`):

```python
reward_processing_classes=processing_class,
```

Both trainers then score completions the same way, through the legacy `get_reward` helper, using the policy's processing class and the policy's raw token ids (`xpo_trainer.py:257`, `nash_md_trainer.py:326`):

```python
_, model_scores, _ = get_reward(
    self.reward_funcs, model_data["input_ids"], self.processing_class.pad_token_id, context_length
)
```

Nothing on that path consults `reward_processing_classes`.

### Why it matters

`OnlineDPOTrainer` decodes completions and re-tokenizes them with `reward_processing_classes`, so its reward model may use a different vocabulary from the policy. `XPOTrainer` and `NashMDTrainer` feed policy token ids straight into the reward model, so the reward model must share the policy's vocabulary.

That constraint is undocumented and fails late. Passing a reward model with a smaller vocabulary produces an embedding lookup error far from the cause:

```
IndexError: index out of range in self
```

I hit this while building a harness for #2228, using a Llama reward model with a Qwen policy. The trainers' own tests avoid it by building the reward head from the same model id as the policy, which is why it is not caught in CI.

### Options

1. Drop the parameter from `XPOTrainer` and match `NashMDTrainer`, which reflects what the reward path actually supports.
2. Make both reward paths honor `reward_processing_classes` by decoding and re-tokenizing the way `OnlineDPOTrainer` does.
3. Keep the signatures and raise at construction when a reward processing class is supplied that differs from the policy's, so the failure is immediate and explains itself.

Option 1 is the smallest change and matches the repository's preference for handling only the cases that exist today. Option 3 costs a few lines and converts a confusing `IndexError` into a clear message.

I am happy to open a PR for whichever the maintainers prefer. Flagging rather than assuming, since this is an API decision rather than a defect with one obvious fix.

amanyagami:
# What does this PR do?

`XPOTrainer` and `NashMDTrainer` (both subclasses of `experimental.online_dpo.OnlineDPOTrainer`) score completions
with the `get_reward()` helper, which runs the reward model directly on the **policy**-tokenized `input_ids`:

```python
# xpo_trainer.py / nash_md_trainer.py
_, model_scores, _ = get_reward(
    self.reward_funcs, model_data["input_ids"], self.processing_class.pad_token_id, context_length
)
```

Unlike `OnlineDPOTrainer._calculate_rewards_from_functions`, which decodes completions with the policy's tokenizer
and *re-tokenizes* them per reward function with `reward_processing_class` before scoring, `XPOTrainer` and
`NashMDTrainer` never decode/re-tokenize: they feed the policy's token IDs straight into the reward model's forward
pass. That's only correct if the reward model shares the exact same vocabulary (token -> ID mapping) as the policy
model — otherwise the reward model scores the wrong tokens, which either silently produces meaningless rewards or
crashes deep inside an embedding lookup with an unhelpful `IndexError`.

On top of that:
- `XPOTrainer` accepts and documents `reward_processing_classes`, but never actually uses it anywhere in
  `_compute_rewards` — passing a reward tokenizer that differs from the policy's just wasn't a way to avoid the
  bug, and there was no signal to the user that it silently does nothing.
- `NashMDTrainer` doesn't accept `reward_processing_classes` at all, and hardcodes
  `reward_processing_classes=processing_class` in its call to `super().__init__()`, so it always feeds the
  policy's tokenizer to the (potentially different) reward model, with no way to override it and no validation that
  this is actually safe.

## Why this fix (see issue for the 3 options considered)

The issue lists three options: (1) implement full decode/re-tokenize per reward model like `OnlineDPOTrainer` does,
(2) add validation + document the shared-vocab constraint, or (3) something else. I went with **(2)**:

`_compute_rewards` in both trainers uses `get_reward()`, a single-model helper built around raw `query_responses`
token IDs (used for computing sequence lengths, building the attention mask from `pad_token_id`, etc.) — not
`OnlineDPOTrainer`'s per-reward-function decode/apply-chat-template/re-tokenize path. Rewriting `_compute_rewards`
to decode and re-tokenize would touch the padding/masking/EOS-detection logic in both trainers and is a much larger,
higher-risk diff for two experimental trainers I can't validate on GPU. Since the issue explicitly allows the
safer, minimal option when a full rewrite is too large, I added a validation check + documentation instead of
reimplementing the reward pipeline.

## Changes

- Added `validate_reward_processing_class_shares_vocab()` to `trl/experimental/utils.py`: given a reward function,
  its processing class, and the policy's processing class, it's a no-op for non-model reward functions and raises a
  clear `ValueError` if a model-based reward function's tokenizer vocabulary doesn't match the policy's.
- `XPOTrainer.__init__`: call the new validation after resolving `self.reward_funcs` to a single model. Documented
  the shared-vocab constraint in the `reward_processing_classes` docstring entry.
- `NashMDTrainer.__init__`: added the missing `reward_processing_class` argument (singular, since NashMD only
  supports one reward model) so callers can now point it at the reward model's own tokenizer. When not provided, it
  still defaults to `processing_class`, preserving the existing default behavior exactly. Also call the new
  validation, and documented the constraint in the docstring.

### Why this doesn't break existing behavior

For both trainers, when `reward_processing_class(es)` isn't explicitly overridden with something incompatible, the
validation either compares a tokenizer to itself (`NashMDTrainer`'s unchanged default) or compares the
auto-loaded reward-model tokenizer to the policy tokenizer, which only raises when they were already silently
mismatched (previously broken, just with a worse failure mode later). A genuinely matching setup is unaffected.

## Tests

Added regression tests in `tests/experimental/test_xpo_trainer.py` and `tests/experimental/test_nash_md_trainer.py`:
- `test_reward_processing_class_vocab_mismatch_raises` (both trainers): passing a reward tokenizer with a different
  vocabulary (a tiny GPT-2 tokenizer against a Qwen2 policy) raises `ValueError` mentioning "vocabulary".
- `test_reward_processing_class_matching_vocab_is_accepted` (XPO) / `test_reward_processing_class_can_be_passed_explicitly`
  (NashMD): explicitly passing a matching processing class is accepted and stored.
- `test_reward_processing_class_defaults_to_processing_class` (NashMD): omitting `reward_processing_class` still
  defaults to `processing_class`, confirming the previous default behavior is preserved.

I confirmed all 5 new tests fail on `main` (2 with `TypeError: unexpected keyword argument` for NashMD, since the
parameter didn't exist; the 2 mismatch tests fail because nothing validates the mismatch) and pass with the fix.

### Test output (after fix)

Full suite for both trainers (includes the existing training/PEFT tests, which run actual tiny-model training steps
on CPU, in addition to the 5 new reward-tokenizer tests):

```
$ python -m pytest tests/experimental/test_xpo_trainer.py tests/experimental/test_nash_md_trainer.py -q
..........................                                               [100%]
26 passed, 1 warning in 334.71s (0:05:34)
```

Just the new reward-tokenizer tests:

```
$ python -m pytest tests/experimental/test_xpo_trainer.py tests/experimental/test_nash_md_trainer.py -k "reward_processing_class" -v
tests/experimental/test_xpo_trainer.py::TestXPOTrainer::test_reward_processing_class_vocab_mismatch_raises PASSED
tests/experimental/test_xpo_trainer.py::TestXPOTrainer::test_reward_processing_class_matching_vocab_is_accepted PASSED
tests/experimental/test_nash_md_trainer.py::TestNashMDTrainer::test_reward_processing_class_vocab_mismatch_raises PASSED
tests/experimental/test_nash_md_trainer.py::TestNashMDTrainer::test_reward_processing_class_can_be_passed_explicitly PASSED
tests/experimental/test_nash_md_trainer.py::TestNashMDTrainer::test_reward_processing_class_defaults_to_processing_class PASSED
5 passed in 16.22s
```

I confirmed these 5 fail on `main` (`git stash` the source changes, keep the tests): 2 with `TypeError: unexpected
keyword argument 'reward_processing_class'` for `NashMDTrainer` (the parameter didn't exist), and the 2 mismatch
tests with no exception raised at all (nothing validated the mismatch).

`pre-commit run --files trl/experimental/utils.py trl/experimental/xpo/xpo_trainer.py trl/experimental/nash_md/nash_md_trainer.py tests/experimental/test_xpo_trainer.py tests/experimental/test_nash_md_trainer.py`
passes (ruff check, ruff format, doc-builder style check all green).

Fixes #6951

## Before submitting

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **Medium Risk**
> Init-time validation can reject previously accepted (but incorrect) reward/policy tokenizer pairings; changes affect experimental online RLHF training paths where reward scoring is security-sensitive for training quality but not auth or data exfiltration.
> 
> **Overview**
> **XPOTrainer** and **NashMDTrainer** score completions by passing policy **`processing_class`** token IDs straight into **`get_reward`**, so a mismatched reward tokenizer can silently corrupt rewards or crash in embeddings. This PR adds **`validate_reward_processing_class_shares_vocab`** in **`trl/experimental/utils.py`** and runs it at trainer init (after unwrapping DDP/DeepSpeed-wrapped reward models).
> 
> **NashMDTrainer** gains an optional **`reward_processing_class`** (default **`processing_class`**) instead of always forcing the policy tokenizer. Docstrings document the shared-vocabulary requirement for both trainers.
> 
> Regression tests cover vocab mismatch **`ValueError`**, matching vocabs, DDP-wrapped reward models, **`ProcessorMixin`** unwrapping, and NashMD defaults.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4243e9dbc1dcd828f28f2bc43f414c882e9269ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amanyagami:
Both issues flagged by Cursor Bugbot are fixed in fbfba768:

1. **Validation skipped after model wrapping** — `self.reward_funcs` can already be wrapped (DeepSpeed engine / DDP) by `Accelerator.prepare_model` inside `super().__init__()` by the time the check runs. Fixed by unwrapping with `self.accelerator.unwrap_model()` at both call sites before the `isinstance(reward_func, PreTrainedModel)` check. Added a regression test that wraps the reward model in a real single-process `DistributedDataParallel` (accelerate's `extract_model_from_parallel` only recognizes real wrapper types, not a hand-rolled stand-in) and confirms the check still fires.

2. **Reward processor vocab lookup crashes** — the helper now unwraps a `ProcessorMixin` on the reward side the same way it already did on the policy side, so `NashMDTrainer` defaulting `reward_processing_class` to a `ProcessorMixin` `processing_class` (VLM policies) no longer raises `AttributeError` from `get_vocab()`. Added direct unit tests for both the matching-vocab (must not raise) and mismatched-vocab (must still raise `ValueError`) cases.

Both new regression tests were confirmed to fail without their respective fix and pass with it. Full `tests/experimental/test_xpo_trainer.py` + `test_nash_md_trainer.py` + `test_utils.py` suite: 34 passed, 1 pre-existing unrelated skip. `ruff check`/`ruff format --check` clean.

qgallouedec:

PR #6968, opened 2026-08-29
Fix mixed-image Online DPO generation

### LINKED ISSUE #6967: Online DPO vLLM server fails on mixed image and text-only batches
### Reproduction

On current `main` at `6630e17a42976fb1db84034e94e0cf058e8baa21`, Online DPO's vLLM server path fails for a conversational VLM batch that mixes rows with and without images.

`training_step` inserts an unfilled image placeholder into every conversational prompt whenever the batch has an `image` column. `_generate_vllm_server` then passes `images=None` for a row whose image value is `None`:

```python
from trl.data_utils import prepare_multimodal_messages

prompt = [
    {
        "role": "user",
        "content": [
            {"type": "image"},
            {"type": "text", "text": "Say hello"},
        ],
    }
]

prepare_multimodal_messages(prompt, images=None)
```

This produces:

```text
ValueError: Number of images provided (0) does not match number of image placeholders (1).
```

The failure is specific to the no-image rows of a mixed batch. Rows with an image still need the placeholder populated before `vllm_client.chat`.

I searched open issues and PRs for this failure and found no overlapping implementation. I will submit a focused fix that removes only unfilled image placeholders from rows whose image value is `None`, preserves embedded image blocks, and adds a server-path regression test for a mixed batch.

### System Info

- Platform: Windows-10-10.0.22631-SP0
- Python: 3.10.11
- TRL: 1.10.0.dev0 (`6630e17a42976fb1db84034e94e0cf058e8baa21`)
- PyTorch: 2.13.0+cu130
- Transformers: 5.14.1
- Accelerate: 1.14.0
- Datasets: 5.0.0

### Checklist

- [x] Any code provided is minimal, complete, and reproducible
- [x] Any code provided is properly formatted in code blocks

DaoyuanLi2816:
# What does this PR do?

Fixes #6967.

Online DPO's image handling added a placeholder to every prompt whenever a batch contained an `image` column. Mixed image/text-only batches therefore sent an unfilled image placeholder through one or more generation paths: server mode raised a placeholder-count error, colocate mode paired an image token with no multimodal input, and the Transformers processor received `None` as an image.

This change fixes the problem where it originates in `training_step`, following the same pattern as the sibling trainers:

- normalize each row with `prepare_multimodal_messages` and its own image list;
- leave text-only rows without image placeholders while preserving image blocks that already contain an embedded image;
- pass the prepared messages directly to vLLM server mode;
- pass empty image lists, rather than `[None]`, to the Transformers processor.

Regression coverage drives `training_step` into the server path and separately covers colocate and Transformers generation. The server test includes a dataset image row, a text-only row, and a row with an already embedded image.

This also follows up on the review finding in #6765: https://github.com/huggingface/trl/pull/6765#discussion_r3864591931

Validation:

```text
python -m pytest \
  tests/experimental/test_online_dpo_trainer.py::test_training_step_prepares_mixed_image_batch_for_vllm_server \
  tests/experimental/test_online_dpo_trainer.py::test_generate_transformers_with_mixed_image_batch \
  tests/experimental/test_online_dpo_trainer.py::test_generate_vllm_colocate_with_mixed_image_batch \
  -q --no-cov
3 passed

python -m pytest \
  tests/experimental/test_online_dpo_trainer.py::TestOnlineDPOVisionTrainer::test_online_dpo_vlm_trainer \
  -q --no-cov
2 passed

pre-commit ruff-check and ruff-format
passed

git diff --check
passed
```

## Before submitting

- [ ] This PR fixes a typo or improves the docs (not applicable).
- [x] I read the contributor guideline, Pull Request section.
- [x] The bug and intended implementation are tracked in #6967.
- [x] No user-facing documentation update is required for this regression fix.
- [x] I added the necessary regression tests.

## AI writing disclosure

## Who can review?

### DIFF
--- a/tests/experimental/test_online_dpo_trainer.py
+++ b/tests/experimental/test_online_dpo_trainer.py
@@ -35,16 +35,19 @@ if is_vision_available():
     from transformers import AutoModelForImageTextToText, AutoProcessor
 
 
-def test_generate_vllm_server_with_mixed_image_batch(monkeypatch):
+def test_training_step_prepares_mixed_image_batch_for_vllm_server(monkeypatch):
+    class StopAfterGeneration(Exception):
+        pass
+
     class DummyProcessor:
         def __call__(self, **kwargs):
             self.text = kwargs["text"]
-            return {"input_ids": torch.tensor([[1], [2]])}
+            return {"input_ids": torch.tensor([[1], [2], [3]])}
 
     class DummyClient:
         def chat(self, messages, **kwargs):
             self.messages = messages
-            return {"completion_ids": [[3], [4], [5], [6]]}
+            return {"completion_ids": [[4], [5], [6], [7], [8], [9]]}
 
     monkeypatch.setattr(
         online_dpo_trainer,
@@ -59,8 +62,9 @@ def test_generate_vllm_server_with_mixed_image_batch(monkeypatch):
     trainer = OnlineDPOTrainer.__new__(OnlineDPOTrainer)
     trainer.state = SimpleNamespace(global_step=0)
     trainer._move_model_to_vllm = lambda: None
+    trainer._tokenizer = SimpleNamespace(eos_token_id=0, pad_token_id=0)
     trainer.processing_class = DummyProcessor()
-    trainer.accelerator = SimpleNamespace(is_main_process=True, process_index=0)
+    trainer.accelerator = SimpleNamespace(is_main_process=True, process_index=0, device="cpu")
     trainer.num_generations = 2
     trainer.repetition_penalty = 1.0
     trainer.temperature = 1.0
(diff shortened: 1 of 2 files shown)


behroozazarkhalili:
Verdict: correct with changes. The fix is real and the test fails before and passes after, but it patches the third of three affected call sites instead of the one place that creates the bad prompt, and two other generation paths in the same trainer still fail on the same batch.

## What I checked

New test on PR head (70f0231d):

    pytest tests/experimental/test_online_dpo_trainer.py::test_generate_vllm_server_with_mixed_image_batch -q
    1 passed, 2 warnings in 16.49s        RC_HEAD=0

Same test against base source 6630e17a with the PR's tests kept:

    ValueError: Number of images provided (0) does not match number of image placeholders (1).
    trl/data_utils.py:105: ValueError
    1 failed, 2 warnings in 12.36s        RC_BASE=1

That is the exact traceback in #6967, so RED/GREEN holds and the test reproduces the reported failure.

Cursor Bugbot's finding ("text-only prompts keep image tokens" in prompts_text and prompt_ids) was valid against the first commit and is fixed by the second. Running the head test against the 4066f7fa source gives `AssertionError: assert ['image text', 'image text'] == ['image text', 'text']`, so the new third assertion is what closes it. That comment can be marked resolved.

`ruff check` and `ruff format --check` at the CI-pinned 0.13.3 pass on both changed files.

## Findings

1. `trl/experimental/online_dpo/online_dpo_trainer.py:1143-1155` is the root cause. `training_step` inserts `{"type": "image"}` into every user message as soon as the batch has an `image` column, without looking at that row's own image value. I replicated the loop verbatim on a mixed batch and the `image=None` row comes back as `[{'type': 'image'}, {'type': 'text', 'text': 'Say hello'}]`. Every sibling trainer instead builds prompts with `prepare_multimodal_messages(prompt, images=image_list)` per row (`grpo_trainer.py:2438-2441`, `rloo_trainer.py:1389-1392`, `distillation_trainer.py:1605-1608`, `grpo_with_replay_buffer_trainer.py:138-141`, `gold_trainer.py:1357-1360`). Running that pattern on the same mixed batch gives `[('text', False)]` for the None row: no placeholder, nothing to strip. Replacing the hand-rolled loop with the sibling call fixes all three generation paths at once, deletes the 16 new lines rather than adding them, and brings this trainer into line with the rest of the repo.

2. `online_dpo_trainer.py:981-982` still fails on the same batch. The non-vLLM transformers path does `kwargs = {"images": [[img] for img in images]}`, which is `[[<PIL>], [None]]` for a mixed batch. Passing exactly that to `AutoProcessor.from_pretrained("trl-internal-testing/tiny-Idefics2ForConditionalGeneration")` raises `TypeError: Incorrect format used for image. Should be an url linking to an image, a base64 string, a local path, or a PIL image.` A user who hits #6967 without vLLM gets a different crash on the same data.

3. `online_dpo_trainer.py:744-751`, the vLLM colocate path, is also unfixed. The stray placeholder renders straight into `prompts_text`, then the row is appended as a bare string with no `multi_modal_data`. `apply_chat_template` on the None-image row returns `'User:<image>Say hello<end_of_utterance>\nAssistant:'`, so vLLM receives an image token with no image for that prompt.

4. `tests/experimental/test_online_dpo_trainer.py:38-84` has a surviving mutant. Deleting the `or "image" in part` guard, so that every image part is stripped from a None-image row, still passes (`1 passed`). The PR body claims embedded image blocks are preserved, and nothing tests it. Adding a third row whose content already carries `{"type": "image", "image": <obj>}` and asserting it survives would close that.

5. `tests/experimental/test_online_dpo_trainer.py:64-80` constructs the trainer with `OnlineDPOTrainer.__new__` and hand-sets twelve attributes, then calls the private `_generate_vllm_server` with placeholders already in place. That pins the test to the current location of the fix: if the fix moves to `training_step` as suggested in finding 1, this test still fails, because for this entry point the upstream-fixed source is identical to base. A test that drives `training_step` (or at least builds the prompts the way `training_step` does) would survive the better fix.

## Not reproduced

Nothing in the PR body or the Bugbot comment failed to reproduce.

## Unevaluable

The end-to-end mixed-batch failure of the colocate path and of the transformers path inside a real `training_step` call: both need a VLM plus, for colocate, a vLLM install and a GPU. Findings 2 and 3 are grounded in the processor and chat-template calls those paths make, not in a full training step. `probe.sh` step 4 runs the trainer's VLM suite on a compute node.

DaoyuanLi2816:
@behroozazarkhalili addressed the review in a9d918c. The fix now normalizes messages once in `training_step` with `prepare_multimodal_messages`, so server, colocate, and Transformers generation all receive consistent mixed-batch inputs. I also replaced the private server-only test with a `training_step` regression, added colocate and Transformers coverage, and included an embedded-image row. The 3 focused regressions and both existing VLM trainer cases pass, along with Ruff check/format. Ready for another look.

DaoyuanLi2816:
@behroozazarkhalili Could you re-review a9d918c when you have a chance? It addresses your September 3 feedback, with normalization moved to training_step and coverage for all three generation paths. The update has been waiting since September 4; the Actions workflows still need contributor approval.

behroozazarkhalili:
Re-reviewed a9d918c6. Verdict: the September 3 findings are all addressed, and I have no new blocking finding. This is ready from my side.

## What I checked

- The three new tests pass on the PR head (`pytest tests/experimental/test_online_dpo_trainer.py -k mixed_image`: 3 passed). The full file on a CPU compute node: 21 passed, 3 skipped, 0 failed (the 3 skips are the GPU and vLLM gated tests).
- `ruff check` and `ruff format --check` at the CI-pinned 0.13.3 pass on both files; the pinned doc-builder style check passes too.
- The branch merges cleanly onto current main (100 commits behind, `git merge-tree` reports no conflicts).

## Findings from September 3, one by one

1. Root cause in `training_step`: fixed. The hand-rolled placeholder loop is gone and prompts go through `prepare_multimodal_messages(prompt, images=image_list)` per row, the same call the GRPO, RLOO, distillation and GOLD trainers make. A `None` image becomes an empty list, so no placeholder is inserted and nothing has to be stripped afterwards. The `_generate_vllm_server` patch from the first round is correctly removed.
2. Transformers path: fixed, and I confirmed it on real processors rather than the dummy. With a small PIL image and a mixed two-row batch, `processor(text=..., images=[[img], []])` succeeds on both `tiny-Qwen2VLForConditionalGeneration` (input_ids `[2, 26]`, pixel_values present) and `tiny-Idefics2ForConditionalGeneration` (`[2, 335]`). The base shape `[[img], [None]]` fails on both: Qwen2VL raises `TypeError: only a single or a list of entries is supported but got type=<class 'NoneType'>`, Idefics2 raises `The number of images in the text [1, 0] and images [1, 1] should be the same`.
3. Colocate path: fixed by the same change, since the stray placeholder no longer reaches `prompts_text`; the existing `image is not None` branch then does the right thing. The new colocate test pins the two shapes (`multi_modal_data` dict for the image row, bare string for the text row).
4. Embedded-image guard mutant: no longer applicable, the guard does not exist. The training_step test now carries a third row with an embedded `{"type": "image", "image": obj}` block and asserts it survives, which is the coverage I asked for.
5. Test drives `training_step`: done. The test enters through `training_step` with a real `_generate_vllm` call and checks the prompts handed to generation, the messages sent to the client, and the processor text.

## One pre-existing limit, not from this PR

The `vision_inputs` block in `training_step` still calls the processor with `text=[""] * len(image_lists)`. On Idefics2 that fails regardless of this PR (`The number of images in the text [0, 0] and images [1, 1]` on base with a full-image batch, `[0, 0]` vs `[1, 0]` on this head with a mixed batch), because Idefics2 counts image tokens in the text. Qwen2VL accepts the dummy text. So the transformers VLM path on Idefics2 was already broken before this PR and stays broken after it; worth a separate issue, not a blocker here.

## Unevaluable

An end-to-end mixed-batch `training_step` on a real VLM with vLLM (server or colocate) still needs a GPU and a vLLM install; the three paths are covered here at the unit level with the real `prepare_multimodal_messages` and, for the transformers path, real processors.

PR #6971, opened 2026-08-29
fix(data): handle all-empty BFD packing

### LINKED ISSUE #6970: Bug: pack_dataset BFD fails when all sequences are empty
## Reproduction

`pack_dataset` filters empty sequences before the BFD binning step. When every typed token sequence is empty, the bin list is also empty. Passing the resulting untyped Python list of indices to Arrow makes it a `null` array, so both BFD strategies fail instead of returning an empty packed dataset.

```python
from datasets import Dataset, Features, Sequence, Value
from trl import pack_dataset

dataset = Dataset.from_dict(
    {"input_ids": [[], []]},
    features=Features({"input_ids": Sequence(Value("int64"))}),
)

pack_dataset(dataset, seq_length=4, strategy="bfd")
```

Changing `strategy` to `"bfd_split"` produces the same failure.

Output:

```
pyarrow.lib.ArrowNotImplementedError: Function 'array_take' has no kernel matching input types (list<item: int64>, null)
```

The expected result is an empty packed dataset with the original list column and an empty `seq_lengths` column.

## System Info

- Platform: macOS-26.5.2-arm64-arm-64bit-Mach-O
- Python version: 3.13.5
- TRL version: 1.13.0.dev0+6630e17
- PyTorch version: 2.13.0 (MPS available)
- Transformers version: 5.16.1
- Datasets version: 5.0.1
- PyArrow version: 25.0.1

## Checklist

YZJF:
# What does this PR do?

Fixes #6970.

`pack_dataset(..., strategy="bfd")` and `"bfd_split"` filter empty input sequences before binning. If every typed input sequence is empty, this leaves no bins. The empty Python list passed to `pyarrow.compute.take` is inferred as a `null` index array, which raises `ArrowNotImplementedError` instead of producing an empty packed dataset.

Construct the take indices as an explicit `int64` Arrow array. This preserves the existing behavior for non-empty indices and lets both BFD strategies return an empty dataset with `seq_lengths` when all input sequences are empty.

The follow-up regression also covers a non-empty input whose second `Dataset.map` batch becomes all empty: `[[1, 2], [3, 4], [], []]` with `batch_size=2`.

## Tests

- RED before the production change: the focused test fails for both `bfd` and `bfd_split` with `ArrowNotImplementedError: Function 'array_take' has no kernel matching input types (list<item: int64>, null)`.
- GREEN: `python -m pytest tests/test_data_utils.py::TestPackDatasetBfd::test_with_all_empty_sequences -q` — 4 passed.
- Related suite: `python -m pytest tests/test_data_utils.py -k 'TestPackDatasetBfd' -q` — 11 passed.
- `pre-commit run --files tests/test_data_utils.py` and `git diff --check` — passed.

## Before submitting

- [x] Was this discussed/approved via a GitHub issue? Fixes #6970.
- [x] Did you make sure to update the documentation with your changes? No documentation change is needed for this internal empty-input fix.

## AI writing disclosure

## Who can review?


behroozazarkhalili:
Verdict: the fix is correct. One adjacent pre-existing gap is worth deciding on before merge.

## What I checked

pyarrow 23.0.1, datasets 4.8.4, transformers 5.11.0, `trl` resolved to the PR worktree.

RED/GREEN. The new test ids on the PR head:

```
$ python -m pytest "tests/test_data_utils.py::TestPackDatasetBfd::test_with_all_empty_sequences" -q
2 passed, 1 warning in 9.99s
```

Same test ids against a copy of the tree with `trl/data_utils.py` reverted to 6630e17a and the new tests kept:

```
E   pyarrow.lib.ArrowNotImplementedError: Function 'array_take' has no kernel matching input types (list<item: int64>, null)
FAILED tests/test_data_utils.py::TestPackDatasetBfd::test_with_all_empty_sequences[bfd]
FAILED tests/test_data_utils.py::TestPackDatasetBfd::test_with_all_empty_sequences[bfd_split]
2 failed, 1 warning in 12.91s
```

The failure text matches the traceback in #6970 exactly, so the test is a real oracle for the reported bug.

Issue scenario. I ran the issue's snippet directly on both trees. Pre-fix it raises the `array_take` error for `bfd` and `bfd_split`; post-fix both return `{'input_ids': [], 'seq_lengths': []}` with features `{'input_ids': List(Value('int64')), 'seq_lengths': List(Value('int32'))}`.

Beyond the issue's case I also probed a mixed dataset (`[[], [1,2,3], []]`), an all-empty second batch (`[[1,2],[3,4],[],[]]` with `batch_size=2`), a zero-row dataset, and the `wrapped` strategy, on both trees. The all-empty-second-batch case fails pre-fix with the same error and passes post-fix, so the fix also covers the more likely real-world shape where only one `map` batch is entirely empty. `wrapped` was already fine on both trees.

Regression and lint. `pytest tests/test_data_utils.py -k PackDataset` gives `11 passed` on the head. `ruff check` and `ruff format --check` at the CI-pinned 0.13.3 both pass on the two changed files.

Sibling consistency. `pc.take` appears only at `trl/data_utils.py:808`, and both call sites (`trl/trainer/sft_trainer.py:1664`, `trl/experimental/gold/gold_trainer.py:2230`) go through the shared `pack_dataset`, so there is no duplicated copy to keep in step.

## Findings

1. `trl/data_utils.py:929` (Medium, pre-existing, not a regression). A dataset with zero rows still fails for both BFD strategies, with a different error: `ValueError: Columns ['seq_lengths'] not in the dataset. Current columns in the dataset: ['input_ids']`. `Dataset.map` never invokes `_pack_bfd` on an empty dataset, so `seq_lengths` is never created, but line 929 appends it to `format["columns"]` unconditionally. I confirmed this fails identically on 6630e17a, so it is not caused by this PR. It is the immediately adjacent case to the one being fixed though, and a user who filters a dataset down to nothing lands on it. Suggestion: either guard line 929 on the column actually being present, or state in the PR that the zero-row case is deliberately left for a separate change.

2. `tests/test_data_utils.py:1343` (Low). The new test covers only a dataset that is entirely empty. The case that actually reaches production, one all-empty `map` batch inside a non-empty dataset, is what the fix also repairs and is not asserted. Suggestion: add `[[1, 2], [3, 4], [], []]` with `map_kwargs={"batch_size": 2}` as a second parametrization, which fails pre-fix and passes post-fix.

3. `tests/test_data_utils.py:1343` (Low). No equivalent case for `strategy="wrapped"`, which I verified already returns `{'input_ids': []}` on both trees. Adding it to `TestPackDatasetWrapped` would pin that behavior rather than leave it incidental.

4. `trl/data_utils.py:808` (Low, informational). The test does not pin the index dtype: mutating `pa.int64()` to `pa.int32()` still passes all 11 packing tests. Since indices are bounded by the `map` batch size, the two are behaviorally equivalent and `int64` is the right choice; no change requested.

## Not reproduced

Nothing in the PR body was contradicted. The claims of RED before, GREEN after, and `9 passed` for `TestPackDatasetBfd` all match what I measured (`TestPackDatasetBfd` collects exactly 9 tests on the head).

## Unevaluable

An end-to-end `SFTTrainer` run with `packing=True` on an all-empty dataset. It needs a model download and more memory than the node I reviewed on allows; the probe script covers it if a maintainer wants it run.

qgallouedec:
closing, see https://github.com/huggingface/trl/issues/6970#issuecomment-5548937237

Issue #6973, opened 2026-08-29
VLLMClient sets no request timeout, so a stalled vLLM server hangs training indefinitely


behroozazarkhalili:
### Summary

`VLLMClient` issues every HTTP request without a timeout, so a vLLM server that accepts a connection and then stalls hangs the training process indefinitely rather than failing. The retry policy that looks like it covers this cannot fire, because without a timeout a stalled read never raises the exception the retry is watching for.

### Where

`trl/generation/vllm_client.py`. None of the ten call sites passes `timeout=`, and `_get`/`_post` forward `**kwargs` without supplying a default:

```python
def _get(self, url: str, **kwargs) -> dict:
    response = self.session.get(url, **kwargs)      # :246
    ...
def _post(self, url: str, **kwargs) -> dict:
    response = self.session.post(url, **kwargs)     # :252
```

The ten sites: `:243` `/v1/models`, `:297` `get_world_size`, `:325` render, `:419` and `:655` `/v1/completions`, `:445` and `:572` generate/chat, `:756` `start_weight_update`, `:760` `finish_weight_update`, `:826` `reset_prefix_cache`. `check_server` at `:273` also calls `requests.get(url)` bare.

`requests` defaults to `timeout=None`, which blocks forever.

### Why the retry does not save it

The session mounts a retry strategy at `:216`:

```python
retry_strategy = Retry(
    total=5,
    connect=5,
    read=5,   # retry failures while reading the response after the connection was successfully established
    ...
```

`read=5` retries a read that *fails*. A read with no timeout does not fail, it blocks, so this branch is unreachable for the stall case. The comment above it says the strategy exists to keep "an otherwise healthy training run" alive through transient failures, which is exactly the case it silently does not cover.

### Impact

Any server-side stall after the TCP connection is established. A wedged engine, a GPU fault on the server, a hung weight transfer, or a network path that blackholes established connections all present the same way: the trainer sits in `recv` with no error, no log line, and no progress. On multi-rank runs the other ranks then hit their NCCL timeout, which is what the symptom is usually reported as. #3433 is one instance of that shape, though its own root cause was a TRL-side worker that no longer exists.

### The design question

A blanket timeout is wrong: `/v1/completions` legitimately runs for minutes on long rollouts, and a fixed ceiling would break real training. The split that looks right is by endpoint class:

| class | endpoints | suggested |
|---|---|---|
[...]

PR #6977, opened 2026-08-29
Bound the check_server health probe so connection_timeout can fire

behroozazarkhalili:
Server readiness checks could hang on a stalled connection or start another probe after the deadline. I bounded health probes, added deadline checks and retry interval validation, and aligned both async clients' probes and sleeps with the remaining budget. Tests against stalled, unavailable, and healthy local servers pass, including cases that fail or hang without the fix.
Refs #6973.

---

> [!NOTE]
> **Medium Risk**
> Changes only health-check polling in vLLM clients, but fixes behavior trainers rely on at startup; mis-tuned timeouts could fail connects that previously hung indefinitely.
> 
> **Overview**
> Fixes vLLM client startup waits that could **block past** `connection_timeout` / `server_timeout` when `/health` accepts TCP but never responds, returns **503** while loading, or when retry/poll sleeps overshot the deadline.
> 
> **`VLLMClient.check_server`** now caps each `/health` request with `timeout=min(retry_interval, remaining budget)`, treats non-200 (including 503) as “not up” when enforcing `total_timeout`, clamps inter-retry sleep to the remaining budget, re-checks the deadline before the next probe, uses **`time.monotonic()`** for deadlines, and rejects non-positive `retry_interval`.
> 
> The same deadline semantics are applied to **`wait_for_server_ready`** on the async GRPO and async distillation `VLLMClient` copies (bounded probes and sleeps, monotonic clock).
> 
> **Tests** add hermetic `TestCheckServerHealthProbe` cases (stalled socket, persistent 503, healthy/retry paths, no probe after deadline, invalid interval) and parametrized async-client readiness tests—no live vLLM or GPU required.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53dcd3ae262b23d17611becf55b2c1f45ab3d958. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/generation/vllm_client.py
```diff
@@ -270,7 +270,9 @@ def check_server(self, total_timeout: float = 0.0, retry_interval: float = 2.0):
 
         while True:
             try:
-                response = requests.get(url)
+                # The probe must be bounded: a server that accepts the connection and then stalls would otherwise
+                # block here forever, and `total_timeout` is only checked in the except branch below.
+                response = requests.get(url, timeout=retry_interval)
```
Jokasa7: [P1] This still never applies `total_timeout` when the server responds normally with a non-200 status. The elapsed-time check remains inside `except RequestException`, so a vLLM instance that repeatedly returns 503 loops and sleeps forever even though it is not up. On this exact head, a deterministic 503 response with `total_timeout=0.10` and `retry_interval=0.02` was still running after 0.375 seconds and returned only after the response changed to 200. Please check the remaining deadline after every unsuccessful probe (and avoid a final request/sleep extending past it), then add a 503 regression alongside the stalled-socket case.

Jokasa7 (commented):
Thanks for adding a bounded probe and a hermetic stalled-connection test. I checked the exact method from this head against the other unhealthy-server outcome that `requests` can return normally: an HTTP response whose status is not 200. That path still bypasses the only elapsed-time check, so the documented total deadline remains ineffective for a server that is reachable but reports that it is not ready.

Using the exact `check_server()` body with a deterministic `requests.get` stub that returned 503, `total_timeout=0.10`, and `retry_interval=0.02`, the call was still running after 0.375 seconds. It returned only after the stub was changed to 200. Applying the deadline to every unsuccessful attempt and adding a 503 regression would make this PR cover the full health-probe contract it describes.

trl/generation/vllm_client.py
behroozazarkhalili: Confirmed, and the scope is wider than 503. Fixed in fe8d46eb.

Reproducing your case against a local server on the previous head, `total_timeout=0.10` and `retry_interval=0.02` was still running after 5.0s. A server answering 404 hangs identically, so the defect is any non-200 status rather than 503 specifically: the `else` branch fell through to the retry sleep without ever consulting the deadline. The check now sits where the raising and non-raising branches converge, so both are bounded.

Your parenthetical about the final request and sleep extending past the deadline was also correct, and it was the larger error. A single probe was bounded only by `retry_interval`, so `total_timeout=0.1` with `retry_interval=5.0` against a stalled socket raised at 5.008s, overshooting by 4.908s. Both the request and the sleep are now clamped to the remaining budget. The clamp is conditional rather than a bare `min`, because `total_timeout` defaults to `0.0`: an unconditional clamp would pass `timeout=0.0` on the default path and break every healthy connection. A non-positive remainder therefore still gets one full attempt, which is what `0.0` has always meant, and the docstring now states that instead of implying a hard zero-second bound.

Two more changes came out of checking the guards rather than the behaviour:

- The tests did not actually cover either clamp. Deleting the sleep clamp or the request clamp left all three passing, because the `< 5.0` bound was too loose to notice. The bound is now derived from the class constants (`RETRY_INTERVAL / 2` with `TOTAL_TIMEOUT = 0.1`, `RETRY_INTERVAL = 5.0`), and three mutants are each killed by the test that targets them: dropping the sleep clamp, dropping the request clamp, and moving the deadline check back inside `except`.
- Deadline arithmetic moved from `time.time()` to `time.monotonic()`, so a system clock adjustment cannot shift the deadline.

Two related problems are real but deliberately not in this PR:

- `total_timeout=float("nan")` never satisfies `elapsed >= total_timeout` and loops forever, and `retry_interval <= 0` raises `ValueError` rather than `ConnectionError`. Both reproduce. `AGENTS.md` asks contributors not to add defensive code or handle cases that do not exist today, and no caller passes either value, so validating them here would work against that guidance. Happy to add it if a maintainer prefers.
- A `requests` timeout bounds inter-byte inactivity, not total duration, so a response delivered slowly enough can outlast the deadline regardless of the clamps. Closing that needs a different transport-level mechanism and is a larger change than this bugfix.

Regression coverage: `test_unavailable_server_raises_rather_than_looping` for the 503 path, alongside the existing stalled-socket and healthy-server cases. Local run is 8 passed, 35 skipped.

qgallouedec:
Have you actually hit this in a real training run ?

qgallouedec:
does vllm, in some setting accepts the connection and then never replies?

behroozazarkhalili:
No. #6973 came from reading `vllm_client.py`, not from a run, and the stall in the PR was reproduced against a synthetic socket that accepts and never answers. I have not seen vLLM do that on `/health`. The reports I can find show the reverse shape: in vllm-project/vllm#39863 and #49237 the API server keeps returning 200 on `/health` while the engine core is dead, so `check_server` passes and the hang lands on `/v1/completions`, which this PR does not touch. So the change here is a contract fix for the documented `connection_timeout` rather than a fix for an observed failure, and the observed vLLM failures point at the data-plane sites instead. If you would rather not carry a guard for a case nobody has hit, I am fine closing this and keeping #6973 focused on the completions path.

behroozazarkhalili:

Issue #1085, opened 2023-12-12
DPO trainer question


hengjiUSTC:
If we are using online dpo dataset like: https://huggingface.co/datasets/Anthropic/hh-rlhf for DPOTrainer. And using a random llama2 chat model from huggingface, which means the sft data used is different from dpo dataset. 

According to original dpo article:
> In practice, existing preference datasets are used to avoid generating new samples and collecting human preferences. When specific preference data isn't available, the reference model is initialized by maximizing the likelihood of preferred completions. This helps reduce the distribution shift between the true reference distribution and the one used in DPO. 

In this case, do we need to do SFT one more time on choose data prompts before doing DPO?  And practically, does it necessary?

lvwerra:
Maybe @lewtun or @edbeeching can help who worked quite a bit on different DPO setups.

yibomiao:
same question

Issue #1082, opened 2023-12-11
if ref model not set，value model is peft model，output always same？


WusamX:
<img width="1035" alt="image" src="https://github.com/huggingface/trl/assets/20549190/d988f076-a5c8-4ec5-81aa-d878b39d2ed3">

lvwerra:
I don't understand what the issue is, can you elaborate?

PR #6979, opened 2026-08-30
Pin the invariant suite's FA2 hub kernel to the v2 branch

qgallouedec:
The nightly invariant job has been red since 2026-08-27: both sft_fa2 members crash in the FA2 varlen backward. Root cause is upstream, not TRL: transformers 5.16.1 resolves `kernels-community/flash-attn2` to the kernel repo's v3 branch, whose stable-ABI builds fail in the backward whenever num_heads != num_heads_kv (huggingface/kernels-community#1085). Qwen2.5-0.5B is GQA (14 q / 2 kv heads), so the suite hits it every run. Fails on every torch tested (2.10-2.13), MHA is unaffected.

This pins the test's `attn_implementation` to `kernels-community/flash-attn2@v2`, the maintained cxx11 line (v1 is frozen since 2026-06-24 and has no torch 2.13 builds). The reference had been recorded on v1 (transformers 5.12.1); it is re-recorded here on v2, on HF Jobs h200 with torch 2.11.0+cu128 and transformers 5.16.1, the same torch/CUDA line as the nightly runner. Both members pass against the new reference. Known-good sha if v2 ever moves: db6b5174.

Drop the pin once https://github.com/huggingface/kernels-community/issues/1085 is fixed.

See https://huggingface.slack.com/archives/C06MG47BTLG/p1787980538918559

---

> [!NOTE]
> **Low Risk**
> Test-only invariant harness and reference JSON; no training or library runtime behavior changes for users.
> 
> **Overview**
> **Pins the invariant suite’s Flash Attention 2 hub kernel to `kernels-community/flash-attn2@v2`** so `sft_fa2` / `sft_fa2_padfree` avoid the default v3 kernel that crashes in the varlen backward on GQA models (e.g. Qwen2.5-0.5B). A shared `FA2_KERNEL` constant replaces the unpinned string in both members, with inline notes on when to drop the pin.
> 
> **Re-records `references/sft_fa2.json`** for that pin and a newer stack (torch 2.11, transformers 5.16.1, H200), updating per-step loss and grad_norm baselines.
> 
> **Updates `tests/invariant/README.md`** to note that most references are H100 while `sft_fa2` was captured on H200, with hardware recorded in each snapshot’s `env` block.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02d87c85f0c7a1742039c1beb02476f378fedf1d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lewtun (approved):
Good fix, LGTM!

albertvillanova (changes requested):
Thanks for the fast diagnosis and for filing the upstream issue with a reproducer. I verified the root cause chain end to end and it holds: transformers `6724a56b81` (#47863) introduced `_FLASH_ATTN_KERNEL_VERSION_MAPPING = {2: 3}`, shipped in 5.16.x, and the nightly logs bracket it exactly. 08-26 ran transformers 5.15.1 and passed 10/10; 08-27, 08-28, 08-29 and 08-31 ran 5.16.1 and fail 2/10 in `varlen_bwd`, from build `torch-stable-abi210-cu128-x86_64-linux` in snapshot `a394b68d`, which is the current head of the kernel repo's `v3` branch. Also confirmed: `db6b5174` in the description is indeed the current `v2` head.

Test-only, minimal, with a documented removal condition. I would like two things changed before it lands.

### 1. The reference was recorded on `v1`, not `v2`

The PR body and the commit message both say "The v2 builds are what the `sft_fa2` reference was recorded against". I do not think that is right, from two independent directions:

- **Code path.** Before the bump, `load_and_register_attn_kernel` did `version = 0 if is_fa4 else 1`, and kernel versions are branches named `v<N>` (danieldk in huggingface/kernels-community#1082). `references/sft_fa2.json` records `transformers 5.12.1`, well before 5.16, so the bare id resolved to branch **v1**. The last green nightly, 08-26 on transformers 5.15.1, also ran v1.
- **Build inventory.** That same snapshot records torch **2.10.0+cu128**, and `v2` ships no torch 2.10 build at all. The recording run could not have used v2.

This is not a claim that the pin is wrong. With the real build matrices, `v2` is defensible as the forward-looking choice, since `v1` is frozen at 2026-06-24 and has no torch 2.13 builds while `v2` is the maintained cxx11 line:

| branch | torch | cu128 variants |
|---|---|---|
| v1 | 2.10, 2.11, 2.12 | torch210, torch211 |
| v2 | 2.11, 2.12, 2.13 | torch211 only |
| v3 | stable-ABI, any torch >= 2.9 | stable-abi210 |

So either pin `@v1`, which is literally the pre-breakage state, or keep `@v2` and describe it as a deliberate move to the maintained branch, noting that the reference was recorded on v1 and that the class absorbs the change. Worth being explicit about the latter, because this class's tolerances (`loss 1.5e-2`, `grad_norm 5.0`) are sized to FA2 plus bf16 noise, roughly 50x looser on grad_norm than the eager `sft` class, so passing across a kernel-line change is spending headroom that exists for a different purpose.

### 2. Please record what makes the pin work

Because a revision is given, transformers sets `version=None` and skips its compatibility resolution entirely, so the pin is that branch's build matrix or nothing. `v2` ships cu128 for **torch 2.11 only**. The invariant runner is cu128 on torch 2.11, so it works today, but the runner installs unpinned latest (5.15.1 to 5.16.1 between two consecutive runs) and its config is not in this repo. The next torch bump there re-breaks the suite with a load-time "no build variant" error rather than a backward crash. One line next to `FA2_KERNEL` naming that constraint, plus a pointer to #1082 alongside #1085, would save the next person the whole re-derivation.

### Minor

- `test_invariant` compares only `steps`; the `config` and `env` blocks in the snapshot are provenance and are never read. So the `references/sft_fa2.json` edit changes no behavior, and it makes the snapshot describe an invocation that never ran. Fine to keep for readability, but the PR body should state that the reference values were not re-recorded.
- Optional, low risk either way: the suite pins `MODEL_REVISION` to a sha, and `@rev` accepts a sha just as well, so `@db6b5174` would make the kernel as pinned as the model. `v2` has not moved since 2026-07-09, so this is a nice-to-have.
- The description says verification was on H100, while the nightly runs on H200. The class was green on H200 through 08-26 so I expect no issue, just confirming that was intentional.

### Follow-ups, not for this PR

- The pin makes TRL's own flash-attention detection miss. `sft_trainer.py:1171` does an exact-match `model.config._attn_implementation in FLASH_ATTENTION_VARIANTS`, and transformers keeps the `@v2` suffix in `_attn_implementation`, so `sft_fa2_padfree` will now log "the attention implementation is not set to a supported Flash Attention variant". It is warning-only, both call sites are `logger.warning` and the collator keys off `self.padding_free`, so the trajectory comparison is unaffected. Still worth fixing, since it puts a false warning in this suite's own logs and hits any user who pins a kernel revision. Matching on the repo part before `@` would do it. (`FLASH_ATTENTION_VARIANTS` is also currently unused in `dpo_trainer.py`.)
- `v2` has since gained torch 2.13 CUDA builds, so the xfail reasons in `tests/test_sft_trainer.py:1101` and `tests/test_dpo_trainer.py:1095` ("no build variant for torch 2.13") are stale, and those tests may be unblockable with the same pin. The catch is the CUDA line: v2's torch 2.13 builds are cu126/cu130/cu132, with no cu128.

qgallouedec:
Thanks.

> The reference was recorded on `v1`, not `v2`

Correct (transformers 5.12.1, torch 2.10, and v2 has no torch 2.10 build). Fixed the claim in the description. Rather than spend the class headroom on a kernel-line change, I re-recorded the reference on v2 in 3b94152f: against the old v1 snapshot the v2 run was at max |dloss| 7.2e-3 (tol 1.5e-2) and max |dgrad_norm| 2.0 (tol 5.0), so half the budget. Recorded on HF Jobs. Both members pass against the new reference.

> Please record what makes the pin work

Added in 7260c5dd: pinning skips build resolution, v2 ships cu128 for torch 2.11 only, link to huggingface/kernels-community#1082.

> the `references/sft_fa2.json` edit changes no behavior

Now a real run, not an edited one.

> `@db6b5174` would make the kernel as pinned as the model

Keeping the branch, for two reasons. The suite runs on unpinned latest torch/transformers on purpose, to catch upstream drift like this one; freezing the kernel at a sha would hide the next v2 build change instead of surfacing it. And a sha freezes the build matrix: the next runner torch bump would fail at load time with "no build variant", the branch gets the new build appended.

> verification was on H100, while the nightly runs on H200

Was intentional (only box I had at hand). Moot now, the new reference is from H200.

> The pin makes TRL's own flash-attention detection miss

Agreed, but separate PR: strip the `@rev` suffix before the FLASH_ATTENTION_VARIANTS check, and drop the unused set in `dpo_trainer.py`.

> the xfail reasons in `tests/test_sft_trainer.py:1101` and `tests/test_dpo_trainer.py:1095` are stale

Agreed, separate PR. v2's torch 2.13 builds are cu126/cu130/cu132 only, so they may stay xfailed on cu128 with a corrected reason.

Two more things:
- huggingface/kernels-community#1125 merged on 08-31 and v3 was rebuilt, but the stable-abi cu128 builds are still 48628c8, so the nightly binary is still the broken one. Pin stays until cu128 is rebuilt and the suite is green on it.
- `python tests/invariant/test_invariant.py` has been broken since 0884c691 (relative import), it has to be `python -m tests.invariant.test_invariant`. README fix in a separate PR.

lewtun:
The new changes also LGTM, thanks for the thorough review @albertvillanova !

albertvillanova (approved):
Verified rather than taking the numbers on trust: the new snapshot differs from the old one only in `attn_implementation`, the env is the nightly's own torch/CUDA line, and recomputing old against new gives max |Δloss| 7.15e-3 and max |Δgrad_norm| 2.02, matching what you report. Worth noting the mean Δloss was 5.1e-4 against a `residual_tol` of 1e-3, so the drift was systematic and re-recording was the right call over spending the budget. The build-matrix comment is accurate: v2 ships cu128 for torch 2.11 only.

You are right about keeping the branch over a sha, a sha would suppress the drift signal the suite exists to produce.

Two nits, neither blocking:
- The new reference is H200, while `sft.json`, `dpo.json` and the README say H100 80GB (and `references/env.lock` does not exist).
- "Drop once #1085 is fixed" understates it, the cu128 builds also have to be rebuilt, still `48628c8` today.

LGTM.

qgallouedec:
both addressed @albertvillanova

albertvillanova:
I confirm that TRL invariant tests are green again after the merge of this PR.

PR #6985, opened 2026-08-31
Match hub kernels by repo id when checking the attention implementation

### LINKED ISSUE #6984: Padding-free warning fires for revision-pinned hub kernels (`repo_id@revision`)
### Describe the bug

`SFTTrainer` decides whether the attention implementation supports padding-free training with an exact-match membership test against a hardcoded set of bare repo ids:

```python
# trl/trainer/sft_trainer.py
use_flash_attention = model.config._attn_implementation in FLASH_ATTENTION_VARIANTS
```

Transformers, however, accepts a revision suffix on hub kernels, `repo_id@revision` (and a kernel name, `repo_id:kernel_name`), and keeps the full string in `config._attn_implementation`. So `kernels-community/flash-attn2@v2` misses the set and the trainer warns:

> Padding-free training is enabled, but the attention implementation is not set to a supported Flash Attention variant. [...]

The same applies to the packing check a few lines below.

This matters right now because pinning a kernel revision is the current workaround for https://github.com/huggingface/kernels-community/issues/1085, where the v3 stable-ABI builds (the default resolution since transformers 5.16.1) crash in the backward pass for GQA models. The revision syntax is what the `kernels` maintainer recommends (https://github.com/huggingface/kernels-community/issues/1082#issuecomment), and TRL itself adopts it for the invariant suite in #6979. So TRL currently warns users against its own documented workaround.

### Scope

Warning only. Both uses are `logger.warning`, and the collator's padding-free behavior is driven by `self.padding_free`, so no training behavior changes. It is misleading output, not a silent correctness problem.

### Reproduction

```python

SFTTrainer(
    args=SFTConfig(
        padding_free=True,
        max_length=None,
        model_init_kwargs={"attn_implementation": "kernels-community/flash-attn2@v2", "dtype": "bfloat16"},
    ),
    train_dataset=...,
)
```

### Expected behavior

A revision-pinned or kernel-name-suffixed hub kernel is recognized as the same implementation as its bare repo id, and no warning is emitted.

### Fix

Compare against the repo id alone, by stripping the `@revision` and `:kernel_name` suffixes before the lookup.

### Notes

- `FLASH_ATTENTION_VARIANTS` is also defined in `trl/trainer/dpo_trainer.py` but is unused there, so DPO has nothing to fix. Worth removing separately.
- Independently of this, the underlying upstream breakage (fine-tuning any GQA model through the flash-attn2 hub kernel, including the silent `flash_attention_2` fallback when the `flash-attn` package is not installed) deserves a documentation note. Tracked separately.

albertvillanova:
This PR stops `SFTTrainer` from warning that a revision-pinned hub attention kernel is an unsupported attention implementation.

Fix #6984.

### Motivation

Transformers accepts a revision and a kernel name on a hub kernel spec (`repo_id@revision:kernel_name`) and keeps the full string in `config._attn_implementation`, while `FLASH_ATTENTION_VARIANTS` holds bare repo ids. A pinned kernel such as `kernels-community/flash-attn2@v2` therefore missed the lookup, and both the padding-free and the BFD packing checks warned that the attention implementation was not supported.

Pinning a revision is the current workaround for the flash-attn2 v3 stable-ABI backward crash on GQA models (huggingface/kernels-community#1085), it is the syntax recommended by the `kernels` maintainer (huggingface/kernels-community#1082), and it is what the invariant suite adopts in #6979. TRL was therefore warning users against its own workaround.

The impact is misleading output only: padding-free behavior is driven by a separate attribute, so no training behavior changes.

### Solution

Strip the revision and kernel name suffixes before the lookup, so a hub kernel is matched by its repo id.

### Changes

- Match the attention implementation against `FLASH_ATTENTION_VARIANTS` by repo id, so revision-pinned and kernel-name-suffixed hub kernels are recognized.
- Add a regression test covering the bare repo id, the revision-pinned form, and `eager` as a positive control, so the assertion cannot pass vacuously if the warning text changes.
- Leave `dpo_trainer.py` untouched: its copy of `FLASH_ATTENTION_VARIANTS` is unused, so removing that dead constant belongs in a separate PR.

---

> [!NOTE]
> **Low Risk**
> Init-time warning logic only; no changes to data collation or training paths.
> 
> **Overview**
> **SFTTrainer** no longer treats revision-pinned or kernel-named hub specs (e.g. `kernels-community/flash-attn2@v2`) as unsupported Flash Attention when **padding-free** is on. It normalizes `config._attn_implementation` to the repo id (strip `@revision` and `:kernel_name`) before checking membership in `FLASH_ATTENTION_VARIANTS`, so users stop getting the misleading “not set to a supported Flash Attention variant” warning while still using a supported kernel.
> 
> This is **warning-only** at trainer init; collator and training behavior are unchanged.
> 
> A new parametrized test asserts the padding-free Flash Attention warning is **absent** for bare and `@v2` flash-attn2 hub ids and **present** for `eager`, without loading any kernel.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 843fa5eddc76b0de8276380ee7215773b82d3f43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -1073,6 +1073,26 @@ class TestSFTTrainer(TrlTestCase):
                 model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5", args=training_args, train_dataset=dataset
             )
 
+    @pytest.mark.parametrize(
+        "attn_implementation, expect_warning",
+        [
+            ("kernels-community/flash-attn2", False),
+            ("kernels-community/flash-attn2@v2", False),
+            ("eager", True),
+        ],
+    )
+    def test_padding_free_warns_only_for_unsupported_attention(self, attn_implementation, expect_warning, caplog):
+        dataset = load_dataset("trl-internal-testing/zen", "standard_language_modeling", split="train[:2]")
+        model = AutoModelForCausalLM.from_pretrained("trl-internal-testing/tiny-Qwen2ForCausalLM-2.5")
+        # Only the config value matters here: the warning is emitted at init, so no attention kernel is ever loaded.
+        model.config._attn_implementation = attn_implementation
+        training_args = SFTConfig(output_dir=self.tmp_dir, padding_free=True, max_length=None, report_to="none")
+
+        with caplog.at_level("WARNING", logger="trl.trainer.sft_trainer"):
+            SFTTrainer(model=model, args=training_args, train_dataset=dataset)
+
+        assert ("supported Flash Attention variant" in caplog.text) == expect_warning
+
     def test_train_with_iterable_dataset(self):
         dataset = load_dataset("trl-internal-testing/zen", "standard_language_modeling", split="train", streaming=True)
 
--- a/trl/trainer/sft_trainer.py
+++ b/trl/trainer/sft_trainer.py
@@ -1168,7 +1168,10 @@ class SFTTrainer(_BaseTrainer):
         # BFD packing requires padding-free mode; otherwise, the collator outputs padded attention masks, causing
         # FlashAttention to ignore position_ids and recompute them incorrectly from the padded attention mask.
         self.padding_free = args.padding_free or (args.packing and args.packing_strategy in {"bfd", "bfd_split"})
-        use_flash_attention = model.config._attn_implementation in FLASH_ATTENTION_VARIANTS
+        # A hub kernel can be requested with a revision and/or a kernel name (`repo_id@revision:kernel_name`), while
+        # the variants above are bare repo ids, so compare against the repo id alone.
+        attn_implementation = model.config._attn_implementation.split("@")[0].split(":")[0]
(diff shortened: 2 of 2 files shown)


albertvillanova:
Merging this PR as a follow-up PR to: 
- https://github.com/huggingface/trl/pull/6979#pullrequestreview-5065126641

> The pin makes TRL's own flash-attention detection miss. `sft_trainer.py:1171` does an exact-match `model.config._attn_implementation in FLASH_ATTENTION_VARIANTS`, and transformers keeps the `@v2` suffix in `_attn_implementation`, so `sft_fa2_padfree` will now log "the attention implementation is not set to a supported Flash Attention variant". It is warning-only, both call sites are `logger.warning` and the collator keys off `self.padding_free`, so the trajectory comparison is unaffected. Still worth fixing, since it puts a false warning in this suite's own logs and hits any user who pins a kernel revision. Matching on the repo part before `@` would do it.

PR #6986, opened 2026-08-31
Document the FlashAttention 2 backward failure on GQA models

albertvillanova:
This PR documents the FlashAttention 2 Hub kernel failure on grouped-query attention models, and the revision pin that works around it.

### Motivation

Since Transformers v5.16, `kernels-community/flash-attn2` resolves to the kernel repository's `v3` branch, whose Torch stable ABI builds fail in the backward pass when `num_attention_heads != num_key_value_heads`, so fine-tuning any grouped-query attention model through this kernel crashes at the first optimizer step (huggingface/kernels-community#1085).

Two things make this hard on users. The error surfaces as an opaque `aten::sum` stable ABI dispatcher failure with no mention of attention, the kernel, or the model configuration. And it is not limited to users who request the kernel explicitly: `attn_implementation="flash_attention_2"` falls back to the same Hub kernel when the `flash-attn` package is not installed.

Several TRL pages recommend this kernel, including the padding-free section and the GKD page, which points Gemma users to it, so the docs currently route users into the failure with no warning.

### Solution

Describe the issue and the `@v2` pin once in the Kernels Hub guide, and point to it from the pages that recommend the kernel. The section is self-contained so it can be removed in one place once the upstream issue is fixed.

### Changes

- Add a section to the Kernels Hub guide describing the symptom, the affected models, the silent `flash_attention_2` fallback, and the `kernels-community/flash-attn2@v2` pin, with a link to the upstream issue.
- Point to that section from the padding-free recommendation in the memory usage guide.
- Point to that section from the Gemma recommendation on the GKD page, and from the attention kernel recommendation in the speeding up training guide.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime or training logic modified.
> 
> **Overview**
> **Documents a training-breaking regression** when using the Hub `kernels-community/flash-attn2` kernel (and the `flash_attention_2` fallback without a local `flash-attn` install) on **grouped-query attention** models after Transformers v5.16.
> 
> Adds a dedicated **“FlashAttention 2 backward pass on GQA models”** section in `kernels_hub.md` that explains the v3 backward failure (`aten::sum` at the first optimizer step), who is affected, and the workaround **`kernels-community/flash-attn2@v2`**, with a link to [kernels-community#1085](https://github.com/huggingface/kernels-community/issues/1085).
> 
> **Cross-links** that section from other guides that already recommend the kernel: Gemma/GKD usage tips, padding-free + FlashAttention in the memory guide, and Hub kernels in the speeding-up-training guide—so users hit the warning before following those examples.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3a86316d641366bb3a582948826893232a5e2623. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/gkd_trainer.md
+++ b/docs/source/gkd_trainer.md
@@ -29,6 +29,7 @@ The authors find that on-policy data (high `lmbda`) performs better and the opti
 
 > [!WARNING]
 > Make sure that `attn_implementation="kernels-community/flash-attn2"` when training [Gemma models](https://huggingface.co/models?other=gemma2). Otherwise you will encounter NaNs in the logits due to the [soft capping technique](https://huggingface.co/blog/gemma2#soft-capping-and-attention-implementations) adopted by this architecture.
+> Gemma models use grouped-query attention, so this kernel currently needs to be pinned; see [FlashAttention 2 backward pass on GQA models](kernels_hub#flashattention-2-backward-pass-on-gqa-models).
 
 The basic API is as follows:
 
--- a/docs/source/kernels_hub.md
+++ b/docs/source/kernels_hub.md
@@ -46,6 +46,24 @@ trl sft ... --attn_implementation kernels-community/flash-attn2
 > [!TIP]
 > Now you can leverage faster attention backends with a pre-optimized kernel for your hardware configuration from the Hub, speeding up both development and training.
 
+## FlashAttention 2 backward pass on GQA models
+
+> [!WARNING]
+> Since Transformers v5.16, `kernels-community/flash-attn2` resolves to the kernel repository's `v3` branch, whose Torch stable ABI builds fail in the backward pass when `num_attention_heads != num_key_value_heads`, that is, on any model using grouped-query attention (most recent models). The forward pass is unaffected, so inference works and training fails at the first optimizer step with `RuntimeError: torch_call_dispatcher( "aten::sum", ... ) API call failed`.
+
+This also affects `attn_implementation="flash_attention_2"` when the `flash-attn` package is not installed, because Transformers then falls back to the same Hub kernel.
+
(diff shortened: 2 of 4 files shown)


docs/source/gkd_trainer.md
AmineDiro, line 1: Is this specific to the gkd trainer ?

AmineDiro (approved):

docs/source/gkd_trainer.md
albertvillanova: No, it is not specific to GKD: the failure hits any grouped-query attention model. The general note lives in the Kernels Hub guide, and this line was only meant as a pointer to it, added because this page actively recommends the FlashAttention 2 kernel for Gemma, which is GQA.

Fair question though, because the scope has narrowed upstream since I opened this PR. The fix (huggingface/kernels-community#1125) is merged and rebuilt for cu126, cu130 and cu132, and I verified it on cu130 at both the kernel and the transformers level. Only the cu128 builds still ship the pre-fix binary. So the note as written is now too broad, and I am reworking the PR: the per-page pointers are the perishable part and will likely go away.

albertvillanova:
Superseded by #6998.

@AmineDiro your question is what prompted this: the note was not GKD-specific, and pinning the per-page pointers to one incident was the wrong shape. Meanwhile the upstream fix (huggingface/kernels-community#1125) landed and was rebuilt for cu126, cu130 and cu132, so this PR's text became inaccurate as well as short-lived.

#6998 keeps the durable part, namely how a kernel version is resolved and how to pin one, in a single section, and reduces the regression itself to one scoped warning about CUDA 12.8 that can be removed on its own once huggingface/kernels-community#1085 closes. Opening it fresh rather than rewriting this one, so the title and diff match what gets reviewed.

Issue #6989, opened 2026-09-01
Docker image version tags are mutable: huggingface/trl:1.9.2 was rebuilt on 65 pushes after its release


albertvillanova:
### Problem

`docker-build.yml` builds and pushes `huggingface/trl:${VERSION}` on **every push to main**, where `VERSION` is whatever PyPI currently reports as the latest release. A version tag is therefore rewritten on every merge until the next release, and does not identify what is inside it.

`huggingface/trl:1.9.2` was rebuilt on all 65 pushes to main between its release and the 1.10.0 release:

- 1.9.2 released to PyPI: `2026-07-28T10:27:43`
- `huggingface/trl:1.9.2` last written: `2026-08-13T01:26:23`, digest `sha256:d57701ac4f62da0dca4890407a755e8757b4fadd6eab9ccd99b4706d4df60843`
- 1.10.0 released to PyPI: `2026-08-13T01:26:35`, twelve seconds later

So the image published as 1.9.2 is the one built sixteen days after 1.9.2 shipped. `huggingface/trl:1.10.0` has been overwritten on every push to main since. The workflow sets `cancel-in-progress: false`, so queued runs are not skipped, they all execute.

Two things compound this:

1. The Dockerfile installs `trl[liger,peft,vlm]` unpinned, so the tag is a claim about the version, not a constraint on it. The transitive dependency set is whatever PyPI resolved at build time.
2. Builds rely on the runner group's warm layer cache, which is not declared anywhere. When it is warm the tag content is frozen; when it is evicted the resolution changes. The digest above moved at least once mid-window, so the content genuinely varies.

### Why it matters

Anyone pinning a version tag expects it to be immutable. `trl-jobs` hardcodes `image="huggingface/trl"` in `trl_jobs/cli.py`, so this is consumed in code and not only in the docs.

The dev image is not affected: `huggingface/trl:dev` tracks main and is meant to be mutable. See #6933 for that discussion.

### Proposal

1. Build the released image on `release: published` instead of on every push to main, keeping the dev image on push to main.
2. Take the version from the release tag instead of querying the PyPI API.
3. Pin the installed version to the tag being built.

GitHub releases are published after the PyPI upload (v1.10.0 at `01:36:04` versus PyPI at `01:26:35`), so a release trigger does not race the index.

qgallouedec:
Fixed by #6990 #6991 and #6992

PR #6990, opened 2026-09-01
Split the released and dev Docker image builds into separate workflows

albertvillanova:
This PR splits the Docker image builds into two workflows, so that the released image is built once per release instead of on every push to main, while the dev image keeps building on every push to main.

Part of #6989.

### Motivation

Both images were built from a single workflow triggered on every push to main. That is correct for `huggingface/trl:dev`, which tracks main and is meant to be mutable, but wrong for the released image: it rewrites `huggingface/trl:${VERSION}` on every merge until the next release, so a version tag does not identify what is inside it. #6989 has the details, including `huggingface/trl:1.9.2` being rebuilt on all 65 pushes to main between its release and 1.10.0.

The two images now have genuinely different release cadences, so a single trigger cannot serve both. Splitting the workflows lets each declare its own, which is also how the test workflows are already organized in this repository: one file per trigger, from `tests.yml` to `tests_latest.yml` and `tests_python_versions.yml`.

GitHub releases are published after the PyPI upload (v1.10.0 at `01:36:04` versus PyPI at `01:26:35`), so the release trigger does not race the index.

Removing the dev image was considered and rejected in #6933: it has consumers, and per-push builds are the behavior its users expect. This PR leaves it untouched.

### Changes

- Trigger `docker-build.yml` on `release: published` instead of on push to main, and keep only the released image job
- Add `docker-build-dev.yml` with the dev image job, triggered on push to main as before
- Give the dev workflow its own concurrency group, so a release build does not queue behind an in-flight dev build

### Note

The `release: published` trigger cannot be exercised before merging, since `workflow_dispatch` runs the job but not the trigger. I will dispatch the workflow manually once this is merged to confirm the job still builds and pushes, and the trigger itself is confirmed by the next release.

---

> [!NOTE]
> **Low Risk**
> CI-only workflow split with no application or auth changes; main operational risk is release Docker builds not running until the next published release unless manually dispatched.
> 
> **Overview**
> **Splits TRL Docker CI so release tags are built on publish, while `huggingface/trl:dev` still tracks `main`.**
> 
> Adds **`.github/workflows/docker-build-dev.yml`**, which runs on every push to `main` (and `workflow_dispatch`) and only builds/pushes `huggingface/trl:dev`, with its own concurrency group `docker-image-build-dev`.
> 
> Updates **`docker-build.yml`** to drop the `trl-dev` job and change the trigger from `push` to `main` to **`release: published`** (still allows manual dispatch). The remaining job keeps building the versioned image from PyPI plus the `huggingface/trl` tag, so versioned Docker images are no longer rebuilt on every merge to `main`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ca448cc53e85c4cadf8628e59fa7c38509fc60b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/docker-build-dev.yml
+++ b/.github/workflows/docker-build-dev.yml
@@ -1,38 +1,26 @@
-name: Build TRL Docker image
+name: Build TRL dev Docker image
 
 on:
-  release:
-    types: [published]
+  push:
+    branches:
+      - main
   workflow_dispatch:
 
 concurrency:
-  group: docker-image-builds
+  group: docker-image-build-dev
   cancel-in-progress: false
 
 env:
   CI_SLACK_CHANNEL: ${{ secrets.CI_PUSH_MAIN_CHANNEL }}
 
 jobs:
-  trl:
-    name: "Build and push TRL Docker image"
+  trl-dev:
+    name: "Build and push TRL Dev Docker image"
     runs-on:
       group: aws-general-8-plus
     steps:
       - name: Checkout code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
-
-      - name: Get TRL version from PyPI
-        run: |
-          VERSION=$(curl -s https://pypi.org/pypi/trl/json | jq -r .info.version)
-          if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+([a-zA-Z0-9._-]*)?$ ]]; then
-            echo "Invalid version format: $VERSION"
-            exit 1
-          fi
-          if [[ ${#VERSION} -gt 50 ]]; then
-            echo "Version string too long: $VERSION"
-            exit 1
-          fi
-          echo "VERSION=$VERSION" >> $GITHUB_ENV
 
       - name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


albertvillanova:
Verified before merge, so the release trigger is not first exercised on a real release.

**The `release: published` trigger fires and runs this workflow.** Tested on a throwaway release in a fork, with the stack applied: https://github.com/albertvillanova/trl/actions/runs/33474463321

```
event=release tag_name=v0.0.0-verify VERSION=0.0.0-verify
```

The release and its tag have been deleted.

That run also settles which ref a release event resolves the workflow from. The fork's default branch still had the old push triggered workflow, with no `release:` trigger at all, so nothing would have fired if release events used the default branch. The run is on `headBranch=v0.0.0-verify`, so **the workflow is resolved at the tagged commit**. This stack therefore has to be merged before the next release tag is cut, which it will be.

**The split workflow still builds on the real runner.** Separately verified on `aws-general-8-plus` with pushing disabled: https://github.com/huggingface/trl/actions/runs/33473658476

Nothing was pushed to Docker Hub by any of this: `huggingface/trl` tags are unchanged.

qgallouedec (approved):
Nice

PR #7003, opened 2026-09-02
Rewrite the long context guide

qgallouedec:
> [!NOTE]
> **Low Risk**
> Documentation and marketing copy only; no library or training behavior changes.
> 
> **Overview**
> **Reframes long-context training docs** around a runnable million-token example instead of a parallelism reference manual.
> 
> **README** and **docs index** swap the “What’s New” blurb from **DistillationTrainer** to the updated [long context guide](https://huggingface.co/docs/trl/long_context_training).
> 
> The **`long_context_training.md`** rewrite leads with **Run it** (`accelerate launch` + `examples/sft_qwen3_8b_1m_context/`) and sample metrics, then **What just happened**—a fixed-order walkthrough of four bottlenecks: **chunked NLL loss** (default in `SFTConfig`), **YaRN** via `model_init_kwargs["rope_parameters"]`, **gradient checkpointing offload**, and **context parallelism** (`cp_size` on FSDP2 vs `sp_size` on DeepSpeed). Long standalone Ring Attention / ALST-Ulysses sections, benchmark tables, and full YAML walkthroughs are trimmed in favor of narrative, memory profiles, and short config snippets; **Further reading** links replace the old resource lists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88dcfc1ffb4d29139c87828249fa45be2326e7f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


adithya-s-k (approved):

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
+Can we compute the loss without materializing the whole logits matrix? Yes, we can proceed in chunks.
 
-1. **Accelerate 1.11.0 or higher** is required for Ring Attention / Context Parallelism support
-2. **FSDP2 (PyTorch FSDP v2)** is required as the distributed training backend
-3. **SDPA attention** - Flash Attention is currently not supported
-4. **Sequence length divisibility** - sequences must be divisible by `cp_size * 2`. This is automatically handled using the `pad_to_multiple_of` parameter in the data collator.
+### The massive logits matrix: chunk it
 
-#### Configuration
+To cut the head off this peak, we can compute the loss in chunks. Instead of computing the logits for the whole sequence at once, we take a few hundred rows at a time, compute the loss for those, and sum the pieces. This way, we never have to hold the entire logits matrix in memory at once.
```
kashif:
```suggestion
Instead of computing the logits for the whole sequence at once, we take a few hundred rows at a time, compute the loss for those, and sum the pieces. The entire logits matrix is never held in memory at once.
```

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
 
-##### Accelerate Configuration
+<img src="https://huggingface.co/datasets/trl-lib/documentation-images/resolve/main/long_context_chunked_loss.png" alt="The same computation one chunk of rows at a time, each chunk freed before the next, summing into the same loss"/>
 
-Use one of the provided accelerate config files (e.g. [`context_parallel_2gpu.yaml`](https://github.com/huggingface/trl/blob/main/examples/accelerate_configs/context_parallel_2gpu.yaml) for 2 GPUs):
+Instead of one big multiplication, we do several smaller ones. The memory saving is large enough that the loss will never be the peak again.
```
kashif:
```suggestion
One big multiplication becomes several smaller ones, and the memory saving is large enough that the loss is never the peak again.
```

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
-    per_device_train_batch_size=1,
-    ...
-)
+```yaml
+parallelism_config:
+  parallelism_config_sp_size: 4
 ```
 
-Then, launch your training script with the appropriate accelerate config file:
+The two knobs do not cross over today: `cp_size` requires FSDP2 and `sp_size` only runs under DeepSpeed. Pick the backend and the method follows. The rest of this section is the FSDP2 path, which is what the example at the top of this guide uses: those four GPUs are not independent workers on four batches, they are one group sharing a single sequence.
```
kashif:
```suggestion
The two knobs do not cross over today: `cp_size` requires FSDP2 and `sp_size` only runs under DeepSpeed. Pick the backend and the method follows. The rest of this section follows the FSDP2 path, which is what the example at the top of this guide uses: its GPUs are not independent workers on separate batches, they are one group sharing a single sequence.
```

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
+And that is the last of the four levers:
 
-### Further Reading
+<img src="https://huggingface.co/datasets/trl-lib/documentation-images/resolve/main/long_context_cp_scaling.png" alt="Peak memory against sequence length on one GPU and on four with context parallelism, the four-GPU line reaching a million tokens where the single GPU stops just past 256k"/>
 
-#### General Resources
-- [Hugging Face Blog: Understanding Ulysses and Ring Attention](https://huggingface.co/blog/exploding-gradients/ulysses-ring-attention) - Detailed comparison of Ring Attention vs Ulysses approaches
-- [Accelerate: Context Parallelism Guide](https://huggingface.co/docs/accelerate/concept_guides/context_parallelism)
-- [Hugging Face Blog: Enabling Long-Context Training with Sequence Parallelism in Axolotl](https://huggingface.co/blog/axolotl-ai-co/long-context-with-sequence-parallelism-in-axolotl)
+One card stops just past 256k. Four of them take the whole million!, and land at 63.6 GB with room to spare. The sequence this guide opened with fits.
```
kashif:
```suggestion
One card stops just past 256k. Four of them take the whole million!, landing at 63.6 GB with room to spare. The sequence this guide opened with fits.
```

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
-> [!NOTE]
-> **Technical Note on Parallelism Configuration:**
-> - **DeepSpeed ALST/Ulysses** uses `sp_size` with `sp_backend="deepspeed"` in both YAML and Python API
-> - **Ring Attention (FSDP2)** uses `cp_size` with `cp_backend="torch"`
->
-> The Trainer automatically accounts for both CP and SP when calculating effective batch sizes and training metrics.
+<img src="https://huggingface.co/datasets/trl-lib/documentation-images/resolve/main/long_context_cp.png" alt="A sequence of tokens overflowing a single GPU and running out of memory, then the same sequence cut into four slices that each fit on one GPU"/>
 
-#### Requirements and Limitations
+Give each GPU one slice of the sequence. With four of them, every GPU holds 262,144 tokens instead of 1,048,576. The four tensors shrink with it: 19 GB rather than 76, which alongside the model's 30 GB is a card that fits.
```
kashif:
```suggestion
Give each GPU one slice of the sequence. With four of them, every GPU holds 262,144 tokens instead of 1,048,576. The four tensors shrink with it: 19 GB rather than 76, which alongside the model's 30 GB fits on the card.
```

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
-    --dtype bfloat16 \
-    --attn_implementation flash_attention_2 \
-    --output_dir output-alst-4gpu \
-    --logging_steps 10 \
-    --report_to trackio
-```
+1. Sequences have to be padded to a multiple of `cp_size * 2`, so `pad_to_multiple_of=8` (in the [`SFTConfig`]) for four GPUs.
+2. The causal-SDPA requirement rules out packing, which relies on a block-diagonal mask to keep documents from reading each other, and TRL raises if you ask for both. It also rules out models whose layers use sliding-window or chunked attention, which accelerate refuses: OpenAI GPT-OSS, Gemma 3 and 4, Qwen3.5 and later. Qwen3 and Qwen3-MoE are full attention throughout, which is why they are the models here.
+
+Passing the slices around costs less than you would expect. At 131k tokens on one node, a step takes 34.6 s across two GPUs, 17.8 s across four and 9.5 s across eight. Each doubling of the group nearly halves the step, and two to eight recovers 3.7x of a possible 4x.
```
kashif:
```suggestion
Passing the slices around costs less than you might expect. At 131k tokens on one node, a step takes 34.6 s across two GPUs, 17.8 s across four and 9.5 s across eight. Each doubling of the group nearly halves the step time, and going from two to eight recovers 3.7x of a possible 4x.
```

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
+```
 
-5. **Monitor memory usage** across all GPUs to ensure balanced workload
+The same measurement, with the rescaled run on top:
 
-#### Training at 1M tokens and beyond
+<img src="https://huggingface.co/datasets/trl-lib/documentation-images/resolve/main/long_context_rope_pertoken.png" alt="The same per-token loss with a second run using rescaled positions, which stays flat where the first one climbs"/>
 
-With the setup above plus a few levers, SFT scales to million-token sequences on a single 8×H100 node. Verified configurations:
+The rescaled run stays flat all the way to 160k. Over the last 20k tokens it averages 2.8 against 7.3 without the rescaling.
```
kashif:
```suggestion
The rescaled run stays flat all the way to 160k. Over the last 20k tokens it averages a loss of 2.8, against 7.3 without the rescaling.
```

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
-> - **Ring Attention (FSDP2)**: Uses `cp_size` to split sequences. With `max_seq_length=8192` and `cp_size=4`, each GPU processes 2048 tokens.
-> - **ALST/Ulysses (DeepSpeed)**: Uses `sp_size` (with `sp_backend="deepspeed"`) to split sequences. With `max_seq_length=8192` and `sp_size=2`, each GPU processes 4096 tokens.
->
-> The Trainer automatically accounts for context parallelism when calculating batch sizes and training metrics.
+- **`num_tokens` is 1.049e6.** That is one sequence, not a batch of short ones. Every token attends to every earlier token, across the whole million.
+- **Loss starts at 4.3**, which is a normal starting loss. A run that is subtly misconfigured for long context starts around 10 instead, and we will come back to why.
+- **380 s per step.** One step is one sequence, so that is a little over six minutes per sequence.
 
-### Choosing Between Ring Attention and Ulysses
+Naively, this sequence needs 288 GB per GPU. The rest of this guide is how it gets to 56.
```
kashif:
```suggestion
Naively, this sequence needs 288 GB per GPU. The rest of this guide is the story of how that comes down to 56.
```

docs/source/long_context_training.md
AmineDiro: we can talk about about __ways__ to chunk: row/ columns or both. 
rows are straightforward because softmax is computed rowwise independtly 👍🏼 

Also there is an arithmetic intensity roofline , chunking too small means wasted MFU

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
+| FSDP2 | ✅ | ❌ |
+| DeepSpeed | ❌ | ✅ |
+| Setting | `cp_size` | `sp_size` |
+| What gets split | the tokens | the attention heads |
+| How far it scales | any number of GPUs | up to the number of KV heads, 8 here |
+| Attention | SDPA, causal only | SDPA or FlashAttention |
+| Requires | accelerate 1.11 | accelerate 1.12, DeepSpeed 0.18.1 |
 
-Use the provided accelerate config file ([`alst_ulysses_4gpu.yaml`](https://github.com/huggingface/trl/blob/main/examples/accelerate_configs/alst_ulysses_4gpu.yaml)):
+**On FSDP2, CP.** Each GPU keeps its own slice, and the other slices come to it in turn, so every token eventually sees every earlier token.
```
kashif:
```suggestion
**On FSDP2, use CP.** Each GPU keeps its own slice, and the other slices come to it in turn, so every token eventually sees every earlier token.
```

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
-  parallelism_config_sp_seq_length_is_variable: true
-  parallelism_config_sp_attn_implementation: flash_attention_2
+  parallelism_config_cp_size: 4
 ```
 
-##### Training Configuration
-
-```python
-from trl import SFTConfig
+**On DeepSpeed, SP.** Instead of moving slices around, it reshuffles the batch just before attention so each GPU holds every token but only a quarter of the attention heads, then shuffles back afterwards.
```
kashif:
```suggestion
**On DeepSpeed, use SP.** Instead of moving slices around, it reshuffles the batch just before attention so each GPU holds every token but only a quarter of the attention heads, then shuffles back afterwards.
```

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
 
-These three were measured against each other with a hand-tuned activation offload rather than the config
-above, so read them for the scaling ratio and not as absolute step times.
+The model and its optimizer account for a fixed 22 GB, whatever the sequence length. Above that, each red band is one layer's saved activation, 0.47 GB apiece. They stack up as the forward writes them, sit there, and come off one at a time as the backward consumes them. The grey on top is everything recomputed and thrown away within milliseconds.
 
-The levers, roughly in the order you will need them as model size or sequence length grows:
+TRL already reduces them for you, and it is on by default. Gradient checkpointing keeps only what goes into each layer and throws away everything the layer computes inside itself: the attention scores, the wide MLP intermediate, the norms. When the backward pass needs them it runs the layer a second time to get them back. What survives is one saved tensor per layer, `sequence x hidden` each, and at long sequence lengths that is still the largest thing on the card.
```
AmineDiro, line 126: > runs the layer a second time to get them back.

It runs all 0..layer, right ? It's a tradeoff between recompute and memory, grad checkpoint mean O(N**2) layer recomputation  if I am not mistaken

docs/source/long_context_training.md
```diff
@@ -1,368 +1,221 @@
...
-3. **Attention implementation** - Flash Attention 2 recommended (clean output), SDPA works as fallback
-4. **Sequence length divisibility** - sequences must be divisible by `sp_size`. Use `pad_to_multiple_of` in your training config.
-5. **Parallelism configuration** - You must ensure `dp_replicate_size × dp_shard_size × sp_size = num_processes`
+**The obvious objection is attention**. Every token has to attend to every earlier token, and no GPU has the earlier tokens. So the GPUs exchange what they are missing, and what comes out is exactly what one enormous GPU would have computed.
 
-#### Configuration
+There are two ways to run that exchange, context parallelism (CP) and Ulysses sequence parallelism (SP).
 
-##### Accelerate Configuration
+| | CP | SP |
```
AmineDiro, line 175: cool ! very clear table 👍🏼

docs/source/long_context_training.md
AmineDiro, line 1: The whole post is super cool . I think it would be good to also release it as a blogpost as it reads like a walkthrough to optimize long context SFT !

AmineDiro (approved):

docs/source/long_context_training.md
qgallouedec: No, only that one layer. Each decoder layer is wrapped individually (`GradientCheckpointingLayer`), so every layer saves its own input and the backward re-runs just that layer from it. One extra forward pass over the model in total, not N**2.

The tradeoff you're describing shows up if you checkpoint sparsely, saving every k-th layer and replaying from the nearest one. transformers exposes that as `every_n_layers`, check huggingface/transformers#48200 ;)

docs/source/long_context_training.md
qgallouedec: Good point on rows. Added a clause: the loss is a sum over rows and softmax is row-wise, so chunking rows is exact and needs no online correction.

But leaving column chunking out. It needs an online softmax, which is cce/liger territory and a different technique from this one.

Same for the roofline. TRL's chunk size is a fixed 256 for now (`_CHUNKED_LM_HEAD_CHUNK_SIZE`), not
something you set, so there's nothing for a reader to act on (again, for now, we can make it an arg of sft, if only there is a real value).

I'll add a sentence saying it's sized to keep the matmuls big

qgallouedec: 88dcfc1f

PR #7017, opened 2026-09-02
AsyncGRPO: PEFT/LoRA support with adapter-only vLLM sync

AmineDiro:
Adds LoRA training to `AsyncGRPOTrainer`, and a second weight-sync path that pushes **only the adapter** to vLLM (~1% of the bytes) instead of merging it into the base model and re-sending every parameter.

## What this adds

- **`peft_config` argument** on `AsyncGRPOTrainer`, mirroring `GRPOTrainer`.
- **Adapter-only weight sync.** Each sync saves the adapter and registers it with the vLLM server over `/v1/load_lora_adapter`, leaving base weights untouched. No NCCL transfer group is built at all in this mode.
- **Merged-weight sync as the fallback arm** of the same branch, so a PEFT model still trains against a server that cannot serve adapters.
- **Versioned adapter publishing** — each sync publishes `trl-policy-v{N}`; the rollout worker requests that name so generation actually uses the current policy.
- **Adapter validation** (`validate_lora_for_vllm_sync`) that rejects configurations vLLM cannot serve as a plain adapter and falls back to merged sync.
- **`round_lora_rank`** helper for vLLM's fixed `max_lora_rank` set.
- **`model_version` persisted in `rollout_state.json`**, so adapter numbering continues across a resume instead of restarting at `v1`.
- Docs section in `async_grpo_trainer.md` and a runnable example, `examples/async_grpo_lora_math/`.

## Design choices

- **No new config field.** The mode is derived: a PEFT model plus a server reporting a `lora_config`. Decided once at init on rank 0 and broadcast, since a per-rank decision would deadlock the collective save against the merged path.
- **A new adapter name per sync, never in-place.** `load_inplace=True` makes the server re-read the adapter from disk on every request ([vllm#41482](https://github.com/vllm-project/vllm/pull/41482)), and vLLM keys its prefix cache on the adapter name alone ([vllm#42125](https://github.com/vllm-project/vllm/issues/42125)) — so an in-place swap would serve the previous policy's KV blocks. A new name is a new cache namespace, so no `reset_prefix_cache()` is needed.
- **Load the adapter before bumping the version.** The rollout worker derives the requested adapter from `model_version`; bumping first would name an adapter the server does not have yet.
- **Keep `max_staleness + 1` versions servable**, matching the `--max-loras` requirement asserted at init. A trajectory that began under an older policy finishes under it, and no in-flight request can name an unloaded adapter.
- **Adapter files outlive their unload by one sync.** vLLM resolves `lora_path` lazily inside the engine, where a missing directory is a fatal `HFValidationError` (it falls back to treating the path as a Hub repo id), not a 404.
- **Absolute adapter paths.** The path is resolved by the server's process, which has its own working directory and may be on another machine.
- **Every rank runs the adapter save.** Materialising a sharded parameter is a collective; a rank-0-only call deadlocks rather than erroring.
- **Reject rather than mis-serve.** `modules_to_save`, DoRA, trained bias, `target_parameters`, and LoRA on `lm_head`/`embed_tokens` all fall back to merged sync with a warning, because the merged path handles them correctly.
- **Shared helpers live in `trl/trainer/utils.py`** (`round_lora_rank`, `validate_lora_for_vllm_sync`, `save_lora_adapter`) so the follow-up `VLLMGeneration` port reuses them unchanged.

### 1. Long run — reward rises, adapter-only sync throughout

**Setup.** `Qwen/Qwen2.5-Math-1.5B` · `sail/Sanity-Test-R1D-1.5B` · LoRA `r=1, alpha=2, target_modules="all-linear"`, `lr=4e-5` (the reference recipe, `oat/scripts/lora/bf16_grpo_tis_lora.sh`) · FSDP2, 4 trainer ranks · vLLM tensor-parallel 2 · 500 steps.

📊 https://huggingface.co/spaces/aminediroHF/async-grpo-sanity-lora-r1d-1.5b

image

**Result.** Reward `0.164 → 0.524` (+219%), monotonic across all ten 50-step blocks. 501 adapter publishes, **0** merged-weight transfers, `ratio ∈ [0.9998, 1.0]` over all 500 steps — i.e. the policy vLLM served matched the policy the trainer scored, for every sync. Reproduced independently by a 100-step run.

### 2. Resume with `SHARDED_STATE_DICT` — works

**Setup.** As above but LoRA `r=32, alpha=64` · FSDP2 **2** trainer ranks, `fsdp_state_dict_type: SHARDED_STATE_DICT` · vLLM TP=2 · 80 steps → checkpoint → fresh vLLM server → resume to 120.

📊 https://huggingface.co/spaces/aminediroHF/async-grpo-lora-resume-ab-sharded

image

**Result.** Reward continues across the resume: last 20 steps before `0.4527` → first 20 after `0.4858`. Both halves of every adapter tensor are bit-identical across the boundary.

### 3. Resume with `FULL_STATE_DICT` — broken upstream

**Setup.** Identical to (2) except one line: `fsdp_state_dict_type: FULL_STATE_DICT`.

📊 https://huggingface.co/spaces/aminediroHF/async-grpo-lora-resume-ab-full

image

**Result.** Reward drops on resume: `0.4355` → `0.3454`, losing **38%** of everything the run had learned, in one step, with no error or warning. Cause is upstream, not in this PR: with FSDP2 + PEFT the `FULL_STATE_DICT` checkpoint stores only rank 0's shard of each adapter tensor, so a resume reverts the other ranks' half (`lora_B` → its zero-init, `lora_A` → a fresh random draw). Measured: the checkpoint file holds exactly 0.5000 of the adapter's elements on 2 ranks, while the complete adapter sits unread beside it in `adapter_model.safetensors`.

Tracked in **https://github.com/huggingface/accelerate/pull/4206**.

**Until that lands, use `fsdp_state_dict_type: SHARDED_STATE_DICT` for FSDP2 + PEFT.** That path uses torch distributed checkpoint, which handles `DTensor` natively — every rank writes and reads its own shard, so nothing needs gathering.

## Notes

- Adapter-only sync is not a blanket win: vLLM serves a LoRA adapter at roughly 1.33–1.39× the time-per-output-token of merged weights, and generation usually dominates RL wall-clock. What it buys is a much shorter pause per sync, ~1% of the sync bandwidth, and QLoRA + FSDP2 working at all.
- `<output_dir>/.vllm_lora/` is a serving cache, not an artifact store — versions outside the staleness window are deleted. Set `save_strategy` if the trained adapter needs to survive the run.

---

> [!NOTE]
> **Medium Risk**
> Touches distributed weight sync, rollout policy versioning, and training logprob paths for PEFT; mis-sync would skew GRPO importance ratios without always failing loudly.
> 
> **Overview**
> **AsyncGRPO** now supports **LoRA training** via a new `peft_config` on `AsyncGRPOTrainer`, with a second weight-sync path that publishes only the adapter to vLLM when the server runs with `--enable-lora`, instead of merging and NCCL-streaming the full model.
> 
> At init the trainer probes `/server_info` (`select_adapter_sync`) and broadcasts the choice: **adapter sync** writes versioned adapters under `output_dir/.vllm_lora/`, loads them through `VLLMClient.load_lora_adapter`, keeps `max_staleness + 1` policy versions registered, and the rollout worker requests `trl-policy-v{N}` as the generation model. **Merged sync** remains the fallback (no LoRA server, unservable PEFT configs, or data-parallel vLLM), with merge/unmerge around NCCL transfer and shared `_iter_vllm_named_params` for manifest/stream alignment. Checkpoints persist **`model_version`** in `rollout_state.json` for safe resume.
> 
> Supporting changes: reject LoRA on `lm_head` (chunked logprobs), `validate_lora_for_vllm_sync` / `save_lora_adapter` / `round_lora_rank`, chunked LM-head **DTensor** gathering and dtype fixes in `trl/trainer/utils.py`, docs + **`examples/async_grpo_math/async_grpo_lora_math.py`**, and broad tests including `FakeVLLMServer`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e733944ade86fa489e0e9e19c75161704f09363. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/source/async_grpo_trainer.md
```diff
@@ -97,6 +97,47 @@ CUDA_VISIBLE_DEVICES=0 VLLM_SERVER_DEV_MODE=1 vllm serve Qwen/Qwen3.5-2B \
...
+    --logprobs-mode processed_logprobs \
+    --weight-transfer-config '{"backend":"nccl"}' \
+    --enable-lora \
+    --max-lora-rank 32 \
+    --max-loras 5
+```
+
+`VLLM_ALLOW_RUNTIME_LORA_UPDATING=1` exposes the endpoint the trainer posts each new adapter to. Add `VLLM_WORKER_MULTIPROC_METHOD=spawn` if you serve tensor-parallel, and keep `--weight-transfer-config` either way: the trainer only chooses a sync mode when it starts, by which point the server is already up, and merged sync is the fallback.
+
+`--max-lora-rank` must be one of `1, 8, 16, 32, 64, 128, 256, 320, 512`. It sets the highest rank the server can serve rather than the rank it will serve, so an `r=4` adapter works fine under `8`. `--max-loras` must be at least `max_staleness + 1`, since the trainer keeps that many adapter versions registered at once. That way a rollout that started under an older policy can finish under it, instead of switching policies mid-generation. The trainer checks both values when it starts and tells you what to restart the server with.
```
qgallouedec: why don't we want to "switch policies mid-generation"? isn't it how full model training works?

docs/source/async_grpo_trainer.md
```diff
@@ -97,6 +97,47 @@ CUDA_VISIBLE_DEVICES=0 VLLM_SERVER_DEV_MODE=1 vllm serve Qwen/Qwen3.5-2B \
...
+
+There is no flag to choose between the two modes. The trainer probes the server on startup and uses adapter sync if the model is a plain LoRA model and the server reports LoRA support. Anything else falls back to merged sync and logs the reason: a server started without `--enable-lora`, a data-parallel server (`/v1/load_lora_adapter` only reaches the replica that answers it, so the others keep serving the base model; tensor parallelism is fine), or an adapter vLLM cannot serve as-is, which means `modules_to_save`, DoRA, `bias != "none"`, `target_parameters`, or LoRA on `lm_head` / `embed_tokens`.
```
qgallouedec: I don't think this is useful in the doc.

docs/source/async_grpo_trainer.md
```diff
@@ -97,6 +97,47 @@ CUDA_VISIBLE_DEVICES=0 VLLM_SERVER_DEV_MODE=1 vllm serve Qwen/Qwen3.5-2B \
...
+
+> [!WARNING]
+> `<output_dir>/.vllm_lora/` is a serving cache, not where your trained adapter is kept. It only holds the most recent versions, and each one is deleted shortly after it falls out of the staleness window. Checkpoints are what preserve the adapter you trained, so do not turn them off: with `save_strategy="no"` a multi-hour run finishes and leaves nothing behind. `output_dir` also has to be readable by the server and has to outlive the job, so node-local scratch will not work, even though it would keep syncs off a slow shared filesystem.
+
+> [!TIP]
+> Adapter sync is not always faster. vLLM generates roughly 1.33–1.39x slower per token when it serves a LoRA adapter than when it serves merged weights, and generation is usually what dominates wall-clock time in RL. What you get in return is a much shorter pause at each sync and about 1% of the sync bandwidth. It is also the only way to run QLoRA with FSDP2, because PEFT cannot merge a quantized weight that FSDP2 has sharded. If you are not using QLoRA and generation throughput is your bottleneck, merged sync is often still the better choice.
```
qgallouedec: interesting

tests/experimental/test_async_grpo_trainer.py
qgallouedec, line 601: It's getting uglier and more fragile. I think this way of testing is awful, but for now it's fine, just as a temporary measure: we'll remove it in the future.

tests/experimental/test_async_grpo_trainer.py
```diff
@@ -509,6 +514,31 @@ def test_inner_training_loop_sets_dataset_start_index_from_file(self):
...
+        trainer.rollout_worker._loop_kwargs = {}
+        trainer.train_dataset = Dataset.from_dict({"prompt": list(range(100))})
+        trainer.accelerator = MagicMock()
+        trainer.accelerator.is_main_process = False
+        trainer._groups_before_resume = 0
+
+        with patch.object(_BaseTrainer, "_inner_training_loop", return_value=None):
+            trainer._inner_training_loop(resume_from_checkpoint=checkpoint_dir)
+
+        assert trainer.model_version == 0
```
qgallouedec, line 694: same here

tests/experimental/test_async_grpo_trainer.py
```diff
@@ -1185,3 +1215,480 @@ def test_epoch_stop_is_fork_independent(self):
...
+
+
+@require_peft
+class TestSaveLoraAdapter(TrlTestCase):
+    def _model_and_env(self, is_main_process=True):
+        from peft import LoraConfig, get_peft_model
+
+        model = AutoModelForCausalLM.from_pretrained("trl-internal-testing/tiny-Qwen2ForCausalLM-2.5")
+        model = get_peft_model(model, LoraConfig(target_modules=["q_proj", "v_proj"]))
+        accelerator = SimpleNamespace(device=torch.device("cpu"), is_main_process=is_main_process)
```
qgallouedec: why cpu?

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -818,6 +861,8 @@ def __init__(
         if args.use_liger_kernel:
             raise NotImplementedError("`use_liger_kernel` is not supported yet.")
 
+        _is_quantized_model = getattr(model, "is_loaded_in_4bit", False) or getattr(model, "is_loaded_in_8bit", False)
```
qgallouedec: it seems that that pr half-addresses qlora: it doesn't add quantization config to the trainer, but it motivates some  changes for qlora. Maybe add qlora in a follow up pr?

trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -48,6 +51,9 @@
     nanmin,
     pad,
     patch_chunked_lm_head,
+    round_lora_rank,
+    save_lora_adapter,
+    validate_lora_for_vllm_sync,
```
qgallouedec: these are only use here, so I suggest that we just move them here

qgallouedec: we might move them in the future to utils, only when they are use in more that one place.

qgallouedec (commented):
Thanks, a few changes and questions. Also, do you know if https://github.com/vllm-project/vllm/pull/53751 can simplify the whole thing?

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: yeah I thought I'd open up a follow up PR on GRPO to use the same adapter only mechanism.

tests/experimental/test_async_grpo_trainer.py
AmineDiro: I agree with you, actually I am not really proud of these tests. I can work on something more robust for LoRA that looks more like an integrration test. I never liked mocking in the first place. This is more interface tests that will work anyway.

tests/experimental/test_async_grpo_trainer.py
AmineDiro: I will work on removing in mocking nonsense 👍🏼

trl/experimental/async_grpo/async_grpo_trainer.py
AmineDiro: thats fair, I didnt tests quantized model, the `_is_quantized_model` here is just to make sure we use higher precision for lora ( model.dtype). 

Thinking about it, I  can drop it now because it might interact weirdly with config dtype that is used to load the model.

AmineDiro:
Thanks a lot @qgallouedec  for the quick review ! 

> Thanks, a few changes and questions. Also, do you know if https://github.com/vllm-project/vllm/pull/53751 can simplify the whole thing?

Actually I am not sure this sparse weight transfer is aimed for LoRA adapters. I need to test it but I checked vllm issues/PRs and it seems to me that lora weight transfer isn't supported yet : 
https://github.com/vllm-project/vllm/issues/20149 
https://github.com/vllm-project/vllm/issues/4068

I'll fix up the tests as I am also not really happy with the current code quality tbh.

docs/source/async_grpo_trainer.md
AmineDiro: >? isn't it how full model training works?

in Full model training we can weight sync during a decode request. We call `pause` on the engine with a `mode=keep` and then we swap the weights and continues decoding under the new policy (kv cache preserved)

For LoRA sync, the re-prefill (after the pause) runs with the adapter the request named at **submit time** so  policy-vN ( if we are at N+1_ for exam. That is precisely why vN has to stay loaded: the whole sequence is re-encoded through vN and  finishes under vN. 

We could use the same name for the adapter ` /v1/load_lora_adapter` with `load_inplace=true`  to force reuse the kv cache prefilled with the old adapter and continue with the new one, but there is an issue where the adapter is re-read from disk on every batch :/ see : https://github.com/vllm-project/vllm/pull/41482

docs/source/async_grpo_trainer.md
AmineDiro: dropping

docs/source/async_grpo_trainer.md
AmineDiro: The overhead of going through disk kills the speed. I'll take a look if I can contribute NCCL LoRA weight transfer which should be waaay faster

AmineDiro:
@qgallouedec [a511e7f](https://github.com/huggingface/trl/pull/7017/commits/a511e7f518976e6d9a9511ad9ec098432ecdd1e8) I revamped the tests, these should be waay more readable now and test the vllm lora support using a stub vllm client directly against the trainer and not Mock class

qgallouedec:
Nice, the tests read much better without the mocks.

two things.
- on the helpers, i'd still move them into `async_grpo_trainer.py`. it's not just that they have one caller: trl/trainer/utils.py is main code and the only consumer is experimental, which is the wrong direction for the boundary When the GRPO follow-up lands and there are two callers, that's the moment to promote them to utils.

And the stub accelerator hardcodes torch.device("cpu"), can you use torch_device? otherwise a device mismatch in save_lora_adapter can't show up on a gpu runner.

qgallouedec:
Thanks for addressing my points!

I think one bugbot finding is real.

Head LoRA gives wrong logprobs, silently. `SFTTrainer` raises for this exact case rather than allowing it, we should probably do this as well here.

The other two are worth a look rather than a fix from me

Separately, this now collides with #7077 and #7075. Worth syncing with @kashif on the order.

AmineDiro:
> Head LoRA gives wrong logprobs, silently. SFTTrainer raises for this exact case rather than allowing it, we should probably do this as well here.

Oh I see, I misread the error, and checked  PEFT. When we set `all-linear` it  doesnt wrap lm_head, so the default recipe works no problem . The error occurs if someone sets  an explicit  "lm_head" in target_modules or a regex that matches it.

> Separately, this now collides with https://github.com/huggingface/trl/pull/7077 and https://github.com/huggingface/trl/pull/7075. Worth syncing with @kashif on the order.

I'll ping @kashif  on this as this should be a special case it

AmineDiro:
Fix in 4e73394 . Good to go 🛫

qgallouedec (approved):
let's go!!

PR #7008, opened 2026-09-02
fix: fall back when model config omits architecture metadata

YZJF:
## What does this PR do?

Fixes #7006.

`create_model_from_path` promises to fall back to `AutoModelForCausalLM` when it cannot infer a model architecture. It previously indexed optional `config.architectures` and accessed optional `config.auto_map` before that fallback, so string model paths could crash when either metadata field was absent or empty.

This change treats empty architecture metadata as unknown, safely reads the optional remote-code mapping, and continues through the existing CausalLM fallback. It adds regression coverage for both `architectures=None` and `architectures=[]`.

The follow-up removes an unnecessary defensive `getattr` around `config.architectures` and makes the VLM online-DPO example fall back to `AutoModelForImageTextToText` when its config has no architecture name. The separate `subfolder` forwarding work in #7026 is intentionally not duplicated here.

## Before submitting

- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request), Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

## AI writing disclosure

We welcome the use of AI tools to help with contributions. For transparency and to help us improve our review process, please indicate the level of AI involvement in this PR.

- [ ] No AI usage: the PR was written entirely by a human.
- [x] AI-assisted: some parts were suggested or improved by AI, but the PR was written and reviewed by a human.
- [ ] AI-generated: the PR was mostly or fully generated by an AI tool.

## Testing

- `python -m pytest tests/test_utils.py::TestCreateModelFromPath::test_falls_back_to_causal_lm_without_architecture_metadata -q` — 2 passed.
- `python -m pytest tests/test_utils.py -q` — 140 passed, 97 skipped.
- Source-level `None`/empty-list smoke and `py_compile` for the VLM example — passed.
- `pre-commit run --files trl/trainer/utils.py examples/online_dpo_visual_math/online_dpo_visual_math.py` — passed.

## Who can review?

Anyone in the community is free to review the PR once the tests have passed.

PR #7026, opened 2026-09-03
Fix config loading options when loading models

### LINKED ISSUE #7025: create_model_from_path ignores subfolder when inferring the model architecture
### System Info

- TRL main: `839728927a093e0cc9c4821c28461b877e3daeb4` (`1.9.0.dev0`)
- Linux / WSL2, Python 3.12.13
- PyTorch 2.13.0+cu126, Transformers 5.13.1, Accelerate 1.14.0
- CPU-only reproduction, no downloads or GPU required

### Reproduction

`create_model_from_path` forwards `subfolder` to the model loader, but not to the preceding `AutoConfig.from_pretrained` call used to choose the model class. A checkpoint stored only in a subdirectory therefore fails before the model loader is reached.

Run against the main commit above:

```python
import os
from pathlib import Path
from tempfile import TemporaryDirectory

os.environ["HF_HUB_OFFLINE"] = "1"
os.environ["CUDA_VISIBLE_DEVICES"] = ""

from transformers import AutoModelForCausalLM, GPT2Config, GPT2LMHeadModel
from trl.trainer.utils import create_model_from_path

with TemporaryDirectory() as directory:
    config = GPT2Config(vocab_size=32, n_positions=16, n_embd=16, n_layer=1, n_head=2, bos_token_id=0, eos_token_id=1)
    GPT2LMHeadModel(config).save_pretrained(Path(directory) / "checkpoint")
    kwargs = dict(subfolder="checkpoint", local_files_only=True, device_map=None)
    AutoModelForCausalLM.from_pretrained(directory, **kwargs)  # succeeds
    create_model_from_path(directory, **kwargs)  # fails while loading AutoConfig
```

Traceback (local path prefixes shortened):

```text
  File "repro_trl_subfolder_submission.py", line 16, in <module>
    create_model_from_path(directory, **kwargs)
  File "trl/trainer/utils.py", line 1182, in create_model_from_path
    config = AutoConfig.from_pretrained(model_id, trust_remote_code=kwargs.get("trust_remote_code", False))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "transformers/models/auto/configuration_auto.py", line 421, in from_pretrained
    raise ValueError(
ValueError: Unrecognized model in /tmp/tmpw6_g2oyy. Should have a `model_type` key in its config.json.
```

The same checkpoint loads through TRL when its subdirectory is passed directly as `model_id`.

### Expected behavior

When `subfolder` is provided, architecture inference should read the configuration from the same directory as the model weights.

I plan to submit a focused fix that forwards `subfolder` to this configuration lookup, with offline regression tests saving a tiny model at the root and in a subdirectory and comparing the loaded logits. This does not change the architecture fallback in #7008 or the processor-loading behavior in #6978.

### Checklist

- [x] I checked existing issues and PRs for this problem.
- [x] System information is included.
- [x] The reproduction is minimal and self-contained.
- [x] Code is formatted in code blocks.
- [x] The traceback includes all frames.

DaoyuanLi2816:
## What does this PR do?

Fixes #7025.

Forward `subfolder`, `revision`, `cache_dir`, `token`, and `local_files_only` to the configuration lookup in `create_model_from_path`, so architecture inference reads the same checkpoint directory as the model loader. Previously, a checkpoint saved only in a subdirectory failed during `AutoConfig.from_pretrained`, even though Transformers loaded it successfully with the same arguments.

The regression test saves a tiny GPT-2 locally and compares the loaded logits with the original for root-directory and subdirectory checkpoints. Before the fix, the subdirectory case fails and the root case passes; both pass after the fix. No Hub fixture or GPU is needed for these cases.

Following the collaborator review, the same lookup now also respects the checkpoint revision, cache location, authentication option, and offline-only setting. Two additional offline tests load a tiny GPT-2 from a temporary Hub-cache snapshot at a non-default revision, both at the root and in a subfolder. A forwarding test checks that model-only options stay out of the config lookup; the test class now inherits `TrlTestCase`.

This does not change the architecture fallback in #7008 or processor loading in #6978. The #7008/#7026 changes are semantically independent but touch adjacent lines: whichever lands second needs to preserve both changes. If #7008 lands first, I will merge the updated main into this branch without rewriting reviewed commits.

## Validation

Current head (Python 3.10, Transformers 5.14.1, CPU/offline):

- Three added cases fail before the loading-option fix; the two original subfolder cases still pass.
- After the fix: 66 passed, 174 deselected in the focused CPU utility selection, including all five model-loading cases. This is not the full test suite.
- Repository-pinned pre-commit hooks: Ruff check, Ruff format, and doc-builder style passed.
- `git diff --check` passed.
- Private Hub authentication and remote-code checkpoints were not exercised; `token` forwarding is checked with a dummy value and mocked loaders.

## Before submitting

- [x] Read the contributor guidelines.
- [x] Linked the current-main reproduction in #7025; no prior maintainer approval is claimed.
- [x] Added regression tests for the bug and unchanged root-directory behavior.
- [x] Documentation reviewed; this restores an existing model-loading argument and adds no new API.

---

> [!NOTE]
> **Low Risk**
> Narrow bugfix in model-loading helper with regression tests; no API changes beyond restoring correct argument forwarding.
> 
> **Overview**
> Fixes checkpoint loading when weights live in a **subfolder** by aligning config discovery with model loading in `create_model_from_path`.
> 
> When architecture is inferred, **`AutoConfig.from_pretrained`** now receives the same Hub/path kwargs as **`from_pretrained`** on the model (`subfolder`, `revision`, `cache_dir`, `token`, `local_files_only`, plus existing `trust_remote_code`). Previously config was read from the repo root while the model could load from a subdirectory, causing failures for subdirectory-only checkpoints.
> 
> Adds **`TestCreateModelFromPathSubfolder`** with local GPT-2 regression tests (root vs `checkpoint` subfolder, offline cached revision) and a mocked test that config kwargs are forwarded without leaking model-only options.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e21e7166d95f780c72e9f6aaefd71716b5291f2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -25,7 +25,7 @@ import torch.nn.functional as F
 import transformers
 from datasets import IterableDataset
 from packaging.version import Version
-from transformers import AutoConfig, AutoModelForCausalLM, PretrainedConfig
+from transformers import AutoConfig, AutoModelForCausalLM, GPT2Config, GPT2LMHeadModel, PretrainedConfig
 from transformers.testing_utils import torch_device
 from transformers.utils import is_peft_available
 
@@ -36,6 +36,7 @@ from trl.trainer.utils import (
     adjusted_mfu,
     compute_flops_per_token,
     compute_mfu,
+    create_model_from_path,
     entropy_from_logits,
     flush_left,
     generate_model_card,
@@ -113,6 +114,26 @@ class TestUseAdapter(TrlTestCase):
 
         assert torch.equal(output_1, expected_1)
         assert torch.equal(output_2, expected_2)
+
+
+class TestCreateModelFromPathSubfolder:
+    @pytest.mark.parametrize("subfolder", ["", "checkpoint"])
+    def test_loads_model_and_config_from_same_directory(self, tmp_path, subfolder):
+        config = GPT2Config(
+            vocab_size=32, n_positions=16, n_embd=16, n_layer=1, n_head=2, bos_token_id=0, eos_token_id=1
+        )
+        model = GPT2LMHeadModel(config).eval()
+        model.save_pretrained(tmp_path / subfolder)
+
+        loaded_model = create_model_from_path(
+            str(tmp_path), subfolder=subfolder, local_files_only=True, device_map=None
+        )
+
+        input_ids = torch.tensor([[1, 2, 3]])
+        with torch.no_grad():
+            expected = model(input_ids).logits
+            actual = loaded_model(input_ids).logits
+        torch.testing.assert_close(actual, expected)
 
 
 class TestPad(TrlTestCase):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -116,7 +116,7 @@ class TestUseAdapter(TrlTestCase):
         assert torch.equal(output_2, expected_2)
 
 
-class TestCreateModelFromPathSubfolder:
+class TestCreateModelFromPathSubfolder(TrlTestCase):
     @pytest.mark.parametrize("subfolder", ["", "checkpoint"])
     def test_loads_model_and_config_from_same_directory(self, tmp_path, subfolder):
         config = GPT2Config(
@@ -134,6 +134,56 @@ class TestCreateModelFromPathSubfolder:
             expected = model(input_ids).logits
             actual = loaded_model(input_ids).logits
         torch.testing.assert_close(actual, expected)
+
+    @pytest.mark.parametrize("subfolder", ["", "checkpoint"])
+    def test_loads_config_from_requested_cached_revision(self, tmp_path, subfolder):
+        model_id = "trl-offline-tests/tiny-gpt2"
+        revision = "b" * 40
+        snapshot = tmp_path / "models--trl-offline-tests--tiny-gpt2" / "snapshots" / revision
+        config = GPT2Config(
+            vocab_size=32, n_positions=16, n_embd=16, n_layer=1, n_head=2, bos_token_id=0, eos_token_id=1
+        )
+        model = GPT2LMHeadModel(config).eval()
+        model.save_pretrained(snapshot / subfolder)
+
+        # No default-revision snapshot exists, and the cache is outside the user's Hub cache.
+        with patch("huggingface_hub.constants.HF_HUB_OFFLINE", True):
+            loaded_model = create_model_from_path(
+                model_id,
+                revision=revision,
+                cache_dir=str(tmp_path),
+                subfolder=subfolder,
+                local_files_only=True,
+                device_map=None,
+            )
+
+        input_ids = torch.tensor([[1, 2, 3]])
+        with torch.no_grad():
+            expected = model(input_ids).logits
+            actual = loaded_model(input_ids).logits
+        torch.testing.assert_close(actual, expected)
+
+    def test_forwards_config_loading_options_without_model_only_kwargs(self):
+        config = GPT2Config(architectures=["GPT2LMHeadModel"])
+        loading_kwargs = {
+            "subfolder": "checkpoint",
+            "revision": "custom-revision",
+            "cache_dir": "custom-cache",
+            "token": "test-token-not-a-credential",
+            "local_files_only": True,
+            "trust_remote_code": True,
+        }
+        with (
+            patch("trl.trainer.utils.AutoConfig.from_pretrained", return_value=config) as config_loader,
+            patch("transformers.GPT2LMHeadModel.from_pretrained") as model_loader,
+        ):
+            result = create_model_from_path("model-id", **loading_kwargs, device_map=None, attn_implementation="eager")
+
+        config_loader.assert_called_once_with("model-id", **loading_kwargs)
+        model_loader.assert_called_once_with(
+            "model-id", **loading_kwargs, device_map=None, attn_implementation="eager", dtype=torch.float32
+        )
+        assert result is model_loader.return_value
 
 
 class TestPad(TrlTestCase):
--- a/trl/trainer/utils.py
+++ b/trl/trainer/utils.py
@@ -1179,7 +1179,11 @@ def create_model_from_path(
         kwargs["device_map"] = None if PartialState().device.type == "cpu" else "auto"
     if architecture is None:
         # Best effort to infer architecture from config, but we fall back to AutoModelForCausalLM if we can't find it
-        config = AutoConfig.from_pretrained(model_id, trust_remote_code=kwargs.get("trust_remote_code", False))
+        config = AutoConfig.from_pretrained(
+            model_id,
+            subfolder=kwargs.get("subfolder", ""),
+            trust_remote_code=kwargs.get("trust_remote_code", False),
+        )
         architecture = getattr(transformers, config.architectures[0], None)
         if architecture is None:
             # Remote-code checkpoint: the architecture name lives in the dynamic module, not in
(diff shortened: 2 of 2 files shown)


behroozazarkhalili:
Verdict: the fix is correct. One coordination item with #7008 (textual merge conflict) is worth resolving before merge.

## What I checked

Environment: transformers 5.11.0, `HF_HUB_OFFLINE=1`, `CUDA_VISIBLE_DEVICES=""`, CPU only, `PYTHONPATH` pinned to the PR worktree.

GREEN on PR head, exactly the new test ids:

```
$ python -m pytest tests/test_utils.py::TestCreateModelFromPathSubfolder -p no:randomly -q
..                                                                       [100%]
2 passed in 20.15s
```

RED on base (`8397289` `trl/trainer/utils.py` restored, PR tests kept):

```
E       ValueError: Unrecognized model in /tmp/.../test_loads_model_and_config_fr1. Should have a `model_type` key in its config.json.
FAILED tests/test_utils.py::TestCreateModelFromPathSubfolder::test_loads_model_and_config_from_same_directory[checkpoint]
1 failed, 1 passed in 13.67s
```

That is the same `ValueError` and the same frame as the traceback in #7025, so the test is a real oracle for the reported scenario and the empty-subfolder case guards the pre-existing path.

Mutant check: replacing `subfolder=kwargs.get("subfolder", "")` with the hardcoded `subfolder="checkpoint"` at `trl/trainer/utils.py:1184` does not pass:

```
FAILED tests/test_utils.py::TestCreateModelFromPathSubfolder::test_loads_model_and_config_from_same_directory[]
1 failed, 1 passed in 12.08s
```

So the parametrization is doing work and a narrower or constant fix is rejected.

Sibling scan: `create_model_from_path` is the only definition in the tree (`trl/trainer/utils.py:1145`) and the 20 call sites in `trl/trainer/*` and `trl/experimental/*` all go through it, so there is no duplicated copy of this block to keep in sync. `subfolder` appears nowhere else in `trl/` besides the new line.

`ruff check` and `ruff format --check` at the CI-pinned 0.13.3 pass on both changed files.

## Findings

1. `trl/trainer/utils.py:1182-1186` conflicts textually with open PR #7008, which rewrites the two lines immediately below the same `AutoConfig.from_pretrained` call. Merging #7008 into this head produces a real conflict in `trl/trainer/utils.py` (`CONFLICT (content)`), while `tests/test_utils.py` auto-merges. The two changes are semantically orthogonal: keeping this PR's multi-line `AutoConfig.from_pretrained` call together with #7008's `architecture_names` lines resolves it. Whichever lands second should rebase; it may be worth saying so in both PRs.

2. `trl/trainer/utils.py:1182-1186` forwards only `subfolder` and `trust_remote_code`, so a Hub checkpoint loaded with `revision`, `cache_dir`, `token`, or `local_files_only` still has its architecture inferred from the default revision and the default cache. This is pre-existing and out of scope for the linked issue, but `revision` in particular can select a different `config.architectures` than the weights being loaded. A follow-up that forwards the loading kwargs the config lookup shares would close the class of bug rather than this one instance.

3. `tests/test_utils.py:119` declares `TestCreateModelFromPathSubfolder` without inheriting `TrlTestCase`. 20 of the 23 test classes in this file inherit it, though two others (`TestChunkedLogProbFunction`, `TestPatchChunkedLMHead`) also do not, and the class only needs `tmp_path`, which it takes as a fixture. Cosmetic only.

## Not reproduced

- No regression for `subfolder=None`. `AutoConfig.from_pretrained(path, subfolder=None)`, `AutoModelForCausalLM.from_pretrained(path, subfolder=None)`, and `create_model_from_path(path, subfolder=None)` all raise the identical `TypeError: expected str, bytes or os.PathLike object, not NoneType`, so the new argument does not introduce a failure mode the weight loader did not already have.
- No behaviour change for callers that do not pass `subfolder`: the default `""` reproduces the previous call, and the empty-subfolder parametrization passes on base as well as head.

## Unevaluable

- Real Hub checkpoints with a `subfolder` plus a non-default `revision`, and remote-code checkpoints whose `auto_map` lives in a subfolder. Both need network access; neither is required to accept this fix.
- The pinned doc-builder style hook, which does not cover `trl/trainer/utils.py` changes of this shape in any case.

DaoyuanLi2816:
@behroozazarkhalili Addressed the loading-options follow-up in 1e21e71: config inference now also receives `revision`, `cache_dir`, `token`, and `local_files_only`, without forwarding model-only kwargs. Added offline cache/revision regressions for root and subfolder checkpoints plus an argument-forwarding check; these three cases fail before the fix. The updated CPU utility selection passes (66 passed, 174 deselected), as do the pinned Ruff/doc-builder hooks. The class now inherits `TrlTestCase`. Private-Hub authentication itself was not tested.

The architecture fallback remains owned by #7008. Its change and this config-loading change should both be retained when resolving the adjacent-line conflict. If #7008 lands first, I will merge the updated main into this branch without rewriting the reviewed commits; I am noting that coordination on #7008 as well.

behroozazarkhalili:
Re-reviewed 1e21e716. Verdict: the follow-up closes finding 2 and finding 3; finding 1 (the textual conflict with #7008) is still live and is the only thing left, and it is coordination rather than a defect.

## What I checked

Environment: transformers 5.11.0, `HF_HUB_OFFLINE=1`, CPU only, `PYTHONPATH` pinned to the PR worktree.

GREEN on the PR head, the five new test ids:

```
pytest tests/test_utils.py::TestCreateModelFromPathSubfolder -p no:randomly -q
5 passed in 23.98s
```

RED with `trl/trainer/utils.py` restored to the base `83972892` and the PR tests kept:

```
FAILED ...test_loads_model_and_config_from_same_directory[checkpoint]
FAILED ...test_loads_config_from_requested_cached_revision[]
FAILED ...test_loads_config_from_requested_cached_revision[checkpoint]
FAILED ...test_forwards_config_loading_options_without_model_only_kwargs
4 failed, 1 passed
```

The one that passes on base is the empty-subfolder same-directory case, which is the pre-existing path and is meant to pass on both trees. The two `requested_cached_revision` cases are the new oracle for finding 2: the cache holds only a non-default revision snapshot, so a config lookup that ignores `revision` or `cache_dir` cannot find a config at all. Both fail on base and pass on head.

`ruff check` and `ruff format --check` at the CI-pinned 0.13.3 pass on both files, and the pinned doc-builder style check passes.

## Findings from September 3

1. Conflict with #7008: still present. `git merge-tree` of #7008's current head onto this head reports a content conflict in `trl/trainer/utils.py`; this head merges cleanly onto current main (82 commits behind, no conflicts). The resolution is unchanged: keep this PR's multi-line `AutoConfig.from_pretrained(...)` call and #7008's `architecture_names` lines below it. Whichever lands second rebases.
2. Loading options: fixed. `subfolder`, `revision`, `cache_dir`, `token`, `local_files_only` and `trust_remote_code` now reach the config lookup, and the forwarding test pins that model-only kwargs (`device_map`, `attn_implementation`) do not, while the weight loader still receives everything plus `dtype`.
3. `TrlTestCase`: fixed, the class inherits it now.

## Not reproduced

Nothing in the update failed to reproduce. The default values match the transformers defaults (`revision="main"`, `local_files_only=False`, `cache_dir=None`, `token=None`), so a caller that passes none of them gets the same lookup as before.

## Unevaluable

Real Hub checkpoints with a non-default revision still need network access; the cached-snapshot test stands in for that by building the cache layout by hand under `HF_HUB_OFFLINE`.

PR #7030, opened 2026-09-03
Add the missing trailing newline to the dev Dockerfile

albertvillanova:
This PR adds the missing trailing newline to `docker/trl-dev/Dockerfile`.

### Motivation

The file has never ended with a newline since it was added in #3931, and #4522 changed its last line without adding one. Its sibling `docker/trl/Dockerfile` had the same problem, which was fixed in #6992 as part of the pinning change, because that PR already rewrote the affected line. This one was left out of that PR on purpose, to keep it to the files it was actually changing.

So this is the last of the pair, and it makes the two Dockerfiles consistent again.

Without the newline, every diff touching the last line shows a `\ No newline at end of file` marker on both sides, and tools that read the file line by line see a final line with no terminator.

### Changes

- Add the trailing newline to the dev image Dockerfile

Note that there are around 60 files in the repository with no trailing newline, most of them chat templates under `trl/chat_templates`. This PR deliberately does not touch those: a trailing newline can matter for a rendered chat template, so that would need its own reasoning rather than a blanket sweep. There is also no `end-of-file-fixer` hook in `.pre-commit-config.yaml`, so nothing enforces this today.

---

> [!NOTE]
> **Low Risk**
> Whitespace-only edit to a Dockerfile; no runtime or build behavior change.
> 
> **Overview**
> Adds a **trailing newline** to `docker/trl-dev/Dockerfile` so the file ends with a proper line terminator, matching `docker/trl/Dockerfile` and avoiding `\ No newline at end of file` noise in future diffs on the last line.
> 
> There is **no change** to image build steps or installed packages—the final `RUN uv pip install ...` line is unchanged aside from the newline.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 703dda75ebbf8f8e70b7c784b9608142a8f9d6ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:

PR #7032, opened 2026-09-03
fix(cli): honor CLI accelerate config precedence

### LINKED ISSUE #7031: Bug: CLI --accelerate_config does not override YAML config
## Reproduction

`TrlParser` documents that command-line arguments override values supplied by a YAML config file. This does not hold for the launcher-only `accelerate_config` option when the YAML file and the CLI both provide the documented two-token spelling.

The following MRE exercises the real `trl sft` command and only mocks the process launcher, so it does not start a training job:

```python
from pathlib import Path
from tempfile import TemporaryDirectory
from unittest.mock import patch

from trl.cli import main

with TemporaryDirectory() as directory:
    config_path = Path(directory) / "config.yaml"
    config_path.write_text("accelerate_config: zero2\\n")

    with patch("trl.cli.accelerate_launcher.launch_training_script") as launch:
        main(["sft", "--config", str(config_path), "--accelerate_config", "zero3"])

    launch_args = launch.call_args.kwargs["launch_args"]
    assert Path(launch_args[1]).name == "zero3.yaml"
    assert "--accelerate_config" not in launch_args
```

On `main` at `4506f4af8ccc5b7f8b337f589a696a19913d8a4a`, the first assertion fails because the launcher receives `zero2.yaml` from the YAML file, and the command-line `--accelerate_config zero3` pair remains in its forwarded arguments:

```text
AssertionError: assert 'zero2.yaml' == 'zero3.yaml'
```

## Expected behavior

The command-line value should take precedence: the launcher should receive `--config_file .../zero3.yaml`, and neither `--accelerate_config` token should remain in its arguments.

## System Info

- Platform: macOS-26.5.2-arm64-arm-64bit-Mach-O
- Python: 3.13.5
- PyTorch: 2.14.0
- Transformers: 5.16.1
- Accelerate: 1.14.0
- Datasets: 5.0.1
- TRL: 1.13.0.dev0
- Accelerator: CPU

## Checklist

- [x] I have checked that my issue isn't already filed.
- [x] I have included my system information.
- [x] Any code provided is minimal, complete, and reproducible.
- [x] Any code provided is properly formatted in code blocks, with no screenshots.
- [x] Any traceback provided is complete.

YZJF:
# What does this PR do?

Fixes #7031.

When a YAML config provides `accelerate_config` and the command line provides the documented
`--accelerate_config <name>` form, `TrainingCommand` concatenates the YAML-only launcher
arguments first. The resolver therefore selects the YAML value and forwards the CLI pair
unchanged. This removes the YAML pair when the CLI explicitly supplies that option, so the
documented command-line precedence is preserved.

The change is deliberately limited to the two-token form. It does not modify the separate
equals-sign syntax work.

This follow-up adds the requested YAML-only regression: a config containing
`accelerate_config: zero2` still forwards `zero2.yaml` to the launcher when no CLI override is
present.

## Validation

- `python -m pytest -p no:rerunfailures tests/test_cli_training.py -q` — 2 passed.
- `HF_DATASETS_OFFLINE=1 HF_HUB_OFFLINE=1 python -m pytest -p no:rerunfailures tests/test_cli_utils.py -q` — 35 passed.
- `python -m pytest -p no:rerunfailures tests/test_cli.py -k "help_no_type_error or vllm_serve_config_file" -q` — 8 passed.
- `pre-commit run --files tests/test_cli_training.py` — passed.

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request-pr), Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? The related bug report is #7031.

## AI writing disclosure


behroozazarkhalili:
Verdict: the fix is correct for the reported scenario, with one test-coverage change worth making before merge.

## What I checked

Worktree at PR head `15e09fa5`, base `4506f4af`. Interpreter `/scratch/ermia/venvs/hf_trl/bin/python`, `PYTHONPATH` pinned to the worktree (`import trl` resolves to `/scratch/ermia/tmp/rv-7032/trl/__init__.py`).

RED/GREEN on the exact new test id:

```
# PR head
python -m pytest -p no:rerunfailures tests/test_cli_training.py -q
1 passed, 1 warning in 8.02s

# same test against the base copy of trl/cli/commands/training.py (PR tests kept)
>       assert Path(launch_args[1]).name == "zero3.yaml"
E       AssertionError: assert 'zero2.yaml' == 'zero3.yaml'
1 failed, 1 warning in 11.20s
```

That is the exact assertion and message quoted in #7031, so the test is a real oracle for the reported bug.

Precedence matrix through `trl.cli.main` with only `launch_training_script` mocked, on the PR head (`launch_args`, `training_script_args`):

```
A yaml-only  : (['--config_file', '.../zero2.yaml'], ['--config', '.../config.yaml'])
B cli-only   : (['--config_file', '.../zero3.yaml'], ['--config', '.../config.yaml'])
C both 2tok  : (['--config_file', '.../zero3.yaml'], ['--config', '.../config.yaml'])
D cli equals : (['--config_file', '.../zero2.yaml', '--accelerate_config=zero3'], [...])
E extra arg  : (['--config_file', '.../zero3.yaml', '--num_processes', '2', '--num_processes', '4'], [...])
```

A, B and C are the wanted behaviour. E confirms that other launcher-only options already got command-line precedence from `accelerate`'s own last-wins parsing, so scoping the fix to `--accelerate_config` is the right size: the bug came specifically from `launch_args.index(...)` picking the first occurrence in `trl/cli/accelerate_config.py:29`.

The `del config_remaining[i : i + 2]` slice is safe because `set_defaults_with_config` always emits `--key value` pairs (`trl/scripts/utils.py:385-387`) and YAML keys are unique, so `index` finds the only occurrence.

`ruff check` and `ruff format --check` at the CI-pinned 0.13.3 pass on both changed files. `tests/test_cli_utils.py` is 35 passed, matching the PR description.

## Findings

1. `tests/test_cli_training.py:19` (medium): the test does not pin the config-only path, so it does not protect the guard it is testing. Dropping just `"--accelerate_config" in cli_remaining and` from `trl/cli/commands/training.py:65` is a plausible simplification a later reader could make, and the suite stays green (`1 passed`) while `trl sft --config cfg.yaml` with `accelerate_config: zero2` in the YAML then launches with `launch_args == []`, that is with no `--config_file` at all, silently losing the DeepSpeed config. Suggestion: add the yaml-only assertion to the same test, for example run `main(["sft", "--config", str(config_path)])` and assert `Path(launch_args[1]).name == "zero2.yaml"`.

2. `trl/cli/commands/training.py:65` (medium): the guard only recognises the two-token command-line spelling. With `accelerate_config: zero2` in the YAML and `--accelerate_config=zero3` on the command line (row D above) the YAML value still wins and the equals token survives into `launch_args`, where the accelerate parser rejects it: `Accelerate launch command: error: unrecognized arguments: --accelerate_config=zero3` (SystemExit 2). This is not a regression, base behaves the same, and the PR description says the equals form is out of scope, but the issue claims documented command-line precedence is restored and that is only true for one of the two spellings. Suggestion: say so in the PR description, or handle it by matching `arg.startswith("--accelerate_config=")` as well.

3. `trl/cli/accelerate_config.py:29,46` (low): the first-occurrence choice that causes the bug lives in the resolver, and the PR fixes it one layer up by special-casing a single flag in the command. Making the resolver pick the last occurrence and strip every `--accelerate_config` pair produces the same results with no change to `training.py`. I ran that variant against base: the new test passes and rows A, B, C, E are identical to the PR head. It keeps the precedence rule in the one function that owns it, which reads better against the leanness rule in AGENTS.md. Either shape is defensible.

4. `tests/test_cli_training.py:1` (low): a new file for a single test, when `tests/test_cli.py` already holds the closest analogue, `test_vllm_serve_config_file` at `tests/test_cli.py:129`, using the same mock-the-launcher pattern. Suggestion: move the test there. That file also uses `yaml.dump` rather than a handwritten `write_text`.

No sibling copies to keep in sync: `resolve_accelerate_config_argument` has exactly one call site (`trl/cli/commands/training.py:68`), and `VllmServeCommand` (`trl/cli/commands/vllm_serve.py:35-38`) does not merge remaining strings this way.

## Not reproduced

- No regression in the yaml-only or cli-only paths on the PR head (rows A and B match base).
- No behaviour change for other launcher-only options (row E).

## Unevaluable

- The end-to-end `accelerate launch` with a real DeepSpeed config: no GPU on the machine used, and every check here mocks the launcher.

PR #7038, opened 2026-09-03
Fail fast for MoE auxiliary loss with Liger GRPO

### LINKED ISSUE #7009: GRPOTrainer silently drops the MoE auxiliary loss when use_liger_kernel=True
### Summary

`GRPOTrainer` with a Mixture-of-Experts model and `use_liger_kernel=True` drops the router
load-balancing auxiliary loss from the training objective. There is no error and no warning, and
`trainer.aux_loss_enabled` stays `True` throughout, so nothing in the run indicates the term is gone.

`DPOTrainer` and `KTOTrainer` already raise for this exact combination. GRPO's copy of that block has no
such guard.

### Why it happens

`compute_loss` forks on the Liger flag and returns, so `_compute_loss` never runs:

https://github.com/huggingface/trl/blob/main/trl/trainer/grpo_trainer.py#L3033

`_compute_loss` is the only place the auxiliary loss is added:

https://github.com/huggingface/trl/blob/main/trl/trainer/grpo_trainer.py#L3346-L3348

`compute_liger_loss` never references `aux_loss`, `router_aux_loss_coef`, or `output_router_logits`,
and `LigerFusedLinearGRPOLoss` takes no router argument. The auxiliary loss is requested only via
`compute_aux_loss=self.aux_loss_enabled` inside `_get_per_token_logps_and_entropies`, which the Liger
path does not call.

`aux_loss_enabled` is still computed and still `True`:

https://github.com/huggingface/trl/blob/main/trl/trainer/grpo_trainer.py#L788-L789

### How easy this is to hit

`router_aux_loss_coef` defaults to `0.001`, so `aux_loss_enabled` is `True` for any MoE model with no
configuration at all. Turning on `use_liger_kernel` is the only thing a user has to do to lose the
term. The `GRPOConfig` docstring says the auxiliary loss "is added to the training loss with this
weight", and the `aux_loss` metric documentation does not mention a Liger exception.

### Measurement

A 2x2 over `{standard, Liger} x {router_aux_loss_coef 0.0, 0.1}` on
`trl-internal-testing/tiny-Qwen3MoeForCausalLM`, one H100, seed 42, everything else identical:

| arm | `aux_loss_enabled` | `aux_loss` logged | loss |
|---|---|---|---|
| standard, coef 0.0 | False | no | 0.0 |
| standard, coef 0.1 | True | yes, 2.239327907562256 | 0.22393278777599335 |
| Liger, coef 0.0 | False | no | 1.4901161193847656e-08 |
| Liger, coef 0.1 | True | no | 1.4901161193847656e-08 |

Changing the coefficient moves the standard loss by 0.2239. It moves the Liger loss by exactly zero:
the two Liger values are bit-identical (`0x3e50000000000000`). So the coefficient has no effect at
all under Liger, which is what separates a missing metric from a missing term in the objective.

Two checks on the numbers. The standard delta matches `coef * aux_loss` to float32 resolution
(0.22393278777599335 against 0.2239327907562256, differing by 2.98e-09), so at the first step the
whole standard-path loss is the auxiliary term. And `reward_std` is 1.113 in all four arms, so no arm
is degenerate. The Liger value of `2**-26` is the fused kernel's rounding around a zero policy loss,
and it does not hide a 0.2239 term.

### Reproduction

`tests/test_dpo_trainer.py::test_init_fails_with_moe_aux_loss_and_liger` is already the test for this,
for DPO. The same shape against GRPO does not raise:

```python

trainer = GRPOTrainer(
    model="trl-internal-testing/tiny-Qwen3MoeForCausalLM",
    args=GRPOConfig(output_dir="out", use_liger_kernel=True, report_to="none"),
    train_dataset=dataset,
)
assert trainer.aux_loss_enabled  # True, yet the term never reaches the loss
trainer.train()
assert not any("aux_loss" in log for log in trainer.state.log_history)
```

### What DPO and KTO do

```python
if self.aux_loss_enabled and self.use_liger_kernel:
    raise ValueError(
        "Liger DPO loss does not support the Mixture-of-Experts load-balancing auxiliary loss, because it "
        "fuses the loss without materializing the router logits. Either set `router_aux_loss_coef` to `0.0` "
        "to disable the auxiliary loss, or set `use_liger_kernel` to False."
    )
```

`trl/trainer/dpo_trainer.py:911` and `trl/trainer/kto_trainer.py:914`, each covered by
`test_init_fails_with_moe_aux_loss_and_liger`. The reason in that message is exactly why the GRPO
fused loss cannot carry the term either. GRPO has neither the guard nor the test.

GRPO does already fail loudly for four other Liger gaps: the entropy bonus, a PEFT adapter on
`lm_head`, prompt-learning PEFT, and entropy-quantile masking. The auxiliary loss is the one that was
missed.

### Scope

Affected, because they inherit `compute_loss` and its Liger fork:

- `GRPOTrainer`
- `trl/experimental/gmpo` and `trl/experimental/gspo_token`, which override only `_compute_loss`
- `trl/experimental/grpo_with_replay_buffer`, which overrides no loss method at all

Not affected:

- `SFTTrainer` sets `output_router_logits` on the config and passes it as a forward kwarg
  independently of its Liger block, so the model still returns the auxiliary loss
- async GRPO rejects `use_liger_kernel` outright
- `RLOOTrainer` has no Liger path
- `MiniLLMTrainer` replaces `compute_loss` entirely and never reaches the fork

Separately, `gmpo` and `gspo_token` contain no reference to `aux_loss` at all, so their
`_compute_loss` drops the term even without Liger. That has a different root cause than this issue,
the same one as #7005: a self-contained `_compute_loss` override that never calls `super()` and so
never reaches the parent's tail block. I have added it there rather than here, since #7005 already
covers those trainers.

### Suggested fix

Add the DPO and KTO guard to the GRPO constructor, next to the four Liger checks already there, with
the matching test. If a silent drop is preferred over a hard failure for backward compatibility, a
`warning_once` at construction would at least make it visible.

I am happy to open the PR.

LOGO127:
# What does this PR do?

`GRPOTrainer` now fails fast when a Mixture-of-Experts model enables the load-balancing auxiliary loss together with the Liger fused GRPO loss.

The Liger GRPO path does not materialize router logits, so the auxiliary loss cannot be included in the training objective. Before this change, the run proceeded without the auxiliary term and without a warning.

The guard matches the existing DPO/KTO behavior. The `GRPOConfig` documentation now explains the required `router_aux_loss_coef=0.0` setting, and a focused initialization regression test covers the incompatibility.

Fixes #7009

## Before submitting

- [x] Was this discussed/approved via a GitHub issue? This change follows the reproducer and proposed direction in [#7009](https://github.com/huggingface/trl/issues/7009).
- [x] Did you make sure to update the documentation with your changes? The `GRPOConfig` parameter documentation now describes the Liger restriction.
- [x] Did you write any new necessary tests? Added `test_init_fails_with_moe_aux_loss_and_liger`.

## AI writing disclosure

- [x] AI-generated: the PR was mostly or fully generated by an AI tool. The changed lines and validation results were reviewed before submission.

## Who can review?

Anyone in the community is welcome to review this focused GRPO/Liger compatibility change once the tests have passed.

## Tests

- `uvx ruff check trl/trainer/grpo_trainer.py trl/trainer/grpo_config.py tests/test_grpo_trainer.py` — passed
- `uvx ruff format --check trl/trainer/grpo_trainer.py trl/trainer/grpo_config.py tests/test_grpo_trainer.py` — passed
- `git diff --check` — passed
- `pytest tests/test_grpo_trainer.py::TestGRPOTrainer::test_init_fails_with_moe_aux_loss_and_liger` — skipped locally because `liger-kernel` is not installed
- Manual construction with `trl-internal-testing/tiny-Qwen3MoeForCausalLM` — confirmed the expected `ValueError` before the Liger dependency check

---

> [!NOTE]
> **Low Risk**
> Initialization-only validation and documentation; no change to training loops unless the incompatible config was previously used silently.
> 
> **Overview**
> **GRPO** now raises at trainer init if a MoE model would use the default load-balancing auxiliary loss (`router_aux_loss_coef` non-zero) together with **`use_liger_kernel=True`**, instead of training without that term and no warning. The error message tells users to set **`router_aux_loss_coef=0.0`** or turn off the Liger kernel, matching the existing DPO/KTO guards.
> 
> **`GRPOConfig`** docs for **`router_aux_loss_coef`** note that the auxiliary loss is unsupported with the Liger kernel. A new **`test_init_fails_with_moe_aux_loss_and_liger`** test (behind **`@require_liger_kernel`**) asserts the **`ValueError`** on a tiny MoE model.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 232b24647c2844999d11abd8b7d9c2dd79d0f202. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -935,6 +935,25 @@ class TestGRPOTrainer(TrlTestCase):
                 peft_config=lora_config,
             )
 
+    @require_liger_kernel
+    def test_init_fails_with_moe_aux_loss_and_liger(self):
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        # The MoE auxiliary loss is on by default; it is incompatible with the Liger fused loss.
+        training_args = GRPOConfig(
+            output_dir=self.tmp_dir,
+            use_liger_kernel=True,
+            report_to="none",
+        )
+
+        with pytest.raises(ValueError, match="does not support the Mixture-of-Experts load-balancing auxiliary loss"):
+            GRPOTrainer(
+                model="trl-internal-testing/tiny-Qwen3MoeForCausalLM",
+                reward_funcs="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
+                args=training_args,
+                train_dataset=dataset,
+            )
+
     @require_peft
     def test_liger_kernel_with_peft_non_lm_head_target_allowed(self):
         # The lm_head guard must only fire when the adapter actually wraps lm_head. An adapter that targets other
--- a/trl/trainer/grpo_config.py
+++ b/trl/trainer/grpo_config.py
@@ -45,7 +45,8 @@ class GRPOConfig(_BaseConfig):
         router_aux_loss_coef (`float`, *optional*, defaults to `0.001`):
             Coefficient of the load-balancing auxiliary loss. Only has an effect when training a Mixture-of-Experts
             (MoE) model; for other models it does nothing. The auxiliary loss is added to the training loss with this
-            weight. Set to `0.0` to disable it.
+            weight. Set to `0.0` to disable it. The auxiliary loss is not supported with the Liger kernel; set this
+            value to `0.0` when `use_liger_kernel=True`.
         disable_dropout (`bool`, *optional*, defaults to `False`):
             Whether to disable dropout in the model. This is useful for training with a reference model, as it prevents
             the model from generating different logprobs for the same input.
@@ -444,7 +445,8 @@ class GRPOConfig(_BaseConfig):
         metadata={
             "help": "Coefficient of the load-balancing auxiliary loss. Only has an effect when training a "
             "Mixture-of-Experts (MoE) model; for other models it does nothing. The auxiliary loss is added to the "
-            "training loss with this weight. Set to `0.0` to disable it."
+            "training loss with this weight. Set to `0.0` to disable it. The auxiliary loss is not supported when "
+            "`use_liger_kernel=True`."
         },
     )
     disable_dropout: bool = field(
--- a/trl/trainer/grpo_trainer.py
+++ b/trl/trainer/grpo_trainer.py
@@ -788,6 +788,12 @@ class GRPOTrainer(_BaseTrainer):
         is_moe = getattr(text_config, "output_router_logits", None) is not None
         self.aux_loss_enabled = is_moe and args.router_aux_loss_coef != 0.0
         self.router_aux_loss_coef = args.router_aux_loss_coef
+        if self.aux_loss_enabled and self.use_liger_kernel:
+            raise ValueError(
+                "Liger GRPO loss does not support the Mixture-of-Experts load-balancing auxiliary loss, because it "
+                "fuses the loss without materializing the router logits. Either set `router_aux_loss_coef` to `0.0` "
+                "to disable the auxiliary loss, or set `use_liger_kernel` to False."
+            )
         self.scale_rewards = args.scale_rewards
         self.importance_sampling_level = args.importance_sampling_level
         self.off_policy_mask_threshold = args.off_policy_mask_threshold


LOGO127:
Maintainers, the guard is intentionally fail-fast and follows the existing DPO/KTO pattern: when `router_aux_loss_coef > 0` with `use_liger_kernel=True`, GRPO raises before trainer setup because the fused Liger loss cannot materialize router logits. The config docs now state the limitation and show `router_aux_loss_coef=0.0` as the compatible setting.

The required PR-template check and Cursor Bugbot review pass. Locally, Ruff and diff checks pass; the focused test is skipped only when Liger is unavailable, and a manual tiny-MoE constructor check verifies the expected error. Happy to adjust the wording or add a fixture if preferred.

PR #7020, opened 2026-09-02
Remove PPOTrainer

qgallouedec:
Removes PPO entirely: `PPOTrainer`, `PPOConfig`, and `modeling_value_head.py` (`PreTrainedModelWrapper`, `AutoModelForCausalLMWithValueHead`, `AutoModelForSeq2SeqLMWithValueHead`), plus its tests, doc page, and the examples

Bit of a moment for TRL: `PPOTrainer` predates every PR in this repo. It landed in dfb6a580 on 2020-03-28, the commit that first added the library, back when the package was called `lm_ppo`. It is the oldest thing in TRL and the last piece of the original codebase.

Why now?:

- It's unmaintained. No feature work in over a year. Every PPO-specific commit since #4482 is a drive-by fix: #6121, #6202, #5151, #5245, #4945.
- It's the only trainer we never aligned on the input format. Every other trainer takes text or conversations; PPO still takes tokenized `input_ids`, which is why it sat alone in the dataset format table as "Tokenized language modeling".
- We record basically no usage.
- It's a magnet for automated bug hunters. The reports are real but they land on code nobody uses, and triaging them costs maintainer time.
- Nobody loses anything. `from trl import PPOTrainer` already stopped working in v1.10 (#4482 moved it to `trl.experimental.ppo`), so anyone actually running PPO is pinned to an old TRL: ms-swift is on `trl>=0.15,<1.0`, verl on `trl<=0.9.6`, and the older tutorials sit around `trl==0.7.1`. Those installs keep resolving exactly as they do today.

Also cleaned up in the same pass:

- `create_reference_model` stays (BCO, A2PO, Online DPO use it), so its tests move from `tests/experimental/test_modeling_value_head.py` into `tests/experimental/test_utils.py`, retargeted at a plain `AutoModelForCausalLM`.
- `require_torch_gpu_if_bnb_not_multi_backend_enabled` and `is_bitsandbytes_multi_backend_available` in `tests/testing_utils.py` had no caller left.

---

> [!NOTE]
> **High Risk**
> This is a breaking removal of a long-standing public experimental API and ~2k+ lines of RL training code; anyone still importing `trl.experimental.ppo` will fail at import time despite prior deprecation to experimental.
> 
> **Overview**
> **Removes PPO from TRL** — the experimental `trl.experimental.ppo` package (`PPOTrainer`, `PPOConfig`, value-head wrappers), the `ppo_sentiment` and `ppo_tldr` examples, and the large PPO test suite.
> 
[...]

PR #7041, opened 2026-09-04
Pool PPO micro-batch statistics by token and sample counts

### LINKED ISSUE #7012: PPOTrainer statistics average per-micro-batch means without weighting by micro-batch size
### Summary

`PPOTrainer` reduces its per-micro-batch statistics with an unweighted mean over slots. Each slot holds a `masked_mean` of its own micro-batch, so when the last micro-batch of a minibatch is short, it carries the same weight as a full one and the reported metric is not the token or sample mean it is taken to be.

### Where

`trl/experimental/ppo/ppo_trainer.py`. The micro-batch loop stores one masked mean per slot, and the reduction at 897-904 averages those slot means:

```python
metrics["policy/approxkl_avg"]  = self.accelerator.gather_for_metrics(approxkl_stats).mean().item()
metrics["policy/clipfrac_avg"]  = self.accelerator.gather_for_metrics(pg_clipfrac_stats).mean().item()
metrics["loss/policy_avg"]      = self.accelerator.gather_for_metrics(pg_loss_stats).mean().item()
metrics["loss/value_avg"]       = self.accelerator.gather_for_metrics(vf_loss_stats).mean().item()
metrics["val/clipfrac_avg"]     = self.accelerator.gather_for_metrics(vf_clipfrac_stats).mean().item()
metrics["policy/entropy_avg"]   = self.accelerator.gather_for_metrics(entropy_stats).mean().item()
metrics["val/ratio"]            = self.accelerator.gather_for_metrics(ratio_stats).mean().item()
```

A mean of per-group means equals the pooled mean only when the groups are equal in size.

### When the micro-batches are uneven

`exact_div` requires `num_mini_batches` to divide `per_device_train_batch_size * gradient_accumulation_steps`. It does not require `per_device_train_batch_size` to divide `local_mini_batch_size`, so a short final micro-batch is reachable from a valid config:

```python
from trl.experimental.ppo import PPOConfig

args = PPOConfig(
    output_dir="/tmp/ppo-uneven", per_device_train_batch_size=4,
    gradient_accumulation_steps=3, num_mini_batches=2, num_ppo_epochs=1,
    use_cpu=True, bf16=False, report_to="none",
)
local_batch = args.per_device_train_batch_size * args.gradient_accumulation_steps
local_mini_batch = local_batch // args.num_mini_batches
sizes = [min(args.per_device_train_batch_size, local_mini_batch - s)
         for s in range(0, local_mini_batch, args.per_device_train_batch_size)]
print(local_batch, local_mini_batch, sizes)
```

Output: `12 6 [4, 2]`. The config is accepted, and the two micro-batches of each minibatch hold 4 and 2 samples.

### Size of the error

With slot values 1.0 and 3.0 over micro-batches of 4 and 2 samples, computed with exact rationals:

| quantity | value |
| --- | --- |
| what the code reports, `mean(1, 3)` | `2` |
| size-weighted mean, `(4*1 + 2*3) / 6` | `5/3` |
| relative error | `1/5`, that is 20% |

The error is exactly zero when every micro-batch is full, which is why the default `num_mini_batches=1` never shows it.

### Not the same as the zero-slot bug

`stats_shape` used to size its last axis by `gradient_accumulation_steps` while the loop writes `ceil(local_mini_batch_size / per_device_train_batch_size)` slots, leaving unwritten zeros in the average. #6861 corrects that axis. The weighting described here is independent: it is present both before and after that change, since both versions take an unweighted mean over whatever slots exist.

### Possible direction

Store a numerator and an active count per slot and reduce as `sum / count` for the token-level metrics, and weight the sample-level metrics by the real micro-batch size. `val/ratio_var` needs its own decision, since a variance over slot means is not a variance over samples either.

behroozazarkhalili:
Fixes #7012.

### What

`PPOTrainer` reduced its per-update statistics with an unweighted mean over micro-batch slots. Each slot holds a `masked_mean` (or plain mean) of its own micro-batch, so with a short final micro-batch the reported number was a mean of means, not the token or sample mean it is documented as. Unwritten slots, when `stats_shape` is larger than the loop writes, also contributed zeros.

The seven means now store weighted numerators and reduce as total numerator over total count:

- `loss/policy_avg`, `policy/clipfrac_avg` weight by the active response-token count of the micro-batch.
- `loss/value_avg`, `val/clipfrac_avg` weight by the active value-token count (`padding_mask_p1`).
- `policy/approxkl_avg`, `policy/entropy_avg`, `val/ratio` are per-token statistics: each slot now stores a `masked_mean` over the response tokens times the response-token count, and the reduction divides by the gathered response-token count. Before, the plain `.mean()` in those three slots included pad positions, and the first version of this PR only reweighted that mean by micro-batch size (corrected in 42c71bff).
- `val/ratio_var` keeps its variance over slot means, on its own buffer, because the issue leaves that population undefined, but now only over the slots the loop wrote: the response-token count buffer marks them, and the unwritten zeros used to inflate it (with `learning_rate=0` under a 3-over-2 split every ratio is 1 and the logged variance was 0.2667 instead of 0).

### Test

`test_uneven_micro_batch_stats_are_pooled` runs one CPU update with `per_device_train_batch_size=4, gradient_accumulation_steps=3, num_mini_batches=2`, which splits each minibatch into micro-batches of 4 and 2 samples. It spies on the module-level `masked_mean` and `selective_log_softmax` to record the live tensors entering each statistic, pools them independently, and compares with the logged metrics.

### Verification

On a compute node (job 58032838):

- fixed tree: the test passes, and the full `tests/experimental/test_ppo_trainer.py` gives 57 passed, 2 skipped;
- base `ppo_trainer.py`: the test fails at the first pooled comparison;
- mutant dividing `loss/policy_avg` by the sample count instead of the token count: fails;
- mutant with the entropy weight off by one: fails;
- ruff 0.13.3 check and format: clean.

`test_ratio_var_ignores_unwritten_stats` trains with `learning_rate=0`, `gradient_accumulation_steps=3` and `num_mini_batches=2` and asserts `val/ratio_var` is 0; on the previous commit it reads 0.2667.

### Limits

- Multi-process aggregation was not run here. The count and numerator buffers go through the same `gather_for_metrics` path as before.
- This overlaps with #6861, which changes the same stat buffers (per-update allocation, NaN for unwritten slots). Whichever merges second needs a small conflict resolution; the count-based reduction here makes the unwritten-slot handling unnecessary for the seven means.

---

> [!NOTE]
> **Medium Risk**
> Changes how PPO training metrics are reduced across micro-batches, which affects monitoring and any logic keyed on those logs; training loss/backprop paths are unchanged but reported KL/entropy/ratio values will differ when batching is uneven.
> 
> **Overview**
> **PPO logged averages** no longer take an unweighted mean over micro-batch slots (which skewed uneven 4 vs 2 splits and let empty slots pull metrics toward zero). Each step now accumulates **weighted numerators**—policy-side stats (`loss/policy_avg`, clip fractions, approx KL, entropy, `val/ratio`) scale by active **response-token** count; value-side stats use **`padding_mask_p1`** counts—and divides by the gathered totals after `gather_for_metrics`.
> 
> **Per-micro-batch computation** also aligns with masking: approximate KL and entropy use `masked_mean` on the response mask instead of raw `.mean()`, and ratio logging uses masked means.
> 
> **`val/ratio_var`** is still variance across per-slot mean ratios, but only slots with `padding_count > 0`, so unwritten buffer zeros no longer inflate variance (e.g. LR=0 case).
> 
> **Tests** add `test_uneven_micro_batch_stats_are_pooled` (monkeypatched `masked_mean` / `selective_log_softmax` vs logged metrics) and `test_ratio_var_ignores_unwritten_stats`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42c71bff9ee2485b88db2160ec56c03554165796. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/experimental/test_ppo_trainer.py
+++ b/tests/experimental/test_ppo_trainer.py
@@ -767,7 +767,7 @@ def test_uneven_micro_batch_stats_are_pooled(tmp_path, monkeypatch):
 
     trainer.train()
 
-    assert [values.shape[0] for values, _ in masked_inputs] == [4, 4, 4, 4, 2, 2, 2, 2] * 2
+    assert [values.shape[0] for values, _ in masked_inputs] == [4] * 7 + [2] * 7 + [4] * 7 + [2] * 7
     assert [values.shape[0] for values in logprobs_diffs] == [4, 2, 4, 2]
     assert [values.shape[0] for values in entropy_logits] == [4, 2, 4, 2]
 
@@ -777,18 +777,15 @@ def test_uneven_micro_batch_stats_are_pooled(tmp_path, monkeypatch):
         return (numerator / denominator).item()
 
     metrics = trainer.state.log_history[-1]
-    logprobs_diff = torch.cat(logprobs_diffs)
-    logits = torch.cat(entropy_logits)
-    prob_dist = torch.nn.functional.softmax(logits, dim=-1)
-    entropy = torch.logsumexp(logits, dim=-1) - torch.sum(prob_dist * logits, dim=-1)
-    assert metrics["policy/approxkl_avg"] == pytest.approx(0.5 * logprobs_diff.square().mean().item(), abs=1e-7)
-    # masked_mean writes value loss, value clipfrac, policy loss, then policy clipfrac for every micro-batch.
-    assert metrics["loss/value_avg"] == pytest.approx(0.5 * pooled_mean(masked_inputs[0::4]), abs=1e-7)
-    assert metrics["val/clipfrac_avg"] == pytest.approx(pooled_mean(masked_inputs[1::4]), abs=1e-7)
-    assert metrics["loss/policy_avg"] == pytest.approx(pooled_mean(masked_inputs[2::4]), abs=1e-7)
-    assert metrics["policy/clipfrac_avg"] == pytest.approx(pooled_mean(masked_inputs[3::4]), abs=1e-7)
-    assert metrics["policy/entropy_avg"] == pytest.approx(entropy.mean().item(), abs=1e-7)
-    assert metrics["val/ratio"] == pytest.approx(logprobs_diff.exp().mean().item(), abs=1e-6)
+    # masked_mean writes value loss, value clipfrac, policy loss, policy clipfrac, approximate KL, entropy, then
+    # ratio for every micro-batch.
+    assert metrics["loss/value_avg"] == pytest.approx(0.5 * pooled_mean(masked_inputs[0::7]), abs=1e-7)
+    assert metrics["val/clipfrac_avg"] == pytest.approx(pooled_mean(masked_inputs[1::7]), abs=1e-7)
+    assert metrics["loss/policy_avg"] == pytest.approx(pooled_mean(masked_inputs[2::7]), abs=1e-7)
+    assert metrics["policy/clipfrac_avg"] == pytest.approx(pooled_mean(masked_inputs[3::7]), abs=1e-7)
+    assert metrics["policy/approxkl_avg"] == pytest.approx(0.5 * pooled_mean(masked_inputs[4::7]), abs=1e-7)
+    assert metrics["policy/entropy_avg"] == pytest.approx(pooled_mean(masked_inputs[5::7]), abs=1e-7)
+    assert metrics["val/ratio"] == pytest.approx(pooled_mean(masked_inputs[6::7]), abs=1e-6)
 
 
 def test_ratio_var_ignores_unwritten_stats(tmp_path):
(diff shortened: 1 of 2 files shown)


behroozazarkhalili:
Closing: `PPOTrainer` was removed from the tree in #7020 (main 700b845c), so the stats-buffer weighting fix here has no target. Nothing else in this PR applied outside `trl/experimental/ppo`. Issue #7012 is closed for the same reason.

PR #7042, opened 2026-09-04
Log GRPO policy_loss at the optimizer-window scale for every loss type

behroozazarkhalili:
GRPO logged `policy_loss` at different scales depending on the loss type. I restored the optimizer window scale for CISPO, DAPO, and VESPO during training while leaving the returned loss unchanged. The regression tests independently compute the objective and check both values, failing on the base code and passing with the fix.
Fixes #7011.

---

> [!NOTE]
> **Low Risk**
> Metrics-only adjustment with no change to the backward loss; limited to GRPO logging for three loss types plus a unit test.
> 
> **Overview**
> **Fixes inconsistent `policy_loss` metrics** for `cispo`, `dapo`, and `vespo` in `GRPOTrainer._compute_loss`. Those branches folded gradient accumulation and `steps_per_generation` into the loss normalizer before recording `policy_loss`, so logs reflected a single micro-batch contribution while other loss types logged the pre–gradient-accumulation (optimizer-window) scale.
> 
> The change **multiplies the detached `policy_loss` by `current_gradient_accumulation_steps` in train mode** (1.0 in eval) after that normalization. **Returned loss and optimization are unchanged**—only what gets appended to `_metrics["train"]["policy_loss"]` is rescaled.
> 
> Adds **`test_policy_loss_uses_optimizer_window_scale`**, parametrized over the three loss types, with hand-built inputs and explicit expectations for both `_compute_loss` output and the logged `policy_loss` under `gradient_accumulation_steps=2` and `steps_per_generation=4`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd5fb1f492766632475fafe191eaa5ad2cbbb6a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #7053, opened 2026-09-04
Log GRPO loss-path metrics once per optimizer window

### LINKED ISSUE #7047: GRPO loss-path metrics average micro-batch entries, so uneven accumulation windows are weighted by their size
### Summary

`GRPOTrainer` appends its loss-path metrics (`policy_loss`, `entropy`, and every metric scaled by `current_gradient_accumulation_steps` in `_compute_loss`) once per micro-batch, and `log()` takes a flat mean over the entries of the logging interval. When the interval contains accumulation windows of different sizes, that mean weights each window by its micro-batch count instead of counting each optimizer step once. The logged value is then a different number from the optimizer-window average, while the base `loss` reported by transformers is unaffected.

### Where the uneven windows come from

transformers shortens the last window of an epoch to the remainder and sets `current_gradient_accumulation_steps` to that size (`transformers/trainer.py`, 5.11.0):

```python
remainder = steps_in_epoch % self.args.gradient_accumulation_steps
if remainder == 0:
    remainder = self.args.gradient_accumulation_steps
...
num_batches = (
    self.args.gradient_accumulation_steps if update_step != (num_update_steps_per_epoch - 1) else remainder
)
...
self.current_gradient_accumulation_steps = len(batch_samples)
```

So with `gradient_accumulation_steps=3` and an epoch of four micro-batches the windows are three micro-batches then one. Each micro-batch entry in `self._metrics["train"]` is already scaled to its own window (`accumulation_factor = self.current_gradient_accumulation_steps`), so the entries carry the right values. The bias comes from the count: the long window contributes three entries and the short one contributes one, and `log()` divides by four.

### Measured on main (5dd51e4d, transformers 5.11.0, CPU)

`trl-internal-testing/tiny-Qwen2ForCausalLM-2.5`, four prompts from `trl-internal-testing/zen`, `per_device_train_batch_size=2`, `gradient_accumulation_steps=3`, `steps_per_generation=2`, `num_generations=2`, `max_completion_length=4`, `entropy_coef=0.1`, `logging_strategy="epoch"`, a deterministic reward. The instrumented trainer records each micro-batch's `current_gradient_accumulation_steps` and the entry appended to `_metrics`.

| micro-batch | window size | `policy_loss` entry |
| --- | --- | --- |
| 1 | 3 | 0.0 |
| 2 | 3 | 0.0 |
| 3 | 3 | 0.0 |
| 4 | 1 | -0.0022285581 |

| aggregation | `policy_loss` |
| --- | --- |
| logged (flat mean over 4 entries) | -0.00055713951587677 |
| mean over the 2 optimizer windows | -0.00111427903175354 |

The logged value is half the per-window mean, which is exactly the three-to-one entry weighting. `entropy` is appended unscaled (`global_masked_mean(entropies)`), so it only sees the count weighting, not the window-size scaling: 11.929378747940063 logged versus 11.929380893707275 per window, a gap that is small because the four micro-batch entries are nearly identical. The base `loss` does not have the problem: transformers sums the scaled micro-batch losses and divides by the optimizer-step count, and its logged value matches the per-window mean to float rounding (-1.194052338600159 versus -1.194052368402481).

### Scope

Every entry on this path has the property: the normalizer branches in `_compute_loss` (around lines 3244 to 3269 on main), the entropy accumulation factor (around 3287), and the `policy_loss` capture. GMPO, MiniLLM and the replay-buffer trainer inherit `_compute_loss`, so they log the same numbers. #7042 corrects the scale of one `policy_loss` branch to match the others and deliberately leaves the aggregation alone, so this issue is the follow-up.

### Suggested fix

Accumulate each metric's micro-batch contributions into a per-window sum and append to `self._metrics` once, at the window boundary (`self._step % self.current_gradient_accumulation_steps == 0` already exists in `training_step`), so the entry count equals the optimizer-step count and `log()`'s mean is exact. Training is untouched; only the logged values change, and only when an interval mixes window sizes.

I can open the PR.

behroozazarkhalili:
GRPO gave longer gradient accumulation windows more weight when averaging loss metrics. I changed GRPO and its GMPO and GSPO-token copies to record one value per optimizer window, and tests for equal and uneven windows pass. This overlaps #7042 on the same lines and will need a small conflict resolution.
Fixes #7047.

---

> [!NOTE]
> **Low Risk**
> Changes only how training metrics are aggregated for logging; optimization math is explicitly unchanged. Clip-ratio and other per-micro-batch diagnostics are untouched.
> 
> **Overview**
> Fixes incorrect **logged** `policy_loss`, `entropy`, `kl`, and `aux_loss` when gradient accumulation uses uneven windows (e.g. a shortened last window), while leaving the actual training loss and gradients unchanged.
> 
> In train mode, micro-batch metric values are now scaled by `1 / current_gradient_accumulation_steps`, summed in `_loss_window_metrics`, and appended to `_metrics` only when `accelerator.sync_gradients` is true—so `log()`’s mean is one value per optimizer step with equal weight per window. Eval still logs per batch. `train()` clears the accumulator at the start to avoid stale partial sums after an interrupted run. The same logging block is applied in the GMPO and GSPO-token GRPO trainer copies.
> 
> Adds `test_policy_loss_is_averaged_over_accumulation_windows` to assert window-weighted averaging vs the old flat micro-batch mean.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c3d4dc4616c3316e1bbc361b03ad86e9fa1ab9a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


behroozazarkhalili:

PR #7066, opened 2026-09-04
Add `log_entropy` to skip the per-token entropy metric in SFT, GRPO and RLOO

### LINKED ISSUE #4184: [Feature] Add a bool enable/disable flag for `entropy logging`
### Feature request

Please add an option to disable `entropy logging`

### Motivation

`entropy logging` Added in https://github.com/huggingface/trl/pull/3940
This is increased train time by %40

I think we should add a bool flag enable/disable it

behroozazarkhalili:
# What does this PR do?

Adds `log_entropy` (default `True`) to `SFTConfig`, `GRPOConfig` and `RLOOConfig` so the per-token entropy metric can be switched off.

The metric is computed on every forward pass today, and it is not free: on the CPU measurement in the issue thread, `entropy_from_logits` over a `(6, 256, 151936)` logits tensor took 2342 ms against 1090 ms for `selective_log_softmax`, and fusing the two passes did not help because both are bandwidth-bound on the logits. The only way to avoid the cost is to not compute it.

With `log_entropy=False`:

- `SFTTrainer` skips `entropy_from_logits` on the `nll` path and skips the per-chunk entropy reduction on the `chunked_nll` path (`_chunked_cross_entropy_loss` and `_patch_chunked_ce_lm_head` take a `compute_entropy` argument; the entropy sum is `None` when disabled). `mean_token_accuracy` is unchanged.
- `GRPOTrainer`, `RLOOTrainer`, and the GMPO and GSPO-token copies of `_compute_loss` request entropies only when the loss needs them (`top_entropy_quantile < 1.0` or an active entropy bonus in GRPO; never in RLOO, which has no such option) and do not append the `entropy` metric. The four GRPO-family blocks are identical.
- DPO, KTO and TPO are out of scope; they compute entropy at their own loss sites and are left as they are.

Tests: `test_log_entropy` in `test_sft_trainer.py` (over `nll` and `chunked_nll`), `test_grpo_trainer.py` and `test_rloo_trainer.py` assert the `entropy` key is absent from every `log_history` entry when the flag is off and present by default. A mutant that appends the SFT metric unconditionally fails exactly the `nll-False` row. Docs for the three trainers name the flag next to the `entropy` metric.

This is shape 1 of the two offered on the issue on 2026-08-05 (a config flag, as the issue title asks). Shape 2 (derive the metric from need and drop it by default) would remove the metric for everyone, which loses a useful entropy-collapse signal, so the flag keeps the default behaviour unchanged.

Fixes #4184

## Before submitting

## AI writing disclosure

## Who can review?

@qgallouedec @albertvillanova

---

> [!NOTE]
> **Low Risk**
> Opt-in performance optimization with default behavior unchanged; training paths that require entropy for regularization or masking still compute it.
> 
> **Overview**
> Adds **`log_entropy`** (default `True`) on `SFTConfig`, `GRPOConfig`, and `RLOOConfig` so training can avoid the expensive per-token entropy pass when you only care about loss and other metrics.
> 
> **SFT:** With `log_entropy=False`, the trainer skips `entropy_from_logits` on the `nll` path and passes `compute_entropy=False` through chunked CE (`_chunked_cross_entropy_loss` / `_patch_chunked_ce_lm_head`). `mean_token_accuracy` is unchanged.
> 
> **GRPO / RLOO (and GMPO / GSPO-token GRPO variants):** Entropy is still computed when the loss needs it (`top_entropy_quantile`, entropy bonus on GRPO); otherwise it follows `log_entropy`. The logged `entropy` metric is only recorded when `log_entropy` is true.
> 
> Docs for the three trainers note the flag next to the `entropy` metric; new parametrized tests assert presence/absence of `entropy` in `log_history`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1fd1cf9336dabae21a693f80dac5a14138ecfb31. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/source/grpo_trainer.md
+++ b/docs/source/grpo_trainer.md
@@ -193,7 +193,7 @@ While training and evaluating, we record the following metrics:
 - `sampling/importance_sampling_ratio/mean`: The average constrained importance sampling ratio. Logged only when `use_vllm=True` and `vllm_importance_sampling_correction=True`.
 - `sampling/importance_sampling_ratio/max`: The largest constrained importance sampling ratio. Logged only when `use_vllm=True` and `vllm_importance_sampling_correction=True`.
 - `policy_loss`: The policy gradient loss value (before any entropy bonus). Logged when `entropy_coef` is nonzero or `use_adaptive_entropy=True`.
-- `entropy`: Average entropy of token predictions across generated completions. (If `mask_truncated_completions=True`, masked sequences tokens are excluded.)
+- `entropy`: Average entropy of token predictions across generated completions. (If `mask_truncated_completions=True`, masked sequences tokens are excluded.) Set `log_entropy=False` to skip computing it.
 - `entropy_coef`: The current entropy regularization coefficient. Logged when `entropy_coef` is nonzero or `use_adaptive_entropy=True`. Updated once per optimizer step when `use_adaptive_entropy=True`.
 - `aux_loss`: The load-balancing auxiliary loss of a Mixture-of-Experts model, before it is scaled by `router_aux_loss_coef` and added to the loss. Logged only when the model is a MoE model and `router_aux_loss_coef` is nonzero.
 - `kl`: The average KL divergence between the model and the reference model, calculated over generated completions. Logged only if `beta` is nonzero.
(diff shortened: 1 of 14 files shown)


qgallouedec:
closing see https://github.com/huggingface/trl/issues/4184#issuecomment-5546295458

PR #7088, opened 2026-09-07
Group Dependabot GitHub Actions and pre-commit updates in a single PR

albertvillanova:
This PR makes Dependabot open a single pull request for all our CI pins, instead of one per package ecosystem.

### Motivation

`doc-builder` is pinned twice: as reusable workflows in `.github/workflows` and as the hook `rev` in `.pre-commit-config.yaml`. Since these live in two different Dependabot ecosystems (`github-actions` and `pre-commit`), each bump arrived as two separate pull requests that had to be merged together, as in #7078 and #7079.

### Solution

Dependabot multi-ecosystem groups combine updates from several ecosystems into one pull request. The three update entries (workflow actions, composite action actions, and pre-commit hooks) now belong to a single group, so both `doc-builder` pins always move together.

The per-entry `schedule` keys are dropped because the group provides the schedule. The `cooldown` and `ignore` settings stay per entry, as they are not supported at group level.

### Changes

- Add a `github-actions-and-pre-commit` multi-ecosystem group with a weekly schedule
- Assign the `github-actions` and `pre-commit` update entries to that group
- Remove the now redundant `actions` and `local-actions` single-ecosystem groups

---

> [!NOTE]
> **Low Risk**
> Only `.github/dependabot.yml` changes how update PRs are grouped; no application or workflow logic is modified.
> 
> **Overview**
> Dependabot will now open **one weekly PR** for CI dependency bumps instead of separate PRs per ecosystem.
> 
> The config adds a `multi-ecosystem-groups` entry (`github-actions-and-pre-commit`) and wires all three update blocks—root workflows, `/.github/actions/*`, and pre-commit—to that group. Per-entry **weekly `schedule`** keys and the old single-ecosystem **`groups`** (`actions`, `local-actions`) are removed in favor of top-level **`patterns`** and **`multi-ecosystem-group`**. **Cooldown** and the **Ruff pre-commit ignore** are unchanged on each entry.
> 
> This keeps paired pins (e.g. `doc-builder` in workflows and `.pre-commit-config.yaml`) on the same bump.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef73d9633aaf31ca3381b5b2b56af674aa8bcfe7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #7091, opened 2026-09-07
Merge the two Dependabot github-actions entries into one

albertvillanova:
This PR merges the two Dependabot `github-actions` update entries into a single one.

### Motivation

The two entries were kept separate in #6897 only to avoid touching the entry that was already working, since a malformed `dependabot.yml` stops all updates for the repository without reporting anything. They have carried identical `cooldown`, `patterns` and group settings ever since, and every policy change has to be applied twice.

Since #7088 both entries feed the same pull request, as confirmed by #7089, so the split no longer has any observable effect.

### Solution

Keep one `github-actions` entry listing both locations in `directories`: `/` for the workflows and the root `action.yml`, and the `/.github/actions/*` glob for the actions pinned inside our local composite actions.

### Changes

- Merge the `/.github/actions/*` entry into the `/` entry, listing both under `directories`
- Drop the duplicated `cooldown`, `patterns` and `multi-ecosystem-group` settings

---

> [!NOTE]
> **Low Risk**
> Dependabot config-only change with no application code impact; main risk is a YAML mistake that could silently stop updates, but the merged layout matches the prior dual-entry coverage.
> 
> **Overview**
> Consolidates the two identical **Dependabot `github-actions`** update blocks into a **single** entry so cooldown, patterns, and `multi-ecosystem-group` only need to be maintained once.
> 
> That entry now uses **`directories`** with **`/`** (workflows and root `action.yml`) and **`/.github/actions/*`** (pins inside local composite actions), replacing the previous split where each path had its own block. Behavior should stay the same now that both paths already shared the same grouped PR policy.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c218cac2e16cb3181308dc8602d6596cc544b0c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,17 +7,10 @@ multi-ecosystem-groups:
       interval: "weekly"
 
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
-    cooldown:
-      default-days: 7
-    patterns: ["*"]
-    multi-ecosystem-group: "github-actions-and-pre-commit"
-
-  # Actions pinned inside our local composite actions. The "/" directory above
-  # does not cover them: it only scans .github/workflows and the root action.yml.
+  # "/" only covers .github/workflows and the root action.yml, hence the composite actions glob.
   - package-ecosystem: "github-actions"
     directories:
+      - "/"
       - "/.github/actions/*"
     cooldown:
       default-days: 7


albertvillanova:
Merging this PR as a simplification of the Dependabot config for security updates.

Issue #7093, opened 2026-09-07
CI emits 462 `transformers` `Trainer` log warnings: "The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config"


albertvillanova:
### What

CI emits a `transformers` log warning from `Trainer.train()` on essentially every trainer test:

```
[transformers] The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config. The model config and generation config were aligned accordingly, being updated with the tokenizer's values. Updated tokens: {'eos_token_id': 151645, 'bos_token_id': None, 'pad_token_id': 151643}.
```

462 lines in a single job, across 22 distinct `Updated tokens: {...}` variants. It is by far the largest single source of warning output in the run (about 84% of all warning lines).

CI run: https://github.com/huggingface/trl/actions/runs/34088542293/job/101637190203

Top variants by count:

| Count | `Updated tokens` |
|---|---|
| 313 | `{'eos_token_id': 151645, 'bos_token_id': None, 'pad_token_id': 151643}` |
| 24 | `{'eos_token_id': 151645}` |
| 21 | `{'eos_token_id': 1, 'bos_token_id': 2, 'pad_token_id': 0}` |
| 14 | `{'eos_token_id': 200002, 'bos_token_id': 199998, 'pad_token_id': 199999}` |
| 13 | `{'eos_token_id': 151645, 'bos_token_id': None, 'pad_token_id': 151645}` |
| 12 | `{'bos_token_id': None, 'pad_token_id': 151643}` |
| 65 | 16 further variants |

Affected tests: the warning is emitted from `Trainer.train()`, so it fires in every test that trains, across `tests/test_sft_trainer.py`, `tests/test_dpo_trainer.py`, `tests/test_grpo_trainer.py`, `tests/test_rloo_trainer.py`, `tests/test_kto_trainer.py`, `tests/test_reward_trainer.py` and `tests/test_distillation_trainer.py`.

Note: unlike the warnings tracked in #6447, #6466, #6555, #6580, #6848, #6849 and #7018, this one is a `logging` record, not a `warnings` warning. It does not appear in the pytest warnings summary and cannot be silenced with a `filterwarnings` entry in `pyproject.toml`. It reaches the job log because `make test` passes `-s`.

The tests themselves pass; for now this is warning noise only.

### When it started

[...]

PR #7097, opened 2026-09-07
Sync the model config pad token id with the tokenizer in the trainers

albertvillanova:
This PR keeps `model.config` and `model.generation_config` in agreement with the tokenizer's pad token in the SFT, DPO, KTO, GRPO, RLOO and Distillation trainers.

### Motivation

When the tokenizer has no pad token, these trainers fall back to the eos token and set it on the tokenizer, but leave the model configs untouched. `Trainer.train()` then finds the divergence, rewrites both configs itself, and logs it as a change the user did not make:

```
The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config. The model config and generation config were aligned accordingly, being updated with the tokenizer's values. Updated tokens: {'pad_token_id': 2}.
```

Part of #7093.

### Solution

Mirror the pad token onto `model.config` and `model.generation_config` right where the trainers set it on the tokenizer. The alignment that `Trainer` performs at train time then finds nothing to change, so the end state is identical, while the model stays consistent with the tokenizer from the moment the trainer is built and for any code reading the config before `train()`.

`RewardTrainer` needs no change: it already sets `model.config.pad_token_id`, and a sequence classification model has no generation config.

### Changes

- Mirror the tokenizer's pad token onto the model config and generation config wherever the SFT, DPO, KTO, GRPO, RLOO and Distillation trainers assign it
- Add a regression test to each of those six trainers, using a model whose tokenizer has no pad token so the fallback is exercised

---

> [!NOTE]
> **Low Risk**
> Small init-time config alignment with unchanged training behavior; covered by new regression tests.
> 
> **Overview**
> When trainers assign a pad token on the tokenizer (including the common **eos fallback** when none is defined), they now also set **`model.config.pad_token_id`** and **`model.generation_config.pad_token_id`** to the same value. This applies to **SFT**, **DPO**, **KTO**, **GRPO**, **RLOO**, and **Distillation** trainers—**DPO** and **SFT** mirror again after resolving an explicit `args.pad_token` for the data collator.
> 
> The goal is to avoid Hugging Face **`Trainer.train()`** detecting a tokenizer/config mismatch and logging an automatic realignment the user did not request, while keeping configs consistent for any code that reads them before training starts.
> 
> Each of the six trainers gets a regression test (`test_pad_token_id_synced_with_model_config`) using a Mistral fixture whose tokenizer has no pad token.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8328b88892b94312b4ff7d84cb4ad488234e0c9d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -3948,13 +3948,7 @@ class TestGRPOTrainer(TrlTestCase):
         def reward_func(completions, **kwargs):
             return [0.0] * len(completions)
 
-        training_args = GRPOConfig(
-            output_dir=self.tmp_dir,
-            per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
-            num_generations=3,  # reduce the number of generations to reduce memory usage
-            max_completion_length=8,  # reduce the completion length to reduce memory usage
-            report_to="none",
-        )
+        training_args = GRPOConfig(output_dir=self.tmp_dir, report_to="none")
         trainer = GRPOTrainer(
             model="trl-internal-testing/tiny-MistralForCausalLM-0.2",
             reward_funcs=reward_func,
--- a/tests/test_grpo_trainer.py
+++ b/tests/test_grpo_trainer.py
@@ -3940,6 +3940,33 @@ class TestGRPOTrainer(TrlTestCase):
         assert len(trainer.reward_processing_classes) == 1
         assert trainer.reward_processing_classes[0] == single_processing_class
 
+    def test_pad_token_id_synced_with_model_config(self):
+        # This model's tokenizer has no pad token, so the trainer falls back to the eos token. The model configs must
+        # follow: otherwise `Trainer` realigns them at train time and reports it as a change the user did not make.
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        def reward_func(completions, **kwargs):
+            return [0.0] * len(completions)
+
+        training_args = GRPOConfig(
+            output_dir=self.tmp_dir,
+            per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
+            num_generations=3,  # reduce the number of generations to reduce memory usage
+            max_completion_length=8,  # reduce the completion length to reduce memory usage
+            report_to="none",
+        )
+        trainer = GRPOTrainer(
+            model="trl-internal-testing/tiny-MistralForCausalLM-0.2",
+            reward_funcs=reward_func,
+            args=training_args,
+            train_dataset=dataset,
+        )
+
+        pad_token_id = trainer.processing_class.pad_token_id
+        assert pad_token_id is not None
+        assert trainer.model.config.pad_token_id == pad_token_id
+        assert trainer.model.generation_config.pad_token_id == pad_token_id
+
 
 @require_vision
 class TestGRPOTrainerVLM(TrlTestCase):
--- a/tests/test_rloo_trainer.py
+++ b/tests/test_rloo_trainer.py
@@ -1854,13 +1854,7 @@ class TestRLOOTrainer(TrlTestCase):
         def reward_func(completions, **kwargs):
             return [0.0] * len(completions)
 
-        training_args = RLOOConfig(
-            output_dir=self.tmp_dir,
-            per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
-            num_generations=3,  # reduce the number of generations to reduce memory usage
-            max_completion_length=8,  # reduce the completion length to reduce memory usage
-            report_to="none",
-        )
+        training_args = RLOOConfig(output_dir=self.tmp_dir, report_to="none")
         trainer = RLOOTrainer(
             model="trl-internal-testing/tiny-MistralForCausalLM-0.2",
             reward_funcs=reward_func,
--- a/tests/test_rloo_trainer.py
+++ b/tests/test_rloo_trainer.py
@@ -1846,6 +1846,33 @@ class TestRLOOTrainer(TrlTestCase):
         assert len(trainer.reward_processing_classes) == 1
         assert trainer.reward_processing_classes[0] == single_processing_class
 
+    def test_pad_token_id_synced_with_model_config(self):
+        # This model's tokenizer has no pad token, so the trainer falls back to the eos token. The model configs must
+        # follow: otherwise `Trainer` realigns them at train time and reports it as a change the user did not make.
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        def reward_func(completions, **kwargs):
+            return [0.0] * len(completions)
+
+        training_args = RLOOConfig(
+            output_dir=self.tmp_dir,
+            per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
+            num_generations=3,  # reduce the number of generations to reduce memory usage
+            max_completion_length=8,  # reduce the completion length to reduce memory usage
+            report_to="none",
+        )
+        trainer = RLOOTrainer(
+            model="trl-internal-testing/tiny-MistralForCausalLM-0.2",
+            reward_funcs=reward_func,
+            args=training_args,
+            train_dataset=dataset,
+        )
+
+        pad_token_id = trainer.processing_class.pad_token_id
+        assert pad_token_id is not None
+        assert trainer.model.config.pad_token_id == pad_token_id
+        assert trainer.model.generation_config.pad_token_id == pad_token_id
+
 
 @require_vision
 class TestRLOOTrainerVLM(TrlTestCase):
--- a/tests/test_distillation_trainer.py
+++ b/tests/test_distillation_trainer.py
@@ -1241,12 +1241,7 @@ class TestDistillationTrainer(TrlTestCase):
         # follow: otherwise `Trainer` realigns them at train time and reports it as a change the user did not make.
         dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
 
-        training_args = DistillationConfig(
-            output_dir=self.tmp_dir,
-            per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
-            max_completion_length=8,  # reduce the completion length to reduce memory usage
-            report_to="none",
-        )
+        training_args = DistillationConfig(output_dir=self.tmp_dir, report_to="none")
         trainer = DistillationTrainer(
             model="trl-internal-testing/tiny-MistralForCausalLM-0.2",
             teacher_model="trl-internal-testing/tiny-MistralForCausalLM-0.2",
--- a/tests/test_distillation_trainer.py
+++ b/tests/test_distillation_trainer.py
@@ -1236,6 +1236,29 @@ class TestDistillationTrainer(TrlTestCase):
 
         torch.testing.assert_close(loss, reference, rtol=1e-4, atol=1e-6)
 
+    def test_pad_token_id_synced_with_model_config(self):
+        # This model's tokenizer has no pad token, so the trainer falls back to the eos token. The model configs must
+        # follow: otherwise `Trainer` realigns them at train time and reports it as a change the user did not make.
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        training_args = DistillationConfig(
+            output_dir=self.tmp_dir,
+            per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
+            max_completion_length=8,  # reduce the completion length to reduce memory usage
+            report_to="none",
+        )
+        trainer = DistillationTrainer(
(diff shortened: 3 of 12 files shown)


qgallouedec:
thks for digging into this. I'm not sure we should merge it though.

`Trainer` already does exactly this alignment at train time, unconditionally, so the end state is identical.
The only thing the PR changes is that the user no longer sees the warning.

That warning is meant to be seen. It's telling the model user that their tokenizer and their model config disagree, which is a real thing to know about and to fix on the model side.

On the CI noise itself I've never had a good answer to this one. One option is to align the tokens on the tiny model fixtures so the warning never fires, but I don't love that either: the fixtures are meant to mimic real models, and real models do have this divergence. hiding it in the fixtures means we stop exercising the path our users actually hit.

albertvillanova:
Thanks @qgallouedec for the careful read. Let me make the case, then agree with you on the fixtures.

**Scope first**: this touches `pad_token_id` only. `align_special_tokens` still runs on every `train()` and still reports EOS and BOS. A user who added `<|im_end|>` still sees the message; a real Qwen still sees `bos_token_id: None`, which is the alarming one (https://github.com/huggingface/transformers/issues/48592). What disappears is one key, in a message that in most cases still prints.

**The contract is different from `transformers`.** Upstream's warning assumes the user brought a model *and* a tokenizer, so any disagreement is theirs, and the message is a receipt for their own edit. Our common entry point is `SFTTrainer(model="mistralai/Mistral-7B-v0.1")`: the user names one artifact, and *we* mutate the tokenizer a few lines before handing it to `Trainer` (`sft_trainer.py:1222`), because `DataCollatorForLanguageModeling` needs a concrete id. Base Mistral shipping no pad token is not a defect. We manufacture the divergence, silently, and `Trainer` then reports it back to the user as something their tokenizer did.

**And they cannot act on it.** There is nothing to fix on the model side; `mistralai/Mistral-7B-v0.1` is correct as published, and "fix it on the model side" would mean patching an upstream checkpoint to accommodate our collator. Fixing the tokenizer means adding a pad token, which is what we just did for them. Passing `pad_token=` in `SFTConfig` still fires, so a documented, supported argument produces a warning when used. The only remedy that works is to pre-set `model.config.pad_token_id` to whatever our internal fallback would have picked, i.e. to reimplement this PR by hand. A warning that fires on nearly every run with no action attached is how we lose the EOS signal we do want read.

**It is not only a log line.** `align_special_tokens` is called from `_inner_training_loop` only. Between `__init__` and `train()`, and permanently if the user never trains, the collator pads with `2` while `model.config.pad_token_id` and `model.generation_config.pad_token_id` are `None`. So `evaluate()`, `predict()`, `save_model()` and a bare `trainer.model.generate(...)` see a model whose config contradicts the tokenizer it was built with. I would want this fixed even without the message.

**We already do this in one trainer**: `reward_trainer.py:552` sets `model.config.pad_token_id` right after assigning the tokenizer's pad, for a functional reason. It is the only one of the eight that closes the loop. This finishes an existing convention rather than introducing one.

**On CI noise, you are right and I want to be honest about it**: this is not the change that quiets CI. Per the attribution in #7093, 4 fixtures stop warning and 8 lose one key; the 313-line dominant variant is entirely fixture-side. I would not merge this for the line count.

**On the fixtures**, that is #7098, and I think it lands on your side: it copies the *reference model's* ids rather than making the fixture self-consistent, so we keep exercising the divergence real Qwen users hit. It mirrors previously merged PRs.

If you are still not convinced, the question I could not answer is what a user of `SFTTrainer(model="mistralai/Mistral-7B-v0.1")` is supposed to do when they read that line.

qgallouedec:
OK, now I see it, and I agree, this is the right fix.

The thing I wanted to check before approving was whether we could accidentally overwrite a pad token id that the checkpoint had deliberately set, in the cases where the tokenizer and the config disagree.

It happens, but rarely: 4 out of 53. Everywhere else `config.pad_token_id` is already `None`, so we are filling a blank rather than replacing a value.

I don't think that's a big deal, but not 100% sure, it happened in the past that we messed up with token id alignment and it caused weird never-ending generations.

Full breakdown below.

## A. No pad token anywhere -> TRL falls back to EOS and creates the divergence

8 models.

| model | tokenizer | config | generation_config |
|---|---|---|---|
| `EleutherAI/pythia-14m` | None | None | None |
| `meta-llama/Llama-3.1-8B-Instruct` | None | None | None |
| `meta-llama/Llama-3.2-1B-Instruct` | None | None | None |
| `meta-llama/Meta-Llama-3-8B-Instruct` | None | None | None |
| `mistralai/Mistral-7B-Instruct-v0.1` | None | None | None |
| `mistralai/Mistral-7B-Instruct-v0.2` | None | None | None |
| `openai-community/gpt2` | None | None | None |
| `tiiuae/falcon-7b-instruct` | None | None | None |

## B. Tokenizer has one, `config.pad_token_id` is None -> divergence ships with the checkpoint

21 models.

| model | tokenizer | config | generation_config |
|---|---|---|---|
| `OpenGVLab/InternVL3-8B-hf` | 151643 (`<\|endoftext\|>`) | None | None |
| `Qwen/Qwen2-VL-2B-Instruct` | 151643 (`<\|endoftext\|>`) | None | 151643 |
| `Qwen/Qwen2.5-32B-Instruct` | 151643 (`<\|endoftext\|>`) | None | 151643 |
| `Qwen/Qwen2.5-Coder-0.5B` | 151643 (`<\|endoftext\|>`) | None | None |
| `Qwen/Qwen2.5-VL-3B-Instruct` | 151643 (`<\|endoftext\|>`) | None | 151643 |
| `Qwen/Qwen3-30B-A3B` | 151643 (`<\|endoftext\|>`) | None | 151643 |
| `Qwen/Qwen3-4B` | 151643 (`<\|endoftext\|>`) | None | 151643 |
| `Qwen/Qwen3-4B-Instruct-2507` | 151643 (`<\|endoftext\|>`) | None | 151643 |
| `Qwen/Qwen3-8B` | 151643 (`<\|endoftext\|>`) | None | 151643 |
| `Qwen/Qwen3-VL-2B-Instruct` | 151643 (`<\|endoftext\|>`) | None | 151643 |
| `Qwen/Qwen3.5-0.8B` | 248044 (`<\|endoftext\|>`) | None | n/a |
| `Qwen/Qwen3.5-4B` | 248044 (`<\|endoftext\|>`) | None | n/a |
| `Qwen/Qwen3.6-35B-A3B` | 248044 (`<\|endoftext\|>`) | None | 248044 |
| `Qwen/Qwen3.8-27B` | 248044 (`<\|endoftext\|>`) | None | 248044 |
| `deepseek-ai/DeepSeek-R1` | 1 (`<｜end▁of▁sentence｜>`) | None | None |
| `deepseek-ai/DeepSeek-R1-0528` | 1 (`<｜end▁of▁sentence｜>`) | None | None |
| `deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B` | 151643 (`<｜end▁of▁sentence｜>`) | None | None |
| `google/gemma-3-4b-it` | 0 (`<pad>`) | None | 0 |
| `llava-hf/llava-v1.6-mistral-7b-hf` | 32001 (`<pad>`) | None | None |
| `meta-models/Muse-Glimmer-30B` | 200018 (`<\|finetune_right_pad\|>`) | None | 200018 |
| `someorgtoo-hf/onyx-hf-converted` | 200018 (`<\|finetune_right_pad\|>`) | None | 200018 |

## C. Two values are set and they point at different tokens

4 models.

| model | tokenizer | config | generation_config |
|---|---|---|---|
| `HuggingFaceTB/SmolVLM2-2.2B-Instruct` | 2 (`<\|im_end\|>`) | 128002 | 2 |
| `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16` | 11 (`<\|im_end\|>`) | 0 | 0 |
| `nvidia/NVIDIA-Nemotron-3-Ultra-550B-A55B-BF16` | 11 (`<\|im_end\|>`) | 0 | 0 |
| `nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16` | 11 (`<\|im_end\|>`) | 0 | 0 |

## D. No contradiction

20 models.

| model | tokenizer | config | generation_config |
|---|---|---|---|
| `CohereLabs/aya-expanse-8b` | 0 (`<PAD>`) | 0 | 0 |
| `CohereLabs/tiny-aya-earth` | 0 (`<PAD>`) | 0 | 0 |
| `HuggingFaceM4/Idefics3-8B-Llama3` | 128002 (`<\|reserved_special_token_0\|>`) | 128002 | 128002 |
| `HuggingFaceM4/idefics2-8b` | 0 (`<unk>`) | 0 | 0 |
| `LiquidAI/LFM2-1.2B` | 0 (`<\|pad\|>`) | 0 | 0 |
| `LiquidAI/LFM2.5-230M` | 0 (`<\|pad\|>`) | 0 | 0 |
| `LiquidAI/LFM2.5-VL-3B` | 124893 (`<\|pad\|>`) | 124893 | 124893 |
| `allenai/Olmo-3-7B-Think` | 100277 (`<\|pad\|>`) | 100277 | None |
| `facebook/opt-1.3b` | ? (`<pad>`) | 1 | 1 |
| `google/diffusiongemma-26B-A4B-it` | 0 (`<pad>`) | 0 | 0 |
| `google/gemma-2-2b-it` | 0 (`<pad>`) | 0 | 0 |
| `google/gemma-4-E2B-it` | 0 (`<pad>`) | 0 | 0 |
| `google/gemma-7b-it` | 0 (`<pad>`) | 0 | 0 |
| `google/paligemma-3b-pt-224` | 0 (`<pad>`) | 0 | 0 |
| `llava-hf/llava-1.5-7b-hf` | 32001 (`<pad>`) | 32001 | 32001 |
| `microsoft/Phi-3-mini-4k-instruct` | 32000 (`<\|endoftext\|>`) | 32000 | 32000 |
| `microsoft/Phi-3.5-mini-instruct` | 32000 (`<\|endoftext\|>`) | 32000 | 32000 |
| `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16` | None | 0 | 0 |
| `openai/gpt-oss-20b` | 199999 (`<\|endoftext\|>`) | 199999 | 199999 |
| `zai-org/GLM-4.5` | 151329 (`<\|endoftext\|>`) | 151329 | 151329 |

tests/test_grpo_trainer.py
```diff
@@ -3940,6 +3940,33 @@ def test_single_reward_model_with_single_processing_class(self):
...
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        def reward_func(completions, **kwargs):
+            return [0.0] * len(completions)
+
+        training_args = GRPOConfig(
+            output_dir=self.tmp_dir,
+            per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
+            num_generations=3,  # reduce the number of generations to reduce memory usage
+            max_completion_length=8,  # reduce the completion length to reduce memory usage
```
qgallouedec: nit, we don't train, so we don't need these

tests/test_rloo_trainer.py
```diff
@@ -1846,6 +1846,33 @@ def test_single_reward_model_with_single_processing_class(self):
...
+        dataset = load_dataset("trl-internal-testing/zen", "standard_prompt_only", split="train")
+
+        def reward_func(completions, **kwargs):
+            return [0.0] * len(completions)
+
+        training_args = RLOOConfig(
+            output_dir=self.tmp_dir,
+            per_device_train_batch_size=3,  # reduce the batch size to reduce memory usage
+            num_generations=3,  # reduce the number of generations to reduce memory usage
+            max_completion_length=8,  # reduce the completion length to reduce memory usage
```
qgallouedec: same here

qgallouedec (approved):
thanks

albertvillanova:
Thanks for your detailed review, @qgallouedec. I'm addressing it.

albertvillanova:
@qgallouedec, about the 4 models in category C, I think we are safe, for a reason your table makes clear: `align_special_tokens` performs the *same* overwrite at `train()` today, unconditionally. So for those four models the end state after training is identical with or without this PR. All that changes is the moment it happens, `__init__` instead of `train()`.

The window in between is where it actually helps. Right now the collator is already padding with the tokenizer's id while `config.pad_token_id` still holds the checkpoint's, so anything that reads the config before the first training step, `evaluate()`, `predict()`, `save_model()`, or a bare `generate()`, sees a model contradicting the batches it is being fed. For SmolVLM2 that is padding with `2` while the config says `128002`. Closing that window is the part I would want even setting the warning aside.

On the never-ending generations, worth noting the direction: the id we write is the one the tokenizer used to build the padded batch, so the config ends up agreeing with the data rather than diverging from it. If that ever goes wrong it would be a tokenizer whose pad token is wrong for the model, which is a different bug and one `Trainer` would land us in anyway.

Nits addressed: dropped the training arguments from the GRPO and RLOO tests, since neither trains. I applied the same to the distillation test, which had the same dead arguments.

PR #5443, opened 2026-04-03
Update vLLM minimum supported version to 0.11.0

albertvillanova:
Update vLLM minimum supported version to 0.11.0.

This PR updates the minimum supported version of vLLM in TRL from 0.10.2 to 0.11.0 and simplifies related compatibility logic throughout the codebase. The changes remove conditional imports and branching for vLLM versions below 0.11.0, standardize the use of `StructuredOutputsParams` and the `"structured_outputs"` key, and update documentation and dependency constraints accordingly.

These changes ensure a simpler and more maintainable codebase by dropping support for legacy vLLM version and unifying the handling of structured outputs.

### Changes

**vLLM Version Support Updates:**

- Raised the minimum supported vLLM version from 0.10.2 to 0.11.0 in documentation, dependency specifications (`pyproject.toml`), and version checks throughout the codebase. 

**Code Simplification and Cleanup:**

- Removed all conditional imports and logic branches that handled vLLM versions below 0.11.0, including the use of `GuidedDecodingParams` and the `"guided_decoding"` key. Now, only `StructuredOutputsParams` and the `"structured_outputs"` key are used.

- Updated all relevant code to consistently use `StructuredOutputsParams` and the `"structured_outputs"` key for structured outputs, removing references to `"guided_decoding"`.

**Import and Dependency Adjustments:**

- Cleaned up imports by removing unnecessary version checks and imports of `Version` from `packaging`, and ensuring direct imports of `StructuredOutputsParams` where needed.

---

> [!NOTE]
> **Medium Risk**
> Medium risk: drops support for vLLM<0.11.0 and standardizes structured output handling, which could break environments pinned to older vLLM or custom `generation_kwargs` that relied on the legacy `guided_decoding` key.
> 
> **Overview**
> Updates TRL’s vLLM support window by raising the minimum supported vLLM version to `0.11.0` (docs, `pyproject.toml`, and `is_vllm_available` warning/version gate).
> 
> Simplifies vLLM integration code by removing compatibility branches for vLLM `<=0.10.2` and standardizing structured-output configuration to always use `StructuredOutputsParams` with the `"structured_outputs"` key across trainers/utilities and the `trl vllm-serve` server path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c71bd96282b2db7197e43165af984a1d98dc9c72. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7102, opened 2026-09-07
Require peft>=0.13.0 and drop the 0.12 version guards

qgallouedec:
I think it's high time to raise the PEFT floor from 0.8.0 (Jan 2024) to 0.13.0 (Sep 2024) and removes the version guards it makes dead.

- `autocast_adapter_dtype` was added in PEFT 0.12.0. The `Version(peft.__version__) >= Version("0.12")` check appeared in 13 places (7 stable trainers, 5 experimental trainers, `experimental/utils.py`); all are now unconditional.
- 0.13.0 rather than 0.12.0 so the floor is no older than 2 years.
- Drops 9 imports that became unused (`import peft`, `packaging.version.Version`).

---

> [!NOTE]
> **Low Risk**
> Dependency floor bump may break installs pinned to older PEFT; training logic is simplified, not changed, for supported versions.
> 
> **Overview**
> Raises the minimum **PEFT** dependency from `0.8.0` to **`0.13.0`** in `pyproject.toml` (both the `peft` extra and the bundled dev dependency set).
> 
> Because **`autocast_adapter_dtype`** is guaranteed on supported PEFT versions, TRL drops **`Version(peft.__version__) >= 0.12`** checks across stable and experimental trainers (DPO, GRPO, SFT, KTO, RLOO, reward, distillation, BCO, CPO, ORPO, TPO, IW-OPD) and in `experimental/utils.py`. **DeepSpeed ZeRO-3 + non-quantized PEFT** still passes `autocast_adapter_dtype=False`; **sharded QLoRA (4-bit)** still disables autocast without a version gate.
> 
> Unused **`import peft`** (only needed for `__version__`) and **`packaging.version.Version`** imports tied to those guards are removed where applicable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 702618258ad04ef11002dc5902dab18996ddee8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova (approved):
LGTM, the change is correct and complete. Verified on `70261825`.

## One thing I would like us to settle separately: the criterion for raising a floor

Not a request to change anything here. It is about the rule the next one of these gets reviewed against.

Every floor bump in the repo so far has been feature-driven, with the required feature named in the title or body: `datasets>=4.7.0` for the Json dtype (#5376), `kernels>=0.14.0` for `trust_remote_code`, `vllm>=0.11.0` (#5443). This PR adds a second, different rationale: the floor should be no older than two years. I think that rationale is defensible, and it happens to match SPEC 0, the Scientific Python standard that superseded NEP 29 and recommends dropping dependency releases older than 24 months. But it is not written down anywhere, and it is in direct tension with something I wrote myself: in #6447 I stated the opposite rule for bitsandbytes, "Do not bump the floor solely to drop this filter: wait for an independent reason to require 0.50.0."

### Why this needs to be written down rather than settled per PR

- **Contributors cannot tell in advance whether the work is wanted.** #6893 raises the bitsandbytes floor to drop a warning filter, which is exactly what #6447 says not to do, and it has been open for two weeks with no way to resolve it in either direction. That is not the contributor's mistake. There was nothing for them to read.
- **Any maintainer should be able to review one of these.** Without a written rule, a floor bump needs a judgment call rather than a check, so it queues behind whoever is presumed to hold the position instead of being reviewable by whoever picks it up. A criterion turns the review into something verifiable from the diff.
- **The floor is a public promise.** It is what users install against, so deciding it case by case means it drifts in whichever direction the most recent PR argued.

### The two questions

1. **When is a bump justified at all?** Feature needed, dead guards or shims removed, or calendar staleness. Whichever we pick, a reviewer should be able to check it mechanically instead of forming an opinion per PR.
2. **Guard or bump?** Guards exist so the floor can stay low and newer features light up only on newer installs, so guarding should be the default and a bump should be argued on what it deletes.

I will open an issue with a concrete proposal for both questions so we can discuss it there.

albertvillanova (commented):
Opened #7116 for the criterion discussion.

PR #7121, opened 2026-09-09
New example: [REDACTED]

qgallouedec:
LoRA SFT on [REDACTED] on 8 nodes with `SFTTrainer`, sharded at load time by transformers' `DistributedConfig`.

Same shape as #6869's examples.

Needs huggingface/transformers#48204.

---

> [!NOTE]
> **Low Risk**
> Adds documentation and standalone example/Slurm scripts only; no changes to TRL training library code paths.
> 
> **Overview**
> Adds a new **`sft_redacted`** example for **LoRA SFT** on a large MoE-style checkpoint using **`SFTTrainer`** on **8×8 GPUs**, with load-time sharding via transformers **`DistributedConfig`** (`fsdp_size=64`, `ep_size=8` expert parallelism).
> 
> The training script loads **allenai/tulu-3-sft-mixture** (500k rows) before the model, enforces a **pre-release transformers** build for `ep_size`, and sets packing, parallel tokenization, **`save_only_model`**, and Trackio logging. A **Slurm** job file launches **`torchrun`** across nodes with IPv4 rendezvous.
> 
> The examples index in **`docs/source/example_overview.md`** is updated with a row for this folder.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c53d461d768c75978be8428bdf81ceb4d2dfa201. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


examples/sft_moe_expert_parallel/sft_moe_expert_parallel.slurm
```diff
@@ -0,0 +1,20 @@
...
+#SBATCH --time=08:00:00
+
+MASTER=$(scontrol show hostnames "$SLURM_JOB_NODELIST" | head -1)
+# ahostsv4 avoids link-local IPv6 addresses, which torchrun cannot rendezvous on
+MASTER_IP=$(getent ahostsv4 "$MASTER" | awk 'NR==1{print $1}')
+
+srun torchrun \
+    --nnodes "$SLURM_NNODES" \
+    --nproc_per_node 8 \
+    --node_rank "$SLURM_NODEID" \
```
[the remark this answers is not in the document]
qgallouedec: half right. The expansion happens, but --node_rank is ignored under rdzv_backend=c10d; torchrun logs exactly that. Removed anyway, it was dead.

examples/sft_moe_expert_parallel/sft_moe_expert_parallel.py
```diff
@@ -0,0 +1,83 @@
...
+#     "trl @ git+https://github.com/huggingface/trl.git",
+#     "transformers @ git+https://github.com/huggingface/transformers.git@ep-fsdp-2d-mesh",
+#     "peft @ git+https://github.com/huggingface/peft.git",
+#     "trackio",
+# ]
+# ///
+
+# docstyle-ignore
+"""
+LoRA SFT of a large MoE on 64 H100s across 8 nodes.
```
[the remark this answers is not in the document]
qgallouedec: correct, real test failure, fixed.

examples/sft_moe_expert_parallel/sft_moe_expert_parallel.py
```diff
@@ -0,0 +1,83 @@
...
+trains on its own slice of the batch and tokens travel to the rank owning their expert. Measured: 22.3 s/step at
+per-device batch 4 (23k tokens/s), 43.7 s with gradient accumulation 2, loss 1.26 -> 0.6 over 100 steps on tulu-3.
+
+Stage the checkpoint on node-local disk first: sharded loading took 29 s from NVMe and did not finish in 40 min
+from a network mount. Gradient checkpointing has to stay on: weights, optimizer state and FSDP buffers already
+take 66-70 GB per GPU. Needs transformers `ep-fsdp-2d-mesh` (#48204) and peft main (#3578), as the header pins.
+
+Launch from this directory:
+
+    sbatch sft_moe_expert_parallel.slurm
```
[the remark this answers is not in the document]
qgallouedec: declining. No example in the repo launches via uv run, and both sibling multi-node examples use the same srun torchrun.

examples/sft_moe_expert_parallel/sft_moe_expert_parallel.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Read the dataset before loading the model: 64 GPUs holding a loaded model is the most expensive place to wait
+# on I/O.
+train_dataset = load_dataset("allenai/tulu-3-sft-mixture", split="train[:500000]")
+
+training_args = SFTConfig(
+    output_dir="sft_moe_expert_parallel",
+    model_init_kwargs={
+        "dtype": torch.bfloat16,
+        "distributed_config": DistributedConfig(
```
AmineDiro, line 61: does this work out of the box with accelerate ? imo, we should reconsider the interaction between running with torchrun vs accelerate like we discussed to catch weird behavior. For example : 
- setting this in training args + run with accelerate with a different config. 
- run with accelerate + distributed_config
- accelerate has different FSDP or deepspeed ?

examples/sft_moe_expert_parallel/sft_moe_expert_parallel.py
```diff
@@ -0,0 +1,83 @@
...
+            tp_size=8, fsdp_size=8, enable_expert_parallel=True, expert_parallel_dispatch=True
```
AmineDiro, line 62: so this depends on : https://github.com/huggingface/transformers/pull/48518 ?

albertvillanova: Yes, this depends on transformers#48518. Verified by ancestry, `ep-token-dispatch` is merged into the pinned `ep-fsdp-2d-mesh` branch, which is where expert_parallel_dispatch is defined (absent from transformers main).

albertvillanova (commented):
Thanks.

Codex raised three points: 2 were fixed, and 1 declined:
- **PEP 723 header vs sbatch:** declined. I think the decline is factually right but leaves the actual gap open.
  - The PEP 723 header is inert. `sbatch → srun torchrun` never reads it, so its pins install nothing. Note the tension: `examples/README.md` requires the header, and #6869's two SLURM examples have no header at all. Three files, three answers. One decision would help.

## Findings
1. **Merge ordering.** The only install path is transformers#48204, a draft umbrella branch that will disappear when its constituent PRs land. Separately, main's docs/source/distributing_training.md has no expert-parallelism or DistributedConfig prose at all: that section comes from #6869, still open. Merging this first puts an index row in main pointing at a concept documented nowhere in TRL's docs, and a launcher pattern (srun torchrun) that contradicts the only multi-node guidance main has (srun accelerate launch).
1. **No environment guard (should fix).** Verified on current transformers: `DistributedConfig(..., expert_parallel_dispatch=True)` raises `TypeError: DistributedConfig.__init__() got an unexpected keyword argument 'expert_parallel_dispatch'`. On 64 ranks, and after `load_dataset("allenai/tulu-3-sft-mixture", split="train[:500000]")` has already been paid for, since the dataset load precedes SFTConfig. The sibling already in main (`sft_qwen3_8b_1m_context`) sets the precedent: a 4-line check right after the imports raising a `RuntimeError` that names the branch to install. That also makes Codex point above harmless.
1. **Exactly one checkpoint, at the last step, never in `output_dir`.** `save_strategy` defaults to "steps" and save_steps to 500, equal to max_steps, so there is a single save into `output_dir/checkpoint-500`, and no `trainer.save_model()`. At the script's own configuration (grad accum 2 → 43.7 s/step) that is ~6.1 h against --time=08:00:00, before preprocessing. A wall-clock kill or preemption at step 480 leaves nothing after 64 H100-hours. The LoRA adapter "saves in seconds" per #6869's own docstring, so a real save_steps cadence is near-free insurance. Related: the measurements are quoted at 100 steps, so was the save path ever exercised? The sibling chose `save_strategy="no"` plus an explicit save_model, so the two examples now show two different save recipes with no stated reason.
1. **`dataset_num_proc` unset.** TRL prepares the dataset inside `main_process_first()` with `num_proc=args.dataset_num_proc` (`sft_trainer.py:1453-1455`), so 500k rows get tokenized and bfd-packed single-threaded on rank 0 while 63 ranks idle, on nodes where the launcher already requests `--cpus-per-task=64`. This is the exact failure mode the script's own comment warns about, and it eats the 8 h budget.
1. **`.slurm` drift across the three launchers.** This copy drops `--node_rank "$SLURM_NODEID"`; #6869's two copies still carry it (same dead flag, same reason). This copy also drops `export HF_SHARD_PREFETCH=4`, which both siblings set, while prescribing node-local staging instead and never mentioning the prefetch knob. Per the repo's consistency rule, propagate the node_rank removal and reconcile the two loading recipes, otherwise the three files read as drift rather than intent.
1. **Router aux loss, probably unintended.** `router_aux_loss_coef` defaults to 0.001, so `aux_loss_enabled` is True for any MoE and `output_router_logits=True` is forwarded every step (`sft_trainer.py:1410-1419`). With LoRA on `q/k/v/o` only, the router and the experts are frozen, so the load-balancing term cannot rebalance anything; it only perturbs the attention adapters. Under `expert_parallel_dispatch` it is also a per-rank quantity, unlike the siblings' shared-batch EP. If unintended, `router_aux_loss_coef=0.0` drops the term and the router-logit memory. If intended, that is exactly what the docstring should say.

### Nits
- the headline number (22.3 s/step, 23k tok/s) describes a configuration the script does not use; I checked the arithmetic and it is internally consistent (64 × 4 × 2048 / 22.3 = 23.5k tok/s), it just is not this config. 
- `# docstyle-ignore` is inert under `examples/` (the doc-builder hook is scoped to `^(trl|tests|docs/source)/`). 
- `trl @ git+...` where the 1M example pins plain "trl", and nothing here appears to need trl main.

## Verified correct

Worth saying explicitly, since several claims in the docstring are load-bearing:
- The block-diagonal-mask claim is accurate: find_packed_sequence_indices plus packed_sequence_mask_function build it from position_ids when attention_mask is None.
- save_only_model=True is precisely what upstream's guard demands, and the comment paraphrases the upstream error faithfully.
- expert_parallel_dispatch's three preconditions (train_sampling_strategy == "random", sized dataset, dispatch_batches False) all hold: TRL leaves the strategy at the default and does not override _get_train_sampler.
- peft pin is right: #3578 merged 2026-08-27, v0.20.0 released 2026-07-28.
- The model_init_kwargs path (new here, the siblings call from_pretrained directly) is safe only because TRL forces device_map=None when distributed_type is MULTI_GPU, which is what torchrun gives. Running the same script single-process would inject device_map="auto" alongside distributed_config.
- report_to="trackio" with a non-JSON model_init_kwargs is safe: trackio's to_json_safe converts DistributedConfig via to_dict() and torch.bfloat16 via str(). Adjacent, not this PR: args.to_json_string() does raise TypeError: Object of type DistributedConfig is not JSON serializable, so report_to="tensorboard" would break on this and on the already-merged 1M example.

## CI

Everything passes except "Tests with latest dependencies": test_grpo_trainer.py::TestGRPOTrainerVLM::test_train_vlm_with_pad_to_multiple_of, ValueError: Image features and image tokens do not match, tokens: 513, features: 512. A docs-and-examples diff cannot cause a GRPO VLM failure, and only the latest-deps job is red (dev, minimum and no-optional all pass), so it is an upstream-latest issue on a 4-commit-stale base. Rebase and re-run to clear it. mergeable_state is blocked, no human approval yet.

qgallouedec:
thanks

> No environment guard (should fix).

It's because it's hip, when transformers releases, we will guard the transformers version. In the meantime:
Fixed in 2cee1514

> `dataset_num_proc` unset.

Set to 32.

> Exactly one checkpoint, at the last step, never in `output_dir`.

Fixed in 73cdd790, save_steps=50

> Router aux loss, probably unintended.

Intended, keeping it. The router weights are frozen but its inputs are not, so routing still drifts under LoRA. But maybe I'm missing your point?

> The only install path is transformers#48204, a draft umbrella branch that will disappear when its constituent
> PRs land.

Yes, because it's WIP, we're working on this with the transformers team, in the meantime, to allow the team to test, we have to work with umbrella branches.

> Merging this first puts an index row in main pointing at a concept documented nowhere in TRL's docs

Out of scope here, #6869 is still wip.

> a launcher pattern (srun torchrun) that contradicts the only multi-node guidance main has

torchrun is what accelerate launch runs underneath, see the internal thread. Nothing here needs an accelerate config, the model is sharded at load time by DistributedConfig.

> `.slurm` drift across the three launchers.

Same answer, #6869 is wip, I'd ignore that branch for now.

> the headline number (22.3 s/step, 23k tok/s) describes a configuration the script does not use

Right. Setting gradient_accumulation_steps=1 so the docstring describes what was run: 14.1 s/step, 131k
tokens/step, 40.2 GB.

> `# docstyle-ignore` is inert under `examples/`

removed. For the records it exists in other examples.

> `trl @ git+...` where the 1M example pins plain "trl", and nothing here appears to need trl main.

It was needed before v1.13.0 because of #7101. Now we can remove.

examples/sft_moe_expert_parallel/sft_moe_expert_parallel.py
qgallouedec: this should be gone after the transformers release

examples/sft_moe_expert_parallel/sft_moe_expert_parallel.py
qgallouedec: this is a wider question addressed internally, I'm closing this thread

examples/sft_redacted/sft_redacted.py
```diff
@@ -0,0 +1,87 @@
...
+    save_only_model=True,
+    # About 13 min of work at risk. The adapter is 194 MB, so the save itself takes seconds.
+    save_steps=50,
+    report_to="trackio",
+)
+trainer = SFTTrainer(
+    model=MODEL,
+    args=training_args,
+    train_dataset=train_dataset,
+    peft_config=LoraConfig(r=16, lora_alpha=32, target_modules=["q_proj", "k_proj", "v_proj", "o_proj"]),
```
AmineDiro, line 85: To be clear, all the experts and the router stay frozen, right ? Because the model is quite big, most of the params probably (90% are expert weights) so the LoRA adapter are on attention only which could end up to be very small.

Can this recipe work with something like :
```python
LoraConfig(target_modules=[], target_parameters=["mlp.experts.gate_up_proj", "mlp.experts.down_proj"])
```
🤔 ?

PR #7127, opened 2026-09-09
Sync the model config eos token id with the tokenizer in the trainers

albertvillanova:
This PR keeps `model.config` and `model.generation_config` in agreement with the tokenizer's eos token in the SFT and Reward trainers, as #7097 did for the pad token.

### Motivation

`SFTConfig.eos_token` and `RewardConfig.eos_token` let the user pick the token that marks the end of a turn or sequence. The trainers set it on the tokenizer and leave the model configs untouched. `Trainer.train()` then finds the divergence, rewrites both configs itself, and logs it as a change the user did not make:

```
The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config. The model config and generation config were aligned accordingly, being updated with the tokenizer's values. Updated tokens: {'eos_token_id': 151644}.
```

Part of #7093.

### Solution

Mirror the eos token onto the model configs right where the trainers set it on the tokenizer. The alignment that `Trainer` performs at train time then finds nothing to change, so the end state is identical, while the model stays consistent with the tokenizer from the moment the trainer is built and for any code reading the configs before `train()`.

The generation config may hold several eos tokens, and any of them halts generation, so the new id is added to the ones already there instead of replacing them. `align_special_tokens` preserves them the same way. The only divergence from it is that an id the list already contains is not appended a second time.

`RewardTrainer` needs the model config only: a sequence classification model has no `generation_config` attribute at all, which is what `align_special_tokens` guards on.

No other trainer is affected, since `eos_token` is exposed by `SFTConfig` and `RewardConfig` only.

### Changes

- Mirror the tokenizer's eos token onto the model config, and onto the generation config where there is one, wherever the SFT and Reward trainers assign it
- Add a regression test to each, covering the causal LM path where the existing eos tokens are preserved and the sequence classification path where there is no generation config

---

> [!NOTE]
> **Low Risk**
> Init-time config alignment only; behavior at train time was already corrected by Trainer, with tests locking in tokenizer/config consistency.
> 
> **Overview**
> When **`SFTConfig.eos_token`** or **`RewardConfig.eos_token`** is set, the trainers already update the tokenizer; this PR **also aligns model configs at init** so Hugging Face **`Trainer`** does not log a spurious “tokenizer vs model config” realignment at **`train()`** time.
> 
> **`RewardTrainer`** copies the tokenizer’s **`eos_token_id`** onto **`model.config`** only (sequence-classification models have no generation config). **`SFTTrainer`** sets **`model.config.eos_token_id`** and **prepends** the new id to **`generation_config.eos_token_id`** when it is not already present, keeping any existing EOS ids for generation.
> 
> Regression tests assert config sync for a custom **`<|im_start|>`** EOS token on both trainers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b88511b21fb05036ac53f2f4c22ffffc17841b46. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_reward_trainer.py
+++ b/tests/test_reward_trainer.py
@@ -1014,3 +1014,19 @@ class TestRewardTrainer(TrlTestCase):
         for n, param in previous_trainable_params.items():
             new_param = trainer.model.get_parameter(n)
             assert not torch.equal(param, new_param), f"Parameter {n} has not changed."
+
+    def test_eos_token_id_synced_with_model_config(self):
+        # The trainer sets the requested eos token on the tokenizer. The model config must follow: otherwise
+        # `Trainer` realigns it at train time and reports it as a change the user did not make.
+        dataset = load_dataset("trl-internal-testing/zen", "standard_implicit_prompt_preference", split="train")
+
+        training_args = RewardConfig(output_dir=self.tmp_dir, eos_token="<|im_start|>", report_to="none")
+        trainer = RewardTrainer(
+            model="trl-internal-testing/tiny-Qwen2ForSequenceClassification-2.5",
+            args=training_args,
+            train_dataset=dataset,
+        )
+
+        eos_token_id = trainer.processing_class.eos_token_id
+        assert eos_token_id == 151644  # <|im_start|>
+        assert trainer.model.config.eos_token_id == eos_token_id
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -2292,6 +2292,22 @@ class TestSFTTrainer(TrlTestCase):
         assert trainer.model.config.pad_token_id == pad_token_id
         assert trainer.model.generation_config.pad_token_id == pad_token_id
 
+    def test_eos_token_id_synced_with_model_config(self):
+        # The trainer sets the requested eos token on the tokenizer. The model configs must follow: otherwise
+        # `Trainer` realigns them at train time and reports it as a change the user did not make.
+        dataset = load_dataset("trl-internal-testing/zen", "standard_language_modeling", split="train")
+
+        training_args = SFTConfig(output_dir=self.tmp_dir, eos_token="<|im_start|>", report_to="none")
+        trainer = SFTTrainer(
+            model="trl-internal-testing/tiny-Qwen2ForCausalLM-2.5", args=training_args, train_dataset=dataset
+        )
+
+        eos_token_id = trainer.processing_class.eos_token_id
+        assert eos_token_id == 151644  # <|im_start|>
+        assert trainer.model.config.eos_token_id == eos_token_id
+        # The model's own eos tokens are kept, since any of them halts generation
+        assert trainer.model.generation_config.eos_token_id == [151644, 151645, 151643]
+
 
 @pytest.mark.slow
 @require_torch_accelerator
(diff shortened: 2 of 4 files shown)


albertvillanova:
## Measured against main

Run on this branch: https://github.com/huggingface/trl/actions/runs/34334811276/job/102411616255
Run on main for comparison: https://github.com/huggingface/trl/actions/runs/34333006767/job/102405754044

| | main | this branch |
|---|---:|---:|
| total realignment lines | 457 | 457 |
| lines mentioning `eos_token_id` | 124 | 124 |
| lines mentioning `bos_token_id` | 433 | 433 |
| lines mentioning `pad_token_id` | 28 | 28 |
| tests passed | 2391 | 2393 |

Not one variant moved. That is the correct result, and it is worth spelling out why, because it means this PR should not be sold as a noise reduction.

**Nothing exercised this path before.** `grep "eos_token=" tests/` returns nothing outside the two tests added here, so `SFTConfig.eos_token` and `RewardConfig.eos_token` had no coverage at all. There were no existing lines for the fix to remove.

**The evidence that it works is an absence.** The run gains two tests and stays at exactly 457 lines. Both new tests set `eos_token="<|im_start|>"` (151644) on a Qwen fixture, which without this change would make `Trainer` report `{'eos_token_id': 151644}` once each:

```
occurrences of 151644 in any realignment line: 0
PASSED tests/test_reward_trainer.py::TestRewardTrainer::test_eos_token_id_synced_with_model_config
PASSED tests/test_sft_trainer.py::TestSFTTrainer::test_eos_token_id_synced_with_model_config
```

Two trainings that set a new eos, and no realignment message from either.

**The 124 remaining eos lines are not ours.** They come from fixtures whose reference config disagrees with its own tokenizer, the same class as the bos residual. `tiny-Cohere2ForCausalLM` is the clearest case: `<EOS_TOKEN>` in the config, `<|END_OF_TURN_TOKEN|>` in the tokenizer. I have corrected the note on #7093 that claimed otherwise.

qgallouedec:
LGTM. 
The fact that a config list gets flattened to a single id (eg Llama-3.1-8B-Instruct ships `[128001, 128008, 128009]`) is upstream behaviour you're reproducing, not something this introduces.

Same nit as #7097: "the end state is unchanged" describes what `Trainer` does elsewhere rather than what the line does, and the reason this PR exists is that the state before `train()` is not the same. Something like "the collator and the model must agree on eos from construction" reads better.

From experimence, proper handling of generation's config eos is super important, so this one is more critical than it might look

qgallouedec (approved):

albertvillanova:
Thanks for the review.

Agreed on the comment: it argues why the line is safe instead of saying what it does, and "the end state is unchanged" reads oddly in a PR whose whole point is that the state before `train()` is not the same. The same sentence already sits on main from the pad token work, so it should change everywhere at once rather than only here. I'll handle that separately so this PR stays as reviewed.

Agreed on the eos criticality too, and the honest state is that these tests exercise only one of the three shapes `generation_config.eos_token_id` takes: a list. The int shape is the most common one in the wild and is not covered here. I'll extend the coverage separately.

PR #7128, opened 2026-09-09
Sync the model config pad token id with the tokenizer in the experimental trainers

albertvillanova:
This PR extends #7097 to the experimental trainers that set the tokenizer's pad token themselves.

### Motivation

When the tokenizer has no pad token, these trainers fall back to the eos token and set it on the tokenizer, but leave the model configs untouched. `Trainer.train()` then finds the divergence, rewrites both configs itself, and logs it as a change the user did not make:

```
The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config. The model config and generation config were aligned accordingly, being updated with the tokenizer's values. Updated tokens: {'pad_token_id': 2}.
```

#7097 closed this in the main trainers. The same block is duplicated in the experimental ones, and `AGENTS.md` asks the copies to stay aligned.

Part of #7093.

### Solution

Mirror the pad token onto the model config and generation config right where the trainer assigns it on the tokenizer, exactly as the main trainers now do. The alignment that `Trainer` performs at train time then finds nothing to change, so the end state is identical, while the model stays consistent with the tokenizer from the moment the trainer is built.

Applied to Online DPO, SDPO, TPO, SDFT, SSD, CPO, IW-OPD, Async GRPO and Async Distillation.

GOLD needs no change: it delegates to `SFTTrainer`, which #7097 already covers. ORPO never assigns a pad token to the processing class, so there is no divergence to mirror.

Teacher and reward models are untouched, consistent with #7097: `Trainer` aligns its own model only, and the reward paths already set `config.pad_token_id`.

### Changes

- Mirror the tokenizer's pad token onto the model config and generation config wherever these nine trainers assign it
- Add the same regression test as #7097 to the seven trainers that can be constructed without an accelerator

---

> [!NOTE]
> **Low Risk**
> Init-time config alignment only; training behavior should match the previous end state after Trainer’s automatic fix, with no auth or data-path changes.
> 
> **Overview**
> Experimental trainers that assign the tokenizer’s pad token (typically **EOS** when none exists) now also set **`model.config.pad_token_id`** and **`model.generation_config.pad_token_id`** at init, matching the fix from #7097 in the main trainers.
> 
> This avoids Hugging Face **`Trainer`** realigning configs at **`train()`** and logging a misleading “tokenizer tokens differ from model config” warning. **CPO**, **Online DPO**, **SDPO**, **SDFT**, **SSD**, **TPO**, **IW-OPD**, **Async GRPO**, and **Async Distillation** are updated; regression tests **`test_pad_token_id_synced_with_model_config`** were added for seven of them (Mistral tiny model, no pad token).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 455cd227177a6646915210f410464ac8ac62ed9d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/experimental/test_cpo_trainer.py
+++ b/tests/experimental/test_cpo_trainer.py
@@ -296,3 +296,18 @@ class TestCPOTrainer(TrlTestCase):
             new_param = trainer.model.get_parameter(n)
             if param.sum() != 0:
                 assert not torch.equal(param, new_param)
+
+    def test_pad_token_id_synced_with_model_config(self):
+        # This model's tokenizer has no pad token, so the trainer falls back to the eos token. The model configs must
+        # follow: otherwise `Trainer` realigns them at train time and reports it as a change the user did not make.
+        dataset = load_dataset("trl-internal-testing/zen", "standard_preference", split="train")
+
+        training_args = CPOConfig(output_dir=self.tmp_dir, report_to="none")
+        trainer = CPOTrainer(
+            model="trl-internal-testing/tiny-MistralForCausalLM-0.2", args=training_args, train_dataset=dataset
+        )
+
+        pad_token_id = trainer.processing_class.pad_token_id
+        assert pad_token_id is not None
+        assert trainer.model.config.pad_token_id == pad_token_id
+        assert trainer.model.generation_config.pad_token_id == pad_token_id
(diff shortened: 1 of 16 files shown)


albertvillanova:
Merging this PR as a simple consistency alignment between experimental and core trainers.

PR #7001, opened 2026-09-01
Add support for vLLM 0.28.0

qgallouedec:
Checked the v0.28.0 release notes against what TRL uses. One breaking change, in weight transfer; everything else (sleep/wake, dev-mode HTTP routes, `vllm serve` flags, network utils) is unchanged.

### Breaking: static trainer-side weight transfer removed (vllm-project/vllm#50902, tracked in #6732)

0.28.0 drops `NCCLWeightTransferEngine.trainer_init` / `trainer_send_weights` and `NCCLTrainerSendWeightsArgs`, and moves `packed` from the per-update info to the init info. On main, `import trl.experimental.async_grpo` fails against 0.28.0.

Fix: the helpers those statics wrapped still exist (`nccl_common.trainer_init`, `packed_tensor.packed_nccl_broadcast_producer`), so `VLLMClient` and the two async `WeightTransferClient`s call them directly behind `is_vllm_available(min_version="0.28.0")`, and put `packed: True` in the init info on >= 0.28.0 (update info below). Same NCCL group, same packed broadcast, same threading. The branch goes away when 0.27.1 ages out (Jan 2027).

Not migrating to the new stateful `NCCLTrainerWeightTransferEngine` here: it only exists from 0.28.0, and its one-shot `send_weights()` fits neither `VLLMClient.weight_update()` (per-tensor calls, Online DPO) nor the async clients' HTTP-on-main-thread timeout design. Still #6732.

## Support window

image

0.28.0 released 26 Aug 2026, ages out ~26 Jan 2027. Next drop: 0.19.0 on **3 Sep 2026**.

### Tests

4×H100, torch 2.13.0+cu130. `tests/test_vllm_client_server.py`: **39 passed, 1 xfailed** on 0.28.0 and on 0.27.1 (both branches). Async GRPO 3-step smoke against `vllm serve` 0.28.0: 4 weight syncs, all 200, `weight_sync_s` 0.13.

Side note: the TP=2 server takes 184-270 s to start on hopper-prod on either version, right at the test's 240 s `connection_timeout`; I raised it locally for these runs, not committed.

### 0.28.0

```
$ pytest -v tests/test_vllm_client_server.py
========== test session starts ==============================
platform linux -- Python 3.13.13, pytest-9.1.1, pluggy-1.6.0 -- /fsx/qgallouedec/miniconda3/envs/trl/bin/python
cachedir: .pytest_cache
rootdir: /fsx/qgallouedec/trl-wt-vllm028
configfile: pyproject.toml
[...]

PR #7131, opened 2026-09-09
[AsyncGRPO] Gather FSDP2 weights before NCCL send to vLLM

### LINKED ISSUE #7103: AsyncGRPO FSDP2 weight sync hangs: interleaved full_tensor() and is_checkpoint_format=True
## Summary

`AsyncGRPOTrainer` + FSDP2 hangs on the first (and later) NCCL weight syncs to `vllm serve` whenever more than one trainer rank participates. This is separate from the offload / server-OOM hangs in #6797, though the symptom is the same: NCCL watchdog, no useful trainer error.

Hit on a 4×8 H200 cluster (16 trainer + 16 sampler, Qwen3-32B and Qwen3-1.7B, FSDP2 + activation checkpointing, vLLM 0.26, `weight_sync_steps=1`). After the fixes below, 3-step and 10-step 16T+16S smokes complete and weight sync is ~2s on 32B.

### 1. FSDP collectives interleaved with rank-0 vLLM send

`_sync_weight` still does:

```python
    self.weight_transfer.send_weights(self._streaming_iter())
else:
    for _ in self._streaming_iter():
        pass
```

`full_tensor()` is a default-PG all-gather. `send_weights()` then blocks rank 0 on a **different** NCCL group (trainer ↔ vLLM) and on HTTP (`/update_weights`). Non-0 ranks immediately issue the next `full_tensor()`. Rank 0's last-enqueued default-PG work falls behind and the watchdog fires, e.g.

```
Watchdog caught collective operation timeout: WorkNCCL(..., OpType=COALESCED, ...)
Rank 0 last enqueued work << other ranks
```

`init_weight_transfer()` is also rank-0-only with no barrier before `_sync_weight()`, so other ranks can enter the first gather while rank 0 is still in the vLLM NCCL handshake.

### 2. `start_weight_update(is_checkpoint_format=True)`

NCCL streaming is not a checkpoint-file reload. With `True`, vLLM workers wait for a checkpoint path that never arrives, so `/update_weights` never returns after the broadcast (HTTP timeout on the trainer).

A third mismatch that hangs **packed** NCCL: FSDP2 + AC upcasts parameters to fp32 and injects `_checkpoint_wrapped_module` into names. vLLM still expects the pre-wrap bf16 names/dtypes.

## Expected

All FSDP ranks finish `full_tensor()` together, then rank 0 sends already-gathered bf16 tensors. `start_weight_update` must use `is_checkpoint_format=False`.

## Environment

- trl `main` (reproduced on current `_streaming_iter` / `is_checkpoint_format=True`)
- torch 2.11 + cu130, vLLM 0.26, Accelerate FSDP2
- 16 trainer ranks, 16 sampler GPUs, TP=4

Related: #6797 (offload `pack_tensors` hang and missing timeout; different root cause).

sfc-gh-truwase:
# What does this PR do?

`AsyncGRPOTrainer` + FSDP2 hangs on NCCL weight sync whenever more than one trainer rank participates. Two bugs compose:

1. `_sync_weight` interleaved `full_tensor()` (default process group) with rank-0 `send_weights()` (vLLM NCCL + HTTP). Non-0 ranks raced ahead on the next all-gather and the trainer process group timed out.
2. `start_weight_update(is_checkpoint_format=True)` is wrong for NCCL streaming: vLLM workers wait for a checkpoint path that never arrives, so `/update_weights` never returns.

This PR gathers every FSDP shard on all ranks first, then rank 0 sends already-gathered tensors. It also casts to the mixed-precision dtype, strips `_checkpoint_wrapped_module` from names so packed NCCL matches the server, and barriers after rank-0-only `init_weight_transfer()`.

Fixes #7103

Replaces #7104 (closed by the first-time-contributor template check).

Verified on 16 trainer + 16 sampler H200 (Qwen3-1.7B and Qwen3-32B FSDP2). After this path, 32B weight sync is ~2.3s and training proceeds past the first step.

## Before submitting

## AI writing disclosure

## Who can review?

@kashif

Made with [Cursor](https://cursor.com)


albertvillanova (commented):
Thanks for the detailed issue and for running this at 16T+16S scale, that kind of report is hard to come by. I went through the four changes one at a time against the vLLM sources, and I think only the first one is doing work. Details and evidence below.

## 1. Gather-then-send: right diagnosis, but it drops the memory bound

The underlying concern is real. On rank 0 the generator is consumed from the NCCL daemon thread, so default-PG `full_tensor()` all-gathers interleave with vLLM-PG broadcasts and with the blocking `/update_weights` HTTP call, while non-0 ranks run ahead. That is a plausible watchdog-timeout shape.

But `_gather_weight_items()` deletes exactly the property the replaced code was written to have. The comment it removes says so: *"frees it once the generator advances, avoiding materializing the full model in memory"*. Every gathered parameter is now held in a list on rank 0 before a single byte is sent, so for the Qwen3-32B case in the description that is on the order of 64 GiB of bf16 on rank 0's device, on top of its shard, optimizer state and activations. It is asymmetric too: rank 0 OOMs while every other rank is comfortable. It fits on H200s; it will not fit on 80 GB parts, which is the common configuration.

The interleaving can be removed without giving that up, for instance by keeping the per-parameter gather loop in lockstep on all ranks and handing rank 0's tensors to the send thread through a small bounded queue. Only a few tensors are live at a time and the all-gathers stay ordered across ranks.

## 2. `is_checkpoint_format=True` -> `False` is incorrect, and inert on the wire

`is_checkpoint_format` does not mean "reload from a checkpoint path". From `vllm/v1/worker/gpu_model_runner.py::reload_weights`:

> `is_checkpoint_format`: set to False if weights have already been processed into kernel format (repacking, renaming, etc.)

`True` routes the stream through `model.load_weights()`, which does the HF-to-vLLM name mapping and the fused `qkv_proj` / `gate_up_proj` packing. `False` takes the other branch:

```python
for name, loaded_weight in weights_iterator:
    param = _get_parameter_for_reload(model, name)   # model.get_submodule(...)
    param.copy_(loaded_weight)
```

with a warning that the weights must be *"in kernel format and already sharded"*. TRL streams HF-named, unsharded tensors (`model.layers.0.self_attn.q_proj.weight`), so `False` would look up a submodule vLLM does not have and copy TP-unsharded shapes. `True` is the correct value.

The reason your run still worked is that the flag never reaches the worker. In vLLM 0.26, 0.27 and current main the route is:

```python
@router.post("/start_weight_update")
async def start_weight_update(raw_request: Request):
    await engine_client(raw_request).start_weight_update()   # body is never read
```

and `Worker.start_weight_update(self)` takes no such argument. The parameter was dropped from the route after 0.24. So on vLLM >= 0.26 this change is a no-op, and on 0.22 to 0.24, still inside TRL's `>= 0.22.0` floor, it would select the direct-copy path and break the sync. `reload_weights` already defaults to `True`, which is what we want.

## 3. `_sync_weight_dtype(args)` reads the wrong knob

Model precision in this trainer is `AsyncGRPOConfig.dtype` (`"auto" | "bfloat16" | "float16" | "float32"`, default `"float32"`), not the `bf16` / `fp16` autocast flags. The two are orthogonal, so the helper is wrong in both directions:

* `dtype="bfloat16"` with `bf16=False`: every parameter is upcast to fp32 before the send, the manifest advertises `float32`, the NCCL payload doubles, and the dtype-mismatch warning in `init_weight_transfer` fires against a bf16 server.
* `dtype="float32"` (the default) with `bf16=True`: fp32 master weights are silently downcast to bf16 on the wire, which is precisely the trainer-generator precision gap the `float32` default exists to close (the config docstring cites https://huggingface.co/papers/2510.26788 for it).

If a cast is genuinely needed under FSDP2 mixed precision, it should target the dtype the manifest already records, so the two sites cannot disagree by construction. Separately, `args: Any` plus two `getattr` calls on a dataclass whose fields always exist runs against the repository's guidance on defensive `hasattr` / `getattr`.

## 4. `_vllm_param_name` is not a behaviour change

The description lists `_checkpoint_wrapped_module` stripping as part of the fix, but it is already on main, and the new two-step replace is equivalent to the existing one-liner on every name shape I tried, including nested wrappers and the leading-prefix case:

```
module.model.layers.0.self_attn.q_proj.weight                       same
model.layers.0._checkpoint_wrapped_module.self_attn.q_proj.weight   same
_checkpoint_wrapped_module.model.embed_tokens.weight                same
model.layers.0._ckpt_wrapped._ckpt_wrapped.x.weight                 same
```

So `test_vllm_param_name_strips_wrappers` passes against unmodified main. It would be worth checking that a new test fails without its fix. Note also that main now derives these names in one place, `_iter_vllm_named_params`, so a second helper in AsyncGRPO and a third in distillation moves away from that.

## The 4 GiB pack buffer

This one is not in the description, and I do not think it is a correctness fix. vLLM's `pack_tensors` treats `buffer_size_bytes` as a soft threshold: an oversize tensor only triggers `warnings.warn`, and producer and consumer use the identical `total_bytes > buffer_size_bytes` break rule, with the consumer allocating the actual chunk size rather than the nominal buffer. A 1.56 GB Qwen3-32B `embed_tokens` row therefore transfers correctly at the 1 GiB default; it lands in its own chunk and logs a warning.

What the bump costs is real: chunks grow to about 4 GiB, and with `packed_num_buffers=2` in flight that is roughly 8 GiB of transient allocation on the trainer and on every vLLM worker, against a server whose OOM headroom the existing error message already calls out.

It also breaks vLLM >= 0.28. `with_packed_nccl_buffer()` injects the key into the per-round update info, but 0.28 moved it to the init info. `NCCLWeightTransferUpdateInfo` now carries only `names` / `dtype_names` / `shapes`, and its docstring states that the buffer geometry *"is a must-agree wire param carried on the init info ... not here"*. The server does `self.update_info_cls(**update_dict)` in `parse_update_info`, so the extra key surfaces as `ValueError: Invalid update_info`. Main had to stop sending `packed` in that same payload for this reason (`weight_transfer.py`, the `_HAS_STATEFUL_TRAINER_ENGINE` branch), so this reintroduces the same class of key.

## Smaller points

* The `wait_for_everyone()` added after `init_weight_transfer()` cannot be load-bearing: `_sync_weight` does a rank-0-only `pause()` and then `wait_for_everyone()` before any gather, so non-0 ranks already block there. The justifying comment describes a race the existing barrier prevents.
* Code quality would fail CI as it stands. `ruff check` reports two import-sorting errors (the new `vllm_client` / `weight_transfer` imports sit before `async_rollout_worker` in both test files) and `ruff format` reformats one over-length assert in `test_async_distillation_trainer.py`.
* The two `weight_transfer.py` copies get different comments above the same constant, and the `_weight_update_info` assignment gets a comment in AsyncGRPO but not in distillation. Duplicated blocks in this repository are expected to match word for word.
* The branch is behind main in a way that matters here: #7017 split `_sync_weight` into `_sync_weight_merged` / `_sync_weight_lora` and centralised name derivation, and #7001 added the vLLM 0.28 path. The AsyncGRPO side will need reworking rather than a textual rebase.

## Suggestion

The interleave you identified in #7103 is worth fixing and the evidence for it is good. I would keep that one change, reshape it so the gather stays bounded rather than materialising the whole model on rank 0, and drop the other four, since three of them are no-ops or regressions and the pack-buffer change breaks 0.28. That also makes the 16T+16S validation attributable to the change it actually exercised.

sfc-gh-truwase:
@albertvillanova thanks for the review and flagging these problems. Will address them.

sfc-gh-truwase:
@albertvillanova please take another look

PR #6192, opened 2026-06-27
Align CPO/ORPO/BCO/Distillation/TPO with DPO: fix ZeRO-3 + PEFT mixed-dtype error

DaoyuanLi2816:
Align the remaining experimental trainers with DPO: fix the ZeRO-3 + PEFT dtype mismatch for non-quantized models.

#6091 fixed this `TypeError` for the core trainers (DPO/GRPO/RLOO/Reward/SFT) and #6093 extended it to KTO. The same `get_peft_model(model, peft_config)` call without the guard is still present in the **CPO, ORPO, BCO, Distillation and TPO** experimental trainers, so they remain affected. This PR applies the identical fix to those five trainers.

Follow-up to:
- #6091
- #6093

Part of:
- #4786

### Changes

**PEFT integration and dtype handling:**

* Detect quantized models via the `is_loaded_in_4bit` / `is_loaded_in_8bit` attributes, stored in `_is_quantized_model`.
* When using ZeRO-3 with PEFT on a non-quantized model (and `peft >= 0.12.0`), pass `autocast_adapter_dtype=False` to `get_peft_model` so LoRA adapter parameters stay in the base model's dtype (e.g. `bf16`) instead of being upcast to fp32. This prevents the mixed-dtype `TypeError` raised by ZeRO-3's `_allgather_params_coalesced` on the first optimizer step.
* The guard is a no-op outside that exact scenario: single-GPU / non-ZeRO-3 / quantized / older-PEFT runs keep calling `get_peft_model(model, peft_config)` unchanged, and each trainer's existing QLoRA bf16 handling is left in place.

**Dependencies:**

* Import `peft` when available, to support the version gate (and `packaging.version.Version` in the distillation trainer).

**Documentation and comments:**

* Added the same explanatory comment block used in DPO/KTO, with links to the relevant TRL (#6089) and upstream DeepSpeed issues.

### Verification

* `ruff check` and `ruff format` pass on all five files.
* The existing `@require_peft` tests for CPO, ORPO, BCO and TPO pass locally (11 passed), exercising the modified `get_peft_model` path on the single-GPU (no-op) branch where `autocast_adapter_dtype` is not passed, confirming no regression to the default code path.

---

> [!NOTE]
> **Low Risk**
> Narrow, conditional change at PEFT wrap time; mirrors an already-shipped pattern in core trainers with no-op behavior outside ZeRO-3 + non-quantized PEFT.
> 
> **Overview**
[...]

PR #5665, opened 2026-04-27
Fix missing PEFT availability check when passing peft_config to experimental trainers

albertvillanova:
Fix missing PEFT availability check when passing peft_config to experimental trainers.

This PR adds improved error handling for cases where a `peft_config` is provided without the required `peft` library installed in both the distillation and TPO experimental trainers. This ensures users receive a clear, actionable error message if they attempt to use PEFT features without having the necessary dependencies.

Related to:
- #5664

### Changes

Error handling improvements for PEFT:

* Added a check in `trl/experimental/distillation/distillation_trainer.py` to raise an `ImportError` with installation instructions if `peft_config` is provided but the `peft` library is missing.
* Added a similar check in `trl/experimental/tpo/tpo_trainer.py` to ensure the `peft` library is installed when `peft_config` is used, raising an informative `ImportError` if not.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to trainer initialization validation and error messaging around PEFT, with minimal impact beyond earlier/clearer failures for misconfigured setups.
> 
> **Overview**
> Standardizes PEFT setup across experimental trainers by **validating `peft_config` early** and failing fast with clearer errors.
> 
> When `peft_config` is provided, trainers now (1) raise an actionable `ImportError` if `peft` isn’t installed (with `pip install trl[peft]` guidance) and (2) enforce that `peft_config` is a `peft.PeftConfig` instance (e.g. `LoraConfig`), before proceeding; several trainers also update PEFT imports to include `PeftConfig` and keep the existing guard against passing an already-wrapped `PeftModel` alongside a new config.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 70b898fc295cba624ac507f0f6c9e06ad8b868a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #5629, opened 2026-04-22
Use `PreTrainedTokenizerBase` for tokenizer type hints

qgallouedec:
In TRL, type hints for tokenizer parameters should use `PreTrainedTokenizerBase`, not `PreTrainedTokenizer`.

In transformers, `PreTrainedTokenizer` is an alias for `PythonBackend` (slow tokenizer only). The fast tokenizer (`TokenizersBackend`) inherits directly from `PreTrainedTokenizerBase` and is NOT a `PreTrainedTokenizer`. Since `AutoTokenizer.from_pretrained(...)` returns a fast tokenizer by default for most models, `PreTrainedTokenizer` hints fail to cover the common case. `PreTrainedTokenizerBase` is the common ancestor and exposes `.apply_chat_template`, `.chat_template`, `.vocab`, `.eos_token`, etc.

**How to apply:** When adding or editing a function that accepts a tokenizer (not model-specific), use `PreTrainedTokenizerBase`. Only use `PreTrainedTokenizer` if the code truly requires slow-tokenizer-only behavior (rare).

---

> [!NOTE]
> **Low Risk**
> Low risk type-hint-only change; runtime behavior should be unchanged, but downstream type checkers may surface new/changed annotations.
> 
> **Overview**
> Broadens tokenizer type annotations from `PreTrainedTokenizer` to `PreTrainedTokenizerBase` across chat-template utilities (e.g. `clone_chat_template`, `get_training_chat_template`, `parse_response`) and BCO dataset tokenization (`_tokenize`) so fast tokenizers and processors are covered.
> 
> Cleans up related typing imports (drops `TYPE_CHECKING`/`Optional` usage) and updates the `ProcessingClassT` `TypeVar`/docstrings to match the new base type.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 87c2dbf1337b90a99cef6235aa7cd8b08450b84e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #7136, opened 2026-09-09
Remove the experimental BCO trainer

qgallouedec:
Removes `trl.experimental.bco` (`BCOTrainer`, `BCOConfig`), added in #1599.

## Why

**Low adoption.** 11 genuine trainings in the last 2 months as per our telemetry, and 1 model created this year: https://huggingface.co/models?other=bco,trl&sort=created

**High maintenance cost for that.** 1,600 lines carrying their own reference-model handling, DeepSpeed preparation, tokenization, and eval loop, plus a UDM density-ratio classifier that is the only reason TRL depends on scikit-learn and joblib at all. 106 commits touched it; the recent ones are all upkeep (#7102 peft floor, #6507 `TQDM_DISABLE`, #6192 ZeRO-3 + PEFT mixed dtype, #5665/#5666/#5683 PEFT plumbing, #5629 tokenizer type hints, #5567 processor support). It left the stable API 2025-10-21 (#4312) and the deprecated alias was dropped 2026-02-10 (#5045); no feature work since.

## Notes

`trl.experimental` has no stability guarantee, so no deprecation cycle. The code stays in git history for anyone wanting to revive it.

The `bco_pair` DPO loss stays: it is the paired form of the same objective, is used by MPO, and does not need the trainer. The [paper](https://huggingface.co/papers/2404.04656) entry stays in `paper_index.md` for that reason, minus the pointer to the unpaired trainer.

It is not a migration path though. `bco_pair` is a paired loss on `DPOTrainer`, and unpaired data is exactly what `BCOTrainer` was for. Those users go to `KTOTrainer`, same unpaired preference dataset type (same row in the dataset format table), in the stable API. UDM has no equivalent there: the density-ratio reweighting does not carry over.

BCO was the only consumer of the rest of what goes with it: the `bco` and `scikit` optional dependency groups (`scikit-learn`, `joblib`), `is_joblib_available`, and the `require_sklearn` test marker.

---

> [!NOTE]
> **Medium Risk**
> Removes a public experimental API (`BCOTrainer`) without a deprecation cycle; users on unpaired BCO must switch trainers, though paired `bco_pair` on DPO is unaffected.
> 
> **Overview**
> **Removes the experimental `BCOTrainer` / `BCOConfig` stack** (`trl.experimental.bco`), including ~1,600 lines of trainer code, the full test suite, and the `bco_trainer` docs and navigation entries.
> 
> Docs are updated so BCO is no longer documented as a standalone trainer: the dataset-format table and index drop `BCOTrainer`, and DPO/`paper_index` no longer point unpaired users at the removed trainer. **`DPOTrainer`’s `loss_type="bco_pair"` is unchanged** (paired BCO objective still lives in DPO loss code).
> 
> Packaging and test infra shed BCO-only dependencies: optional extras `bco` and `scikit` (`scikit-learn`, `joblib`), plus `is_joblib_available`, `require_sklearn`, and `BCOTrainer` from the trainer registry. Unpaired preference workflows are expected to use **`KTOTrainer`** instead; UDM density-ratio reweighting from the old trainer has no direct replacement.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac92491759fa1c3553572d68264c242f866cba45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova (commented):
I'm not opposed to this, but I'd like us to be more careful about the reasoning, because whatever we settle here becomes the template for the other trainers sitting in `trl.experimental` for the same reason.

**On the telemetry number.** We should say exactly what is being counted because I read a different figure than the one here: 74 genuine users in the last month. Also, the window disqualifies the metric for this question:  every BCO data point we have was collected after we demoted it, removed it from the docs index and made it print an experimental warning. It cannot distinguish "nobody wants this" from "we told everyone to stop".

None of this means BCO is widely used. It probably isn't. It means the numbers as presented can't carry the weight of the decision, and low usage of something we stopped supporting is partly something we caused. That inference deserves caution generally: it can reflect limited interest, but also gaps in support or usability. For BCO specifically, UDM (the thing that distinguishes it from KTO) requires the user to bring their own embedding model, embedding tokenizer and `prepare_model` glue, with no example and no CLI path. Near-zero adoption is what I would expect from that regardless of demand.

**On the maintenance cost.** The commit count overstates the marginal cost and invites the obvious rebuttal. Of the commits touching the BCO path, only a small fraction are BCO-specific; the rest are repo-wide sweeps (#5802 touched 88 files, #6978 42, #6180 24), where BCO's cost is one more file in a sed-and-review pass. The real argument is the one our contributing guidance makes: duplication is accepted but consistency is mandatory, so each extra trainer is a permanent tax on every sweep, and 1,600 lines of copied reference-model, DeepSpeed and tokenization code is a large surface to keep aligned. That case stands on its own. Likewise, scikit-learn and joblib are optional extras behind a lazy availability check, only needed when `embedding_func` is passed, so stock `pip install trl` never pulled them; the win is deleting two extras, not shrinking anyone's install.

**What I'd rather do.** #4223 moved these trainers with the stated plan to discuss later which get promoted, which stay and which get removed. That discussion never happened, #4223 is closed, and we are now making the call one PR at a time in threads only the people who already agree are reading.

2 ideas:
1. Open a follow-up RFC and settle it once for everything in `trl.experimental`, agreeing the bar before the list so each case is decided by a rule rather than re-argued: time in experimental, no open issues or PRs from non-maintainers, no downstream consumers, and notice in the previous release. Worth separating the two populations in there too, since they arrived under different terms: the trainers we demoted in the #4223 wave had users on the stable API who never opted into "may vanish in any release", while everything incubated directly into experimental did. Notice is owed to the first group and not the second.
   - Happy to open it if you agree on this
1. What about implementing a "removal" cycle (analog to the deprecation one) for every trainer we plan to remove? Give BCO one release of notice and merge it: a `FutureWarning` in `BCOTrainer.__init__` naming `KTOTrainer` as the replacement and the target version, plus a line in the release notes. 
   -  That is the same three lines that used to sit in `trl/trainer/bco_trainer.py` until #5045, which told BCO users the old import path would go away in 0.29 and pointed them at #4223.
   - One cycle turns "removed without warning" into a record we can point at, and it gives us the one measurement telemetry can't: whether anyone turns up. If nobody does, the removal is unarguable.

CC: @huggingface/trl

qgallouedec:
Thks, fair on the reasoning, I'll fix the body.

On the telemetry gap: unfiltered I get the same ~70 you do, dropping the runs that report `device=cpu` takes it to ~10. Nobody genuinely trains on CPU, and #6996 shows what those rows are, the repro there was run on CPU by someone hunting bugs.

Which is the cost I'd really want in the RFC. #6996 and #6999 are an open issue plus a +145/-16 fix for the BCO encoder-decoder path, and the reporter said it plainly when I asked how he found it: 
> Bug hunt, not real use. I have never trained a BCO encoder-decoder model, and there is no user waiting on this that I know of.

Not isolated: 14 issues from that account since June, 11 still open, plus #6572 and #6581, the same BCO fix submitted twice. My point is triage costs, and every trainer we keep is more surface for machine-generated findings on code nobody runs. That is what's eating maintainer time.

It's also why I don't read `trl.experimental` as somewhere things can sit indefinitely. The overview frames it as a staging area with a conditional promotion path, removable in any release without prior deprecation, (+ #4223 has listed BCO as "May be removed" since 2025-10-07). Parking trainers there forever isn't the spirit of it.

So I'd push back on the `FutureWarning`. Not because notice is worthless, but because a cycle is a process we then have to run: pick the target version, land the warning, remember it next release, chase it when the release slips, answer this again in the removal PR. For 1,600 lines with no users, that overhead is exactly what we're trying to get back. A line in the release notes is enough.

RFC: yes, please open it, and let's settle the demoted vs incubated split there.

albertvillanova (commented):
That reconciles it, thanks. ~70 unfiltered is what I was reading, and the CPU filter explains the rest. Worth putting both numbers and the filter in the body, and the same CPU-filtered count for one trainer nobody disputes, so a reader can calibrate without asking

On triage, you're right and my commit-count point doesn't touch it. Sweeps are cheap, attention isn't, and the volume scales with agents deployed rather than with users. #6999 is the cost in the present tense: an open +145/-16 on the encoder-decoder path that someone would have to review, for a path the author says nobody is waiting on. Merging this makes that review and #6996 moot, which is a good argument.

Fair on the `FutureWarning` too. Your objection is to the two-step, and that's an accurate description of what a manual cycle costs. A release-notes line is proportionate here. One small ask: point that line at the RFC, so the notice doubles as an invitation to the discussion where it actually matters.

The thing I'd carry into the RFC is that the triage argument isn't really about BCO. Machine findings land on unmaintained code, and there are ~25 modules in `trl.experimental` that fit. This removes one surface. Followed to its conclusion, if report volume decides what stays, then whoever runs the most agents is choosing what TRL keeps. I'd rather we set the bar ourselves. That's also the answer to the process cost you describe: picking a version, landing a warning, remembering it, chasing the slip, re-arguing it in the removal PR is the cost of deciding case by case. Deciding once is how it goes away.

Opening the RFC now, with the demoted vs incubated split and triage surface as one of the axes.

albertvillanova:
RFC:
- #7182

albertvillanova (approved):
Approving. My earlier concerns were about the reasoning rather than the outcome, and the numbers you added settle them: the CPU filter explains the gap I was reading, and every axis from #7182 points the same way here. Usage is at the bottom among modules old enough to have been discovered, the one issue filed against it came from a bug hunt rather than from use, nothing in the major training frameworks imports `BCOTrainer`, no owner turned up in either thread, and KTO covers the same unpaired preference data in the stable API.

One thing worth adding to the body before merging, not a blocker: `bco_pair` staying is not a migration path, since it is the paired loss on `DPOTrainer` and the unpaired workflow is precisely what `BCOTrainer` was for. `KTOTrainer` is where those users go, same dataset type, now stable. Worth saying so, with the caveat that UDM has no equivalent there, so the record shows what does and does not carry over.

qgallouedec:
thanks done

Issue #7137, opened 2026-09-09
Align tiny test model configs with their reference models


albertvillanova:
### What

The tiny models under `trl-internal-testing` are built by the scripts in `scripts/generate_tiny_models/` from architecture arguments only:

```python
config = Qwen3Config(
    vocab_size=len(tokenizer.vocab),
    hidden_size=8,
    num_attention_heads=4,
    num_key_value_heads=2,
    num_hidden_layers=2,
    intermediate_size=32,
)
```

Everything not passed falls back to the config class defaults, while the tokenizer and generation config are copied from the real model. So a tiny model can carry `eos_token_id=None` where the model it stands in for uses `151645`, a scalar `eos_token_id=2` where the real one ships a list of three ids, generic RoPE where the real one uses `rope_type: llama3` frequency scaling, or an untied `lm_head` where the real one ties.

The consequence is that CI exercises configurations that do not exist in the wild, and code paths every user hits have no coverage at all.

`print_config_diff` already prints these differences on every generation run, and `_DIFF_IGNORE` never filtered them. They were simply never acted on, apart from four one-off PRs: #5638, #5706, #5707 and #5716.

### Goal

For each tiny model, the only differences against its reference should be the deliberate size reduction (`hidden_size`, `num_hidden_layers`, `intermediate_size`, `num_attention_heads`, `num_key_value_heads`, `head_dim`) plus fields derived from them, such as `layer_types`, whose length tracks `num_hidden_layers`.

`vocab_size` is mirrored despite affecting size: it is a matrix dimension rather than a token count, and reproducing the reference's padding is part of the fidelity.

### Open

- [x] #7130 `tie_word_embeddings`, for the four tiny models where it diverges
- [x] #7134 Qwen2.5-Coder
- [x] #7135 DeepSeek-R1-Distill
- [x] #7138 FalconMamba, correcting the reference
- [x] #7139 FalconMamba config, on top of #7138

The eight tiny models aligned earlier are tracked on #7093, since they also reduced the warning noise there.

### Notes

No tiny model change is needed for Cohere2, Nemotron or DeepSeek-V3: their configs already match their references, and the residual difference comes from the reference disagreeing with its own tokenizer.

albertvillanova:
`tiny-FalconMambaForCausalLM` cannot be aligned the way the others were, because it is compared against the wrong model.

[...]

PR #2287, opened 2024-10-27
🤏 New models for tests

qgallouedec:
# What does this PR do?

Fixes #2177

The models currently used for testing lack uniformity and practicality. Specifically:  
- They are sourced from various namespaces (`trl-internal-testing`, `facebook`, `philschmid`, `hf-internal-testing`, etc.).  
- They are not all small.  
- Their naming conventions are inconsistent (e.g., `tiny-random-MistralForCausalLM`, `dummy-GPT2-correct-vocab`, `gpt2`, `pythia-14m`).  
- There is no existing script for generating these models.  

This PR introduces the following improvements:  
- A script to create tiny models specifically for testing purposes.  
- A uniform naming convention for these models, all placed under a single namespace (`trl-internal-testing`).

When approved, I'll move every models into `trl-internal-testing` namespace (instead of `qgallouedec`). And probably remove all the old testing models in `trl-internal-testing`.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/trl/tree/main/docs).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #7138, opened 2026-09-09
Point the tiny FalconMamba generator at a FalconMamba reference

albertvillanova:
`tiny-FalconMambaForCausalLM` is generated against `tiiuae/falcon-7b-instruct`, a Falcon attention model, while the tiny model itself is `FalconMambaForCausalLM`, a Mamba SSM. This PR points it at `tiiuae/falcon-mamba-7b-instruct`.

It is deliberately a one-line change. Regenerating the tiny model on top of it picks up the tokenizer, chat template and generation config from the right model family, which is the substance of the fix. Aligning the config values follows in #7139, so the tokenizer change and the config change can be reviewed and regenerated separately.

Part of #7137.

@qgallouedec you introduced this entry in #2287, so I would value your opinion on the choice of reference, and on whether the mismatch was intentional for some reason I have missed.

### How it got here

The pairing has been wrong since the entry was first written. `453db5cd` (2024-11-25, #2287) created the single-file generator with:

```python
("tiiuae/falcon-7b-instruct", FalconMambaConfig, FalconMambaForCausalLM, None),
```

The list is `(model_id, config_class, model_class, suffix)`, so a Falcon repo id was paired with the FalconMamba classes from the start. That entry then survived unchanged, including through the split into per-model scripts in #5637, which carried it over faithfully. `tiiuae/falcon-mamba` has never appeared anywhere in the repository: `git log --all -S "falcon-mamba"` returns nothing.

It is an understandable choice rather than a random one. `tiiuae/falcon-7b-instruct` supplies a working tokenizer with a chat template and the right 65024 vocabulary, so the generated model loads and passes its smoke test, and nothing downstream cared which config the numbers came from. `print_config_diff` only arrived with #5637, and a 46-row diff scrolls past exactly like a 6-row one.

### Why the instruct variant

Both FalconMamba releases were candidates:

```
tiiuae/falcon-mamba-7b            chat_template=NO   bos=0 eos=11 pad=11
tiiuae/falcon-mamba-7b-instruct   chat_template=yes  bos=8 eos=11 pad=0
```

The tiny model has a chat template today, inherited from `falcon-7b-instruct`, and it is used where that matters: it appears in `tests/test_chat_template_utils.py` as `id="falconmamba"`, and in `test_sft_trainer.py` and `test_data_utils.py`. Pointing at the base variant would remove a capability the tiny model currently has.

This also matches the rest of the series. Of the 35 tiny causal-LM models, 32 have a chat template; the three without are `tiny-GPT2LMHeadModel`, `tiny-GPTNeoXForCausalLM` and `tiny-OPTForCausalLM`, whose families have no chat release. The naming is not the guide here: only five references carry "Instruct" in the name, yet `Qwen/Qwen3-8B`, `zai-org/GLM-4.5`, `deepseek-ai/DeepSeek-R1` and the Nemotrons all ship chat templates.

Checked that the assertion in the tool-calling suite still holds, since a different template could have changed it:

```
trl-internal-testing/tiny-FalconMambaForCausalLM  supports_tool_calling=False
tiiuae/falcon-mamba-7b-instruct                   supports_tool_calling=False
```

### Effect on the diff

The row count against the old reference was meaningless, since the two config schemas are unrelated: `alibi`, `multi_query`, `parallel_attn` and `ffn_hidden_size` on one side against `state_size`, `conv_kernel`, `expand` and `time_step_rank` on the other.

```
[config_diff] tiiuae/falcon-7b-instruct        vs tiny  46 differences
[config_diff] tiiuae/falcon-mamba-7b-instruct  vs tiny  11 differences
```

The correction alone accounts for 35 of them. The remaining 11 are what #7139 addresses, taking it to 6, all of them the deliberate size reduction.

### Note on regeneration

Unlike the other models in this series, this one changes the tokenizer as well: the two repos ship different `tokenizer.json`, `tokenizer_config.json` and `special_tokens_map.json`. That is unavoidable once the reference is corrected, and it is the reason this is a PR of its own rather than being bundled with the config alignment.

### DIFF
--- a/scripts/generate_tiny_models/for_causal_lm/falcon_mamba_for_causal_lm.py
+++ b/scripts/generate_tiny_models/for_causal_lm/falcon_mamba_for_causal_lm.py
@@ -27,7 +27,7 @@ from .._common import (
 
 check_transformers_version()
 
-MODEL_ID = "tiiuae/falcon-7b-instruct"
+MODEL_ID = "tiiuae/falcon-mamba-7b-instruct"
 
 tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
 generation_config = GenerationConfig.from_pretrained(MODEL_ID)


qgallouedec (approved):
Right ref, thanks

I can't remember if there were a reason behind #2287, FalconMamba came later and I probably reused the nearest Falcon entry. Splitting the reference from the values is right. LGTM.

albertvillanova:

- [x] https://huggingface.co/trl-internal-testing/tiny-FalconMambaForCausalLM/discussions/1

albertvillanova:
Tested with `MODEL_REVISIONS` pointing at the regenerated model, `trl-internal-testing/tiny-FalconMambaForCausalLM` at `refs/pr/1`, on top of current `main`. The suite is green: https://github.com/huggingface/trl/actions/runs/34463973072

PR #7167, opened 2026-09-10
Reword the token sync comments to say what the line does

albertvillanova:
This PR rewords the comment that accompanies the token sync added in #7097, #7128 and #7127, following the review on https://github.com/huggingface/trl/pull/7127#issuecomment-5603967168.

### Motivation

The comment argues why the line is safe rather than saying what it does:

```python
# Mirror the pad token onto the model configs: `Trainer` runs the same alignment at train time, so the end
# state is unchanged, but the model stays consistent with the tokenizer from the moment it is built.
```

"The end state is unchanged" is a claim about what `Trainer` does elsewhere, and it reads oddly next to a line whose whole purpose is that the state before `train()` is *not* the same. A reader of the trainer wants to know why the assignment is there, not why it is harmless.

### Solution

Say what the line does, and keep the parts that carry information a reader cannot get from the code:

```python
# The model must agree with the tokenizer on the pad token from construction, so mirror it onto the model
# configs.
```

The eos variants keep their second half, since the append semantics and the absent generation config on sequence classification models are both non-obvious:

```python
# The model must agree with the tokenizer on the eos token from construction, so mirror it onto the model
# configs. The generation config may hold several eos tokens, any of which halts generation, so the new
# one is added to the existing ones instead of replacing them.
```

```python
# The model must agree with the tokenizer on the eos token from construction, so mirror it onto the model
# config (a sequence classification model has no generation config).
```

Comments only. No code changes, no behavior changes.

### Changes

- Reword all 18 occurrences: 16 pad token ones across the SFT, DPO (x2), KTO, GRPO, RLOO and Distillation trainers and the nine experimental trainers, plus the two eos token ones in the SFT and Reward trainers

---

> [!NOTE]
> **Low Risk**
> Comment-only change with no runtime or training behavior impact.
> 
> **Overview**
> This PR **only rewords comments** next to the existing tokenizer→model token ID assignments (`model.config.pad_token_id`, `generation_config`, and eos variants). **No executable code changes.**
> 
> Across **SFT, DPO, KTO, GRPO, RLOO, Distillation, Reward**, and **nine experimental trainers**, the old text that argued the assignments were redundant with Hugging Face `Trainer` at train time is replaced with wording that states the **purpose**: keep **model config aligned with the tokenizer from construction**.
> 
> **Pad-token** sites use a shorter two-line comment. **EOS** paths in **SFT** and **Reward** keep extra notes (append-to-`eos_token_id` list behavior; no `generation_config` on sequence-classification models).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ae22e28ac0f25c3e8b0f97db814a8399928fe43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/experimental/async_distillation/async_distillation_trainer.py
+++ b/trl/experimental/async_distillation/async_distillation_trainer.py
@@ -990,8 +990,8 @@ class AsyncDistillationTrainer(_BaseTrainer):
             )
         if processing_class.pad_token is None:
             processing_class.pad_token = processing_class.eos_token
-        # Mirror the pad token onto the model configs: `Trainer` runs the same alignment at train time, so the end
-        # state is unchanged, but the model stays consistent with the tokenizer from the moment it is built.
+        # The model must agree with the tokenizer on the pad token from construction, so mirror it onto the model
+        # configs.
         model.config.pad_token_id = processing_class.pad_token_id
         model.generation_config.pad_token_id = processing_class.pad_token_id
 
--- a/trl/experimental/async_grpo/async_grpo_trainer.py
+++ b/trl/experimental/async_grpo/async_grpo_trainer.py
@@ -858,8 +858,8 @@ class AsyncGRPOTrainer(_BaseTrainer):
             )
         if processing_class.pad_token is None:
             processing_class.pad_token = processing_class.eos_token
-        # Mirror the pad token onto the model configs: `Trainer` runs the same alignment at train time, so the end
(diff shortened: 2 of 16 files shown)


albertvillanova:
Merging this PR as a simple rewording following a previous review.

Issue #7169, opened 2026-09-10
Where the GPU runner hours go in the Tests workflow


albertvillanova:
Opening this as a place to record measurements rather than as a bug report.

`tests.yml` books five GPU jobs per run: four on `aws-g6e-4xlarge` and one on `aws-g5-12xlarge-cache`, around 77 runner-minutes for a full matrix. That capacity is shared, so runner time spent on a run that cannot change any merge decision is queue time for every other PR in flight.

All figures below come from the Actions API and `git log`, over 2026-08-26 to 2026-09-10: a 15-day window covering 700 `tests.yml` runs and 114 commits landed on `main`. Roughly double them for a monthly rate.

## 1. Superseded PR runs are never cancelled

341 PR runs were superseded by a newer commit on the same branch and kept running to completion. Those runs consumed 352 GPU-hours in total, of which **64 GPU-hours were spent after the superseding run had already been created**, testing commits that were already obsolete.

`build_pr_documentation.yml` declares a `concurrency` group with `cancel-in-progress`. `tests.yml` and `tests-experimental.yml` do not. The 64 hours is the reclaimable part, not the full 352.

This one needs no decision: it copies a block the repo already uses, and the `head_ref` fallback leaves pushes to `main` and to `ci-*` branches untouched. Open as #7170.

## 2. The `paths:` filter applies to `pull_request` but not to `push`

`tests.yml` filters `pull_request` events by path, so a docs-only PR correctly shows no GPU jobs. The `push:` trigger carries no filter, so the same change runs the full matrix once it lands on `main`.

**27 of the 114 commits** that landed in the window would not have matched the `pull_request` filter. 10 of them touched nothing but `.md` files. At one matrix each that is roughly **35 GPU-hours**, spent testing commits the PR-side filter had already judged irrelevant.

This one is not a simple mirror of the filter: `push:` also covers `ci-*` branches, which are used to force a run on demand. Sharing one filter between the two triggers means deciding what a `ci-*` push should do when its diff does not match, and that is a policy question rather than a workflow edit.

Open as #7171, where the `ci-*` question is the thing to settle.

## 3. Consecutive merges each trigger a full matrix on `main`

113 runs were started by pushes to `main` in the window, consuming 145 GPU-hours. **71 of them, 63%, were superseded by the next merge before they finished.** Measured the same way as item 1, **77 GPU-hours** were spent on those runs after the next merge had already landed. This is the largest of the three items.

[...]

PR #7174, opened 2026-09-10
Key the doc build concurrency group on the pull request number

albertvillanova:
## What does this PR do?

`build_pr_documentation.yml` keys its concurrency group on `github.head_ref`, which is the head branch name without the fork owner. Two open PRs from different forks that both use a branch called `patch-1` therefore land in the same group, and a push on one cancels the in-flight doc build on the other.

`github.event.pull_request.number` is unique per pull request. This workflow only triggers on `pull_request`, so the `github.run_id` fallback is never reached, but it is kept so the block stays identical to the one in `tests.yml` and `tests-experimental.yml`. Happy to drop it if you would rather the expression say only what this workflow needs.

Bugbot found the same pattern on https://github.com/huggingface/trl/pull/7170#discussion_r3979963943, where it had been copied from here. No two of the 182 currently open PRs share a head branch name, but 130 of them come from forks, so the collision is reachable.

---

> [!NOTE]
> **Low Risk**
> Single CI workflow concurrency expression change with no runtime or application impact.
> 
> **Overview**
> **Fixes cross-PR doc build cancellations** when multiple forks use the same branch name (e.g. `patch-1`).
> 
> The **Build PR Documentation** workflow’s `concurrency.group` now uses `github.event.pull_request.number` instead of `github.head_ref`, so each PR gets its own group and pushes on one PR no longer cancel in-flight doc builds on another. The `github.run_id` fallback is unchanged for parity with other workflows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5adea5af0436da0b54983913bd1145c9c109143c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -7,7 +7,7 @@ env:
   TRL_EXPERIMENTAL_SILENCE: 1
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: true
 
 jobs:


albertvillanova:
`build_pr_documentation.yml` was itself copied from the workflow template in the `doc-builder` README, which still carries the `head_ref` form. The same one-liner is proposed there in huggingface/doc-builder#829, so new copies stop inheriting it.

qgallouedec (approved):
Indeed, I learned something here.

On the fallback, `tests.yml` and `tests-experimental.yml` use `|| github.ref` rather than `|| github.run_id`, so it isn't quite identical either way. They also run on `push` where theirs is reachable, while this one is `pull_request`-only so yours stays dead. Keep it for symmetry if you prefer, no strong view.

Worth pushing the one-liner upstream to the doc-builder template too?, since that's where the pattern came from.

albertvillanova:
Thanks, @qgallouedec.

You are right about `main` as it stands: #7172 changed both test workflows to `|| github.ref`, so the three blocks are not identical today. However, #7201 reverts that to `|| github.run_id` in both (see explanation in https://github.com/huggingface/trl/issues/7169#issuecomment-5659190854), which restores the symmetry this PR was leaning on.

Your reachability point holds either way. In `tests.yml` the fallback is load-bearing, since it is what gives every push its own group and keeps `main` runs from cancelling each other. Here it is dead, because this workflow only triggers on `pull_request`. I would still keep it. The value is that the block can be copied between workflows without anyone having to work out which half applies, and copying it without thinking is exactly how the `head_ref` collision travelled from the doc-builder template into this file, and from here into `tests.yml`.

On the template, as I already commented in https://github.com/huggingface/trl/pull/7174#issuecomment-5631720105, that is what I did on Friday:
> The same one-liner is proposed there in huggingface/doc-builder#829, so new copies stop inheriting it.

PR #7134, opened 2026-09-09
Align tiny Qwen2.5-Coder config with Qwen/Qwen2.5-Coder-0.5B

albertvillanova:
This PR aligns the `tiny-Qwen2ForCausalLM-2.5-Coder` generator config with its reference model, `Qwen/Qwen2.5-Coder-0.5B`.

Part of #7137.

Same treatment as #7098, #7106, #7107, #7108, #7109, #7110, #7112 and #7113, for one of the tiny models that emits no warning in the current job and so was left to a second pass.

### Motivation

The script builds `Qwen2Config` from architecture arguments only, so the special token ids fall back to the class defaults, which are `None`. `Trainer.train()` finds the divergence and rewrites the model config, logging:

```
The tokenizer has new PAD/BOS/EOS tokens that differ from the model config and generation config. The model config and generation config were aligned accordingly, being updated with the tokenizer's values. Updated tokens: {'eos_token_id': 151643, 'bos_token_id': None, 'pad_token_id': 151643}.
```

### Solution

Mirror the reference values:

- `vocab_size=151936` (reference padded embedding size; previously `len(tokenizer.vocab) = 151665`)
- `bos_token_id=151643`, `eos_token_id=151643` (reference values; previously `None`)
- `rope_theta=1000000.0` (reference override; class default is `10000.0`)
- `max_window_layers=24` (reference override; class default is `28`)
- `tie_word_embeddings=True` (reference value; class default is `False`)

`tie_word_embeddings` is included here rather than deferred, on the same reasoning as #7130: it is the class default surviving rather than a deliberate choice, and mirroring it now avoids regenerating this tiny model twice. As there, it changes the artefact and not only the config, since `save_pretrained` omits the tied `lm_head.weight`.

`pad_token_id` is left unset because the reference does not set it either.

Once the tiny model is regenerated the message drops from three keys to one, `{'bos_token_id': None}`, which is the reference disagreeing with its own tokenizer and is not fixable here.

### Before

```
[config_diff] Qwen/Qwen2.5-Coder-0.5B vs tiny (11 differences)
  bos_token_id                                     151643                             → None
  eos_token_id                                     151643                             → None
  hidden_size                                      896                                → 8
  intermediate_size                                4864                               → 32
  layer_types                                      ['full_attention', 'full_attention → ['full_attention', 'full_attention
  max_window_layers                                24                                 → 28
  num_attention_heads                              14                                 → 4
  num_hidden_layers                                24                                 → 2
[...]

PR #7186, opened 2026-09-11
Retire three tiny model generators with no test consumer

albertvillanova:
Three tiny models are generated and published but loaded by no test. This PR drops their generators.

Found by cross-referencing every `push_to_hub` call in the 64 generator scripts against the model ids appearing under `tests/`.

| Tiny model | Status |
|---|---|
| `tiny-Qwen2ForCausalLM-2.5-Coder` | never referenced under `tests/`, in any commit |
| `tiny-Qwen3MoeForSequenceClassification` | was a tokenizer source only; #5469 replaced those uses with `tiny-Qwen3MoeForCausalLM` |
| `tiny-PaliGemmaForConditionalGeneration` | added to the DPO vision list already commented out in #2287; #4728 removed the dead comment |

### Why this is worth doing

An unused tiny model is not free. It is regenerated on every alignment pass, and each regeneration costs a Hub PR to open, review and merge, with no test that would notice if the result were broken.

It also makes the pre-merge check for a tiny model PR silently vacuous. `apply_model_revisions` in `tests/conftest.py` injects a revision only when a test loads that exact id, so a `MODEL_REVISIONS` entry for an unused model never fires: the suite goes green whether or not the regenerated artefact is sound. That is what happened on #7134, where the entry would have had no effect and the run was skipped.

### Questions

CC: @huggingface/trl 

**Delete or revive?** Deleting is the claim that these three models are not worth covering. The opposite reading is defensible, particularly for PaliGemma: it is a prefix-LM VLM, a shape no other tiny model in the suite has, and it was disabled from the moment it was added with no reason recorded, unlike the Idefics2 line above it which cites a transformers issue. If you would rather revive one, say which and I will drop that commit and open the coverage work separately. There is one commit per model, so any of the three can be dropped on its own.

**What about the Hub repos?** This PR only removes the generators. The repositories under `trl-internal-testing` stay as they are. Options are to leave them, archive them, or delete them. Leaving them means an id that still resolves, with no script left to regenerate it, which is the state that produced this cleanup in the first place. I would rather not touch the org without a decision from you.

### Changes

- Drop `for_causal_lm/qwen2_for_causal_lm_2_5_coder.py`
- Drop `for_sequence_classification/qwen3_moe_for_sequence_classification.py`
- Drop `for_conditional_generation/paligemma_for_conditional_generation.py`

---

> [!NOTE]
> **Low Risk**
> Deletes maintenance-only scripts with no test consumers; no runtime or library behavior changes.
> 
> **Overview**
> Removes three **unused** tiny-model generator scripts under `scripts/generate_tiny_models/`: Qwen2.5-Coder causal LM, Qwen3 MoE sequence classification, and PaliGemma conditional generation. Each script built a shrunk config, ran smoke checks, and pushed to Hub via `push_to_hub`, but **no test** in the suite loads those published ids anymore.
> 
> This stops regenerating and opening Hub alignment PRs for models that nothing in CI would exercise. **Hub repos are unchanged**—only the local generators are deleted.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ed090de5b36eb2f92d86320f5bbf1ebc9320443. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (approved):
Thanks!

> Delete or revive? 

Delete and optionally revive in the future, low priority. These are old models, I think we should focus adding support for new models in the time we have.

> What about the Hub repos? 

Delete the repo, IMO, every repo found in trl-internal-testing (this includes datasets as well) should have their generator script in huggingface/trl

albertvillanova:
Thanks, @qgallouedec. Agreed on both counts, and on the principle: a repo under `trl-internal-testing` with no generator script in this repository is a repo nobody can reproduce, which is how these three drifted out of use unnoticed.

On the Hub repos, I would have preferred _archiving_ them rather than _deleting_: it would keep the history and leave the ids resolvable for anything that still points at them. That option does **not** exist on the Hub today. It has been requested since July 2024 and is still open. So, agreed on delete them.

However, I'd rather not fold them into this PR. I checked for consumers before deleting, since a repo deletion is not reversible, and the set (repos without a generation script) is not uniform: some are unused by everyone, some turn out to be load-bearing for other projects. I'll open a dedicated issue to go through every repo that has no generation script, so we can discuss, argue and agree one by one before anything is deleted. This PR then does what it says on the tin, which is removing three generators with no consumer.

PR #7195, opened 2026-09-12
Fix two stream races in OffloadActivations

kashif:
# What does this PR do?

Fixes two stream races in `OffloadActivations` that axolotl found and works around with monkeypatches on their end (`activation_checkpointing.py`), rather than upstream:

1. `self.s0` was cached once at construction as the default/current stream. Under FSDP2, checkpoint recompute, or a custom autograd Function launching kernels on its own stream, the actual current stream at pack/unpack time can differ from the one cached at init, so syncs against `self.s0` target the wrong stream. Made `s0` a property queried live instead.
2. `pack_tensor` cloned offset/non-contiguous saved tensors (e.g. cross-entropy's shifted `shift_labels`) on the transfer stream, after already taking the `wait_stream` snapshot that's supposed to order the transfer stream's copy against the compute stream. The clone was enqueued too late to be covered by that wait, so the transfer stream's copy could race it. Moved the clone before the `wait_stream` call so it's covered.

Both are real, rare, timing-dependent races and hard to pin with a deterministic unit test; `tests/test_activation_offloading.py::test_real_hf_model` caught fix #2 being wrong on a first pass (it produced a gradient mismatch), so I'm relying on the existing suite plus repeated runs rather than adding a new flaky one.

## Before submitting

## AI writing disclosure

## Who can review?

cc @qgallouedec @winglian

---

> [!NOTE]
> **Medium Risk**
> Changes low-level GPU stream ordering in activation offload used during training; incorrect sync could cause rare nondeterministic wrong gradients, though the fix aligns with established stream semantics.
> 
> **Overview**
> Fixes two timing-dependent CUDA/XPU/NPU stream races in **`OffloadActivations`** that could corrupt offloaded activations or break gradient correctness (e.g. under FSDP2, checkpoint recompute, or custom autograd).
> 
> **Compute stream (`s0`)** is no longer pinned at construction (`default_stream` / cached `current_stream`). It is now a **live `current_stream()` property** at pack/unpack time so `wait_stream`, `wait_event`, and `record_stream` target the stream that actually produced the saved tensor.
> 
> In **`pack_tensor`**, contiguous **`clone`/`contiguous`** for non-broadcast views (e.g. shifted labels with non-zero storage offset) runs on the **compute stream before** `s1.wait_stream(s0)`, so the transfer stream’s CPU copy cannot race ahead of that materialization. Stride/shape metadata is captured after that prep, still before the async offload on `s1`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3eb5ffb1bd43a7322773f115bab20f5b8c65ada. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/trl/models/activation_offloading.py
+++ b/trl/models/activation_offloading.py
@@ -151,13 +151,6 @@ class OffloadActivations(saved_tensors_hooks):
         self.accelerator_type = (
             torch.accelerator.current_accelerator().type if hasattr(torch, "accelerator") else "cuda"
         )
-        # NOTE: xpu doesn't have `default_stream` API, use `current_stream` instead
-        if self.accelerator_type == "xpu":  # comp stream
-            self.s0 = torch.xpu.current_stream()
-        elif is_torch_npu_available() and self.accelerator_type == "npu":
-            self.s0 = torch.npu.current_stream()
-        else:
-            self.s0 = torch.cuda.default_stream()
 
         # For streaming
         if self.use_streams:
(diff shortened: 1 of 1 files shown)


qgallouedec (approved):
Thanks! Ran both trees on an H100.

The `s0` bug shows deterministically: read `s0` inside a `torch.cuda.stream(custom)` block and main returns the default stream, this branch returns the custom one.

The clone ordering reads right 👍

The race itself I couldn't show. 15 runs per tree, 15 clean here, one unexplained failure on main that didn't reproduce in 12 more attempts. Agree it's not worth a flaky test.

One ask: fix 1 is deterministic even if the race isn't. Three lines, enter a custom stream, assert `offload.s0` is it. Stops someone re-caching it in `__init__`, which is how this got in.

PR #7197, opened 2026-09-13
[MFU] Use device peak flops instead of hardcoded H100's

dwarez:
# What does this PR do?
The async trainers currently compute MFU using the `compute_mfu` function but without passing it the device's flops, therefore the formula will use the default value for it, which is the peak flops for the H100. 

In this PR we define `_PEAK_FLOPS_BY_DEVICE` which maps the device name to its peak flops (I used torchtitan's mapping as the main source of truth and validated most of nvidia's entries with the official specs). Furthermore, the mapping is dtype/precision aware so it can be eventually expanded with lower precision types.

I preferred to avoid exposing the mfu computation when the peak flops could not be retrieved since using a generic denominator would provide false data to the user, but let me know if you'd like this behavior to be different.

I tested these changes using hf jobs with 2xH200 for both trainers on some smoke runs (so the mfu is quite low) but the retrieved peak flops was indeed correct

## Before submitting

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **Low Risk**
> Changes are limited to observability (MFU metrics) and utility lookups; training loss and optimization paths are untouched, with MFU omitted rather than misreported when capacity is unknown.
> 
> **Overview**
> MFU in **AsyncGRPO** and **AsyncDistillation** no longer assumes a fixed H100 peak FLOPs denominator. At trainer init, peak capacity is resolved from the actual training devices and mixed-precision dtype via new helpers **`get_peak_flops`** and **`get_peak_flops_per_device`** in `trl/trainer/utils.py`, backed by a dtype-aware **`_PEAK_FLOPS_BY_DEVICE`** table (NVIDIA, AMD, Trainium/Inferentia, TPU). **`compute_mfu`** now requires an explicit `peak_flops_per_device` (the old H100 default is removed).
> 
> **`perf/mfu_fwd_bwd`** and **`perf/mfu_wall_clock`** are logged only when every rank has a known peak for that precision; otherwise MFU is skipped and throughput/timing metrics still run. Heterogeneous clusters use the mean peak across ranks when all are known.
> 
> Docs for both async trainers expand the Performance section (throughput vs MFU, `_fwd_bwd` vs `_wall_clock`, training-device-only scope, omission behavior). Tests cover the lookup table, distributed “all ranks must be known” rule, and step-metric MFU gating.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e658593874666cde2184e154055a326447308214. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/trainer/utils.py
```diff
@@ -1703,11 +1705,119 @@ def compute_flops_per_token(config: PretrainedConfig, seq_len: int) -> int:
     return 3 * forward_flops
 
 
+# Theoretical dense accelerator throughput. Values and sources follow TorchTitan's BF16 peak-FLOPs lookup, extended
+# with the additional NVIDIA GPUs offered by Hugging Face Jobs. More specific names must precede their prefixes.
+_PEAK_FLOPS_BY_DEVICE = (
```
AmineDiro, line 1710: Niice, I wanted to have this in utils for a long time 😄 ! 
bf16 is enough for now I think, most people will train in this precision

trl/trainer/utils.py
```diff
@@ -1703,11 +1705,119 @@ def compute_flops_per_token(config: PretrainedConfig, seq_len: int) -> int:
...
+        `float` or `None`: Mean peak FLOPs per device, or `None` if any training device or precision is unsupported.
+    """
+    device = accelerator.device
+    if device.type == "cuda":
+        device_name = torch.cuda.get_device_name(device)
+    elif device.type == "xpu":
+        device_name = torch.xpu.get_device_name(device)
+    else:
+        device_name = device.type
+    dtype = {"bf16": torch.bfloat16, "fp16": torch.float16, "no": model_dtype}.get(accelerator.mixed_precision)
```
AmineDiro: we should probably pass the [dtype](https://github.com/huggingface/trl/blob/a6754332bb88316891fd960957c57ff8fd8d5362/trl/experimental/async_grpo/async_grpo_config.py#L177)

tests/test_utils.py
```diff
@@ -1746,6 +1749,112 @@ def test_config_without_head_dim(self):
...
+    @pytest.mark.parametrize(
+        ("device_name", "dtype"),
+        [
+            ("Tesla T4", torch.bfloat16),
+            ("NVIDIA A10G", torch.float32),
+            ("Unknown accelerator", torch.bfloat16),
+            ("NVIDIA L40", torch.bfloat16),
+            ("NVIDIA A1000", torch.bfloat16),
+            ("NVIDIA XA100", torch.bfloat16),
+            ("Intel Data Center GPU Max 15500", torch.bfloat16),
```
AmineDiro: I am really not sure trl works with this device in the general settings. 
We use NCCL groups and I am pretty just this vllm weightransfer is Nvidia only 🤔

AmineDiro (commented):
Thanks for fixing this! 🤗  Just small nits

trl/trainer/utils.py
dwarez: right that makes sense, I changed it and also changed the mapping (the dict associated to the accelerator name) to reflect the same naming (so without `torch.`) to make matching more immediate 👍

tests/test_utils.py
dwarez: yeah I wasn't really sure about that. I saw that TRL has explicit support for XPUs but I guess that not everything works out of the box. I've removed those entries for the time being, we can also add them back eventually

trl/trainer/utils.py
```diff
@@ -1703,11 +1705,105 @@ def compute_flops_per_token(config: PretrainedConfig, seq_len: int) -> int:
...
+
+    Returns:
+        `float` or `None`: Mean peak FLOPs per device, or `None` if any training device or precision is unsupported.
+    """
+    device = accelerator.device
+    if device.type == "cuda":
+        device_name = torch.cuda.get_device_name(device)
+    else:
+        device_name = device.type
+    peak_flops = get_peak_flops(device_name, dtype)
```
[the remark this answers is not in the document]
dwarez: @AmineDiro I guess that's right, I still need to grab the accel `mixed_precision` field to account for mixed precision training (which is the default)

trl/trainer/utils.py
```diff
@@ -1703,11 +1705,105 @@ def compute_flops_per_token(config: PretrainedConfig, seq_len: int) -> int:
...
+    peaks = gather_object([peak_flops])
+    if any(peak is None for peak in peaks):
+        logger.info(
+            "MFU metrics are disabled because the peak FLOPs are unknown for at least one training device or "
+            "precision. Throughput and timing metrics are still reported."
+        )
+        return None
+    return sum(peaks) / len(peaks)
```
qgallouedec, line 1799: I think this is too defensive: mixed GPU models in a single job basically never happens, and even when it does nothing breaks, log builds the metrics dict locally and only the main process reports, so ranks disagreeing about whether to emit `perf/mfu_*` doesn't hang anything.
I'd drop the gather and just use the local device's peak. Same for the `device.type != "cuda"` branch, xla devices report `device.type == "xla"` so it never matches the TPU or trn entries in the table.

dwarez: can I use ` torch.cuda.get_device_name(accelerator.device)` or do we need a check also for TPUs? In that case I can drop tranium entries from the mapping. In case we cannot map we just warn and then omit the metric

Issue #5361, opened 2026-03-24
Multimodal generation hangs in vllm server mode with large images (works in colocate), manual resizing required


SmartFive:
### Reproduction

Description:
When running a multimodal RL training script (e.g., GRPO with Qwen-VL) using TRL and vLLM, I encountered a severe hanging issue specific to the vllm-mode="server".

If I pass original high-resolution images (e.g., 1080x720, 1555x709) to the environment, the generation process completely hangs/freezes in server mode. However, the exact same images process perfectly fine without any issues when using vllm-mode="colocate".

The Workaround:
The only way to prevent the server mode from hanging is to manually resize the images before passing them to the messages:

```
Python
from PIL import Image

MAX_DIM = 1024
# ...
if max(img.size) > MAX_DIM:
    img.thumbnail((MAX_DIM, MAX_DIM)) 
```
Logs showing the problematic image sizes (before resizing):

```
Plaintext
Sample 'WIT_184' Step 0 | Multiple Prompt Length: 4036 tokens | Text Prompt Length: 2212 | Images: 2
  - Image size: (1080, 720), mode: RGB
  - Image size: (1555, 709), mode: RGB
 ```
Expected Behavior:
I suspect the hang in server mode is caused by the massive payload size of Base64 encoded images blocking the JSON serialization or causing an HTTP timeout, which colocate mode avoids by sharing memory/pointers.

However, as a user, I do not want to manually downsample my images, as this loss of resolution might severely degrade the model's visual performance. The framework should ideally handle standard high-res images natively in server mode without freezing (perhaps by optimizing the image transmission protocol, supporting local file path references, or handling the preprocessing dynamically).

### System Info

- Platform: Linux-5.4.0-204-generic-x86_64-with-glibc2.31
- Python version: 3.12.12
- TRL version: 1.0.0.dev0
- PyTorch version: 2.9.1
- accelerator(s): NVIDIA A100-SXM4-80GB, NVIDIA A100-SXM4-80GB, NVIDIA A100-SXM4-80GB, NVIDIA A100-SXM4-80GB, NVIDIA A100-SXM4-80GB, NVIDIA A100-SXM4-80GB, NVIDIA A100-SXM4-80GB, NVIDIA A100-SXM4-80GB
- Transformers version: 4.57.6
- Accelerate version: 1.13.0
- Accelerate config: 
  - compute_environment: LOCAL_MACHINE
  - distributed_type: DEEPSPEED
  - mixed_precision: bf16
[...]

PR #7199, opened 2026-09-13
Reduce connection-pool warnings in the vLLM client

vladbataev:
# What does this PR do?

`VLLMClient` dispatches up to 64 concurrent requests for chat, image preprocessing, and generation from image features, but its HTTP adapter retains only 10 connections per host by default. When connections return to a full pool, urllib3 closes them and logs `Connection pool is full, discarding connection`.

Size the pool to match the existing concurrency defaults using a shared constant, and add regression tests that check the actual HTTP and HTTPS pool capacities.

The intended benefit is fewer pool-full warnings and less unnecessary connection churn. No training speedup is claimed; training throughput and step time have not been measured.

Related to #5361, where this pool-size mismatch was investigated. This PR does not claim to resolve the reported training hang.

## Validation

- Controlled local HTTP smoke test with two batches of 64 concurrent chat requests: **108 pool-full warnings before, 0 after**. New TCP connections in the second batch fell from **54 to 0**. This was a transport check, not a training benchmark.
- `pytest tests/test_vllm_client_server.py -q`: **7 passed, 35 skipped**; vLLM is not installed locally.
- Both new protocol cases fail when the pool-size argument is removed in memory.
- Ruff lint, formatting, and `git diff --check` passed.

## Before submitting

- [x] Read the contributor guidelines.
- [x] Linked the related investigation: #5361.
- [x] Reviewed documentation needs; no public API or default-concurrency changes require documentation updates.
- [x] Added regression tests.

## AI writing disclosure

---

> [!NOTE]
> **Low Risk**
> Transport-layer pooling change only; default concurrency behavior is unchanged and covered by new unit tests.
> 
> **Overview**
> Aligns the vLLM HTTP client’s urllib3 connection pool with its default **64** concurrent request dispatch, which previously used the adapter’s default pool size (~10) and triggered “Connection pool is full, discarding connection” warnings under load.
> 
> Introduces `_DEFAULT_GENERATION_CONCURRENCY` and sets `HTTPAdapter(pool_maxsize=...)` to that value for both `http://` and `https://`. The same constant replaces literal `64` defaults on `image_features`, `_generate_from_features`, and `chat` so pool sizing stays tied to concurrency.
> 
> Adds `TestConnectionPoolSize` (mocked client init) asserting each scheme’s pool `maxsize` is at least the default concurrency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52a8f58a6460cbcac427df71daea847a89f6253e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vladbataev:
Hi @qgallouedec, could you review the PR when you have a chance? It’s a small vLLM client pool-sizing fix to reduce pool-full warnings, with regression tests included.

albertvillanova:

vladbataev:
@albertvillanova I marked this AI-generated to disclose its origin accurately. I’ve read and reviewed the implementation: the client already dispatches up to 64 concurrent requests, and the change aligns the connection pool’s retained capacity with that default.

I understand the concern about contributors being unable to explain their changes. I can discuss this implementation and address review feedback. Would you consider reopening it for review?

albertvillanova:
Reopening. Thanks for engaging, and for disclosing the origin accurately rather than under-reporting it.

The concern behind that policy is the case where a contributor cannot discuss the change or vouch for it. Your description matches the code (the client methods default to 64 concurrent requests while the mounted adapter keeps the default pool of 10), so that concern does not apply here.

I will review it shortly. The expectation from here is the usual one: you address the review comments yourself.

albertvillanova (commented):
Thanks, this holds up. I checked the premise rather than the prose, and it is accurate: `requests` mounts `HTTPAdapter` with `pool_maxsize=10` and `pool_block=False`, so when `image_features`, `_generate_from_features` and `chat` each fan out to 64 threads against a single host, urllib3 opens 64 connections, returns 10 to the pool and closes the other 54 with the warning from `HTTPConnectionPool._put_conn`. Nothing in the repo passes `max_concurrent_requests`, so 64 is what actually runs. Your 108 warnings over two batches and 54 fresh connections in the second batch are exactly what that arithmetic predicts. The server-mode client is built on the main process only, so this trades 10 idle sockets for 64 on one process rather than per rank. Thanks also for keeping #5361 as context rather than a claimed fix; that one is about base64 image payloads and is a separate problem.

Three things before this can go in.

**1. `get_sequence_logprobs` keeps a literal 4.** The constant is named `DEFAULT_MAX_CONCURRENT_REQUESTS`, but `get_sequence_logprobs` has the same parameter and stays at `4`. Was leaving it out deliberate, and what is the reasoning? I have a view, but I would rather hear yours first, because it decides the shape of the change: a constant with that name covering three of the four is the one outcome to avoid. Either it is genuinely the default for every `max_concurrent_requests` and that method adopts it too, or the two numbers mean different things and the constant should be named for what it actually sizes.

**2. The test asserts against itself.** It reads the expected value out of `inspect.signature(...).parameters["max_concurrent_requests"].default`, so the pass condition is derived from the same source lines it is guarding. Lower those defaults to 5 and the test still passes against a pool of 64; rename or drop the parameter and it fails with a `KeyError` rather than a useful message. Assert the number instead: `assert pool.pool.maxsize >= 64`. It is shorter, it is readable without knowing the introspection API, it says what the test is actually for, and it removes the need to reach into the private `_generate_from_features` from a test. This repo prefers the blunt version of a check over the clever one.

**3. Constant naming.** The two existing module-level constants in this file are `_HAS_STATEFUL_TRAINER_ENGINE` and `_HAS_WEIGHT_UPDATE_LIFECYCLE`. If the constant survives point 1, prefix it to match.

Minor, take or leave: one `HTTPAdapter` instance is mounted under both `http://` and `https://`, so both parametrized cases read `pool_maxsize` off the same object and the second adds little.

vladbataev:
Thanks for the review.

I’m keeping `get_sequence_logprobs` at `4`: it dispatches chunks containing multiple sequences, whereas the three generation paths dispatch one request per prompt. Raising that limit would change existing dispatch behavior beyond this pool-sizing fix. I’ve renamed the shared constant to `_DEFAULT_GENERATION_CONCURRENCY` to clarify its scope and make it private.

I agree that removing the signature introspection simplifies the test. I’ve changed the assertion to:

```python
assert pool.pool.maxsize >= _DEFAULT_GENERATION_CONCURRENCY
```

I’d prefer using the constant here because the requirement is that the pool accommodates the configured default generation concurrency. The test still catches the original regression: omitting `pool_maxsize` leaves the actual pool at `10`, below the current default of `64`. I verified that both protocol cases fail when that argument is omitted.

I’ve also retained both protocol cases because they check the two adapter registrations. Omitting only the HTTPS mount leaves HTTPS using Requests’ default pool of `10`, while HTTP remains correctly configured, so the second case catches a separate regression.

qgallouedec:

vladbataev:
@qgallouedec could you clarify why the PR was closed again after the earlier decision to reopen it https://github.com/huggingface/trl/pull/7199#issuecomment-5696155591
I proceeded based on the explicit invitation to continue the review, so I’m confused about the expectation now.

qgallouedec:
My mistake, sorry for the noise. We get a large volume of automated sloppy PRs and triage them in batches, so we do get false positives. Reopening.

albertvillanova (approved):
All three addressed, and you were right to push back on the fourth.

I checked the HTTPS point rather than taking it on trust, and it holds: `requests.Session.__init__` mounts its own default `HTTPAdapter()` under both `https://` and `http://`, so dropping only the `https://` line leaves HTTPS on a pool of 10 while HTTP stays correct. The two cases do guard different registrations. My nit was wrong, keep them both.

The `get_sequence_logprobs` answer is the right one. It dispatches chunks of sequences while the other three dispatch one request per item, so the two numbers genuinely size different things, and `_DEFAULT_GENERATION_CONCURRENCY` names what it sizes without over-claiming. Leaving that method's `4` alone also keeps the PR to the pool-sizing change.

On the assertion, agreed: the constant is the better right-hand side, since a legitimate lowering of the default should not fail the test. My objection was to deriving the expected value through `inspect.signature`, and that is gone. Thanks for checking that it fails without `pool_maxsize` rather than resting on a green run.

PR #7201, opened 2026-09-14
Stop cancelling superseded runs on main

albertvillanova:
## What does this PR do?

#7172 grouped push runs by `github.ref` so that a newer merge cancels the run started by the previous one. It saves the runner time it claimed, but it makes `main` look broken: I thought it was going to be grey instead of red.

GitHub rolls a commit's checks up into the single icon shown in the commit list, and it buckets `cancelled` with failure. The three most recent merges before this PR:

Image

```
FAILURE  cancelled=5 failure=0   Move the latest release tests from T4 to L40S (#7184)
FAILURE  cancelled=5 failure=0   Use the canonical id for the tiny GPTNeoX model (#7185)
FAILURE  cancelled=0 failure=1   Train the tiny DeepSeek-R1-Distill model (#7187)
```

The first two are red for nothing. The third is a real failure, and it is now indistinguishable from the noise. Since #7172 merged, 3 of the 8 commits on `main` carry a red rollup and 2 of them only because their runs were cancelled. A history where most commits show a failing icon teaches everyone to stop reading the icon, which costs more than the runner time the change saved.

There is no way to recolour it. Only `success`, `skipped` and `neutral` avoid red; a workflow job cannot report `neutral` for itself, and a job-level `if:` is evaluated when the job is scheduled, which is before we can know that a later commit is coming.

This restores `github.run_id` as the fallback, so every push gets its own group again. Cancellation of superseded pull request runs, the larger half of #7170 and the part that was not in question, is unchanged. The `!cancelled()` guard added to the Slack steps in #7172 stays and simply becomes inert.

Applied to `tests-experimental.yml` as well so the two blocks stay identical.

Removing redundant `main` runs without attaching cancelled checks to commits needs a mechanism that tests batches before they land rather than after, which is what the GitHub merge queue does. That is a bigger change and a separate conversation.

Part of #7169.

---

> [!NOTE]
> **Low Risk**
> CI workflow-only change; restores per-push runs on main without altering test logic or PR cancel-in-progress behavior.
> 
> **Overview**
> Reverts the **non-PR** GitHub Actions concurrency key from `github.ref` back to **`github.run_id`** in `tests.yml` and `tests-experimental.yml`, with the same one-line change in both workflows.
> 
> On **`main` pushes**, each run now gets its own concurrency group again, so a newer merge no longer **cancels** the previous push’s jobs. That avoids cancelled workflows rolling up as red/failed on the commit list when nothing actually broke. **Pull request** behaviour is unchanged: the group still uses `github.event.pull_request.number`, so superseded PR runs can still be cancelled via `cancel-in-progress: true`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc639918f7699b550b4edbe00e435868c81fcf60. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:
Merging this PR as a CI hotfix.

PR #7212, opened 2026-09-15
Add TailSFT example

kashif:
# What does this PR do?

Adds TailSFT to `SFTTrainer`, following Algorithm 1 from the paper. It includes the initial-loss filter, a small example, docs, and tests.

## Before submitting

- [ ] Was this discussed/approved via a GitHub issue? Please add a link to it if that is the case.

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **Low Risk**
> Changes are limited to a new standalone example script and documentation; core TRL training APIs are not modified in this diff.
> 
> **Overview**
> **Adds a TailSFT GSM8K example** that subclasses [`SFTTrainer`] to implement filtered fine-tuning from the TailSFT paper: it records each sequence’s length-normalized initial-policy loss, then on each training step drops the fraction of the distributed selection batch whose current loss improved most relative to that baseline (`filter_fraction`, with optional linear `ramp` schedule). Training loss is token-averaged only over kept sequences; packing is rejected, and MoE router aux loss is recomputed on survivors only.
> 
> Documentation lists the new `tail_sft_gsm8k` example in the overview table and adds a **Supervised Fine-Tuning** paper-index entry with a short description and `python examples/tail_sft_gsm8k/tail_sft.py` invocation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a666cdfa1645a5add245071d9ea731972f087bf8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/source/paper_index.md
```diff
@@ -1230,6 +1230,30 @@ trainer.train()
 
 Papers relating to the [`SFTTrainer`]
 
+### TailSFT: Filtered Fine-Tuning Improves Post-Training Performance
+
+**📜 Paper**: https://huggingface.co/papers/2608.25756
+
+TailSFT preserves response coverage for subsequent reinforcement learning by filtering the sequences whose
+length-normalized loss has improved the most relative to the initial policy. TRL implements the paper's distributed
+selection-batch filtering with `loss_type="tail_sft"`; the training dataset must contain an `initial_loss` column with
```
lewtun: I wonder if we could automatically compute the `initial_loss` if it's not present in the training dataset or provide a `.predict()` method to enable it? Just an idea and not necessary if it adds too much complexity

examples/sft_tail_sft/sft_tail_sft.py
```diff
@@ -0,0 +1,93 @@
...
+)
+preparation_trainer = SFTTrainer(
+    model=model,
+    args=preparation_args,
+    train_dataset=dataset,
+    processing_class=tokenizer,
+)
+dataset = preparation_trainer.train_dataset
+
+# Algorithm 1 records the initial policy's mean target-token cross-entropy for every sequence.
```
lewtun: Would it make sense to wrap this in a `.predict()` method so that we can simplify the logic a bit and not require two separate `SFTTrainer` instances?

trl/trainer/sft_trainer.py
```diff
@@ -1811,9 +1851,49 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+                num_filtered = min(int(gathered_margins.numel() * filter_fraction + 0.5), gathered_margins.numel() - 1)
+                keep = torch.ones_like(gathered_margins, dtype=torch.bool)
+                keep[gathered_margins.argsort()[:num_filtered]] = False
+                batch_size = margins.size(0)
+                start = self.accelerator.process_index * batch_size
+                local_keep = keep[start : start + batch_size]
+
+                selected_tokens = loss_mask * local_keep.unsqueeze(1)
+                global_num_tokens = self.accelerator.gather(selected_tokens.sum()).sum()
+                loss = (per_token_loss * selected_tokens).sum() / (global_num_tokens / self.accelerator.num_processes)
```
AmineDiro: Pretty sure it works with grad acc. but just checking if it's working

trl/trainer/sft_trainer.py
```diff
@@ -1811,9 +1851,49 @@ def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=N
...
+                if self.aux_loss_enabled:
+                    loss = loss + self.args.router_aux_loss_coef * outputs.aux_loss
```
AmineDiro: `outputs.aux_loss` is computed over all the tokens but loss is _filtered_ here, right ? So the router covers tokens that will get thrown away. 

I wonder if this is really important or not 🤔  ?? ( wdyt @qgallouedec  @kashif  ) 

if we keep it, we are effectively kind of "leaking" the gradient from masked sequences to the model

AmineDiro (commented):
LGTM. Super @kashif 🐐 !! 

Just question about grad acc and MoE models here that can be an edge case

trl/trainer/sft_trainer.py
AmineDiro: Actually thinking about it this `global_num_tokens` is only computed per  within the micro-batch ? If I've understand the paper correctly filtering is gathered accross ranks of 1 microbatch size OR across a whole batch size spanning different grad acc steps ?

qgallouedec:
Thks for the PR. i'm not keen on landing this in `SFTTrainer` though.

The cost isn't the 90 lines, it's that we support them forever. It adds a branch in `compute_loss`, two new config fields, a new reserved dataset column name, and a new set of cross-incompatibilities. The PR already has to explicitly reject packing, padding-free, vision, liger, custom `compute_loss_func`, CP, SP and TP, and that's before the one we raised at journal club (around GAS and DDP) and some open ones (https://github.com/huggingface/trl/pull/7212#discussion_r4016448311 from @AmineDiro). From then on every change to the SFT loss path has to be checked against `tail_sft` too.

And we don't know yet whether this reproduces or is useful outside the paper's setup. (The paper is a few weeks old)

So instead I think the right move is an example

```python

import torch
from torch.utils.data import DataLoader

from trl import ModelConfig, ScriptArguments, SFTConfig, SFTTrainer, TrlParser
from trl.trainer.sft_trainer import DataCollatorForLanguageModeling


@dataclass
class TailSFTScriptArguments(ScriptArguments):
    r"""
    [`ScriptArguments`] with GSM8K as the default dataset.

    Parameters whose default values are overridden:

    > - `dataset_name`: Defaults to `"openai/gsm8k"`.
    > - `dataset_config`: Defaults to `"main"`.
    """

    dataset_name: str = "openai/gsm8k"
    dataset_config: str | None = "main"


@dataclass
class TailSFTConfig(SFTConfig):
    r"""
    [`SFTConfig`] with the filtering schedule of TailSFT.

    Parameters whose default values are overridden:

    > - `per_device_train_batch_size`: Defaults to `2` (a 16-example selection batch on 8 GPUs, as in the paper's math
    >   runs).

    Additional parameters:

    > - `filter_fraction`: Defaults to `0.5`.
    > - `filter_schedule`: Defaults to `"ramp"`.
    """

    per_device_train_batch_size: int = 2
    filter_fraction: float = field(
        default=0.5,
        metadata={"help": "Fraction of each selection batch to drop. `0` recovers standard SFT."},
    )
    filter_schedule: str = field(
        default="ramp",
        metadata={
            "help": "How the filter fraction moves over training. `static` holds it at `filter_fraction`; `ramp` "
            "raises it linearly from 0 at the first step to `filter_fraction` at the last.",
            "choices": ["static", "ramp"],
        },
    )


@dataclass
class TailSFTModelConfig(ModelConfig):
    r"""
    [`ModelConfig`] with the paper's model as default.

    Parameters whose default values are overridden:

    > - `model_name_or_path`: Defaults to `"allenai/Olmo-3-1025-7B"`.
    > - `dtype`: Defaults to `"bfloat16"`.
    """

    model_name_or_path: str = "allenai/Olmo-3-1025-7B"
    dtype: str = "bfloat16"


@dataclass
class DataCollatorForTailSFT(DataCollatorForLanguageModeling):
    """SFT collator that also batches the recorded initial-policy loss of each example."""

    def torch_call(self, examples: list[dict]) -> dict[str, torch.Tensor]:
        output = super().torch_call(examples)
        output["initial_loss"] = torch.tensor([example["initial_loss"] for example in examples])
        return output


class TailSFTTrainer(SFTTrainer):
    """
    [`SFTTrainer`] with the sequence filtering of TailSFT.

    Reads the `initial_loss` column of the training dataset, and records it with the initial policy when the column is
    absent. An evaluation dataset, if any, needs the column too.
    """

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        if self.args.packing:
            raise ValueError("Packing is not supported: filtering is per sequence, and packing concatenates them.")

        # Gradient accumulation requires scaled loss. Normally, loss scaling in the parent class depends on whether the
        # model accepts loss-related kwargs. Since we compute our own loss, this check is irrelevant. We set
        # self.model_accepts_loss_kwargs to False to enable scaling.
        self.model_accepts_loss_kwargs = False

        # Algorithm 1, line 2. Recorded here, once the parent has tokenized the dataset, so the initial loss is
        # measured over exactly the target tokens the training loss will use.
        if "initial_loss" not in self.train_dataset.column_names:
            self.train_dataset = self.train_dataset.add_column("initial_loss", self.record_initial_losses())

        self.data_collator = DataCollatorForTailSFT(
            pad_token_id=self.data_collator.pad_token_id, pad_to_multiple_of=self.args.pad_to_multiple_of
        )

    def _set_signature_columns_if_needed(self):
        # Without this, `initial_loss` is dropped from the dataset before it reaches the collator.
        super()._set_signature_columns_if_needed()
        self._signature_columns = [*self._signature_columns, "initial_loss"]

    @torch.no_grad()
    def record_initial_losses(self) -> list[float]:
        """Length-normalized loss of the initial policy over every training example."""
        # Every rank scores the whole dataset. The pass is redundant across ranks, but it is forward-only and it keeps
        # the recording free of any cross-rank bookkeeping.
        dataloader = DataLoader(
            self.train_dataset,
            batch_size=self.args.per_device_train_batch_size,
            collate_fn=self.data_collator,
        )
        self.model.eval()
        initial_losses = []
        for inputs in dataloader:
            inputs = {key: value.to(self.model.device) for key, value in inputs.items()}
            logits = self.model(input_ids=inputs["input_ids"], attention_mask=inputs["attention_mask"]).logits
            per_token_loss, loss_mask = self.per_token_loss(logits, inputs["labels"])
            initial_losses += (per_token_loss.sum(-1) / loss_mask.sum(-1)).tolist()
        self.model.train()
        return initial_losses

    def per_token_loss(self, logits: torch.Tensor, labels: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
        """Cross-entropy of every next-token prediction, zero outside the target tokens, and the mask of those."""
        shift_logits, shift_labels = logits[..., :-1, :], labels[..., 1:]
        loss = F.cross_entropy(shift_logits.transpose(1, 2), shift_labels, ignore_index=-100, reduction="none")
        return loss, shift_labels != -100

    def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=None):
        initial_loss = inputs.pop("initial_loss")
        if not self.model.training:
            return super().compute_loss(model, inputs, return_outputs, num_items_in_batch)

        labels = inputs.pop("labels")
        outputs = model(**inputs, use_cache=False)
        per_token_loss, loss_mask = self.per_token_loss(outputs.logits, labels)

        # Algorithm 1, line 6: over the selection batch, the examples of a single forward pass across all ranks, drop
        # the fraction whose loss has dropped the furthest below where the initial policy left it. Both losses are
        # length-normalized, so the decision doesn't favor longer or shorter responses.
        fraction = self.args.filter_fraction
        if self.args.filter_schedule == "ramp":
            fraction *= self.state.global_step / max(self.state.max_steps - 1, 1)
        margins = (per_token_loss.sum(-1) / loss_mask.sum(-1)).detach() - initial_loss
        margins = self.accelerator.gather(margins)
        # The paper drops round(world_size * batch_size * f) examples; keep one so the token average stays defined.
        keep = torch.ones_like(margins, dtype=torch.bool)
        keep[margins.argsort()[: min(round(margins.numel() * fraction), margins.numel() - 1)]] = False
        keep = keep.view(self.accelerator.num_processes, -1)[self.accelerator.process_index]

        # Algorithm 1, line 8: token-averaged cross-entropy over the survivors. The average runs over the whole
        # selection batch, so the per-rank losses are scaled to sum to it once the backward pass averages them.
        loss_mask = loss_mask & keep.unsqueeze(-1)
        num_tokens = self.accelerator.gather(loss_mask.sum()).sum()
        loss = (per_token_loss * loss_mask).sum() * self.accelerator.num_processes / num_tokens

        self._metrics["train"]["filtered_fraction"].append(1 - keep.float().mean().item())
        return (loss, outputs) if return_outputs else loss


def main(script_args, training_args, model_args):
    training_args.model_init_kwargs = dict(
        revision=model_args.model_revision,
        attn_implementation=model_args.attn_implementation,
        dtype=model_args.dtype,
    )

    dataset = load_dataset(script_args.dataset_name, script_args.dataset_config, split=script_args.dataset_train_split)
    dataset = dataset.rename_columns({"question": "prompt", "answer": "completion"})

    trainer = TailSFTTrainer(
        model=model_args.model_name_or_path,
        args=training_args,
        train_dataset=dataset,
    )
    trainer.train()

    if training_args.push_to_hub:
        trainer.push_to_hub(dataset_name=script_args.dataset_name)


def make_parser():
    dataclass_types = (TailSFTScriptArguments, TailSFTConfig, TailSFTModelConfig)
    return TrlParser(dataclass_types)


if __name__ == "__main__":
    parser = make_parser()
    script_args, training_args, model_args = parser.parse_args_and_config()
    main(script_args, training_args, model_args)
```

so it kills two birds with one stone:
- we get the paper reproduced, and
- we get an example showing that it is easy to customize TRL, and that the way to do it is to subclass the trainer, not to patch the source or wait for us to add a `loss_type`. `SFTTrainer` stays as it is, no new config, no new reserved column, no new incompatibility matrix.

Would you be up for moving the PR to `examples/tail_sft_gsm8k/` along those lines?

trl/trainer/sft_trainer.py
qgallouedec: ideally it should be across the whole batch, but seems very hard to implement cleanly. The current state is a nice sweet spot

docs/source/paper_index.md
```diff
@@ -1230,6 +1230,39 @@ trainer.train()
...
 
+### TailSFT: Filtered Fine-Tuning Improves Post-Training Performance
+
+**📜 Paper**: https://huggingface.co/papers/2608.25756
+
+TailSFT preserves response coverage for subsequent reinforcement learning by filtering the sequences whose
+length-normalized loss has improved the most relative to the initial policy. The
+[`tail_sft_gsm8k`](https://github.com/huggingface/trl/tree/main/examples/tail_sft_gsm8k) example implements the paper's
+distributed selection-batch filtering by subclassing [`SFTTrainer`]. It records each sequence's initial-policy loss
+before training and filters examples independently in every selection batch.
```
```suggestion
TailSFT preserves response coverage for subsequent reinforcement learning by filtering the sequences whose length-normalized loss has improved the most relative to the initial policy. The [`tail_sft_gsm8k`](https://github.com/huggingface/trl/tree/main/examples/tail_sft_gsm8k) example implements the paper's distributed selection-batch filtering by subclassing [`SFTTrainer`]. It records each sequence's initial-policy loss before training and filters examples independently in every selection batch.
```

docs/source/paper_index.md
```diff
@@ -1230,6 +1230,39 @@ trainer.train()
...
+    lr_scheduler_type="constant_with_warmup",
+    filter_fraction=0.5,  # The selected math and OCI runs ramp to a 0.5 filtering fraction.
+    filter_schedule="ramp",  # "...raises the fraction linearly from 0 at the first step to f at the last"
+)
+trainer = TailSFTTrainer(
+    ...,
+    args=training_args,
+    train_dataset=dataset,
+)
+```
```
````suggestion
```sh
python examples/tail_sft_gsm8k/tail_sft.py
```
````

**Review by qgallouedec — APPROVED** [review]
nice ! 🐐

**Comment by qgallouedec** [comment]
just make sure to update the PR title before merging

File: examples/tail_sft_gsm8k/tail_sft.py [hunk 7]
```diff
@@ -0,0 +1,248 @@
...
+class DataCollatorForTailSFT(DataCollatorForLanguageModeling):
+    """SFT collator that also batches the recorded initial-policy loss of each example."""
+
+    def torch_call(self, examples: list[dict]) -> dict[str, torch.Tensor]:
+        output = super().torch_call(examples)
+        output["initial_loss"] = torch.tensor([example["initial_loss"] for example in examples])
+        return output
+
+
+class TailSFTTrainer(SFTTrainer):
```
AmineDiro, line 121: that's cool !

trl/trainer/sft_trainer.py
AmineDiro: @kashif  this was resolved, any details on this, just curious

kashif: Resolved because the implementation moved to a paper-specific example targeting dense OLMo-3, so the MoE aux-loss case is no longer part of the generic SFTTrainer path.

kashif: Follow-up: I added explicit MoE support in a666cdf. TailSFT now recomputes the native load-balancing loss with filtered sequences masked out, so discarded sequences do not update the router.

AmineDiro: Oh, okay. I wasn't really sure if we should keep the discarded sequence aux_loss because I don't have a strong view on whether the routing ability over tokens itself should be subject to filtering 👍🏼

PR #7213, opened 2026-09-15
Await async reward functions written as callable classes

albertvillanova:
This PR makes an asynchronous reward function written as a callable class run asynchronously, instead of being dispatched as a synchronous one.

### Motivation

A reward function can take three shapes in TRL: a module-level function, a `functools.partial`, or a callable class instance. `get_callable_name` names all three, and `AsyncGRPOTrainer` documents the callable class as one of the picklable forms to use, since everything handed to the rollout worker is pickled.

Asynchronous reward functions are recognised with `inspect.iscoroutinefunction`, which inspects the object it is given and not its `__call__`. For an instance of a class whose `__call__` is `async def` it returns `False`, so the reward function is dispatched as a synchronous one:

- `GRPOTrainer` and `RLOOTrainer` call it inline and then iterate what came back, raising `TypeError: 'coroutine' object is not iterable`.
- The AsyncGRPO rollout worker hands it to a thread, where the call only creates the coroutine, and the un-awaited coroutine object is what lands in that reward function's column.

The one asynchronous form that survives pickling is therefore the one that does not run.

### Solution

`is_async_callable` answers whether calling an object returns a coroutine, over the same three shapes `get_callable_name` covers: it unwraps `functools.partial` the same way, then falls back to the object's `__call__`. It replaces `inspect.iscoroutinefunction` wherever a reward function is dispatched.

Tool dispatch keeps `inspect.iscoroutinefunction`. Tools are keyed by `tool.__name__`, so they are always plain functions or bound methods, which it reads correctly. Environment `get_reward` methods are bound methods as well, and are left alone for the same reason.

### Changes

- Add `is_async_callable` to `trl/trainer/utils.py`, beside `get_callable_name`
- Use it for reward function dispatch in `GRPOTrainer`, `RLOOTrainer` and the AsyncGRPO rollout worker
- Cover the detection of each callable shape, and add an async callable class reward function to the async reward tests of all three

---

> [!NOTE]
> **Low Risk**
> Targeted dispatch fix for reward functions with expanded test coverage; behavior unchanged for plain sync/async functions and models.
> 
> **Overview**
> Adds **`is_async_callable`** next to **`get_callable_name`**, unwrapping **`functools.partial`** and treating a class with **`async def __call__`** as async—something **`inspect.iscoroutinefunction`** misses on instances.
> 
> **GRPOTrainer**, **RLOOTrainer**, and the AsyncGRPO rollout worker’s **`_score_group`** now use it to decide whether reward functions are awaited on the event loop vs run in a thread, fixing broken training when rewards are picklable **async callable classes** (coroutines were returned un-awaited or iterated as sync results).
> 
> Tests cover each callable shape, an async-class scoring path in async GRPO, and an **`AsyncCallableReward`** in the existing GRPO/RLOO mixed sync/async training tests; async rollout tests also use **`get_callable_name`** for reward metric keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e26732284aba3c907ff0d48e21c11f3a1b6570e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (approved):
clean, thanks!

PR #7214, opened 2026-09-15
Link the upstream issue behind the trufflehog digest pin

albertvillanova:
This PR records why the TruffleHog scanner image is pinned by digest alongside the action SHA, and when that pin can go away.

### Motivation

The action is a wrapper: it runs the scanner as a container chosen by the `version` input, so the action SHA pins the wrapper while the `version` input pins the binary that actually scans. The comment already states that the two must move together, but not why the coupling exists nor when it ends, so the extra pin reads as unexplained maintenance.

It also costs something on every bump: Dependabot updates the action SHA and leaves the `version` input untouched, and since the wrapper stays compatible with the older image the scan keeps passing while silently running the previous scanner.

### Changes

- Link the upstream issue about the `version` default in the secret scanning workflow
- State the condition under which the digest pin becomes redundant

---

> [!NOTE]
> **Low Risk**
> Comment-only workflow documentation with no runtime or security behavior changes.
> 
> **Overview**
> Updates **comments only** in the TruffleHog secret-scanning workflow—no behavior change to checkout, `version`, or `extra_args`.
> 
> The comment block now explains that the action SHA pins the wrapper while the `version` digest pins the scanner image (because `:latest` is mutable), that **Dependabot** only bumps the action SHA so both pins must be synced manually, and links **upstream issue #4857**. It also records when to drop the digest pin: once the action defaults `version` to its own release.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a8f86d32c48fa4a8f642816b59b6b50cdc743e48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:
Merging this PR as a simple change in a CI comment.

Issue #6151, opened 2026-06-23
Support router-based multi-teacher (MOPD) distillation in DistillationTrainer


mohanpb:
### Feature request

`DistillationTrainer` currently supports teacher-student distillation with a single teacher. A useful extension would be router-based multi-teacher distillation, where each example is routed to one teacher based on a dataset column such as `domain`, `task`, or `data_source`.

This would support workflows where a single student is trained from multiple specialist teachers, for example:

- math examples -> math teacher
- code examples -> code teacher
- general instruction examples -> general teacher

A possible API could be:

```python
DistillationConfig(
    lmbda=1.0,
    beta=1.0,
    teacher_model_names_or_paths={
        "math": "path-or-hub-id-to-math-teacher",
        "code": "path-or-hub-id-to-code-teacher",
        "general": "path-or-hub-id-to-general-teacher",
    },
    teacher_routing_column="domain",
)
```

Dataset example:

```python
{
    "messages": [...],
    "domain": "math",
}
```

Initial scope could be limited to:

> one example -> one route key -> one teacher

This would not require teacher aggregation, dynamic routing, teacher debate, or teacher updates. The main addition would be selecting the appropriate teacher per example before applying the existing distillation loss.

Relevant references:

- On-Policy Distillation: https://thinkingmachines.ai/blog/on-policy-distillation/
- MiMo-V2-Flash technical report, which introduces Multi-Teacher On-Policy Distillation: https://arxiv.org/abs/2601.02780

### Motivation

This would make `DistillationTrainer` more useful for mixed-domain post-training setups where different teacher models are strong in different domains.

Today, users can distill from a single teacher, but if they have separate specialist teachers for math, code, reasoning, safety, or general instruction, there does not seem to be a simple way to route each example to the appropriate teacher within the same training run.

Router-based multi-teacher distillation would allow users to train one deployable student model from multiple specialist teachers while keeping the first version simple and close to the existing `DistillationTrainer` flow.

### Your contribution

I would like to work on this and submit a PR if the maintainers are open to the feature.

I can start with a small scoped implementation that extends DistillationTrainer rather than adding a new trainer, preserves the existing single-teacher behavior, and adds routing from a dataset column to one configured teacher per example.

albertvillanova:
Hey @mohanpb, thanks for the thoughtful, well-scoped proposal — MOPD is a neat direction. 

[...]

PR #7215, opened 2026-09-15
Implement MOPD for the `DistillationTrainer`

cmpatino:
# What does this PR do?

Adds multi-teacher on-policy distillation (MOPD, https://huggingface.co/papers/2606.30406) to the`DistillationTrainer` as an opt-in `teacher_models` argument. Each dataset row selects its teacher through a `teacher_id` column, and the student is distilled against that teacher's full next-token distribution with the existing chunked divergence loss.

The single-teacher `teacher_model` path is unchanged. Registering one teacher through`teacher_models` is bitwise identical to `teacher_model=` (measured on single GPU, DDP and DeepSpeed ZeRO-2).

## Design
- Teachers live on CPU. At each generation boundary, every teacher present in the batch is moved to the accelerator once, scores its rows for all microbatches, and is moved back. Only the completion hidden states are kept on CPU. Device memory during scoring is therefore bounded by the largest teacher, not by the sum of teachers.
- The loss splits each microbatch's rows by teacher and calls the existing `_chunked_divergence_loss` once per group, uploading each teacher's output head as needed. Only the heads of the teachers present in the microbatch are on the device during the loss. Row order is never changed.
- Each checkpoint carries `teacher_manifest.json` (id, source, resolved commit, dtype). Resuming with a different teacher set under the same ids raises.

## Assumptions and restrictions
- All teachers fit in host RAM at once, together with one generation batch of hidden-state
  targets (`per_device_train_batch_size × gradient_accumulation_steps × max_completion_length × teacher hidden_size × dtype bytes`).
- Every teacher shares the student's tokenizer (compared by canonical serialization, padding configuration excluded) and vocabulary size.
- One teacher per row; no aggregation across teachers and no per-teacher loss weights.
- Teachers must be unquantized and loaded without device dispatch or offload hooks; quantized and dispatched teacher backends have not been integrated or validated. `device_map=None` is allowed.
- Text-only students; VLM students keep using `teacher_model`.
- DeepSpeed ZeRO-3 is rejected: the teacher head is a plain device tensor and per-teacher loss groups would issue a rank-dependent number of collectives. ZeRO-1/2 and DDP are supported; under a DeepSpeed mixed-precision plugin, teachers are cast to the engine dtype (parameters only, as the engine does).

Feature requested in https://github.com/huggingface/trl/issues/6151

## Before submitting

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **Overview**
> Unable to generate summary.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed46483c8f356ac7606de713a2e5a35989129d62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/trainer/distillation_trainer.py
```diff
@@ -406,12 +452,28 @@ def __init__(
         quantization_config: "BitsAndBytesConfig | None" = None,
         peft_config: "PeftConfig | None" = None,
         tools: list[Callable] | None = None,
+        teacher_models: "dict[str, str | PreTrainedModel] | None" = None,
```
qgallouedec: api-wise, I feel it's strange to have `teacher_model` and `teacher_models`. Why not simply re-using `teacher_model` and support list/dict?

docs/source/paper_index.md
```diff
@@ -1678,6 +1678,8 @@ training_args = GKDConfig(
...
 
+The synchronous [`DistillationTrainer`] implements the same fusion stage for locally loaded teachers: passing more than one entry in `teacher_models` enables MOPD, with each row's `teacher_id` column selecting which (already-trained, frozen) teacher scores it. See [Multiple teachers (MOPD)](distillation_trainer#multiple-teachers-mopd).
```
diogomsmiranda: Another comment mentions this and I agree that it might make more sense to use this as the behaviour for teacher_model (maybe renaming this?) instead of adding another argument like you have described here: https://github.com/huggingface/trl/pull/7215/changes#diff-3627d7ea9d494ee0001a18064dd53255aed938151295a4bb3af0c0e7cc23b8feR269 

You say that to enable MOPD you just have to instantiate `teacher_models` instead of `teacher_model`. But what happens when `teacher_models={}` or when it's 1 model, does it default back to 1 teacher kd?

trl/trainer/distillation_trainer.py
```diff
@@ -406,12 +417,27 @@ def __init__(
...
     ):
         if args is None:
             model_name = model if isinstance(model, str) else get_config_model_id(model.config)
             model_name = model_name.split("/")[-1]
             args = DistillationConfig(f"{model_name}-Distillation")
 
+        # Multi-teacher mode is opted into by `teacher_models` and is exclusive with the singular teacher entry
+        # points, so a run can never carry two teacher sources with different lifecycles. Checked before any loading.
+        if teacher_models is not None and (teacher_model is not None or args.teacher_model_name_or_path is not None):
```
diogomsmiranda: I think you might be missing a check somewhere here to see if `teacher_models` is not empty. Else you are letting it fail on teacher_id verification or if the dataset doesn't have one then there's no fail guard.

trl/trainer/distillation_trainer.py
```diff
@@ -1853,6 +2118,91 @@ def _compute_loss(self, unwrapped_student, inputs, num_items_in_batch):
...
+            loss = student_hidden_states.new_zeros((), dtype=torch.float32)
+            entropy_sum = student_hidden_states.new_zeros((), dtype=torch.float32)
+            n_valid = student_hidden_states.new_zeros((), dtype=torch.int64)
+            teacher_stats = student_hidden_states.new_zeros((2, len(self.teacher_models)), dtype=torch.float32)
+            # The group losses are summed, so they must share one denominator. `num_items_in_batch` is already one;
+            # where it is `None` (evaluation) the loss would otherwise normalize each group by its own valid-token
+            # count, and the sum of the group means would scale with the number of teachers in the microbatch. The
+            # count matches the one the loss computes from the same mask, so a single group reduces to the legacy
+            # `mean` exactly.
+            denominator = loss_mask.sum().clamp(min=1) if num_items_in_batch is None else num_items_in_batch
```
diogomsmiranda, line 2199: Hey, I revisited the paper that you are using to base this implementation I had a question about the normalisation relative to the paper's Eq. 1. The paper first normalizes each trajectory by its own valid length and then averages trajectories however here you normalize the total divergence by the total valid-token count instead.

So trajectory 'i' receives weight proportional to number of valid tokens, rather than uniform weight 1/B with B being the number of trajectories in the batch.  Routed domains might have systematically different trajectory lengths, bigger trajectories will have higher weights. Although this is equivalent when the size of batches per device equals 1 when >1 it might produce unwanted multi-teacher specific behaviour.

Is this intentional to be coherent with the current `DistillationTrainer`?

diogomsmiranda: For this I believe you would need to track the number of trajectories per batch which you might be able to do in similar faction to what you do with `num_items_in_batch`?

trl/trainer/distillation_trainer.py
cmpatino: I agree, so I included the change in `34874dcc`. 

`teacher_model` now takes either a single teacher (unchanged behavior) or a dict[str, str | PreTrainedModel] that routes rows by their teacher_id column. I kept it to dicts and not list/dict, since routing needs the ids.

I also updated the docstring to emphasize that the type also controls how we handle the teacher. A single teacher keeps the teacher model in the accelerator, while a mapping keeps the teachers on the CPU and moves them one at a time.

The catch is that we'd also move the teachers to a CPU even if the user specifies a mapping with one model (see the example below). I think that's a tradeoff that would keep the API consistent, and would allow us to test parity between single-teacher and MOPD.

For example, I tested the scenario below, and the results are identical, except for how we move the teacher from the CPU in the second case.

```python
teacher_model=teachers["a"]    # single-teacher path
teacher_model={"only": teachers["a"]}    # mapping with one entry
```

trl/trainer/distillation_trainer.py
cmpatino: Good catch, thanks!

docs/source/paper_index.md
cmpatino: That's a good point. I changed the API as I described in [the comment above](https://github.com/huggingface/trl/pull/7215#discussion_r4024553853)

> I agree, so I included the change in `34874dcc`.
> 
> `teacher_model` now takes either a single teacher (unchanged behavior) or a dict[str, str | PreTrainedModel] that routes rows by their teacher_id column. I kept it to dicts and not list/dict, since routing needs the ids.
> 
> I also updated the docstring to emphasize that the type also controls how we handle the teacher. A single teacher keeps the teacher model in the accelerator, while a mapping keeps the teachers on the CPU and moves them one at a time.
> 
> The catch is that we'd also move the teachers to a CPU even if the user specifies a mapping with one model (see the example below). I think that's a tradeoff that would keep the API consistent, and would allow us to test parity between single-teacher and MOPD.
> 
> For example, I tested the scenario below, and the results are identical, except for how we move the teacher from the CPU in the second case.
> 
> ```python
> teacher_model=teachers["a"]    # single-teacher path
> teacher_model={"only": teachers["a"]}    # mapping with one entry
> ```

trl/trainer/distillation_trainer.py
cmpatino: You're right. The paper averages per-trajectory means, while this PR divides the summed divergence by the total number of valid tokens, so longer trajectories weigh more. However, I implemented this version on purpose to keep parity with the single-teacher mode. 

I agree that routing adds a tilt toward the teacher whose domain produces longer completions. However, I'd rather not change the aggregation in this PR to keep it contained to the new feature that keeps parity with the existing behavior.

We can work on a PR that enables choosing the loss type for both multi- and single-teacher paths, as we do in GRPO. The implementation can follow the trajectory count alongside `num_items_in_batch` as you suggest.

trl/trainer/distillation_trainer.py
```diff
@@ -849,13 +955,160 @@ def __init__(
...
+        per_teacher = (self.args.teacher_model_init_kwargs_by_teacher or {}).get(teacher_id, {})
+        init_kwargs = {**common_init_kwargs, **per_teacher}
+        init_kwargs.setdefault("trust_remote_code", self.args.trust_remote_code)
+        # Teachers stay on CPU and are uploaded one at a time, so they are always loaded unmapped.
+        init_kwargs["device_map"] = None
+        commit = None
+        if not isinstance(source, str):
+            teacher, source = source, get_config_model_id(source.config)
+        else:
+            teacher = create_model_from_path(source, **init_kwargs)
```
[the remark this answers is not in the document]
cmpatino: Confirmed, thanks. Fixed in `df69c79f` by mirroring the single-teacher path's `setdefault("revision", ...)`. A test covers both directions through the manifest's resolved commit.

trl/trainer/distillation_trainer.py
diogomsmiranda: lgtm! Once this is merged I can open a draft PR with the option to choose the loss type if you agree.

kashif (approved):
Looks good to me. The synchronous MOPD implementation is consistent with the existing DistillationTrainer design, preserves the single-teacher path, and has strong unit and distributed coverage.

One small nonblocking improvement that could be good to add: validate that every key in teacher_model_init_kwargs_by_teacher exists in the teacher_model mapping. Currently a typo in a teacher ID is silently ignored, which could cause that teacher to load with the common/default revision or dtype instead of the intended override. This should only require a small set-difference check and a focused test.

docs/source/distillation_trainer.md
```diff
@@ -262,6 +262,92 @@ Tested with:
...
+    model="Qwen/Qwen2.5-0.5B-Instruct",
+    teacher_model={
+        "math": "Qwen/Qwen2.5-Math-1.5B-Instruct",
+        "code": "Qwen/Qwen2.5-Coder-1.5B-Instruct",
+    },
+    train_dataset=dataset,
+    args=DistillationConfig(
+        output_dir="mopd-student",
+        # Route one teacher to a specific revision while the rest load normally.
+        teacher_model_init_kwargs_by_teacher={"code": {"revision": "<revision-name>"}},
```
qgallouedec, line 295: `teacher_model_init_kwargs_by_teacher`? simply `teacher_model_init_kwargs` instead no?

docs/source/distillation_trainer.md
```diff
@@ -262,6 +262,92 @@ Tested with:
...
+    ),
+)
+trainer.train()
+```
+
+### Requirements
+
+- Every teacher must share the student's vocabulary, i.e. the same `vocab_size`. Prompts are rendered once with the student's tokenizer, and every teacher scores those exact token IDs, so a teacher trained on a different tokenizer is training the student against the wrong tokens. Use [GOLD](gold_trainer) for cross-tokenizer distillation.
```
qgallouedec, line 303: just to be precise, I don't think that sharing the same vocab size is enough. It should share the same vocabulary. So imo just
```suggestion
- Every teacher must share the student's vocabulary. Prompts are rendered once with the student's tokenizer, and every teacher scores those exact token IDs, so a teacher trained on a different tokenizer is training the student against the wrong tokens. Use [GOLD](gold_trainer) for cross-tokenizer distillation.
```

docs/source/distillation_trainer.md
```diff
@@ -262,6 +262,92 @@ Tested with:
...
+- Teachers must be unquantized teachers loaded without device dispatch or offload hooks. Teachers move between CPU and the accelerator as complete modules, and the loss reads their output heads as dense matrices; quantized and dispatched teacher backends have not been integrated or validated. A `quantization_config` and an active `device_map` are rejected at initialization, per teacher after the per-teacher overrides are merged, and every loaded teacher is checked for quantization and device dispatch — including one loaded from a checkpoint that carries its quantization in its own config. `device_map=None` is allowed.
```
qgallouedec, line 304:
```suggestion
- Teachers must be unquantized loaded without device dispatch or offload hooks. Teachers move between CPU and the accelerator as complete modules, and the loss reads their output heads as dense matrices; quantized and dispatched teacher backends have not been integrated or validated. A `quantization_config` and an active `device_map` are rejected at initialization, per teacher after the per-teacher overrides are merged, and every loaded teacher is checked for quantization and device dispatch — including one loaded from a checkpoint that carries its quantization in its own config. `device_map=None` is allowed.
```

tests/distributed/test_distillation_trainer_multi_teacher.py
qgallouedec, line 1: I reviewed this test file and the other, and I think it's not how we want to test that. Mostly because we test on GPU (here it's trying to test on CPU), and it contains flaky patterns like the `SimpleNamespace`, and private method testing that we try to avoid as much as possible.
So let's remove these two files. And instead either:
- add a test case in `test_distributed.py` following the existing convention
- or, if it's to hard, because you need to change the script or I don't know, let's just add this in a follow-up PR.

trl/trainer/distillation_trainer.py
```diff
@@ -396,7 +406,7 @@ class DistillationTrainer(_BaseTrainer):
     def __init__(
         self,
         model: "str | PreTrainedModel | PeftModel",
-        teacher_model: str | PreTrainedModel | None = None,
+        teacher_model: "str | PreTrainedModel | dict[str, str | PreTrainedModel] | None" = None,
```
qgallouedec, line 409: why not

```suggestion
        teacher_model: str | PreTrainedModel | dict[str, str | PreTrainedModel] | None = None,
```
?

qgallouedec (commented):
Initial review, I'll need more time

PR #7216, opened 2026-09-15
Reduce head_dim in the tiny Gemma generator

albertvillanova:
The tiny Gemma model is not scaled down in attention. `GemmaConfig.head_dim` defaults to `256` and the script never passes it, so the tiny model keeps the reference's head dimension while its hidden size is 8:

| | `hidden_size` | `num_attention_heads` | `head_dim` | `q_proj` |
|---|---|---|---|---|
| `google/gemma-7b-it` | 3072 | 16 | 256 | (4096, 3072) |
| `tiny-GemmaForCausalLM` | 8 | 4 | 256 | (1024, 8) |

The attention projections are 128 times wider than the hidden size they read from, a shape no real Gemma has. `print_config_diff` cannot see it, because the reference is also `256` and the row never appears.

#7137 lists `head_dim` among the fields the tiny models are meant to reduce, so this is a missed scale-down
rather than a config divergence, and it is fixed here on its own, before the alignment in the follow-up PR.

### Solution

Pass `head_dim=2`, which is `hidden_size // num_attention_heads` and restores `q_proj` to `(8, 8)`.

### Effect

```
q_proj:  (1024, 8)  ->  (8, 8)
params:  2,098,728  ->  2,049,960
```

The remaining parameters are almost entirely the `256000 x 8` embedding, so the absolute saving is small; the point is the shape.

```
[config_diff] google/gemma-7b-it vs tiny (8 differences)
  head_dim                                         256                                → 2
  hidden_act                                       gelu                               → gelu_pytorch_tanh
  hidden_size                                      3072                               → 8
  intermediate_size                                24576                              → 32
  num_attention_heads                              16                                 → 4
  num_hidden_layers                                28                                 → 2
  num_key_value_heads                              16                                 → 2
  rope_scaling                                     None                               → <missing>
```

This is the state *after* this PR, not its delta. The only change here is `head_dim`, which is why the row count goes up from 7 to 8: it now shows as the deliberate reduction it should always have been. The `hidden_act` and `rope_scaling` rows are pre-existing on `main` and untouched here; `hidden_act` is fixed in the follow-up, and `rope_scaling` is a reference key `GemmaConfig` does not model.


albertvillanova:

- [x] https://huggingface.co/trl-internal-testing/tiny-GemmaForCausalLM/discussions/1

albertvillanova:
Tested with `MODEL_REVISIONS` pointing at the regenerated model, `trl-internal-testing/tiny-GemmaForCausalLM` at `refs/pr/1`, on top of current `main`: https://github.com/huggingface/trl/actions/runs/34982142238

The suite is green, with a single unrelated CUDA OOM failure on a different tiny model.

albertvillanova:
Merging this PR as a CI tiny-model shape fix and as part of an approved series.

PR #7226, opened 2026-09-16
Document that AsyncGRPO tools must be module-level functions

albertvillanova:
This PR corrects the picklability guidance for AsyncGRPO, which recommends two callable forms that a tool cannot take.

### Motivation

Everything handed to the rollout worker is pickled, so the user guide, the `reward_funcs` docstring and the `TypeError` raised by `AsyncRolloutWorker.start()` all say to pass a module-level function, a `functools.partial`, or a callable class instance. That is right for `reward_funcs` and for `environment_factory`, and wrong for `tools`.

The model calls a tool by name, so the worker registers tools under their `__name__`. Neither a `functools.partial` nor a callable class instance has a `__name__`, so following that advice for a tool raises an `AttributeError` inside the rollout process the first time it collects the tools for a batch, which surfaces as a worker failure rather than as an error against the tool that caused it.

The error message is the sharpest of the three: it is raised precisely when a user has just had an unpicklable tool rejected and is looking for the fix, so it hands them a form that trades a clear failure at startup for an obscure one during generation.

### Solution

Name the exception in all three places: tools are registered under their `__name__`, so a tool has to be a module-level function. Methods discovered from an `environment_factory` are unaffected, since they carry their own name. The pickling requirement itself is unchanged, and it still applies to tools.

### Changes

- Correct the picklability warning in the AsyncGRPO guide
- Correct the same guidance in the `reward_funcs` docstring and in the `TypeError` raised when the worker starts

---

> [!NOTE]
> **Low Risk**
> Documentation and error-message text only; no runtime behavior changes.
> 
> **Overview**
> Clarifies AsyncGRPO picklability guidance so **tools** are not lumped in with `reward_funcs` and `environment_factory`.
> 
> The user guide, `AsyncGRPOTrainer` `reward_funcs` docstring, and the `TypeError` from `AsyncRolloutWorker.start()` now state that tools are registered under their **`__name__`**, so each tool must be a **module-level function**—not `functools.partial` or a callable class instance, which lack a reliable `__name__` and would fail during rollout. Pickling rules for tools are unchanged; only the documented callable shapes are corrected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 801b5efb11ba73b967fab95d6b8fa88a93f7805b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (approved):
pinging @AmineDiro for viz, but lgtm!

AmineDiro (approved):

PR #7227, opened 2026-09-16
Reduce head_dim in the tiny Gemma2 generator

albertvillanova:
The tiny Gemma2 model is not scaled down in attention, the same defect fixed for Gemma in #7216. `Gemma2Config.head_dim` defaults to `256` and the script never passes it, so the tiny model keeps the reference's head dimension while its hidden size is 8:

| | `hidden_size` | `num_attention_heads` | `head_dim` | `q_proj` |
|---|---|---|---|---|
| `google/gemma-2-2b-it` | 2304 | 8 | 256 | (2048, 2304) |
| `tiny-Gemma2ForCausalLM` | 8 | 4 | 256 | (1024, 8) |

The attention projections are 128 times wider than the hidden size they read from, a shape no real Gemma2 has. `print_config_diff` cannot see it, because the reference is also `256` and the row never appears.

#7137 lists `head_dim` among the fields the tiny models are meant to reduce, so this is a missed scale-down rather than a config divergence, and it is fixed here on its own, before the alignment in the follow-up PR.

### Solution

Pass `head_dim=2`, which is `hidden_size // num_attention_heads` and restores `q_proj` to `(8, 8)`.

`query_pre_attn_scalar` is reduced to `2` alongside it. It is the attention scaling divisor (`self.scaling = config.query_pre_attn_scalar ** -0.5`), and the reference sets it to `256`, equal to its `head_dim`, which is the usual `1/sqrt(head_dim)` relationship. Mirroring the reference's value while shrinking `head_dim` would keep a piece of the 2B model's geometry in a model whose heads are 2 wide, so the relationship is preserved instead of the number.

### Effect

```
q_proj:   (1024, 8)  ->  (8, 8)
params:   2,098,760  ->  2,049,992
scaling:   0.062500  ->  0.707107
```

As with Gemma, the remaining parameters are almost entirely the `256000 x 8` embedding, so the absolute saving is small. The point is the shape.

```
[config_diff] google/gemma-2-2b-it vs tiny (11 differences)
  cache_implementation                             hybrid                             → <missing>
  eos_token_id                                     [2 items] [1, 107]                 → 1
  head_dim                                         256                                → 2
  hidden_act                                       gelu_pytorch_tanh                  → <missing>
  hidden_size                                      2304                               → 8
  intermediate_size                                9216                               → 32
  layer_types                                      [26 items] ['sliding_attention', … → [2 items] ['sliding_attention', '…
  num_attention_heads                              8                                  → 4
  num_hidden_layers                                26                                 → 2
  num_key_value_heads                              4                                  → 2
  query_pre_attn_scalar                            256                                → 2
```

This is the state after this PR, not its delta. The only changes here are `head_dim` and `query_pre_attn_scalar`, which is why the row count goes up from 9 to 11: both now show as the deliberate reductions they should always have been. The other rows are pre-existing on `main` and untouched here. `eos_token_id` is fixed in the follow-up; `cache_implementation` and `hidden_act` are reference keys that `Gemma2Config` does not model.

---

> [!NOTE]
> **Low Risk**
> Only changes the tiny Gemma2 Hub generation script; no runtime library or production model behavior unless the Hub artifact is regenerated.
> 
> **Overview**
> The tiny **Gemma2** generator now passes **`head_dim=2`** (`hidden_size // num_attention_heads`) so attention projections scale with the 8-wide hidden state instead of keeping the default **256**, fixing **`q_proj`** from **(1024, 8)** to **(8, 8)**.
> 
> **`query_pre_attn_scalar=2`** is set alongside **`head_dim`** so attention scaling stays **`1/sqrt(head_dim)`** rather than mirroring the 2B model’s **256** divisor on 2-wide heads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ade1674c97c1dd2c63c6f863c1f12a43314708d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:

- [x] https://huggingface.co/trl-internal-testing/tiny-Gemma2ForCausalLM/discussions/1

albertvillanova:
Tested with `MODEL_REVISIONS` pointing at the regenerated model, `trl-internal-testing/tiny-Gemma2ForCausalLM` at `refs/pr/1`, on top of current `main`: https://github.com/huggingface/trl/actions/runs/35082990088

The suite is green.

albertvillanova:
Merging this PR as a CI tiny-model shape fix and as part of an approved series.

PR #7230, opened 2026-09-16
Make the end-of-turn loss mask warning actionable

itwangwang:
# What does this PR do?

Expands the `assistant_only_loss` end-of-turn warning to state the consequence
and point at the check users can run before launching a training run.
Message text only — no behavioral change.

## Why

I hit this while fine-tuning Qwen3.5-4B with QLoRA for a structured tool-calling
task (natural-language ops requests -> JSON tool calls).

The current warning says the end-of-turn token is missing from the loss mask, but
not what that leads to. I missed it in the training log and only found out after a
2h20m training run plus a full evaluation pass: strict JSON rate came out at 79.7%,
below both the zero-shot (100%) and few-shot (100%) baselines, because the model had
never learned to stop and kept generating past the JSON object.

This is exactly what #5988 described when it introduced the check:

> SFT runs fine, loss goes down, and the bug only shows up at inference as endless
> generations.

The detection logic is already there and works correctly — it fired for me. The gap
is that the message doesn't convey that this is a silent failure whose cost lands
hours later, so it's easy to scroll past among the other startup logs. Naming the
consequence and pointing at `is_chat_template_stop_token_trained()` would have saved
the run.

Fixes # N/A — no existing issue; reporting from first-hand usage.

## Before submitting

- [x] Did you read the [[contributor guideline](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request)](https://github.com/huggingface/trl/blob/main/CONTRIBUTING.md#create-a-pull-request), Pull Request section?

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **Low Risk**
> Log message only in `SFTTrainer.__init__`; no logic, defaults, or training path changes.
> 
> **Overview**
> **Expands the startup warning** when `assistant_only_loss=True` and `is_chat_template_stop_token_trained()` fails, so the message is easier to act on without changing any training behavior.
> 
> The added text explains that **training loss can look fine while the problem only shows up at inference** (the model may not learn to stop). It also lists two fixes: set `assistant_only_loss=False`, or put the assistant end-of-turn token inside `{% generation %}...{% endgeneration %}` in the chat template.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ca5121f1eaa8adfef3f2ddad940d3750213a6b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/trainer/sft_trainer.py
```diff
@@ -1273,7 +1273,10 @@ def __init__(
         ):
             logger.warning(
                 "The chat template does not include the assistant turn's end-of-turn token in the loss mask; "
-                "the model may not learn to stop."
+                "the model may not learn to stop. Training will look healthy - the loss still goes down - but "
+                "generations may run until `max_new_tokens` at inference time, which is especially damaging for "
+                "structured outputs such as JSON or tool calls. To check this before launching a run, call "
+                "`is_chat_template_stop_token_trained(processing_class)` from `trl.chat_template_utils`."
```
qgallouedec: This text says how to *check* for the problem, not what to do about it, so it's not quite "actionable". The user has already had it detected, that warning is the check. 

What a user can actually change is `assistant_only_loss`, or the template. Maybe something like:

> either set `assistant_only_loss=False` to train on the full sequence, or edit the chat template so the end-of-turn token falls inside `{% generation %}...{% endgeneration %}`.

itwangwang: You're right — the warning *is* the check, so pointing back at it was circular.
Updated to your suggestion. I kept one short clause on the training loss still
looking healthy, since that's what made me scroll past it in the first place,
but dropped the rest.

qgallouedec (approved):
thanks

PR #7232, opened 2026-09-16
Pass ffn_dim instead of intermediate_size in the tiny OPT generator

albertvillanova:
The tiny OPT model is not scaled down in its feed-forward block. The script passes `intermediate_size=32`, but `OPTConfig` has no such field: OPT spells it `ffn_dim`. The value is silently stored as an extra attribute and `ffn_dim` keeps its class default of `3072`, so the FFN is built at 3072 wide on a hidden size of 8.

```
fc1:     (3072, 8)   <- 3072 is the OPTConfig default, not the 32 the script asked for
params:  523,696     of which ~98k is the oversized FFN
```

`print_config_diff` did show this, as `ffn_dim 8192 → 3072` next to a stray `intermediate_size <missing> → 32`, but the pair reads like a naming quirk rather than a model that was never shrunk.

This is the same class of missed scale-down as #7216 and #7227, where `head_dim` kept a constant class default. The difference is the cause: there the default was a constant that happened to match the reference, here the script names the field wrong so the intended value never lands at all.

### Solution

Pass `ffn_dim=32` instead of `intermediate_size=32`.

### Effect

```
fc1:     (3072, 8)  ->  (32, 8)
params:   523,696   ->  420,336
```

```
[config_diff] facebook/opt-1.3b vs tiny (9 differences)
  activation_dropout                               0.0                                → <missing>
  ffn_dim                                          8192                               → 32
  hidden_size                                      2048                               → 8
  num_attention_heads                              32                                 → 4
  num_hidden_layers                                24                                 → 2
  num_key_value_heads                              <missing>                          → 2
  prefix                                           </s>                               → None
  vocab_size                                       50272                              → 50265
  word_embed_proj_dim                              2048                               → 8
```

This is the state after this PR, not its delta. Two rows resolve at once: `ffn_dim` becomes the deliberate reduction it was meant to be, and the stray `intermediate_size` row disappears. The rest are pre-existing on `main` and untouched here. `vocab_size` and `num_key_value_heads` are handled in the follow-up; `activation_dropout` and `prefix` are reference keys that `OPTConfig` does not model.

`word_embed_proj_dim` is correct as it stands and is left alone. Its class default is `None`, resolving to `hidden_size`, so it tracks the scale-down on its own, and the reference has `word_embed_proj_dim == hidden_size == 2048`, meaning no `project_in`/`project_out` layers. The tiny model reproduces that. Worth noting for future references: `facebook/opt-350m` has `hidden_size=1024` against `word_embed_proj_dim=512`, and a tiny model built from that one would need the field set explicitly to keep the projection layers.

---

> [!NOTE]
> **Low Risk**
> One-line change in an offline tiny-model generation script; no runtime library or production inference path.
> 
> **Overview**
> Fixes the tiny OPT causal-LM generator so the feed-forward block actually scales down: **`OPTConfig` is constructed with `ffn_dim=32` instead of `intermediate_size=32`**, which OPT does not recognize.
> 
> Previously the intended width was ignored and `ffn_dim` stayed at its default (~3072), leaving an oversized FFN on `hidden_size=8`. After the change, the tiny checkpoint should match the intended ~32-wide FFN and drop parameter count accordingly; regenerating the Hub tiny model is required for the fix to apply there.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 239e3faaf632a74db9a142745230844254a8f6ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:
This PR requires #7238 to land first.

Regenerating the tiny OPT model is currently impossible: `check_dtype_pattern` calls `HfApi.get_safetensors_metadata` on `facebook/opt-1.3b`, which predates safetensors and ships only `pytorch_model.bin`, so `NotASafetensorsRepoError` propagates and the generator dies before `push_to_hub`. #7238 catches it and turns the check into a skip.

Once that is merged, this PR's Hub regeneration and CI verification will follow here as usual.

qgallouedec:
LGTM

qgallouedec (approved):

albertvillanova:

- [x] https://huggingface.co/trl-internal-testing/tiny-OPTForCausalLM/discussions/1

Model size reduction: from 1_051_128 to 844_368.

albertvillanova:
Tested with `MODEL_REVISIONS` pointing at the regenerated model, `trl-internal-testing/tiny-OPTForCausalLM` at `refs/pr/1`, on top of current `main`: https://github.com/huggingface/trl/actions/runs/35189278734

The suite is green.

PR #7238, opened 2026-09-16
Skip the dtype check when the reference is not a safetensors repo

albertvillanova:
`check_dtype_pattern` aborts the generator whenever the reference model is not a safetensors repo.

`facebook/opt-1.3b` predates safetensors and ships only `flax_model.msgpack`, `pytorch_model.bin` and `tf_model.h5`, so `HfApi.get_safetensors_metadata` raises and the exception is never caught:

```
huggingface_hub.errors.NotASafetensorsRepoError: 'facebook/opt-1.3b' is not a safetensors repo.
Couldn't find 'model.safetensors.index.json' or 'model.safetensors' files.
```

The traceback lands before `print_config_diff` and `push_to_hub`, so `python -m scripts.generate_tiny_models.for_causal_lm.opt_for_causal_lm` cannot complete at all, with or without `--create-pr`. The tiny OPT model on the Hub predates this check being added, which is why nobody hit it until now.

The check is diagnostics: it prints per-tensor dtype mismatches and nothing gates on its output. A reference simply not being a safetensors repo says nothing about the tiny model, so it should not be able to stop the run.

### Solution

Catch `NotASafetensorsRepoError`, print that the check was skipped, and return.

```
[dtype_check] facebook/opt-1.3b: not a safetensors repo, skipping
```

The per-tensor comparison is genuinely unavailable here, so it is reported as a skip rather than a pass. A config-level `torch_dtype` comparison was considered as a fallback and rejected: that field is a single scalar for the whole checkpoint, which is blind to the mixed-precision patterns this function exists to catch (its docstring cites Qwen3.5 keeping `linear_attn.A_log` in fp32 inside a bf16 checkpoint), so reporting it as a pass would overclaim.

`facebook/opt-1.3b` is the only non-safetensors reference among the generator scripts today; every other tiny model regenerates fine.

### Verification

The OPT generator now runs end to end:

```
[smoke_test] OPTForCausalLM: OK (output shape (1, 4, 50265))
[dtype_check] facebook/opt-1.3b: not a safetensors repo, skipping
[config_diff] facebook/opt-1.3b vs tiny (10 differences)
```

and a safetensors reference is unaffected:

```
[dtype_check] Qwen/Qwen3-8B: all matched tensors have the reference dtype
```

No tiny model changes, so no Hub regeneration is needed for this PR.

---

> [!NOTE]
> **Low Risk**
> Diagnostic-only path in an internal script; behavior for safetensors references is unchanged.
> 
> **Overview**
> **`check_dtype_pattern`** in the tiny-model generator no longer crashes when the reference Hub repo has no safetensors weights (e.g. `facebook/opt-1.3b`).
> 
> It now catches **`NotASafetensorsRepoError`** from `get_safetensors_metadata`, logs a skip message, and returns so generation can continue to config diff and hub push. Safetensors references still get the full per-tensor dtype comparison.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63d16534631d6e69128e90de90a8e2c04ecd780c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
so for these we check by hand? I'm ok with it of so.

qgallouedec (approved):

PR #7243, opened 2026-09-16
Move DPO to chunked log probabilities

kashif:
# What does this PR do?

Moves DPO onto the shared chunked log-probability path while keeping the trainer loss math in one place. This removes the duplicated fused DPO loss and lifts its restrictions on mixed losses, f-divergences, and precomputed reference log-probs.

Part of #7063.

## Before submitting

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **High Risk**
> This rewrites the default memory-efficient DPO training path (log-prob computation, gradients, and distributed wrapping) while removing the old fused implementation; regressions could affect loss values, FSDP2/ZeRO-3 training, and reference precomputation.
> 
> **Overview**
> **DPO with `use_liger_kernel=True` now uses the same chunked log-probability path as other TRL trainers** instead of the removed `FusedLinearDPOLoss` / Liger fused kernels. Policy and reference log-probs are computed via backbone hidden states and `_ChunkedLogProbFunction` (no full `[batch, seq, vocab]` logits), then the existing `_compute_loss` preference math runs unchanged.
> 
> **Capability and packaging changes:** mixed `loss_type` lists, non-default `f_divergence_type`, and `precompute_ref_log_probs=True` are allowed on the chunked path; **`use_weighting=True` is newly rejected** (replacing the old f-divergence block). Docs reflect this. **`trl/losses/dpo_loss.py`**, **`fused_linear_preference.py`**, and **`tests/test_dpo_loss.py`** are deleted; `FusedLinearDPOLoss` is dropped from public exports.
> 
> **Distributed / correctness:** FSDP2 is supported by gathering sharded `lm_head` DTensors before projection; distributed **`test_dpo_liger` no longer xfail on `fsdp2`**. Reference log-prob precompute and training go through `_forward_redirection` when wrapped (ZeRO-3/FSDP/DDP). Trainer tests add parity checks (loss types, f-divergences, precomputed refs, `output_multiplier`, bf16, PEFT guards).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 55f4e3c5b5aaead7df72012852f664e6333ee390. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/distributed/test_distributed.py
```diff
@@ -206,14 +206,7 @@ def test_dpo_precompute_ref_log_probs(self, config, get_config_path):
...
             ),
-            pytest.param(
-                "fsdp2",
-                marks=pytest.mark.xfail(
-                    reason="Liger DPO loss reads `lm_head.weight` and runs the backbone directly, which is "
-                    "incompatible with FSDP2's DTensor-sharded parameters (mixed Tensor/DTensor ops).",
-                    strict=True,
-                ),
-            ),
+            "fsdp2",
```
qgallouedec, line 209: cool!

trl/trainer/dpo_trainer.py
```diff
@@ -797,69 +800,52 @@ def __init__(
...
 
-        # Liger loss
+        # Chunked log-probability path
         self.use_liger_kernel = args.use_liger_kernel
         if self.use_liger_kernel:
             if not is_liger_kernel_available():
                 raise ImportError(
                     "You set `use_liger_kernel=True` but the liger kernel is not available. "
                     "Please install liger-kernel first: `pip install liger-kernel`"
                 )
```
qgallouedec: do we still need this?

qgallouedec (commented):
Benchmarked it (single H100, Qwen3-0.6B, bs=4, seq 512)

| config | median step | peak |
|---|---|---|
| main, fused, inline ref | 0.2243 s (n=4) | 7.05 GB |
| this PR, inline ref | 0.2457 s (n=6) | 7.05 GB |
| this PR, `precompute_ref_log_probs=True` | 0.1971 s (n=2) | 4.83 GB |

+9.5% per step at equal config, -12.1% with precomputed ref logps, which main rejects outright. Memory unchanged in the first case, 31% lower in the second.

So nothing unexpected!

A few things:
- For the record, I found that `compute_loss` redirects on `is_zero3 or self.is_fsdp_enabled`, so plain DDP falls through to `_compute_loss(unwrapped_model, ...)` and `prepare_for_backward()` never fires. That is inherited from main so leaving it is right. I'll open an issue to fix it across trainers.

- Docs need updating, `dpo_trainer.md` has a "Compatibility and constraints" section listing, under `use_liger_kernel=True`: only a single `loss_type`, no `compute_metrics`, no `precompute_ref_log_probs`. This PR lifts the first and the third and adds a new one (`use_weighting=True`), so three lines there are now wrong.

- Same section is where the f-divergence lift belongs, it was only ever enforced in code.

qgallouedec:
Related to #7245: DPO silently skips DDP gradient synchronization with `use_liger_kernel=True`, and this PR is what makes it fixable.

`compute_loss` redirects only on `is_zero3 or self.is_fsdp_enabled`, so under plain DDP the loss runs on the unwrapped model, `DistributedDataParallel.forward()` never runs, its reducer is never armed, and gradients are never all-reduced. Measured on 2 GPUs with different data per rank: 310/310 parameters differ across ranks, on main and on this branch alike.

On main the one-line fix does not work. Adding `or model is not unwrapped_model` makes every rank raise `NotImplementedError: Cannot access storage of TensorWrapper`, because `FusedLinearDPOFunction` runs `torch.func.grad_and_value` and that cannot run inside `DistributedDataParallel.forward()`.

On this branch the same change gives 0/310. `_ChunkedLogProbFunction` is a plain `autograd.Function`, so the redirect is fine.

So the DPO half of #7245 belongs here, as one line in `compute_loss`:

```python
if is_zero3 or self.is_fsdp_enabled or model is not unwrapped_model:
```

KTO is going separately, it is already on the chunked path.

qgallouedec:
Thanks!, both points addressed. Three things left:

- The guard has to match KTO on main (#7247): `if self.is_fsdp_enabled or model is not unwrapped_model:`, and the `deepspeed_plugin` / `is_zero3` lines go. Same block, same text.
- Merge main, the branch predates #7247.
- Min-versions CI is red on all five gradient tests, and it's not noise: the mismatch sits on a few rows of one 1.2M-element parameter. Reduction-order differences don't look like that. Worth finding what the old torch does differently there before touching the tolerance.

trl/trainer/dpo_trainer.py
kashif: Yes, still needed: same restriction as KTO's already-merged chunked migration (#7075), for the same reason - compute_metrics needs the full logits, which the chunked path never materializes.

trl/trainer/dpo_trainer.py
```diff
@@ -1763,17 +1749,16 @@ def evaluate(
...
-                # Under ZeRO-3, `lm_head.weight` is sharded and the fused loss reads it directly (bypassing the
-                # module), so run the loss inside the engine's forward via `_forward_redirection` to arm the parameter
-                # coordinator's gather/reduce hooks.
+                # The chunked projection reads `lm_head.weight` directly, but distributed wrappers still need their
+                # forward to run so their parameter materialization and gradient synchronization hooks are armed.
                 deepspeed_plugin = self.accelerator.state.deepspeed_plugin
                 is_zero3 = deepspeed_plugin is not None and deepspeed_plugin.zero_stage == 3
                 unwrapped_model = self.accelerator.unwrap_model(model)
-                if is_zero3 or self.is_fsdp_enabled:
+                if is_zero3 or self.is_fsdp_enabled or model is not unwrapped_model:
```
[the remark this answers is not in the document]
kashif: Resolved: just matched this guard to KTO's on main (#7247) directly, so both trainers redirect on the same condition now.

kashif:
Three things from the last review:

- Guard now matches KTO's on main (#7247) exactly: dropped `is_zero3`/`deepspeed_plugin`, same condition, same comment.
- Rebased on main (also picks up #7247 and #7077, which had both landed since).
- Min-versions CI: root-caused it rather than just loosening the tolerance. Reproduced with the pinned accelerate/datasets/transformers versions on ROCm - passes cleanly, so it's not a dependency-version issue. The failing parameter is always `lm_head.weight`; this test model's `hidden_size=8` makes the chunked backward's weight-grad GEMM (roughly `(chunk_size, N) x (N, 8)`) unusually narrow, a shape the non-chunked path's full-width GEMM never produces. That's consistent with cuBLAS picking a different, less precise algorithm for it specifically on Ampere+ (observed up to 3.4e-2 on the CI's L40S), not a chunking bug - ROCm gives an exact match. Gave just that one parameter a wider tolerance instead of loosening all of them.

qgallouedec (commented):
Thanks! A few drift points left:

- `_get_per_token_logps_and_entropies`: the write-back still uses `zeros_like` + `t[mask] = v`. KTO and GRPO moved to `new_zeros(mask.shape).masked_scatter(mask, v)` in #7077, which landed after this branch. Please take KTO's version verbatim and port `test_chunked_logps_stay_differentiable_when_all_masked` with it.
- Docs: can the liger restrictions use KTO's shape, a plain list of what is restricted (`compute_metrics`, PEFT on `lm_head`, prompt-learning PEFT, MoE aux loss, plus `use_weighting` here)? Drop the sentence about what is now supported. Also add KTO's "omitted when `use_liger_kernel=True`" note to `logits/chosen` and `logits/rejected`, since they are no longer logged.
- Tests: `test_liger_loss_types_match_non_liger_loss` is KTO's `test_liger_loss_matches_non_liger_loss`, keep the name. `test_chunked_logps_honor_output_multiplier_without_calling_lm_head` covers code shared with KTO but has no KTO counterpart, either add it there too or leave it for a follow-up.

kashif:
Thanks @qgallouedec, addressed in 3018d47a and bc50642f:

- mask write-back now uses `new_zeros(mask.shape).masked_scatter(...)`, same as KTO, so `_get_per_token_logps_and_entropies` is identical in both trainers
- ported `test_chunked_logps_stay_differentiable_when_all_masked` from KTO
- renamed to `test_liger_loss_matches_non_liger_loss` to match KTO
- docs: the `use_liger_kernel=True` constraints now have the same shape as KTO (plus `use_weighting`), and the `logits/*` metrics say they're omitted on the chunked path

KTO doesn't have the output_multiplier test yet. I'll add it in a follow-up so this PR stays DPO-only.

albertvillanova (changes requested):
Thanks, the KTO alignment is now exact: `_get_per_token_logps_and_entropies` and `compute_loss` diff clean against main.

Two things before this can go in:

1. The `lm_head.weight` tolerance can go back to 5e-4. The min-versions failure on c2196d5f5 was not cuBLAS: it reproduces bit for bit (0.0337 at index (2664, 4)) only when `test_train_with_liger` runs earlier in the same process, and passes when the parity tests run alone. On the current head it no longer happens: at a uniform 5e-4 the parity tests pass alone, right after `test_train_with_liger`, in the exact sequence the failing worker ran, and in the full suite on the min-versions setup (https://github.com/huggingface/trl/actions/runs/35850265163, https://github.com/huggingface/trl/actions/runs/35848993881). So the special case, and the comment explaining it, can be dropped, which also keeps the test identical to KTO's.

2. Please keep the `is_liger_kernel_available()` guard. KTO and GRPO both still have it, so dropping it here is drift. If it is redundant with transformers' own check, that is worth a separate PR across all three trainers.

Smaller:
- `assert_verbose_allclose` and `HFAlignmentLoss` in `tests/testing_utils.py` were only used by `tests/test_dpo_loss.py`, so they are dead now.
- `@require_liger_kernel` on the `_assert_liger_loss_matches` helper does nothing.
- `mean_token_accuracy` is also not logged on the chunked path; it could get the same "omitted" note as `logits/*`.

Nice catch on `use_weighting`: on main the Liger path silently ignored it, so the new error is a fix, not only a restriction.

kashif:
Thanks @albertvillanova, all done in 3f4cfc64:

1. `lm_head.weight` tolerance is back to a uniform 5e-4, and I took KTO's comment so the helper matches KTO's. Passes alone and right after `test_train_with_liger`, same as you saw. Thanks for tracking down the real cause.
2. Put the `is_liger_kernel_available()` guard back, same text as KTO.

Smaller:
- removed `assert_verbose_allclose` and `HFAlignmentLoss` from `tests/testing_utils.py`. `HFDistillationLoss` too, it's been dead since #7301 dropped `test_jsd_loss.py`
- dropped `@require_liger_kernel` from the helper
- `mean_token_accuracy` gets the same "omitted" note as `logits/*`

albertvillanova (commented):
Follow-up on the `lm_head.weight` tolerance: the cause is the `ACCELERATE_MIXED_PRECISION` leak between tests fixed in #7321. c2196d5f5 predates it, and the branch has it since the last main merge, which is why the parity tests now pass at 5e-4 in the failing sequence. The same leak is behind the GRPO failure #7351 addresses.

albertvillanova (approved):
Thanks, all addressed: the gradient check now matches KTO's exactly, the guard is back, and the dead test helpers are gone. LGTM.

PR #6372, opened 2026-07-13
Fix DPO/KTO `use_liger_kernel` under DeepSpeed ZeRO-3

qgallouedec:
`use_liger_kernel=True` + DeepSpeed ZeRO-3 in `DPOTrainer` crashes. This mirrors the GRPO bug fixed in #5891.

- ✅ liger + zero2
- ✅ no-liger + zero3
- ❌ liger + zero3 → crash

Closes #4544.

## Why

The Liger fused DPO loss reads `lm_head.weight` directly instead of going through `model.forward()`. Under ZeRO-3 this breaks in **two** places:

1. **Forward.** ZeRO-3 shards every parameter to `numel 0` and only re-materializes it inside its owning module's forward hook. Since Liger accesses `lm_head.weight` by attribute (never calling the `lm_head` module), the gather hook never fires and the fused matmul gets an empty shard → `size mismatch, got input (s2x2048), vec (0)`.
2. **Backward.** Even once the weight is gathered for the forward, the fused loss produces a *dense* weight gradient. If the fused loss doesn't run inside the DeepSpeed engine's `forward()`, ZeRO-3's pre-forward hooks are never armed, so the backward gradient reduction isn't set up and grad accumulation fails → `size of tensor a (0) must match the size of tensor b (H)`.

## Fix

Apply the same two-part treatment GRPO already uses (#5891):

1. **`maybe_gather_lm_head_ctx`** (new shared helper in `trainer/utils.py`) all-gathers the `lm_head` weight/bias for the duration of the fused call. No-op when not on ZeRO-3, or when the weight is already gathered (tied embeddings). Deduplicates by identity so a shared policy/reference head (PEFT with no separate `ref_model`) isn't passed to `GatheredParameters` twice.
2. Run `_compute_loss_liger` through **`_ForwardRedirection`** so the fused loss executes inside the engine's `forward()`, arming the ZeRO-3 pre-forward hooks. The redirection is gated to ZeRO-3 / FSDP: on plain DDP it would wrap Liger's `torch.func` call inside `DDP.forward()` and error, so there we call `_compute_loss_liger` directly on the unwrapped model.

Also routes GRPO's inline gather through the shared `maybe_gather_lm_head_ctx` (was a duplicated block), per the repo's consistency guideline.

## Verified (2×H100)

`test_dpo_liger` (new, mirrors `test_grpo_liger`):

| backend | result |
|---|---|
| ddp | ✅ pass |
| zero2 | ✅ pass |
[...]

PR #7246, opened 2026-09-16
Redirect the liger loss through the distributed wrapper under DDP

### LINKED ISSUE #7245: DPO and KTO skip DDP gradient synchronization with use_liger_kernel=True
### Reproduction

`compute_loss` in `DPOTrainer` and `KTOTrainer` runs the loss on the **unwrapped** model, and only enters `_forward_redirection` for ZeRO-3 and FSDP:

```python
unwrapped_model = self.accelerator.unwrap_model(model)
if is_zero3 or self.is_fsdp_enabled:
    return self._forward_redirection(model, unwrapped_model, self._compute_loss, unwrapped_model, inputs, return_outputs)
return self._compute_loss(unwrapped_model, inputs, return_outputs)
```

Under plain DDP neither condition holds, so `DistributedDataParallel.forward()` never runs. `Reducer::prepare_for_backward()` is therefore never called, `expect_autograd_hooks_` stays false, and the reducer's autograd hooks return early without all-reducing. Every rank keeps its own gradients.

Needs liger-kernel installed, since `use_liger_kernel=True` is gated on it.

```python
# ddp_grad_sync.py
import os

import torch
import torch.distributed as dist
from datasets import Dataset
from transformers import AutoModelForCausalLM, AutoTokenizer, TrainerCallback

MODEL = "trl-internal-testing/tiny-Qwen3ForCausalLM"
USE_LIGER = os.environ.get("USE_LIGER", "1") == "1"
rank = int(os.environ["RANK"])

tokenizer = AutoTokenizer.from_pretrained(MODEL)
# Each rank gets different data, so unsynchronised gradients are guaranteed to differ.
dataset = Dataset.from_list(
    [
        {
            "prompt": f"rank {rank} question {i} about something",
            "chosen": f"a good answer from rank {rank}",
            "rejected": f"a bad answer from rank {rank}",
        }
        for i in range(32)
    ]
)
model = AutoModelForCausalLM.from_pretrained(MODEL, dtype=torch.bfloat16)
training_args = DPOConfig(
    output_dir="ddp_grad_sync_out",
    use_liger_kernel=USE_LIGER,
    max_steps=1,
    max_length=128,
    bf16=True,
    save_strategy="no",
    report_to="none",
)
trainer = DPOTrainer(model=model, args=training_args, train_dataset=dataset, processing_class=tokenizer)


class GradProbe(TrainerCallback):
    def on_pre_optimizer_step(self, args, state, control, **kwargs):
        unwrapped = trainer.accelerator.unwrap_model(trainer.model)
        n_differing, n_total, max_diff = 0, 0, 0.0
        for _, param in unwrapped.named_parameters():
            if param.grad is None:
                continue
            grad = param.grad.detach().float()
            gathered = [torch.zeros_like(grad) for _ in range(dist.get_world_size())]
            dist.all_gather(gathered, grad)
            diff = (gathered[0] - gathered[1]).abs().max().item()
            n_total += 1
            max_diff = max(max_diff, diff)
            if diff > 0:
                n_differing += 1
        if dist.get_rank() == 0:
            print(f"use_liger_kernel={USE_LIGER} params={n_total} differing={n_differing} max_abs_diff={max_diff:.3e}")


trainer.add_callback(GradProbe())
trainer.train()
```

```sh
USE_LIGER=1 torchrun --nproc_per_node=2 ddp_grad_sync.py
USE_LIGER=0 torchrun --nproc_per_node=2 ddp_grad_sync.py
```

### Outcome

```
use_liger_kernel=True  params=25 differing=25 max_abs_diff=1.265e-01
use_liger_kernel=False params=25 differing=0  max_abs_diff=0.000e+00
```

Same on `Qwen/Qwen3-0.6B`: 310/310 differing with liger, 0/310 without.

Expected `differing=0` in both cases. Instead an N-GPU DDP run is N independent single-GPU runs at `per_device_train_batch_size`, and the saved checkpoint is rank 0's. No error and no warning.

### Scope

- `trl/trainer/dpo_trainer.py` and `trl/trainer/kto_trainer.py`, both in `compute_loss`.
- Only with `use_liger_kernel=True` and plain DDP. Single GPU, FSDP and ZeRO-3 are fine. I did not check ZeRO-1/2.
- Introduced in #6372, which unwrapped the model to fix ZeRO-3. Before it, `compute_loss` passed the wrapped model and DDP was correct. Present in v1.10.0, v1.11.0 and v1.12.0.
- `tests/distributed/test_distributed.py::test_dpo` does parametrize `ddp`, but it never passes `--use_liger_kernel`, and `run_command` only asserts the process exits 0, which a silent desync does.

Both files already use the DDP-aware condition elsewhere: `kto_trainer.py:1322` and GRPO in #7077 both do `self.is_fsdp_enabled or model is not unwrapped_model`.

### System Info

- TRL: main @ b7df6a8f
- transformers: 5.17.0.dev0
- PyTorch: 2.13.0+cu130
- Python: 3.13.13
- 2x NVIDIA H100 80GB, plain DDP via torchrun

qgallouedec:
## What does this PR do?

Fixes #7245.

`compute_loss` in `DPOTrainer` and `KTOTrainer` entered `_forward_redirection` only for ZeRO-3 and FSDP, and otherwise ran the loss on the unwrapped model. Under plain DDP that means `DistributedDataParallel.forward()` never runs, so its reducer is never armed by `prepare_for_backward()`, the autograd hooks return early, and gradients are never all-reduced. Each rank keeps its own.

Measured on 2 GPUs, one step, different data per rank, comparing every parameter's gradient across ranks:

| | params differing | max abs diff |
|---|---|---|
| `use_liger_kernel=True`, before | 310 / 310 | 7.5e-02 |
| `use_liger_kernel=True`, after | 0 / 310 | 0.0 |
| `use_liger_kernel=False` | 0 / 310 | 0.0 |

Introduced in #6372, which unwrapped the model to fix ZeRO-3. Present in v1.10.0, v1.11.0 and v1.12.0.

The condition now matches the one `compute_ref_log_probs` already uses in both files, and that `GRPOTrainer` uses in #7077.

`is_zero3` is kept rather than folded in: under ZeRO-3 the model is a `DeepSpeedEngine`, so `model is not unwrapped_model` should already cover it, but deepspeed is xfailed against the transformers version I have, so I left the existing term alone rather than change a path I could not exercise. Worth simplifying separately.

---

> [!NOTE]
> **Medium Risk**
> Changes distributed training gradient synchronization for Liger-enabled DPO/KTO; fixes a silent DDP bug but touches core `compute_loss` paths used on every training step.
> 
> **Overview**
> Fixes incorrect multi-GPU training when **DPO** or **KTO** use `use_liger_kernel=True`. `compute_loss` now routes the Liger loss through `_forward_redirection` whenever the passed `model` is a distributed wrapper (`model is not unwrapped_model`), in addition to ZeRO-3 and FSDP—so DDP’s `forward()` runs and gradient all-reduce is armed.
> 
> Regression tests were added for **DPO** and **KTO** that mock a wrapped model and assert `compute_loss` calls `_forward_redirection`, matching the pattern already used for reference log-probs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e02eb31718adb75e152cf316bbf3adbe778bd172. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
Superseded. Splitting this in two, because the same fix does not work for both trainers.

KTO is #7247, verified: 310/310 parameters differing across ranks before, 0/310 after, and zero3/zero2/ddp all green.

DPO cannot take this change on main. The fused loss runs `torch.func.grad_and_value` internally, so running it inside `DistributedDataParallel.forward()` raises `NotImplementedError: Cannot access storage of TensorWrapper` on every rank. It turns a silent desync into a crash. It works once #7243 moves DPO onto the chunked path, so it goes there.

Tracked in #7245.

PR #7248, opened 2026-09-17
Read the MoE auxiliary loss coefficient from the model config

### LINKED ISSUE #7222: router_aux_loss_coef is silently ignored on MoE configs without output_router_logits
### Reproduction

`router_aux_loss_coef` silently does nothing on half of the MoE architectures, because whether a model counts as a MoE is inferred from a config field many MoE configs do not declare:

```python
# trl/trainer/sft_trainer.py
is_moe = getattr(text_config, "output_router_logits", None) is not None
self.aux_loss_enabled = is_moe and self.args.router_aux_loss_coef != 0.0
```

```python
from datasets import Dataset

dataset = Dataset.from_dict({"prompt": ["What is 2+2?"] * 8, "completion": [" Four."] * 8})
for model_id in ("hf-internal-testing/tiny-random-Qwen3MoeForCausalLM",
                 "hf-internal-testing/tiny-random-DeepseekV3ForCausalLM"):
    trainer = SFTTrainer(
        model=model_id,
        args=SFTConfig(output_dir="out", max_steps=2, logging_steps=1, report_to=[],
                       router_aux_loss_coef=0.001),
        train_dataset=dataset,
    )
    trainer.train()
    keys = {k for h in trainer.state.log_history for k in h}
    print(f"{model_id}: aux_loss logged = {'aux_loss' in keys}")
```

### Outcome

```
tiny-random-Qwen3MoeForCausalLM: aux_loss logged = True
tiny-random-DeepseekV3ForCausalLM: aux_loss logged = False
```

Both are MoE, both were asked for the same coefficient, one gets the term and the other does not. No warning either way.

On transformers main, 34 MoE configs declare `output_router_logits` and 30 do not. The second group includes DeepSeek-V2/V3/V3.2, GLM4-MoE, GLM4V-MoE, Qwen3-VL-MoE, Mistral4, dots1, Kimi-Linear and Step3p7. For all of those, setting `router_aux_loss_coef` is a no-op.

### Expected behavior

Either detect MoE from something every MoE has, e.g. `num_experts` / `num_local_experts` on the text config, or keep the current check and warn when `router_aux_loss_coef` is non-zero but the term cannot be applied. Silently ignoring the argument is the part to avoid: the run looks configured and is not.

### System info

trl 1.13.0.dev0, transformers 5.18.0.dev0, peft 0.21.0, accelerate 1.15.0, torch 2.13.0, one H100.

This issue is not open to external contribs

qgallouedec:
Fixes #7222.

`router_aux_loss_coef` was keyed on `output_router_logits` in the config, which was wrong in both directions: silently ignored on Qwen3-VL-MoE and GLM4V-MoE, a crash on Llama 4. Digging into it turned into a few transformers issues and PRs; the conclusion, and this PR, is that the config is the right marker once transformers keeps it honest.

**Principle.** The load-balancing loss is a training-time term, so the training lib owns it. What TRL needs from the model is two config attributes: `output_router_logits` for "this model returns its router logits", `router_aux_loss_coef` for the coefficient the architecture was trained with. Bias-balanced families (DeepSeek-V3, GLM4-MoE, Kimi, dots1, ...) declare no coefficient, so the term stays off on them without any list of models.

| | before | after |
|---|---|---|
| default coefficient | flat `0.001` for everyone | the architecture's own: 0.01 for OLMoE and FlexOlmo, 0.0001 for GLM4V-MoE, 0.0 for bias-balanced and dense models |
| Qwen3-VL-MoE, GLM4V-MoE | ignored | on, they declare a coefficient |
| DeepSeek-V3 and the bias-balanced families | `0.001` silently dropped | `0.0`, off on purpose |
| enabled when | config has `output_router_logits` | config has `output_router_logits` and the resolved coefficient is non-zero |
| explicit coefficient on a model without router logits (dense, Llama 4 on older transformers) | silently ignored, or a crash on `outputs.aux_loss` | `ValueError` at init |

Explicit `router_aux_loss_coef` still wins. The MoE block is byte-identical across SFT, GRPO, RLOO, DPO, KTO and AsyncGRPO.

**Why the config key is trustworthy now.** On transformers `main` it was wrong for 21 of 62 MoE classes. huggingface/transformers#48920 (16 causal MoEs plus DeepSeek-V2, HunYuan-V1-MoE, Mistral4, Llama 4, Cohere2-MoE) and huggingface/transformers#48885 (the VLM wrappers) make `hasattr(text_config, "output_router_logits")` agree with "returns `router_logits`" on all 62. Related: huggingface/transformers#48886 (output types), #48889 (Llama 4's dead fields), #48911.

**Not in this PR.** TRL still reads `outputs.aux_loss`, so an explicit coefficient on a bias-balanced model, which returns router logits and no loss, has nothing to add yet. Computing the term ourselves from `router_logits` is the follow-up, and the point of asking transformers for the logits.

Release note: bias-balanced MoEs that used to pick up the flat `0.001` now resolve to `0.0`; set `router_aux_loss_coef` explicitly to keep the term.

---

> [!NOTE]
> **Medium Risk**
> Changes default MoE training loss for many setups (bias-balanced MoEs drop the old 0.001 default) and adds a hard error when aux coef is set on non-MoE models.
> 
> **Overview**
> **MoE router auxiliary loss** no longer defaults to a flat `0.001` everywhere. `router_aux_loss_coef` is optional (`None`) on SFT, DPO, GRPO, KTO, RLOO, and AsyncGRPO configs; trainers resolve it from the model’s `text_config` when unset, and explicit values still override.
> 
> Auxiliary loss is enabled only when the config exposes `output_router_logits` **and** the resolved coefficient is non-zero—so bias-balanced MoEs with no declared coef stay off, while architectures that declare one (e.g. Qwen3 MoE) pick it up automatically. Setting `router_aux_loss_coef` on a model without router logits now raises **`ValueError` at init** instead of being ignored.
> 
> `SFTTrainer` additionally keeps `text_config.output_router_logits` and `text_config.router_aux_loss_coef` in sync with the resolved settings. New SFT tests cover architecture defaults, explicit override, and the dense-model error path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98d88b9502a278d26463b32c3af307f6c12c099b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


trl/experimental/async_grpo/async_grpo_trainer.py
```diff
@@ -1080,11 +1080,17 @@ def __init__(
...
-        # MoE load-balancing auxiliary loss, applied to Mixture-of-Experts models (no effect otherwise)
+        # MoE load-balancing auxiliary loss. `output_router_logits` in the config means the model returns its router
+        # logits; `router_aux_loss_coef` is the architecture's own coefficient, 0.0 when the config declares none.
         text_config = model.config.get_text_config()
-        is_moe = getattr(text_config, "output_router_logits", None) is not None
-        self.aux_loss_enabled = is_moe and args.router_aux_loss_coef != 0.0
-        self.router_aux_loss_coef = args.router_aux_loss_coef
+        coef = args.router_aux_loss_coef
+        self.router_aux_loss_coef = getattr(text_config, "router_aux_loss_coef", 0.0) if coef is None else coef
+        if coef and not hasattr(text_config, "output_router_logits"):
```
AmineDiro, line 1088: Just checking, `output_router_logits` is always an attribute from transfomers for MoEs ?

AmineDiro (approved):

trl/experimental/async_grpo/async_grpo_trainer.py
qgallouedec: Not today, no, and that's the reason this PR exists. I dug into it in huggingface/transformers#48886 and huggingface/transformers#48911.

Swept the 76 MoE architectures in transformers main: 
- 14 have no `output_router_logits` on the config at all
- 3 declare it but return nothing
- 17 models compute router logits in the gate and drop them before the output huggingface/transformers#48911
- 4 VLM wrappers read the flag off kwargs instead of the config huggingface/transformers#48885.

So the marker is wrong in both directions rn, which is why `router_aux_loss_coef` was silently ignored on Qwen3-VL-MoE and crashed on Llama 4. 
huggingface/transformers#48920 and huggingface/transformers#48885 make the attribute mean "returns router logits" on all of them. 

Until those land, the 2 VLM MoEs keep the term off by default (as before), and an explicit coefficient raises instead of being ignored, which is the better failure.

Issue #7244, opened 2026-09-16
Fused Triton kernels for the loss head


qgallouedec:
### Feature request

Ship fused Triton kernels for TRL's loss head, with the current torch implementations kept as fallback:

1. logprob + entropy from logits (`selective_log_softmax` + `entropy_from_logits`), used by GRPO/RLOO/DPO and ~15 other trainers
2. chunked CE returning loss + entropy + token accuracy (SFT `loss_type="chunked_nll"`)
3. `_ChunkedLogProbFunction` (KTO, async GRPO)

The loss head is the only hot path TRL owns: transformers already kernelizes the model internals (RMSNorm, SwiGLU, Linear, attention) and nothing on the Hub covers what comes after the decoder.

Measured on one H100, bf16, V=151936, H=4096, 8192 tokens:

- GRPO/RLOO/DPO logprob path: `selective_log_softmax` 3.50 ms + `entropy_from_logits` 9.33 ms = 12.8 ms per pass over `[8, 1024, 151936]`, plus 2.32 GiB of logits. A 40-line fused Triton kernel (online logsumexp + gather + entropy, forward only) does the same in 0.89 ms, 14x, and lands closer to the fp32 reference than the current bf16 path. GRPO runs this 2 to 4 times per step.
- SFT `chunked_nll` (the default), chunk 256, fwd+bwd: 167 ms, 13.1 GiB peak. For reference, liger FLCE from the Hub is 192 ms / 6.1 GiB, the naive materialized-logits CE is 59 ms / 19.9 GiB, and the bare `lm_head` GEMM fwd+bwd is 42 ms. A fused CE should sit around 60-80 ms at ~6 GiB, so 2.5x faster than today's default at half its memory.
- `_ChunkedLogProbFunction`: 495 ms fwd+bwd vs 58 ms for the naive path, 5.6 vs 10.6 GiB.

Reusing an existing kernel does not work: `kernels-community/liger-kernels` FLCE returns the loss only, while we need per-token logprobs, entropy, token accuracy, the `num_items_in_batch` reduction, temperature, `logit_scale` and `final_logit_softcapping`.

One thing to settle first: in-tree Triton vs a `torch-universal` kernel repo on the Hub (`build.toml` with `universal = true`, no compile step). The Hub gives rocm/xpu variants and no hard triton dependency, but needs network: `get_kernel(repo, version=N)` hits the Hub even when the snapshot is already cached, so a default-on path needs a warm-up and a fallback.

[...]

PR #7253, opened 2026-09-17
Use fused kernels for logprobs and entropy

kashif:
# What does this PR do?

Uses the fused logprob + entropy kernel maintained in `kernels/` and published to [trl-lib/trl-losses](https://huggingface.co/kernels/trl-lib/trl-losses). It fuses temperature scaling and row masking for GRPO, RLOO, DPO, KTO, and TPO, while keeping the current PyTorch implementation as the fallback.

The kernel source and publishing workflow landed in #7266. Related to #7244.

## Before submitting

## AI writing disclosure

## Who can review?

---

> [!NOTE]
> **Medium Risk**
> Touches shared loss utilities and multiple RL trainers’ forward/metric paths; incorrect masking, temperature, or kernel fallback could change training numerics, though PyTorch fallbacks and new tests mitigate this.
> 
> **Overview**
> Adds optional runtime loading of the Hub **`trl-lib/trl-losses`** fused kernel (when `kernels` is installed) for selective log-probabilities and entropy, with **cached one-shot load** (including failed/offline attempts) and **no Hub access under `torch.compile`**.
> 
> **`selective_log_softmax`** and new **`selective_log_softmax_and_entropy`** gain **`temperature`** and **`row_mask`** so trainers can mask completion tokens and scale logits inside the fused path instead of extra tensor ops. PyTorch paths remain the fallback when the kernel is missing or inputs are unsupported (e.g. non-CUDA/XPU).
> 
> **DPO, GRPO, KTO, RLOO, and TPO** are wired to the combined API where entropy is logged (or GRPO entropy bonus needs grads), and to masked **`selective_log_softmax`** for reference/KL paths—removing separate **`entropy_from_logits`** calls on shifted logits. **GRPO/RLOO** stop pre-dividing logits by temperature before logprob computation.
> 
> Docs and **`kernels/`** examples now describe auto-load for those trainers and use **`trust_remote_code=True`** for **`get_kernel`**. Tests cover Hub load caching, kernel dispatch vs fallback, and compile-safe behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f15178df970b143abe36b100816c0f9bcf16a1c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec:
The Hub repo `trl-lib/trl-losses` contains a README only. The publish workflow from #7266 failed on its single run with a 403 (token lacks `job.write` on `trl-lib`). So `get_kernel` raises.

qgallouedec:
Thanks, 2 commits since, neither touches the blocker. `trl-lib/trl-losses` is still `.gitattributes` + `README.md`, and `publish-kernels` has still run once and failed with the 403 (`missing permissions: job.write` on `trl-lib`). The fused path is unreachable everywhere, so this merges as a no-op.

CI being 11/11 green is not evidence either way: the CI env has no `kernels` installed (we should install kernels btw), so neither the fused path nor `tests/test_kernels.py` runs.

Fix the token permission and re-run the workflow, then this is reviewable end to end.

qgallouedec:
Correction! I was wrong about the Hub repo. All three backends are there. The 403 only means future kernel changes won't republish, not that this PR is a no-op.

But the fused path is still off, for a different reason, and it took a GPU run to see it because the failure is swallowed.

`_load_trl_loss_kernel` gates on `transformers.utils.is_kernels_available()`, which enforces `0.17.0 <= kernels < 0.18.0`. Our own `kernels` extra is `transformers[kernels]`, which pins `kernels>=0.16.0,<0.17`. So installing TRL the documented way gives 0.16.x, the gate returns False, and the kernel is never even tried, although `get_kernel` works fine on 0.16.1. Measured:

| installed kernels | `is_kernels_available()` | `_load_trl_loss_kernel()` |
|---|---|---|
| 0.16.1 (what our extra installs) | False | None, fused path off |
| 0.17.1 | True | loaded |

That gate is transformers' window for its own model kernels and says nothing about whether `get_kernel` works. Gating on the import instead would drop the coupling.

On 0.17.1 the path does work. A few numbers (Qwen-size logits `[8, 1024, 151936]`, bf16, H100):

| | fwd | fwd+bwd | peak |
|---|---|---|---|
| fused | 1.1 ms | 5.4 ms | 9.9 GiB |
| torch | 14.5 ms | 45.3 ms | 16.9 GiB |

Accuracy, both paths against an fp32 reference, `[2, 256, 151936]`, max abs error on log-probs:

| dtype | T | fused | torch |
|---|---|---|---|
| bf16 | 1.0 | 1.9e-06 | 6.3e-02 |
| bf16 | 0.7 | 3.8e-06 | 1.5e-01 |
| fp32 | 1.0 | 3.8e-06 | 1.9e-06 |

Two things still to fix:

- The swallow. I got `None` with no reason and had to bisect it by hand; a `logger.warning_once` with the exception would have said "kernels 0.16.1 outside the supported window" immediately.
- CI installs no `kernels`, so 11/11 green exercises neither the fused path nor `tests/test_kernels.py`. Adding it to the test extra would cover both, once the version window agrees with itself? I'm fine with doing it in another PR though

None of the above is really blocking here, so I'm approving

qgallouedec (approved):

PR #7259, opened 2026-09-17
Pass the FSDP2 ignored_params kwarg only on torch 2.7 or later

albertvillanova:
`prepare_fsdp` passes `ignored_params` to `fully_shard` unconditionally, but that keyword only exists from torch 2.7.0. On torch 2.5 and 2.6 the call raises `TypeError`, so FSDP2 is unusable there even though every other keyword TRL passes has been accepted since 2.5.

| keyword | 2.5 | 2.6 | 2.7 |
|---|---|---|---|
| `mesh`, `reshard_after_forward`, `mp_policy`, `offload_policy` | yes | yes | yes |
| `ignored_params` | no | no | yes |

This is a guard that was lost in transit rather than one that was never written. The block mirrors `accelerate`'s `fsdp_utils.py`, down to the comment about `MixedPrecisionPolicy`, and accelerate gates the same keyword:

```python
# `ignored_params` is only supported in torch >= 2.7.0
if is_torch_version(">=", "2.7.0") and fsdp2_plugin.ignored_modules is not None:
    fsdp2_kwargs["ignored_params"] = get_parameters_from_modules(...)
```

### No information is lost on older torch

The keyword defaults to `None`, so omitting it is equivalent to passing `None`, and `get_parameters_from_modules` returns an empty set when `ignored_modules` is `None`. The only case where old torch would change behaviour is a non-empty ignored set, and every path that reaches it already warns:

- From accelerate 1.11.0, `FullyShardedDataParallelPlugin.__post_init__` warns and sets `ignored_modules = None` on torch < 2.7.
- Below accelerate 1.11.0, the branch directly above this one already warns and sets `ignored_params = None`.

So no third warning is added here, which would mean two warnings for one situation.

The comparison uses `Version(torch.__version__)`, matching `Version(accelerate.__version__)` two lines above and matching accelerate's own semantics, which also compare the parsed version without stripping pre-release markers.

### Changes

- Build the `fully_shard` keywords as a dict and insert `ignored_params` only on torch >= 2.7.0
- Import `torch`, which the module did not bind (it imported only `torch.nn as nn`)

---

> [!NOTE]
> **Low Risk**
> Narrow compatibility fix for distributed FSDP2 wrapping; behavior on torch 2.7+ is unchanged, and older torch already treats ignored modules as unset via accelerate or existing warnings.
> 
> **Overview**
> **FSDP2 setup on PyTorch 2.5/2.6** no longer crashes when TRL wraps a model: `prepare_fsdp` was always passing `ignored_params` into `fully_shard`, but that argument exists only from **torch 2.7.0**, which caused a `TypeError` on earlier releases.
> 
> The FSDP2 path now builds a `fsdp2_kwargs` dict for the supported keywords and **adds `ignored_params` only when** `Version(torch.__version__) >= 2.7.0`, aligned with accelerate’s gating. A top-level **`import torch`** was added so the version check can run (the file previously only imported `torch.nn`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1678750232b315af44196de412ba6836eae889c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:
Merging this PR as a low-risk torch version compatibility fix: it stops passing a keyword that torch below 2.7 does not accept, with no behavior change on current torch.

PR #7260, opened 2026-09-17
Guard the torch 2.6 APIs used by FSDP2 and online DPO

albertvillanova:
Stacked on #7259, which guards the `ignored_params` keyword. This one guards the three remaining places where TRL reaches for a torch 2.6 API, so that FSDP2 and the online DPO vLLM path work on torch 2.5 as well.

The FSDP2 names became public in `torch.distributed.fsdp` only in 2.6, and `torch.accelerator` only exists from 2.6:

| API | 2.4 | 2.5 | 2.6 |
|---|---|---|---|
| `torch.distributed.fsdp.FSDPModule` | no | no | yes |
| `torch.distributed.fsdp.fully_shard`, `MixedPrecisionPolicy` | no | no | yes |
| `torch.accelerator` | no | no | yes |
| `torch.distributed._composable.fsdp.*` (private) | yes | yes | yes |

### Why all three and not just `torch.accelerator`

`FSDPModule` is imported at module level in `trl/models/utils.py`, and that module defines `prepare_fsdp` and `unwrap_model_for_generation`, which `dpo`, `grpo`, `rloo`, `kto`, `distillation` and `callbacks` all import. So on torch 2.5 those six cannot be imported at all, and guarding only the runtime call would leave the import failing while appearing to add tolerance.

### On the fallbacks

FSDP2 lived at `torch.distributed._composable.fsdp` before it was promoted, and that path still exports all three names today, so the pre-2.6 branch is the same implementation under its older name rather than a reimplementation.

For `torch.accelerator` the fallback is `"cuda"`, matching what `trl/models/activation_offloading.py` already does for the identical call. That sibling uses `hasattr(torch, "accelerator")`; this PR uses an explicit version comparison instead, since a `hasattr` check on a name that appeared in a known release is a version check in disguise.

### Changes

- Import `FSDPModule` from the private path before torch 2.6, next to the existing `accelerate` version-conditional import
- Same for `fully_shard` and `MixedPrecisionPolicy` at their use site
- Fall back to `"cuda"` for the accelerator type before torch 2.6

---

> [!NOTE]
> **Low Risk**
> Version-gated imports and a cuda fallback for device detection; behavior on torch 2.6+ is unchanged and pre-2.6 uses the prior private FSDP2 module path.
> 
> **Overview**
> Adds **PyTorch 2.5 compatibility** for code paths that previously assumed APIs introduced in **2.6**.
> 
> In `trl/models/utils.py`, **`FSDPModule`** is no longer imported unconditionally from the public `torch.distributed.fsdp` package. On torch **&lt; 2.6**, it imports from **`torch.distributed._composable.fsdp`** instead, and the FSDP2 **`fully_shard` / `MixedPrecisionPolicy`** imports at runtime follow the same split. That fixes import failures for trainers that pull in this module (DPO, GRPO, online DPO, etc.) when FSDP2 is used on 2.5.
> 
> In **`online_dpo_trainer.py`**, vLLM server communicator setup no longer calls **`torch.accelerator`** on older torch; it uses **`torch.accelerator.current_accelerator().type`** on **≥ 2.6** and defaults **`accelerator_type` to `"cuda"`** otherwise, matching the existing pattern in activation offloading.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa4eae1424d9646058ade8b2aacc37859c8c077d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:
Merging this PR as a low-risk torch version compatibility fix: it adds a fallback for torch below 2.6, with no behavior change on current torch.

Issue #7202, opened 2026-09-14
Audit the trl-internal-testing repos that have no generation script


albertvillanova:
Follow-up to #7186, where we agreed on the principle: every repo under `trl-internal-testing` should have its generation script in this repository, datasets included. A repo nobody can regenerate is a repo nobody can reason about.

Deleting a Hub repo is irreversible, and the set is not uniform, so this issue goes through them one by one. Nothing gets deleted until it is agreed here.

### What was checked

Every repo in the org cross-referenced against the generation scripts in `scripts/generate_tiny_models/` and `scripts/`, and against consumers in this repository, `transformers`, `peft`, `diffusers`, `trl-internal`, and public code on GitHub.

**75 model repos, 61 generation scripts. 14 models have no script.** They fall into three groups, and only the third is about deletion.

#### Used by our own tests, so they need a script rather than deletion

- [ ] `tiny-BloomForCausalLM` — 9 references in `tests/`, also used by `transformers`
- [ ] `tiny-Phi3ForCausalLM` — `tests/test_sft_trainer.py:2682` (note `-3` and `-3.5` do have scripts)

#### Used by another project, so keep, and agree who owns the script

- [ ] `tiny-random-LlamaForCausalLM` — `peft`, 27 references
- [ ] `tiny-Llama4ForCausalLM` — `peft`, 12 references
- [ ] `tiny-Qwen3_5ForConditionalGeneration` — `trl-internal`, 17 references
- [ ] `tiny-DbrxForCausalLM` — `transformers`

#### No consumer found anywhere, so candidates for deletion

- [ ] `dummy-GPT2-correct-vocab`
- [ ] `tiny-Qwen2AudioForConditionalGeneration`
- [ ] `tiny-DeepseekV4ForCausalLM`
- [ ] `tiny-VoxtralForConditionalGeneration`
- [ ] `tmp-tiny-DbrxForCausalLM` — the name suggests a leftover from a debugging session
- [ ] `tiny-Qwen2ForCausalLM-2.5-Coder` — generator removed in #7186
- [ ] `tiny-PaliGemmaForConditionalGeneration` — generator removed in #7186
- [ ] `tiny-Qwen3MoeForSequenceClassification` — generator removed in #7186

[...]

PR #7262, opened 2026-09-17
Stop depending on the tiny Bloom model in the chat template tests

albertvillanova:
`trl-internal-testing/tiny-BloomForCausalLM` is the last tiny model in this repository that cannot be given a generation script.

`transformers` uses it at `tests/models/auto/test_tokenization_auto.py:544`, in a test named `test_auto_tokenizer_loads_bloom_repo_without_tokenizer_class`, added by huggingface/transformers#42563, the tokenization refactor that removed `BloomTokenizer`. The repo's `tokenizer_config.json` still declares `"tokenizer_class": "BloomTokenizer"`, and the test asserts `AutoTokenizer` resolves it to a fast backend anyway. Its value is that its metadata is stale, so regenerating it with current `transformers` would destroy the thing being tested, and a generation script is exactly a recipe for producing current metadata.

Our own use of it needs none of that. All nine references want one property, a tokenizer with no chat template, and `tiny-GPT2LMHeadModel` has that property and does have a generation script.

### Changes

- `TestCloneChatTemplate`, four tests, eight references: use `tiny-GPT2LMHeadModel`
- `TestSupportsToolCalling`: drop the Bloom entry, since `tiny-GPT2LMHeadModel` sits on the next line and covers the same case

Nothing in those assertions is Bloom-specific. They check the cloned template's `eos_token`, the rendered prompt, and that the resized embeddings reach a multiple of 123 or 124. `model.vocab_size` is set generically by `resize_token_embeddings` in `PreTrainedModel`, not by either model class, and `clone_chat_template` uses `pad_to_multiple_of` for embedding resizing but never needs a pad token, which GPT-2 lacks and Bloom has.

Part of #7202, where the repo is proposed for transfer to `hf-internal-testing` rather than deletion.

---

> [!NOTE]
> **Low Risk**
> Test-only fixture changes with no production code or runtime behavior impact.
> 
> **Overview**
> Swaps the **no chat template** fixture in `TestCloneChatTemplate` from `trl-internal-testing/tiny-BloomForCausalLM` to `trl-internal-testing/tiny-GPT2LMHeadModel` across four tests (eight model/tokenizer loads). Assertions stay the same: cloned `eos_token`, embedding resize multiples, and rendered prompts.
> 
> In `TestSupportsToolCalling`, removes the Bloom parametrized case because GPT-2 already covers the same “no chat template → does not support tool calling” behavior.
> 
> This drops TRL’s dependency on the Bloom tiny repo, which is intentionally frozen for upstream `transformers` tokenizer tests and cannot be regenerated with a generation script.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81cafcc6759c8da4bbcb66581d94b65ce4117a9a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qgallouedec (approved):
thanks

AmineDiro (approved):

PR #7277, opened 2026-09-18
Move the code quality check to its own workflow

albertvillanova:
## What does this PR do?

`Check code quality` runs `pre-commit run --all-files`, which has nothing to do with the test suite, but it lives inside `tests.yml` and `tests-experimental.yml` and is therefore gated behind their `paths:` filters. That has two consequences.

It does not run when it should. A pull request that changes only `docs/source/` matches neither filter, so the whole workflow is skipped and the `doc-builder-style` hook never runs on the files it targets. #7177 is an example: its only checks were Bugbot, the doc build, trufflehog and the template check.

And it exists twice, byte for byte, in the two workflows.

This moves it into its own `quality.yml` with no paths filter, so it runs on every pull request and on every push to `main`, and removes both copies. The job name stays `Check code quality`, and no required status checks are configured on `main`, so nothing that refers to it by name breaks.

Part of #7169.

---

> [!NOTE]
> **Low Risk**
> CI-only refactor with no application code changes; the job behavior and name are unchanged.
> 
> **Overview**
> **Extracts the pre-commit “Check code quality” job into a dedicated `quality.yml` workflow** so it is no longer tied to the path-filtered `tests.yml` and `tests-experimental.yml` pipelines.
> 
> The new workflow runs on every pull request (non-draft) and on pushes to `main`, using the same Ubuntu + Python 3.12 setup, pre-commit cache, and `pre-commit run --all-files` step as before. The duplicate `check_code_quality` job is removed from both test workflows, so doc-only or other PRs that skip tests still get lint/format hooks (e.g. `doc-builder-style`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 644a393ba7521ab64a02f93e8c5f95168d9c81e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


albertvillanova:
Merging this PR as a CI refactoring, required as precondition by PRs #7281 and #7283.

